PR #343, opened 2018-04-27
feat: add tabbed navigation for development site

### LINKED ISSUE #288: Enable tabbed navigation per component
see  comp - end goal is we should be able to navigate from:
A. examples page with multiple examples of a component
B. Detail pages with documentation, accessibility, etc

This task should provide the mechanism to create multiple views per component and add tabbed interface to app-bar.

nicholasrice:
closes #288

### DIFF
--- a/packages/fast-development-site-react/src/components/site/action-bar.tsx
+++ b/packages/fast-development-site-react/src/components/site/action-bar.tsx
@@ -6,6 +6,7 @@ import { RouteComponentProps } from "react-router";
 import { ComponentViewTypes } from "./component-view";
 import { glyphExamples, glyphPage } from "@microsoft/fast-glyphs-msft";
 import ComponentViewToggle from "./component-view-toggle";
+import manageJss, { ComponentStyles, IJSSManagerProps, IManagedClasses } from "@microsoft/fast-jss-manager-react";
 
 export interface IActionBarProps extends RouteComponentProps<any> {
     /*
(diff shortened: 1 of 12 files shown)


packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -93,87 +95,125 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
...
                     {this.getPaneCollapseToggle()}
-                    {this.getSlotItems(this, ShellSlot.pane)}
+                    {this.getChildrenBySlot(this, ShellSlot.pane)}
                     <ul className={this.props.managedClasses.site_paneToc}>
-                        {this.getRootToc(this.props.children, SiteSlot.category, path, "/")}
+                        {this.getRootToc(this.props.children, SiteSlot.category, route.route, "/")}
                     </ul>
                 </ShellPane>
-                <ShellCanvas>
+                <ShellCanvas >
```
chrisdholt: nit: seems like an errant space

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -93,87 +95,125 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
...
             </ShellInfoBar>
         );
     }
 
     /**
      * Gets all of the potential routes as strings to be used to build the shell
      */
-    private getRoutes(items: JSX.Element, baseRoute: string, slot: string, routes?: IComponentRoutes[]): IComponentRoutes[] {
-        let currentRoutes: IComponentRoutes[] = routes;
+    private getRoutes(items: JSX.Element, baseRoute: string, slot: SiteSlot, routes?: IComponentRoute[]): IComponentRoute[] {
```
chrisdholt: Do we need to check that routes exist here?

packages/fast-development-site-react/src/components/site/action-bar.tsx
```diff
@@ -0,0 +1,78 @@
...
+                <div className={this.props.managedClasses.actionBar_componentViewToggles}>
+                    <ComponentViewToggle
+                        to={this.props.match.path}
+                        onClick={this.onComponentViewChangeCallback(ComponentViewTypes.detail)}
+                        label="View detail"
+                        current={this.isAriaCurrent(ComponentViewTypes.detail)}
+                        glyph={glyphPage}
+                    />
+                    <ComponentViewToggle
+                        to={this.props.match.path + `${ComponentViewTypes[ComponentViewTypes.examples]}/`}
```
janechu: could stick the this.props.match.path in with the backticks to do  \`${this.props.match.path}${ComponentViewTypes[ComponentViewTypes.examples]}/\`

packages/fast-development-site-react/src/components/site/index.tsx
nicholasrice: na, it is handled if it is undefined

PR #346, opened 2018-04-28
feat(markdown): add utility to convert markdown to msft component strings

### LINKED ISSUE #289: Migrate FW.Framework markdown-to-react utility
Convert markdown to react-msft components (where available, standard HTML where its not). Port from FW.Framework and create new package in fast-dna named *fast-markdown-react-msft*.

Add build step to create these files.

janechu:
Closes #289

### DIFF
--- a/packages/fast-markdown-msft-react/src/index.spec.ts
+++ b/packages/fast-markdown-msft-react/src/index.spec.ts
@@ -1,0 +1,94 @@
+import * as MarkdownIt from "markdown-it";
+import FASTMarkdownIt from "./index";
+
+const md: any = new MarkdownIt({
+    html: true,
+    linkify: true,
+    typographer: true,
+    xhtmlOut: true
+});
+
+md.use(FASTMarkdownIt);
+
+const newlineRegex: RegExp = /\r\n|\n/g;
+
+describe("Convert markdown to HTML", function(): void {
+
+    test("should convert a link to a hyperlink with a blank target", function(): void {
+        const markdown: string = "https://test.com";
+        expect(md.render(markdown).replace(newlineRegex, "")).toBe(
+            /* tslint:disable-next-line */
+            "<Typography tag=\"p\" typeLevel={2}><Hypertext text=\"https://test.com\" href=\"https://test.com\" target=\"_blank\">https://test.com</Hypertext></Typography>"
+        );
+    });
+
+    test("should convert a (c) to a copywrite symbol", function(): void {
+        const markdown: string = "(c)";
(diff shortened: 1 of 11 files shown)


packages/fast-markdown-msft-react/src/index.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    });
+
+    test("should convert a markdown unordered bullet list to a list", function(): void {
+        const markdown: string = "* list item";
+        expect(md.render(markdown).replace(newlineRegex, "")).toBe(
+            "<ul><li><Typography tag=\"p\" typeLevel={2}>list item</Typography></li></ul>"
+        );
+    });
+
+    test("should convert a markdown ordered list to a fluent web list", function(): void {
```
chrisdholt: Probably don't want FW here or throughout

packages/fast-markdown-msft-react/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    attrPush: (attributes: string[]) => void;
+}
+
+class FastMarkdownIt {
+    constructor(md: MarkdownIt) {
+        md.core.ruler.push("fast", (new CreateRule(md) as any));
+        md.renderer.rules.paragraph_open = function(): string {
+            return `<Typography tag="p" typeLevel={2}>`;
+        };
+        md.renderer.rules.heading_open = function(tokens: ITokens<IToken>, idx: number): string {
```
chrisdholt: If this is specific to FAST MSFT, should we not use the Heading component? If this isn't specific to MSFT, can you update the ReadMe language?

chrisdholt: Highlighted, it put it where it wanted. I'll keep an eye out for it.

packages/fast-markdown-msft-react/src/rule.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    constructor(md: MarkdownIt) {
+        return (this.replace as any);
+    }
+
+    public replace = (state: IState): void => {
+        this.renderComponents(state);
+    }
+
+    private renderComponents = (state: IState): void => {
```
nalogiudice: Can you leave a comment somewhere in this file explaining what it is doing?

janechu: resolved

janechu:
@awentzel its specific to the MSFT exported components, this can only be used if we are importing the MSFT react package because of the differences in markdown between our react packages. For example, there is a Heading in MSFT but not in base

PR #353, opened 2018-04-30
docs: update README.md to be correct

awentzel:

### DIFF
--- a/packages/fast-jss-utilities/README.md
+++ b/packages/fast-jss-utilities/README.md
@@ -1,2 +1,2 @@
 # FAST - JSS Utilities
-Utilities specific to JSS (JavaScript Style Sheets) packages intended to be shared across different frameworks. For example, React, Angular, or Vue.
+This package (@microsoft/fast-jss-utilities) is a collection of utilities intended to be used with JSS (JavaScript Style Sheets) projects.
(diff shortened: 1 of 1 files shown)


packages/fast-jss-utilities/README.md
```diff
@@ -1,2 +1,2 @@
-# FAST components styles - MSFT
-`@microsoft/fast-components-styles-msft` is a collection of JSS (JavaScript Style Sheets) objects and style utilities that power msft components in the FAST ecosystem. The component styles are intended to be compiled with [JSS](https://github.com/cssinjs/jss) and used with base components such as`@microsoft/fast-components-base-react`.
+# FAST - JSS Utilities
+Utilities specific to JSS (JavaScript Style Sheets) packages intended to be shared across different frameworks. For example, React, Angular, or Vue.
```
chrisdholt: I think we can *keep* and get rid of a few things from before:

`@microsoft/fast-jss-utilities` is a collection of utilities intended to be used with JSS (JavaScript Style Sheets) projects.

Doesn't have to be the above, not sure we need to mention anything about frameworks as it's more directly a set of utilities for JSS. Frameworks don't necessarily enter into the picture.

Issue #355, opened 2018-05-01
Style text field with initial JSS


nalogiudice:
Style the form generator text field using JSS manager to have more of a Microsoft style and layout.
This will be a good first test, and then other elements can be styled.

PR #356, opened 2018-05-01
feat(form generator): adds styling scaffolding and JSS for text field

nalogiudice:
Closes [#355](https://github.com/Microsoft/fast-dna/issues/355)


packages/fast-form-generator-react/package.json
```diff
@@ -88,6 +88,10 @@
     "yargs": "^11.0.0"
   },
   "dependencies": {
-    "@microsoft/fast-jest-snapshots-react": "^1.0.0"
+    "@microsoft/fast-components-class-name-contracts-base": "^1.0.0",
+    "@microsoft/fast-glyphs-msft": "^1.0.0",
+    "@microsoft/fast-jest-snapshots-react": "^1.0.0",
+    "@microsoft/fast-jss-manager": "^1.0.0",
+    "@microsoft/fast-jss-utilities": "^1.0.0"
```
janechu: you can remove fast-jest-snapshots-react from dependencies, if anything they should be in devDependencies, can also remove fast-glyphs-msft since we will be using just the pseudo elements for styling.

packages/fast-form-generator-react/src/class-name-contracts/form-item.select.class-name-contract.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Select class name contract
+ */
+export default interface IFormItemSelectClassNameContract {
+    form_item_select_wrapper: string;
```
janechu: We can just call the wrapper formItemSelect also we can change all of these to formItemSelect_elementName because I believe the intent is that the formItemSelect which is scoped to this file is the block and the HTML element you're referencing should be the element in the BEM syntax

packages/fast-form-generator-react/src/styles/form-item.select.style.ts
```diff
@@ -0,0 +1,64 @@
+import { Direction, ellipsis, localizeSpacing, toPx } from "@microsoft/fast-jss-utilities";
```
janechu: Not a comment on this line, but I would not put these in their own folder, the can just be in src/form/ with the rest of the files

packages/fast-form-generator-react/package.json
nalogiudice: Yep - whoops I preemptively added the glyphs package. Will fo on the other.

packages/fast-form-generator-react/src/styles/form-item.select.style.ts
nalogiudice: ok

packages/fast-form-generator-react/package.json
```diff
@@ -54,6 +54,7 @@
   "devDependencies": {
     "@microsoft/fast-components-react-base": "^1.0.0",
     "@microsoft/fast-development-site-react": "^1.0.0",
+    "@microsoft/fast-jest-snapshots-react": "^1.0.0",
     "@microsoft/fast-jss-manager-react": "^1.0.0",
```
janechu: need to add fast-jss-manager-react to peer dependencies

janechu: also missing dependency for the fast-component-react-base in dev and peer

packages/fast-form-generator-react/src/form/form-item.select.style.ts
```diff
@@ -0,0 +1,70 @@
+import { Direction, ellipsis, localizeSpacing, toPx } from "@microsoft/fast-jss-utilities";
+
+function applypaddingStyle(): any {
+    return {
+        padding: localizeSpacing(Direction.ltr)(`${toPx(8)} ${toPx(24)} ${toPx(8)} ${toPx(8)}`)
```
chrisdholt: I think we'll want direction to be compiled at runtime instead of hard coded, correct?

nalogiudice: This was only done because of code climate issue - I can change back @janechu / @chrisdholt

chrisdholt: No - just talked with @janechu on this; can you create a task to pass this from a context so that this gets compiled correctly at runtime? No need to change at this point in time, but we'll want to go back and update.

packages/fast-form-generator-react/src/form/form-item.select.tsx
```diff
@@ -1,6 +1,14 @@
 import * as React from "react";
 import IFormItemCommon from "./form-item";
+import { default as SelectStyles } from "./form-item.select.style";
```
janechu: you can actually just import styles, there's no need for namespacing because we aren't importing multiple styles into this file

packages/fast-form-generator-react/src/form/form-item.select.style.ts
```diff
@@ -0,0 +1,70 @@
...
+        "&:before": {
+            right: toPx(15),
+            transform: "rotate(45deg)"
+        },
+        "&:after": {
+            right: toPx(22),
+            transform: "rotate(-45deg)"
+        }
+    },
+    form_item_select_input: {
```
janechu: these names should all match eg formItemSelect_input and remove _wrapper on the object key above

packages/fast-form-generator-react/src/form/form-item.select.style.ts
```diff
@@ -0,0 +1,70 @@
+import { Direction, ellipsis, localizeSpacing, toPx } from "@microsoft/fast-jss-utilities";
+
+function applypaddingStyle(): any {
+    return {
+        padding: localizeSpacing(Direction.ltr)(`${toPx(8)} ${toPx(24)} ${toPx(8)} ${toPx(8)}`)
+    };
+}
+
+const styles: any = {
```
janechu: this should not be any, see fast-components-styles-msft/src/checkbox/index.ts for example, should be something like `ComponentStyles<ISelectClassNameContract, IDesignSystem>`

nalogiudice: Good catch - updating

packages/fast-form-generator-react/src/form/form-item.select.style.ts
nalogiudice: Well, not we know how this didn't error

nalogiudice:
Closing to see if new branch resolves build issue

PR #362, opened 2018-05-03
feat(form): add the form generator to the development site

### LINKED ISSUE #291: Wire schema-form-generator to active component on the Examples page
On the page with numerous examples, I should be able to select any example and make changes via the schema form generator.

janechu:
Closes #291

### DIFF
--- a/packages/fast-development-site-react/src/components/site/action-bar.tsx
+++ b/packages/fast-development-site-react/src/components/site/action-bar.tsx
@@ -4,9 +4,9 @@ import { IDevSiteDesignSystem } from "../design-system";
 import { Link, withRouter } from "react-router-dom";
 import { RouteComponentProps } from "react-router";
 import { ComponentViewTypes } from "./component-view";
-import { glyphExamples, glyphPage } from "@microsoft/fast-glyphs-msft";
+import { glyphBuildingblocks, glyphExamples, glyphPage } from "@microsoft/fast-glyphs-msft";
 import ComponentViewToggle from "./component-view-toggle";
-import manageJss, { ComponentStyles, IJSSManagerProps, IManagedClasses } from "@microsoft/fast-jss-manager-react";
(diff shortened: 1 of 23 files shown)


janechu:
Just a note, I have removed a few dependencies that cause circular dependencies in our packages. For example we cant rely on fast-components-react-base because that is used by fast-components-react-msft and if the development site uses the form generator which uses the base, it will become circular.

packages/fast-development-site-react/src/components/site/action-bar.tsx
```diff
@@ -15,30 +15,87 @@ export interface IActionBarProps extends RouteComponentProps<any> {
...
+    return {
+        position: "relative",
+        height: "40px",
+        border: "none",
+        background: "none",
+        padding: "0",
+        "& span": {
+            width: "16px",
+            height: "16px",
+            marginRight: "5px",
```
nalogiudice: `toPx()` ?

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ICSSRules,
+    IJSSManagerProps,
+    IManagedClasses
+} from "@microsoft/fast-jss-manager-react";
+import { toPx } from "@microsoft/fast-jss-utilities";
+import devSiteDesignSystemDefaults, { IDevSiteDesignSystem } from "../design-system";
+
+export interface IComponentWrapperManagedClasses {
+    componentWrapper: string;
+    componentWrapper__active: string;
```
nalogiudice: I thought we were only doing a single _ ?

packages/fast-development-site-react/src/components/site/action-bar.tsx
janechu: fixed

Issue #304, opened 2018-04-20
MSFT checkbox styling does not work across browsers 


chrisdholt:
### Expected behavior 
Checkbox looks like a MSFT checkbox on all browsers.

### Actual behavior
Checkbox inherits the native styles of whatever browser it is opened in. It only appears similar to a MSFT checkbox in MSFT browsers (where MSFT design language is being applied).

### Steps to reproduce the problem
Boot fast-components-react-msft in chrome, firefox, safari and notice that the checkbox appears to be a native checkbox.
### Specifications (hardware, software, configurations)
* Include device make, model, OS version: 
* Include browser and version: 
* Include screen orientation: 
* Include screen size: 
* Include page URL: 
* Include localization: 

*Please attach screenshots or video of the problem for quickest resolution*

nalogiudice:
Closed with [#371](https://github.com/Microsoft/fast-dna/pull/371)

PR #371, opened 2018-05-04
feat(checkbox): update styling and incorrect states

nalogiudice:
<body>

- Update styles for cross-browser support
- Fixes incorrect state handler

<footer>
Closes [#304](https://github.com/Microsoft/fast-dna/issues/304)
 

### DIFF
--- a/packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
+++ b/packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
@@ -6,7 +6,7 @@ exports[`checkbox checkbox: 0 1`] = `
 >
   <input
     className="checkbox-input"
-    defaultChecked={undefined}
+    defaultChecked={false}
     disabled={undefined}
     onChange={[Function]}
     type="checkbox"
@@ -23,7 +23,7 @@ exports[`checkbox checkbox: 1 1`] = `
 >
   <input
     className="checkbox-input"
-    defaultChecked={undefined}
+    defaultChecked={false}
     disabled={undefined}
     onChange={[Function]}
     type="checkbox"
@@ -40,7 +40,7 @@ exports[`checkbox checkbox: 2 1`] = `
 >
   <input
     className="checkbox-input"
-    defaultChecked={undefined}
+    defaultChecked={false}
     disabled={undefined}
     onChange={[Function]}
     type="checkbox"
@@ -57,7 +57,7 @@ exports[`checkbox checkbox: 3 1`] = `
 >
   <input
     className="checkbox-input"
-    defaultChecked={undefined}
+    defaultChecked={false}
     disabled={undefined}
     onChange={[Function]}
     type="checkbox"
@@ -108,7 +108,7 @@ exports[`checkbox checkbox: 6 1`] = `
 >
   <input
     className="checkbox-input"
-    defaultChecked={undefined}
+    defaultChecked={false}
     disabled={true}
     onChange={[Function]}
     type="checkbox"
--- a/packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
+++ b/packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
@@ -5,10 +5,10 @@ exports[`checkbox checkbox: 0 1`] = `
   className="checkbox"
 >
   <input
-    checked={undefined}
     className="checkbox-input"
+    defaultChecked={undefined}
     disabled={undefined}
-    onChange={undefined}
+    onChange={[Function]}
     type="checkbox"
   />
   <span
@@ -22,10 +22,10 @@ exports[`checkbox checkbox: 1 1`] = `
   className="checkbox"
 >
   <input
-    checked={undefined}
     className="checkbox-input"
+    defaultChecked={undefined}
     disabled={undefined}
-    onChange={undefined}
+    onChange={[Function]}
     type="checkbox"
   />
   <span
@@ -39,10 +39,10 @@ exports[`checkbox checkbox: 2 1`] = `
   className="checkbox"
 >
   <input
-    checked={undefined}
     className="checkbox-input"
+    defaultChecked={undefined}
     disabled={undefined}
-    onChange={undefined}
+    onChange={[Function]}
     type="checkbox"
   />
   <span
@@ -56,10 +56,61 @@ exports[`checkbox checkbox: 3 1`] = `
   className="checkbox"
 >
   <input
-    checked={undefined}
     className="checkbox-input"
+    defaultChecked={undefined}
     disabled={undefined}
-    onChange={undefined}
+    onChange={[Function]}
+    type="checkbox"
+  />
+  <span
+    className="checkbox-label"
+  />
+</label>
+`;
+
+exports[`checkbox checkbox: 4 1`] = `
+<label
+  className="checkbox"
+>
+  <input
+    className="checkbox-input"
+    defaultChecked={true}
+    disabled={undefined}
+    onChange={[Function]}
+    type="checkbox"
+  />
+  <span
+    className="checkbox-label"
+  />
+</label>
+`;
+
+exports[`checkbox checkbox: 5 1`] = `
+<label
+  className="checkbox"
+>
+  <input
+    className="checkbox-input"
+    defaultChecked={true}
+    disabled={undefined}
+    onChange={[Function]}
+    type="checkbox"
+  />
+  <span
+    className="checkbox-label"
+  />
+</label>
+`;
+
+exports[`checkbox checkbox: 6 1`] = `
+<label
+  className="checkbox checkbox-disabled"
+>
+  <input
+    className="checkbox-input"
+    defaultChecked={undefined}
+    disabled={true}
+    onChange={[Function]}
     type="checkbox"
   />
   <span
--- a/packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
+++ b/packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
@@ -5,7 +5,9 @@ exports[`checkbox checkbox: 0 1`] = `
   className="checkbox"
 >
   <input
+    checked={false}
     className="checkbox-input"
+    disabled={null}
     onChange={[Function]}
     type="checkbox"
   />
@@ -20,7 +22,9 @@ exports[`checkbox checkbox: 1 1`] = `
   className="checkbox"
 >
   <input
+    checked={false}
     className="checkbox-input"
+    disabled={null}
     onChange={[Function]}
     type="checkbox"
   />
@@ -35,7 +39,9 @@ exports[`checkbox checkbox: 2 1`] = `
   className="checkbox"
 >
   <input
+    checked={false}
     className="checkbox-input"
+    disabled={null}
     onChange={[Function]}
     type="checkbox"
   />
@@ -50,7 +56,9 @@ exports[`checkbox checkbox: 3 1`] = `
   className="checkbox"
 >
   <input
+    checked={false}
     className="checkbox-input"
+    disabled={null}
     onChange={[Function]}
     type="checkbox"
   />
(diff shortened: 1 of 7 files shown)


packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -5,51 +5,52 @@ import { CheckboxHTMLTags, CheckboxProps, ICheckboxHandledProps, ICheckboxManage
...
      */
-    public applyIndeterminateState(): any {
-         if (this.props.indeterminate !== undefined && this.inputRef.current) {
-            this.inputRef.current.indeterminate = this.props.indeterminate;
-        }
+    constructor(props: CheckboxProps) {
+        super(props);
+
+        this.state = {
+            checked: this.props.checked
```
janechu: The prop `checked` is optional but the state checked is required, therefore its possible that state checked can be undefined, can you make sure that this is always a boolean?

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -66,8 +67,8 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
                     className={get(this.props, "managedClasses.checkbox_input")}
                     type="checkbox"
                     ref={this.inputRef}
-                    onChange={this.props.onChange}
-                    checked={this.props.checked}
+                    onChange={this.handleCheckboxChange}
```
janechu: What will happen if I want to know when this has changed? I can no longer pass an onChange here because its being controlled by the checkbox.

nalogiudice: Adding back in

packages/fast-components-react-base/src/checkbox/checkbox.tsx
nalogiudice: Roger

packages/fast-components-react-base/src/checkbox/checkbox.tsx
chrisdholt: Looks like you need to update jest tests after adding this back in.

nalogiudice: Should be good now.

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_script:
   - ./cc-test-reporter before-build
 script:
   - lerna run tslint
-  - lerna run unit-tests --maxWorkers=4
+  - travis_wait lerna run unit-tests --maxWorkers=4
```
awentzel: Let's remove this as it's not related to the PR changes. I have a PR already up too that is configuring travis for publishing and it can go in there.

nalogiudice: Ok. It did finally allow this test to past.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +90,36 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+    private get tag(): string {
+        return CheckboxHTMLTags[this.props.tag] || CheckboxHTMLTags.label;
+    }
+
+    /**
+     * Apply indeterminate state to items that are indeterminate.
+     * This method should be called after render because it relies on element references.
+     */
+    private applyIndeterminateState(): void {
+         if (this.props.indeterminate !== undefined && this.inputRef.current && this.state.checked) {
```
nicholasrice: `!isUndefined(this.props.indeterminate)` for consistency

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -3,53 +3,56 @@ import * as ReactDOM from "react-dom";
...
+        this.inputRef = React.createRef();
     }
 
     /**
-     * Called the when a component is mounted
+     * React life-cycle method
      */
     public componentDidMount(): void {
         this.applyIndeterminateState();
+        if (!isUndefined(this.props.checked)) {
```
nicholasrice: You're already setting state in the constructor - why do you need to call setState in `componentDidMount`?

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +90,36 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+    private applyIndeterminateState(): void {
+         if (this.props.indeterminate !== undefined && this.inputRef.current && this.state.checked) {
+            this.inputRef.current.indeterminate = this.props.indeterminate;
+        }
+    }
+
+    /**
+     * Handles onChange as a controlled component
+     */
+    private handleCheckboxChange = (e: any): void => {
```
nicholasrice: can you type the `e` param accurately? I think it should be `React.ChangeEvent<HTMLInputElement>`

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -66,8 +69,8 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
                     className={get(this.props, "managedClasses.checkbox_input")}
                     type="checkbox"
                     ref={this.inputRef}
-                    onChange={this.props.onChange}
-                    checked={this.props.checked}
```
nicholasrice: why are you removing the `checked` property in the render method? It seems now that the `checked` prop never gets used except to set `defaultChecked` - is that intentional?

packages/fast-components-react-base/src/checkbox/checkbox.tsx
nalogiudice: Sure

packages/fast-components-react-base/src/checkbox/checkbox.tsx
nalogiudice: Removing - not sure why this was here.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
nalogiudice: Originally I was just coming in to update the styling cross-browser, but then noticed this page errored due to controlled/uncontrolled state handling. I've re-wired it again, I believe it handles both states now.

nicholasrice: I still don't understand why you removed the `checked` prop though - this means that the checkbox cannot be manually assigned a value

nalogiudice: It's still there. The `defaultChecked` is for the controlled value, while the `checked` is for the uncontrolled.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +87,62 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
     }
+
+    /**
+     * Generates the correct attributes
+     */
+    private generateAttributes(): object {
+        const attributes: object = {};
+
+        /* tslint:disable:no-string-literal */
+        if (this.props.disabled) {
```
nicholasrice: you should just put this in the render method: `disabled={this.props.disabled || null}`

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +87,62 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+            attributes["disabled"] = "disabled";
+        }
+
+        // If the author is not setting the checked state, we should let the checkbox manage itself
+        if (!isUndefined(this.props.checked)) {
+            if (this.props.onChange) {
```
nicholasrice: It seems unintuitive to me to respect the `checked` prop *only* when `onChange` is passed. As an author, I would expect passing `checked` would *always* set the checked property, regardless of if I'm handling the change callback.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +87,62 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+                attributes["checked"] = !!this.props.checked;
```
nicholasrice: you shouldn't need to use string accessors anymore - this was a work-around that stemmed from not having accurate typings in the other repository.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +87,62 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
 
         return super.generateClassNames(classes);
     }
+
+    /**
+     * Generates the correct attributes
+     */
+    private generateAttributes(): object {
```
nicholasrice: I'd like to get away from the "generateAttributes" pattern - it was mainly implemented as a way to work around inaccurate typings and it was always a bit confusing as to when an attribute would be added in the render function and when it would be added via the `generateAttributes` method. I don't think you'll really need this though, with the other changes.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
nalogiudice: Ok - made some changes. Should be good.

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +88,34 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+         if (!isUndefined(this.props.indeterminate) && this.inputRef.current && this.state.checked) {
+            this.inputRef.current.indeterminate = this.props.indeterminate;
+        }
+    }
+
+    /**
+     * Handles onChange as a controlled component
+     */
+    private handleCheckboxChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+        this.setState({checked: !this.state.checked});
```
nicholasrice: The only issue now is if an author sets the `checked` property, and the component is clicked by the user, the property is then inconsistent with the state of the UI.

Props should be the single source of truth - if an author sets the checked prop, the UI needs to represent that value until the author changes it. It should not update internally if it is being controlled from the outside.

janechu: +1, this is the difference between controlled and uncontrolled

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +88,34 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+
+        if (this.props.onChange) {
+            this.props.onChange(e);
```
janechu: can this be the targets value?

packages/fast-components-react-base/src/checkbox/__snapshots__/checkbox.spec.ts.snap
```diff
@@ -56,14 +65,77 @@ exports[`checkbox checkbox: 3 1`] = `
   className="checkbox"
 >
   <input
-    checked={undefined}
-    className="checkbox-input"
-    disabled={undefined}
-    onChange={undefined}
+    checked={false}
+    className="checkbox_input"
```
janechu: @nicholasrice any idea why this isn't hashing? should it or should it not be?

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -87,6 +98,36 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
...
+     * Handles onChange as a controlled component
+     */
+    private handleCheckboxChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+        if (isUndefined(this.props.checked)) {
+            this.setState({checked: !this.state.checked});
+        }
+
+        if (this.props.onChange) {
+            this.props.onChange(e);
+        }
```
janechu: there should be a check when props are updated to also change the state since checked is based only on state, you can use this lifecycle hook: https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

PR #374, opened 2018-05-05
chore(build): stream all processes on travis and max workers for jest in each package

chrisdholt:


.travis.yml
```diff
@@ -16,14 +16,8 @@ before_script:
...
 script:
-  - lerna run tslint
-  - lerna run unit-tests --maxWorkers=4
-  - lerna run build:production
-  - lerna run build:dist
-  - lerna run build:app
-  - lerna run build:interface
-  - lerna run build:chrome
-  - lerna run eslint
+  - lerna run test --stream
```
nalogiudice: Why was it decided to reset this? You'll need to travis-wait to fix the timeout issue:
`- travis_wait lerna run test --stream`

chrisdholt: finished in 10 minutes (60% faster). 

Moved workers to each package.json and streaming the entire thing.

nalogiudice: Ok I'll rebase and see if this works on my PR

PR #398, opened 2018-05-09
feat(textfield) : add new component

### LINKED ISSUE #176: Add text field component
### Expected behavior 
Migrate https://fluentweb.com/components/text-field to FAST-DNA 

- Ensure style is maintained
- Ensure behavior is maintained
- Ensure accessibility is maintained

For testing follow https://github.com/Microsoft/fast-dna/wiki/the-process

emathiasen:
Closes #176

### DIFF
--- a/packages/fast-components-react-base/src/text-field/text-field.props.ts
+++ b/packages/fast-components-react-base/src/text-field/text-field.props.ts
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { IManagedClasses, ITextfieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { IManagedClasses, ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 
 export enum TextFieldType {
    date = "date",
@@ -17,7 +17,7 @@ export enum TextFieldType {
    week = "week"
 }
 
-export interface ITextfieldHandledProps {
+export interface ITextFieldHandledProps {
     /**
      * The disabled state
      */
@@ -31,7 +31,7 @@ export interface ITextfieldHandledProps {
     /**
      * Placeholder Text for input field
      */
-    placeholder: string;
+    placeholder?: string;
 
     /**
      * The text field input type
@@ -44,6 +44,6 @@ export interface ITextfieldHandledProps {
     value?: string;
 }
 
-export interface ITextfieldUnhandledProps extends React.HTMLAttributes<HTMLInputElement> {}
-export interface ITextfieldManagedClasses extends IManagedClasses<ITextfieldClassNameContract> {}
-export type TextfieldProps = ITextfieldHandledProps & ITextfieldUnhandledProps & ITextfieldManagedClasses;
+export interface ITextFieldUnhandledProps extends React.HTMLAttributes<HTMLInputElement> {}
+export interface ITextFieldManagedClasses extends IManagedClasses<ITextFieldClassNameContract> {}
+export type TextFieldProps = ITextFieldHandledProps & ITextFieldUnhandledProps & ITextFieldManagedClasses;
--- a/packages/fast-components-react-base/src/text-field/text-field.props.ts
+++ b/packages/fast-components-react-base/src/text-field/text-field.props.ts
@@ -24,11 +24,6 @@ export interface ITextFieldHandledProps {
     disabled?: boolean;
 
     /**
-     * The error state
-     */
-    error?: boolean;
-
-    /**
      * Placeholder Text for input field
      */
     placeholder?: string;
@@ -37,11 +32,6 @@ export interface ITextFieldHandledProps {
      * The text field input type
      */
     type?: TextFieldType;
-
-    /**
-     * The text field value
-     */
-    value?: string;
 }
 
 export interface ITextFieldUnhandledProps extends React.HTMLAttributes<HTMLInputElement> {}
--- a/packages/fast-components-react-base/src/text-field/text-field.props.ts
+++ b/packages/fast-components-react-base/src/text-field/text-field.props.ts
@@ -1,0 +1,49 @@
+import * as React from "react";
+import { IManagedClasses, ITextfieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export enum TextFieldType {
+   date = "date",
+   email = "email",
+   hidden = "hidden",
+   month = "month",
+   number = "number",
+   password = "password",
+   range = "range",
+   search = "search",
+   tel = "tel",
+   text = "text",
+   time = "time",
+   url = "url",
+   week = "week"
+}
+
+export interface ITextfieldHandledProps {
+    /**
+     * The disabled state
+     */
+    disabled?: boolean;
+
+    /**
+     * The error state
+     */
+    error?: boolean;
+
+    /**
+     * Placeholder Text for input field
+     */
+    placeholder: string;
+
+    /**
+     * The text field input type
+     */
+    type?: TextFieldType;
+
+    /**
+     * The text field value
+     */
+    value?: string;
+}
+
+export interface ITextfieldUnhandledProps extends React.HTMLAttributes<HTMLInputElement> {}
+export interface ITextfieldManagedClasses extends IManagedClasses<ITextfieldClassNameContract> {}
+export type TextfieldProps = ITextfieldHandledProps & ITextfieldUnhandledProps & ITextfieldManagedClasses;
--- a/packages/fast-components-react-base/src/text-field/text-field.schema.json
+++ b/packages/fast-components-react-base/src/text-field/text-field.schema.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json-schema.org/schema#",
-    "title": "Text Field",
+    "title": "Text field",
     "description": "A Text Field component's schema definition.",
     "type": "object",
     "id": "text-field",
@@ -11,7 +11,7 @@
             "type": "boolean"
         },
         "type": {
-            "title": "Text Field type",
+            "title": "Input type",
             "type": "string",
             "enum": [
                 "date",
@@ -30,7 +30,7 @@
             ]
         },
         "placeholder": {
-            "title": "Text Field placeholder",
+            "title": "Placeholder",
             "type": "string"
         }
     }
--- a/packages/fast-components-react-base/src/text-field/text-field.schema.json
+++ b/packages/fast-components-react-base/src/text-field/text-field.schema.json
@@ -6,5 +6,40 @@
     "id": "text-field",
     "children": true,
     "properties": {
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "error": {
+            "title": "Error",
+            "type": "boolean"
+        },
+        "type": {
+            "title": "Text Field type",
+            "type": "string",
+            "enum": [
+                "date",
+                "email",
+                "hidden",
+                "month",
+                "number",
+                "password",
+                "range",
+                "search",
+                "tel",
+                "text",
+                "time",
+                "url",
+                "week"
+            ]
+        },
+        "placeholder": {
+            "title": "Text Field placeholder",
+            "type": "string"
+        },
+        "value" : {
+            "title": "Text Field value",
+            "type": "string"
+        }
     }
 }
--- a/packages/fast-components-react-base/src/text-field/text-field.schema.json
+++ b/packages/fast-components-react-base/src/text-field/text-field.schema.json
@@ -10,10 +10,6 @@
             "title": "Disabled",
             "type": "boolean"
         },
-        "error": {
-            "title": "Error",
-            "type": "boolean"
-        },
         "type": {
             "title": "Text Field type",
             "type": "string",
@@ -35,10 +31,6 @@
         },
         "placeholder": {
             "title": "Text Field placeholder",
-            "type": "string"
-        },
-        "value" : {
-            "title": "Text Field value",
             "type": "string"
         }
     }
--- a/packages/fast-components-react-base/src/text-field/text-field.schema.json
+++ b/packages/fast-components-react-base/src/text-field/text-field.schema.json
@@ -1,0 +1,10 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Text Field",
+    "description": "A Text Field component's schema definition.",
+    "type": "object",
+    "id": "text-field",
+    "children": true,
+    "properties": {
+    }
+}
--- a/packages/fast-components-react-base/src/text-field/text-field.tsx
+++ b/packages/fast-components-react-base/src/text-field/text-field.tsx
@@ -2,17 +2,17 @@ import * as React from "react";
 import * as ReactDOM from "react-dom";
 import Foundation, { HandledProps } from "../foundation";
 import {
-    ITextfieldHandledProps,
-    ITextfieldManagedClasses,
-    ITextfieldUnhandledProps,
+    ITextFieldHandledProps,
+    ITextFieldManagedClasses,
+    ITextFieldUnhandledProps,
     TextFieldType
 } from "./text-field.props";
-import { IManagedClasses, ITextfieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { IManagedClasses, ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 import { get } from "lodash-es";
 
 /* tslint:disable-next-line */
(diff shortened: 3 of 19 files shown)


packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -0,0 +1,34 @@
+import { IDesignSystem } from "../design-system";
+import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
+import { ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { toPx } from "@microsoft/fast-jss-utilities";
+import * as Chroma from "chroma-js";
+
+const styles: ComponentStyles<ITextFieldClassNameContract, IDesignSystem> = {
+    textField: {
+        color: "#757575",
```
nalogiudice: You can use chroma to mix colors to get this result. See button for examples.

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -0,0 +1,34 @@
...
+        outline: `${toPx(1)} solid transparent`,
+        fontSize: toPx(14),
```
nalogiudice: Should use applyType to add fontSize and lineHeight

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -0,0 +1,34 @@
...
+        fontWeight: "100",
+        "&:focus": {
+            outline: `${toPx(1)} solid #616161`,
+            border: `${toPx(1)} solid #616161`
+        },
+        "&:disabled": {
+            color: "#CCCCCC",
+            cursor: "not-allowed",
+            background: (config: IDesignSystem): string => {
+                return Chroma(config.backgroundColor).css();
```
nalogiudice: You should just return `config.background` here - it is already in the correct format

packages/fast-components-styles-msft/src/text-field/index.ts
nalogiudice: This would go for all hardcoded hex values.

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -0,0 +1,69 @@
...
+     * Generates class names
+     */
+    protected generateClassNames(): string {
+        return super.generateClassNames(get(this.props, "managedClasses.textField"));
+    }
+
+    /**
+     * Generates handled attributes based on props
+     */
+    private generateHandledAttributes(): Partial<ITextFieldHandledProps> {
```
nalogiudice: We are trying to get away from the `generateAttributes` method for a few reasons. The props can go directly on the render function.

packages/fast-components-react-base/src/text-field/text-field.props.ts
```diff
@@ -0,0 +1,49 @@
+import * as React from "react";
+import { IManagedClasses, ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export enum TextFieldType {
+   date = "date",
```
nalogiudice: @nicholasrice do we want to do it this way?

packages/fast-components-react-base/src/text-field/text-field.tsx
nalogiudice: See checkbox for examples

packages/fast-components-styles-msft/src/text-field/index.ts
emathiasen: Could this be moved into a utility? Seems like its going to be used frequently.

packages/fast-components-react-base/src/text-field/text-field.props.ts
nicholasrice: The enum looks right to me, is that what you're reffering to?

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -0,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
```
nicholasrice: I think you need to rebase to master - these files are already in. If you could do a round on this PR and ensure that only the intended files are changed that would be good, cause there are a bunch of seemingly unrelated changes to this PR.

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -0,0 +1,74 @@
+import { IDesignSystem } from "../design-system";
+import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
+import { ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { toPx } from "@microsoft/fast-jss-utilities";
+import { applyType } from "../utilities/typography";
+import * as Chroma from "chroma-js";
+
+function applyMixedColor(incomingProperty: string, mixValue?: number, alpha?: number): ICSSRules<IDesignSystem> {
```
nicholasrice: `incomingProperty` isn't really doing anything for this function but reducing extensibility. Instead, you could change to:

```
function applyMixedColor(minvalue?: number, alpha?: number): (config: IDesignSystem) => string {
    return function applyColors(config: IDesignSystem): string {
       if (alpha) {
           return Chroma.mix(config.foregroundColor, config.backgroundColor, mixValue).alpha(alpha).css();
       } else {
           return Chroma.mix(config.foregroundColor, config.backgroundColor, mixValue).css();
       }
   }
}
```

and later apply it:
```
{
    outlineColor: applyMixedColor(0.38)
}
```

Doing that would expand the scope of use of the function and reduce code complexity

.github/ISSUE_TEMPLATE/Bug_report.md
emathiasen: rebased with 68b3ee8

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLInputElement> {
+        return (
+            <input
+                {...this.unhandledProps()}
+                className={this.generateClassNames()}
+                disabled={this.props.disabled}
```
chrisdholt: `disabled={this.props.disabled || null}`

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -0,0 +1,50 @@
...
+                placeholder={this.props.placeholder}
```
chrisdholt: `placeholder={this.props.placeholder || null}`

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -0,0 +1,50 @@
...
+    TextFieldType
+} from "./text-field.props";
+import { IManagedClasses, ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { get } from "lodash-es";
+
+/* tslint:disable-next-line */
+class TextField extends Foundation<ITextFieldHandledProps & IManagedClasses<ITextFieldClassNameContract>, React.HTMLAttributes<HTMLInputElement>, {}> {
+    protected handledProps: HandledProps<ITextFieldHandledProps & IManagedClasses<ITextFieldClassNameContract>> = {
+        disabled: void 0,
+        error: void 0,
```
chrisdholt: If we've pulled error, can we pull it from here as well?

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -0,0 +1,50 @@
...
+        placeholder: void 0,
+        managedClasses: void 0,
+        type: void 0,
+        value: void 0
```
chrisdholt: value won't be passed if we are handling it but not applying it to the element.

If we don't plan on doing anything with it, let's remove this from here and props to ensure users can pass it as expected.

packages/fast-components-react-base/src/text-field/text-field.props.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface ITextFieldHandledProps {
+    /**
+     * The disabled state
+     */
+    disabled?: boolean;
+
+    /**
+     * The error state
+     */
+    error?: boolean;
```
chrisdholt: can we pull error from here as well

packages/fast-components-react-base/src/text-field/text-field.props.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * The text field input type
+     */
+    type?: TextFieldType;
+
+    /**
+     * The text field value
+     */
+    value?: string;
```
chrisdholt: If we aren't handling this in the component, we can pull from props.

chrisdholt: Don't forget to pull from schema's as well.

packages/fast-components-react-base/src/text-field/text-field.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Text Field",
```
nalogiudice: Text field (de-capitalize field)

packages/fast-components-react-base/src/text-field/text-field.schema.json
```diff
@@ -0,0 +1,37 @@
...
+    "type": "object",
+    "id": "text-field",
+    "children": true,
+    "properties": {
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "type": {
+            "title": "Text Field type",
```
nalogiudice: Just "Type" - though this could be "Input type" as well

packages/fast-components-react-base/src/text-field/text-field.schema.json
```diff
@@ -0,0 +1,37 @@
...
+                "search",
+                "tel",
+                "text",
+                "time",
+                "url",
+                "week"
+            ]
+        },
+        "placeholder": {
+            "title": "Text Field placeholder",
```
nalogiudice: Just "Placeholder"

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -0,0 +1,45 @@
+import { IDesignSystem } from "../design-system";
+import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
+import { ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { toPx } from "@microsoft/fast-jss-utilities";
+import { applyType } from "../utilities/typography";
+import * as Chroma from "chroma-js";
+
+function applyMixedColor(mixValue?: number, alpha?: number): (config: IDesignSystem) => string {
+    return function applyColors(config: IDesignSystem): string {
```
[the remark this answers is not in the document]
nalogiudice: Not your concern, but we may want to make this a utility... thoughts @nicholasrice @chrisdholt ?

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -0,0 +1,45 @@
...
+            return Chroma.mix(config.foregroundColor, config.backgroundColor, mixValue).css();
+        }
+    };
+}
+
+const styles: ComponentStyles<ITextFieldClassNameContract, IDesignSystem> = {
+    textField: {
+        ...applyType("t7", "vp1"),
+        color: applyMixedColor(0.46),
+        fontWeight: "100",
```
nalogiudice: We have a utility for font weights that we should use

nalogiudice:
Also, noticing there is no hover state

packages/fast-components-react-base/src/text-field/text-field.props.ts
```diff
@@ -0,0 +1,39 @@
+import * as React from "react";
+import { IManagedClasses, ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export enum TextFieldType {
+   date = "date",
+   email = "email",
+   hidden = "hidden",
+   month = "month",
+   number = "number",
```
nalogiudice: We usually have a separate component for number field, so I think this should be omitted... again thoughts?

packages/fast-components-styles-msft/src/text-field/index.ts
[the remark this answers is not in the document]
awentzel: Already filed issues from Code Climate to resolve this.  I should have notified you guys here.  This is a new feature just implemented last week. can be resolved at later date.

packages/fast-components-react-base/src/text-field/text-field.props.ts
awentzel: @nalogiudice : If you're saying that text-field should be one word, then should that also apply to the file name?  Also, I'm not understanding what you mean by Number?  Do you have an example or file we can look at?

packages/fast-components-styles-msft/src/text-field/index.ts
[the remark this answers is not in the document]
chrisdholt: Yep - we'll break this out in a separate task as we'll want to resolve other places this can be used.

packages/fast-components-react-base/src/text-field/text-field.props.ts
chrisdholt: With unhandled props I think we can apply all functionality we need to to ensure both text and number fields are supported with a single component. I know previously we did that, although I don't recall the exact reason. Is there a specific scenario you can think of that would require two different components here?

emathiasen:
@nalogiudice [Figma](https://www.figma.com/file/Kc8GyAEJw6EGBQFlEvrsnsq7/FAST-Components---MS) comp didn't show a hover state.. We can definitely add one if other components have a similar interaction. 
Looping in @Falkicon should add a hover state for the Text field component?

packages/fast-components-styles-msft/src/text-field/index.ts
emathiasen: JSS expecting fontWeight as a string. Should utility be updated to reflect this or should it be converted in the components handling of that property?

packages/fast-components-react-base/src/text-field/text-field.props.ts
nalogiudice: I think we may be able to get by with unhandled props. Perhaps an example that tests that would be beneficial in this case. I.E. uses min, max, step etc... to verify it works as expected.

@awentzel if you look on fluentweb you'll see we have a number field component and a text field component, as separate components. But, @chrisdholt 's point is valid I believe, with my caveat above.

Text field is two words, no hyphen

packages/fast-components-styles-msft/src/text-field/index.ts
nalogiudice: Interesting. Not really in the scope of the PR but we could either simply `toString` the functions' results, or change it to return a string. I wonder if it is being used elsewhere and needs to be a number in that case.

nicholasrice: we may just need to update the typings for a stylesheet - I think right now it explicitly expects a string but sometimes numbers are valid.

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -0,0 +1,69 @@
...
+    ITextFieldUnhandledProps,
+    TextFieldType
+} from "./text-field.props";
+import { IManagedClasses, ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { get } from "lodash-es";
+
+/* tslint:disable-next-line */
+class TextField extends Foundation<ITextFieldHandledProps & IManagedClasses<ITextFieldClassNameContract>, React.HTMLAttributes<HTMLInputElement>, {}> {
+    protected handledProps: HandledProps<ITextFieldHandledProps & IManagedClasses<ITextFieldClassNameContract>> = {
+        disabled: void 0,
```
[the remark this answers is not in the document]
nalogiudice: @chrisdholt has this been resolved?

packages/fast-components-react-msft/src/text-field/examples.data.ts
```diff
@@ -0,0 +1,38 @@
+import { ISnapshotTestSuite } from "@microsoft/fast-jest-snapshots-react";
+import { IManagedClasses } from "@microsoft/fast-jss-manager-react";
+import TextField from "./index";
+import { ITextFieldHandledProps, TextFieldType } from "@microsoft/fast-components-react-base";
+import * as schema from "@microsoft/fast-components-react-base/dist/text-field/text-field.schema.json";
+
+export default {
+    name: "text-field",
```
nalogiudice: Should be `text field` as well

PR #401, opened 2018-05-10
chore(circle-ci): add initial configuration

awentzel:
Experimentation to investigate if Circle CI will better meet our CI/CD needs where Travis CI is falling short.


.circleci/config.yml
```diff
@@ -0,0 +1,73 @@
...
+          paths:
+            - ./node_modules   
+      - run: 
+          name: Test
+          command: |
+            lerna run test --stream
+
+  "test-coverage":
+    environment:
+      CC_TEST_REPORTER_ID=08a773cb4ea5811add5a45e12873e5cd2634c005568705cc37abfd5217617a32
```
janechu: should this be a token?

.circleci/run-build-locally.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+curl --user ${CIRCLE_CI_TOKEN}: \
+    --request POST \
+    --form revision=403b42af280ec3023a06826f4edaceda8a1f8a2d\
```
janechu: should this be a token?

.circleci/config.yml
awentzel: It's not required and almost all examples manage it this way.

.circleci/run-build-locally.sh
awentzel: This is only for local debugging.  Each time someone wanted to debug locally they would update the revision value with the git log id.  I don't think it should be a token as that would be more difficult to manage.

PR #409, opened 2018-05-10
fix(jss): fix errors caused by un-linked JSS rules

### LINKED ISSUE #342: JSS sometimes throws "link" error
image

Current repro in Toggle and also in ComponentViewToggle in development-site. 

It is preventing proper styling of components.

nicholasrice:
closes #342 

There are still 4 warnings in toggle caused by `-ms-high-contrast` media queries - I'm working with the JSS author on how best we can resolve these warnings.

### DIFF
--- a/packages/fast-components-styles-msft/src/toggle/index.ts
+++ b/packages/fast-components-styles-msft/src/toggle/index.ts
@@ -52,6 +52,11 @@ const styles: ComponentStyles<IToggleClassNameContract, IDesignSystem> = {
         width: toPx(10),
         height: toPx(10)
     },
+    "@media screen and (-ms-high-contrast: active)": {
+        toggle_button: {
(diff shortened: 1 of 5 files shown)


packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -10,9 +10,8 @@ const styles: ComponentStyles<IToggleClassNameContract, IDesignSystem> = {
         "& span": {
             userSelect: "none",
             marginTop: "0",
-            paddingBottom: "0"
-        },
-        "&[aria-disabled='true']": {
+            paddingBottom: "0" },
+        "&[aria-disabled=\"true\"]": {
```
nalogiudice: Interesting - the single quotes were causing the issue?

nicholasrice: yep :/

PR #411, opened 2018-05-10
fix(button): fixes an issue where href was not being passed as part of the created anchor element

### LINKED ISSUE #410: Button does not pass href as props to anchors
**Describe the bug**
Button does not pass href as props.

We are handling href to generate an anchor instead of a button, so we need to enumerate href to ensure the props are passed to the created element and not just referenced internally to generate the tag.

chrisdholt:
closes #410 

<body>

<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/button/__snapshots__/button.spec.ts.snap
+++ b/packages/fast-components-react-base/src/button/__snapshots__/button.spec.ts.snap
@@ -3,6 +3,7 @@
 exports[`button button: 0 1`] = `
 <button
   className="button"
+  href={null}
 >
   Button text
 </button>
--- a/packages/fast-components-react-base/src/button/button.tsx
+++ b/packages/fast-components-react-base/src/button/button.tsx
@@ -30,6 +30,7 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IButtonCla
             <this.tag
                 {...this.unhandledProps()}
                 className={this.generateClassNames()}
+                href={this.props.href || null}
                 {...this.renderDisabledAttribute()}
             >
                 {this.props.children}


packages/fast-components-react-base/src/button/__snapshots__/button.spec.ts.snap
```diff
@@ -3,6 +3,7 @@
 exports[`button button: 0 1`] = `
 <button
   className="button"
+  href={null}
```
nalogiudice: This feels weird to me. href should never be written to the button DOM, yet our snapshot shows it. Maybe it would be better to return href as an attribute only when the tag is an anchor

chrisdholt: I'm fine with that but we should treat all attributes the same (IE, checkbox disabled - disabled={this.props.disabled || null})

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -28,7 +28,7 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
                 {...this.unhandledProps()}
                 className={this.generateClassNames()}
                 managedClasses={this.props.managedClasses}
-                href={this.props.href}
+                href={this.props.href || null}
```
nicholasrice: I don't think you need to pass null here - the `BaseButton` should handle the `undefined` consistently

packages/fast-components-react-base/src/button/__snapshots__/button.spec.ts.snap
chrisdholt: And as a note, while it's in the snapshot, it's not in the DOM.

nicholasrice: It does look weird - I think that is just a product of snapshots though - react handles it correctly

packages/fast-components-react-msft/src/button/button.tsx
chrisdholt: updated (with force...sorry)

PR #413, opened 2018-05-11
chore(schemas): add schemas for components to packages which use the development site package

### LINKED ISSUE #375: Add JSON schemas to packages using the development site
Add the JSON schemas for the components in fast-components-react-msft and fast-components-react-base components.

**Adding schema files**: you can add these into the fast-components-react-base and fast-components-react-msft, these should match interfaces for each base or msft component, in base for example if you were to add a schema to button you would add it in src/button/button.schema.json

**Examples**: find example schemas in fast-form-generator-react/app/components

**Integrating into the webpack app**: In the components file examples.data.ts import the schema for example button would be import * as schema from "./button.schema.json"; and add to the default export the line schema, (which is the same as schema: schema)

janechu:
Closes #375


build/copy.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Utility for copy files to directories.
+ * Usage: node build/copy.js --src %srcPathGlob% --dist %distDirectory%
```
nicholasrice: should probably be "dest" - "dist" implies distribution

build/copy.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+const srcPaths = argv._.concat(argv.src);
+
+/**
+ * Function to copy a src file a given path
+ */
+function copyPath(srcPath) {
+    const firstDirectoryRegex = /[^\/]*/;
+
+    if (argv.dist) {
```
nicholasrice: you should use `path.join` - this may not work in Windows cause of different path delimiters. In general, use Node's `path` module to do path manipulation and you should be good.

packages/fast-components-react-base/app/index.tsx
```diff
@@ -16,10 +18,12 @@ const root: HTMLElement = document.createElement("div");
 root.setAttribute("id", "root");
 document.body.appendChild(root);
 
+const formChildOptions: IFormChildOption[] = formChildFromExamplesFactory(examples);
+
 /* tslint:disable */
 function render(): void {
     ReactDOM.render(
-        <Site title={"FAST components base"}>
+        <Site title={"FAST components base"} formChildOptions={formChildOptions} frameworks={["react"]}>
```
nicholasrice: enum? I thought you had made one.

packages/fast-development-site-react/src/utilities/form-child-from-examples-factory.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface IExample {
+    name: string;
+    component: any;
+    schema: JSON;
+    data: any[];
+}
+
+export default function componentFactory(examples: IExample): IFormChildOption[] {
+    const formChildOptions: IFormChildOption[] = [];
```
nicholasrice: you can just use `.map` to avoid creating this variable:

```
return Object.keys(examples).map((key: string) => {
    return {
        // ....
    };
});
```

packages/fast-components-react-base/src/checkbox/checkbox.schema.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Checkbox",
+    "description": "A checkbox component's schema definition.",
+    "type": "object",
+    "id": "checkbox",
+    "properties": {
+        "tag": {
+            "title": "HTML element tag",
```
nalogiudice: nit: prefer `HTML tag` (how it is on current fluentweb)

packages/fast-components-react-base/src/hypertext/hypertext.schema.json
```diff
@@ -0,0 +1,14 @@
...
+    "description": "A hypertext component's schema definition.",
+    "type": "object",
+    "id": "hypertext",
+    "children": true,
+    "properties": {
+        "href": {
+            "title": "HTML href attribute",
+            "type": "string"
+        }
+    }
```
nalogiudice: required?

packages/fast-components-react-base/src/image/image.schema.json
```diff
@@ -0,0 +1,60 @@
...
+        "src": {
+            "title": "HTML src attribute",
+            "type": "string"
+        },
+        "srcSet": {
+            "title": "HTML srcSet attribute",
+            "type": "string"
+        },
+        "vp1": {
+            "title": "Image address at viewport 1",
```
nalogiudice: Shorten to "Viewport 1 image".

Same with all below

packages/fast-components-react-base/src/toggle/toggle.schema.json
```diff
@@ -0,0 +1,38 @@
...
+        "selectedString": {
+            "title": "Selected text",
+            "type": "string"
+        },
+        "statusLabelId": {
+            "title": "HTML status label id attribute",
+            "type": "string"
+        },
+        "unselectedString": {
+            "title": "Unselected text",
```
nalogiudice: Missing required?

packages/fast-components-react-base/src/typography/typography.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Typography",
+    "description": "A typography component's schema definition.",
+    "type": "object",
+    "id": "typography",
+    "children": true,
+    "properties": {
+        "tag": {
+            "title": "HTML element tag",
```
nalogiudice: Same as other comment - prefer HTML tag

packages/fast-components-react-msft/src/heading/heading.schema.json
```diff
@@ -0,0 +1,35 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Heading",
+    "description": "A heading component's schema definition.",
+    "type": "object",
+    "id": "heading",
+    "children": true,
+    "properties": {
+        "tag": {
+            "title": "HTML element tag",
```
nalogiudice: Same here - prefer HTML tag

packages/fast-development-site-react/src/utilities/form-child-from-examples-factory.ts
janechu: fixed

packages/fast-components-react-msft/src/heading/heading.schema.json
janechu: fixed

packages/fast-components-react-base/src/typography/typography.schema.json
janechu: fixed

packages/fast-components-react-base/src/toggle/toggle.schema.json
janechu: fixed

packages/fast-components-react-base/src/image/image.schema.json
janechu: fixed

packages/fast-components-react-base/src/checkbox/checkbox.schema.json
janechu: fixed

build/copy-schemas.js
```diff
@@ -0,0 +1,30 @@
...
+ * Utility for copying schemas to their dist folders.
+ * Usage: node build/copy-schemas.js
+ */
+const path = require("path");
+const fs = require("fs");
+const glob = require("glob");
+
+const rootDir = path.resolve(process.cwd());
+const srcSchemaPaths = "src/**/*.schema.json";
+const destDir = "dist";
```
nicholasrice: should this also be "dest"?

janechu: no its for the dist folder for our distribution

Issue #417, opened 2018-05-11
Add all SVGs, additional JSS, and general examples


nalogiudice:
Add a general examples page that shows all form generator components together (more like they would be on the live site). Update all SVGs to be proper. Tweak JSS as needed.

PR #418, opened 2018-05-11
feat(form generator): add SVGs, additional JSS, and general examples

nalogiudice:
Closes [#417](https://github.com/Microsoft/fast-dna/issues/417)
<body>
Adds a general examples page that shows all form generator components together (more like they would be on the live site, updates all SVGs to be proper, and tweaks JSS as needed.
<footer>
 

### DIFF
--- a/packages/fast-form-generator-react/app/components/general-example/general-example.spec.tsx
+++ b/packages/fast-form-generator-react/app/components/general-example/general-example.spec.tsx
@@ -4,6 +4,7 @@ import { ISnapshotTestSuite } from "@microsoft/fast-jest-snapshots-react";
 import { getExample } from "@microsoft/fast-permutator";
 import Form, { IFormProps } from "../../../src/form";
 import * as generalExampleSchema from "./general-example";
+import config from "./general-example.config";
 
 const name: string = "general-example";
 
@@ -11,7 +12,8 @@ const exampleData: IFormProps = {
     schema: generalExampleSchema,
     data: getExample(generalExampleSchema),
     /* tslint:disable-next-line */
-    onChange: (data: any): void => {}
+    onChange: (data: any): void => {},
+    componentMappingToPropertyNames: config
 };
 
 const examples: ISnapshotTestSuite<IFormProps> = {
--- a/packages/fast-form-generator-react/app/components/general-example/general-example.spec.tsx
+++ b/packages/fast-form-generator-react/app/components/general-example/general-example.spec.tsx
@@ -1,0 +1,27 @@
+import * as React from "react";
+import { generateSnapshots } from "@microsoft/fast-jest-snapshots-react";
+import { ISnapshotTestSuite } from "@microsoft/fast-jest-snapshots-react";
+import { getExample } from "@microsoft/fast-permutator";
+import Form, { IFormProps } from "../../../src/form";
+import * as generalExampleSchema from "./general-example";
+
+const name: string = "general-example";
+
+const exampleData: IFormProps = {
+    schema: generalExampleSchema,
+    data: getExample(generalExampleSchema),
+    /* tslint:disable-next-line */
+    onChange: (data: any): void => {}
+};
+
+const examples: ISnapshotTestSuite<IFormProps> = {
+    name,
+    component: Form,
+    data: [
+        exampleData
+    ]
+};
+
+describe(name, () => {
+    generateSnapshots(examples);
+});
--- a/packages/fast-form-generator-react/app/components/general-example/general-example.tsx
+++ b/packages/fast-form-generator-react/app/components/general-example/general-example.tsx
@@ -3,7 +3,7 @@ import Array, { IArrayProps } from "../arrays/arrays";
 import Objects, { IObjectsProps } from "../objects/objects";
 import Theme, { IThemeProps } from "../theme/theme";
 
-export enum tagEnum {
+export enum GeneralExampleTags {
     button = "button",
     span = "span"
 }
@@ -13,7 +13,7 @@ export interface IGeneralExampleProps {
     alignVertical: string;
     level: number;
     level2: number;
-    tag: tagEnum;
+    tag: GeneralExampleTags;
     title: string;
     details: string;
     text: string;
--- a/packages/fast-form-generator-react/app/components/general-example/general-example.tsx
+++ b/packages/fast-form-generator-react/app/components/general-example/general-example.tsx
@@ -1,0 +1,38 @@
+import * as React from "react";
+
+export enum tagEnum {
+    button = "button",
+    span = "span"
+}
+
+export interface IGeneralExampleProps {
+    alignHorizontal: string;
+    alignVertical: string;
+    level: number;
+    tag: tagEnum;
+    title: string;
+    details: string;
+    text: string;
+    checkbbox: boolean;
+}
+
+/**
+ * This test components API should have:
+ * - a number of required property which maps to a configuration which will organise them by weight
+ */
+export default class GeneralExample extends React.Component<IGeneralExampleProps, {}> {
+    public render(): JSX.Element {
+        return (
+            <this.props.tag>
+                {this.props.alignHorizontal}
+                {this.props.alignVertical}
+                {this.props.level}
+                {this.props.title}
+                {this.props.details}
+                {this.props.tag}
+                {this.props.text}
+                {this.props.checkbbox}
+            </this.props.tag>
+        );
+    }
+}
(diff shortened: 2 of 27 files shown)


packages/fast-form-generator-react/app/components/general-example/general-example.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+import Form, { IFormProps } from "../../../src/form";
+import * as generalExampleSchema from "./general-example";
+import config from "./general-example.config";
+
+const name: string = "general-example";
+
+const exampleData: IFormProps = {
+    schema: generalExampleSchema,
+    data: getExample(generalExampleSchema),
+    /* tslint:disable-next-line */
```
chrisdholt: does this need to be here?

packages/fast-form-generator-react/app/components/general-example/general-example.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from "react";
+import Array, { IArrayProps } from "../arrays/arrays";
+import Objects, { IObjectsProps } from "../objects/objects";
+import Theme, { IThemeProps } from "../theme/theme";
+
+export enum tagEnum {
```
chrisdholt: We are PascalCasing enums

packages/fast-form-generator-react/app/index.tsx
```diff
@@ -81,7 +81,7 @@ export default class App extends React.Component<{}, IAppState> {
             <DesignSystemProvider designSystem={designSystemDefaults}>
                 <div>
                     <div
-                        style={{width: "300px", height: "100vh", background: "rgb(244, 245, 246)", float: "left"}}
+                        style={{width: "300px", minHeight: "100vh", padding: "0 8px", background: "rgb(244, 245, 246)", float: "left"}}
```
chrisdholt: Comment: I don't think we want this as the min-height. I think we'll want a pixel value. At a certain point, we should require a scroll or the view is going be totally broken.

packages/fast-form-generator-react/src/form/form-item.array.style.ts
```diff
@@ -31,8 +33,25 @@ const styles: ComponentStyles<IFormItemArrayClassNameContract, {}> = {
         ...applyAriaHiddenStyles(),
         ...applyPopupMenuStyles(),
         "& li": {
```
chrisdholt: Do we still need to target the `li` if we have no styles bound to it and the only styles in it are nested?

packages/fast-form-generator-react/src/form/form-item.array.style.ts
```diff
@@ -41,6 +60,18 @@ const styles: ComponentStyles<IFormItemArrayClassNameContract, {}> = {
         "& li button": {
```
chrisdholt: nit: prefer we target a class here.

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -234,15 +234,15 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
         // we have nothing to add or delete
         if (items.length === 0) {
             return [
-                (<li key={0}>No actions available</li>)
+                (<li key={0}><span>No actions available</span></li>)
```
chrisdholt: if this is for styling can we target a class instance instead of the element?

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -120,28 +121,39 @@ export function applyPopupMenuStyles(): ICSSRules<{}> {
...
-                opacity: ".6",
-                pointerEvents: "none",
-                top: toPx(12),
-                width: toPx(16),
-                height: toPx(16),
-                left: toPx(12),
-                /* tslint:disable-next-line */
-                background: "url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTMiIGhlaWdodD0iMTYiIHZpZXdCb3g9IjAgMCAxMyAxNiIgdmVyc2lvbj0iMS4xIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHhtbG5zOnhsaW5rPSJodHRwOi8vd3d3LnczLm9yZy8xOTk5L3hsaW5rIj48ZyBpZD0iQ2FudmFzIiBmaWxsPSJub25lIj48ZyBpZD0iJiMyMzg7JiMxNTk7JiMxMzE7Ij48cGF0aCBkPSJNIDE0IDQuMjg5MDZMIDE0IDE2TCAxIDE2TCAxIDBMIDkuNzEwOTQgMEwgMTQgNC4yODkwNlpNIDEwIDRMIDEyLjI4OTEgNEwgMTAgMS43MTA5NEwgMTAgNFpNIDEzIDE1TCAxMyA1TCA5IDVMIDkgMUwgMiAxTCAyIDE1TCAxMyAxNVoiIHRyYW5zZm9ybT0idHJhbnNsYXRlKC0xIDApIiBmaWxsPSJibGFjayIvPjwvZz48L2c+PC9zdmc+) center no-repeat"
+        "& li": {
+            "& > button": {
```
chrisdholt: if we target a class we do not need the direct child selector. Resolving in the dom becomes easier.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -120,28 +121,39 @@ export function applyPopupMenuStyles(): ICSSRules<{}> {
...
+                "&:focus": {
+                    outline: "none"
+                },
+                "&:hover": {
+                    backgroundColor: colors.grayBackground
+                }
             },
-            "&:focus": {
-                outline: "none"
+            "& > span": {
```
chrisdholt: same comment here about preferring a class instead an element.

packages/fast-form-generator-react/app/index.tsx
nalogiudice: This is just for the test site (fixes background not extending the entire way) - I think it's fine here.

packages/fast-form-generator-react/src/form/form-item.children.tsx
nalogiudice: Seems like unneeded weight/complexity to add classes for everything, but I can change...

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -120,28 +121,39 @@ export function applyPopupMenuStyles(): ICSSRules<{}> {
...
+                cursor: "pointer",
+                display: "block",
+                ...localizePadding(12, 12, 12, 36),
+                position: "relative",
+                border: "none",
+                background: "transparent",
+                textAlign: "left",
+                width: "100%",
+                ...ellipsis(),
+                "&:before": {
```
nicholasrice: `&::before`

packages/fast-form-generator-react/app/components/general-example/general-example.spec.tsx
nalogiudice: Yep

packages/fast-form-generator-react/src/form/form-item.array.style.ts
nalogiudice: Refactored and removed

packages/fast-form-generator-react/src/form/form-item.array.style.ts
nalogiudice: added

packages/fast-form-generator-react/src/form/form-item.children.tsx
nalogiudice: was able to use another preexisting class

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: Had a conversation with @nicholasrice about this, since these are shared functions used in several components. The only way to solve this would to be to break out all the individual elements into their own functions and then add all of those classes in the components that use them - which isn't necessarily better. Really, there should be a few more components (like pop-up menu) that gets the right styling. That is out of the scope of this task though. Leaving as is for now.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: See above comment.

Issue #414, opened 2018-05-11
React Image component throws several errors


nicholasrice:
- Launch `fast-components-react-msft` development site.
- Open image examples page
- Notice errors
![image](https://user-images.githubusercontent.com/3213292/39906604-1ec14c8c-5490-11e8-8929-b3fb9379d7ef.png)

PR #422, opened 2018-05-15
fix(image): fix incorrect data and logic causing error in srcSet 'w' value

nalogiudice:
Closes [#414](https://github.com/Microsoft/fast-dna/issues/414)
<body>

<footer>
 


packages/fast-components-react-base/src/image/image.tsx
```diff
@@ -107,7 +107,7 @@ class Image extends Foundation<IImageHandledProps & IManagedClasses<IImageClassN
         const vp: number = breakpoints[`vp${breakpointsId}`];
         const minWidth: number = vp;
 
-        if (!srcset) {
+        if (!srcset || this.props.src) {
```
janechu: looks like if src exists there is no src element?

nicholasrice: yea what is the purpose of `this.props.src`? it doesn't seem to drive any attribute, which seems un-intuitive.

nalogiudice:
@janechu & @nicholasrice - I removed that `this.props.src` logic

janechu:
Do we want to have any logic dictating what prop combinations should go in? Opening this up for discussion, I think it should be reflected in the interface and the schema.

nalogiudice:
I see image as having three sets of data that match three configurations: 

1. if vp props **exist**, it is a `<picture></picture>` element
2. if _only_ src exists, it is a `<image></image>` element
3. if src, srcSet, **and** sizes exist, it is `<image></image>` with srcSet

The current logic (which I havent changed) drives based on `src` meaning that the reason the `<picture></picture>` wasn't showing up earlier is because `src` existed in the example props that had the vp props.

PR #424, opened 2018-05-15
chore(dev-site): fix dev site spacing and styling issues for the left pane and header

chrisdholt:
<body>

<footer>
 

### DIFF
--- a/packages/fast-development-site-react/src/components/shell/pane.tsx
+++ b/packages/fast-development-site-react/src/components/shell/pane.tsx
@@ -27,8 +27,7 @@ const style: ComponentStyles<IShellPaneManagedClasses, IDevSiteDesignSystem> = {
         width: toPx(40),
     },
     shellPane__expanded: {
-        width: toPx(300),
-        padding: toPx(12)
+        width: toPx(300)
     },
     shellPane__hidden: {
         display: "none"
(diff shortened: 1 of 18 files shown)


packages/fast-development-site-react/src/components/shell/pane.tsx
```diff
@@ -27,8 +27,7 @@ const style: ComponentStyles<IShellPaneManagedClasses, IDevSiteDesignSystem> = {
         width: toPx(40),
     },
     shellPane__expanded: {
-        width: toPx(300),
-        padding: toPx(12)
+        width: toPx(300)
```
nalogiudice: Removing the padding would affect the form generator configuration style... is this accounted for?

chrisdholt: The form generator is using the same style...ok. I'll figure out a way to address this singular instance.

nalogiudice: Another option would be to bake this style into the form generator surrounding div or form element

chrisdholt: Proposal is up. Added an optional prop for className to enable styling within a layout so it doesn't have to be baked into the form itself.

nalogiudice: Cool looks good.

PR #438, opened 2018-05-16
feat(dev-site): adds error boundary to dev site component wrapper to prevent entire views from breaking

### LINKED ISSUE #436: Add error boundaries to component factory in dev site
Proposal is to add an error boundary outside each component instance in the dev site component factory. 

Wrapping each example in an error boundary would allow us and other users to better scope issues that are isolated to individual component instances, without breaking the entire experience due to incorrect examples, broken instances, etc.

chrisdholt:
<body>
Adds an error boundary to the component example wrapper to isolate errors and prevent an issue where the entire view is unable to load due to component specific errors.

closes #436 
<footer>
 

### DIFF
--- a/packages/fast-development-site-react/src/utilities/error-boundary.tsx
+++ b/packages/fast-development-site-react/src/utilities/error-boundary.tsx
@@ -3,9 +3,9 @@ import { toPx } from "@microsoft/fast-jss-utilities";
 import { IDevSiteDesignSystem } from "../../src/components/design-system";
 import manageJss, { ComponentStyles, IJSSManagerProps, IManagedClasses } from "@microsoft/fast-jss-manager-react";
 
-/*tslint:disable:no-empty-interface*/
+/* tslint:disable:no-empty-interface */
 export interface IErrorBoundaryProps {}
-/*tslint:disable:no-empty-interface*/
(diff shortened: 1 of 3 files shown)


chrisdholt:
error_boundary

nalogiudice:
@chrisdholt curious what happens if the error is super long.

packages/fast-development-site-react/src/utilities/error-boundary.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from "react";
+import { toPx } from "@microsoft/fast-jss-utilities";
+import { IDevSiteDesignSystem } from "../../src/components/design-system";
+import manageJss, { ComponentStyles, IJSSManagerProps, IManagedClasses } from "@microsoft/fast-jss-manager-react";
+
+/*tslint:disable:no-empty-interface*/
+export interface IErrorBoundaryProps {}
+/*tslint:disable:no-empty-interface*/
```
nicholasrice: should this be enable?

chrisdholt: Just making super sure it was disabled... :)

Updated.

Issue #430, opened 2018-05-16
Add hover and focus states for form generator items


nalogiudice:
Use fluentweb.com as a base

PR #449, opened 2018-05-18
feat(form generator): add focus/hover states

nalogiudice:
Closes [#430](https://github.com/Microsoft/fast-dna/issues/430)
<body>

<footer>
 

### DIFF
--- a/packages/fast-form-generator-react/src/form/form-item.array.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.array.tsx
@@ -87,6 +87,7 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
             <div className={this.props.managedClasses.formItemArray}>
                 <div>
                     <h3>{this.getLabelText()}</h3>
+                    {/* TODO: #460 Fix "identical-code" */}
                     <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
                     <ul aria-hidden={this.state.hideOptionMenu} className={this.props.managedClasses.formItemArray_actionMenu}>
                         {this.renderArrayMenuItems()}
--- a/packages/fast-form-generator-react/src/form/form-item.array.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.array.tsx
@@ -87,7 +87,7 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
             <div className={this.props.managedClasses.formItemArray}>
                 <div>
                     <h3>{this.getLabelText()}</h3>
-                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}>Options</button>
+                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
                     <ul aria-hidden={this.state.hideOptionMenu} className={this.props.managedClasses.formItemArray_actionMenu}>
                         {this.renderArrayMenuItems()}
                     </ul>
--- a/packages/fast-form-generator-react/src/form/form-item.children.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.children.tsx
@@ -263,9 +263,9 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
         }).map((option: any, index: number): JSX.Element => {
             return (
                 <li key={uniqueId()}>
-                    <a onClick={this.clickComponentFactory("add", option)}>
-                        {option.name}
-                    </a>
+                    <button onClick={this.clickComponentFactory("add", option)}>
+                        <span>{option.name}</span>
+                    </button>
                 </li>
             );
         });
@@ -345,7 +345,7 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
                 <div className={this.props.managedClasses.formItemChildren_existingChildren}>
                     <div>
                         <h3>Building blocks</h3>
-                        <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}>Options</button>
+                        <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
                         <ul className={this.props.managedClasses.formItemChildren_optionMenu} aria-hidden={this.state.hideOptionMenu}>
                             {this.getActionMenuChildItems()}
                         </ul>
--- a/packages/fast-form-generator-react/src/form/form-item.children.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.children.tsx
@@ -345,6 +345,7 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
                 <div className={this.props.managedClasses.formItemChildren_existingChildren}>
                     <div>
                         <h3>Building blocks</h3>
+                        {/* TODO: #460 Fix "identical-code" */}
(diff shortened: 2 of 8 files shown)


janechu:
For the align horizontal, align vertical and theme swapping we want those to have 2px borders, there should be one color for hover and one color for focus state, double borders are too busy.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -27,6 +34,18 @@ export function localizePadding(top: number, right: number, bottom: number, left
     };
 }
 
+export function insetStrongBoxShadow(color: string): ICSSRules<{}> {
```
awentzel: All functions should have comments describing what it is and what is does.

awentzel: Once I understand more about what this does I may suggest a different function name.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -27,6 +34,18 @@ export function localizePadding(top: number, right: number, bottom: number, left
...
+    return {
+        boxShadow: `inset ${toPx(0)} ${toPx(0)} ${toPx(0)} ${toPx(1)} ${color}`
+    };
+}
+
+export function insetHoverBoxShadow(): ICSSRules<{}> {
+    return {
+        boxShadow: `inset ${toPx(0)} ${toPx(0)} ${toPx(2)} ${colors.hover}`
+    };
+}
```
chrisdholt: Could these functions be combined into one such as "applyBoxShadow"? I get that it's fairly specific to this scenario, but do we need two functions for a singular property? Could we trim this down to one function?

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -16,6 +22,7 @@ export const colors: any = {
     gray: "#8A8A8A",
     grayBackground: "rgba(0, 0, 0, 0.04)",
     boxShadow: "rgba(0, 0, 0, 0.08)",
+    hover: "rgba(0,0,0,.3)",
```
chrisdholt: super nit, but can we align our spacing on these as they are all the same? IE - Space after comma

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -47,7 +66,13 @@ export function applyInputStyle(): ICSSRules<{}> {
         boxShadow: `inset ${toPx(0)} ${toPx(0)} ${toPx(4)} ${colors.boxShadow}`,
```
chrisdholt: If we had a singular function, we could use it here as well.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: yeah I was actually already doing that

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: sure

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: yep

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: I actually feel like this would be very redundant. Most of them will simply repeat the title... aka "applyInputStyles" - "Applies the general input styles"... doesn't really add any context.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
chrisdholt: perfect

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -27,6 +35,30 @@ export function localizePadding(top: number, right: number, bottom: number, left
     };
 }
 
+export function boxShadow(
+    offsetX: number,
```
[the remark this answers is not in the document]
nalogiudice: Code climate wanted to break out boxShadow into a single function, now that it is in a function, it is complaining about the number of arguments... I'm going to say this is a "Won't Fix"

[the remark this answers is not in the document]
awentzel: I think Code Climate has a good point here.  Maybe speak with Rice on ways to improve this function. Alternatively, add a comment `// TODO #123 : title...` inside the code after you create a new issue describing the problem, the reason the function is the way it is or the proposed fix. I'm ok with tackling this issue later given the current priorities.

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -87,7 +87,7 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
             <div className={this.props.managedClasses.formItemArray}>
                 <div>
                     <h3>{this.getLabelText()}</h3>
-                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}>Options</button>
+                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
```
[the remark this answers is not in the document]
awentzel: Create issue and todo statement and it can be refactored as a single utility function in the future.

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -345,7 +345,7 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
                 <div className={this.props.managedClasses.formItemChildren_existingChildren}>
                     <div>
                         <h3>Building blocks</h3>
-                        <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}>Options</button>
+                        <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
```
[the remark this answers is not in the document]
awentzel: Create issue and todo statement and it can be refactored as a single utility function in the future.

janechu:
The merge message is incorrectly formatted, makr sure there is no space between your type and parenthesis.

nalogiudice:
@awentzel TODOs have been added

packages/fast-form-generator-react/src/form/form-section.style.ts
```diff
@@ -78,13 +78,18 @@ const styles: ComponentStyles<IFormSectionClassNameContract, {}> = {
         },
         "&[aria-pressed='true']": {
```
nicholasrice: not in your changeset, but this is going to throw that link error - can you change to `&[aria-pressed=\"true\"]` while you're in here?

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -27,6 +38,34 @@ export function localizePadding(top: number, right: number, bottom: number, left
     };
 }
 
+// TODO: #459: Fix "argument_count"
```
nicholasrice: lets do this one now - just make it accept a config object instead of a set of arguments.

nicholasrice: ew thats actually really ugly - maybe leave as is - I think code climate is wrong on this one.

packages/fast-form-generator-react/src/form/form-section.style.ts
nalogiudice: whoops yep

PR #458, opened 2018-05-18
docs: update animation readme

awentzel:


packages/fast-animation/README.md
```diff
@@ -1,13 +1,9 @@
-# FAST animation library (Beta)
+# FAST Animation library
```
nalogiudice: nit: animation should not be capitalized

PR #467, opened 2018-05-22
feat(layouts): create configurable breakpoint tracker utility

### LINKED ISSUE #358: Create breakpoint tracker utility
- Should only be initialized once (implementation from previous repo exports the initialized instance)
- Should include default breakpoints but allow each instance to be configurable
- Should allow for up to 6 breakpoints, but not require all 6
- Should follow a subscriber model

chrisdholt:
closes #358 

<body>

<footer>
 


packages/fast-layouts-react/src/utilities/breakpoint-tracker.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    let subscriber: any;
+
+    beforeEach(() => {
+        subscriber = {
+            onBreakpointChanged: (notification: BreakpointTrackerCallback): void => {
+                return;
+            }
+        };
+    });
+
```
nicholasrice: the tests here don't really match the description - I'm not sure you really need to test for your description though, you can probably just remove these.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    test("should initialize automatically without the 'new' keyword", (): void => {
+        expect(breakpointTracker).toBeDefined();
+        expect(breakpointTracker).toBeTruthy();
+    });
+
+    test("should successfully track subscribers", (): void => {
```
nicholasrice: You actually shouldn't need unit tests for this. That "mocking is a code smell" article had some really useful insights for me for unit testing:

What you're really doing here is testing the internals of the tracker - which is the exact type of thing you *don't* want to test for. If I change up the way the internals this code works but *don't* change the interface with the code, then all of my tests should still pass because I haven't broken the API. If you test the internals though, then I need to update the code *and* the tests event though the API doesn't change - thus more work for no benefit.

A better way to test this is to add a subscription and ensure that the callback get's called when it is subscribed, and not get called after it is unsubscribed. Then you're testing subscription without probing the internal data structure of the component.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * The current breakpoint.
+     */
+    private breakpoint: number;
+
+    /**
+     * The current breakpoint.
+     */
+    private _breakpointConfig: Breakpoints = defaultBreakpoints;
```
nicholasrice: why the underscore?

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     *
+     */
+    private subscriptions: BreakpointTrackerCallback[];
+
+    /**
+     * Stores breakpoint config
+     */
+    private get breakpointConfig(): Breakpoints {
```
nicholasrice: There isn't a real reason to use a getter/setter here because you're not doing any manipulation during the get/set process. You should just access the property directly.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    /**
+     * The current breakpoint.
+     */
+    private _breakpointConfig: Breakpoints = defaultBreakpoints;
+
+    /**
+     * Store if requestAnimationFrame can be used
+     */
+    private useRequestAnimationFrame: boolean;
```
nicholasrice: It looks like our whole browser matrix now supports rAF (https://caniuse.com/#feat=requestanimationframe) - maybe we should remove the fallback and just rely on it?

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
chrisdholt: @nicholasrice throws an exception about naming conflicts with the getter and setter.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
chrisdholt: removing and renaming underscored private var to this.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
chrisdholt: Agreed - Pulling this out. I could see it technically being helpful, but in reality I think we stick with our current browser matrix for consistency and b/c for any open source projects, they likely aren't working in browsers that dont.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
chrisdholt: renaming based on convo regarding get/set

PR #470, opened 2018-05-23
feat(detail view): add detail view

### LINKED ISSUE #295: Integrate markdown-to-react utility into component-documentation page
Should render component documentation from its markdown files.

janechu:
Closes #295


packages/fast-components-react-base/src/toggle/examples.data.tsx
```diff
@@ -3,11 +3,28 @@ import { IManagedClasses } from "@microsoft/fast-components-class-name-contracts
...
+    detailData: {
+        managedClasses: {
+            toggle: "toggle",
+            toggle_label: "toggle_label",
+            toggle_wrapper: "toggle_wrapper",
+            toggle_input: "toggle_input",
+            toggle_button: "toggle_button",
+        },
+        children: "Toggle",
+        id: "toggle01",
```
nalogiudice: Should id's be different in detailData vs data? You are using the same ids below

nalogiudice: This is elsewhere as well

janechu: they don't need to be in this case since the views are never shown at the same time so the ids are always unique on the page

packages/fast-components-react-base/src/image/README.MD
```diff
@@ -1,31 +1,8 @@
-# Implementation
+## Best practices
+Use *image* to break up text-heavy pages, to illustrate products for sale, or to improve storytelling on a page. The use of *image* should always support the page’s story and, when possible, contribute to that story. Size the *image* to complement the rest of the page content. Use the **picture** option to employ a selection of image versions for different viewport sizes.
 
-## Implementation example
-```javascript
+### Style guidance
+Don't use *image* as a substitute for *hero*, *feature*, or *product* and *content placements*. Don't use an image if it reduces the effectiveness of the page.
```
chrisdholt: FYI - created an issue #473  to scrub documentation for non-existent components. We'll handle this in a different PR.

packages/fast-components-react-base/src/toggle/examples.data.tsx
nalogiudice: cool

PR #476, opened 2018-05-25
chore(perf): remove unnecessary instances of lodash in favor of lodash-es and add bundle analyzer to react-msft for debugging perf

chrisdholt:
<body>

<footer>
 

### DIFF
--- a/packages/fast-components-react-msft/webpack.config.js
+++ b/packages/fast-components-react-msft/webpack.config.js
@@ -15,7 +15,7 @@ module.exports = {
         publicPath: "/",
         filename: "[name].js"
     },
-    mode: "production",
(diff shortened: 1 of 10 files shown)


packages/fast-components-react-msft/webpack.config.js
```diff
@@ -41,6 +42,10 @@ module.exports = {
             onBuildStart: [
                 `npm run convert:readme`
             ]
+        }),
+        new BundleAnalyzerPlugin({
```
janechu: can we add this to the base package as well?

chrisdholt: yes we can.

PR #478, opened 2018-05-26
chore(perf): reduces production bundle size by enabling webpack tree shaking and removing duplicated dependencies

### LINKED ISSUE #367: add sideEffects: false to each package.json for webpack 4 users
### Expected behavior 
webpack can deeply prune unused exports because no side effects exists between reexports in any module packages

### Actual behavior
webpack 4 relies on having a `sideEffects: false` field in the package.json of resolved package to provide this treatment. I did a full analysis of every import/export in each package and it looks like no sideEffects actually occur at any re-export level, therefore its safe and a nonbreaking change to add this property. 

(PS: I'll send PR 😏 )

### Steps to reproduce the problem
N/A
### Specifications (hardware, software, configurations)
N/A
* Include device make, model, OS version: 
* Include browser and version: 
* Include screen orientation: 
* Include screen size: 
* Include page URL: 
* Include localization: 

*Please attach screenshots or video of the problem for quickest resolution*

### LINKED ISSUE #477: Enable tree shaking and Aliasing of dependencies in webpack to reduce overall production bundle size

chrisdholt:
closes #477 , closes #367 
<body>

**Sizes**

- Stat: 1.06MB

- Parse: 478.93KB

- GZipped: 131.66KB

**Before**
before_webpack_bundle_analyzer

**After**
after_webpack_bundle_analyzer

<footer>
 


packages/fast-components-react-base/package.json
```diff
@@ -85,6 +86,7 @@
     "react-dom": "^16.3.0"
   },
   "dependencies": {
-    "@microsoft/fast-components-class-name-contracts-base": "^1.6.0"
+    "@microsoft/fast-components-class-name-contracts-base": "^1.6.0",
+    "uglifyjs-webpack-plugin": "^1.2.5"
```
nicholasrice: this doesn't seem right - why would a webpack plugin be a dependency?

chrisdholt: Should we dev-, I'll update

chrisdholt: And I don't think we need it there - this is a mistake.

chrisdholt: Updated!

PR #485, opened 2018-05-30
feat(localization): adds direction rtl/ltr updating

### LINKED ISSUE #270: Add language changing support
We should be able to switch languages in for testing, this should also switch RTL.

There should be:
- A dropdown in the information bar that should contain a subset of languages including some RTL and LTR languages
- Selecting a language should change the language of the examples shown
- Any change in LTR should be passed as a callback so that users can update their design systems and other page level items accordingly.

janechu:
Closes #270

### DIFF
--- a/packages/fast-development-site-react/app/app.tsx
+++ b/packages/fast-development-site-react/app/app.tsx
@@ -19,6 +19,7 @@ import PolymerHeading from "./components/polymer-heading";
 import { ISiteCategoryProps } from "../src/components/site/category";
 import ParagraphDocs from "./components/paragraph/.tmp/documentation";
 import ButtonDocs from "./components/button/.tmp/documentation";
+import { FrameworkEnum } from "../src/components/site/dev-tools";
 
 export interface IAppState {
     direction: Direction;
@@ -38,8 +39,12 @@ export default class App extends React.Component<{}, IAppState> {
         }
     ];
 
+    private frameworks: FrameworkEnum[];
+
     constructor(props: {}) {
         super(props);
+
+        this.frameworks = [FrameworkEnum.react, FrameworkEnum.angular];
 
         this.state = {
             direction: Direction.ltr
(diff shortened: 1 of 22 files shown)


packages/fast-development-site-react/app/design-system/index.ts
```diff
@@ -0,0 +1,14 @@
+export enum Direction {
```
nicholasrice: You've defined this several times - is it not sharable?

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -30,16 +30,24 @@ import {
...
 
+export enum Direction {
+    rtl = "rtl",
+    ltr = "ltr"
+}
+
 export interface ISiteProps {
     title: string;
     formChildOptions: IFormChildOption[];
+    onUpdateLTR?: (ltr: Direction) => void;
```
nicholasrice: yea lets change this - you're updating document direction, not "ltr"

janechu: good point, fixing

packages/fast-development-site-react/app/app.tsx
```diff
@@ -19,152 +20,171 @@ import { ISiteCategoryProps } from "../src/components/site/category";
...
         return (
-            <SiteCategoryItem
-                slot={index === 0 ? "canvas-detail-view-example" : "canvas-example-view"}
-                key={index}
-                data={componentDataItem}
-            />
+            <Site
+                title={"FAST Documentation site"}
+                formChildOptions={this.formChildOptions}
+                frameworks={["react" as any, "angular" as any]}
```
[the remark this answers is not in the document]
janechu: fixed

packages/fast-development-site-react/app/design-system/index.ts
janechu: wasn't sure where to put it initially, now fixed, exported from the fast-application-utilities

packages/fast-development-site-react/app/components/button/button.tsx
```diff
@@ -1,20 +1,38 @@
...
 }
 
-class Button extends React.Component<IButtonProps, {}> {
+export interface ITestButtonClassNameContract {
+    button: string;
+}
+
+const ButtonStyles: ComponentStyles<ITestButtonClassNameContract, IDesignSystem> = {
+    button: {
+        width: "50px",
```
nalogiudice: `toPx()`?

janechu: not necessary to import the utility library here since these are just test components to verify the functionality of the dev site

Issue #482, opened 2018-05-30
Checkbox Indeterminate state indicator is visually broken in firefox


chrisdholt:
**Describe the bug**

**To Reproduce**
1. Go to **checkbox** in Firefox
2. Look at the indeterminate checkbox example

**Expected behavior**
Indeterminate state indicator matches design across all browsers

PR #487, opened 2018-05-30
fix(checkbox): aligns input to center of label text

nalogiudice:
Closes [#482](https://github.com/Microsoft/fast-dna/issues/482)
<body>

<footer>
 

### DIFF
--- a/packages/fast-components-styles-msft/src/checkbox/index.ts
+++ b/packages/fast-components-styles-msft/src/checkbox/index.ts
@@ -105,7 +105,7 @@ const styles: ComponentStyles<ICheckboxClassNameContract, IDesignSystem> = {
         marginTop: toPx(2)
     },
     checkbox_disabled: {
-        "& $checkbox_input, & $checkbox_label": {
+        "& $checkbox_input, & $checkbox_label, & $checkbox_span": {
             cursor: "default",
             opacity: ".6"
         }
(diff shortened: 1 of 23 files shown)


packages/fast-components-styles-msft/src/checkbox/index.ts
```diff
@@ -21,6 +21,7 @@ const styles: ComponentStyles<ICheckboxClassNameContract, IDesignSystem> = {
         boxSizing: "content-box",
         margin: "0",
         zIndex: "1",
+        alignSelf: "center",
```
chrisdholt: I think we want the parent to dictate this. Per our convo earlier, I have this in a branch with other fixes. I think the parent should set the alignment of the children to ensure that both are centered vertically with one another.

packages/fast-components-styles-msft/src/checkbox/index.ts
```diff
@@ -81,6 +82,7 @@ const styles: ComponentStyles<ICheckboxClassNameContract, IDesignSystem> = {
         display: "inline-block",
         width: toPx(20),
         height: toPx(20),
+        alignSelf: "center",
```
chrisdholt: We shouldn't need to do this in multiple places, the parent can ensure both of these align to one another.

PR #492, opened 2018-05-31
chore(animation): fixes incorrect npm registry config and unmarks package as private

chrisdholt:
<body>

<footer>
 

### DIFF
--- a/packages/fast-animation/.npmignore
+++ b/packages/fast-animation/.npmignore
@@ -1,11 +1,14 @@
 # Source files
+__tests__/
+
+# Source files
 client/
 lib/
 webpack/
 www/
 
 # Configs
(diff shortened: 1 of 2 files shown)


packages/fast-animation/.npmignore
```diff
@@ -1,5 +1,10 @@
-client
-webpack
+# Source files
+client/
```
janechu: could we add __test__/, node_modules, coverage, tslint.json in here as well
can remove .eslintrc, we dont appear to have that anymore

chrisdholt: I'll add test/, coverage/ and tslint...node modules is never touched in npm publish, even if you try and forcefully include it.

PR #495, opened 2018-06-01
feat: add sketch utility and design kit

### LINKED ISSUE #494: Add sketch support for react-msft

nicholasrice:
closes #494


packages/fast-components-react-msft/build/create-sketch-library.js
```diff
@@ -0,0 +1,38 @@
...
+
+const sources = [
+    "http://localhost:7001/components/button/",
+    "http://localhost:7001/components/checkbox/",
+    "http://localhost:7001/components/divider/",
+    "http://localhost:7001/components/heading/",
+    "http://localhost:7001/components/label/",
+    "http://localhost:7001/components/toggle/",
+    "http://localhost:7001/components/typography/",
+    "http://localhost:7001/components/hypertext/",
```
awentzel: Should these instead be made relative?  How would I do this in a production environment?

packages/fast-components-react-msft/build/create-sketch-library.js
```diff
@@ -0,0 +1,38 @@
...
+].map(value => {
+    return {
+        url: value,
+        selectors: "[data-sketch-symbol]"
```
awentzel: I expect we'll have additional export types.  Should we plan for this in the name?  data-exports-sketch or something?  What is a symbol in this context?

packages/fast-components-react-msft/build/create-sketch-library.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+});
+
+const config = {
+    name: "Fast-dna MSFT design kit",
```
awentzel: The name: should be "FAST-DNA" whenever used.  Also, should these page widths be dynamic based on actual browser size at the time the export is ran?

packages/fast-components-react-msft/src/divider/__snapshots__/divider.spec.ts.snap
```diff
@@ -3,6 +3,7 @@
 exports[`divider divider: 0 1`] = `
 <hr
   className="divider-0-1-1"
+  data-sketch-symbol="Divider"
```
awentzel: Is it possible that the "data-sketch-symbol" value could be derived from the component that the script is executed on in a way to eliminate the need for this attribute altogether?

packages/fast-sketch-library/readme.md
```diff
@@ -0,0 +1,40 @@
+# FAST-DNA sketch library
```
awentzel: Remove "-DNA" to be consistent with how we are naming other components as a title of the readme. But, very happy to see a readme and the right details to use.  :-)

packages/fast-sketch-library/readme.md
```diff
@@ -0,0 +1,40 @@
...
+`npm i -D @microsoft/fast-sketch-library`
+
+Set up a node script to extract symbols from a library
+
+```js
+const extractSymbolLibrary = require("@microsoft/fast-sketch-library").extractSymbolLibrary;
+const fs = require("fs");
+
+const config = {
+    name: "Fast-dna MSFT design kit",
```
awentzel: Capitalize FAST-DNA

packages/fast-sketch-library/readme.md
```diff
@@ -0,0 +1,40 @@
...
+    pageWidth: 1600,
+    pageHeight: 1600,
+    sources: [
+        {
+            url: "http://your-website.com",
+            selectors: ".some-element-selector"
+        }
```
awentzel: if you determine in other comments a change is needed here then we should update this to match.

packages/fast-sketch-library/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+import { Browser, Page } from "puppeteer";
+import * as fs from "fs";
+import * as path from "path";
+import { Page as SketchPage } from "@brainly/html-sketchapp";
+
+/**
+ * Store in-page script content as a string to be loaded into the browser
+ */
+const aSketchPage: string = fs.readFileSync(path.resolve(__dirname, "./aSketchPage.js")).toString();
+
```
awentzel: Should interface comments be added to fully understand the purpose of the interface itself?

packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -76,11 +76,13 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
     /**
      * Renders the component
      */
-    public render(): React.ReactElement<HTMLInputElement> {
+    public render(): React.ReactElement<HTMLElement> {
         return (
-            <this.tag className={this.generateClassNames()}>
+            <this.tag
+                {...this.unhandledProps()}
```
chrisdholt: I think this is the right move, but I wonder if there is anything else we need to add as props for the input now that we've moved unhandled props. Thoughts?

packages/fast-components-react-msft/build/create-sketch-library.js
chrisdholt: Curious on this as well, as right now I believe the same thing that is rendered into dev is what we compile and render as our production docs site right now.

nicholasrice: Nope. You point it any URL - if you want to use it in production, you point it at a production URL.

packages/fast-sketch-library/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "@microsoft/fast-sketch-library",
+  "description": "A Node.js utility for quickly extracting Sketch symbol libraries from a website",
+  "version": "1.0.0",
```
chrisdholt: Always worth another look, but I don't think that this has any side effects.

Can you add `sideEffects: false` to the package.json so that we can enable tree shaking for ourselves and others?

packages/fast-sketch-library/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "prepare": "npm run clean:dist && tsc -p ./tsconfig.json && webpack",
+    "test": "npm run tslint",
+    "tslint": "tslint -c ./tslint.json 'src/**/*.ts{,x}' '**/*.ts{,x}'",
+    "tslint:fix": "tslint -c ./tslint.json --fix 'src/**/*.ts{,x}' '**/*.ts{,x}'"
+  },
+  "license": "MIT",
+  "readme": "README.md",
+  "devDependencies": {
+    "@brainly/html-sketchapp": "^3.1.0",
+    "@microsoft/fast-tslint-rules": "^1.6.0",
```
chrisdholt: technically at 1.7 if you want to change it.

packages/fast-components-react-msft/build/create-sketch-library.js
nicholasrice: A symbol is a sketch concept. I feel like `data-exports-sketch` doesn't really change the meaning - I feel like `data-sketch-symbol` is descriptive and appropriate - if the scope of requirements changes we can easily change this, but without a clearer idea of what it would change to I don't know what else to make it.

packages/fast-components-react-msft/src/divider/__snapshots__/divider.spec.ts.snap
nicholasrice: I don't think it would give us the clarity we want. This value is harvested and applied as a Symbol name in sketch (and also used a selector to find sketch symbol sources), and right now we don't have the consistency between component values to do this accurately.

packages/fast-sketch-library/readme.md
nicholasrice: This is the readme - I don't think it needs to reflect our actual implementation and I think it's actually better if it is generic.

packages/fast-sketch-library/src/aSketchPage.ts
```diff
@@ -0,0 +1,51 @@
+import { Base, nodeToSketchLayers, Page, SymbolMaster, Text,  } from "@brainly/html-sketchapp";
+import { ISymbolLibrarySource } from "./index";
+const symbolNameDataAttribute: string = "data-sketch-symbol";
+
+export function getAsketchSymbols(source: ISymbolLibrarySource): JSON[] {
+    const selectors: string = Array.isArray(source.selectors) ? source.selectors.join(", ") : source.selectors;
+    const nodes: Element[] = Array.from(document.querySelectorAll(selectors));
+
+    return nodes.map((node: Element) => {
+        const rect: ClientRect = node.getBoundingClientRect();
```
chrisdholt: Generic comment in a specific place. In our implementation we expect the DOM, but we can't count on that for others. Should these functions ensure that the DOM is available, or should we note that this utility expects access to the DOM, so those implementing it should be aware of that?

packages/fast-sketch-library/src/index.ts
nicholasrice: good call  - updated.

packages/fast-sketch-library/readme.md
nicholasrice: actually changed this because I think it is better if it generic

packages/fast-sketch-library/readme.md
nicholasrice: done

packages/fast-components-react-msft/build/create-sketch-library.js
nicholasrice: done

packages/fast-components-react-base/src/checkbox/checkbox.tsx
nicholasrice: I'm not sure - I think we're squared away with event listeners because the events will bubble. If there is anything we're missing we can add it fairly easily

packages/fast-sketch-library/package.json
nicholasrice: sure

packages/fast-sketch-library/src/aSketchPage.ts
nicholasrice: I can make a note - yea this is 100% only executable in the dom, and honestly should error if the DOM isn't available so it doesn't fail silently.

packages/fast-sketch-library/package.json
nicholasrice: updated.

packages/fast-sketch-library/src/aSketchPage.ts
chrisdholt: Sounds good - just figure at minimum we should note that it's a requirement due to the internals.

packages/fast-sketch-library/readme.md
awentzel: This was generic about the overall code review not the readme. But, i get it.

PR #501, opened 2018-06-06
feat(development-site): add component for title and component for title with brand color applied

### LINKED ISSUE #283: Create new component so "Site-Title" can be styled properly
The site component passes a single string for the title prop that can't be split and styled properly based on the Figma comp.

janechu:
Closes #283


packages/fast-components-react-msft/app/app.tsx
```diff
@@ -25,10 +29,12 @@ export default class App extends React.Component<{}, {}> {
     public render(): JSX.Element {
         return (
             <Site
-                title={"FAST Microsoft component documentation"}
                 formChildOptions={formChildOptions}
                 onUpdateDirection={this.handleUpdateDirection}
             >
+                <SiteTitle slot={"title"}>
+                    <SiteTitleBrand>FAST</SiteTitleBrand> Microsoft component documentation
```
chrisdholt: We may want to check, I think one of the things Jason wanted was to shorten this string a bit - perhaps even to just "FAST documentation" for now (since it's currently the only one up)

packages/fast-development-site-react/src/components/site/title.tsx
```diff
@@ -0,0 +1,46 @@
...
+import manageJSS, { ComponentStyles, IJSSManagerProps, IManagedClasses } from "@microsoft/fast-jss-manager-react";
+import { IDevSiteDesignSystem } from "../design-system";
+
+
+export interface ISiteTitleProps {
+    slot?: string;
+}
+
+export interface ISiteTitleBrandedManagedClasses {
+    titleAcronym: string;
```
chrisdholt: Do we want to pick something a bit more extensible? It's an acronym for our use case, but this may not be the same for other teams or users. Perhaps we could do a sync tomorrow AM and land something?

chrisdholt: It's technically a "branded" string, so perhaps we want to go with something generic like that?

packages/fast-components-react-msft/app/app.tsx
nalogiudice: +1 for shorter

packages/fast-development-site-react/src/components/site/title.tsx
janechu: resolved

packages/fast-components-react-msft/app/app.tsx
janechu: resolved

packages/fast-components-react-base/app/index.tsx
```diff
@@ -23,7 +25,10 @@ const formChildOptions: IFormChildOption[] = formChildFromExamplesFactory(exampl
 /* tslint:disable */
 function render(): void {
     ReactDOM.render(
-        <Site title={"FAST base component documentation"} formChildOptions={formChildOptions}>
+        <Site formChildOptions={formChildOptions}>
+            <SiteTitle slot={"title"}>
```
nicholasrice: Is there an enum for slots? If not I can file a followup task

janechu: no enum, file away!

Issue #502, opened 2018-06-06
Add favicon and title to documentation site


chrisdholt:
<img width="458" alt="favicon_example" src="https://user-images.githubusercontent.com/13071055/41062487-d0013886-698a-11e8-99fc-b6f56d900a5e.png">

PR #506, opened 2018-06-06
feat(development site): add favicon for development site

nalogiudice:
<body>

<footer>
 
Closes [#502](https://github.com/Microsoft/fast-dna/issues/502)


packages/fast-development-site-react/webpack.config.js
```diff
@@ -18,6 +18,10 @@ module.exports = {
     },
     module: {
         rules: [
+            {
```
nicholasrice: it may be better to use somthing like https://github.com/jantimon/favicons-webpack-plugin - it will cover more bases

packages/fast-development-site-react/app/index.html
```diff
@@ -4,6 +4,7 @@
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width,initial-scale=1">
         <title>FAST React development app</title>
+        <link rel="shortcut icon" href="../src/assets/images/favicon.ico">
```
nicholasrice: we wouldn't want to source this from "./src" because it won't be bundled when webpack builds.

chrisdholt: +1

packages/fast-development-site-react/webpack.config.js
nalogiudice: @Falkicon created only a 16x16 favicon ico (not based off of a larger image), so I don't think we can use the above yet.

PR #508, opened 2018-06-06
feat(JSS): update JSS manager to support function stylesheets

### LINKED ISSUE #316: Update JSSManager to accept a function as a stylesheet
JSSManager should accept a function that returns a JSS stylesheet. 

This enables us to localize property names based on design-system inputs. All stylesheets that are a function would need to be mapped to dynamic stylesheets.

nicholasrice:
closes #316


packages/fast-jss-manager-react/src/manage-jss.spec.tsx
```diff
@@ -33,6 +33,17 @@ const dynamicStyles: ComponentStyles<any, any> = {
     }
 };
 
+const stylesheetResolver: ComponentStyles<any, any> = (config: any): any => {
```
awentzel: Seems like this might be a good candidate for a function comments

packages/fast-jss-manager-react/src/manage-jss.spec.tsx
```diff
@@ -33,6 +33,17 @@ const dynamicStyles: ComponentStyles<any, any> = {
...
+        resolvedStylesClass: {
+            background: "green",
+            color: (): string => {
+                return "yellow";
+            }
+        }
+    };
+};
+
 const staticAndDynamicStyles: ComponentStyles<any, any> = {
```
awentzel: comments would be great here.  what does this do?

packages/fast-jss-manager-react/src/manage-jss.tsx
```diff
@@ -8,7 +8,7 @@ import jss, { stylesheetManager } from "./jss";
 import { SheetsManager, StyleSheet } from "jss";
 import { IDesignSystemProviderProps } from "./design-system-provider";
 import * as propTypes from "prop-types";
-import { ClassNames, ComponentStyles, IManagedClasses } from "@microsoft/fast-jss-manager";
+import { ClassNames, ComponentStyles, ComponentStyleSheet, IManagedClasses } from "@microsoft/fast-jss-manager";
 import { isEqual, merge, omit } from "lodash-es";
 
 // hoist-non-react-statics does not seem to be a properly formatted ES6 module, so we need to require it instead
```
awentzel: Is this a known bug?  Seems like something we should tag with todo and issue number so we can clean up at later time when the dependency is fixed.

packages/fast-jss-manager-react/src/manage-jss.tsx
```diff
@@ -131,13 +123,43 @@ function manageJss<S, C>(styles?: ComponentStyles<S, C>): <T>(Component: React.C
...
+                    };
+                }, (): void => {
+                    if (this.hasStyleSheet()) {
+                        this.state.styleSheet.attach().update(this.designSystem);
+                    }
+                });
+            }
+
             /**
              * Creates a jss stylesheet from the dynamic portion of an associated style object and any style object passed
```
awentzel: nit: JSS is referenced should be the same casing.  So either jss or JSS.  I'd suggest JSS since it's an acronym.  Though in title comments i'd use "feat(jss)..." for consistency and cleaner change logs.  Just my thoughts

packages/fast-jss-manager/src/index.ts
```diff
@@ -1,6 +1,12 @@
 import * as CSS from "csstype";
 import { ClassNames, IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
 export { ClassNames, IManagedClasses };
+
+/**
+ * Define a static CSS Rule
+ */
+export type ICSSStaticRule =  CSS.Properties | string;
```
awentzel: Suggest using @params for all type, functions, etc as well like what was done below.  Good practice to declare in comments the inputs and outputs to fully understand how it's used.

packages/fast-jss-manager-react/src/manage-jss.spec.tsx
nicholasrice: sure.

packages/fast-jss-manager-react/src/manage-jss.tsx
nicholasrice: sure

packages/fast-jss-manager-react/src/manage-jss.spec.tsx
nicholasrice: fixed.

PR #509, opened 2018-06-06
fix(permutator): fix an issue where lodash was being referred to and not lodash-es

### LINKED ISSUE #507: Fix the issue where we cannot add children in the form generator
Currently clicking the children in the form generator does not result in the children being added to the component, this should work as intended.

janechu:
Resolves #507

### DIFF
--- a/packages/fast-permutator/__tests__/test-utils/index.js
+++ b/packages/fast-permutator/__tests__/test-utils/index.js
@@ -1,5 +1,5 @@
-const get = require('lodash/get'),
-    permutator = require('../../src/index');
+const get = require('lodash-es').get;
+const permutator = require('../../src/index');
 
 function prettyJSON(obj) {
     return JSON.stringify(obj, null, 2);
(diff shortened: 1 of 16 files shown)


packages/fast-permutator/__tests__/test-utils/index.js
```diff
@@ -1,5 +1,5 @@
-const get = require('lodash/get'),
-    permutator = require('../../src/index');
+const get = require('lodash-es').get;
```
chrisdholt: @janechu Is there a reason for the syntax change here back to dot notation? This ends up increasing our dependency on lodash-es from 9kb to 32kb.

chrisdholt: Yep...this resolves it.

chrisdholt:
@janechu can you make one other change? Can you remove the aliasing in the fast-components-react-msft package that moves lodash to lodash-es. It's not needed.

janechu:
Removed the redundant lodash alias

PR #517, opened 2018-06-08
feat(fast-components-react-msft): add localized styles and enable ltr/rtl swapping on documentation site

### LINKED ISSUE #486: Add direction to the design system of Microsoft styles
- Direction (LTR/RTL) should be updated when the development site dropdown is triggered, this functionality already exists with a handler being passed but we need to add this configuration to the design system
- Any use of localization should accept the new design system configuration for the direction (eg. localizedSpacing)

chrisdholt:
<body>
closes #486 
<footer>
 

### DIFF
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -69,6 +69,8 @@ export default class App extends React.Component<{}, IAppState> {
     private handleUpdateDirection = (direction: Direction): void => {
         const newDir: Direction = this.state.direction === Direction.ltr ? Direction.rtl : Direction.ltr;
 
+        if (this.state.direction === newDir) { return; }
+
         this.setState({
             direction: newDir
         });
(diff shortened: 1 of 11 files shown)


packages/fast-components-react-msft/app/app.tsx
```diff
@@ -56,6 +67,12 @@ export default class App extends React.Component<{}, {}> {
     }
 
     private handleUpdateDirection = (direction: Direction): void => {
-        // TODO: #486 change the design system direction (ltr)
+        const newDir: Direction = this.state.direction === Direction.ltr ? Direction.rtl : Direction.ltr;
```
nicholasrice: should this instead update `state.direction` to the value of the `direction` argument?

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -56,6 +67,12 @@ export default class App extends React.Component<{}, {}> {
     }
 
     private handleUpdateDirection = (direction: Direction): void => {
```
nicholasrice: how is this `handleUpdateDirection` used? It seems like "direction" should really just be a prop - this function doesn't get called anywhere either

chrisdholt: Synced offline - not seen b/c the code was folded.

Issue #293, opened 2018-04-20
Create infobar for footer of Fast apps


nicholasrice:
Should display:
- Currently selected component
- component status (under consideration, in development, complete) (verify these with @Falkicon)

eljefe223:
#542

PR #558, opened 2018-06-18
feat(fast-development-site): Create infobar for fast site

eljefe223:
closes: #293 

### DIFF
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -26,8 +26,7 @@ const formChildOptions: IFormChildOption[] = formChildFromExamplesFactory(exampl
 
 const hypertextStyles: ComponentStyles<IColumnClassNamesContract, undefined> = {
     hypertext: {
-        marginLeft: "8px",
-        marginRight: "8px",
+        margin: "0 8px",
         display: "inline-block",
         lineHeight: "1",
         whiteSpace: "nowrap"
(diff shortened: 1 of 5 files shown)


packages/fast-development-site-react/src/components/site/category.tsx
```diff
@@ -6,6 +6,13 @@ export interface ISiteCategoryProps {
     name: string;
     component?: any;
     schema?: any;
+    status?: Status;
+}
+
+export enum Status {
```
chrisdholt: @eljefe223 Can you add documentation for each of these status' in the readme. We should likely also reference it in the readme for fast-components-react-msft and base since these status' will correlate directly to the components.

eljefe223: Added to #8

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -54,9 +63,9 @@ export default class App extends React.Component<{}, IAppState> {
                 <SiteCategory slot={"category"} name={"Components"}>
                     {componentFactory(examples, Object.assign({}, DesignSystemDefaults, {direction: this.state.direction}))}
                 </SiteCategory>
-                <div slot="info-bar">
+                <div slot="info-bar" style={{display: "flex", alignItems: "center"}}>
```
janechu: This should be a class not inline styles

Issue #545, opened 2018-06-14
Space is missing above titles on configure side pane


Falkicon:
On FAST Documentation, the proper 20px spacing is missing in the configure side pane. Currently, this applies to "Building blocks" and "Add building blocks".

See example of header spacing in side pane here: 
https://www.figma.com/file/0MUSNF4N5VnE4Mmo0qX0GYtE/FAST-Application

PR #570, opened 2018-06-21
fix(fast-development-site) Update styling to config panel to match design

eljefe223:
Closes: #545 
 

### DIFF
--- a/packages/fast-form-generator-react/src/form/form-item.align-horizontal.style.ts
+++ b/packages/fast-form-generator-react/src/form/form-item.align-horizontal.style.ts
@@ -1,6 +1,13 @@
-import { applyInputBackplateStyle, applyLabelStyle, applyWrapperStyle, colors } from "../utilities/form-input.style";
+import {
+    applyInputBackplateStyle,
+    applyInputWrapperStyle,
+    applyLabelStyle,
+    applyWrapperStyle,
+    colors
+} from "../utilities/form-input.style";
 import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
 import { IFormItemAlignHorizontalClassNameContract } from "../class-name-contracts/";
+import { toPx } from "@microsoft/fast-jss-utilities";
 
 /* tslint:disable */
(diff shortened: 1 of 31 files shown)


packages/fast-form-generator-react/src/form/form-item.align-horizontal.style.ts
```diff
@@ -1,6 +1,13 @@
...
+import {
+    applyInputBackplateStyle,
+    applyInputWrapperStyle,
+    applyLabelStyle,
+    applyWrapperStyle,
+    colors
+} from "../utilities/form-input.style";
 import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
 import { IFormItemAlignHorizontalClassNameContract } from "../class-name-contracts/";
+import { toPx } from "@microsoft/fast-jss-utilities";
```
nicholasrice: do you actually need `toPx`?

packages/fast-form-generator-react/src/form/form-item.align-horizontal.tsx
```diff
@@ -21,7 +21,7 @@ class FormItemAlignHorizontal extends React.Component<IFormItemComponentMappingT
                 >
                     {this.props.label}
                 </label>
-                <div>
+                <div className={this.props.managedClasses.formItemAlignHorizontal_input}>
```
janechu: @nicholasrice technically this isn't the input, should we still be using this type of naming?

packages/fast-form-generator-react/src/form/form-item.align-vertical.style.ts
```diff
@@ -1,4 +1,10 @@
-import { applyInputBackplateStyle, applyLabelStyle, applyWrapperStyle, colors } from "../utilities/form-input.style";
+import {
+    applyInputBackplateStyle,
+    applyInputWrapperStyle,
+    applyLabelStyle,
+    applyWrapperStyle,
+    colors
```
janechu: looks like we are only using the imported applyInputWrapperStyle, not the rest

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -85,8 +85,8 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
     public render(): JSX.Element {
         return (
             <div className={this.props.managedClasses.formItemArray}>
-                <div>
-                    <h3>{this.getLabelText()}</h3>
+                <div className={this.props.managedClasses.formItemArray_header}>
+                    <h3 style={{margin: 0}}>{this.getLabelText()}</h3>
```
janechu: I think we shouldn't be inlining styles here, either we can target it via the formItemArray_header or create a new class and attach it

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -392,8 +392,8 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
         if (typeof currentChildren !== "undefined") {
             return (
                 <div className={this.props.managedClasses.formItemChildren_existingChildren}>
-                    <div>
-                        <h3>Building blocks</h3>
+                    <div className={this.props.managedClasses.formItemChildren_header}>
+                        <h3 style={{margin: 0}}>Building blocks</h3>
```
janechu: again, could be targetted via the formItemChildren_header or using a new class

packages/fast-form-generator-react/src/form/form-section.style.ts
```diff
@@ -40,17 +44,19 @@ const styles: ComponentStyles<IFormSectionClassNameContract, {}> = {
                 right: "0"
             },
             "& a": {
-                display: "block",
-                ...localizePadding(12, 18, 12, 1)
```
janechu: you can remove the import for this since it looks like we aren't using it in the file any longer

packages/fast-form-generator-react/src/form/form-item.align-horizontal.style.ts
eljefe223: no we don't I added for some inline styles that I ended up moving.

packages/fast-form-generator-react/src/form/form-item.align-horizontal.tsx
nicholasrice: yea it should probably be something more like "inputContainer"

packages/fast-form-generator-react/src/form/form-item.align-vertical.style.ts
janechu: Nevermind, I misread that this had been added on individual lines

PR #573, opened 2018-06-21
fix(fast-development-site) Move design kit link behind Site Menu

### LINKED ISSUE #14: Nested JSS classes without any css rules does not export the classname
### Expected behavior 
One should be able to nest a JSS class without defining root style rules
```
{
    root: {
        "& $nested: { color: "red" }
   },
   nested: {}
}
```
### Actual behavior
There is no class exposed under the `nested` key of the classes object.

eljefe223:
Closes #14 
 

### DIFF
--- a/packages/fast-development-site-react/src/components/site/menu.tsx
+++ b/packages/fast-development-site-react/src/components/site/menu.tsx
@@ -16,6 +16,7 @@ export interface ISiteNavManagedClasses {
     siteMenu_nav: string;
     siteMenu_button: string;
     siteMenu_buttonGlyph: string;
+    siteMenu_ul: string;
 }
 
 const style: ComponentStyles<ISiteNavManagedClasses, IDevSiteDesignSystem> = {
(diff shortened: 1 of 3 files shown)


packages/fast-development-site-react/src/components/site/menu.tsx
```diff
@@ -25,6 +26,9 @@ const style: ComponentStyles<ISiteNavManagedClasses, IDevSiteDesignSystem> = {
     },
     siteMenu_nav: {
         position: "absolute",
+        background: "#FFFFFF",
+        zIndex: "5",
+        boxShadow: `outline ${toPx(-2)} 0 ${toPx(4)} ${toPx(4)} rgba(0, 0, 0, 0.2)`,
```
nalogiudice: We could use Chroma here to get the desired background boxshadow color.
`boxShadow: (config: IDesignSystem): string => {
      return `outline ${toPx(-2)} 0 ${toPx(4)} ${toPx(4)} ${Chroma(config.foregroundColor).alpha(0.2).css()}`;
  }`

chrisdholt: I think with our overall goal of removing Chroma where we can (because we aren't using fast-colors yet), we should hold on this. We have quite a few places we could use it here, but I think we intentionally left off the dependency. We definitely need a better way to manipulate the colors here as most all of them are hard coded or built into the design system for single use cases. I think for now though, we should hold on using Chroma here and come up with a viable solution.

PR #587, opened 2018-06-26
fix(fast-development-site): changed foreground color of <a/> tags to brand color

### LINKED ISSUE #581: Hyperlinks should be branch color and not blue on Fast Sites
Hyperlinks should be branch color and not blue on Fast Sites

eljefe223:
closes #581 
 

### DIFF
--- a/packages/fast-development-site-react/src/components/site/category-documentation.tsx
+++ b/packages/fast-development-site-react/src/components/site/category-documentation.tsx
@@ -13,7 +13,12 @@ export interface ICategoryDocumentationManagedClasses {
 
 const style: ComponentStyles<ICategoryDocumentationManagedClasses, IDevSiteDesignSystem> = {
     documentationPanel: {
-        maxWidth: toPx(1000)
+        maxWidth: toPx(1000),
+        "& a": {
+            color: (config: IDevSiteDesignSystem): string => {
+                return `${config.brandColor}`;
+            }
+        }
     }
 };
 


packages/fast-development-site-react/src/components/site/category-documentation.tsx
```diff
@@ -13,7 +13,12 @@ export interface ICategoryDocumentationManagedClasses {
 
 const style: ComponentStyles<ICategoryDocumentationManagedClasses, IDevSiteDesignSystem> = {
     documentationPanel: {
-        maxWidth: toPx(1000)
+        maxWidth: toPx(1000),
+        "& a": {
+            color: (config: IDevSiteDesignSystem): string => {
+                return `${config.brandColor}`;
```
nalogiudice: You don't need to interpolate - you can just return `config.brandColor`

chrisdholt:
@eljefe223 Is there a reason that we decided not to take care of this in the markdown editor?

packages/fast-development-site-react/src/components/site/category-documentation.tsx
```diff
@@ -13,7 +13,12 @@ export interface ICategoryDocumentationManagedClasses {
 
 const style: ComponentStyles<ICategoryDocumentationManagedClasses, IDevSiteDesignSystem> = {
     documentationPanel: {
-        maxWidth: toPx(1000)
+        maxWidth: toPx(1000),
+        "& a": {
```
chrisdholt: What happens if an anchor comes in without an href? There would be no difference between an active anchor and a disabled anchor. What if someone else is using this package and mapping their own components to the dev site here? I don't think a global style within the documentation panel is the right way to solve this.

nicholasrice: Are you suggesting passing in a styled component into the dev site?

chrisdholt: I believe we already have a library that converts markdown into our own components (fast-markdown) and this could be leveraged in the MSFT library when we create the <Documentation /> component. We don't need to do the work in the dev site, it can be specific to the project using it.

PR #591, opened 2018-06-27
feat(fast-development-site): add component transparency toggle to info bar

### LINKED ISSUE #556: Add Transparency toggle to footer
- transparency toggle - this should change the background of the component viewing area so users can differentiate between the background of a component and the background of the viewing area.

eljefe223:
closes #556 

### DIFF
--- a/packages/fast-development-site-react/src/components/site/component-wrapper.tsx
+++ b/packages/fast-development-site-react/src/components/site/component-wrapper.tsx
@@ -7,14 +7,13 @@ import manageJss, {
     IManagedClasses
 } from "@microsoft/fast-jss-manager-react";
 import { ErrorBoundary, IErrorBoundaryProps } from "../../utilities";
-import { glyphBuildingblocks } from "@microsoft/fast-glyphs-msft";
 import { toPx } from "@microsoft/fast-jss-utilities";
 import devSiteDesignSystemDefaults, { IDevSiteDesignSystem } from "../design-system";
 import { ComponentViewTypes } from "./component-view";
 
 export interface IComponentWrapperManagedClasses {
     componentWrapper: string;
-    componentWrapper__Transparent: string;
+    componentWrapper__transparent: string;
     componentWrapperExamples: string;
     componentWrapper__active: string;
 }
@@ -25,11 +24,11 @@ export interface IComponentWrapperProps<T> {
     singleRow?: boolean;
     designSystem: T;
     view: ComponentViewTypes;
-    transparent: boolean;
+    transparentBackground?: boolean;
     onClick?: (activeIndex: number) => void;
 }
 
-/* tslint:disable */
+/* tslint:disable-next-line */
 const checker: string = "url('data:image/gif;base64,R0lGODlhEAAQAPAAANfX1wAAACH5BAEAAAEALAAAAAAQABAAAAIfhG+hq4jM3IFLJhoswNly/XkcBpIiVaInlLJr9FZWAQAh/wtYTVAgRGF0YVhNUDw/eHBhY2tldCBiZWdpbj0i77u/IiBpZD0iVzVNME1wQ2VoaUh6cmVTek5UY3prYzlkIj8+Cjx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8iIHg6eG1wdGs9IkFkb2JlIFhNUCBDb3JlIDUuNi1jMTM4IDc5LjE1OTgyNCwgMjAxNi8wOS8xNC0wMTowOTowMSAgICAgICAgIj4KIDxyZGY6UkRGIHhtbG5zOnJkZj0iaHR0cDovL3d3dy53My5vcmcvMTk5OS8wMi8yMi1yZGYtc3ludGF4LW5zIyI+CiAgPHJkZjpEZXNjcmlwdGlvbiByZGY6YWJvdXQ9IiIvPgogPC9yZGY6UkRGPgo8L3g6eG1wbWV0YT4KPD94cGFja2V0IGVuZD0iciI/PgH//v38+/r5+Pf29fTz8vHw7+7t7Ovq6ejn5uXk4+Lh4N/e3dzb2tnY19bV1NPS0dDPzs3My8rJyMfGxcTDwsHAv769vLu6ubi3trW0s7KxsK+urayrqqmop6alpKOioaCfnp2cm5qZmJeWlZSTkpGQj46NjIuKiYiHhoWEg4KBgH9+fXx7enl4d3Z1dHNycXBvbm1sa2ppaGdmZWRjYmFgX15dXFtaWVhXVlVUU1JRUE9OTUxLSklIR0ZFRENCQUA/Pj08Ozo5ODc2NTQzMjEwLy4tLCsqKSgnJiUkIyIhIB8eHRwbGhkYFxYVFBMSERAPDg0MCwoJCAcGBQQDAgEAADs=')";
 
 const componentWrapperBorder: string = `${toPx(1)} solid rgb(204, 204, 204)`;
@@ -39,10 +38,8 @@ const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSys
         padding: toPx(24),
         borderBottom: componentWrapperBorder,
     },
-    componentWrapper__Transparent: {
-        backgroundSize: `${toPx(8)} ${toPx(8)}`,
-        background: checker,
-        backgroundRepeat: "repeat",
+    componentWrapper__transparent: {
+        background: `transparent ${toPx(8)}/${toPx(8)} ${checker} repeat`
     },
     componentWrapperExamples: {
         "&:last-child:not(:nth-child(3n + 3)), &:nth-child(3n + 1)": {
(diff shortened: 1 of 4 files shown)


packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -7,12 +7,14 @@ import manageJss, {
...
 } from "@microsoft/fast-jss-manager-react";
 import { ErrorBoundary, IErrorBoundaryProps } from "../../utilities";
+import { glyphBuildingblocks } from "@microsoft/fast-glyphs-msft";
 import { toPx } from "@microsoft/fast-jss-utilities";
 import devSiteDesignSystemDefaults, { IDevSiteDesignSystem } from "../design-system";
 import { ComponentViewTypes } from "./component-view";
 
 export interface IComponentWrapperManagedClasses {
     componentWrapper: string;
+    componentWrapper__Transparent: string;
```
chrisdholt: should this be lowercase t?

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -7,12 +7,14 @@ import manageJss, {
     IManagedClasses
 } from "@microsoft/fast-jss-manager-react";
 import { ErrorBoundary, IErrorBoundaryProps } from "../../utilities";
+import { glyphBuildingblocks } from "@microsoft/fast-glyphs-msft";
```
chrisdholt: I don't see us using this anywhere, do we need it? If not, please remove it

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -23,16 +25,25 @@ export interface IComponentWrapperProps<T> {
     singleRow?: boolean;
     designSystem: T;
     view: ComponentViewTypes;
+    transparent: boolean;
```
chrisdholt: I'm not sure this should be a required prop - Seems more like something you would opt into.

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -23,16 +25,25 @@ export interface IComponentWrapperProps<T> {
...
 const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSystem> = {
     componentWrapper: {
         display: "block",
         padding: toPx(24),
         borderBottom: componentWrapperBorder,
     },
+    componentWrapper__Transparent: {
+        backgroundSize: `${toPx(8)} ${toPx(8)}`,
+        background: checker,
+        backgroundRepeat: "repeat",
```
chrisdholt: can we shorthand this to just "background"?

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -91,7 +102,9 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
     }
 
     private getClassNames(): string {
-        const classNames: string = this.props.managedClasses.componentWrapper;
+        const classNames: string = this.props.transparent
```
chrisdholt: with the other change to this in your PR, this will get even more complex and difficult to read. Suggest breaking this up here as well.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -217,6 +222,19 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
...
+        background: "none",
+        height: toPx(32),
+        width: toPx(32),
+        outline: "0",
+        verticalAlign: "middle"
+    },
+    site_transparencyToggleButtonIcon: {
+        display: "block",
+        verticalAlign: "middle",
+        fontSize: "0"
```
chrisdholt: do we need to define fontSize as "0"?

nicholasrice: It looks like this is using text-based styles to achieve vertical alignment, is that right? Is it possible to use flexbox instead? Flexbox seems to be more consistent across browsers.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -768,6 +798,12 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         });
     }
 
+    private handleComponentTransparent = (): void => {
```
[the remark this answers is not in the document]
nicholasrice: or `handleTransparencyToggleClick` - that would convey that this is an event handler. You could also plumb the event into the function definition `private handleTransparencyToggleClick = (e: React.MouseEvent<HTMLButtonElement): void => { ... }` to make that more illustrative.

packages/fast-development-site-react/src/components/site/index.tsx
chrisdholt: +1 on Nick's comment. On the opposite side of this we're using flexbox to accomplish the same thing. As a side note, I think the parent div here is floated, and we should avoid that since we can use flexbox there as well.

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -23,16 +25,25 @@ export interface IComponentWrapperProps<T> {
     singleRow?: boolean;
     designSystem: T;
     view: ComponentViewTypes;
+    transparent: boolean;
     onClick?: (activeIndex: number) => void;
 }
 
+/* tslint:disable */
+const checker: string = "url('data:image/gif;base64,R0lGODlhEAAQAPAAANfX1wAAACH5BAEAAAEALAAAAAAQABAAAAIfhG+hq4jM3IFLJhoswNly/XkcBpIiVaInlLJr9FZWAQAh/wtYTVAgRGF0YVhNUDw/eHBhY2tldCBiZWdpbj0i77u/IiBpZD0iVzVNME1wQ2VoaUh6cmVTek5UY3prYzlkIj8+Cjx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8iIHg6eG1wdGs9IkFkb2JlIFhNUCBDb3JlIDUuNi1jMTM4IDc5LjE1OTgyNCwgMjAxNi8wOS8xNC0wMTowOTowMSAgICAgICAgIj4KIDxyZGY6UkRGIHhtbG5zOnJkZj0iaHR0cDovL3d3dy53My5vcmcvMTk5OS8wMi8yMi1yZGYtc3ludGF4LW5zIyI+CiAgPHJkZjpEZXNjcmlwdGlvbiByZGY6YWJvdXQ9IiIvPgogPC9yZGY6UkRGPgo8L3g6eG1wbWV0YT4KPD94cGFja2V0IGVuZD0iciI/PgH//v38+/r5+Pf29fTz8vHw7+7t7Ovq6ejn5uXk4+Lh4N/e3dzb2tnY19bV1NPS0dDPzs3My8rJyMfGxcTDwsHAv769vLu6ubi3trW0s7KxsK+urayrqqmop6alpKOioaCfnp2cm5qZmJeWlZSTkpGQj46NjIuKiYiHhoWEg4KBgH9+fXx7enl4d3Z1dHNycXBvbm1sa2ppaGdmZWRjYmFgX15dXFtaWVhXVlVUU1JRUE9OTUxLSklIR0ZFRENCQUA/Pj08Ozo5ODc2NTQzMjEwLy4tLCsqKSgnJiUkIyIhIB8eHRwbGhkYFxYVFBMSERAPDg0MCwoJCAcGBQQDAgEAADs=')";
```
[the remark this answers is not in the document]
eljefe223: Unfortunately on my monitors I was seeing strange artifacting sticking with current implementation

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -217,6 +221,15 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
         alignItems: "center",
         justifyContent: "center"
     },
+    site_transparencyToggleButton: {
+        border: "none",
+        background: "none",
+        height: toPx(32),
+        width: toPx(32),
+        outline: "0",
+        verticalAlign: "middle",
```
chrisdholt: I'm not sure we still need verticalAlign middle here still. If we do - I think there are better solutions for it.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -217,6 +221,15 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
         alignItems: "center",
         justifyContent: "center"
     },
+    site_transparencyToggleButton: {
```
chrisdholt: can we add "cursor: pointer" here to ensure all browsers show that this is actionable?

chrisdholt: @eljefe223 making sure you saw this last one

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -49,6 +49,7 @@ export interface ISiteProps {
     frameworks?: FrameworkEnum | FrameworkEnum[];
     activeFramework?: FrameworkEnum;
     collapsed?: boolean;
+    transparent?: boolean;
```
janechu: I think the comment may have gotten lost somewhere but this should be more clear as to what is transparent, eg `componentBackgroundTransparent` like you have below.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -664,6 +684,12 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
     private renderInfoBarConfiguration(): JSX.Element {
         return (
             <div className={this.props.managedClasses.site_infoBarConfiguration}>
+                <button
+                    onClick={this.handleTransparencyToggleClick}
+                    className={this.props.managedClasses.site_transparencyToggleButton}
+                >
+                    <span dangerouslySetInnerHTML={{__html: glyphTransparency}}/>
```
janechu: I think we can just use the base64 string instead of having a glyph

packages/fast-development-site-react/src/components/site/index.tsx
janechu: as this is a new prop, please also add to the documentation.

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: Discussed offline we want to keep these separate for consistency

PR #595, opened 2018-06-27
fix(fast-development-site): correctly display component view border

### LINKED ISSUE #594: Examples view of component border is not displaying correctly
Examples view of component border is not displaying correctly

eljefe223:
Closes #594 
 

### DIFF
--- a/packages/fast-development-site-react/src/components/site/component-wrapper.tsx
+++ b/packages/fast-development-site-react/src/components/site/component-wrapper.tsx
@@ -44,11 +44,8 @@ const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSys
         background: checker
     },
     componentWrapperExamples: {
-        "&:last-child:not(:nth-child(3n + 3)), &:nth-child(3n + 1)": {
+        "&:nth-child(3n + 1), &:nth-child(3n + 2)": {
             borderRight: componentWrapperBorder,
-        },
-        "&:nth-child(3n + 3)": {
-            borderLeft: componentWrapperBorder
         },
     },
     componentWrapper__active: {
@@ -107,11 +104,13 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
             ? `${this.props.managedClasses.componentWrapper__transparent} ${classNames}`
             : classNames;
 
-        return this.props.active && this.props.view === ComponentViewTypes.examples
-            ? `${classNames} ${this.props.managedClasses.componentWrapperExamples} ${this.props.managedClasses.componentWrapper__active}`
-            : this.props.view === ComponentViewTypes.detail
-            ? `${classNames} ${this.props.managedClasses.componentWrapper__active}`
+        classNames = this.props.active
+            ? `${this.props.managedClasses.componentWrapper} ${this.props.managedClasses.componentWrapper__active}`
+            : this.props.managedClasses.componentWrapper;
+        return this.props.view === ComponentViewTypes.examples
+            ? `${classNames} ${this.props.managedClasses.componentWrapperExamples}`
             : classNames;
+
     }
 
     private handleClick = (): void => {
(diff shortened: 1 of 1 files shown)


packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -91,13 +88,14 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
     }
 
     private getClassNames(): string {
-        const classNames: string = this.props.managedClasses.componentWrapper;
+        const classNames: string = this.props.active
+            ? `${this.props.managedClasses.componentWrapper} ${this.props.managedClasses.componentWrapper__active}`
+            : this.props.managedClasses.componentWrapper ;
```
chrisdholt: can we break this up once more since you're making the change here?

```
const classNames: string = this.props.managedClasses.componentWrapper;

classNames = this.props.active ? `${classNames} ${this.props.managedClasses.componentWrapper__active}` : classNames;
```

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -34,12 +34,9 @@ const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSys
         borderBottom: componentWrapperBorder,
     },
     componentWrapperExamples: {
-        "&:last-child:not(:nth-child(3n + 3)), &:nth-child(3n + 1)": {
+        "&:nth-child(3n + 1), &:nth-child(3n + 2)": {
```
chrisdholt: Going to pull this down for visual confirmation to confirm localization doesn't break this.

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -91,13 +88,14 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
...
 
-        return this.props.active && this.props.view === ComponentViewTypes.examples
-            ? `${classNames} ${this.props.managedClasses.componentWrapperExamples} ${this.props.managedClasses.componentWrapper__active}`
-            : this.props.view === ComponentViewTypes.detail
-            ? `${classNames} ${this.props.managedClasses.componentWrapper__active}`
+        return this.props.view === ComponentViewTypes.examples
+            ? `${classNames} ${this.props.managedClasses.componentWrapperExamples}`
             : classNames;
+
     }
```
janechu: nit: extra line

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
chrisdholt: @eljefe223 I think this only fixes part of the problem. This ends up breaking in RTL, so we likely need to support localization here as well.

eljefe223: That is because this fix is in coordination with the other fix. You need to pull in my other RTL changes.

awentzel:
Just a little comment on the title of the PR.  Instead of saying what is broken I'd suggest always saying what the user gets.  This is helpful when users/contributors are reading the change logs.  For example, I'd suggest changing the title too "fix(fast-development-site): update component border style to ..."

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -107,10 +104,11 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
             ? `${this.props.managedClasses.componentWrapper__transparent} ${classNames}`
             : classNames;
 
-        return this.props.active && this.props.view === ComponentViewTypes.examples
-            ? `${classNames} ${this.props.managedClasses.componentWrapperExamples} ${this.props.managedClasses.componentWrapper__active}`
-            : this.props.view === ComponentViewTypes.detail
-            ? `${classNames} ${this.props.managedClasses.componentWrapper__active}`
+        classNames = this.props.active
+            ? `${this.props.managedClasses.componentWrapper} ${this.props.managedClasses.componentWrapper__active}`
+            : this.props.managedClasses.componentWrapper;
```
janechu: I think @chrisdholt previous comment still applies here, we are going to have a situation where if the component is active and transparent you will be overwriting and not including the transparent class. You can just replace `this.props.managedClasses.componentWrapper` on 108 and 109 with `classNames`.

PR #605, opened 2018-06-28
feat(fast-development-site): add functionality for titles to be collapsible on configure pane

### LINKED ISSUE #548: Update titles to be collapsible on configure pane.

eljefe223:
Closes #548 
 


chrisdholt:
@eljefe223 can you add the correlating "PR" label to this

packages/fast-form-generator-react/src/form/form-section.tsx
```diff
@@ -273,35 +274,18 @@ class FormSection extends React.Component<IFormSectionProps & IManagedClasses<IF
...
      * Gets the categories
      */
     private getCategories(formItemParameters: IFormItemParameters[]): JSX.Element[] {
         const categories: JSX.Element[] = [];
         const categoryParams: IFormCategories[] = getCategoryParams(formItemParameters, this.props.orderByPropertyNames);
 
-        for (const category of categoryParams) {
+        for (let i: number = 0, categoryParamLength: number = categoryParams.length; i < categoryParamLength; i++) {
             const categoryFormItems: JSX.Element[] = [];
+            const contentId: string = uniqueId(this.convertToHyphenated(categoryParams[i].title));
```
janechu: Is it necessary to use a prefix?

packages/fast-form-generator-react/src/form/form-section.tsx
```diff
@@ -314,13 +298,24 @@ class FormSection extends React.Component<IFormSectionProps & IManagedClasses<IF
             }
 
             categories.push(
-                this.generateFormItemCategories(categoryFormItems, category.title)
+                <FormCategory
+                    id={contentId}
+                    key={i}
+                    Expandable={true}
+                    isExpanded={true}
```
janechu: Why do we pass the expanded as a prop if its always true? Why are we also passing expandable as true? Should the user be controlling this from a passed prop instead? I would assume that this should be part of the `IFormOrderByPropertyNamesCategories` interface which I would expect to be something like `expandable` and `expanded` which are both optional.

packages/fast-form-generator-react/src/form/form-category.style.ts
```diff
@@ -0,0 +1,56 @@
...
+        "&:hover": {
+            cursor: "pointer"
+        },
+        "&::after": {
+            content: `url('data:image/svg+xml;base64,${dropdownArrow}')`,
+            fill: "white",
+            position: "absolute",
+            right: toPx(11),
+            top: toPx(11),
+            verticalAlign: "middle",
```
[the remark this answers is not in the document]
eljefe223: No it is not. Removing

packages/fast-form-generator-react/src/form/form-section.tsx
eljefe223: Expanded is not always true if you don't set it that <FormCategory/> would be collapsed. You need to explicitly set isExpanded

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Passes the category title
+     */
+    title: string;
+
+    /**
+     * Is the category expandable
+     */
+    Expandable?: boolean;
```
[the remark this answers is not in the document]
chrisdholt: nit: lowercase* because one word

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+     */
+    id?: string;
+}
+
+export interface IFormCategoryState {
+    isCategoryExpanded: boolean;
+}
+
+/**
+ * Schema form component definition
```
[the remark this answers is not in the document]
chrisdholt: +1

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+            : this.props.managedClasses.formCategory__collapsed
+            : this.props.managedClasses.formCategory;
+
+        return classNames;
+    }
+
+    private renderHeader(): JSX.Element {
+        const header: JSX.Element = this.props.Expandable
+        ? {...this.renderHeaderButton()}
+        : {...this.renderHeaderTitle()};
```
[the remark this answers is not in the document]
chrisdholt: Instead of creating a single function here, can we just reference these inline in place?

```
{ this.props.expandable ? this.renderHeaderButton() : this.renderHeaderTitle() }
```

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+     */
+    Expandable?: boolean;
+
+    /**
+     * Is the category expanded
+     */
+    isExpanded?: boolean;
+
+    /**
+     * Passes the id
```
nicholasrice: Can you make the id comment a bit more descriptive? - the prop isn't actually doing any passing itself

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+ * @extends React.Component
+ */
+/* tslint:disable-next-line */
+class FormCategory extends React.Component<IFormCategoryProps & IManagedClasses<IFormCategoryClassNameContract>, IFormCategoryState> {
+
+    constructor(props: IFormCategoryProps & IManagedClasses<IFormCategoryClassNameContract>) {
+        super(props);
+
+        if (this.props.Expandable) {
+            this.state = {
```
nicholasrice: You probably want to set the state regardless so that it has the expected values (and is not undefined) - you can change to `this.state = { isCategoryExpanded: !!this.props.isExpanded };` to do this and remove the if statement. The `!!` return a boolean even if `this.props.isExpanded` is undefined.

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+     */
+    public render(): JSX.Element {
+        // Exit if the array is only one item long and that item is null or undefined
+        if (this.props.categoryItem.length < 1 && this.props.categoryItem[0] === null ||
+            this.props.categoryItem.length < 1 && this.props.categoryItem[0] === undefined) {
+            return;
+        }
+
+        return (
+            <div key={this.props.title}>
```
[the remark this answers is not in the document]
nicholasrice: Except for id is currently optional - should it be required?

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private generateContainerAttributes(): React.HtmlHTMLAttributes<HTMLDivElement> {
+        if (!this.props.Expandable) {
+            return;
+        }
+
+        const attributes: Partial<React.HtmlHTMLAttributes<HTMLDivElement>> = {};
+        attributes["aria-hidden"] = !this.state.isCategoryExpanded;
+        attributes.id = this.props.id;
```
nicholasrice: it seems odd to me that we would only apply the id if the component is also expandable - maybe just apply the id regardless?

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+        );
+    }
+
+    private generateContainerAttributes(): React.HtmlHTMLAttributes<HTMLDivElement> {
+        if (!this.props.Expandable) {
+            return;
+        }
+
+        const attributes: Partial<React.HtmlHTMLAttributes<HTMLDivElement>> = {};
+        attributes["aria-hidden"] = !this.state.isCategoryExpanded;
```
nicholasrice: If you move the id to the render function, you could also move the aria-hidden logic into the render function and remove the `generateContainerAttributes` method all together.

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,136 @@
...
+            isCategoryExpanded: !this.state.isCategoryExpanded
+        });
+    }
+
+    private getClassNames(): string {
+        const classNames: string = this.props.Expandable
+            ? this.state.isCategoryExpanded
+            ? this.props.managedClasses.formCategory
+            : this.props.managedClasses.formCategory__collapsed
+            : this.props.managedClasses.formCategory;
```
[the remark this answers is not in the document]
nicholasrice: where does `null` come from? If returning `null` is possible you should also add it to the type definition of the function - `private getClassNames(): string | null`

packages/fast-form-generator-react/src/form/form-category.tsx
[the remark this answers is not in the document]
nicholasrice: +1 for @chrisdholt suggestion.

packages/fast-form-generator-react/app/components/weight-properties/weight-properties.config.ts
```diff
@@ -9,6 +9,7 @@ export const weightPropertiesWeight: IFormOrderByPropertyNamesProps = {
     categories: [
         {
             title: "Content",
+            expandable: true,
```
chrisdholt: @eljefe223 do we not need to pass id here as a prop?

packages/fast-form-generator-react/src/form/form-section.tsx
```diff
@@ -89,7 +90,7 @@ class FormSection extends React.Component<IFormSectionProps & IManagedClasses<IF
         this.state = {
             schema: schemaState,
             oneOfAnyOf: oneOfAnyOfState,
-            sections: getSchemaSubsections({schema: schemaState, oneOfAnyOf: oneOfAnyOfState}, this.props)
+            sections: getSchemaSubsections({schema: schemaState, oneOfAnyOf: oneOfAnyOfState}, this.props),
```
chrisdholt: nit: can we remove this?

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,117 @@
...
+                </div>
+            </div>
+        );
+    }
+
+    private generateContainerAttributes(): React.HtmlHTMLAttributes<HTMLDivElement> {
+        if (!this.props.expandable) {
+            return;
+        }
+
```
nicholasrice: I still think the ID should always be applied, regardless of the value of `this.props.expandable`

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,118 @@
...
+                >
+                    {this.props.categoryItem}
+                </div>
+            </div>
+        );
+    }
+
+    private generateContainerAttributes(): React.HtmlHTMLAttributes<HTMLDivElement> {
+        const attributes: Partial<React.HtmlHTMLAttributes<HTMLDivElement>> = {};
+        attributes.id = this.props.id;
```
chrisdholt: This could be done in one definition:

```
const attributes: Partial<React.HtmlHTMLAttributes<HTMLDivElement>> = {
    id: this.props.id
};
```

packages/fast-form-generator-react/src/form/form-category.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+        if (!this.props.expandable) {
+            return attributes;
+        }
+
+        attributes["aria-hidden"] = !this.state.expanded;
```
chrisdholt: As an FYI, we could return once from this function by changing to check for expandable:

```
if (this.props.expandable) {
    attributes["aria-hidden"] = !this.state.expanded;
}
```

PR #606, opened 2018-06-28
fix(fast-development-site): update background image size

eljefe223:


packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -39,7 +39,7 @@ const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSys
         borderBottom: componentWrapperBorder,
     },
     componentWrapper__transparent: {
-        background: `transparent ${toPx(8)}/${toPx(8)} ${checker} repeat`
+        background: `transparent 0 ${toPx(16)}/${toPx(16)} ${checker} repeat`
```
janechu: There is no need to state repeat as it is the initial value see [spec](https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat).

I don't see the value of transparent being specified when the background will fill the entire image.

eljefe223: The gif is already 16 x 16. I simply set the background to point to the gif and removed everything else.

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -41,7 +41,7 @@ const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSys
         borderBottom: componentWrapperBorder,
     },
     componentWrapper__transparent: {
-        background: `transparent ${toPx(8)}/${toPx(8)} ${checker} repeat`
+        background: `${checker}`
```
janechu: You don't need template literals here since there are no calculations inside to be made, it can just be `background: checker`.

Issue #599, opened 2018-06-28
The + icon is misaligned from the text in the building blocks add children section


janechu:
The + icon is misaligned in Edge browser, there appears to be no top value.

![edge-but-form-generator-misaligned-children.png](https://images.zenhubusercontent.com/59557d9061b4ca3b0a72c253/d1e5bb4b-0cac-430b-a667-a2f590d757c0)

PR #626, opened 2018-07-02
fix(fast-form-generator-react): update + glyph positioning

nalogiudice:
Closes [#599](https://github.com/Microsoft/fast-dna/issues/599)

<body>

<footer>
 

### DIFF
--- a/packages/fast-form-generator-react/src/utilities/form-input.style.ts
+++ b/packages/fast-form-generator-react/src/utilities/form-input.style.ts
@@ -305,6 +305,7 @@ export function applyPopupMenuStyles(): ICSSRules<{}> {
 export const thickLine: ICSSRules<{}> = {
     "& > span": {
         position: "relative",
+        paddingTop: toPx(3),
         paddingBottom: toPx(3),
         display: "block",
         ...ellipsis(),
@@ -467,7 +468,8 @@ export function applyAddItemStyle(): ICSSRules<{}> {
             width: toPx(16),
             height: toPx(16),
             left: "0",
-            top: toPx(11)
+            top: "50%",
+            transform: "translateY(-50%)"
         }
     };
 }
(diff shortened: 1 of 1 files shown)


packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -466,7 +466,8 @@ export function applyAddItemStyle(): ICSSRules<{}> {
             background: pinkPlus,
             width: toPx(16),
             height: toPx(16),
-            left: "0"
+            left: "0",
+            top: toPx(11)
```
chrisdholt: fixing to 11px threw this off in chrome for me when testing. If we need to absolutely position this for all browsers, let's consider a more standard approach:

```
top: "50%",
transform: "translateY(-50%)"
```

awentzel:
Scope should include hyphens if a package name per documentation

packages/fast-form-generator-react/src/utilities/form-input.style.ts
nalogiudice: Wait really? I saw no change in Chrome positioning at all.

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -466,7 +467,9 @@ export function applyAddItemStyle(): ICSSRules<{}> {
             background: pinkPlus,
             width: toPx(16),
             height: toPx(16),
-            left: "0"
+            left: "0",
+            top: "50%",
+            transform: "translateY(-50%)"
```
eljefe223: I assume this guarantees that the glyph is positioned center vertically in its parent?

nalogiudice: Correct.

PR #630, opened 2018-07-03
feat(fast-components-styles-msft): set default styles in the case no design system is provided

### LINKED ISSUE #535: Create defaults for styles that receive data from a design system provider
User should always see a style applied regardless of the presence a design system provider wrapper.

Currently we are not using fallbacks for our JSS msft styles. In case there is no config or the property has not been provided through the config we should have a default fallback built in.

The best solution might be to use the `get` utility from lodash-es to check.

Example for returning a property of config called `foregroundColor`:
```
color: (config) => {
    return get({config}, "config.foregroundColor") || designSystemDefaults.foregroundColor;
}
```

chrisdholt:
<body>
closes #535 
<footer>
 

### DIFF
--- a/packages/fast-components-styles-msft/src/text-field/index.ts
+++ b/packages/fast-components-styles-msft/src/text-field/index.ts
@@ -1,13 +1,18 @@
-import { IDesignSystem } from "../design-system";
+import designSystemDefaults, { IDesignSystem } from "../design-system";
 import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
 import { ITextFieldClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 import { toPx } from "@microsoft/fast-jss-utilities";
+import { isEmpty } from "lodash-es";
 import { applyType } from "../utilities/typography";
 import { fontWeight } from "../utilities/fonts";
 import { applyMixedColor } from "../utilities/colors";
 
 function applyTextFieldMixedColor(mixValue?: number): (config: IDesignSystem) => string {
-    return (config: IDesignSystem): string => applyMixedColor(config.foregroundColor, config.backgroundColor, mixValue);
+    return (config: IDesignSystem): string => {
+        config = isEmpty(config) ? designSystemDefaults : config;
+
+        return applyMixedColor(config.foregroundColor, config.backgroundColor, mixValue);
+    };
 }
 
 const styles: ComponentStyles<ITextFieldClassNameContract, IDesignSystem> = {
@@ -28,7 +33,7 @@ const styles: ComponentStyles<ITextFieldClassNameContract, IDesignSystem> = {
         },
         "&:disabled": {
             background: (config: IDesignSystem): string => {
-                return config.backgroundColor;
+                return config.backgroundColor || designSystemDefaults.backgroundColor;
             },
             borderColor: applyTextFieldMixedColor(0.863),
             color: applyTextFieldMixedColor(0.80),
--- a/packages/fast-components-styles-msft/src/toggle/index.ts
+++ b/packages/fast-components-styles-msft/src/toggle/index.ts
@@ -1,21 +1,27 @@
-import { IDesignSystem } from "../design-system";
+import designSystemDefaults, { IDesignSystem } from "../design-system";
 import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
 import { applyLocalizedProperty, toPx } from "@microsoft/fast-jss-utilities";
 import { typeRamp } from "../utilities/typography";
 import { IToggleClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { isEmpty } from "lodash-es";
 import * as Chroma from "chroma-js";
 import { Direction } from "@microsoft/fast-application-utilities";
 
 function setFloatAndClear(clearFloat: boolean = true): ICSSRules<IDesignSystem> {
     return {
-        float: (config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right",
-        clear: clearFloat ? ((config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right") : null
+        float: (config: IDesignSystem): string => {
+            return (isEmpty(config) || config.direction === Direction.ltr) ? "left" : "right";
+        },
+        clear: (config: IDesignSystem): string => {
+            return clearFloat ? (isEmpty(config) || config.direction === Direction.ltr ? "left" : "right") : null;
+        }
     };
 }
 
-/* tslint:disable:max-line-length */
+/* tslint:disable-next-line */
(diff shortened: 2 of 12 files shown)


packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -1,13 +1,18 @@
...
 import { toPx } from "@microsoft/fast-jss-utilities";
+import { isEmpty } from "lodash-es";
 import { applyType } from "../utilities/typography";
 import { fontWeight } from "../utilities/fonts";
 import { applyMixedColor } from "../utilities/colors";
 
 function applyTextFieldMixedColor(mixValue?: number): (config: IDesignSystem) => string {
-    return (config: IDesignSystem): string => applyMixedColor(config.foregroundColor, config.backgroundColor, mixValue);
+    return (config: IDesignSystem): string => {
+        config = isEmpty(config) ? designSystemDefaults : config;
```
janechu: config appears to be passed and then re-assigned, I think to be safe we would want to instead not reassign this and do something more like:

```
const colorConfig: string[] = [
    isUndefined(config.foregroundColor) ? designSystemDefaults.foregroundColor : config.foregroundColor,
    isUndefined(config.backgroundColor) ? designSystemDefaults.backgroundColor : config.backgroundColor,
    mixValue
];

return applyMixedColor(...colorConfig);
```

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -1,21 +1,27 @@
...
+import { isEmpty } from "lodash-es";
 import * as Chroma from "chroma-js";
 import { Direction } from "@microsoft/fast-application-utilities";
 
 function setFloatAndClear(clearFloat: boolean = true): ICSSRules<IDesignSystem> {
     return {
-        float: (config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right",
-        clear: clearFloat ? ((config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right") : null
+        float: (config: IDesignSystem): string => {
+            return (isEmpty(config) || config.direction === Direction.ltr) ? "left" : "right";
```
janechu: As before, I don't think isEmpty is safe, it may have other values but still be missing config.direction. Here I think you can just check for config.

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -1,21 +1,27 @@
...
+            return clearFloat ? (isEmpty(config) || config.direction === Direction.ltr ? "left" : "right") : null;
+        }
     };
 }
 
-/* tslint:disable:max-line-length */
+/* tslint:disable-next-line */
 const styles: ComponentStyles<IToggleClassNameContract, IDesignSystem> = (config: IDesignSystem): ComponentStyleSheet<IToggleClassNameContract, IDesignSystem> => {
-/* tslint:enable:max-line-length */
+    config = isEmpty(config) ? designSystemDefaults : config;
```
janechu: Again, I don't think we should be re-assigning a passed value. We should assign this as safely as possible so using a new variable is fine.

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -34,18 +40,16 @@ const styles: ComponentStyles<IToggleClassNameContract, IDesignSystem> = (config
...
         },
         toggle_wrapper: {
-            position: "relative"
+            position: "relative",
+            marginTop: "0",
+            ...setFloatAndClear()
+        },
+        toggle_statusLabel: {
+            ...setFloatAndClear(false),
+            [applyLocalizedProperty("margin-left", "margin-right", config.direction)]: toPx(5)
```
janechu: I think we are mixing and matching our tasks here, I would think this PR would only be about adding the defaults for the design system but it appears to include toggle fixes as well, can we address this in another PR or is there a reason they need to be included in this one?

chrisdholt: The styling will be broken without passing a class here. Why - I don't know, but I can remove. We just need to accept that there will be a gap in coverage for default styles for toggle. With a class, things pass as expected; with previous changes, it fails to pass the styles.

chrisdholt: So, by not assigning a class here, the task is complete but broken for toggle. I'm fine with that if need be.

packages/fast-components-styles-msft/src/text-field/index.ts
chrisdholt: I think the task may have been a bit misleading. 
```
In case there is no wrapping design system config we should have a default fallback built in.
```
First, let me explain the reasoning for addressing "config".

Config is only being reassigned if there is no design system provided (per the current task description). The reason that I'm using isEmpty() here rather than isUndefined() is that in the event of a design system provider or design system not being present, "config" comes back as an empty object. isEmpty() will check undefined, null, and an empty object and return true - which is why I chose to use it.

Based on the current task description (if no design system) it seemed more convenient and cleaner to pass defaults via config rather than pass individual params in the event there is no design system.

If there is a value missing from the design system - will there not be compile errors? Currently everything in the interface we expect is required, so if I define a partial instance, I get compile errors.

Where there are no errors is where there is no concept of a wrapping design system.

I can update to address every individual instance, I just need to know how to update the task description so that there is alignment between the ask and the final PR.

packages/fast-components-styles-msft/src/toggle/index.ts
chrisdholt: Let's land the above note on line 10 of text-field re: isEmpty() prior to digging in on this.

packages/fast-components-styles-msft/src/toggle/index.ts
chrisdholt: Let's land the convo above prior to any follow-up on this

packages/fast-components-styles-msft/src/text-field/index.ts
janechu: Couple notes:

1. You could have a design system that only has one value, but not all the ones you want. eg. Here you need `foregroundColor`, what if config is just `{backgroundColor: "purple"}`. Then you will be passing undefined as the `foregroundColor` as it is missing.
2. I am uncomfortable with re-assigning config as in JavaScript when you pass an object it passes a value taht is a reference to an object, you will risk manipulating config at it's source.

janechu: I would think that you want to check for the specific value you are looking for, and if not fallback to the default value, rather than replace the entire config which may or may not exist.

chrisdholt: 1. Valid and accurate but I'm still not sure that you won't run into compile issues with a partial design system when it exists.
2. Fair point; expectation was that config does not exist and chose isEmpty due to it technically always being passed as an empty object.

Either way with 1 & 2, I think the confusion came down to the details in the task. I read it as a lack of a design system altogether and considering the potential compile issues due to a partial design system being invalid, I set config to a default value. 

Based on this, it seems that what we want is:

"In the event that a value passed is not defined, set a fallback."

I'll abandon the PR for now and resubmit based on that scenario.

PR #633, opened 2018-07-04
feat(fast-components-styles-msft): set default config values as a fallback if there is a missing or incomplete design system

### LINKED ISSUE #535: Create defaults for styles that receive data from a design system provider

chrisdholt:
closes #535 

<body>

<footer>
 

### DIFF
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -1,8 +1,9 @@
-import { IDesignSystem } from "../design-system";
+import designSystemDefaults, { IDesignSystem } from "../design-system";
 import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
 import { IButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 import { IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 import { applyLocalizedProperty, Direction, localizeSpacing, toPx } from "@microsoft/fast-jss-utilities";
+import { get } from "lodash-es";
 import { applyType } from "../utilities/typography";
 import { applyMixedColor } from "../utilities/colors";
 import * as Chroma from "chroma-js";
@@ -15,14 +16,14 @@ function applyTransaprentBackplateStyles(): ICSSRules<IDesignSystem> {
             ...applyTransaprentBackground()
         },
         color: (config: IDesignSystem): string => {
-            return config.brandColor;
+            return get(config, "brandColor") || designSystemDefaults.brandColor;
         },
         "&:disabled span::before, &[aria-disabled] span::before": {
             background: "transparent"
         },
         "&:focus span::before, &:active span::before, &:hover span::before": {
             background: (config: IDesignSystem): string => {
-                return config.brandColor;
+                return get(config, "brandColor") || designSystemDefaults.brandColor;
             }
         },
         "&:disabled, &[aria-disabled]": {
@@ -44,7 +45,10 @@ function applyTransaprentBackground(): ICSSRules<IDesignSystem> {
 function applyPropertyDrivenColor(incomingProperty: string, mixValue?: number, alpha?: number): ICSSRules<IDesignSystem> {
     return {
         [incomingProperty]: (config: IDesignSystem): string => {
-            return applyMixedColor(config.foregroundColor, config.backgroundColor, mixValue, alpha);
+            const backgroundColor: string = get(config, "backgroundColor") || designSystemDefaults.backgroundColor;
+            const foregroundColor: string = get(config, "foregroundColor") || designSystemDefaults.foregroundColor;
+
+            return applyMixedColor(foregroundColor, backgroundColor, mixValue, alpha);
         }
     };
 }
@@ -70,7 +74,7 @@ const styles: ComponentStyles<IMSFTButtonClassNameContract, IDesignSystem> = (co
             verticalAlign: "bottom",
             transition: "all 0.2s ease-in-out",
             color: (): string => {
-                return config.backgroundColor;
+                return get(config, "backgroundColor") || designSystemDefaults.backgroundColor;
             },
             ...applyPropertyDrivenColor("backgroundColor", 0.46),
             "&:hover": {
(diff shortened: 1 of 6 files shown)


packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -15,14 +16,14 @@ function applyTransaprentBackplateStyles(): ICSSRules<IDesignSystem> {
             ...applyTransaprentBackground()
         },
         color: (config: IDesignSystem): string => {
-            return config.brandColor;
+            return get(config, "brandColor") || designSystemDefaults.brandColor;
```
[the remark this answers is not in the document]
chrisdholt: Correct. I just logged out these two while within the context of the development site (design system exists):

1. `get(config, "backgroundColor");` - resolves with design system property
2. `get(config, "config.backgroundColor")` - always returns undefined

If number 2 were the case, our usage of get with managed classes would likely be incorrect.

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -1,21 +1,28 @@
...
+import { get } from "lodash";
 import * as Chroma from "chroma-js";
 import { Direction } from "@microsoft/fast-application-utilities";
 
 function setFloatAndClear(clearFloat: boolean = true): ICSSRules<IDesignSystem> {
     return {
-        float: (config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right",
-        clear: clearFloat ? ((config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right") : null
+        float: (config: IDesignSystem): string =>
+            get(config, "direction") || designSystemDefaults.direction === Direction.ltr ? "left" : "right",
```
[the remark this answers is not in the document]
chrisdholt: Updated.

packages/fast-components-styles-msft/src/button/index.ts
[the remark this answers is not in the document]
chrisdholt: Feel free to pull down and confirm. You can comment out the dev site and just render the components without a design system as a way to test.

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -1,21 +1,38 @@
...
+import designSystemDefaults, { IDesignSystem } from "../design-system";
 import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
 import { applyLocalizedProperty, toPx } from "@microsoft/fast-jss-utilities";
 import { typeRamp } from "../utilities/typography";
 import { IToggleClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { get } from "lodash";
 import * as Chroma from "chroma-js";
 import { Direction } from "@microsoft/fast-application-utilities";
 
 function setFloatAndClear(clearFloat: boolean = true): ICSSRules<IDesignSystem> {
```
nicholasrice: I think you should separate float and clear. It isn't super intuitive from a readers perspective as to what this does when any particular param is passed.

If you make two functions (each one does one thing) and uses them when needed you will increase readability and reduce code complexity.

Also, since these are localizing floats and clears, we should probably make that clear with the function name.

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -68,60 +85,60 @@ const styles: ComponentStyles<IToggleClassNameContract, IDesignSystem> = (config
...
-                        return Chroma(config.brandColor).alpha(0.8).css();
+                        return Chroma(brandColor).alpha(0.8).css();
                     },
                     borderColor: (): string => {
-                        return Chroma(config.brandColor).alpha(0.8).css();
+                        return Chroma(brandColor).alpha(0.8).css();
                     }
                 },
                 "&:focus": {
                     borderColor: (): string => {
```
nicholasrice: none of these need to be functions anymore because the config is already available - you can just assign the value to the key.

packages/fast-components-styles-msft/src/toggle/index.ts
chrisdholt: Updated based on this feedback.

packages/fast-components-styles-msft/src/toggle/index.ts
chrisdholt: updated this across all other stylesheet instances (toggle, checkbox, button)

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -1,21 +1,39 @@
...
+import { get } from "lodash";
 import * as Chroma from "chroma-js";
 import { Direction } from "@microsoft/fast-application-utilities";
 
-function setFloatAndClear(clearFloat: boolean = true): ICSSRules<IDesignSystem> {
+function setLocalizedFloat(): ICSSRules<IDesignSystem> {
     return {
-        float: (config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right",
-        clear: clearFloat ? ((config: IDesignSystem): string => config.direction === Direction.ltr ? "left" : "right") : null
+        float: (config: IDesignSystem): string => {
```
nicholasrice: what if I want to float something right in LTR? Maybe just a generic "applyLocalizedValue" would work for both of these - does that exist yet?:

```
float: applyLocalizedValue("left", "right"),
clear: applyLocalizedValue("left", "right")
```

chrisdholt: yeah - this makes way more sense. At this point, the function really doesn't provide a lot of value.

chrisdholt: @nicholasrice updated with applyLocalizedProperty from jss-utils package.

packages/fast-components-styles-msft/src/button/index.ts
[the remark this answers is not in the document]
janechu: Sure, that's why my earlier suggestion was `get({config}, "config.backgroundColor")`.

PR #636, opened 2018-07-05
feat(fast-css-editor-react): add default editor component and position component

### LINKED ISSUE #597: Create a package for an editor for JSS styles
The user should be able to see a range of UI components that will allow them to update/add JSS values.

There should be a package that allows the mapping of provided JSS styles to a component or set of components that allow the user to edit them.

This should exist as standalone components (eg, positioning) and also as a default component that uses all components.

The package should:
- Use Jest snapshots
- Have adequate data examples
- Have adequate documentation of the components and their uses
- Have a default export JSSEditor that uses all of the individual components
- Have a positioning component that allows the user to select between static and absolute position values, provides a UI for choosing/showing which left/right/top/bottom values are available or selected and inputs for changing them.
- A callback should be provided for all exports to allow the user to update their data which will them be re-mapped to the JSS editor UI.

janechu:
Resolves #597

### DIFF
--- a/packages/fast-css-editor-react/app/app.tsx
+++ b/packages/fast-css-editor-react/app/app.tsx
@@ -28,28 +28,38 @@ export enum EditorView {
 }
 
 class App extends React.Component<{}, IAppState> {
-    public readonly state: IAppState = {
-        editorDetail: {
-            position: {
-                position: PositionValue.static
-            }
-        },
-        editorExample: {
-            position: {
+    public state: IAppState;
+
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            editorDetail: {
+                position: {
+                    position: PositionValue.static
+                },
+                onChange: this.handleEditorDetailUpdate
+            },
+            editorExample: {
+                position: {
+                    position: PositionValue.absolute,
+                    left: 4,
+                    top: 12
+                },
+                onChange: this.handleEditorExampleUpdate
+            },
+            positionDetail:{
+                position: PositionValue.static,
+                onChange: this.handlePositionDetailUpdate}
+            ,
+            positionExample: {
                 position: PositionValue.absolute,
-                left: 4,
-                top: 12
+                left: 1,
+                top: 5,
+                onChange: this.handlePositionExampleUpdate
             }
-        },
-        positionDetail: {
-            position: PositionValue.static
-        },
-        positionExample: {
-            position: PositionValue.absolute,
-            left: 1,
-            top: 5
-        }
-    };
+        };
+    }
 
     public render(): JSX.Element {
         return (
@@ -101,15 +111,6 @@ class App extends React.Component<{}, IAppState> {
                 </SiteCategory>
             </Site>
         );
-    }
-
-    public componentWillMount(): void {
-        this.setState({
-            editorDetail: Object.assign({}, this.state.editorDetail, {onChange: this.handleEditorDetailUpdate}),
-            editorExample: Object.assign({}, this.state.editorExample, {onChange: this.handleEditorExampleUpdate}),
-            positionDetail:  Object.assign({}, this.state.positionDetail, {onChange: this.handlePositionDetailUpdate}),
-            positionExample: Object.assign({}, this.state.positionExample, {onChange: this.handlePositionExampleUpdate})
-        });
     }
 
     private handleEditorDetailUpdate = (updateEditorDetail: any): void => {
--- a/packages/fast-css-editor-react/app/app.tsx
+++ b/packages/fast-css-editor-react/app/app.tsx
@@ -20,6 +20,13 @@ export interface IAppState {
     positionExample: ICSSPositionProps;
 }
 
+export enum EditorView {
+    detail = "editorDetail",
+    example = "editorExample",
+    positionDetail = "positionDetail",
+    positionExample = "positionExample"
+}
+
 class App extends React.Component<{}, IAppState> {
     public readonly state: IAppState = {
         editorDetail: {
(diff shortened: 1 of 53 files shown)


packages/fast-css-editor-react/package.json
```diff
@@ -0,0 +1,96 @@
...
+    "webpack-cli": "^2.0.11",
+    "webpack-dev-server": "^3.1.0"
+  },
+  "peerDependencies": {
+    "lodash-es": "^4.0.0",
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0"
+  },
+  "dependencies": {
+    "@microsoft/fast-application-utilities": "^1.8.0",
```
chrisdholt: Can any of the fast-* dependencies be peer deps?

packages/fast-css-editor-react/src/position/index.tsx
```diff
@@ -0,0 +1,173 @@
+import * as React from "react";
+import { IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+import manageJss, { IJSSManagerProps } from "@microsoft/fast-jss-manager-react";
+import style, { ICSSPositionClassNameContract } from "./position.style";
+
+export enum PositionValue {
+    static = "static",
+    absolute = "absolute"
```
chrisdholt: are we currently only providing partial values here for a reason? Otherwise, I would expect all currently supported position values to be valid enums.

packages/fast-css-editor-react/webpack.config.js
```diff
@@ -0,0 +1,56 @@
...
+    ],
+    resolve: {
+        extensions: [".js", ".tsx", ".ts", ".json"],
+        alias: {
+            fbjs: path.resolve('./node_modules/fbjs'),
+            'lodash-es': path.resolve('./node_modules/lodash-es'),
+            react: path.resolve('./node_modules/react'),
+            'react-dom': path.resolve('./node_modules/react-dom'),
+        }
+    },
```
chrisdholt: Does this get run through webpack for production independently of an app? If we don't need to package an "App" instance of this, can we remove all the alias resoltion from 40-48?

packages/fast-css-editor-react/src/position/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+    onChange: (positionValues: any) => void;
+}
+
+class CSSPosition extends React.Component<ICSSPositionProps & IManagedClasses<ICSSPositionClassNameContract>, {}> {
+    private positionKey: string = "position";
+
+    public render(): JSX.Element {
+        return (
+            <div className={this.props.managedClasses.cssPosition}>
+                <span className={this.props.managedClasses.cssPosition_selectContainer}>
```
chrisdholt: Not a blocker, but <span> here seems better served as a <div>. I could see using the <span> if we want it to function as an inline-element here, but since we are applying styles to it, it seems out of place. 

Total nit. Not blocking. More curious than anything.

packages/fast-css-editor-react/src/position/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+                updatedProps[prop] = this.props[prop];
+            }
+        });
+
+        updatedProps[updatedPropKey] = parseInt(updatedPropValue, 10);
+
+        return updatedProps;
+    }
+
+    private getExcludedLocation(updatedPropKey: Location): Location {
```
chrisdholt: Do we need to add a simple comment above these functions to provide a little insight into their purpose? Obviously not full JSDoc syntax, but similar to our approach with all other packages?

packages/fast-css-editor-react/app/app.tsx
```diff
@@ -0,0 +1,145 @@
...
+                    </SiteCategory>
+                </SiteCategory>
+            </Site>
+        );
+    }
+
+    public componentWillMount(): void {
+        this.setState({
+            editorDetail: Object.assign({}, this.state.editorDetail, {onChange: this.handleEditorDetailUpdate}),
+            editorExample: Object.assign({}, this.state.editorExample, {onChange: this.handleEditorExampleUpdate}),
```
nicholasrice: Any reason you're not doing this in the constructor? Also, state isn't really readonly though you've assigned it as such. I think you should do all of the initial state setup in the constructor.

packages/fast-css-editor-react/src/position/README.md
```diff
@@ -0,0 +1,36 @@
+# CSSPosition
+The `CSSPosition` component shows the CSS position value as a select, it will also provide a callback for updated this value. When set to "absolute" it will include a UI for allowing the user to select left/right/top/bottom and enter numbers for pixel positioning. The components position will default to static if not set.
```
nicholasrice: it will also provide a callback for *updating* this value.

packages/fast-css-editor-react/src/position/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+    public render(): JSX.Element {
+        return (
+            <div className={this.props.managedClasses.cssPosition}>
+                <span className={this.props.managedClasses.cssPosition_selectContainer}>
+                    <select
+                        className={this.props.managedClasses.cssPosition_selectContainer_select}
+                        data-location={this.positionKey}
+                        onChange={this.handleOnChange}
+                        value={this.props.position ? this.props.position : PositionValue.static}
+                    >
```
nicholasrice: why not relative and fixed?

janechu: Per @Falkicon we are only doing static and absolute at this time.

packages/fast-css-editor-react/src/position/index.tsx
janechu: Not sure I understand your comment?

packages/fast-css-editor-react/src/position/index.tsx
janechu: they will be added as we decide which ones we will light up

packages/fast-css-editor-react/webpack.config.js
janechu: Yes, we need this for our build to check on the app.

packages/fast-css-editor-react/src/position/index.tsx
chrisdholt: Just curious on the choice of wrapping select in a span here if it's not inline content

packages/fast-css-editor-react/src/position/position.style.ts
```diff
@@ -0,0 +1,91 @@
...
+    cssPosition: {
+    },
+    cssPosition_selectContainer: {
+        display: "flex",
+        position: "relative",
+        "&::before, &::after": {
+            top: toPx(12),
+            width: toPx(1),
+            height: toPx(10),
+            content: "''",
```
nalogiudice: I think you may have to do it like" `content: "\"\""`

janechu: No need, this is how we are currently doing it.

nalogiudice: It's actually mixed. @nicholasrice changed some JSS to this, so maybe he can shed some light on it

PR #640, opened 2018-07-05
feat(fast-jss-utilities): add a utilitiy to convert hex values to rgb(a)

### LINKED ISSUE #639: Add a utility to convert hex values to rgb or rgba
Add a utility to convert hex values to either RGB or RGBA color values.

Users should be able to pass design system values with an opacity and return a valid color.

chrisdholt:
<body>
closes #639 
<footer>
 

### DIFF
--- a/packages/fast-jss-utilities/src/color.ts
+++ b/packages/fast-jss-utilities/src/color.ts
@@ -1,0 +1,13 @@
+export function convertHexToRgb(hex: string, alpha?: number): string {
+    if (!hex || !hex.match(`^#?(([0-9a-fA-F]{2}){3}|([0-9a-fA-F]){3})$`)) {
+        throw new Error("Received hex value is not valid");
+    }
+
+    const newHex: string = hex.charAt(0) === "#" ? hex.substring(1) : hex;
+
+    const red: number = parseInt(newHex.length === 3 ? newHex.slice(0, 1).repeat(2) : newHex.slice(0, 2), 16);
+    const green: number = parseInt(newHex.length === 3 ? newHex.slice(1, 2).repeat(2) : newHex.slice(2, 4), 16);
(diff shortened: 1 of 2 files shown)


packages/fast-jss-utilities/src/color.ts
```diff
@@ -0,0 +1,13 @@
+export function convertHexToRgb(hex: string, alpha?: number): string {
```
nalogiudice: nit: Just a thought - why not just `hexToRgb` I don't think any understanding is lost and it is shorter.

nicholasrice: Part of me thinks that we shouldn't expose the alpha logic in this function. It doesn't really seem part of the scope of converting hex to rgb. 

Also, chroma already does color conversion - which we take a dependency on. Is there a reason we aren't just using chromas utilities?

chrisdholt: At some point I think there was a conversation that over time we want to reduce our dependency on Chroma where we can for smaller things like this. We can't tree shake it, so if we just wanted this functionality, we are still taking the whole thing.

If that's not a huge concern, I'll just abandon this. 

There is currently an issue on chroma for converting to ES6 to support this, but it's been open for over a year. We could always contribute it in

nicholasrice: Yea on one hand adding this now reduces the amount of work we'ed need to do later. On the other, it adds code in the interim and moving away from chroma isn't super high priority at the moment. 

I think i'm in favor of just using chroma for now and doing a migration if/when it makes sense, but I'm open to other's thoughts on the matter.

chrisdholt: I'm fine with this. I'll just close it out.

chrisdholt:
Closing as moving from Chroma is not a high pri and there is more work to do to fully address it.

Issue #271, opened 2018-04-17
Add theme support to FAST Apps footer


janechu:
Theme swapping on a global level should be supported.

There should be:
- A theme light and theme dark icon in the information bar
- The root level theme should swap when the theme light/dark buttons are toggled

PR #655, opened 2018-07-11
feat(fast-development-site): add component theme support

eljefe223:
Closes: #271 
 

### DIFF
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -40,6 +40,11 @@ export interface IAppState {
     theme: string;
 }
 
+const themes: any[] = [
+    {id: "light", displayName: "light", background: DesignSystemDefaults.backgroundColor},
+    {id: "dark", displayName: "dark", background: DesignSystemDefaults.foregroundColor}
+];
+
 export default class App extends React.Component<{}, IAppState> {
     constructor(props: {}) {
         super(props);
@@ -56,9 +61,7 @@ export default class App extends React.Component<{}, IAppState> {
                 formChildOptions={formChildOptions}
                 onUpdateDirection={this.handleUpdateDirection}
                 onUpdateTheme={this.handleUpdateTheme}
-                /* tslint:disable:max-line-length */
-                themes={[{name: "light", background: DesignSystemDefaults.backgroundColor}, {name: "dark", background: DesignSystemDefaults.foregroundColor}]}
-                /* tslint:enable:max-line-length */
+                themes={themes}
             >
                 <SiteMenu slot={"header"}>
                     <SiteMenuItem>
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -8,6 +8,7 @@ import Site, {
     formChildFromExamplesFactory,
     IFormChildOption,
     ISiteProps,
+    ITheme,
     SiteCategory,
     SiteCategoryIcon,
     SiteCategoryItem,
@@ -40,7 +41,7 @@ export interface IAppState {
     theme: string;
 }
 
-const themes: any[] = [
+const themes: ITheme[] = [
     {id: "light", displayName: "light", background: DesignSystemDefaults.backgroundColor},
     {id: "dark", displayName: "dark", background: DesignSystemDefaults.foregroundColor}
 ];
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -37,6 +37,7 @@ const hypertextStyles: ComponentStyles<IHypertextClassNameContract, undefined> =
 
 export interface IAppState {
     direction: Direction;
+    theme: string;
 }
 
 export default class App extends React.Component<{}, IAppState> {
@@ -44,7 +45,8 @@ export default class App extends React.Component<{}, IAppState> {
         super(props);
 
         this.state = {
-            direction: Direction.ltr
+            direction: Direction.ltr,
+            theme: "light"
         };
     }
 
@@ -53,6 +55,8 @@ export default class App extends React.Component<{}, IAppState> {
             <Site
                 formChildOptions={formChildOptions}
                 onUpdateDirection={this.handleUpdateDirection}
+                onUpdateTheme={this.handleUpdateTheme}
+                themes={["light", "dark"]}
             >
                 <SiteMenu slot={"header"}>
                     <SiteMenuItem>
@@ -73,10 +77,20 @@ export default class App extends React.Component<{}, IAppState> {
                     </SiteCategoryIcon>
                 </SiteCategory>
                 <SiteCategory slot={"category"} name={"Components"}>
-                    {componentFactory(examples, Object.assign({}, DesignSystemDefaults, {direction: this.state.direction}))}
+                    {componentFactory(examples, {...this.generateDesignSystem()})}
                 </SiteCategory>
             </Site>
         );
+    }
+
+    private generateDesignSystem(): IDesignSystem {
+        const designSystem: Partial<IDesignSystem> = {
+            direction: this.state.direction,
+            foregroundColor: this.state.theme === "dark" ? DesignSystemDefaults.backgroundColor : DesignSystemDefaults.foregroundColor,
+            backgroundColor: this.state.theme === "dark" ? DesignSystemDefaults.foregroundColor : DesignSystemDefaults.backgroundColor
+        };
+        
+        return Object.assign({}, DesignSystemDefaults, designSystem);
     }
 
     private handleUpdateDirection = (direction: Direction): void => {
@@ -87,5 +101,11 @@ export default class App extends React.Component<{}, IAppState> {
         this.setState({
             direction: newDir
         });
+    }
+    
+    private handleUpdateTheme = (theme: string): void => {
+        this.setState({
+            theme
+        })
     }
 }
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -56,7 +56,7 @@ export default class App extends React.Component<{}, IAppState> {
                 formChildOptions={formChildOptions}
                 onUpdateDirection={this.handleUpdateDirection}
                 onUpdateTheme={this.handleUpdateTheme}
-                themes={["light", "dark"]}
+                themes={[{name: "light", background: DesignSystemDefaults.backgroundColor}, {name: "dark", background: DesignSystemDefaults.foregroundColor}]}
             >
                 <SiteMenu slot={"header"}>
                     <SiteMenuItem>
--- a/packages/fast-components-react-msft/app/app.tsx
+++ b/packages/fast-components-react-msft/app/app.tsx
@@ -56,7 +56,9 @@ export default class App extends React.Component<{}, IAppState> {
                 formChildOptions={formChildOptions}
                 onUpdateDirection={this.handleUpdateDirection}
                 onUpdateTheme={this.handleUpdateTheme}
+                /* tslint:disable:max-line-length */
                 themes={[{name: "light", background: DesignSystemDefaults.backgroundColor}, {name: "dark", background: DesignSystemDefaults.foregroundColor}]}
+                /* tslint:enable:max-line-length */
             >
                 <SiteMenu slot={"header"}>
                     <SiteMenuItem>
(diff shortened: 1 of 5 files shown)


janechu:
Please add documentation for the new callback and the themes property.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -83,13 +85,19 @@ export interface ISiteState {
...
     locale: string;
+    theme: IThemeValues;
 }
 
 export enum SiteSlot {
     category = "category",
     categoryIcon = "category-icon"
 }
 
+export interface IThemeValues {
```
janechu: I don't think "Values" adds ... ironically... value here, I think it can just be `ITheme`

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -603,6 +612,7 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
                             key={index}
                             onClick={this.handleComponentClick}
                             index={index}
+                            background={this.state.theme.background}
```
janechu: This should error if theme is not provided, we should check to make sure theme is available, if not set this to null.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -755,6 +793,18 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         });
     }
 
+    private handleThemeUpdate = (e: any): void => {
+        if (this.props.onUpdateTheme) {
+            this.props.onUpdateTheme(e.target.value);
+        }
+
+        const found: IThemeValues = this.props.themes.find(function(item: IThemeValues): boolean { return item.name === e.target.value; });
```
janechu: for readability would prefer the contents of find are formatted more like so:
```
this.props.themes.find((item: ITheme): boolean {
    return item.name === e.target.value;
});
```

janechu: I would rename this, found is not very intuitive for what it is, `selectedTheme` might be better

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -83,13 +85,19 @@ export interface ISiteState {
...
+    theme: IThemeValues;
 }
 
 export enum SiteSlot {
     category = "category",
     categoryIcon = "category-icon"
 }
 
+export interface IThemeValues {
+    name: string;
```
janechu: seems like we are using name like a uuid, a better pattern might be:
```
{
    uuid: string;
    displayName: string;
    background: string;
}
```

I would prefer a separation of concerns here where the eventual value of the selected option is different from the string being displayed.

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -53,6 +55,10 @@ export default class App extends React.Component<{}, IAppState> {
             <Site
                 formChildOptions={formChildOptions}
                 onUpdateDirection={this.handleUpdateDirection}
+                onUpdateTheme={this.handleUpdateTheme}
+                /* tslint:disable:max-line-length */
+                themes={[{name: "light", background: DesignSystemDefaults.backgroundColor}, {name: "dark", background: DesignSystemDefaults.foregroundColor}]}
```
chrisdholt: can we clean this up by defining these as cost values outside the class? Then we don't ned to disable max line length

packages/fast-development-site-react/src/components/site/index.tsx
nalogiudice: +1

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -71,7 +72,7 @@ const styles: ComponentStyles<IComponentWrapperManagedClasses, IDevSiteDesignSys
 class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDesignSystem> & IManagedClasses<IComponentWrapperManagedClasses>, {}> {
     public render(): JSX.Element {
         return (
-            <div className={this.getClassNames()} onClick={this.handleClick} dir={this.props.dir || null}>
+            <div className={this.getClassNames()} onClick={this.handleClick} dir={this.props.dir || null} style={this.getThemeBackground()}>
```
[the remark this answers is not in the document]
chrisdholt: Yeah I think this would be a good thing to do.

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -97,6 +98,10 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
         );
     }
 
+    private getThemeBackground(): React.CSSProperties {
+        return this.props.transparentBackground ? null : {background: this.props.background};
+    }
```
chrisdholt: getThemeBackground is being called regardless of if there is style or not. It's being set regardless of if background is passed as a prop. Suggest a null check here before going forward.

```
if (!this.props.background) { return };
```

packages/fast-development-site-react/src/components/site/index.tsx
chrisdholt: Rather than doing this eventually, can we do this in this pr? Do we want uuid or just ID. If a user is passing, I think id works better from a clarity standpoint.

Do we want to hand them back the id in this case then rather than the display name? Seems like a better option.

packages/fast-development-site-react/src/components/site/index.tsx
chrisdholt: Yep, this is expecting background as required, but it's optional. No value here should just pass null.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -634,6 +644,7 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
                 <ComponentWrapper
                     key={index}
                     index={index}
+                    background={this.state.theme.background}
```
chrisdholt: Let's fix this instance of background as well.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -722,6 +734,22 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         );
     }
 
+    private renderThemeSelect(): JSX.Element {
+        if (Array.isArray(this.props.themes)) {
```
chrisdholt: I think we should confirm that this isn't an empty array.

```
if (Array.isArray(this.props.themes) && this.props.themes.length > 0) { }
```

packages/fast-development-site-react/src/components/site/index.tsx
janechu: I think we want to pass them the id as it should be unique.

packages/fast-development-site-react/README.md
```diff
@@ -64,6 +64,25 @@ The `Site` component should be used as the wrapping component for the developmen
...
+}
+```
+**themes** - An object that passes a theme `id`,`displayName` and `background`. `background` will be the color teh component renders on top of.
+
+#### Example on usage of `themes`
+```jsx
+const themes: any[] = [
+    {id: "Foo", displayName: "Foo", background: "#000"},
+    {id: "Bar", displayName: "Bar", background: "#FFF"},
+    {id: "Rumple", displayName: "Rumple", background: "#333"}];
```
janechu: Can we make the examples more real-world, so light and dark and can we also remove the `:any[]` since that is TypeScript and although you can use this the aim should be at straight up javascript

packages/fast-development-site-react/README.md
```diff
@@ -64,6 +64,25 @@ The `Site` component should be used as the wrapping component for the developmen
...
 
+**onUpdateTheme** - The callback provided for an update for theme, it well execute the given callback with the selected theme.
+
+#### Example callback to be passed to `onUpdateTheme`
+```jsx
+handleOnUpdateTheme = (theme: string): void {
+    this.setState({theme});
+}
+```
+**themes** - An object that passes a theme `id`,`displayName` and `background`. `background` will be the color teh component renders on top of.
```
janechu: "An object with properties..." would be more accurate, also misspelling of "the" and I would word it more like "The `background` property will be applied to the container of the component examples on the detail and example pages.

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -97,6 +98,12 @@ class ComponentWrapper extends React.Component<IComponentWrapperProps<IDevSiteDe
         );
     }
 
+    private getInlineStyle(): React.CSSProperties {
+        if (!this.props.background) { return; }
```
janechu: This should be a linting issue, the return should be on its own line

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -755,6 +801,20 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         });
     }
 
+    private handleThemeUpdate = (e: any): void => {
```
janechu: I think this should be typed as `Event.OnChange<HTMLSelectElement>`

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -634,6 +649,7 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
                 <ComponentWrapper
                     key={index}
                     index={index}
+                    background={this.state.theme.background || null}
```
janechu: If theme is undefined this will error because it will try to look up background on an undefined property. You can do `this.state.theme && this.state.theme.background ? this.state.theme.background : null`

I would also make the same change on line 620.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -745,7 +781,17 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         });
     }
 
-    private handleLocaleUpdate = (e: any): void => {
+    private renderThemes(): JSX.Element[] {
+        return this.props.themes.map((theme: ITheme, index: number): JSX.Element => {
+            return (
+                <option key={index}>
```
janechu: You can use the id here as they key and also as the value so this should be `<option key={theme.id} value={theme.id}>`

janechu: For reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -755,6 +801,20 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         });
     }
 
+    private handleThemeUpdate = (e: React.ChangeEvent<HTMLSelectElement>): void => {
+        if (this.props.onUpdateTheme) {
+            this.props.onUpdateTheme(e.target.value);
+        }
+
+        const selectedTheme: ITheme = this.props.themes.find((item: ITheme): boolean => {
+            return item.displayName === e.target.value;
```
janechu: After you update the previous comment you would want to change item.displayName to item.id

packages/fast-development-site-react/README.md
```diff
@@ -64,6 +64,25 @@ The `Site` component should be used as the wrapping component for the developmen
 
 **onUpdateDirection** - The callback provided for an update in RTL, it will execute the given callback with "rtl" or "ltr".
 
+**onUpdateTheme** - The callback provided for an update for theme, it well execute the given callback with the selected theme.
+
+#### Example callback to be passed to `onUpdateTheme`
+```jsx
+handleOnUpdateTheme = (theme: string): void {
```
janechu: you should remove `: void` since this is typescript

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -37,14 +37,21 @@ const hypertextStyles: ComponentStyles<IHypertextClassNameContract, undefined> =
 
 export interface IAppState {
     direction: Direction;
+    theme: string;
 }
 
+const themes: any[] = [
```
chrisdholt:
```
ITheme[]
```

Typings ^^^ instead of any[]

packages/fast-development-site-react/app/app.tsx
```diff
@@ -25,8 +25,15 @@ import { FrameworkEnum } from "../src/components/site/dev-tools";
 
 export interface IAppState {
     direction: Direction;
+    theme: string;
 }
 
+const themes: any[] = [
```
chrisdholt:
```
ITheme[]
```

Typings ^^^ instead of any[]

PR #687, opened 2018-07-13
fix(fast-components-react-msft): fix error running jest with components that take a dependency on chroma-js

### LINKED ISSUE #402: Adding snapshots in MSFT to components using Chroma plugin results in error
**Describe the bug**
When adding snapshots in react MSFT package, components using Chroma for JSS styling error in different ways (see screenshot)

**To Reproduce**
1. Go to `packages/fast-components-react-msft`
2. Add a snapshot for toggle (most straighforward error) or button (less helpful error) - anything that uses Chroma)
3. Update jest tests and see error

**Expected behavior**
Not to error and to add snapshots

**Screenshots**
screen shot 2018-05-10 at 12 03 21 pm

chrisdholt:
<body>
closes #402 
<footer>
 

### DIFF
--- a/packages/fast-colors/src/range.ts
+++ b/packages/fast-colors/src/range.ts
@@ -1,4 +1,4 @@
-import * as Chroma from "chroma-js";
+import Chroma from "chroma-js";
 
 /**
  * Interface that describes the configuration options of the color algorithm .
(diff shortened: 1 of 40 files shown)


packages/fast-colors/src/range.ts
```diff
@@ -1,4 +1,4 @@
-import * as Chroma from "chroma-js";
+import { default as Chroma } from "chroma-js";
```
nicholasrice: if this is the fix, can you just do `import Chroma from "chroma-js"`?

chrisdholt: Confirmed and updated for all packages.

PR #689, opened 2018-07-13
fix(toggle): fix labels on toggle to support dark theme

### LINKED ISSUE #688: Toggle label does not swap when design system values are changed
The toggle label does not swap when design system values are changed. The rest of the component themes as expected but the label does not. 

Current behavior: Toggle label does not swap when foreground color and background color are swapped.

Expected behavior: Toggle label color updates when design system values are different.

eljefe223:
Closes #688 
 


chrisdholt:
@eljefe223 Can you update the PR message to be more clear as to what the fix actually is? It looks like toggle already supports the dark theme, but this addresses a different issue than simply adding the theme.

Issue #690, opened 2018-07-13
Hypertext needs dark theme support


eljefe223:
Hypertext needs dark theme support

chrisdholt:
@eljefe223 can you update this issue to give more detail to the actual bug?

PR #691, opened 2018-07-13
fix(hypertext): update foreground to use design system colors for theming support

eljefe223:
Closes: #690 
 
### DIFF
--- a/packages/fast-components-styles-msft/src/hypertext/index.ts
+++ b/packages/fast-components-styles-msft/src/hypertext/index.ts
@@ -16,6 +16,9 @@ const styles: ComponentStyles<IHypertextClassNameContract, IDesignSystem> = {
     hypertext: {
         outline: toPx(0),
         textDecoration: "none",
+        color: (config: IDesignSystem): string => {
+            return get(config, "foregroundColor") || designSystemDefaults.foregroundColor;
+        },
         "&[href]": {
             ...applyHypertextBorder(1),
             color: (config: IDesignSystem): string => {


chrisdholt:
Can you update the PR merge message to be more specific regarding the fix. If I'm looking at this, theming is support on a hypertext with an href. This seems to broad and not a fix as written.

awentzel:
Instead of repeating “fix” in the title you should specify what update the was made for the user.

chrisdholt:
@nicholasrice In this instance, I think that we want hypertext without an href to inherit from it's parent rather than specifying a design system color. It seems like this is an issue that needs to be reconciled within the tool rather than modifying the component.

nicholasrice:
@chrisdholt good call - yes we do.

I agree, this is a tooling bug - the region that has it's theme changing should get assigned a the new foreground color. We may run into interesting issues here - color is applied to the html/body elements and will have a lower specificity than the element we apply the new color too.

awentzel:
@nicholasrice as a tooling bug what is the impact on this PR and current status.  It's now marked as stale by the system.  How do we move forward on this?

chrisdholt:
@awentzel The primary question we are asking in this case is if components that would traditionally inherit styles should instead include them. It's an interesting use case because I think in the case of color (such as this), we may want to do so, but we wouldn't want to do so in the case of something like font-family. Last week @nicholasrice and I decided to think on it for a bit and come back to it. I think we just need to land/confirm our position going forward of all components including colors instead of inheriting them.

Issue #692, opened 2018-07-13
Add visual toggle state to Waffle toggle in views bar. 


nalogiudice:
Transparency toggle needs an active visual state when toggling.

PR #693, opened 2018-07-13
fix(fast-development-site-react): add active styles for transparency toggle

nalogiudice:
Closes [#692](https://github.com/Microsoft/fast-dna/issues/692)

Active state style:
screen shot 2018-07-13 at 12 26 38 pm

Hover brings a slight shift in opacity, like our other form elements.

<body>

<footer>
 

### DIFF
--- a/packages/fast-development-site-react/src/components/site/index.tsx
+++ b/packages/fast-development-site-react/src/components/site/index.tsx
@@ -236,11 +236,22 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
     },
     site_transparencyToggleButton: {
         border: "none",
+        borderRadius: toPx(2),
         background: "none",
         height: toPx(32),
         width: toPx(32),
         cursor: "pointer",
-        outline: "0"
+        outline: "0",
+        marginRight: toPx(4),
+        opacity: ".85",
+        paddingTop: toPx(4),
+        "&:hover": {
+            opacity: "1"
+        },
+        "&[aria-pressed=\"true\"]": {
+            background: "#FFFFFF",
+            opacity: "1"
+        }
     },
     site_transparencyToggleButtonIcon: {
         position: "relative",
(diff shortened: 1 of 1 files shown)


packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -236,11 +236,22 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
...
-        outline: "0"
+        outline: "0",
+        marginRight: toPx(4),
+        opacity: ".85",
+        paddingTop: toPx(4),
+        "&:hover": {
+            opacity: "1"
+        },
+        "&[aria-pressed=\"true\"]": {
+            background: "#FFFFFF",
```
chrisdholt: Is there a reason that we are not using the design system here?

chrisdholt: Is there a reason that we are not using the design system here?

nalogiudice: Because it didn't work. You can see that the design system is used randomly in the file. Not sure the logic behind that, but neither: 
`background: (config: IDevSiteDesignSystem): string => config.backgroundColor,` or 
```
background: (config: IDevSiteDesignSystem): string => {
                return config.backgroundColor;
            }
```
render anything.

chrisdholt: Pulling this down and taking a look. It seems like something else is going on if this isn't working. If we can't figure it out, we can propose putting this in with hard coded values and file an issue, but I'd like us to first try and see what might be going on to cause this issue as it may not be a bug.

chrisdholt: @nalogiudice config.backgroundColor is coming in as undefined which is why it's not working. Can you quickly see if the other parts of this using the design system are also not showing up correctly? If they are working correctly it would be great to see if we are missing something as to why it's working sometimes in this file. If they are both not working, I think we should file a bug and note the bug in this file at the styles object. Then we can push this through

nalogiudice: @chrisdholt confirming that they all return `undefined`. Will add a bug

nalogiudice: [#694](https://github.com/Microsoft/fast-dna/issues/694)

awentzel:
why is adding active styles a fix?  Should it be a feat?  Or it would would be update to fix something?

nalogiudice:
I've updated the PR description - not sure how to make it more verbose that than. Also have updated to feat.

chrisdholt:
Please rebase :)

Issue #702, opened 2018-07-16
MSFT Button should use an enum instead of booleans to describe button subtype


eljefe223:
MSFT Button example should use a dropdown for button type

PR #703, opened 2018-07-16
fix(button): update to use enum for button types

eljefe223:
Closes: #702 
 
BREAKING CHANGE: The API for the fast-components-react-msft button has 4 boolean props (justified, lightweight, outline and primary). These values are mutually exclusive and cannot be applied simultaneously. For this reason an enum is a more appropriate approach. This fix removes those booleans and passes a single prop.

### DIFF
--- a/packages/fast-components-react-msft/README.md
+++ b/packages/fast-components-react-msft/README.md
@@ -18,7 +18,7 @@ document.body.appendChild(root);
 
 function render(): void {
     ReactDOM.render(
-        <Button subtype="primary">
+        <Button appearance="primary">
             Click me!
         </Button>,
         root
--- a/packages/fast-components-react-msft/README.md
+++ b/packages/fast-components-react-msft/README.md
@@ -18,7 +18,7 @@ document.body.appendChild(root);
 
 function render(): void {
     ReactDOM.render(
-        <Button primary={true}>
+        <Button subtype="primary">
             Click me!
         </Button>,
         root
--- a/packages/fast-components-react-msft/src/button/button.props.ts
+++ b/packages/fast-components-react-msft/src/button/button.props.ts
@@ -2,7 +2,7 @@ import * as React from "react";
 import { IButtonHandledProps as IBaseButtonHandledProps } from "@microsoft/fast-components-react-base";
 import { IManagedClasses, IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 
-export enum Appearance {
+export enum ButtonAppearance {
     justified= "justified",
     lightweight= "lightweight",
     outline= "outline",
@@ -13,7 +13,7 @@ export interface IButtonHandledProps extends IBaseButtonHandledProps {
     /**
      * The Button subType
      */
-    appearance?: Appearance;
+    appearance?: ButtonAppearance;
 }
 
 export interface IButtonUnhandledProps extends React.AllHTMLAttributes<HTMLElement> {}
--- a/packages/fast-components-react-msft/src/button/button.props.ts
+++ b/packages/fast-components-react-msft/src/button/button.props.ts
@@ -2,7 +2,7 @@ import * as React from "react";
 import { IButtonHandledProps as IBaseButtonHandledProps } from "@microsoft/fast-components-react-base";
 import { IManagedClasses, IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 
-export enum Subtype {
+export enum Appearance {
     justified= "justified",
     lightweight= "lightweight",
     outline= "outline",
@@ -13,7 +13,7 @@ export interface IButtonHandledProps extends IBaseButtonHandledProps {
     /**
      * The Button subType
      */
-    subtype?: Subtype;
+    appearance?: Appearance;
 }
 
 export interface IButtonUnhandledProps extends React.AllHTMLAttributes<HTMLElement> {}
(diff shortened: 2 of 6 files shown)


awentzel:
Please update PR / Issue title and description to conform to [guidance](https://github.com/Microsoft/fast-dna/wiki/pull-request-guidance) for better changelog reporting.

packages/fast-components-react-msft/src/button/button.props.ts
```diff
@@ -2,26 +2,18 @@ import * as React from "react";
...
-    /**
-     * The lightweight option
-     */
-    lightweight?: boolean;
-
-    /**
-     * The outline option
-     */
-    outline?: boolean;
+export enum Subtype {
```
chrisdholt: Subtype seems odd to me for button, because I expect that it would be a subset of the type attribute. This is really just specific to style. Can we revisit the naming on this?

eljefe223: Yes we should @janechu and I noodled over it for awhile. We cant use style or type. I thought about ButtonStyle but Jane thought that was redundant... we could also do something like "appearance"

nalogiudice: +1 `appearance` could work

janechu: I agree, I like appearance

janechu:
Can the footer message for the breaking change specify what exactly the breaking change is and why it was done?

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -42,21 +39,21 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
     protected generateClassNames(): string {
         let classLocation: string;
 
-        if (this.props.primary) {
+        if (this.props.subtype === Subtype.primary) {
```
janechu: I would prefer using a switch here now that this is isolated to one item we are checking against, we can also skip storing a variable, it can look like:

```
switch (this.props.appearance) {
    case Subtyle.outline:
        return super.generateClassNames(get(this.props, "managedClasses.button_outline"));
    // etc..
    default:
        return null;
}
```

packages/fast-components-react-msft/src/button/button.props.ts
chrisdholt: + 1 - great suggestion.

chrisdholt:
I think we still need to walk through a proposal for this being a breaking change and land when and why we choose to break vs. deprecate props.

packages/fast-components-react-msft/src/button/button.props.ts
```diff
@@ -2,26 +2,18 @@ import * as React from "react";
...
-    /**
-     * The lightweight option
-     */
-    lightweight?: boolean;
-
-    /**
-     * The outline option
-     */
-    outline?: boolean;
+export enum Appearance {
```
chrisdholt: I think we should namespace this specific to button to ensure we don't create a global that we may need to leverage later. Similar components with similar enum's would follow the same pattern.

Thoughts?

chrisdholt: The prop name is fine because it is scoped, but here I think we'll want to specify that this is `ButtonAppearance`

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -40,23 +37,22 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
...
+            case ButtonAppearance.primary:
+                return super.generateClassNames(get(this.props, "managedClasses.button_primary"));
+            case ButtonAppearance.outline:
+                return super.generateClassNames(get(this.props, "managedClasses.button_outline"));
+            case ButtonAppearance.lightweight:
+                return super.generateClassNames(get(this.props, "managedClasses.button_lightweight"));
+            case ButtonAppearance.justified:
+                return super.generateClassNames(get(this.props, "managedClasses.button_justified"));
+            default:
+                return null;
```
chrisdholt: Can we remove some of the duplicated code here and only return if we have classNames? Perhaps:

```
    let className: string;

    if (!this.props.appearance) {
        return;
    }

    switch (this.props.appearance) {
        case ButtonAppearance.primary:
            className = "managedClasses.button_primary";
            break;
        case ButtonAppearance.outline:
            className = "managedClasses.button_outline";
            break;
        case ButtonAppearance.lightweight:
            className = "managedClasses.button_lightweight";
            break;
        case ButtonAppearance.justified:
            className = "managedClasses.button_justified";
            break;
    }

    return super.generateClassNames(get(this.props, className));
}
```

@janechu, thoughts?

janechu: We can avoid the assigning of a variable but I agree we need the if statement.

packages/fast-components-react-msft/README.md
```diff
@@ -18,7 +18,7 @@ document.body.appendChild(root);
 
 function render(): void {
     ReactDOM.render(
-        <Button primary={true}>
+        <Button subtype="primary">
```
janechu: this needs to be updated to appearance

chrisdholt:
A PR has been added to use enzyme for unit testing these components. Let's add unit tests for this; that will fix the broken build and test coverage issue.

packages/fast-components-react-msft/src/button/button.props.ts
```diff
@@ -2,26 +2,18 @@ import * as React from "react";
...
+    justified= "justified",
+    lightweight= "lightweight",
+    outline= "outline",
+    primary= "primary",
+}
 
+export interface IButtonHandledProps extends IBaseButtonHandledProps {
     /**
-     * The primary option
+     * The Button subType
```
chrisdholt: comment should be "The button appearance"

Issue #704, opened 2018-07-17
Click outside to close option menu in fast-form-generator


nalogiudice:
When clicking outside the "options" menu in the configure UI, you can't click out to close it. We should eat a click and close.

PR #705, opened 2018-07-17
feat (fast-form-generator): adds logic to close option menus when clicking outside of menu UI

nalogiudice:
Closes [#704](https://github.com/Microsoft/fast-dna/issues/704)
<body>

<footer>
 

### DIFF
--- a/packages/fast-form-generator-react/src/form/form-item.array.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.array.tsx
@@ -75,10 +75,23 @@ export interface IFormItemArrayState {
  */
 class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasses<IFormItemArrayClassNameContract>, IFormItemArrayState> {
 
+    /**
+     * Store a reference to the menu element
+     */
+    private menuElement: HTMLUListElement;
+
+    /**
+     * Store a reference to the button element
+     */
+    private buttonElement: HTMLButtonElement;
+
+    /**
+     * Store a reference to the span element
+     */
+    private spanElement: HTMLSpanElement;
+
     constructor(props: IFormItemArrayProps & IManagedClasses<IFormItemArrayClassNameContract>) {
         super(props);
-
-        window.addEventListener("click", this.handleWindowClick)
 
         this.state = {
             hideOptionMenu: true
@@ -87,12 +100,19 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
 
     public render(): JSX.Element {
         return (
-            <div className={this.props.managedClasses.formItemArray} ref={this.setRef("rootElement")}>
+            // tslint:disable-next-line:jsx-no-string-ref
+            <div className={this.props.managedClasses.formItemArray}>
                 <div className={this.props.managedClasses.formItemArray_header}>
                     <h3>{this.getLabelText()}</h3>
                     {/* TODO: #460 Fix "identical-code" */}
-                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
-                    <ul aria-hidden={this.state.hideOptionMenu} className={this.props.managedClasses.formItemArray_actionMenu}>
+                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu} ref={this.storeButtonRef}>
+                        <span ref={this.storeSpanRef}>Options</span>
+                    </button>
+                    <ul
+                        ref={this.storeMenuRef}
+                        aria-hidden={this.state.hideOptionMenu}
+                        className={this.props.managedClasses.formItemArray_actionMenu}
+                    >
                         {this.renderArrayMenuItems()}
                     </ul>
                 </div>
@@ -101,8 +121,36 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
         );
     }
 
+    public componentDidUpdate(): void {
+        this.state.hideOptionMenu
+            ? window.removeEventListener("click", this.handleWindowClick)
+            : window.addEventListener("click", this.handleWindowClick);
+    }
+
+    private storeMenuRef = (ref?: HTMLUListElement): void => {
+        if (ref) {
+            this.menuElement = ref;
+        }
+    }
+
+    private storeButtonRef = (ref?: HTMLButtonElement): void => {
+        if (ref) {
+            this.buttonElement = ref;
+        }
+    }
+
+    private storeSpanRef = (ref?: HTMLSpanElement): void => {
+        if (ref) {
+            this.spanElement = ref;
+        }
+    }
+
     private toggleMenu = (): void => {
         this.setState({hideOptionMenu: !this.state.hideOptionMenu});
+    }
+
+    private closeMenu = (): void => {
+        this.setState({hideOptionMenu: true});
     }
 
     /**
@@ -219,10 +267,10 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
 
     /**
      * Handle clicking on the window - closes the Select if the users
-     * is clicking somewhere other than the Select.
+     * is clicking somewhere other than the menu.
      */
-    private handleWindowClick = (e: React.MouseEvent): void => {
-        if (!isDescendant(this.getRef('rootElement'), e.target)) {
+    private handleWindowClick = (e: any): void => {
+        if (!isDescendant(this.menuElement , e.target) && e.target !== this.buttonElement && e.target !== this.spanElement) {
             this.closeMenu();
         }
     }
--- a/packages/fast-form-generator-react/src/form/form-item.children.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.children.tsx
@@ -12,6 +12,7 @@ import styles from "./form-item.children.style";
 import { IFormItemChildrenClassNameContract } from "../class-name-contracts/";
 import manageJss, { IJSSManagerProps } from "@microsoft/fast-jss-manager-react";
 import { IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+import { isDescendant } from "@microsoft/fast-web-utilities";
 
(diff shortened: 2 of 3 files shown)


packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -12,6 +12,7 @@ import styles from "./form-item.array.style";
 import { IFormItemArrayClassNameContract } from "../class-name-contracts/";
 import manageJss, { IJSSManagerProps } from "@microsoft/fast-jss-manager-react";
 import { IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+import { isDescendant } from "@microsoft/fast-web-utilities";
```
chrisdholt: Does this actually work? isDescendant was pulled out of fast-web-utilities in favor of just using element.contains()

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -98,10 +121,38 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
         );
     }
 
+    public componentDidUpdate(): void {
+        this.state.hideOptionMenu
+            ? window.removeEventListener("click", this.handleWindowClick)
+            : window.addEventListener("click", this.handleWindowClick);
```
chrisdholt: Every update will perform this, even if it's not needed. Do we need to do this on update every time the cycle is ran? Is there a reason we should not simply add the event listener on mount and remove on unmount?

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -12,6 +12,7 @@ import styles from "./form-item.children.style";
 import { IFormItemChildrenClassNameContract } from "../class-name-contracts/";
 import manageJss, { IJSSManagerProps } from "@microsoft/fast-jss-manager-react";
 import { IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+import { isDescendant } from "@microsoft/fast-web-utilities";
```
chrisdholt: Same comment here on isDescendant. I'm surprised you have access to this since I don't see it in the actual package.

packages/fast-web-utilities/src/dom.ts
```diff
@@ -5,3 +5,24 @@
...
+    while (!!node) {
+        if (node === parent) {
+            return true;
+        }
+
+        node = node.parentNode;
+    }
+
+    return false;
+}
```
chrisdholt: I see - This was intentionally removed for element.contains. We should be able to accomplish this without this function.

packages/fast-form-generator-react/src/form/form-item.array.tsx
janechu: window click event is IE9+, we may want to use document instead of window because of that. It doesn't hurt to be backwards compatible in this case so I would change it.

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -98,10 +121,38 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
...
+    }
+
+    private storeMenuRef = (ref?: HTMLUListElement): void => {
+        if (ref) {
+            this.menuElement = ref;
+        }
+    }
+
+    private storeButtonRef = (ref?: HTMLButtonElement): void => {
```
janechu: similarly to the renaming I suggested above, I would have this be called setOptionMenuTriggerRef, store isn't really standard.

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -444,6 +498,16 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
         });
     }
 
+    /**
+     * Handle clicking on the window - closes the Select if the users
+     * is clicking somewhere other than the menu.
+     */
+    private handleWindowClick = (e: any): void => {
```
janechu: same comment as before, can be more precise typing than "any"

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -405,10 +437,32 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
         }
     }
 
+    private storeMenuRef = (ref?: HTMLUListElement): void => {
+        if (ref) {
+            this.menuElement = ref;
+        }
+    }
+
+    private storeButtonRef = (ref?: HTMLButtonElement): void => {
```
janechu: same comments as before, this can be more semantic, see my previous comments

packages/fast-web-utilities/src/dom.spec.ts
```diff
@@ -36,3 +36,9 @@ describe("getKeyCode", () => {
         expect(getKeyCode(new KeyboardEvent("keypress", {charCode: 9} as KeyboardEventInit))).toBe(KeyCodes.tab);
     });
 });
+
+describe("isDescendant", () => {
+    test("should correctly handle null", () => {
```
janechu: This does not seem like adequate testing. Please add in sufficient tests for:
- testing an immediate child
- testing for a child further down the tree of a component
- testing for a adjacent element to the one being checked (false)
- testing for a parent being the descendent of its child (false)

janechu: redundant, see @chrisdholt comments

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -84,12 +100,19 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
...
         return (
+            // tslint:disable-next-line:jsx-no-string-ref
             <div className={this.props.managedClasses.formItemArray}>
                 <div className={this.props.managedClasses.formItemArray_header}>
                     <h3>{this.getLabelText()}</h3>
                     {/* TODO: #460 Fix "identical-code" */}
-                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu}><span>Options</span></button>
-                    <ul aria-hidden={this.state.hideOptionMenu} className={this.props.managedClasses.formItemArray_actionMenu}>
+                    <button onClick={this.toggleMenu} aria-expanded={!this.state.hideOptionMenu} ref={this.storeButtonRef}>
+                        <span ref={this.storeSpanRef}>Options</span>
```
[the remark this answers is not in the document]
chrisdholt: element.contains() :)

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -74,6 +90,12 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
         };
     }
 
+    public componentDidUpdate(): void {
+        this.state.hideOptionMenu
+            ? window.removeEventListener("click", this.handleWindowClick)
```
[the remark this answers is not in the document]
chrisdholt: I'd propose we just add it onMount and remove it on unMount.

packages/fast-form-generator-react/src/form/form-item.array.tsx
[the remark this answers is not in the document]
nalogiudice: isDescendant did not work. I can try element.contains()

packages/fast-form-generator-react/src/form/form-item.children.tsx
nalogiudice: I added it to the package

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -214,6 +265,16 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
         this.props.onChange(this.props.dataLocation, arrayMove(this.props.data, oldIndex, newIndex));
     }
 
+    /**
+     * Handle clicking on the window - closes the Select if the users
+     * is clicking somewhere other than the menu.
+     */
+    private handleWindowClick = (e: any): void => {
```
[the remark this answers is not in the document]
nalogiudice: It didn't like any typings...kept on expecting 'slots' on element... there are plenty of any typings in this repo. I can look into it more though

packages/fast-form-generator-react/src/form/form-item.children.tsx
nalogiudice: Yeah I originally tried this but it didn't like it. Will look into more.

packages/fast-form-generator-react/src/form/form-item.array.tsx
```diff
@@ -98,10 +121,38 @@ class FormItemArray extends React.Component<IFormItemArrayProps & IManagedClasse
         );
     }
 
+    public componentDidUpdate(): void {
+        this.state.hideOptionMenu
+            ? window.removeEventListener("click", this.handleWindowClick)
+            : window.addEventListener("click", this.handleWindowClick);
+    }
+
+    private storeMenuRef = (ref?: HTMLUListElement): void => {
```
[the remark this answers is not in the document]
nalogiudice: So a couple of comments. They way these references are stored is how they are done in other places in this package, so those should be updated if they are "wrong". I used this page as a reference but was only able to get it working the way I have here. I will look into it with the other changes though.

packages/fast-web-utilities/src/dom.spec.ts
nalogiudice: Well this was the testing we provided for it in our fluent repo. May be moot now anyway.

packages/fast-form-generator-react/src/form/form-item.array.tsx
nalogiudice: Sure.

packages/fast-form-generator-react/src/form/form-item.array.tsx
nalogiudice: Well, yeah it works or I wouldnt have submitted :)
We'll see what we can do with contains.

PR #707, opened 2018-07-17
fix(fast-css-editor-react): update the api to accept strings for position

### LINKED ISSUE #706: Position should allow any valid CSS values
Currently the user can only put numbers, this does not accurately match what can be applied to these CSS properties.

This will cause a breaking change.

janechu:
Resolves #706

BREAKING CHANGE: The API for the fast-css-editor-react libraries CSSPosition and CSSEditor exports is changing to use strings instead of numbers for the left, right, top and bottom CSS properties. This is a fix needed to map directly to JSS and other CSS systems.

### DIFF
--- a/packages/fast-css-editor-react/app/app.tsx
+++ b/packages/fast-css-editor-react/app/app.tsx
@@ -37,16 +37,16 @@ class App extends React.Component<{}, IAppState> {
             },
             editorExample: {
                 position: PositionValue.absolute,
-                left: 4,
-                top: 12
+                left: "4px",
+                top: "12px"
             },
             positionDetail:{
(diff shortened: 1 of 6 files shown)


packages/fast-css-editor-react/app/app.tsx
```diff
@@ -37,16 +37,16 @@ class App extends React.Component<{}, IAppState> {
             },
             editorExample: {
                 position: PositionValue.absolute,
-                left: 4,
-                top: 12
+                left: "4px",
```
nalogiudice: Just curious, what happens if a user passes just "4" as a string?

Issue #711, opened 2018-07-18
Update style of language and theme menu and add resize behavior


eljefe223:
Update style of language and theme menu. Select menu for locales should say 'English'

chrisdholt:
@eljefe223 What is the "definition of done" for this issue?

PR #712, opened 2018-07-18
feat(fast-development-site): update theme and locale styles for select and resize behavior

eljefe223:
Closes: #711 
 

### DIFF
--- a/packages/fast-development-site-react/package.json
+++ b/packages/fast-development-site-react/package.json
@@ -31,6 +31,7 @@
     "@microsoft/fast-jss-manager-react": "^1.9.2",
     "@microsoft/fast-jss-utilities": "^1.9.2",
     "@microsoft/fast-layouts-react": "^1.9.2",
+    "@microsoft/fast-web-utilities": "^1.9.2",
     "react-syntax-highlighter": "^7.0.4"
   },
   "devDependencies": {
--- a/packages/fast-development-site-react/src/components/site/index.tsx
+++ b/packages/fast-development-site-react/src/components/site/index.tsx
@@ -34,7 +34,7 @@ import {
     Row,
     RowResizeDirection
 } from "@microsoft/fast-layouts-react";
-import { Direction, isRTL } from "@microsoft/fast-application-utilities";
+import { Direction } from "@microsoft/fast-application-utilities";
 
 export enum ComponentViewSlot {
     example = "canvas-example-view",
@@ -46,7 +46,7 @@ export interface ISiteProps {
     formChildOptions: IFormChildOption[];
     onUpdateDirection?: (ltr: Direction) => void;
     onUpdateTheme?: (theme: string) => void;
-    locales?: string[];
+    locales?: ILocaleMapping;
     themes?: ITheme[];
     frameworks?: FrameworkEnum | FrameworkEnum[];
     activeFramework?: FrameworkEnum;
@@ -85,7 +85,7 @@ export interface ISiteState {
     componentBackgroundTransparent: boolean;
     formView: boolean;
     devToolsView: boolean;
-    locale: string;
+    locale: ILocale;
     theme: ITheme;
 }
 
@@ -300,10 +300,29 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
     }
 };
 
+export interface ILocale {
+    displayName: string;
+    direction: Direction;
+}
+
+export interface ILocaleMapping {
+    en: ILocale;
+    [locale: string]: ILocale;
+};
+
 class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClasses>, ISiteState> {
 
     public static defaultProps: Partial<ISiteProps> = {
-        locales: ["English", "en-rtl"]
+        locales: {
+            en: {
+                displayName: "English",
+                direction: Direction.ltr,
+            },
+            "en-rtl": {
+                displayName: "English-RTL",
+                direction: Direction.rtl
+            }
+        }
     };
 
     private localeSelect: React.RefObject<HTMLSelectElement>;
@@ -337,7 +356,7 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
             detailViewComponentData: this.getDetailViewComponentData(),
             formView: true,
             devToolsView: false,
-            locale: "English",
+            locale: this.props.locales.en,
             theme: this.getInitialTheme()
         };
     }
@@ -366,8 +385,8 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
     }
 
     public componentDidMount(): void {
-        localeMeasure = this.visualLength(this.localeRuler, this.localeSelect);
-        themeMeasure = this.visualLength(this.themeRuler, this.themeSelect);
+        localeMeasure = this.calculateSelectWidth(this.localeRuler, this.localeSelect);
+        themeMeasure = this.calculateSelectWidth(this.themeRuler, this.themeSelect);
 
         // If the path we load the site in doesn't match component view, update state
         // to match the path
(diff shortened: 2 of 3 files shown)


packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -780,6 +807,24 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
...
+    private renderMeasureSpan(id: string, displayString: string): JSX.Element {
+        return (
+            <span id={id} aria-hidden="true" className={this.props.managedClasses.site_measureSpan}>
+                {displayString}
+            </span>
+        );
+    }
+
+    private visualLength(id: string, ref: React.RefObject<HTMLSelectElement>): number {
+        const ruler: HTMLElement = document.getElementById(id);
```
nicholasrice: I'd recommend working inside of React's ecosystem and using refs for the rulers also - avoids any possible complication with update cycles and missing elements.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -22,6 +22,7 @@ import ActionBar from "./action-bar";
 import DevTools, { FrameworkEnum } from "./dev-tools";
 import ConfigurationPanel from "./configuration-panel";
 import NotFound from "./not-found";
+import { convertStylePropertyPixelsToNumber } from "@microsoft/fast-web-utilities";
```
janechu: this doesn't appear to get used

packages/fast-development-site-react/package.json
```diff
@@ -31,6 +31,7 @@
     "@microsoft/fast-jss-manager-react": "^1.9.2",
     "@microsoft/fast-jss-utilities": "^1.9.2",
     "@microsoft/fast-layouts-react": "^1.9.2",
+    "@microsoft/fast-web-utilities": "^1.9.2",
```
janechu: does not appear to get used

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -314,7 +337,7 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
             detailViewComponentData: this.getDetailViewComponentData(),
             formView: true,
             devToolsView: false,
-            locale: "en",
+            locale: "English",
```
janechu: this is not a valid locale, valid locales must start with two lowercase letters indicating the language, which can then be followed by various other parameters. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Language for how this is used in a header and https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes for a list of the accepted language values

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -739,9 +768,13 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
                 >
                     <span dangerouslySetInnerHTML={{__html: glyphTransparency}}/>
                 </button>
+                {this.renderMeasureSpan(this.themeRuler, `${this.state.theme.displayName}`)}
                 {this.renderThemeSelect()}
+                {this.renderMeasureSpan(this.localeRuler, `${this.state.locale}`)}
```
janechu: backticks unnecessary since you are not concatenating anything within, see also 771

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -782,6 +817,25 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         }
     }
 
+    private renderMeasureSpan(ref: React.RefObject<HTMLSpanElement>, displayString: string): JSX.Element {
```
janechu: please be more verbose on method name so on first glance a reader can understand what it is for. It can be quite long as this all gets minified for production.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -782,6 +817,25 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
...
+        return (
+            <span ref={ref} aria-hidden="true" className={this.props.managedClasses.site_measureSpan}>
+                {displayString}
+            </span>
+        );
+    }
+
+    private visualLength(refRuler: React.RefObject<HTMLSpanElement>, ref: React.RefObject<HTMLSelectElement>): number {
```
janechu: also here, prefer a longer method name that describes what this is doing

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: @janechu This is being used on line 832 and 833

packages/fast-development-site-react/package.json
eljefe223: See response to other comment

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -289,17 +300,48 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
...
+        locales: {
+            en: {
+                displayName: "English",
+                direction: Direction.ltr,
+            },
+            "en-rtl": {
+                displayName: "English-RTL",
+                direction: Direction.rtl
+            }
+        }
```
chrisdholt: I believe that the locale changes here will be a breaking change for any users who currently use the dev site.

Is there a reason we are including these changes in this PR instead of in a different one?

eljefe223: @chrisdholt The issues Jason put in the One Note was the styling of the select to match the design and 'en' should be 'English'

chrisdholt: Sure - this is a breaking change though. Have you booted this up in the MSFT package to ensure that it works there without changing that package configuration? I think it's going to break.

chrisdholt: I see, it's not breaking currently in MSFT because we are relying on the default props. However, we've changed from a "string" to an interface with an additional key for the props. It isn't breaking our MSFT scenario but I believe this is a breaking change. We can still move forward with it, but I'm more curious why we made these additional changes. It seems to me that we could technically just change "en" to "English" in the MSFT implementation to solve for this.

chrisdholt: Primarily, just offering some insight to consider. I think in this specific case though we ended up changing the entire fundamentals of how the thing works and the stated purpose of that change doesn't map to the issue or PR title. So, looking at this, it seems like the larger refactor which would've been better suited to it's own PR. The breaking change itself doesn't make sense in this context to me, and while I think it's the correct code solution, this will get lost in the changelog as part of this PR/title.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -739,9 +768,13 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
                 >
                     <span dangerouslySetInnerHTML={{__html: glyphTransparency}}/>
                 </button>
+                {this.renderMeasureSpan(this.themeRuler, this.state.theme.displayName)}
```
janechu: this will error, theme can be undefined

packages/fast-development-site-react/src/components/site/index.tsx
janechu: I suggest since this is measuring the display names we call it renderSelectTextStringsForMeasurement and add an additional comment above "Renders a span to measure the display names to make sure the width of the select is consistent"

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -816,6 +870,7 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
     }
 
     private handleLocaleUpdate = (e: React.ChangeEvent<HTMLSelectElement>): void => {
+
```
janechu: nit: unnecessary empty line

packages/fast-development-site-react/src/components/site/index.tsx
awentzel: I'd recommend reversing the naming pattern.  In API's they are easier to understand and learn when things can group together by name.  So if you started with "Measure..." anything related to measuring would show up together in an alphabetized list of function names. 

Either way, always document with function comments, name, description, inputs, outputs.  this will then ensure the name of the function, whatever it is, has a good description for verbosity and examples are nice as well.

Issue #714, opened 2018-07-19
Update cursor for non interactable areas in the UI


eljefe223:
Update cursor for non interactable areas in the UI. 

Update areas of UI that are not interactable that currently show the text insert cursor to be the white arrow cursor.

The areas are "FAST Documentation title".

"Building blocks" heading.

"Component: …" and "Beta" text.

![image](https://user-images.githubusercontent.com/37851214/42953538-85e3c5ae-8b2f-11e8-9706-b31432edfa95.png)

chrisdholt:
Can you update the title and description of this to mention that this is specific to the fast-development-site-react package?

PR #715, opened 2018-07-19
fix(fast-development-site): text cursor for non interactable areas should be default cursor

eljefe223:
Closes: #714 
 

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -204,14 +205,15 @@ const styles: ComponentStyles<ISiteManagedClasses, IDevSiteDesignSystem> = {
     site_paneTocRow: {
         display: "flex",
         flexWrap: "nowrap",
-        alignItems: "center",
```
chrisdholt: we should technically be doing seperate PR's for this or opening an issue and just assigning it as a good first issue for other contributors.

PR #720, opened 2018-07-19
fix(fast-form-generator-react): update the controlled state to respect arrays and objects

### LINKED ISSUE #719: Controlled state of form generator only affects the object location
When the form generator has been implemented in its controlled state the user should be able to navigate to sections as if it was in its uncontrolled state.

Controlled state (passing in the location prop) to the Form should allow the user to activate object prioperties and array items.

janechu:
Resolves #719

### DIFF
--- a/packages/fast-form-generator-react/app/index.tsx
+++ b/packages/fast-form-generator-react/app/index.tsx
@@ -16,13 +16,20 @@ import * as testComponents from "./components";
 export type componentDataOnChange = (e: React.ChangeEvent<HTMLFormElement>) => void;
 
 export interface IAppState {
-    currentComponentSchema: any;
-    currentComponentData: any;
-    currentComponentConfig?: IFormComponentMappingToPropertyNamesProps;
-    currentComponentOrderByPropertyNames?: IFormOrderByPropertyNamesProps;
-    currentComponentAttributeAssignment?: IFormAttributeSettingsMappingToPropertyNames;
-    currentComponent: any;
-    onChange: componentDataOnChange;
+    currentControlledComponent: any;
+    currentControlledComponentSchema: any;
+    currentControlledComponentData: any;
+    currentControlledComponentConfig?: IFormComponentMappingToPropertyNamesProps;
+    currentControlledComponentOrderByPropertyNames?: IFormOrderByPropertyNamesProps;
+    currentControlledComponentAttributeAssignment?: IFormAttributeSettingsMappingToPropertyNames;
+    controlledOnChange: componentDataOnChange;
+    currentUncontrolledComponent: any;
+    currentUncontrolledComponentSchema: any;
+    currentUncontrolledComponentData: any;
+    currentUncontrolledComponentConfig?: IFormComponentMappingToPropertyNamesProps;
+    currentUncontrolledComponentOrderByPropertyNames?: IFormOrderByPropertyNamesProps;
+    currentUncontrolledComponentAttributeAssignment?: IFormAttributeSettingsMappingToPropertyNames;
+    uncontrolledOnChange: componentDataOnChange;
     showExtendedControls: boolean;
     dataLocation: string;
     schemaLocation: string;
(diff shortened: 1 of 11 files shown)


packages/fast-form-generator-react/app/index.tsx
```diff
@@ -52,48 +59,73 @@ export default class App extends React.Component<{}, IAppState> {
...
                     </div>
+                </div>
+                <div style={{float: "left", width: "50%"}}>
                     <div>
-                        <div>
-                            <select onChange={this.handleComponentUpdate}>
-                                {this.getComponentOptions()}
-                            </select>
+                        <div
+                            style={{width: "300px", minHeight: "100vh", padding: "0 8px", background: "rgb(244, 245, 246)", float: "left", display: "inline-block"}}
```
chrisdholt: I think tests are failing due to max-line-length issues here and on line 94. Can we return this from a function or just define it as a constant and pass it here? Function will probably solve both CircleCI and CodeClimate issues

janechu: Thanks, addressed.

chrisdholt:
@janechu Rebase please :)

packages/fast-form-generator-react/app/index.tsx
```diff
@@ -52,54 +59,86 @@ export default class App extends React.Component<{}, IAppState> {
...
 
     public render(): JSX.Element {
         return (
             <DesignSystemProvider designSystem={designSystemDefaults}>
-                <div>
-                    <div
-                        style={{width: "300px", minHeight: "100vh", padding: "0 8px", background: "rgb(244, 245, 246)", float: "left"}}
-                    >
-                        <Form {...this.coerceFormProps()} />
+                <div style={{float: "left", width: "50%"}}>
```
eljefe223: Should we turn this into a class since the same style is being used on lie 109?

packages/fast-form-generator-react/app/index.tsx
```diff
@@ -52,54 +59,86 @@ export default class App extends React.Component<{}, IAppState> {
...
-     * The schema form attribute settings mapping configuration
-     */
-    private attributeSettingsMappingToPropertyNames: IFormAttributeSettingsMappingToPropertyNames;
-
     constructor(props: {}) {
         super(props);
 
         this.childOptions = this.getChildOptions();
-        this.onChange = this.onChange.bind(this);
+        this.controlledOnChange = this.controlledOnChange.bind(this);
```
nicholasrice: Can you instead define these using arrow syntax?

nicholasrice: Never mind, you are.... Are these bind statements necessary?

packages/fast-form-generator-react/app/index.tsx
```diff
@@ -52,54 +59,86 @@ export default class App extends React.Component<{}, IAppState> {
...
+            currentControlledComponentSchema: testComponents.textField.schema,
+            currentControlledComponentData: getExample(testComponents.textField.schema),
+            currentControlledComponentOrderByPropertyNames: void(0),
+            currentControlledComponentAttributeAssignment: void(0),
+            currentUncontrolledComponent: testComponents.textField.component,
+            currentUncontrolledComponentSchema: testComponents.textField.schema,
+            currentUncontrolledComponentData: getExample(testComponents.textField.schema),
+            currentUncontrolledComponentOrderByPropertyNames: void(0),
+            currentUncontrolledComponentAttributeAssignment: void(0),
+            controlledOnChange: this.controlledOnChange,
```
nicholasrice: It seems a little unconventional to me to store these methods in state - can you speak to why this is necessary?

packages/fast-form-generator-react/src/form/form.tsx
```diff
@@ -123,6 +124,10 @@ class Form extends React.Component<IFormProps & IManagedClasses<IFormClassNameCo
         }
     }
 
+    private isChildLocation(location: IFormLocation): boolean {
+        return location && location.dataLocation.slice(location.dataLocation.length - 8, location.dataLocation.length) === "children";
```
nicholasrice: Why does this subtract 8?

packages/fast-form-generator-react/src/form/form.tsx
```diff
@@ -230,12 +233,19 @@ class Form extends React.Component<IFormProps & IManagedClasses<IFormClassNameCo
      * Generate the section to be shown
      */
     private generateSection(): JSX.Element {
+        const lastComponentTrackerIndex: number = this.state.componentTracker.length - 1;
+        const schema: any = isRootLocation(this.state.activeSchemaLocation)
+            ? this.state.componentTracker[lastComponentTrackerIndex].schema
```
nicholasrice: You should store `this.state.componentTracker[lastComponentTrackerIndex].schema` in a variable so you don't access it more than once.

packages/fast-form-generator-react/src/form/form.tsx
janechu: because 8 is the length of the word "children", I'll add a comment

janechu:
Closing this for now, needs some re-work and updating to issue

Issue #726, opened 2018-07-23
Update text area to not use a monospace font and not allow height/width adjustments.


eljefe223:
Update text area to not use a monospace font and not allow height/width adjustments.

PR #727, opened 2018-07-23
fix(fast-form-generator): remove resize behavior and update fontfamily

eljefe223:
Closes: #726 
 
### DIFF
--- a/packages/fast-form-generator-react/src/form/form-item.textarea.style.ts
+++ b/packages/fast-form-generator-react/src/form/form-item.textarea.style.ts
@@ -19,7 +19,7 @@ const styles: ComponentStyles<IFormItemTextareaClassNameContract, {}> = {
         marginTop: toPx(8),
         marginBottom: toPx(2),
         resize: "none",
-        fontFamily: "Segoe UI, SegoeUI, Helvetica Neue, Helvetica, Arial, sans-serif"
+        fontFamily: "inherit"
     }
 };
 
--- a/packages/fast-form-generator-react/src/form/form-item.textarea.style.ts
+++ b/packages/fast-form-generator-react/src/form/form-item.textarea.style.ts
@@ -17,7 +17,9 @@ const styles: ComponentStyles<IFormItemTextareaClassNameContract, {}> = {
     formItemTextarea_textarea: {
         ...applyInputStyle(),
         marginTop: toPx(8),
-        marginBottom: toPx(2)
+        marginBottom: toPx(2),
+        resize: "none",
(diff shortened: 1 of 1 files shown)


packages/fast-form-generator-react/src/form/form-item.textarea.style.ts
```diff
@@ -17,7 +17,9 @@ const styles: ComponentStyles<IFormItemTextareaClassNameContract, {}> = {
     formItemTextarea_textarea: {
         ...applyInputStyle(),
         marginTop: toPx(8),
-        marginBottom: toPx(2)
+        marginBottom: toPx(2),
+        resize: "none",
+        fontFamily: "Segoe UI, SegoeUI, Helvetica Neue, Helvetica, Arial, sans-serif"
```
chrisdholt: Does "inherit" not work here? If it doesn't, we should consider setting a variable somewhere and importing that. Having multiple instances of these strings throughout the package will prove difficult to maintain.

eljefe223: yes "inherit" does work.

packages/fast-form-generator-react/src/form/form-item.textarea.style.ts
```diff
@@ -17,7 +17,9 @@ const styles: ComponentStyles<IFormItemTextareaClassNameContract, {}> = {
     formItemTextarea_textarea: {
         ...applyInputStyle(),
         marginTop: toPx(8),
-        marginBottom: toPx(2)
+        marginBottom: toPx(2),
+        resize: "none",
```
nicholasrice: Did you consider just restricting to vertical resize? I understand why we wouldn't want to do a horizontal resize as it doesn't align with the container but I could see how vertical resize could be nice if typing long paragraphs.

chrisdholt: I think part of the reason for removing here is that it has poor cross-browser support (no IE or Edge support at all). For a consistent experience we discussed simply removing this all up.

nicholasrice: My opinion is that we should keep vertical if its available - I don't think we should remove useful features just get consistency.

chrisdholt: I'm fine with this ^^^

eljefe223: Just verified that `resize: "vertical"` works in firefox and chrome

PR #732, opened 2018-07-24
feat(fast-components-react-msft): add enzyme and unit tests for MSFT specific components

### LINKED ISSUE #731: Add enzyme for unit testing to fast-components-react-msft
Enzyme should be added and used for unit testing components that are created in the MSFT package. Additional tests are not necessary at this time when we are simply exporting a "styled" component, but components that extend or are specific to MSFT should include unit tests.

chrisdholt:
<body>
closes #731 
<footer>
 

### DIFF
--- a/packages/fast-components-react-msft/src/button/button.spec.tsx
+++ b/packages/fast-components-react-msft/src/button/button.spec.tsx
@@ -41,7 +41,7 @@ describe("button", (): void => {
         );
 
         const button: any = rendered.first().shallow();
-        expect(button.prop("aria-hidden")).not.toBe(undefined);
+
         expect(button.prop("aria-hidden")).toEqual(true);
     });
 });
--- a/packages/fast-components-react-msft/src/button/button.spec.tsx
+++ b/packages/fast-components-react-msft/src/button/button.spec.tsx
@@ -1,0 +1,47 @@
+import * as React from "react";
+import * as Adapter from "enzyme-adapter-react-16";
+import { configure, shallow } from "enzyme";
+import examples from "./examples.data";
+import { generateSnapshots } from "@microsoft/fast-jest-snapshots-react";
+import { ButtonHTMLTags } from "@microsoft/fast-components-react-base";
+import {
+    ButtonProps,
+    IButtonHandledProps,
+    IButtonManagedClasses,
+    IButtonUnhandledProps
+} from "./button";
+
+/*
+ * Configure Enzyme
+ */
+configure({adapter: new Adapter()});
+
+describe("button snapshots", (): void => {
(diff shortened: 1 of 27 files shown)


packages/fast-components-react-msft/src/flipper/flipper.spec.tsx
```diff
@@ -0,0 +1,105 @@
...
+    IFlipperManagedClasses,
+    IFlipperUnhandledProps
+} from "./flipper";
+
+/*
+ * Configure Enzyme
+ */
+configure({adapter: new Adapter()});
+
+describe("flipper snapshot", (): void => {
```
janechu: snapshot or snapshots? button has plural.

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+        };
+
+        const props: IButtonHandledProps & IButtonUnhandledProps = {...handledProps, ...unhandledProps};
+
+        const rendered: any = shallow(
+            <Component {...props} />
+        );
+
+        const button: any = rendered.first().shallow();
+        expect(button.prop("aria-hidden")).not.toBe(undefined);
```
nicholasrice: You don't really need both of these assertions

packages/fast-components-react-msft/src/flipper/flipper.spec.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+        const props: IFlipperHandledProps & IFlipperUnhandledProps = {...handledProps, ...unhandledProps};
+
+        const rendered: any = shallow(
+            <Component {...props} />
+        );
+
+        const flipper: any = rendered.first().shallow();
+
+        expect(flipper.prop("aria-labelledby")).not.toBe(undefined);
```
nicholasrice: same here - the `.not.toBe(undefined` is covered with following test - the other one will fail if it is undefined

packages/fast-components-react-msft/src/heading/heading.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+        const rendered: any = shallow(
+            <Component tag={HeadingTag.h3} />
+        );
+        const heading: any = rendered.first().shallow();
+
+        expect(heading.instance().props.tag).not.toBe(undefined);
+        expect(heading.instance().props.tag).toEqual(HeadingTag.h3);
+    });
+
+    test("should return an instance of the typography component", () => {
```
nicholasrice: why do we care about this? If we refactor and change what this inherits from - why is that a problem as long as the prop interface is the same?

PR #739, opened 2018-07-27
feat(fast-components-react-base): add horizontal overflow

### LINKED ISSUE #725: Create a horizontal overflow component
There should be a new component in `fast-layouts-react` for scrolling through multiple items.

User stories:
- Tony is looking for a new computer game to buy for his daughter Amy for her birthday. He goes onto his computer and navigates the browser to a site that sells games. He sees a list of games but not the one his daughter wants so he clicks an arrow button to show more games. He sees the game his daughter would be interested in and clicks on it which directs him to a purchase page.
- Bobby is a blind man, he's searching on the internet for a refrigerator. He visits a site and using keyboard controls he finds a list of refrigerators. He tabs through the list and finds one with the brand he likes and selects it. This redirects him to another page with the full list of information about that refrigerator.
- Julia is a woman who is waiting at a bus stop after dealing with her overbearing boss. She starts browsing the internet and searches for articles about comfort food. She reads an article about healthy eating being stress relieving and at the bottom of the article there is a caption entitled "recipes high in antioxidants" and below is a list of images of food. She does not see food she is interested in so she swipes across to see more. she finds an image of food she is interested in after she has swiped to the next image set and selects it which navigates her to a page with a recipe for that food.

Functional requirements:
- Multiple items are shown depending on the width of the component
- Two buttons which will allow next/previous movement of items, these should show up on the left/right sides as dictated by the LTR of the page. When clicked they will have an animated scrolling behavior, they will scroll the items in the direction indicated until the next set of items is visible. The next set should be the next or previous set that includes the last item that was previously hidden or partially viewable
- Scrolling through items via touch
- Scrolling through items via tab

Technical spec:
- Scrolling distance is calculated when the next/previous buttons are clicked. It will calculate this by determining the last item that is partially visible or if none is partially visible use the next item that is hidden and moving the items until that item is moved all the way to the other side of the carousel while remaining in view.

Implementation example:
```jsx
<HorizontalOverflow>
    <button slot="previous">Previous</button>
    <img src="https://placehold.it/300x300" />
    <img src="https://placehold.it/300x300" />
    <img src="https://placehold.it/300x300" />
    <img src="https://placehold.it/300x300" />
    <img src="https://placehold.it/300x300" />
    <button slot="next">Next</button>
</HorizontalOverflow>
```

Scenarios:
- previous button is not passed
    - do not show previous button
- next button is not passed
    - do not show next button
- no items are passed
    - do not show the component, throw a warning in the console that items must be passed for the component

janechu:
Closes #725


janechu:
I'm a little low on my testing %, @chrisdholt @nicholasrice can you offer some suggestions?

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
...
+        });
+
+        return previous;
+    }
+
+    private getNext(): JSX.Element {
+        let next: JSX.Element = null;
+        const children: React.ReactNode[] = Array.isArray(this.props.children) ? this.props.children : [this.props.children];
+
+        children.forEach((child: JSX.Element) => {
```
nicholasrice: React has an API for interacting with children that this should probably use: https://reactjs.org/docs/react-api.html#reactchildren

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
...
+                }
+            }
+        }
+
+        return `${itemMaxHeight}px`;
+    }
+
+    private getPrevious(): JSX.Element {
+        let previous: JSX.Element = null;
+        const children: React.ReactNode[] = Array.isArray(this.props.children) ? this.props.children : [this.props.children];
```
nicholasrice: Maybe instead of both a `getNext` and `getPrevious` method you introduce a "getChildBySlot" function. Then you can just use that with a different argument value.

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
...
+    items: HTMLElement[];
+    itemsHeight: string;
+}
+
+class HorizontalOverflow extends Foundation<HorizontalOverflowProps, IHorizontalOverflowState> {
+
+    public static getDerivedStateFromProps(
+        nextProps: HorizontalOverflowProps,
+        prevState: IHorizontalOverflowState
+    ): null | Partial<IHorizontalOverflowState> {
```
nicholasrice: I don't think you need to store items in state. What is the benefit of doing that over excluding next/prev buttons in the the render cycle itself?

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.utilities.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+}
+
+/**
+ * Inspired by the github gist contribution: https://gist.github.com/andjosh/6764939
+ */
+export function scrollLeft(element: HTMLDivElement, left: number, duration: number): void {
+    const start: number = element.scrollLeft;
+    const change: number = left - start;
+    const startDate: number = +new Date();
```
nicholasrice: is that `+` supposed to be in the `new Date()` expression?

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.utilities.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    const easeInOutQuad: EaseInOutQuad = function(
+        currentTime: number,
```
nicholasrice: To avoid creating this function every `scrollLeft` call you should require a duration to be passed to it.

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.utilities.ts
nicholasrice: I see what this does - I'd recommend revising to something more human readable: https://stackoverflow.com/questions/14255664/new-date-is-this-good-practice

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.utilities.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    return children.filter((child: JSX.Element) => {
+        return child.props && child.props.slot !== ButtonDirection.previous && child.props.slot !== ButtonDirection.next;
+    });
+}
+
+/**
+ * Inspired by the github gist contribution: https://gist.github.com/andjosh/6764939
+ */
+export function scrollLeft(element: HTMLDivElement, left: number, duration: number): void {
```
nicholasrice: I still don't think this file should exist. If these are fully reusable outside the context of this component then lets figure out file structure that enables that. If they aren't, they should be private members on the class so we're not allowing exports of functions that were never written to be used outside a specific use-case and context.

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
+import * as React from "react";
+import { get } from "lodash-es";
+import manageJss, { ComponentStyles, IJSSManagerProps, IManagedClasses } from "@microsoft/fast-jss-manager-react";
+import Foundation, { IFoundationProps } from "../foundation";
+import { HorizontalOverflowProps } from "./horizontal-overflow.props";
+import { getItems, scrollLeft } from "./horizontal-overflow.utilities";
```
nicholasrice: Based on earlier conversations this seem a bit more like a component than a layout - thoughts?

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
...
+            this.setState({
+                itemsHeight
+            });
+        }
+    }
+
+    private getItemMaxHeight(): string {
+        let itemMaxHeight: number = 0;
+
+        if (get(this.horizontalOverflowItemsRef, "current.childNodes")) {
```
nicholasrice: You're accessing `get(this.horizontalOverflowItemsRef, "current.childNodes")` twice here - you should store that in a variable to avoid that.

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
...
+            display: "flex",
+            alignItems: "center",
+            height: "100%"
+        }
+    }
+};
+
+export interface IHorizontalOverflowState {
+    items: HTMLElement[];
+    itemsHeight: string;
```
nicholasrice: Did you consider making `state.itemsHeight` a number? If we ever need to do math on this value we'll have to convert it back to a number prior to being able to do that.

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    private itemsOnLoad = (): void => {
+        const itemsHeight: string = this.getItemMaxHeight();
+
+        if (itemsHeight !== this.state.itemsHeight) {
+            this.setState({
+                itemsHeight
+            });
+        }
```
nicholasrice: Can you add documentation of each method describing why it exists?

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
janechu: @chrisdholt do you want to weigh in here? The idea is also still nebulous to me. Once we determine criteria we should file an issue to document the differences so we can offer more clarity in future. If we determine this is more of a component library item I can move it to base, I think it would make more sense.

chrisdholt: @nicholasrice can you expand on that? I feel like this really just serves to facilitate a specific layout for UI elements.

chrisdholt: I'm open to having this live as a component in base. It would provide more control and open up the design system values.

packages/fast-layouts-react/src/horizontal-overflow/horizontal-overflow.tsx
janechu: No math gets done on the value (or conceivably should ever be done) so I wasn't worried about it being an issue but I'll make the change, just in case.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,359 @@
...
+                        {this.getItems()}
+                    </ul>
+                </div>
+                <div className={this.props.managedClasses.horizontalOverflow_previous}>
+                    <div>
+                        <div onClick={this.handlePreviousClick}>
+                            {this.getChildBySlot(ButtonDirection.previous)}
+                        </div>
+                    </div>
+                </div>
```
chrisdholt: Do we need the additional `<div>` wrapping the div with the onClick? It seems unneccessary without more context.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,359 @@
...
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <div
+                className={this.props.managedClasses.horizontalOverflow}
+                onLoad={this.itemsOnLoad}
+            >
+                <div style={{height: `${this.state.itemsHeight}px`, position: "relative", overflow: "hidden"}}>
+                    <ul
+                        className={this.props.managedClasses.horizontalOverflow_items}
+                        style={{position: "relative", whiteSpace: "nowrap", overflowX: "scroll", padding: 0, margin: 0}}
```
chrisdholt: Could we break this style out into a function?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,359 @@
...
+
+        if (distanceFromBeginning === maxDistance) {
+            distance = maxDistance;
+        } else {
+            distance = this.getNextDistance(availableWidth, itemWidths, distanceFromBeginning);
+
+            if (distance >= maxDistance) {
+                distance = maxDistance;
+            }
+        }
```
chrisdholt: With the else, it doesn't look like the original distance value ever gets returned without passing through the function on line 195. Could we simplify this a bit? I could be missing something.

```
if (distanceFromBeginning === maxDistance) {
    return maxDistance;
}

distance = this.getNextDistance(availableWidth, itemWidths, distanceFromBeginning);

return distance >= maxDistance ? maxDistance : distance;
```

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,359 @@
...
+    private getWithinMinDistance(distanceFromBeginning: number, availableWidth: number, itemWidths: number[]): number {
+        let distance: number = 0;
+
+        if (distanceFromBeginning === 0) {
+            distance = 0;
+        } else {
+            distance = this.getPreviousDistance(availableWidth, itemWidths, distanceFromBeginning);
+        }
+
+        return distance;
```
chrisdholt: Similar question here on simplifying without the else:

```
let distance: number = 0;

if (distanceFromBeginning === 0) {
    return distance;
}

return this.getPreviousDistance(availableWidth, itemWidths, distanceFromBeginning);
```

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    /**
+     * Gets the maximum distance that can be scrolled
+     */
+    private getMaxScrollDistance(availableWidth: number, itemWidths: number[]): number {
+        const totalWidth: number = itemWidths.reduce((a: number, b: number) => a + b);
+        return totalWidth - availableWidth;
```
chrisdholt: nit: newline above return?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -0,0 +1,359 @@
...
+
+        currentTime--;
+
+        return -changeInValue / 2 * (currentTime * (currentTime - 2) - 1) + startValue;
+    }
+
+    /**
+     * Scrolls the container for the items list
+     */
+    private scrollLeft(element: HTMLUListElement, left: number, duration: number): void {
```
chrisdholt: Do we need to calculate this differently to account for localization? When testing this, next didn't operate as expected, but the previous button was required to fire when moving to RTL.

From the MDN docs:
```
Note that if the element's direction of the element is rtl (right-to-left) then scrollLeft is 0 when the scrollbar is at its rightmost position (at start of the scrolled content) and then increasingly negative as you scroll towards the end of the content.
```

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: Sure, that looks much better

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: Yep, good catch

janechu: see change on line 184

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.node.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+import * as Adapter from "enzyme-adapter-react-16";
+import { configure, shallow } from "enzyme";
+import HorizontalOverflow, {
+    IHorizontalOverflowClassNameContract,
+    IHorizontalOverflowHandledProps,
+} from "./";
+
+/*
+ * Configure Enzyme
+ */
```
nicholasrice: Do we use `.node.spec` anywhere else? I'm guessing you're trying to highlight that these run in a "node" environment?

I think we should keep all our tests together for now until we can discuss organization as part of our e2e testing work.

PR #758, opened 2018-08-03
feat: add color picker to MSFT dev site

nicholasrice:


packages/fast-components-react-msft/app/app.tsx
```diff
@@ -36,14 +39,21 @@ const hypertextStyles: ComponentStyles<IHypertextClassNameContract, undefined> =
     }
 };
 
-export interface IAppState {
+enum Themes {
```
awentzel: Nit:  `Themes` is plural, yet, `Direction` is singular.  Shouldn't `Themes` be `Theme` for consistency?  The usage would then read `Theme.light` not `Themes.light` unless there will be multiple light themes?

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -16,11 +16,14 @@ import Site, {
     SiteMenuItem,
     SiteTitle,
     SiteTitleBrand,
+    ShellSlot
 } from "@microsoft/fast-development-site-react";
 import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
 import { Direction } from "@microsoft/fast-application-utilities";
 import * as examples from "./examples";
 import Hypertext from "../src/hypertext";
+import ColorPicker, { IColorConfig } from "./color-picker";
```
awentzel: nit: In the past we would alphabetize import statements.  Is this still a standard we want to follow?

packages/fast-components-react-msft/app/color-picker.tsx
```diff
@@ -0,0 +1,105 @@
...
+            : "accentColor";
+
+        if (typeof this.props.onColorUpdate === "function") {
+            this.props.onColorUpdate(Object.assign({}, this.props, { [updatedColorKey]: value}));
+        }
+    }
+
+    /**
+     * Ensures that colors are properly formatted
+     */
```
awentzel: If there are input/output parameters I would recommend adding them to the function comments.  Every function, should have comments, inputs, outputs, return statements to help in understanding details for new contributors.

packages/fast-components-react-msft/app/color-picker.tsx
```diff
@@ -0,0 +1,105 @@
...
+    private formatColor(color: string) {
+        // Color inputs don't like three digit hex values, so we need to convert it to 6
```
awentzel: this comment could probably be moved to function comment.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -295,6 +294,22 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
         locales: ["en", "en-rtl"]
     };
 
+    public static getDerivedStateFromProps(props: ISiteProps, state: ISiteState): Partial<ISiteState> | null {
+        if (props.activeTheme) {
+            return props.activeTheme !== state.theme ? { theme: props.activeTheme } : null;
```
awentzel: Rather than returning null.  It might be nice to consider having application logic to handle throwing exceptions rather than null statements.  The exception could be "BusinessExceptionNoTheme" for example.  Then it can be gracefully handled by the calling function and more reusable.  They might go more along the side of application instrumentation/exception management which I have in future planning.  But, just an idea.

nicholasrice: This is a React lifecycle hook that expects a null return value to signal no changes to state, so it's not really up to us as that is the API that React landed on.

packages/fast-components-react-msft/app/color-picker.tsx
nicholasrice: In general I would like to defer to TypeScript to describe the inputs and outputs because those are 100% guaranteed to be accurate - comments sometimes lie.

I think comments describing the API can be helpful if the API is hard to type or hard to understand, but I think that explicitly documenting the API in comments as opt-in based on reasoning above. Our type system is pretty accurate and descriptive so I'd like to utilize that as much as possible to avoid redundant and error prone work.

packages/fast-components-react-msft/app/app.tsx
nicholasrice: We alphabetize inside a statement but I don't think we're alphabetizing the statements themselves. To me, there is a real benefit to that.

packages/fast-components-react-msft/app/color-picker.tsx
nicholasrice: Moved

packages/fast-components-react-msft/app/app.tsx
nicholasrice: Changed to `Theme` to align

packages/fast-development-site-react/src/components/site/index.tsx
awentzel: Thanks for the clarification.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -830,13 +845,15 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
...
-        this.setState({
-            theme: selectedTheme
-        });
+        if (!this.props.activeTheme) {
+            const selectedTheme: ITheme = this.props.themes.find((item: ITheme): boolean => {
+                return item.id === e.target.value;
+            });
+    
+            this.setState({
+                theme: selectedTheme
```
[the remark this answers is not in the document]
nicholasrice: Good eye! I've changed it so that selecting an accent color does not adjust the theme

packages/fast-components-react-msft/app/color-picker.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const styles: ComponentStyles<IColorPickerManagedClasses, IDesignSystem> = {
+    colorPicker: {
+        display: "flex",
+        height: "100%",
+        alignItems: "center"
+    },
+    colorPicker_label: {
+        margin: "0 12px"
```
janechu: we can use the toPx function here

packages/fast-components-react-msft/app/color-picker.tsx
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Ensures that colors are properly formatted
+     * Color input elements don't understand three digit hex values, so we need to convert them to 6
+     */
+    private formatColor(color: string): string {
+        const threeDigitHex: RegExp = /\#([a-fA-F0-9]{3})$/g;
+        const match: string[] | null = threeDigitHex.exec(color);
+
+        return Array.isArray(match)
+            ? `#${match[1].split("").map((charecter: string) => charecter + charecter).join("")}`
```
janechu: *character

nicholasrice: thanks!

packages/fast-components-react-msft/app/color-picker.tsx
nicholasrice: I'd prefer to not  preform an operation if we don't need to.

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -109,9 +137,40 @@ export default class App extends React.Component<{}, IAppState> {
...
-        this.setState({
-            theme
+    private handleUpdateTheme = (theme: Theme): void => {
+        if (theme !== Theme.custom) {
+            this.setState({
+                theme,
+                foregroundColor: theme === Theme.dark ? DesignSystemDefaults.backgroundColor : DesignSystemDefaults.foregroundColor,
+                backgroundColor: theme === Theme.dark ? DesignSystemDefaults.foregroundColor : DesignSystemDefaults.backgroundColor
+            });
+        } else {
```
chrisdholt: Do we need the  { else ?

chrisdholt: Yes, we need the else.

nicholasrice: ;)

PR #761, opened 2018-08-04
feat(fast-components-react-base): add tabs component

### LINKED ISSUE #749: Create tabs as a composable set of components
There should be a ~~tab panel~ tabs component in `fast-components-react-base` that implements the tab ~~panel~~ interface.

User stories:
- Wanda has come back from a long day at work and is trying to order some food online. The restaurant she's looking at offers drinks, main courses and deserts. Not wanting to spend the extra premium on drinks which are currently shown, she clicks on main courses and makes a selection from the items that show up. She completes her order and her food is delivered.
- Amy is trying to find book a flight, she sees the current deals for today and next to it is another section for booking. She clicks the section for booking and the current deals are replaced with a form for her to fill out booking information. She completes the form and her flight has been booked.

Functional requirements:

**Initial view**:
- Multiple selectable items appear next to each other, one item has an "active" state
- Below the selectable items is content that corresponds to the currently selected "active" item

**Selecting an inactive item**:
- When an inactive item is selected, it's state is changed to "active" and the previous "active" item is changed to inactive
- Content changes corresponding to the tab selected
- Items can be set to an active state by mouse click or by left/right arrow keys once the selectable items are in focus

Technical spec:
- Content is swapped when a different tab is selected by the user
- An optional callback is used to inform the user what uuid is for activation

Implementation:
Note: updated to reflect @chrisdholt notes relating to https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel where:
- **Tabs** - the containing component for the tab elements and tab panels
- **TabItem** - the container for both the tab element and the tab panel it controls
- **Tab** - the controlling element
- **TabPanel** - the tab panel containing content relating to the controlling tab element

```diff
-<TabPanels selected="id01">
-    <TabPanel id="id01" slot="tab-panel">
-        <Tab slot="tab">
-            Tab 1
-        </Tab>
-        <TabContent slot="tab-content">
-            Tab 1 content
-        </TabContent>
-    </TabPanel>
-</TabPanels>

-<Tabs selected="id01">
+<Tabs activeId="id01">
+    <TabItem id="id01" slot="tab-item">
+        <Tab slot="tab">
+            Tab 1
+        </Tab>
+        <TabPanel slot="tab-panel">
+            Tab panel content
+        </TabPanel>
+    </TabItem>
+</Tabs>
```

Naming clarification: The HTML attribute role of *tabpanel* although one word is, according to spec, two words "tab panel" and as such we will use this as the basis for naming for the tab panel component when implementing the component name and slot.

Scenarios:
- No ~~TabPanel~~ `TabItem` is passed
    - Do not show ~~TabPanels~~ `Tabs`
- A ~~TabPanel~~ `TabItem` is missing `Tab` or ~~TabContent~~ `TabPanel`
    - Do not show the ~~TabPanel~~ `TabItem`

janechu:
Resolves #749


packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+     * Gets the current index by tab item ID
+     */
+    private getCurrentIndexById = (item: JSX.Element): boolean => {
+        return item.props.id === this.state.activeId;
+    }
+
+    /**
+     * Gets the child by the slot property
+     */
+    private getChildBySlot(children: React.ReactNode, slot: TabSlot): JSX.Element[] {
```
nicholasrice: I feel like `getChildBySlot` could be a generic utility - slots are something we use somewhat frequently

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+        };
+    }
+
+    /**
+     * Renders the component
+     */
+    public render(): JSX.Element {
+        const tabElements: JSX.Element[] = this.renderTabElements();
+
+        if (tabElements.length > 0) {
```
nicholasrice: I think I'm partial to always rendering the HTML that we can render. Deciding to render or not render is really an implementation concern - not a component concern.

I think I'd be a little confused if I dropped the component on a page but it didn't do anything. Its similar to the silent error scenario - I now have to know about the internal code logic to accurately debug. I don't think this is a major use case so I'm not too concerned about it, but I think I prefer to keep the components as simple as possible and leave implementation concerns to the authors.

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    /**
+     * Renders the tab panels
+     */
+    private renderTabPanels(): JSX.Element[] {
+        const tabPanels: JSX.Element[] = [];
+
+        this.getChildBySlot(this.props.children, TabSlot.tabItem).forEach((tabItem: JSX.Element, index: number): void => {
+            const id: string = get(tabItem, "props.id") ? tabItem.props.id : `tab${index}`;
```
nicholasrice: If we have tab components on the page without ID's, we've created two identical ID values.

Do we need to do work here? I kind of think we should just require a unique ID on the item. That way, if there are two unique ID's on the page, it is *always* the fault of the author and not a bug in the component.

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+            tabPanels.push(
+                <div
+                    key={id}
+                    id={id}
+                    className={this.props.managedClasses.tab_panel}
+                    role="tabpanel"
+                    aria-labelledby={id}
+                    aria-hidden={this.state.activeId !== id ? true : false}
```
nicholasrice: nit: `aria-hidden={this.state.activeID !== id}`. The `true : false` is redundant

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,313 @@
...
+                this.activateLast();
+                break;
+        }
+    }
+
+    /**
+     * Activates the previous tab item
+     */
+    private activatePrevious(): void {
+        const items: JSX.Element[] = this.getChildBySlot(this.props.children, TabSlot.tabItem);
```
nicholasrice: There is a decent amount of duplication between `activePrevious`, `activateNext`, `activateFirst`, and `activateLast` - I think you could refactor to a single function that shifts the `activeId` by a particular value instead: `shiftActiveTab(-1)`, `shiftActiveTab(1)`, `shiftActiveTab(this.state.items.length - this.getChildBySlot(this.props.children, TabSlot.tabItem)`, etc

packages/fast-components-react-base/src/tabs/tab-item.props.ts
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * The unique id for the tab item
+     */
+    id: string;
+
+    /**
+     * The slot identifying this component to the TabItem component as a tab-item
+     * the string passed must be "tab-item"
+     */
+    slot: string;
```
nicholasrice: `slot` should be an enum

packages/fast-components-react-base/src/tabs/tab-panel.props.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+
+export interface ITabPanelHandledProps {
+    /**
+     * The slot identifying this component to the TabPanel component as a tab-panel
+     * the string passed must be "tab-panel"
+     */
+    slot: string;
```
nicholasrice: `slot` should also be an enum here

packages/fast-components-react-base/src/tabs/tab-panel.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+import Foundation, { HandledProps } from "../foundation";
+import { ITabPanelHandledProps, ITabPanelUnhandledProps } from "./tab-panel.props";
+
+class TabPanel extends Foundation<ITabPanelHandledProps, ITabPanelUnhandledProps, {}> {
+    public render(): JSX.Element {
+        return (
+            <React.Fragment>
+                {this.props.children}
```
nicholasrice: This doesn't seem to be handling any of the handled props

packages/fast-components-react-base/src/tabs/tab-item.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from "react";
+import Foundation, { HandledProps } from "../foundation";
+import { ITabItemHandledProps, ITabItemUnhandledProps } from "./tab-item.props";
+
+class TabItem extends Foundation<ITabItemHandledProps, ITabItemUnhandledProps, {}> {}
```
nicholasrice: This also isn't handling any of the handled props, or the unhandled props. Also, this has no render method, is that intentional?

packages/fast-components-react-base/src/tabs/examples.data.tsx
```diff
@@ -0,0 +1,117 @@
...
+    TabPanel
+} from "./tabs";
+import schema from "./tabs.schema.json";
+import Documentation from "./.tmp/documentation";
+import { Orientation } from "./tabs.props";
+
+function getTabItem(index: number): JSX.Element {
+    return (
+        <TabItem slot="tab-item" id={`tab0${index}`}>
+            <Tab slot="tab">
```
nicholasrice: It seems a little odd to me that we have a `Tab` component that requires a `slot={"tab"}`. Do we really even need `Tab` and `TabPanel` components? I don't really follow the purpose they serve.

janechu: They are basically buckets in which there is an opportunity for children to be placed. So for example if I wanted to put an SVG in the `Tab` I can do that as well as any content I want in the `TabPanel`. The prop of slot is to follow our convention that we are going to be syncing this up with an API for web components in the future so in order to find an appropriate slot I need the prop. Also there's no way of knowing once the components are minified what they are, in fact you could use any element or a custom component as long as you have the slot prop with the correct string.

janechu: Additionally `TabItem` must contain a `Tab` and a `TabPanel` in order to be a complete item, that's really what the purpose is. If it does not then we don't count it as a viable `TabItem`.

packages/fast-components-react-base/src/tabs/tab-item.tsx
janechu: Yes, it's really just for semantic convenience, it does not even need to be a component it could be a `div` or any other element the user passes since we are only using it as a wrapper to find the contents and verify it contains both the `Tab` and `TabPanel`. I think it's more clear from an implementation perspective that this component exists however because I think it's how the user expects to compose it.

packages/fast-components-react-base/src/tabs/tabs.tsx
janechu: oops, great point

packages/fast-components-react-base/src/tabs/tabs.tsx
janechu: Very true, I think we should have an issue for it though

packages/fast-components-react-base/src/tabs/tabs.tsx
janechu: sounds good

packages/fast-components-react-base/src/tabs/tab-item.tsx
nicholasrice: At some level, I agree but it feels like we're mixing metaphors here. We're exposing a purpose-specific component that, unless *configured correctly*, will not work as expected. I would expect, if these sub-components we're available, that given a nesting structure it would do the right thing.

With this implementation though, it won't. Slots are the only thing that *actually* matter.

janechu: Gotcha, I agree, I've changed this up to not use slots but check the component type as we are doing in `fast-layouts-react` for the `Grid` component.

chrisdholt:
I think we'll want `Tab` and `Tabpanel` to operate as individual components and accept unhandled props. Essentially, we probably need individual components for everything but the combinded `Tabitem`.

packages/fast-components-react-base/src/tabs/tabs.props.ts
```diff
@@ -0,0 +1,53 @@
+import * as React from "react";
+import { IManagedClasses, ITabsClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export enum Orientation {
```
nicholasrice: This is a pretty non-specific name so I think it should either be:
a.) Semantically scoped to Tabs (eg, `TabsOrientation`)
b.) Made generic in a non-tabs specific location

packages/fast-components-react-base/src/tabs/tabs.props.ts
```diff
@@ -0,0 +1,53 @@
...
+     */
+    children?: React.ReactNode | React.ReactNode[];
+
+    /**
+     * The aria-label applied to the tablist for the tab items
+     */
+    label: string;
+
+    /**
+     * The tab update callback
```
nicholasrice: Can you make the `onUpdateTab` comment more descriptive? Why would I want to use this prop? When and why is this function called?

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,336 @@
+import * as React from "react";
+import { get } from "lodash-es";
+import { KeyCodes } from "@microsoft/fast-web-utilities";
+import { IManagedClasses, ITabsClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import Foundation, { HandledProps } from "../foundation";
+import { ITabsHandledProps, ITabsManagedClasses, ITabsUnhandledProps, TabsProps } from "./tabs.props";
+
+export enum TabSlot {
```
nicholasrice: this should probably be `TabsSlot`

packages/fast-components-class-name-contracts-base/src/tabs/index.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * The class name contract for the tabs component
+ */
+export interface ITabsClassNameContract {
+    tab_item: string;
```
nicholasrice: These should be `tabs_tabItem` and `tabs_tabList`

packages/fast-components-class-name-contracts-base/src/tabs/index.ts
```diff
@@ -0,0 +1,18 @@
...
+    tabs: string;
+}
+
+export interface ITabClassNameContract {
+    tab: string;
+    tab__active: string;
+}
+
+export interface ITabPanelClassNameContract {
+    tab_panel: string;
```
nicholasrice: `tabPanel` & `tabPanel__hidden`

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,336 @@
...
+     * Renders the tab elements
+     */
+    private renderTabElements(): JSX.Element[] {
+        const tabElements: JSX.Element[] = [];
+
+        this.getChildBySlot(
+            this.props.children,
+            this.getSlot(TabSlot.tabItem)
+        ).forEach((tabItem: JSX.Element, index: number): void => {
+            tabElements.push(
```
nicholasrice: This is pretty nitty but I think we should avoid mutating functions when possible. This operation is safe, but I think reducing mutating operations where we can helps mitigate folks referencing code examples and applying them to new areas where mutating functions may be problematic.

By using `map` instead of `forEach`, you avoid both mutation and creating a variable - you can just return the result of the `map` function.

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,336 @@
...
+    private activatePrevious(): void {
+        const items: JSX.Element[] = this.getChildBySlot(
+            this.props.children,
+            this.getSlot(TabSlot.tabItem)
+        );
+        const currentItemIndex: number = items.findIndex(this.getCurrentIndexById);
+        const previousItemIndex: number = currentItemIndex > 0 ? currentItemIndex - 1 : items.length - 1;
+        const previousItemId: string = items[previousItemIndex].props.id;
+
+        if (!this.props.activeId) {
```
nicholasrice: All of the "activate" methods share a common pattern of updating state to a particular index and focusing the element. I would reccomend centralizing that logic into a discrete function to reduce duplication and complexity

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,336 @@
...
+                return typeof this.props.tabItemSlot === "string" ? this.props.tabItemSlot : TabSlot.tabItem;
+            case TabSlot.tabPanel:
+                return typeof this.props.tabPanelSlot === "string" ? this.props.tabPanelSlot : TabSlot.tabPanel;
+        }
+    }
+
+    /**
+     * Gets the child by the slot property
+     */
+    private getChildBySlot(children: React.ReactNode, slot: TabSlot | string): JSX.Element[] {
```
nicholasrice: This returns a collection of children, correct? I'd recommend changing the function name to reflect that - this makes it sound like this function will return a single child.

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -0,0 +1,336 @@
...
+        const childBySlot: JSX.Element[] = [];
+
+        React.Children.forEach(children, (child: JSX.Element): void => {
+            if (child.props && child.props.slot === slot) {
+                if (slot === this.getSlot(TabSlot.tabItem)) {
+                    if (
```
nicholasrice: I'd also reccomend moving the bit of logic to filter out `TabItem` elements that don't have both a `tabPanel` and `tab` to another function. The name of this function implies that I'm just returning children that match a particular slot.

That keeps this function single-purpose and provides a semantic name to tie to the behavior of only rendering components that have both `tab` and `tabPanel` elements.

Issue #555, opened 2018-06-15
Change the dev-tools tabType to an enum


janechu:
The tabType should be an enum for better type safety and clarity.

PR #765, opened 2018-08-07
feat(fast-development-site-react): add enums for tabType for better type safety and clarity

nalogiudice:
Closes [#555](https://github.com/Microsoft/fast-dna/issues/555)
<body>

<footer>
 

### DIFF
--- a/packages/fast-development-site-react/src/components/site/dev-tools.tsx
+++ b/packages/fast-development-site-react/src/components/site/dev-tools.tsx
@@ -5,12 +5,12 @@ import { IDevSiteDesignSystem } from "../design-system";
 import CodePreview from "./dev-tools-code-preview";
 import { IFormChildOption } from "./";
 
-export enum FrameworkEnum {
+export enum Framework {
     react = "react",
     angular = "angular"
 }
 
-export enum TabTypeEnum {
+export enum TabType {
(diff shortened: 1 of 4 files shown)


packages/fast-development-site-react/src/components/site/dev-tools.tsx
```diff
@@ -10,7 +10,11 @@ export enum FrameworkEnum {
     angular = "angular"
 }
 
-export type tabType = "Code" | "Properties" | "Schema";
+export enum TabTypeEnum {
```
chrisdholt: I know that we have it on the "FrameworkEnum" above, but I think that's actually a mistake. Adding enum to a typed "enum" seems totally redundant to me. I'm sure you were following the existing pattern, but I don't know that it's the pattern we want.

chrisdholt:
@nalogiudice updating this label to `pr : feature`

packages/fast-development-site-react/src/components/site/dev-tools.tsx
nalogiudice: Sure, I can remove the word Enum - I can see how it is similar to saying  "ATM machine" :)

nalogiudice:
@chrisdholt both enums updated.

PR #767, opened 2018-08-07
fix(fast-web-utilities): fix an issue where getClientRectWithMargin threw an error in EDGE and IE strict modes due to readonly properties

### LINKED ISSUE #764: Element.getBoundingClientRect return value is read-only
It looks like this is the offending line:
https://github.com/Microsoft/fast-dna/blob/master/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx#L116

chrisdholt:
closes #764 
<body>

<footer>
 

### DIFF
--- a/packages/fast-web-utilities/src/html.ts
+++ b/packages/fast-web-utilities/src/html.ts
@@ -17,7 +17,7 @@ export function getClientRectWithMargin(element: HTMLElement): IClientRectWithMa
 
     const rect: ClientRect = element.getBoundingClientRect();
     const style: CSSStyleDeclaration = window.getComputedStyle(element, null);
-    const clone: any = {
+    const clone: IClientRectWithMargin = {
         width: rect.width,
         height: rect.height,
         top: rect.top,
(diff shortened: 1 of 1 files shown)


packages/fast-web-utilities/src/html.ts
```diff
@@ -1,13 +1,30 @@
...
-export function getClientRectWithMargin(element: HTMLElement): ClientRect | DOMRect {
+export function getClientRectWithMargin(element: HTMLElement): IClientRectWithMargin {
     if (!element) {
         return;
     }
 
-    const clone: any = element.getBoundingClientRect();
+    const rect: ClientRect = element.getBoundingClientRect();
     const style: CSSStyleDeclaration = window.getComputedStyle(element, null);
+    const clone: any = {
```
nalogiudice: should this typing be `any` ?

chrisdholt: updated.

PR #769, opened 2018-08-09
fix(hyperlink): correctly handle href as a handled prop and update documentation

### LINKED ISSUE #768: Hypertext: `href` is a handled prop but not handled correctly in code; documentation is incorrect
Hypertext enumerates href in the schema as well as in the prop interface, but it is not listed as a handled prop in the .tsx file, nor is it passed directly to the component. To ensure that the href property is removed from the DOM, this should be passed directly to the component with a value or as null or undefined.

In addition, the documentation gives guidance in the base component where it is not necessary and the guidance given does not fully align with the current implementation guidance for the component.

chrisdholt:
<body>
closes #768
<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/hypertext/README.md
+++ b/packages/fast-components-react-base/src/hypertext/README.md
@@ -1,2 +1,6 @@
 ## Guidance
 The *hypertext* component is an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) element which accepts children. *Hypertexts* most commonly appear as inline elements within the context of a sentence or a paragraph.
+
+
+## Accessibility
+A *hypertext* should have two points of visual distinction when *inline*.
--- a/packages/fast-components-react-base/src/hypertext/README.md
+++ b/packages/fast-components-react-base/src/hypertext/README.md
@@ -1,2 +1,2 @@
 ## Guidance
-The *Hypertext* component is an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) element which accepts children. *Hypertexts* most commonly appear as inline elements within the context of a sentence or a paragraph.
+The *hypertext* component is an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) element which accepts children. *Hypertexts* most commonly appear as inline elements within the context of a sentence or a paragraph.
--- a/packages/fast-components-react-base/src/hypertext/README.md
+++ b/packages/fast-components-react-base/src/hypertext/README.md
@@ -1,6 +1,5 @@
-## Guidance
+# Hypertext
 The *hypertext* component is an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) element which accepts children. *Hypertexts* most commonly appear as inline elements within the context of a sentence or a paragraph.
-
 
 ## Accessibility
 A *hypertext* should have two points of visual distinction when *inline*.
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-base/src/hypertext/README.md
```diff
@@ -1,14 +1,2 @@
...
-Two *hypertexts* can appear consecutively on separate lines of text and wrap. *Hypertext* text must be clear and unique. Multiple *hypertext* to same location should have identical link text.
-
-### Behavioral guidance
-If you need to initiate actions (e.g. download a file), consider using *action trigger* instead. For calls to action (e.g. "buy"), consider using *call to action*, and for the final action in a workflow (e.g. "save"), consider using *button*. Keep *hypertexts* to just a few words, 4-5 max, not entire sentences. Don't include the period if *hypertext* is at the end of a sentence.
-
-## Accessibility
-A *hypertext* is required to have two points of visual distinction when *inline*. The underline and visual state colors are both required for accessibility.
\ No newline at end of file
+## Guidance
+The *Hypertext* component is an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) element which accepts children. *Hypertexts* most commonly appear as inline elements within the context of a sentence or a paragraph.
```
nalogiudice: I feel strongly we should **not** be capitalizing our components.

packages/fast-components-react-msft/src/hypertext/README.md
```diff
@@ -1,11 +1,8 @@
-## Best practices
-Use *hypertext* when you need to navigate the user to more information about the text that was selected. *Hypertexts* are best used for tertiary actions outside the main work flow.
+## Guidance
+The *Hypertext* component is an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) element which accepts children. *Hypertexts* most commonly appear as inline elements within the context of a sentence or a paragraph.
```
nalogiudice: Same with all of these

packages/fast-components-react-base/src/hypertext/README.md
chrisdholt: Since documentation is part of this, I will update here. We should document this somewhere so it's explicit.

packages/fast-components-react-msft/src/hypertext/README.md
chrisdholt: I would assume the whole file based on the previous comment :)

chrisdholt: resolved.

packages/fast-components-react-base/src/hypertext/README.md
chrisdholt: resolved. With exception of those that start a sentence.

packages/fast-components-react-msft/src/hypertext/README.md
```diff
@@ -1,11 +1,8 @@
...
 
 ### Usage
-*Hypertexts* may appear as inline elements. Use inline *hypertext* within the context of a sentence or a paragraph. Within a paragraph, *hypertext* should inherit the same formatting as the neighboring text as far as font size and font weight. Keep *hypertexts* to just a few words, 4-5 max, not entire sentences. Don't include the period if *hypertext* is at the end of a sentence.
-
-### Style guidance
-Two *hypertexts* can appear consecutively on separate lines of text and wrap. *Hypertext* text must be clear and unique. Multiple *hypertext* to same location should have identical link text.
+Use *hypertext* when you need to navigate the user to more information about the text that was selected. *Hypertexts* are best used  as inline elements within the context of a sentence or a paragraph. Keep *hypertexts* to just a few words, 4-5 max, not entire sentences. Don't include the period if *hypertext* is at the end of a sentence. *Hypertext* without an `href` prop should appear as normal text
```
chrisdholt: missing period.

PR #776, opened 2018-08-11
fix: development builds of fast-components-react-msft should not be minified

### LINKED ISSUE #771: fast-components-react-msft should not minify development builds
It seems that `fast-components-react-msft` always runs uglify-js in development builds. This makes it hard to debug issues and minifies all class names.

expected: minified builds are only created during production builds
actual: minified builds are always created

nicholasrice:
Ensures that only production builds are minified.

fixes #771


packages/fast-components-react-msft/package.json
```diff
@@ -20,7 +20,7 @@
     "copy:all": "npm run copy:json-schema && npm run copy:readme",
     "copy:json-schema": "node ../../build/copy-schemas.js",
     "copy:readme": "node ../../build/copy-readme.js",
-    "build:app": "webpack --progress",
+    "build:app": "cross-env NODE_ENV=production webpack --progress",
```
awentzel: There's a new method of doing this I discovered this week `--mode=production` or `--mode=development`.  Instead of using NODE_ENV cross-env etc.  In the latest version of Webpack (4.16.5) you can use the `mode` [option](https://webpack.js.org/concepts/mode/). 

For example:
```js
module.exports = (env, argv) => {
    if(argv.mode === 'development') {
        config.output.pathinfo = true;
        config.output.filename = "assets/scripts/kw-[name].bundle.js";
        console.log('Environment: ' + 'development');
    }
    if(argv.mode === 'production') {
        config.output.filename = "assets/scripts/kw-[hash].min.bundle.js";
        console.log('Environment: ' + 'production');
    }

    return config;
};
```

Then in your package.json file i'm doing this
```json
 "scripts": {
    "serve-pro": "webpack --config ./build/webpack/webpack.config.js --progress --mode=production",
    "serve-dev": "webpack-dev-server --no-inline --config ./build/webpack/webpack.config.js --progress --mode=development"
  },
```

awentzel: I believe it also does some of the minimizing etc automatically now as well, but, not certain.

nicholasrice: Oh cool - I'll take a look at that - gonna close this for the time being then.

Regarding "server-pro" and "serve-dev" - `npm start` is the way I've seen most projects launch the development environment. I'm in favor of leaving our commands as-is and just fixing them to not minify in the development environment.

The automatic minimizing comes from the production mode - that's why I'm able to rip out uglify and the manual usage of it.

Issue #596, opened 2018-06-27
Create accessibility documentation for all react components


janechu:
For the packages:
- fast-components-react-base
- fast-components-react-msft

Add to the README.md files links to the relevant MDN (Mozilla Development Network) page(s) for the element(s) used (such as button HTML element for the Button component).

Also add any necessary Accessibility spec from w3 for example a menu should consult the w3 page https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus.

PR #780, opened 2018-08-14
docs(fast-components-react-base): add accessibility documentation and core documentation clean up

nalogiudice:
Closes [#596](https://github.com/Microsoft/fast-dna/issues/596)
<body>

<footer>
 


packages/fast-components-react-base/src/horizontal-overflow/README.md
```diff
@@ -1,2 +1,5 @@
 # Horizontal overflow
-The *Horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions.
\ No newline at end of file
+The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions.
+
+## Accessibility
+The 'next' and 'previous' buttons are only useful for sighted users, since overflow is a purely visual construct, so they should be hidden from screen readers. If the items in the overflow need to be tabbable, users can add the appropriate event handlers or `tabindex`.
```
chrisdholt: "If the items in the overflow need to be tabbable, users can add the appropriate event handlers or `tabindex`." 

I think we should either improve this and expand on the documentation, or leave this out. As of now it seems generic and not necessarily useful. What is the scenario where something would need to be tabbable and not already in the tabindex? What assumptions are we making about the implementation where adding event handlers or managing tabindex is the provided solution? This seems more confusing than helpful to me.

packages/fast-components-react-base/src/image/README.md
```diff
@@ -1,8 +1,5 @@
 # Image
-Use *image* to break up text-heavy pages, to illustrate products for sale, or to improve storytelling on a page. The use of *image* should always support the page’s story and, when possible, contribute to that story. Size the *image* to complement the rest of the page content. Use the **picture** option to employ a selection of image versions for different viewport sizes.
-
-## Style guidance
-Don't use *image* as a substitute for *hero*, *feature*, or *product* and *content placements*. Don't use an image if it reduces the effectiveness of the page.
+Use *image*, based on the standard [image](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) and [picture](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) elements, to break up text-heavy pages, to illustrate products for sale, or to improve storytelling on a page. Use the **picture** option when using a selection of image versions for different viewport sizes.
```
chrisdholt: Can we find a way to remove this:

"to break up text-heavy pages, to illustrate products for sale, or to improve storytelling on a page. "

That seems very "brand" guidance specific and not necessarily helpful for the base component.

packages/fast-components-react-base/src/label/README.md
```diff
@@ -1,16 +1,5 @@
...
-
-## Usage guidance
-Use sentence case (e.g. "Last name"), but avoid punctuation such as colons. *Label* text should be short and succinct, not a full sentence. When labeling components, write the *label* text as a noun or a concise noun phrase, without using full sentences. Using commas to separate list items is an exception to this rule.
-
-## Behavioral guidance
-When instructional text is needed near a group of controls, consider using caption instead. *Labels* are only intended for identification purposes, not as instructional text.
+*Label* is a replacement for the HTML [label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) or `legend` element. It can be used alone or grouped with other related components. Some components have the *label* built-in, but most do not. *Labels* cannot exceed one line of text.
```
chrisdholt: Do we want to say that this is a "replacement" for the HTML label/legend elements?

packages/fast-components-react-base/src/button/README.md
```diff
@@ -1,2 +1,5 @@
 # Button
-The *Button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element but will appear with button styling.
+*Buttons* are commonly used for final actions that complete a user work flow and the text should communicate what will happen after interaction. Typical use cases include submitting a form, closing a [dialog](/components/dialog/), or saving settings. The *button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element, but will appear with button styling.
```
nicholasrice: The base component doesn't assert any styling requirements so I think you should remove the last half of the last sentence.

packages/fast-components-react-base/src/dialog/README.md
```diff
@@ -1,5 +1,5 @@
 # Dialog
-The *Dialog* component accepts children and can either be a modal, or a non-modal dialog. Modal dialogs prevent users from interacting with the rest of the page content, while non-modal dialogs allow users the ability to continue interacting with the page. The width and height of the dialog content area can be set by props. If no props are passed to set width or height, the dialog width will default to 640px and the height will default to 480px.
+The *dialog* component accepts children and can either be a modal, or a non-modal *dialog*. Modal *dialogs* prevent users from interacting with the rest of the page content, while non-modal *dialogs* allow users the ability to continue interacting with the page. The width and height of the *dialog* content area can be set by props. If no props are passed to set width or height, the *dialog* width will default to 640px and the height will default to 480px.
```
nicholasrice: Not you, but this wording around default values for width and height seem to me that they are just waiting to get out of sync - maybe we should remove here? Defaults will get communicated when typedoc info gets integrated.

packages/fast-components-react-base/src/dialog/README.md
```diff
@@ -1,5 +1,5 @@
...
\ No newline at end of file
+A *dialog* must include *at least* one focusable control. When the *dialog* appears on the screen, keyboard focus should move to the first focusable element. While not required to be focused, a *dialog* should include a label. This can be applied as an `aria-label` using the `label` prop, or by referencing a child within the *dialog* with the `aria-labelledby` attribute using the `labelledBy` prop. Optionally, additional context can be provided to assistive technologies by including an `aria-describedby` attribute which can be applied using the `describedBy` prop. Learn more about *dialog* accessibility at [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role) and [W3C WAI-ARIA documentation](https://www.w3.org/TR/wai-aria-practices/#dialog_modal).
```
nicholasrice: I think this paragraph should only speak to the React props and not the underlying `aria` attributes. My feeling is that it should be sufficient accessibility guidance to document the abstraction and not the implementation details when we're handling the implementation details directly. eg.
"While not required to be focused, a *dialog* should include a label. The ID of that label should be provided to the *dialog's* labelledBy prop...."

packages/fast-components-react-base/src/toggle/README.md
```diff
@@ -1,11 +1,5 @@
...
-
-## Usage guidance
-Only replace the **on** and **off** labels if there are more specific labels for the setting. If there are short, 3-4 character labels that represent binary opposites that are more appropriate for a particular setting, use them. For example, you might use **show** or **hide** if the setting is "Show images." Using more specific labels can help when dealing with different languages.
-
-## Behavioral guidance
-If any extra step is required for changes to be effective, you should use *checkbox* and corresponding "Apply" button instead of *toggle*. Users shouldn’t do something else or go somewhere else in order to experience the *toggle's* effect.
\ No newline at end of file
+## Accessibility
+The functionality of *toggle* is very similar to a typical [checkbox](https://www.w3.org/TR/wai-aria-practices/#checkbox), but without the third state (partially checked). If any extra step is required for changes to be effective, you should use [checkbox](/components/checkbox/) and corresponding "Apply" button instead of *toggle*. Users shouldn’t do something else or go somewhere else in order to experience the *toggle's* effect.
```
chrisdholt: Is this actually accessibility guidance? It seems more generic than a11y related.

packages/fast-components-react-base/src/divider/README.md
```diff
@@ -1,2 +1,5 @@
 # Divider
-The *Divider* component can be an [hr](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) element.
+The *divider* component defines a thematic break or shift of topic within a page. It is an enhancement of the HTML [hr](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) element.
```
chrisdholt: Is this actually an enhancement of the `<hr />` element? It essentially returns an <hr /> and ensures that we handle a11y correctly. I don't see it as enhancing the element in any meaningful way

packages/fast-components-react-base/src/toggle/README.md
nicholasrice: "partially checked" should be changed to "indeterminate" to use accurate language.

packages/fast-components-react-base/src/checkbox/README.md
```diff
@@ -1,2 +1,5 @@
 # Checkbox
-The *Checkbox* component can be a [checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).
+The [checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox) component is an inout form element used to select or deselect items in a list. The list can be a single item with *checkbox* or multiple items, each with their own *checkbox*. *Checkboxes* have three visual states: checked, unchecked, and indeterminate.
```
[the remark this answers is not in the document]
nalogiudice: Sure we can remove the multiple language

packages/fast-components-react-base/src/divider/README.md
```diff
@@ -1,2 +1,5 @@
...
+
+## Accessibility
+The *divider* is intrinsically set as `role="presentation"`, and therefore is not necessary to explicitly include when usage involves changing the look of the page but does not have all the functional, interactive, or structural relevance implied by the `hr` element. When the *divider* separates and distinguishes sections of content or groups of menu items, apply `role="separator"` attribute.
```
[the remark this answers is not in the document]
nalogiudice: Whoops, well this was wrong in our prior documentation as well :/

packages/fast-components-react-base/src/horizontal-overflow/README.md
nalogiudice: @janechu for follow-up, as this was based on her direction to add this info.

packages/fast-components-react-base/src/image/README.md
nalogiudice: Sure will remove.

packages/fast-components-react-base/src/button/README.md
nalogiudice: Sure makes sense - removed.

packages/fast-components-react-base/src/toggle/README.md
janechu: Also here, do not link to checkbox via this path, it breaks inside the repository.

packages/fast-components-react-base/src/dialog/README.md
nalogiudice: Yep, makes sense - updated.

packages/fast-components-react-base/src/dialog/README.md
nalogiudice: Ok, I basically removed anything that we handle anyway, and just left the links to a11y info... is this sufficient?

packages/fast-components-react-base/src/button/README.md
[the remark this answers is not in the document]
nalogiudice: Ok will remove

packages/fast-components-react-base/src/horizontal-overflow/README.md
janechu: This is based on the native scrolling behavior, if you want items in the horizontal overflow to be tabbed through you have to make it so they can be tabbed to, just as it works when scrolling vertically down a page. It's really just a reminder that since this is based on native scrolling behavior all other rules apply. We could leave it out, as long as we have noted that we are using native scrolling behavior.

chrisdholt: I think we can note that this is implemented using the native scrolling behavior but leave anything regarding items being "tabbable". To me it doesn't add any value regarding accessibility of this component. Whether it's in the context of native scrolling or not, if you want to tab to something, it either needs to be a focusable element, or you need to manage it's tabindex. I think us partially addressing this adds more confusion than not addressing managing focus on the children of the component.

nalogiudice: Sure, ok, I'll just include that it is based on native scrolling and that is all.

Issue #785, opened 2018-08-15
The ISnapshotTestSuite interface contains required properties that are not required for snapshot testing


nicholasrice:
**Is your feature request related to a problem? Please describe.**
The `ISnapshotTestSuite` interface contains required props that are irrelevant to snapshot testing. `schema`, `detailData`, and `documentation` props are requirements of our development-site but not the snapshot testing utilities.

**Describe the solution you'd like**
The requirements of the `ISnapshotTestStuite` interface should only contain properties required for generating snapshots - these include `name`, `component`, and `data` - all other properties should be removed. The documentation site should expand on these props with any necessary additions.

chrisdholt:
closed by #788

PR #788, opened 2018-08-15
fix: remove extraneous requirements for snapshot interface

nicholasrice:
#785

### DIFF
--- a/packages/fast-components-react-base/src/tabs/examples.data.tsx
+++ b/packages/fast-components-react-base/src/tabs/examples.data.tsx
@@ -1,5 +1,4 @@
 import * as React from "react";
-import { ISnapshotTestSuite } from "@microsoft/fast-jest-snapshots-react";
 import {
     IManagedClasses,
     ITabClassNameContract,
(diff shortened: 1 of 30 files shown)


packages/fast-development-site-react/src/utilities/component-factory.tsx
```diff
@@ -2,21 +2,65 @@ import * as React from "react";
...
+    /**
+     * The React component to serve as documentation for the component
+     */
+    documentation: JSX.Element;
+}
+
+/**
+ * An interface to describe the examples object consumed by componentFactor
+ */
+export interface IComponentFactorExamples {
```
chrisdholt: Missing the "y": `IComponentFactoryExamples`

packages/fast-components-react-base/src/tabs/examples.data.tsx
```diff
@@ -1,5 +1,4 @@
 import * as React from "react";
-import { ISnapshotTestSuite } from "@microsoft/fast-jest-snapshots-react";
 import {
     IManagedClasses,
     ITabClassNameContract,
```
janechu: can we remove the `IManagedClasses` and `ITabClassNameContract` here too?

nicholasrice: I removed `IManagedClasses` - `ITabClassNameContract` is being used still.

Issue #782, opened 2018-08-14
Add a callback to HorizontalOverflow and an option to remove the onClick element(s)


janechu:
The implementer should know when `HorizontalOverflow` has been scrolled to the beginning or end of the content it contains. It should also have a passed property to optionally remove the previous/next elements if the elements are at the beginning or end of the content.

Suggested implementation:
- Add additional props `onScrollToEnd`, `onScrollToStart`, `autoRemoveClickElements`
- Add render methods for the previous and next div elements that have the `onClick`
- Add a scroll event listener to the items container, be sure to use requestAnimationFrame
- Check in the scroll listener if the items are at the beginning or the end of the items, if they are and `autoRemoveClickElements` has been passed, do not render the corresponding element eg. items are scrolled to the start, remove the previous div element, if `onScrollToStart` has been passed, execute the callback, and vice versa for moving to the end of the items

PR #797, opened 2018-08-22
feat(fast-components-react-base): add callback to horizontal overflow to return and object that informs scroll start and end

nalogiudice:
Closes [#782](https://github.com/Microsoft/fast-dna/issues/782)
<body>

<footer>
 

### DIFF
--- a/packages/fast-components-react-base/package.json
+++ b/packages/fast-components-react-base/package.json
@@ -26,7 +26,7 @@
     "test": "npm run convert:readme && npm run tslint && npm run unit-tests && npm run build:app",
     "tslint": "tslint -c ./tslint.json 'src/**/*.ts{,x}' '**/*.ts{,x}'",
     "tslint:fix": "tslint -c ./tslint.json --fix 'src/**/*.ts{,x}' '**/*.ts{,x}'",
-    "unit-tests": "jest --maxWorkers=4",
+    "unit-tests": "jest --maxWorkers=4 --env=jsdom --setupTestFrameworkScriptFile=raf/polyfill",
     "unit-tests:update": "jest --updateSnapshot",
     "unit-tests:watch": "jest --watch"
   },
@@ -35,7 +35,7 @@
     "coverageThreshold": {
       "global": {
         "statements": 99,
-        "branches": 94,
+        "branches": 93,
         "functions": 100,
         "lines": 99
       }
@@ -81,6 +81,7 @@
     "jest": "^22.4.2",
     "jsdom-global": "^3.0.2",
     "lodash-es": "^4.0.0",
+    "raf-polyfill": "^1.0.0",
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
     "react-test-renderer": "^16.3.0",
@@ -103,6 +104,7 @@
     "@microsoft/fast-web-utilities": "^1.9.2",
     "exenv-es6": "^1.0.0",
     "lodash-es": "^4.0.0",
+    "raf-polyfill": "^1.0.0",
     "react": "^16.3.0",
     "react-dom": "^16.3.0"
   },
--- a/packages/fast-components-react-base/package.json
+++ b/packages/fast-components-react-base/package.json
@@ -104,7 +104,6 @@
     "@microsoft/fast-web-utilities": "^1.9.2",
     "exenv-es6": "^1.0.0",
     "lodash-es": "^4.0.0",
-    "raf-polyfill": "^1.0.0",
     "react": "^16.3.0",
     "react-dom": "^16.3.0"
   },
--- a/packages/fast-components-react-base/package.json
+++ b/packages/fast-components-react-base/package.json
@@ -34,10 +34,10 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "statements": 98,
-        "branches": 93,
-        "functions": 100,
-        "lines": 98
+        "statements": 80,
+        "branches": 80,
+        "functions": 80,
+        "lines": 80
       }
     },
     "coveragePathIgnorePatterns": [
--- a/packages/fast-components-react-base/package.json
+++ b/packages/fast-components-react-base/package.json
@@ -34,10 +34,10 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "statements": 99,
+        "statements": 98,
         "branches": 93,
         "functions": 100,
-        "lines": 99
+        "lines": 98
       }
     },
     "coveragePathIgnorePatterns": [
--- a/packages/fast-components-react-base/package.json
+++ b/packages/fast-components-react-base/package.json
@@ -105,6 +105,7 @@
     "@microsoft/fast-web-utilities": "^1.9.2",
     "exenv-es6": "^1.0.0",
     "lodash-es": "^4.0.0",
+    "raf-throttle": "^2.0.3",
     "react": "^16.3.0",
     "react-dom": "^16.3.0"
   },
--- a/packages/fast-components-react-base/package.json
+++ b/packages/fast-components-react-base/package.json
@@ -82,6 +82,7 @@
     "jsdom-global": "^3.0.2",
     "lodash-es": "^4.0.0",
     "raf-polyfill": "^1.0.0",
+    "raf-throttle": "^2.0.3",
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
     "react-test-renderer": "^16.3.0",
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -2,7 +2,7 @@
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
 ### Usage guidance
-The *horizontal overflow* component always initializes at the start of the overflow contianer &mdash; left side in LTR and right side in RTL.
+The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL.
 
 ## Accessibility
 The 'next' and 'previous' buttons are only useful for sighted users, since overflow is a purely visual construct, so they should be hidden from screen readers.
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -1,5 +1,8 @@
 # Horizontal overflow
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
+### Usage guidance
+The *horizontal overflow* component always starts at the start of the overflow contianer &mdash; left side in LTR and right side in RTL.
+
 ## Accessibility
 The 'next' and 'previous' buttons are only useful for sighted users, since overflow is a purely visual construct, so they should be hidden from screen readers.
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -2,7 +2,7 @@
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
 ### Usage guidance
-The *horizontal overflow* component always starts at the start of the overflow contianer &mdash; left side in LTR and right side in RTL.
+The *horizontal overflow* component always initializes at the start of the overflow contianer &mdash; left side in LTR and right side in RTL.
 
 ## Accessibility
 The 'next' and 'previous' buttons are only useful for sighted users, since overflow is a purely visual construct, so they should be hidden from screen readers.
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
@@ -1,6 +1,9 @@
 import * as React from "react";
 import { IHorizontalOverflowClassNameContract, IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
 
+export type onScrollToStart = () => void;
+export type onScrollToEnd = () => void;
+
 export interface IHorizontalOverflowHandledProps {
     /**
      * The horizontal overflow content
@@ -11,10 +14,27 @@ export interface IHorizontalOverflowHandledProps {
      * The duration the scroll movement should last
      */
     scrollDuration?: number;
+
+    /**
+     * Callback for on scroll start
+     */
+    onScrollToStart?: onScrollToStart;
+
+    /**
+     * Callback for on scroll end
+     */
+    onScrollToEnd?: onScrollToEnd;
+
+    /**
+     * Option to remove previous/next buttons when scroll reaches beginning or end of container
+     */
+    autoRemoveClickElements?: boolean;
 }
 
 export interface IHorizontalOverflowUnhandledProps extends React.AllHTMLAttributes<HTMLElement> {}
 export interface IHorizontalOverflowManagedClasses extends IManagedClasses<IHorizontalOverflowClassNameContract> {}
 export type HorizontalOverflowProps = IHorizontalOverflowHandledProps
     & IHorizontalOverflowUnhandledProps
-    & IHorizontalOverflowManagedClasses;
+    & IHorizontalOverflowManagedClasses
+    & onScrollToStart
+    & onScrollToEnd;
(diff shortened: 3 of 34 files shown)


nalogiudice:
@chrisdholt I believe code climate issues are non-issues.

packages/fast-components-react-base/src/horizontal-overflow/README.md
```diff
@@ -1,5 +1,8 @@
 # Horizontal overflow
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
+### Usage guidance
+The *horizontal overflow* component always starts at the start of the overflow contianer &mdash; left side in LTR and right side in RTL.
```
nicholasrice: Maybe change the first sentence to "The Horizontal overflow component always *initializes*"

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -76,6 +90,62 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+    /**
+     * Callback on scroll change
+     */
+    public onScrollChange = (): IScrollObject => {
+        let scrollObject: IScrollObject;
+        const distanceRemaining: number =
+            this.horizontalOverflowItemsRef.current.scrollWidth - this.horizontalOverflowItemsRef.current.scrollLeft;
+
+        if (this.props.onScrollChange) {
+            this.props.onScrollChange();
```
nicholasrice: I think you want to pass the `scrollObject` as a parameter to `this.props.onScrollChange` - otherwise an implementer cant know anything about the nature of the scroll change.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -76,6 +90,62 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+        }
+
+        if (this.horizontalOverflowItemsRef.current.scrollLeft === 0) {
+            this.getLTR() === Direction.ltr ?
+            scrollObject = {
```
nicholasrice: I don't think you want to assign different values here based on rtl - I would expect "start" to be the left edge in ltr and the right edge in rtl. This implementation would require me to check for "start" in ltr and "end" in rtl when trying to determine if I'm at the beginning of the overflow.

packages/fast-components-react-base/package.json
```diff
@@ -103,6 +104,7 @@
     "@microsoft/fast-web-utilities": "^1.9.2",
     "exenv-es6": "^1.0.0",
     "lodash-es": "^4.0.0",
+    "raf-polyfill": "^1.0.0",
```
nicholasrice: I don't think we need `raf-polyfill` as a peer dependency - just a dev dependency because our shipped code doesn't actually reference it.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -85,6 +155,18 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+    /**
+     * Request's an animation frame if there are currently no open animation frame requests
+     */
+    private requestFrame = (): void => {
+        if (this.openRequestAnimationFrame) {
+            return;
+        }
+
+        this.openRequestAnimationFrame = true;
+        window.requestAnimationFrame(this.onScrollChange);
```
nicholasrice: You might want to look at https://github.com/wuct/raf-throttle instead of manually setting off the rafThrottle - we can then use that package anywhere we need to use raf instead of implementing that logic in every component.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -105,6 +187,22 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+        return (
+            <div className={this.props.managedClasses.horizontalOverflow_previous} onClick={this.handlePreviousClick}>
+                {this.getChildBySlot(ButtonDirection.previous)}
+            </div>
+        );
+    }
+
+    private renderNextButton(): JSX.Element {
+        return (
+            <div className={this.props.managedClasses.horizontalOverflow_next} onClick={this.handleNextClick}>
```
nicholasrice: I think I would prefer to see these buttons not abstracted to a function - the function doesn't perform any logic and reading the render function now makes me track down what exactly `renderNextButton` does. I think it's more simple to just keep this in the render function.

packages/fast-components-react-base/src/horizontal-overflow/README.md
nalogiudice: Sure.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: This was actually part of another task @janechu for guidance

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nicholasrice: Oops - was reading this wrong. Please disregard.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: Yep, done.

packages/fast-components-react-base/package.json
```diff
@@ -81,6 +81,8 @@
     "jest": "^22.4.2",
     "jsdom-global": "^3.0.2",
     "lodash-es": "^4.0.0",
+    "raf-polyfill": "^1.0.0",
+    "raf-throttle": "^2.0.3",
```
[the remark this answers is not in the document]
nalogiudice: added

packages/fast-components-react-base/src/horizontal-overflow/README.md
```diff
@@ -1,5 +1,8 @@
 # Horizontal overflow
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
+### Usage guidance
+The *horizontal overflow* component always initializes at the start of the overflow contianer &mdash; left side in LTR and right side in RTL.
```
janechu: nit: container, but I would actually change this to content

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -1,6 +1,18 @@
...
+ * Scroll interface for consumers
+ * 'start' is when the horizontal overflow scroll is all the way left in LTR (all the way right in RTL)
+ * 'end' is when the horizontal overflow scroll is all the right in LTR (all the way left in RTL)
+ */
+export interface IScrollChange {
+    start: boolean;
+    end: boolean;
+}
+
+export type onScrollChange = (scrollObject: IScrollChange) => void;
```
janechu: I believe our naming convention means this should be pascal case.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -76,6 +90,43 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+
+        if (this.horizontalOverflowItemsRef.current.scrollLeft === 0) {
+            scrollObject = { start: isLtr, end: !isLtr };
+        } else if (distanceRemaining === this.horizontalOverflowItemsRef.current.clientWidth) {
+            scrollObject = { start: !isLtr, end: isLtr };
+        } else {
+            scrollObject = { start: false, end: false };
+        }
+
+        if (this.props.onScrollChange) {
```
janechu: This should be a check at the beginning, if the onScrollChange is not present just return.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -76,6 +90,43 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+    public componentWillUnmount(): void {
+        if (canUseDOM() && this.horizontalOverflowItemsRef.current) {
+            this.horizontalOverflowItemsRef.current.removeEventListener("scroll", this.throttled);
+        }
+    }
+
+    /**
+     * Callback on scroll change
+     */
+    public onScrollChange = (): IScrollChange => {
```
janechu: why is this public?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -76,6 +90,43 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+        let scrollObject: IScrollChange;
+        const isLtr: boolean = this.getLTR() === Direction.ltr;
+        const distanceRemaining: number =
+            this.horizontalOverflowItemsRef.current.scrollWidth - this.horizontalOverflowItemsRef.current.scrollLeft;
+
+        if (this.horizontalOverflowItemsRef.current.scrollLeft === 0) {
+            scrollObject = { start: isLtr, end: !isLtr };
```
janechu: you can just execute the props.onScrollChange

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -76,6 +90,43 @@ class HorizontalOverflow extends Foundation<HorizontalOverflowProps,  React.AllH
...
+            scrollObject = { start: !isLtr, end: isLtr };
+        } else {
+            scrollObject = { start: false, end: false };
+        }
+
+        if (this.props.onScrollChange) {
+            this.props.onScrollChange(scrollObject);
+        }
+
+        return scrollObject;
```
janechu: why does this return anything?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: How would we pass them the scroll object with data if it is at the beginning?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: So consumers can access it...?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: The entire meetings conclusion said we would return an object with scroll data to the consumer.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: Not sure I understand based on where this comment is

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
nalogiudice: Do you mean `OnScrollChange`?

packages/fast-components-react-base/src/horizontal-overflow/README.md
nalogiudice: Sure.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: But that is what you are doing with `this.props.onScrollChange`

janechu: The return value at the end of this does nothing.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: Instead of saving this as a variable you can just execute the this.props.onScrollChange({...data})

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: That is what the callback is for.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
janechu: Yep

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nicholasrice: To put it more illustratively - it's not really the React convention to expose public methods for consumers to call. To do that, you would have to use a `ref` to obtain a reference to the component and imperatively call a function. Also, publicly calling this function would do a consumer no good because this is a function that needs to be called when the scroll changes - not arbitrarily by the consumer.

Instead, it should be a private function used by the class itself and consumers should register a callback (`this.props.onScrollChage`) to handle updates.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nicholasrice: Her point is that this shouldn't return anything - consumers get notified via a registered callback. Because we only care about this data if a callback is registered, we can check for the existence of the callback before doing any work. If the callback doesn't exist, we know we don't need to do any work and we can just exit because the point of this function is to call a consumer-registered callback.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: I guess I am confused - I thought the function had to be public to access - I will change to private.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: @janechu are you saying like this?
```
        if (this.horizontalOverflowItemsRef.current.scrollLeft === 0) {
            this.props.onScrollChange({ start: isLtr, end: !isLtr });
        } else if (distanceRemaining === this.horizontalOverflowItemsRef.current.clientWidth) {
            this.props.onScrollChange({ start: !isLtr, end: isLtr });
        } else {
            this.props.onScrollChange({ start: false, end: false });
        }
```

nicholasrice: yep - that should work

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nicholasrice: That's the point - no one should be accessing it.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: @nalogiudice yep

nalogiudice: Ok

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: Got it

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -274,7 +274,7 @@ describe("horizontal overflow", (): void => {
             renderedWithImages.instance()[
                 "onScrollChange"
             ]()
-        ).not.toBe(undefined);
+        ).toBe(undefined);
```
janechu: why is this changed?

nalogiudice:
Follow up task [#823](https://github.com/Microsoft/fast-dna/issues/823)

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -257,18 +263,38 @@ describe("horizontal overflow", (): void => {
...
-    test("should have an `onLoad` method", () => {
+    test("should unmount appropriately by removing event listener", () => {
         const renderedWithImages: any = mount(
             <HorizontalOverflow managedClasses={managedClasses}>
                 {imageSet1}
             </HorizontalOverflow>
         );
 
+        renderedWithImages.unmount();
+        expect(renderedWithImages.removeEventListener).toBe(undefined);
```
janechu: This does not do anything, I suggest you remove this test.

nalogiudice: resolved

Issue #620, opened 2018-06-29
Component status should be opt-in


janechu:
The user should only see the component status if the developer has enabled it.

The root development site component should have a new prop that enables the viewing of the component status.

PR #798, opened 2018-08-22
feat(fast-development-site-react): add component status as opt-in

nalogiudice:
Closes [#620](https://github.com/Microsoft/fast-dna/issues/620)
<body>

<footer>
 


packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -766,12 +767,22 @@ class Site extends React.Component<ISiteProps & IManagedClasses<ISiteManagedClas
...
             <div className={this.props.managedClasses.site_statusBar}>
                 <span className={this.props.managedClasses.site_statusComponentName}>Component: {this.state.componentName}</span>
-                <span className={this.getClassNames()}/>
-                {this.state.componentStatus}
+                {this.renderStatus()}
             </div>
         );
     }
 
+    private renderStatus(): JSX.Element | null {
```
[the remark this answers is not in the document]
nalogiudice: Maybe a silly question, but is this even needed then?

nalogiudice:
@nicholasrice removed incorrect typing

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -52,6 +52,7 @@ export interface ISiteProps {
     activeFramework?: Framework;
     collapsed?: boolean;
     componentBackgroundTransparent?: boolean;
+    showComponentStatus?: boolean;
```
chrisdholt: Technically this will disappear from anyone's site that currently has it implemented. If they don't want that, this is great...if they are using it, they will need to figure out why it's gone. 

@nicholasrice, @janechu should we default this to true and have it operate as "opt-out" to maintain current behavior?

nicholasrice: good call - yea I would default this to true

nalogiudice: Done

PR #800, opened 2018-08-22
fix(fast-form-generator-react): fix form item select to return the correct value based on type

### LINKED ISSUE #799: Form item select always returns strings, even when enum is a number
Form items select currently always returns a string, even when the enum being passed is typeof number. This creates issues when using the form generator with a component that expects an enum number value. The form generator returns a string and because the property doesn't match the component behavior breaks.

chrisdholt:
<body>
closes #799 
<footer>
 

### DIFF
--- a/packages/fast-form-generator-react/src/form/form-item.select.tsx
+++ b/packages/fast-form-generator-react/src/form/form-item.select.tsx
@@ -58,10 +58,16 @@ class FormItemSelect extends React.Component<IFormItemSelectProps & IManagedClas
      * Handles the onChange of the select element
      */
     private handleChange = (event: React.FormEvent<HTMLSelectElement>): void => {
-        if (this.props.propertyType === "number") {
-            this.props.onChange(this.props.dataLocation, parseInt((event.target as HTMLSelectElement).value, 0));
-        } else {
-            this.props.onChange(this.props.dataLocation, (event.target as HTMLSelectElement).value);
+        switch (this.props.propertyType) {
+            case "boolean":
+                this.props.onChange(this.props.dataLocation, (event.target as HTMLSelectElement).value === "true");
+                break;
+            case "number":
+                this.props.onChange(this.props.dataLocation, parseInt((event.target as HTMLSelectElement).value, 0));
+                break;
+            default:
+                this.props.onChange(this.props.dataLocation, (event.target as HTMLSelectElement).value);
+                break;
         }
     }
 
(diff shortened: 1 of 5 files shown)


packages/fast-form-generator-react/src/form/form-item.select.tsx
```diff
@@ -53,7 +58,11 @@ class FormItemSelect extends React.Component<IFormItemSelectProps & IManagedClas
      * Handles the onChange of the select element
      */
     private handleChange = (event: React.FormEvent<HTMLSelectElement>): void => {
-        this.props.onChange(this.props.dataLocation, (event.target as HTMLSelectElement).value);
+        if (this.props.propertyType === "number") {
+            this.props.onChange(this.props.dataLocation, parseInt((event.target as HTMLSelectElement).value, 0));
```
janechu: the base for parseInt should be 10

chrisdholt: I added due to a tslint error. Should we keep it or remove it and add a tslint disable comment?

janechu: If it throws a tslint error we can leave it in, I think it's more preferable to disabling

packages/fast-form-generator-react/src/form/form-item.select.tsx
```diff
@@ -53,7 +58,17 @@ class FormItemSelect extends React.Component<IFormItemSelectProps & IManagedClas
      * Handles the onChange of the select element
      */
     private handleChange = (event: React.FormEvent<HTMLSelectElement>): void => {
-        this.props.onChange(this.props.dataLocation, (event.target as HTMLSelectElement).value);
+        switch (this.props.propertyType) {
+            case "boolean":
+                this.props.onChange(this.props.dataLocation, (event.target as HTMLSelectElement).value === "true");
+                break;
+            case "number":
+                this.props.onChange(this.props.dataLocation, parseInt((event.target as HTMLSelectElement).value, 0));
```
nicholasrice: I think the "number" case is accidently working - according to MDN the radix should be between 2 and 36. I think you're looking for a radix of 10.

chrisdholt: Not exactly. It can be octal or decimal. Do we want to be explicit just to ensure our exact scenario? The default is actually 0 or undefined.

nicholasrice: Yea I seem to remember this function not working unless a radix was passed, however it seems to now. I think this is moot per our convo - we should change this to not expect a new property and just use JSON to stringify and parse the values. That will align us with the JSON schema spec where an enum can be any valid JSON data.

chrisdholt: Agree - the update has been pushed. You probably just need to refresh :)

PR #803, opened 2018-08-22
feat(fast-form-generator-react): add a utility to convert a data location to a schema location

### LINKED ISSUE #802: Add a conversion utility to get a schema location from a data location
To simplify our API, we need a utility that will convert a data location to its corresponding schema location. It should detect the type and add in all appropriate strings to create the correct lodash location.

janechu:
Resolves #802


packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,156 @@ export function getComponentTracker(
...
+    const schemaLocationStrings: string[] = getSchemaLocationStringsFromDataLocationStrings(dataLocationStrings, schema, data);
+
+    return schemaLocationStrings.join(".").replace(squareBracketRegex, "");
+}
+
+/**
+ * Get an array of schema location strings from an array of data location strings
+ */
+export function getSchemaLocationStringsFromDataLocationStrings(dataLocationStrings: string[], schema: any, data: any): string[] {
+    const squareBracketRegex: RegExp = /\[(\d+?)\]/g;
```
nicholasrice: You define `squareBracketRegex` several times - consider moving this outside of individual function scope.

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,156 @@ export function getComponentTracker(
...
+    let schemaLocationStrings: string[] = [];
+    let reconstitutedDataLocation: string = "";
+
+    for (let i: number = 0; i < dataLocationStrings.length; i++) {
+        const partialData: any = getPartialData(reconstitutedDataLocation, data);
```
nicholasrice: For both calls to `getPartialData` you're putting together location segments - it seems you could simplify by doing the work to convert arguments to dot notation inside of the `getPartialData` function instead. This then turns into:

```
const partialData: any = getPartialData(dataLocationStrings.slice(0, index), data);
const partialSchema: any = getPartialData(schemaLocationStrings, schema);
```

You can then remove `reconstitutedDataLocation` all together.

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,156 @@ export function getComponentTracker(
...
+    }
+
+    return normalizedDataLocation.join(".");
+}
+
+/**
+ * Gets the index from a JSON schemas oneOf/anyOf array that validates against the data
+ */
+export function getValidAnyOfOneOfIndex(oneOfAnyOf: string, data: any, schema: any): number {
+    let validIndex: number;
```
nicholasrice: You should use `findIndex` - it will break out of the loop once the index is found:

```
return schema[oneofAnyOf].findIndex((item) => tv4.validate(data, item);
```

You'll also need to validate for a `-1` return value whereever this is called.

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,156 @@ export function getComponentTracker(
     return updatedComponentTracker;
 }
 
+/**
+ * Creates a schema location from a data location
+ */
+export function mapSchemaLocationFromDataLocation(dataLocation: string, data: any, schema: any): string {
+    const squareBracketRegex: RegExp = /\[(\d+?)\]/g;
+    let normalizedDataLocation: string = dataLocation.replace(squareBracketRegex, `.$1`); // convert all [ ] to . notation
```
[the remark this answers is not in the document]
janechu: I want to standardize the input to make sure its always the correct format. For example, correct lodash notation could be a["b"].0.c, I want to make sure this is always a.b[0].c so that I can split using "." and reconstruct the schema location with these items.

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,160 @@ export function getComponentTracker(
...
+            getSchemaOneOfAnyOfLocationStrings(
+                partialSchema,
+                partialData
+            )
+        );
+
+        if (dataLocationStrings[i] !== "") {
+            schemaLocationStrings = schemaLocationStrings.concat(
+                getSchemaLocationStringsFromDataLocationString({
+                    isObject: typeof partialData === "object",
```
nicholasrice: you might want to change isObject to `typeof partialData === "object" && partialData !== null`, or use lodash's `isPlainObject`

`typeof null === "object"` evaluates to true - yay JavaScript!

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,160 @@ export function getComponentTracker(
...
+        schemaLocationStrings.push(`oneOf.${getValidAnyOfOneOfIndex("oneOf", data, schema)}`);
+    }
+
+    return schemaLocationStrings;
+}
+
+/**
+ * Get an array of schema location strings from a single data location item
+ */
+export function getSchemaLocationStringsFromDataLocationString(config: ISchemaLocationStringsFromDataLocationStringConfig): string[] {
```
nicholasrice: `getSchemaLocationStringsFromDataLocationString` doesn't seem like an accurate name - I'd expect it to accept a string as an argument but it doesn't

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -1,5 +1,14 @@
 import { clone, cloneDeep, get, isEqual, mergeWith } from "lodash-es";
-import { IBreadcrumbItemConfig, IBreadcrumbItemsConfig, IComponentItem, IFormProps, IFormState, LocationOnChange } from "./form.props";
+import * as tv4 from "tv4";
+import { IBreadcrumbItemConfig, IBreadcrumbItemsConfig, IComponentItem, IFormProps, IFormState } from "./form.props";
+
+export interface ISchemaLocationStringsFromDataLocationStringConfig {
```
nicholasrice: I'm not really following the purpose of this interface. Did you add it as a product of Code Climate yelling about too many arguments?

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +227,160 @@ export function getComponentTracker(
...
+
+/**
+ * Converts a data location strings array items into bracket notation
+ */
+export function convertArrayItemsToBracketNotation(dataLocation: string, data: any): string {
+    const normalizedDataLocation: string[] = [];
+    const dataLocations: string[] = dataLocation.split(".");
+    let currentDataLocations: string[] = [];
+
+    for (let i: number = 0; i < dataLocations.length; i++) {
```
nicholasrice: You don't really need `currentDataLocations`:

`const currentDataLocation = dataLocations.slice(0, i).join(".");`

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -1,5 +1,23 @@
 import { clone, cloneDeep, get, isEqual, mergeWith } from "lodash-es";
-import { IBreadcrumbItemConfig, IBreadcrumbItemsConfig, IComponentItem, IFormProps, IFormState, LocationOnChange } from "./form.props";
+import * as tv4 from "tv4";
+import { IBreadcrumbItemConfig, IBreadcrumbItemsConfig, IComponentItem, IFormProps, IFormState } from "./form.props";
+
+export enum RegexType {
```
nicholasrice: I think this is more ceremony than is really necessary - I really just meant:
```
const squareBrackets: RegExp = /\[(\d+?)\]/g;
const oneOfAnyOf: RegExp = /(oneOf|anyOf)\[\d+\]/g;
```

I'd prefer that approach because it doesn't require us to semantically name and export both an enum and a function to retrieve from that enum

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +235,156 @@ export function getComponentTracker(
...
+/**
+ * Creates a schema location from a data location
+ */
+export function mapSchemaLocationFromDataLocation(dataLocation: string, data: any, schema: any): string {
+    const squareBracketRegex: RegExp = getRegex(RegexType.squareBracket);
+    let normalizedDataLocation: string = dataLocation.replace(squareBracketRegex, `.$1`); // convert all [ ] to . notation
+    normalizedDataLocation = convertArrayItemsToBracketNotation(dataLocation, data); // convert back all array items to use [ ]
+    const dataLocationSegments: string[] = normalizedDataLocation.split(".");
+
+    if (dataLocation === "") {
```
nicholasrice: It seems like you should perform this check prior to any other work because you know as soon as the function executes if you should exit

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -218,6 +221,160 @@ export function getComponentTracker(
...
+    if (typeof data === "object" && data !== null && !Array.isArray(data)) {
+        schemaLocationSegments.push("properties");
+    }
+
+    schemaLocationSegments.push(dataLocation);
+
+    // In the case that this is an array and this is not the the last item in the data location string,
+    // add the JSON schema "items" keyword
+    if ((Array.isArray(data) || isDataLocationArrayItem(dataLocation)) && !lastDataLocationSegment) {
+        schemaLocationSegments.push("items");
```
chrisdholt: assuming because it's a "keyword" per your comment that "items" doesn't need to be localized.

PR #805, opened 2018-08-23
feat(paragraph): adds paragraph as a new MSFT component

### LINKED ISSUE #804: Add paragraph as a new MSFT component
**Is your feature request related to a problem? Please describe.**
In the MSFT design language, paragraph makes up a specific subset of the typography system. Currently a user would have to configure the `<Typography />` component each time or create their own instance of the MSFT paragraph.

**Describe the solution you'd like**
For convenience, it seems like it would be helpful to offer this as component, as this is an established "pattern" for MSFT - similar to what has been done with `<Heading />`.

chrisdholt:
<body>
closes #804 
<footer>
 

### DIFF
--- a/packages/fast-components-react-msft/src/paragraph/index.ts
+++ b/packages/fast-components-react-msft/src/paragraph/index.ts
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { IFoundationProps } from "@microsoft/fast-components-react-base";
-import { IHeadingClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { IParagraphClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 import Paragraph, {
     IParagraphHandledProps,
     IParagraphManagedClasses,
--- a/packages/fast-components-react-msft/src/paragraph/index.ts
+++ b/packages/fast-components-react-msft/src/paragraph/index.ts
@@ -1,0 +1,14 @@
+import * as React from "react";
+import { IFoundationProps } from "@microsoft/fast-components-react-base";
+import { IHeadingClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import Paragraph, {
+    IParagraphHandledProps,
+    IParagraphManagedClasses,
+    IParagraphUnhandledProps,
+    ParagraphLevel
+} from "./paragraph";
+import manageJss, { IJSSManagerProps } from "@microsoft/fast-jss-manager-react";
+import { IDesignSystem, ParagraphStyles } from "@microsoft/fast-components-styles-msft";
+
+export default manageJss(ParagraphStyles)(Paragraph);
+export * from "./paragraph";
--- a/packages/fast-components-react-msft/src/paragraph/paragraph.tsx
+++ b/packages/fast-components-react-msft/src/paragraph/paragraph.tsx
@@ -1,0 +1,64 @@
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import { get } from "lodash-es";
+import { Foundation, HandledProps, TypeLevel, TypographyTag } from "@microsoft/fast-components-react-base";
+import {
+    IParagraphHandledProps,
+    IParagraphManagedClasses,
+    IParagraphUnhandledProps,
+    ParagraphLevel
+} from "./paragraph.props";
+import Typography from "../typography";
+import { IManagedClasses, IParagraphClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+
+/* tslint:disable-next-line */
+class Paragraph extends Foundation<IParagraphHandledProps & IManagedClasses<IParagraphClassNameContract>, React.HTMLAttributes<HTMLParagraphElement>, {}> {
+    protected handledProps: HandledProps<IParagraphHandledProps & IManagedClasses<IParagraphClassNameContract>> = {
+        level: void 0,
+        managedClasses: void 0
+    };
+
+    /**
+     * Renders the component
+     */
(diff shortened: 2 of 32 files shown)


packages/fast-components-react-msft/src/paragraph/index.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+import { IFoundationProps } from "@microsoft/fast-components-react-base";
+import { IHeadingClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
```
janechu: do we need to import this?

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
```diff
@@ -0,0 +1,64 @@
...
+    protected generateClassNames(): string {
+        const classes: string = get(this.props, `managedClasses.paragraph_${this.props.level || ParagraphLevel._3}`);
+
+        return super.generateClassNames(classes);
+    }
+
+    /**
+     * Stores level for use in render
+     */
+    private getTypeLevelFromProps(): TypeLevel {
```
nicholasrice: naming nit: I don't think that either "get" and "fromProps" add any semantic value to this function name. It can be reduced to `typeLevel` and it still convey all necessary meaning.

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
```diff
@@ -0,0 +1,64 @@
...
+        switch (this.props.level) {
+            case ParagraphLevel._1:
+                return TypeLevel._5;
+            case ParagraphLevel._2:
+                return TypeLevel._6;
```
nicholasrice: you can remove `case ParagraphLevel._3` here - `default` handles it

packages/fast-components-react-msft/src/paragraph/index.ts
chrisdholt: This was just incorrectly copy pasta'd.

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
```diff
@@ -0,0 +1,64 @@
...
+                {this.props.children}
+            </Typography>
+        );
+    }
+
+    /**
+     * Generates class names based on props
+     */
+    protected generateClassNames(): string {
+        const classes: string = get(this.props, `managedClasses.paragraph_${this.props.level || ParagraphLevel._3}`);
```
nicholasrice: Is there a reason `ParagraphLevel._3` isn't just a default prop? It appears that you could assign a default, just use `this.props.level`

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
chrisdholt: updated

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
chrisdholt: updated

packages/fast-components-styles-msft/src/paragraph/index.ts
```diff
@@ -0,0 +1,18 @@
+import { IDesignSystem } from "../design-system";
+import { ComponentStyles, ICSSRules } from "@microsoft/fast-jss-manager";
+import { IParagraphClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { fontWeight } from "../utilities/fonts";
+
+const styles: ComponentStyles<IParagraphClassNameContract, IDesignSystem> = {
+    paragraph_1: {
+        fontWeight: `${fontWeight.semilight}`
```
janechu: nit: not necessary to use backticks for these

chrisdholt: these are numbers, so I need the backticks for strings.

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
```diff
@@ -0,0 +1,68 @@
...
+    IParagraphManagedClasses,
+    IParagraphUnhandledProps,
+    ParagraphLevel
+} from "./paragraph.props";
+import Typography from "../typography";
+import { IManagedClasses, IParagraphClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+
+/* tslint:disable-next-line */
+class Paragraph extends Foundation<IParagraphHandledProps & IManagedClasses<IParagraphClassNameContract>, React.HTMLAttributes<HTMLParagraphElement>, {}> {
+    public static defaultProps: Partial<IParagraphHandledProps> = {
```
janechu: is this still necessary now that the switch covers a default addition of 3?

chrisdholt: Yes for the className.

packages/fast-components-react-msft/src/paragraph/paragraph.tsx
```diff
@@ -0,0 +1,68 @@
...
+        level: ParagraphLevel._3
+    };
+
+    protected handledProps: HandledProps<IParagraphHandledProps & IManagedClasses<IParagraphClassNameContract>> = {
+        level: void 0,
```
janechu: add children

chrisdholt: Let's sync with @nicholasrice on this as we are inconsistent on this and we should land and file an issue if needed to align things.

chrisdholt: Synced with @nicholasrice, we'll add children in foundation

PR #810, opened 2018-08-24
feat: add contrast based color system

### LINKED ISSUE #809: feat: introduce contrast-based color system
Introduce color system outlined by #738

nicholasrice:
closes #809


packages/fast-components-styles-msft/src/checkbox/index.ts
```diff
@@ -60,47 +61,60 @@ const styles: ComponentStyles<ICheckboxClassNameContract, IDesignSystem> = (conf
...
-                width: toPx(2)
-            },
+                width: "2px"
+           },
             "&::before": {
-                top: toPx(4),
-                left: toPx(11),
-                height: toPx(12),
+                top: "4px",
+                left: "10px",
```
janechu: this was 11, do we want to change it to 10?

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -1,17 +1,61 @@
...
+
+    /**
+     * The value typically used for backgrounds of elements
+     */
     backgroundColor: string;
+
+    /**
+     * The brand color. This value is deprecated in lieu of accentColor
+     * @deprecated
+     *
```
janechu: nit extra commented line

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -1,17 +1,61 @@
...
+     */
     brandColor: string;
+
+    /**
+     * The primary direction of the document
```
janechu: do we want to say document? because you could wrap anything with the design system

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -1,17 +1,61 @@
...
+ * Ensure that all properties of the design system are assigned
+ */
+export const safeDesignSystem: (config: Partial<IDesignSystem>) => IDesignSystem =
+    memoize(
+        (config: Partial<IDesignSystem>): IDesignSystem => {
+            return Object.assign({}, designSystemDefaults, config);
+        }
+    );
+
+// export { memoizedSafeDesignSystem as safeDesignSystem };
```
janechu: still needed?

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -1,43 +1,41 @@
...
     textField: {
         ...applyType("t7", "vp1"),
-        color: applyTextFieldMixedColor(0.46),
+        color: ensureForegroundNormal,
+        background: ensuresBackgroundNormal,
         fontWeight: fontWeight.light.toString(),
-        outline: `${toPx(1)} solid transparent`,
-        border: `${toPx(1)} solid transparent`,
-        borderColor: applyTextFieldMixedColor(0.46),
+        border: `1px solid transparent`,
```
janechu: nit: doesnt need backticks

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -50,23 +51,23 @@ const styles: ComponentStyles<IToggleClassNameContract, IDesignSystem> = (config
...
         toggle_input: {
             position: "relative",
             margin: "0",
-            width: toPx(44),
-            height: toPx(20),
+            width: "44px",
+            height: "20px",
             background: backgroundColor,
-            border: `${toPx(1)} solid`,
+            border: `1px solid`,
```
janechu: nit: backticks unnecessary

packages/fast-jss-utilities/package.json
```diff
@@ -54,7 +59,9 @@
     "typescript": "^3.0.1"
   },
   "dependencies": {
-    "@microsoft/fast-jss-manager": "^2.0.0"
+    "@microsoft/fast-colors": "^2.0.0",
+    "@microsoft/fast-jss-manager": "^2.0.0",
+    "lodash-es": "^4.17.10"
```
janechu: should any of these be peer?

packages/fast-jss-utilities/src/colors.ts
```diff
@@ -0,0 +1,51 @@
...
+export enum WCAGAAContrastRatios {
+    /**
+     * The contrast ratio for normal size text elements and UI elements less than 3px
+     */
+    normal = 4.5,
+
+    /**
+     * The contrast ratio for large size text elements and UI elements 3px or greater
+     */
+    large = 3
```
janechu: i find this a little unintuitive, because I would expect `WCAGAAContrastRatios.normal` to be a standard ratio and `WCAGAAContrastRatios.large` to be a larger ratio, not relating to the size of the UI elements.

packages/fast-components-react-msft/src/button/__snapshots__/button.spec.tsx.snap
```diff
@@ -2,27 +2,27 @@
 
 exports[`button snapshots Button: 0 1`] = `
 <button
-  className="button-0-1-7"
+  className="button-0-1-8"
   data-sketch-symbol="Button - default"
```
nalogiudice: Is this sketch symbol still correct with this naming change?

packages/fast-components-react-msft/src/button/__snapshots__/button.spec.tsx.snap
```diff
@@ -68,72 +68,72 @@ exports[`button snapshots Button: 5 1`] = `
 
 exports[`button snapshots Button: 6 1`] = `
 <button
-  className="button-0-1-79"
+  className="button-0-1-92 button__disabled-0-1-98"
   data-sketch-symbol="Button - default disabled"
```
nalogiudice: Same here, is this sketch symbol still correct with this naming change?

packages/fast-colors/src/contrast.spec.ts
```diff
@@ -18,3 +18,57 @@ describe("contrast", (): void => {
...
+    test("should not throw", (): void => {
+        expect(() => {
+            ensureContrastCast();
+        }).not.toThrow();
+    });
+    test("should return the opperand value if arguments are not valid", (): void => {
+        expect(ensureContrastCast()).toBe(undefined);
+        expect(ensureContrastCast(1)).toBe(undefined);
+        expect(ensureContrastCast(1, "red")).toBe("red");
+        expect(ensureContrastCast(1, 1)).toBe(1);
```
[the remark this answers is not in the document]
nicholasrice: Yep. The reason is that I'm not passing the correct parameter count and type.

packages/fast-jss-utilities/src/colors.ts
nicholasrice: Do you have a suggestion to add clarity? This makes sense to me because a larger element would require a lower contrast ratio because the element is physically more prominent and noticeable.

janechu: maybe `default` or `normal` and `low`?

packages/fast-colors/src/contrast.ts
```diff
@@ -1,20 +1,25 @@
...
+/**
+ * A function to manipulate contrast of two colors.
+ */
+export type ContrastFunction = (targetRatio: number, operandColor: string, referenceColor: string) => string;
+
 /**
  * Adjust the darkness/lightness of a foreground color so that it matches a target contrast ratio against a background color
- * @param targetRatio - The desired contrast ratio to bring the operand to against the reference
+ * @param targetRatio - The desired contrast ratio to bring the operand to against the referenceColor
  * @param operand - The color value to manipulate
```
janechu: *operandColor

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -43,33 +70,111 @@ function applyTransaprentBackground(): ICSSRules<IDesignSystem> {
...
+    );
+    const outlineDisabledBorderColor: string = outlineDisabledColor;
 
     return {
         button: {
             ...applyType("t7", "vp1"),
             boxSizing: "border-box",
-            maxWidth: toPx(374),
-            minWidth: toPx(120),
+            maxWidth: "374px",
```
awentzel: Why is did you remove toPx?  Just curiousl

packages/fast-jss-utilities/src/colors.ts
awentzel: Who is to say what is 'normal' or 'low' as this may vary depending on many accessibility factors.  I would consider sticking with sizes.  'small' 'medium' 'large' and we can extend a few others later if needed. 'extra-large' 'extra-small' for example.

awentzel:
If this is new, does this have code coverage hooked up already?  If not, we should do this before the task is completed.

packages/fast-jss-utilities/src/colors.ts
nicholasrice: Leaving as is but clarifying the enum name.

packages/fast-components-styles-msft/src/button/index.ts
nicholasrice: Its an unnecessary operation that needs to happen at runtime. `toPx` is useful when number manipulation is needed, but if we know the value already there is no reason I know of to take the runtime hit - just define it as a string.

nicholasrice:
@awentzel code coverage exists in the packages that this touches.

PR #814, opened 2018-08-26
chore: update fast-animation's TypeScript

### LINKED ISSUE #796: chore: update fast-animation typescript
TypeScript 3.0 adds DOM typing's for many of the Web Animations API objects which is causing redundant type errors when migrating to 3.0. Some of the other typings seem to not exist, so there will be some refactor work required to update TypeScript.

nicholasrice:
closes #796

This PR implement's partial interfaces for the Keyframe interface - this is due to an outstanding issue in TypeScript.

More details here: #815

### DIFF
--- a/packages/fast-animation/lib/animateGroup/index.ts
+++ b/packages/fast-animation/lib/animateGroup/index.ts
@@ -80,8 +80,8 @@ class AnimateGroup {
     private getAnimationDuration(effectTiming: EffectTiming): number {
         const duration: string | number = effectTiming.duration;
         const sanitizedDuration: number = typeof duration === "string"
-        ? parseFloat(duration)
-        : duration;
+            ? parseFloat(duration)
+            : duration;
 
         return (effectTiming.delay || 0) + (sanitizedDuration || 0);
     }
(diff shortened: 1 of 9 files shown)


packages/fast-animation/lib/animateGroup/index.ts
```diff
@@ -77,8 +77,13 @@ class AnimateGroup {
     /**
      * Returns the cumulative time it will take to complete an animation
      */
-    private getAnimationDuration(effectTiming: AnimationEffectTiming): number {
-        return (effectTiming.delay || 0) + (effectTiming.duration || 0);
+    private getAnimationDuration(effectTiming: EffectTiming): number {
+        const duration: string | number = effectTiming.duration;
```
janechu: isn't this always a number because the enum is always a number for duration?

nicholasrice: Internally we use `number` so we can do maths, however TypeScript defines `EffectTiming.duration` as either a string or a number, so we need to ensure it's a number and parse it if it isn't.

PR #818, opened 2018-08-27
feat(metatext): add metatext as a new MSFT component

### LINKED ISSUE #817: Add metatext as a new component in MSFT package
In the MSFT design language, metatext is a consistent pattern for handling additional information or context regarding other text. 

Story:
```
Metatext is lightweight text that describes other, more prominent text or components nearby. For example, metatext could provide more information about an app title such as the price or publisher.
```

Please create `metatext` as part of the MSFT package; built from `<Typography />` component

chrisdholt:
<body>
closes #817
<footer>
 


packages/fast-components-react-msft/src/metatext/examples.data.tsx
```diff
@@ -0,0 +1,38 @@
...
+import Metatext from "./index";
+import {
+    IMetatextHandledProps,
+    IMetatextManagedClasses,
+    IMetatextUnhandledProps,
+    MetatextProps,
+    MetatextTag
+} from "./metatext.props";
+import schema from "./metatext.schema.json";
+import Documentation from "./.tmp/documentation";
```
awentzel: Can you explain what's happening here?  Why is documentation in a `.tmp` folder?

packages/fast-components-react-msft/src/metatext/metatext.tsx
```diff
@@ -0,0 +1,53 @@
...
+class Metatext extends Foundation<IMetatextHandledProps & IManagedClasses<IMetatextClassNameContract>, React.HTMLAttributes<HTMLSpanElement | HTMLParagraphElement>, {}> {
+    protected handledProps: HandledProps<IMetatextHandledProps & IManagedClasses<IMetatextClassNameContract>> = {
+        managedClasses: void 0,
+        tag: void 0
+    };
+
+    /**
+     * Stores HTML tag for use in render
+     */
+    private get tag(): TypographyTag {
```
nicholasrice: Any reason you're not relying on `defaultProps` here? You could then just pass `this.props.tag` directly to `Typography` and avoid the getter.

packages/fast-components-react-msft/src/metatext/metatext.props.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface IMetatextHandledProps {
+    /**
+     * The metatext content
+     */
+    children?: React.ReactNode | React.ReactNode[];
+
+    /**
+     * The metatext tag
+     */
+    tag: MetatextTag;
```
nicholasrice: this would be optional if you utilized defaultProps

packages/fast-components-react-msft/src/metatext/metatext.tsx
chrisdholt: Established pattern we have had throughout all other components when providing a tag. We're using both a getter and default props in most places. Happy to update, but would ask that we create an issue to align all of them.

packages/fast-components-react-msft/src/metatext/metatext.props.ts
chrisdholt: This should be optional anyway due to the current getter setup. This is just a mistake.

packages/fast-components-react-msft/src/metatext/metatext.tsx
nicholasrice: It's only a necessary pattern in certain circumstances:

The Typography component uses that pattern because we need to be *sure* that the tag passed as props is valid. Because of that, we need to use a function to return the prop. TypeScript does not allow us to use a function's return value as a JSX tag though; `<this.tag()>` is invalid. The workaround is to use a getter. You're just passing a prop *down* to `Typography` though, not creating an HTML element, so it's a different case.

My main point is that you're applying a default value while the props interface explicitly requires the value. If the component has a default, then it shouldn't require a value.

chrisdholt: Not saying it's *necessary*, saying it's how we've managed this with all other typography-derived components. I'll push an update.

chrisdholt: Props was also a mistake, but the above explanation is accurate. We should update elsewhere, but that can be done in another PR.

packages/fast-components-react-msft/src/metatext/examples.data.tsx
awentzel: I think any generated files that are used for deployment should go in a dist or www folder if one of those is already present.  my preference is `dist` since it's more comment.  Also, prefer to have `www` inside that if there is an accompanying website.  But, this can be a future change and has no bearing on this PR.

awentzel:
Rebased and pushed on behalf of Chris since he's currently on vacation.

PR #824, opened 2018-08-28
refactor(fast-form-generator-react): update the navigation system to only use data locations

### LINKED ISSUE #807: Refactor the component tracker
The navigation should be reliable when navigating via breadcrumbs.

Technical requirements:
- Use one object per location item - this is to ensure that every breadcrumb corresponds to a tracked item
- Only use data locations - this should make use of the new utility to derive a schema location from a data location
- Each schema associated with a tracked item should be able to be mapped to a form section

janechu:
Closes #807


packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -232,8 +278,9 @@ export function normalizeDataLocation(dataLocation: string, data: any): string {
 /**
  * Removes any references to array index
  */
-export function normalizeSchemaLocation(schemaLocation: string): string {
-    return schemaLocation.replace(squareBracketsRegex, "");
+export function normalizeSchemaLocation(schemaLocation: string, lastLocationItem?: boolean): string {
+    const schemaLocationWithoutArrayLocation: string = schemaLocation.replace(squareBracketsRegex, "");
+    return lastLocationItem ? schemaLocationWithoutArrayLocation.replace(arrayItemsRegex, "") : schemaLocationWithoutArrayLocation;
```
nicholasrice: I don't think you should add the `lastLocationItem` param here - it makes the function multi-purpose and kind of defeats the point of "normalizing" when you can get multiple outputs. It also makes harder to understand - reading the code, I now need  to go figure out what passing `true` does because it's not inherently descriptive.

I think you should preform the replacement contextually where you need it - if it crops up more than once I would *then* create a new abstraction instead of complicating this one.

Issue #825, opened 2018-08-29
fix: fast-animation should export commonjs


nicholasrice:
A recent change to fast-animation changed the module export to ES6 - this could have a breaking impact on consumers so we should target this change for a major release.

PR #826, opened 2018-08-29
fix: reverting change to export es6 from fast-animation

nicholasrice:
#825

### DIFF
--- a/packages/fast-animation/lib/tsconfig.json
+++ b/packages/fast-animation/lib/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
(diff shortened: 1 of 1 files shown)


packages/fast-animation/lib/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "module": "ES6",
+    "module": "CommonJS",
     "target": "ES6",
```
janechu: shouldnt the target be es5?

nicholasrice: good catch - changed.

PR #832, opened 2018-08-30
chore(fast-components-react-base): update an enzyme test to use mount instead of shallow

awentzel:
**Problem:**
Shallow was causing inconsistent build breaks when running `lerna run test`.  

**Solution:**
update to use mount not shallow for enzyme as a result of inconsistent build breaks with shallow


janechu:
Since this is only an issue in one location can you change the wording of this to something more along the lines of `chore: update an enzyme test to use mount instead of shallow`.

Issue #833, opened 2018-08-30
Create "Progress" component in Base and MSFT


eljefe223:
Progress component is needed in base and MSFT.

Suggested API would be to expose the following props:

value: the value of the progress
minValue: this minimum value
maxValue: the maxValue
should accept children

Suggested functionality:
Indeterminate and determinate states can be determined if the vale props is passed. Roles can be used to display the correct children in either case.

PR #835, opened 2018-08-30
feat(Progress): add progress component

eljefe223:
closes: #833


packages/fast-components-react-msft/src/progress/progress.tsx
```diff
@@ -0,0 +1,80 @@
...
+                value={this.props.value}
+                minValue={this.props.minValue}
+                maxValue={this.props.maxValue}
+            >
+                <div
+                    className={this.props.managedClasses.progress_indicator__determinate}
+                    slot={ProgressType.determinate}
+                >
+                    <div
+                        className={this.props.managedClasses.progress_indicator__determinate_bar}
```
nicholasrice: because `determinate_bar` is referring to an element and not a state, this classname should be `progress_determinateBar`. To be more semantic though, you might name it `progess_determinateValueIndicator`

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    return {
+        progress: {
+            display: "flex",
+            width: "100%",
+            alignItems: "center",
+            height: "10px"
+        },
+        progress_indicator__determinate: {
+            display: "flex",
```
nicholasrice: You've got some common styles between `progress_indicator__determinate` and `progress_indicator__indeterminate` - I would add a base class (`progress_indicator`) that gets applied to both containers that apply the common styles.

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -0,0 +1,109 @@
...
+                animationTimingFunction: "ease-out"
+            },
+            "75%": {
+                opacity: "1",
+            },
+            "80%": {
+                transform: "translateX(100%)",
+                opacity: "0",
+            },
+            "100%": {
```
nicholasrice: nit: `"100%": {}`

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -0,0 +1,109 @@
+import designSystemDefaults, { IDesignSystem } from "../design-system";
+import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
+import { applyLocalizedProperty, Direction, toPx } from "@microsoft/fast-jss-utilities";
+import { IMSFTProgressClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import Chroma from "chroma-js";
+import { get } from "lodash";
+
+/* tslint:disable-next-line */
+const styles: ComponentStyles<IMSFTProgressClassNameContract, IDesignSystem> = (config: IDesignSystem): ComponentStyleSheet<IMSFTProgressClassNameContract, IDesignSystem> => {
+    const backgroundColor: string = get(config, "backgroundColor") || designSystemDefaults.backgroundColor;
```
nicholasrice: Can you give using the contrast helpers a shot? Ping me if you have questions on them.

packages/fast-components-react-msft/src/progress/progress.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    slot={ProgressType.indeterminate}
+                    className={this.props.managedClasses.progress_indicator__indeterminate}
+                >
+                    {this.renderIndeterminateItems(this.generateIndeterminateItems())}
+                </div>
+            </BaseProgress>
+        );
+    }
+
+    public generateIndeterminateItems(): ISpanData[] {
```
nicholasrice: It seems a little heavy-handed to me to define objects and an interface for dot-data when you're really just iterating. I think you could simplify this using the following:

```
private static indicatorDotCount: number = 5
// ...
private renderIndeterminateItems(): JSX.Element[] {
        return new Array(Progress.indicatorDotCount).fill(undefined).map((item: undefined, index: number) => {
            let className: string = this.props.managedClasses.progress_dot;
            className = `${className} ${this.props.managedClasses[`progress_dot__${index + 1}`]}`;
            return (
                <span
                    className={className}
                    key={index}
                />
            );
        });
    }
```

packages/fast-components-react-base/src/progress/README.md
```diff
@@ -0,0 +1,5 @@
+# Progress
+The *Progress* component should be used for an element that displays the progress status for a task that take a long time or consists of several steps. Typical use cases include loading elements or status of a download. The *Progress* component is an [div](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div) element with it `role` set to "progressbar".
```
nalogiudice: please lowercase component names

nalogiudice: "..for a task that takes a long time..."

packages/fast-components-react-base/src/progress/progress.tsx
```diff
@@ -0,0 +1,101 @@
...
+     * renders children based on value prop
+     */
+    private renderChildren(): JSX.Element[] {
+        return this.props.value !== undefined
+            ? this.renderChildElements(ProgressType.determinate)
+            : this.renderChildElements(ProgressType.indeterminate);
+    }
+
+    /**
+     * renders children by slot name
```
nalogiudice: nit: Capitalize start of code comment sentences

packages/fast-components-react-msft/src/progress/README.md
```diff
@@ -0,0 +1,8 @@
+# Progress
+The *Progress* component is an [div](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div) element with it `role` set to "progressbar". Setting the `value` prop will automatically display child elements with their `role` set to "deteminate" if `value` is undefined child elements with their `role` set to "indeteminate" will be displayed.
```
nalogiudice: Same comment about lowercase component names.

packages/fast-components-react-base/src/progress/README.md
eljefe223: I was using Button's read me as guidance. We should file a bug to fix the capitalization there.

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -0,0 +1,105 @@
+import designSystemDefaults, { IDesignSystem, withDesignSystemDefaults } from "../design-system";
+import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
+import { IMSFTProgressClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { ensureBrandNormal, largeContrast } from "../utilities/colors";
+import Chroma from "chroma-js";
```
janechu: nit: chroma doesnt get used

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -0,0 +1,105 @@
...
+    return {
+        progress: {
+            display: "flex",
+            width: "100%",
+            alignItems: "center",
+            height: "10px",
+            textAlign: "left"
+        },
+        progress_determinateValueIndicator: {
+            background: brandColor,
```
janechu: @nicholasrice do you think we should use the contrast ratios here to ensure the brand color is meeting contrast ration with the determinate background color?

nicholasrice: I don't think so. If you look at the component, the accent color (value indicator) is actually on top of the page background. It appears to be filling the bar created by the grey, however it is only on the right edge that the indicator is against the grey - all 3 other edges are on the page background.

PR #838, opened 2018-08-31
chore(toggle): update to coding standards

### LINKED ISSUE #837: Ensure toggle is up to coding standards
Ensure Toggle is implemented correctly and that source is up to coding standards. This component should be able to be used as a reference for other components.

nicholasrice:
closes #837


packages/fast-components-react-base/src/toggle/toggle.spec.tsx
```diff
@@ -38,68 +38,80 @@ describe("toggle", (): void => {
...
         );
 
-        const state: any = rendered.state("checked");
-
-        expect(state).toBe(undefined);
+        expect(rendered.state("selected")).toBe(false);
     });
 
-    test("should correctly operate as an uncontrolled component and handle `onChange` events when `selected` prop is not passed", () => {
+    test("a change event should change the selected state when no `selected` prop is not provided", () => {
```
janechu: nit: can we re-word to read more like "should allow a change event to update the selected state when no `selected` prop is provided" so that our tests can read more like a sentence when starting with the test suites describe string "toggle". This is also a double negative so was a little confusing to read.

packages/fast-components-react-base/src/toggle/toggle.spec.tsx
```diff
@@ -38,68 +38,80 @@ describe("toggle", (): void => {
...
-        const input: any = rendered.find(".toggle-input-class");
-        input.prop("onChange")(); // we have to fire the internal onChange from the input as we aren't passing one
+        rendered.find(inputSelector).simulate("change");
 
-        expect(rendered.state("checked")).toBe(true);
+        expect(rendered.state("selected")).toBe(true);
     });
 
-    test("should correctly operate as a controlled component when `onChange` and `selected` prop is passed", () => {
+    test("should call a regestired callback after a change event", () => {
```
janechu: nit: *registered

packages/fast-components-react-base/src/toggle/toggle.spec.tsx
```diff
@@ -38,68 +38,80 @@ describe("toggle", (): void => {
...
-        expect(onChange).toHaveBeenCalled();
+        controlled
+            .find(inputSelector)
+            .simulate("change");
+
+        expect(onChange).toHaveBeenCalledTimes(2);
     });
 
-    test("should correctly call getDerivedStateFromProps if an updated `selected` prop is passed which differs from the state", () => {
+    test("setting props.selected should prevent Toggle from updating its own selected state", () => {
```
janechu: nit: can we re-word to read "should not update the selected state during a change event when `selected` prop has been provided"

packages/fast-components-react-base/src/toggle/toggle.tsx
```diff
@@ -69,12 +69,11 @@ class Toggle extends Foundation<IToggleHandledProps & IManagedClasses<IToggleCla
                         className={get(this.props, "managedClasses.toggle_input")}
                         type="checkbox"
                         id={this.props.id}
-                        defaultChecked={this.state.checked}
                         aria-describedby={this.props.statusLabelId}
                         disabled={this.props.disabled}
                         value={this.generateToggleStateLabel()}
                         onChange={this.handleToggleChange}
-                        checked={!!this.state.checked}
+                        checked={this.state.selected}
```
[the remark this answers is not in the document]
nicholasrice: `defaultChecked` never did anything because `checked` was always set

PR #842, opened 2018-08-31
chore(checkbox): update to coding standards

### LINKED ISSUE #841: Ensure Checkbox is up to coding standards
Ensure Checkbox is implemented correctly and that source is up to coding standards. This component should be able to be used as a reference for other components.

nicholasrice:
closes #841

### DIFF
--- a/packages/fast-components-react-base/src/checkbox/checkbox.tsx
+++ b/packages/fast-components-react-base/src/checkbox/checkbox.tsx
@@ -3,7 +3,7 @@ import * as ReactDOM from "react-dom";
 import Foundation, { HandledProps } from "../foundation";
 import { CheckboxHTMLTags, CheckboxProps, ICheckboxHandledProps, ICheckboxManagedClasses, ICheckboxUnhandledProps } from "./checkbox.props";
 import { ICheckboxClassNameContract, IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -130,7 +130,7 @@ class Checkbox extends Foundation<ICheckboxHandledProps & ICheckboxManagedClasse
      * Handles onChange as a controlled component
      */
     private handleCheckboxChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
-        if (isUndefined(this.props.checked)) {
+        if (typeof this.props.checked !== "boolean") {
```
nalogiudice: Can you explain this change?

janechu: I would say that its one less import the way it's been done here so preferrable

nicholasrice: The expectation is that "checked" is a boolean. By checking for `undefined`, we're not asserting that our data structure meets our expectations. If `checked` were set to `"foobar"` or `Infinity` or `null`, this could would execute under false assumptions and would do unexpected things.

nalogiudice: Got it

Issue #834, opened 2018-08-30
Create call to action component


nalogiudice:
Create call to action component and all relative testing etc...

chrisdholt:
Is this necessary? How does it differ from adding a glyph to `<Button />` via children or className/classNameContract?

PR #851, opened 2018-09-05
feat(call-to-action): add call to action component

nalogiudice:
Closes [#834](https://github.com/microsoft/fast-dna/issues/834)
<body>

<footer>
 

### DIFF
--- a/packages/fast-components-class-name-contracts-msft/src/button/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/button/index.ts
@@ -8,6 +8,6 @@ export interface IMSFTButtonClassNameContract extends IButtonClassNameContract {
     button_outline: string;
     button_lightweight: string;
     button_justified: string;
-    button_textContainer: string;
+    button_span: string;
     button__disabled: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/button/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/button/index.ts
@@ -8,6 +8,6 @@ export interface IMSFTButtonClassNameContract extends IButtonClassNameContract {
     button_outline: string;
     button_lightweight: string;
     button_justified: string;
-    button_span: string;
+    button_textContainer: string;
     button__disabled: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
@@ -1,15 +1,9 @@
-import { IHypertextClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
-
 /**
  * The class name contract for the call to action component
  */
-export interface ICallToActionClassNameContract extends IHypertextClassNameContract {
-    callToAction_primary: string;
-    callToAction_secondary: string;
+export interface ICallToActionClassNameContract {
+    callToAction: string;
+    callToAction_backer: string;
     callToAction_lightweight: string;
-    callToAction_span: string;
-    "@keyframes cta-glyph-move-out": string;
-    "@keyframes cta-glyph-move-in": string;
-    "@keyframes cta-text-move-in": string;
-    "@keyframes cta-text-move-out": string;
+    glyph: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
@@ -1,0 +1,15 @@
+import { IHypertextClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+/**
+ * The class name contract for the call to action component
+ */
+export interface IMSFTCallToActionClassNameContract extends IHypertextClassNameContract {
+    callToAction_primary: string;
+    callToAction_secondary: string;
+    callToAction_lightweight: string;
+    callToAction_span: string;
+    "@keyframes cta-glyph-move-out": string;
+    "@keyframes cta-glyph-move-in": string;
+    "@keyframes cta-text-move-in": string;
+    "@keyframes cta-text-move-out": string;
+}
--- a/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
@@ -2,10 +2,33 @@
  * The class name contract for the call to action component
  */
 export interface ICallToActionClassNameContract {
+    /**
+     * The root of the call to action component
+     */
     callToAction: string;
-    callToAction_primary: string;
-    callToAction_lightweight: string;
-    callToAction_justified: string;
+
+    /**
+     * The call to action glyph
+     */
+    callToAction_glyph: string;
+
+    /**
+     * The primary appearance modifier
+     */
+    callToAction__primary: string;
+
+    /**
+     * The lightweight appearance modifier
+     */
+    callToAction__lightweight: string;
+
+    /**
+     * The justified appearance modifier
+     */
+    callToAction__justified: string;
+
+    /**
+     * The disabled modifier
+     */
     callToAction__disabled: string;
-    glyph: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
@@ -3,7 +3,7 @@ import { IHypertextClassNameContract } from "@microsoft/fast-components-class-na
 /**
  * The class name contract for the call to action component
  */
-export interface IMSFTCallToActionClassNameContract extends IHypertextClassNameContract {
+export interface ICallToActionClassNameContract extends IHypertextClassNameContract {
     callToAction_primary: string;
     callToAction_secondary: string;
     callToAction_lightweight: string;
--- a/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
@@ -3,7 +3,9 @@
  */
 export interface ICallToActionClassNameContract {
     callToAction: string;
-    callToAction_backer: string;
+    callToAction_primary: string;
     callToAction_lightweight: string;
+    callToAction_justified: string;
+    callToAction__disabled: string;
     glyph: string;
 }
--- a/packages/fast-components-react-msft/package.json
+++ b/packages/fast-components-react-msft/package.json
@@ -73,6 +73,7 @@
     "@microsoft/fast-glyphs-msft": "^2.2.0",
     "@microsoft/fast-jest-snapshots-react": "^2.2.0",
     "@microsoft/fast-jss-manager": "^2.2.0",
+    "@microsoft/fast-jss-utilities": "^2.2.0",
     "@microsoft/fast-sketch-library": "^2.2.0",
     "@microsoft/fast-tslint-rules": "^2.2.0",
     "@types/jest": "^22.2.0",
--- a/packages/fast-components-react-msft/package.json
+++ b/packages/fast-components-react-msft/package.json
@@ -73,7 +73,6 @@
     "@microsoft/fast-glyphs-msft": "^2.2.0",
     "@microsoft/fast-jest-snapshots-react": "^2.2.0",
     "@microsoft/fast-jss-manager": "^2.2.0",
-    "@microsoft/fast-jss-utilities": "^2.2.0",
     "@microsoft/fast-sketch-library": "^2.2.0",
     "@microsoft/fast-tslint-rules": "^2.2.0",
     "@types/jest": "^22.2.0",
--- a/packages/fast-components-react-msft/src/button/button.props.ts
+++ b/packages/fast-components-react-msft/src/button/button.props.ts
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { IButtonHandledProps as IBaseButtonHandledProps } from "@microsoft/fast-components-react-base";
 import { IManagedClasses, IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { ButtonSlot } from "./button";
 
 export enum ButtonAppearance {
     justified= "justified",
--- a/packages/fast-components-react-msft/src/button/button.props.ts
+++ b/packages/fast-components-react-msft/src/button/button.props.ts
@@ -1,7 +1,6 @@
 import * as React from "react";
 import { IButtonHandledProps as IBaseButtonHandledProps } from "@microsoft/fast-components-react-base";
 import { IManagedClasses, IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
-import { ButtonSlot } from "./button";
 
 export enum ButtonAppearance {
     justified= "justified",
--- a/packages/fast-components-react-msft/src/button/button.spec.tsx
+++ b/packages/fast-components-react-msft/src/button/button.spec.tsx
@@ -167,7 +167,7 @@ describe("button", (): void => {
     });
 
     /* tslint:disable-next-line */
-    test("should add an 'before' element when a child is passed with the before slot", () => {
+    test("should add a child element with the 'before' prop into the before slot", () => {
         const props: IButtonHandledProps = {
             appearance: ButtonAppearance.lightweight,
             href: "#",
--- a/packages/fast-components-react-msft/src/button/button.spec.tsx
+++ b/packages/fast-components-react-msft/src/button/button.spec.tsx
@@ -167,7 +167,7 @@ describe("button", (): void => {
     });
 
     /* tslint:disable-next-line */
-    test("should add a child element with the 'before' prop into the before slot", () => {
+    test("should add a child element with the slot prop set to 'before' into the before slot location", () => {
         const props: IButtonHandledProps = {
             appearance: ButtonAppearance.lightweight,
             href: "#",
@@ -183,7 +183,7 @@ describe("button", (): void => {
     });
 
     /* tslint:disable-next-line */
-    test("should add an 'after' element when a child is passed with the after slot", () => {
+    test("should add a child element with the slot prop set to 'after' into the after slot location", () => {
         const props: IButtonHandledProps = {
             appearance: ButtonAppearance.lightweight,
             href: "#",
--- a/packages/fast-components-react-msft/src/button/button.spec.tsx
+++ b/packages/fast-components-react-msft/src/button/button.spec.tsx
@@ -34,8 +34,9 @@ describe("button", (): void => {
             className={"slotBefore"}
             key={"Slot"}
             slot="before"
-            dangerouslySetInnerHTML={{__html: "Before slot"}}
-        />
+        >
+            Before slot
+        </div>
     );
 
     const afterSlotExample: JSX.Element = (
@@ -43,8 +44,9 @@ describe("button", (): void => {
             className={"slotAfter"}
             key={"afterSlot"}
             slot="after"
-            dangerouslySetInnerHTML={{__html: " After slot"}}
-        />
+        >
+            After slot
+        </div>
     );
 
     test("should return an object that includes all valid props which are not enumerated as handledProps", () => {
@@ -68,7 +70,7 @@ describe("button", (): void => {
     });
 
     /* tslint:disable-next-line */
-    test("should set a className that matches button_primary managedClass when ButtonAppearance.primary is passed to the appearance prop", () => {
+    test("should apply a 'primary' html class when appearance is primary", () => {
         const props: IButtonHandledProps = {
             appearance: ButtonAppearance.primary
         };
@@ -87,7 +89,7 @@ describe("button", (): void => {
     });
 
     /* tslint:disable-next-line */
-    test("should set a className that matches button_outline managedClass when ButtonAppearance.outline is passed to the appearance prop", () => {
+    test("should apply an 'outline' html class when appearance is outline", () => {
         const props: IButtonHandledProps = {
             appearance: ButtonAppearance.outline
         };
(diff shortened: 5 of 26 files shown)


packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
```diff
@@ -0,0 +1,15 @@
+import { IHypertextClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+/**
+ * The class name contract for the call to action component
+ */
+export interface IMSFTCallToActionClassNameContract extends IHypertextClassNameContract {
```
chrisdholt: There is no base MSFT component, so I don't think we need to preface this with MSFT.

packages/fast-components-react-msft/src/call-to-action/call-to-action.props.ts
```diff
@@ -0,0 +1,30 @@
...
+     */
+    children?: React.ReactNode | React.ReactNode[];
+
+    /**
+     * The destination address
+     */
+    href: string;
+}
+
+export interface ICallToActionUnhandledProps extends React.AllHTMLAttributes<HTMLElement> {}
```
chrisdholt: If using Hypertext, this should be HTMLAnchorElement. If we were using the `<Button />` component this would be accurate.

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
```diff
@@ -0,0 +1,54 @@
...
+        href: void 0,
+        managedClasses: void 0
+    };
+
+    /**
+     * Renders the component
+     */
+    public render(): JSX.Element {
+        return (
+            <Hypertext
```
chrisdholt: Can you elaborate on the reason you're returning Hypertext here? CTA could be a button...we've definitely seen that before and I believe it's a big reason that we created the button component in the way we did. It seems like we could reduce complexity and code simply by using the MSFT button. There is way more alignment with `<Button />` than there is with `<Hypertext />`.

awentzel: This is a good point, might suggest raising this in today's standup.

packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
nalogiudice: Sure, I suppose so.

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
nalogiudice: Seems like a disconnect. I had a conversation with @janechu and @nicholasrice and this was the conclusion. To extend hypertext in MSFT only, as CTA has always been an anchor (based off Fluent). We do not have a glyph system to use yet, so these are base64 encoded images. The animations add classNames and JSS. Semantically speaking, having a CTA component for usage in Microsoft ecosystems makes a lot more sense to me than strong-arming the button to behave differently when a glyph is added.

chrisdholt: You can have a CTA in an app that needs to be a button. From my viewpoint this is undoubtedly built out of the base button due to functional requirements of being either an anchor or a button. This seems like carry-over from previous implementations and I think there is a lot of reuse between button and CTA in MSFT styles. There is very little delta between these two. The difference is primarily in style enums, which is carryover from a previous implementation. As an aside - I would expect justified to exist here as well. Perhaps CTA is an implementation or extension of Button if need be, or we build it all new with the duplication of Button - at this point I definitely do not see this as Hypertext though.

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -0,0 +1,207 @@
...
+                    animation: {
+                        name: "cta-text-move-out",
+                        duration: "600ms",
+                        timingFunction: animationProperties,
+                        fillMode: "forwards"
+                    }
+                }
+            }
+        },
+        callToAction_span: {
```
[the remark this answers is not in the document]
nalogiudice: Will change3 to `textContainer`

nalogiudice:
Making changes discussed in meeting...will update soon.

packages/fast-components-class-name-contracts-msft/src/button/index.ts
```diff
@@ -8,6 +8,6 @@ export interface IMSFTButtonClassNameContract extends IButtonClassNameContract {
     button_outline: string;
     button_lightweight: string;
     button_justified: string;
-    button_span: string;
+    button_textContainer: string;
```
janechu: This would be a breaking change, lets move this to a release candidate branch for the next major version bump.

packages/fast-components-react-msft/src/button/button.props.ts
```diff
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { IButtonHandledProps as IBaseButtonHandledProps } from "@microsoft/fast-components-react-base";
 import { IManagedClasses, IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { ButtonSlot } from "./button";
```
janechu: imported but doesn't appear to get used?

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -133,4 +152,36 @@ describe("button", (): void => {
 
         expect(button.prop("className")).toEqual(undefined);
     });
+
+    /* tslint:disable-next-line */
+    test("should add an 'before' element when a child is passed with the before slot", () => {
```
janechu: nit: "should add a child element with the 'before' prop into the before slot"

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -52,12 +61,40 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
...
 
-    private generateInnerContent(): React.ReactElement<HTMLSpanElement> | (React.ReactNode | React.ReactNode[]) {
-        if (this.props.appearance === ButtonAppearance.lightweight || this.props.appearance === ButtonAppearance.justified) {
-            return <span className={get(this.props, "managedClasses.button_span")}>{this.props.children}</span>;
+    /**
+     * Renders slotted children in the appropriate slot
+     */
+    private renderChildrenBySlot(slot: ButtonSlot): JSX.Element {
+        if (Array.isArray(this.props.children)) {
+            return this.props.children.map((child: any, index: number) => {
```
janechu: I think we can use the `React.Children` instead and then you can skip the `isArray` check.

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -133,4 +152,36 @@ describe("button", (): void => {
...
+        const rendered: any = mount(
+            <Component {...props} />
+        );
+
+        expect(rendered.instance().props.children[1].props.slot).toBe("before");
+        expect(rendered.find("div.slotBefore").length).toBe(1);
+    });
+
+    /* tslint:disable-next-line */
+    test("should add an 'after' element when a child is passed with the after slot", () => {
```
janechu: can you also change this to match the above test string, something like "should add a child element with the slot prop set to 'after' into the after slot location" (also add similar wording above, my previous comment called it the 'before' prop but its the slot prop)

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -52,12 +61,40 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
...
+                if (!child.props) {
+                    content = child;
+                }
+            });
+        } else {
+            content = this.props.children;
         }
 
-        return this.props.children;
+        return <span className={get(this.props, "managedClasses.button_span")}>{content}</span>;
```
chrisdholt: If we are updating this, can we not refer to this as button_span?

packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * The class name contract for the call to action component
+ */
+export interface ICallToActionClassNameContract {
+    callToAction: string;
+    callToAction_backer: string;
```
janechu: lets change this to backplate to be more semantic

chrisdholt: Can we align these with the actual appearance type? This is a subset of button, so I would assume that if I wanted to override these changes I would do so by targeting the appearance, similar to how we have it in Button.

callToAction_primary: string;
callToAction_secondary: string;

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -52,12 +61,40 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
...
 
-    private generateInnerContent(): React.ReactElement<HTMLSpanElement> | (React.ReactNode | React.ReactNode[]) {
-        if (this.props.appearance === ButtonAppearance.lightweight || this.props.appearance === ButtonAppearance.justified) {
-            return <span className={get(this.props, "managedClasses.button_span")}>{this.props.children}</span>;
+    /**
+     * Renders slotted children in the appropriate slot
+     */
+    private renderChildrenBySlot(slot: ButtonSlot): JSX.Element[] {
+        if (Array.isArray(this.props.children)) {
+            return this.props.children.map((child: any, index: number) => {
```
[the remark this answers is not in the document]
nalogiudice: Yep yep.

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -52,12 +61,40 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
...
+        }
+    }
+
+    /**
+     * Renders the non-slot child content
+     */
+    private renderContent(): JSX.Element {
+        let content: any = {};
+
+        if (Array.isArray(this.props.children)) {
```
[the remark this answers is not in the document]
nalogiudice: Done

packages/fast-components-react-msft/src/button/examples.data.tsx
```diff
@@ -13,6 +13,20 @@ import schema from "./button.schema.json";
...
+    <div
+        slot="before"
+        dangerouslySetInnerHTML={{__html: "<"}}
+    />
+);
+
+const afterSlotExample: JSX.Element = (
+    <div
+        slot="after"
+        dangerouslySetInnerHTML={{__html: " >"}}
```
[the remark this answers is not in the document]
nalogiudice: `<` actually errored, so switch to `{"<"}`

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
```diff
@@ -0,0 +1,86 @@
...
+                {...this.unhandledProps()}
+                className={this.generateClassNames()}
+                {...this.props}
+                jssStyleSheet={styles}
+            >
+                {this.props.children}
+                <div
+                    slot="after"
+                    className={get(this.props, "managedClasses.glyph")}
+                    dangerouslySetInnerHTML={{__html: glyphArrowright}}
```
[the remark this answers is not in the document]
nalogiudice: Yes, it seems this way... the glyph is an SVG and this is the way we have set SVGs elsewhere in the repo (see the app building blocks or just search dangerouslySetInnerHTML - it is all glyph usage)

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    config: IDesignSystem
+): ComponentStyleSheet<ICallToActionClassNameContract, IDesignSystem> => {
+    type ContrastFunction = (operandColor: string, referenceColor: string) => string;
+
+    const designSystem: IDesignSystem = withDesignSystemDefaults(config);
+    const contrastScale: number = designSystem.contrast;
+    const backgroundColor: string = designSystem.backgroundColor;
+    const brandColor: string = designSystem.brandColor;
+    const direction: Direction = designSystem.direction;
+    const scaledEnsureNormalContrast: ContrastFunction = curry(ensureNormalContrast)(contrastScale);
```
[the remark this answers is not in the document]
nalogiudice: Removed unneeded stuffs

packages/fast-components-react-msft/src/call-to-action/call-to-action.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+        const callToAction: any = rendered.first().shallow();
+        // Get the expected className value from the list of generated managed classes
+        const expectedClassName: string = callToAction.instance().props.managedClasses.callToAction_lightweight;
+        const rootClassName: string = callToAction.instance().props.managedClasses.callToAction;
+
+        expect(callToAction.instance().props.appearance).toEqual(CallToActionAppearance.justified);
+        // Generated managedClass should be passed to className
+        expect(callToAction.prop("className")).toBe(`${rootClassName} ${expectedClassName}`);
+    });
```
chrisdholt: Can we add two tests? One to ensure that a custom className can be passed when the appearance prop is set, and one that ensures that it can be passed when it is NOT set?

packages/fast-components-class-name-contracts-msft/src/call-to-action/index.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * The class name contract for the call to action component
+ */
+export interface ICallToActionClassNameContract {
+    callToAction: string;
+    callToAction_primary: string;
```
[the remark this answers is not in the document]
nalogiudice: Ah yes, this wouldn't be a breaking change yet - sorry my mind was on the breaking change branch for these changes - will change here

packages/fast-components-react-msft/src/call-to-action/call-to-action.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+describe("call to action snapshot", (): void => {
+    generateSnapshots(examples);
+});
+
+describe("call to action", (): void => {
+    const Component: React.ComponentClass<ICallToActionHandledProps & ICallToActionManagedClasses> = examples.component;
+    const managedClasses: ICallToActionClassNameContract = {
+        callToAction: "call-to-action",
+        callToAction_backer: "call-to-action-backer",
```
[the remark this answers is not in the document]
nalogiudice: Oddly enough it didn't when I ran locally, but it is wrong - will change.

packages/fast-components-react-msft/src/call-to-action/call-to-action.spec.tsx
nalogiudice: Sure

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Generates class names
+     */
+    protected generateClassNames(): string {
+        let classNames: string = super.generateClassNames(get(this.props, "managedClasses.callToAction"));
+
+        if (this.props.disabled) {
+            classNames += `${classNames} ${get(this.props, "managedClasses.button__disabled")}`;
```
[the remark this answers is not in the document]
nalogiudice: Whoops this should be `callToAction__disabled`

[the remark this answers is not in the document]
chrisdholt: I trust that we need both, but can you provide some context as to why `button__disabled` doesn't do what we need?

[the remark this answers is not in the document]
nalogiudice: We will have to style the glyph, same as all of the other appearance modifiers

packages/fast-components-react-msft/src/call-to-action/call-to-action.spec.tsx
nalogiudice: @chrisdholt tests added

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -28,6 +29,24 @@ describe("button", (): void => {
 
     const href: string = "https://www.microsoft.com";
 
+    const beforeSlotExample: JSX.Element = (
+        <div
+            className={"slotBefore"}
+            key={"Slot"}
+            slot="before"
+            dangerouslySetInnerHTML={{__html: "Before slot"}}
```
nicholasrice: Why are you are you dangerously setting inner html?

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -50,12 +59,34 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
...
+        });
+    }
+
+    /**
+     * Renders the non-slot child content
+     */
+    private renderContent(): JSX.Element {
+        const content: any[] = [];
+
+        React.Children.forEach(this.props.children, (child: JSX.Element, index: number) => {
```
nicholasrice: What you're effectively doing here is filtering out items, so the more semantic way to do this is:
```
React.Children.toArray(this.props.children).filter((() => {
//...
})`
```

Doing it this way has two benefits: the method name accurately describes what you're doing and you avoid mutating functions

packages/fast-components-react-msft/src/call-to-action/call-to-action.spec.tsx
```diff
@@ -0,0 +1,172 @@
...
+
+        const rendered: any = shallow(
+            <Component {...props} />
+        );
+
+        expect(rendered.first().prop("data-my-custom-attribute")).toEqual(true);
+    });
+
+    // tslint:disable-next-line:max-line-length
+    test("should set a className that matches callToAction__primary managedClass when CallToActionAppearance.primary is passed to the appearance prop", () => {
```
nicholasrice: Try to avoid implementation details in the test descriptions:
"should apply a 'primary' html class when appearance is primary" is shorter and describes the *intent* of the test but not how the test should be implemented.

Can you apply this pattern to the other tests as well?

packages/fast-components-react-msft/src/button/button.spec.tsx
nalogiudice: Whoops - this was a hold over from the glyph - fixing.

packages/fast-components-react-msft/src/button/button.tsx
nalogiudice: Sure

packages/fast-components-react-msft/src/call-to-action/call-to-action.spec.tsx
nalogiudice: All updated

nalogiudice:
@nicholasrice all comments addressed

packages/fast-glyphs-msft/.babelrc
```diff
@@ -0,0 +1,11 @@
+{
```
janechu: this needs to be added to the npmignore

packages/fast-components-react-msft/package.json
```diff
@@ -73,6 +73,7 @@
     "@microsoft/fast-glyphs-msft": "^2.2.0",
     "@microsoft/fast-jest-snapshots-react": "^2.2.0",
     "@microsoft/fast-jss-manager": "^2.2.0",
+    "@microsoft/fast-jss-utilities": "^2.2.0",
```
janechu: This should be removed and the styles should be exported from the styles package.

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { ICallToActionClassNameContract, IManagedClasses } from "@microsoft/fast-components-class-name-contracts-msft";
+import { get } from "lodash-es";
+import { glyphArrowright } from "@microsoft/fast-glyphs-msft";
+import { applyLocalizedProperty, localizeSpacing, toPx } from "@microsoft/fast-jss-utilities";
+import { IDesignSystem, withDesignSystemDefaults } from "@microsoft/fast-components-styles-msft";
+import { ComponentStyles } from "@microsoft/fast-jss-manager";
+import { Direction } from "@microsoft/fast-application-utilities";
+
+// Since MSFT button is already styled, we need to override in this way to alter button classes
+const styles: ComponentStyles<Partial<IMSFTButtonClassNameContract>, IDesignSystem> = {
```
janechu: can we move this to the `fast-components-styles-msft` package as a secondary export named `calltToActionButtonOverrides`

chrisdholt: missed this; good catch @janechu

packages/fast-glyphs-msft/.babelrc
nalogiudice: Added

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -0,0 +1,108 @@
+import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
+import { ICallToActionClassNameContract, IMSFTButtonClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { applyLocalizedProperty, Direction, localizeSpacing } from "@microsoft/fast-jss-utilities";
+import { IDesignSystem, withDesignSystemDefaults } from "../design-system/index";
+import { disabledContrast, ensureNormalContrast } from "../utilities/colors";
+
+// Since MSFT button is already styled, we need to override in this way to alter button classes
+export const calltToActionButtonOverrides: ComponentStyles<Partial<IMSFTButtonClassNameContract>, IDesignSystem> = {
```
janechu: spelling * callToActionButtonOverrides

PR #855, opened 2018-09-06
feat: add density to MSFT design system

### LINKED ISSUE #848: Implement desnsity into MSFT design system
The MSFT design system will introduce the concept of density into its components. Density will be used to describe the physical size and space between UI elements. Integrating density in this way will allow re-use of common UI components across a wider range of experiences without having to redefine dimensions.

Integrating density will introduce two new design system variables to the MSFT design system:
- `density`: number
  - Default: 1
  - A multiplier to make adjustments to style values. Reducing the density value will result in a smaller style value creating a more dense experience. Increasing density will have the opposite effect and will create a less dense experience.
- `designUnit`: number
  - Default: 4
  - A value to help derive style values that snap to design grids.

Several utilities will also be implemented to facilitate density integrations, including:
`density(value: number, unit?: "string" = "px"): (designSystem: IDesignSystem) => string`:
The `density` function will accept both a value and optionally a unit - it will return a function that accepts the MSFT design system, which will return the provided value multiplied by the design systems density value.

`interval(value: number, interval: number)`:
Rounds a number to the nearest value that is an interval for a number. This will be useful for snapping numbers to the design grid.
interval(10, 4) -> 8

nicholasrice:
closes #848


packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -47,8 +48,8 @@ const styles: ComponentStyles<IMSFTProgressClassNameContract, IDesignSystem> = (
                 position: "absolute",
                 display: "inline-block",
                 background: brandColor,
-                width: "5px",
-                height: "5px",
+                width: toPx(designSystem.designUnit),
```
nalogiudice: Making sure these values were supposed to change from the prior values (i.e. they are all now `4px`)

nicholasrice: We're going to do a design review, but 5px doesn't really align with how elements are sized in the MSFT design language so I snapped them down a px.

nalogiudice: Cool

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -1,22 +1,23 @@
...
-    /* tslint:disable-next-line */
     const determinateBackgroundColor: string = largeContrast(designSystem.contrast, designSystem.backgroundColor, brandColor);
 
     return {
         progress: {
             display: "flex",
             width: "100%",
             alignItems: "center",
-            height: "10px",
+            height: toPx(designSystem.designUnit),
```
janechu: this is changing to 4px, is this right?

nicholasrice: yes. The vertical height of the visible UI is 4px, however this was set to ten. The content-box should match the physical height of the UI element. If we want to add vertical spacing we should use margin.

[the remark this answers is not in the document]
nicholasrice: Yep, we can use margin if we need more vertical space but I wanted to align the content-box to the height of the children

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -20,33 +21,37 @@ function hoverColor(config: IDesignSystem): string {
...
+        textField: {
+            ...applyType("t7", "vp1"),
+            color: ensureForegroundNormal,
+            background: ensuresBackgroundNormal,
+            fontWeight: fontWeight.light.toString(),
+            border: "1px solid transparent",
+            borderColor: foregroundNormal,
+            boxSizing: "border-box",
+            borderRadius: "2px",
+            padding: "10px",
```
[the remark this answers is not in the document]
nicholasrice: It is being used to set a desired height that *does* snap to the design grid, so I left it as is

PR #869, opened 2018-09-10
feat(dialog): add onDismiss callback to support "soft dismiss" functionality

### LINKED ISSUE #861: Soft dismiss callback on Dialog
Often times "dialogs" or "lightboxes" have soft-dismiss options. I'd like to request a similar feature for Dialog. A new prop, "onDismiss" could be passed in that gets called when a user clicks on the modal overlay. Or a similar functionality that gives the consuming component a way to know when a user has clicked outside the content.

chrisdholt:
<body>
Adds a new prop of `onDismiss`. When passed, the callback will be fired in the following situations:
- When a modal dialog is open and the modal overlay is clicked
- When a modal or non-modal dialog is open and the escape key is pressed

closes #861 
<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/dialog/dialog.props.ts
+++ b/packages/fast-components-react-base/src/dialog/dialog.props.ts
@@ -42,7 +42,7 @@ export interface IDialogHandledProps {
 
     /**
      * The onDismiss callback
-     * Callback is registered on click of the modal or when escape key is pressed
+     * Callback is registered on click of the modal overlay or when escape key is pressed
      */
     onDismiss?: DialogOnDismiss;
 
--- a/packages/fast-components-react-base/src/dialog/dialog.props.ts
+++ b/packages/fast-components-react-base/src/dialog/dialog.props.ts
@@ -41,12 +41,19 @@ export interface IDialogHandledProps {
     modal?: boolean;
 
     /**
+     * The onDismiss callback
+     * Callback is registered on click of the modal or when escape key is pressed
+     */
+    onDismiss?: DialogOnDismiss;
+
+    /**
      * Sets the visibility of the dialog to assistive technologies
      * If true, aria-hidden is false
      */
     visible?: boolean;
 }
 
+export type DialogOnDismiss = (e?: React.MouseEvent | KeyboardEvent) => void;
 export interface IDialogUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
 export interface IDialogManagedClasses extends IManagedClasses<IDialogClassNameContract> {}
 export type DialogProps = IDialogHandledProps & IDialogUnhandledProps & IDialogManagedClasses;
--- a/packages/fast-components-react-base/src/dialog/dialog.tsx
+++ b/packages/fast-components-react-base/src/dialog/dialog.tsx
@@ -110,13 +110,13 @@ class Dialog extends Foundation<IDialogHandledProps & IManagedClasses<IDialogCla
     }
 
     private handleOverlayClick = (event: React.MouseEvent): void => {
-        if (typeof this.props.onDismiss === "function" && this.props.visible) {
+        if (this.props.onDismiss && typeof this.props.onDismiss === "function" && this.props.visible) {
             this.props.onDismiss(event);
         }
     }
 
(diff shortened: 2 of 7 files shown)


packages/fast-components-react-base/src/dialog/dialog.props.ts
```diff
@@ -40,13 +40,20 @@ export interface IDialogHandledProps {
      */
     modal?: boolean;
 
+    /**
+     * The onDismiss callback
+     * Callback is registered on click of the modal or when escape key is pressed
```
nalogiudice: "...on click of modal" means when modal overlay is clicked, correct?

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -50,6 +53,15 @@ class Dialog extends Foundation<IDialogHandledProps & IManagedClasses<IDialogCla
         );
     }
 
+    /**
+     * React life-cycle method
+     */
+    public componentDidMount(): void {
+        if (canUseDOM() && this.props.onDismiss) {
+            window.addEventListener("keydown", this.onDialogKeyboardDismiss);
```
nalogiudice: Don't we need to remove the event listener on unMount?

chrisdholt: yep - updated.

packages/fast-components-react-base/src/dialog/dialog.props.ts
chrisdholt: yes. Updated with more detail.

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -50,6 +53,35 @@ class Dialog extends Foundation<IDialogHandledProps & IManagedClasses<IDialogCla
...
+        }
+    }
+
+    /**
+     * React life-cycle method
+     */
+    public componentDidUpdate(prevProps: Partial<IDialogHandledProps>): void {
+        if (canUseDOM() && !prevProps.onDismiss && this.props.onDismiss) {
```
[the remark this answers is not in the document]
chrisdholt: agreed and updated

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -68,11 +80,26 @@ class Dialog extends Foundation<IDialogHandledProps & IManagedClasses<IDialogCla
...
+
+    private onClickDismiss = (event: React.MouseEvent): void => {
+        if (!this.props.onDismiss || !this.props.visible) {
+            return;
+        }
+
+        this.props.onDismiss(event);
+    }
+
+    private onDialogKeyboardDismiss = (event: KeyboardEvent): void => {
```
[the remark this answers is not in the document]
chrisdholt: updated

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -68,11 +102,24 @@ class Dialog extends Foundation<IDialogHandledProps & IManagedClasses<IDialogCla
...
                 className={get(this.props, "managedClasses.dialog_modalOverlay")}
+                onClick={this.handleOverlayClick}
                 role={"presentation"}
                 tabIndex={-1}
             />
         );
     }
+
+    private handleOverlayClick = (event: React.MouseEvent): void => {
+        if (this.props.onDismiss && this.props.visible) {
```
nicholasrice: you should also make sure `this.props.onDismiss` is a *function*, not just a truthy value. `typeof this.props.onDismiss === "function"`

awentzel:
I don't believe that modal dialogs should support 'escape' keypress to trigger 'onDismiss'. For example 'Cancel' or 'Accpet'.  Now on 'Cancel', it would make sense to trigger 'onDismiss'. So if there is a cancel button I could see that 'escape' would be acceptable.  

However, what if the two buttons were 'Accept' or 'Reject' and neither was defined by 'cancel' action.  You wouldn't want 'escaping' because it wouldn't have a valid return value which is required.

Is this an optional?  What is the default?

PR #872, opened 2018-09-11
chore: use lerna independent mode

nicholasrice:


package.json
```diff
@@ -18,7 +18,7 @@
   "devDependencies": {
     "@types/node": "^9.4.7",
     "glob": "^7.1.2",
-    "lerna": "^3.0.6",
+    "lerna": "^3.3.0",
```
chrisdholt: @nicholasrice will we need to update circle-ci as well with this change or will this all be done locally?

nicholasrice: To my knowledge there is nothing that needs to happen to circle-ci here. This shouldn't affect builds, only publishing.

PR #875, opened 2018-09-12
feat(text-field): deprecates non string or number enum types which are not valid as text field inputs

### LINKED ISSUE #721: Update to Text-field to remove non string/number input types
"Text field" should only include input types for strings and numbers. Meaning that date and range should be new components.

Deliverables:
1. File a new feature to create date component.
2. File a new feature to create range component.
3. Remove the date and range input types from Text Field.
4. Validate the schema is accurate
5. Validate the configurator no longer feels broken.
6. Assign someone to perform the QA review once in PR.

chrisdholt:
<body>
closes #721
<footer>
 


packages/fast-components-react-base/src/text-field/text-field.props.ts
```diff
@@ -2,19 +2,39 @@ import * as React from "react";
...
-   time = "time",
-   url = "url",
-   week = "week"
+    /**
+     * @deprecated as of v2.3.0
+     * Will be deleted in version 3.0.
+     */
+    date = "date",
+    email = "email",
+    hidden = "hidden",
```
nicholasrice: Should we also do hidden? It doesn't seem like hidden is a good fit for this component

packages/fast-components-react-base/src/text-field/text-field.tsx
```diff
@@ -41,6 +41,22 @@ class TextField extends Foundation<ITextFieldHandledProps & IManagedClasses<ITex
...
+        switch (type) {
+            case TextFieldType.date:
+            case TextFieldType.month:
+            case TextFieldType.range:
+            case TextFieldType.time:
+            case TextFieldType.week:
+                /* tslint:disable-next-line */
+                console.warn(`TextFieldType enum "${type}" has been deprecated as of version 2.3.0, and will be removed in version 3.0.0.`);
+                break;
+            default:
```
nicholasrice: You don't need a default case here

chrisdholt: good catch.

packages/fast-components-react-base/src/text-field/text-field.props.ts
chrisdholt: Yeah - I think we can.

janechu:
Can we change the merge title to "fix".

packages/fast-components-react-base/src/text-field/text-field.spec.tsx
```diff
@@ -94,4 +94,64 @@ describe("text-field", (): void => {
 
         expect(rendered.prop("placeholder")).toEqual("Test");
     });
+    /* tslint:disable no-console */
+    test("should throw a warning in the console if `TextFieldType.date` is passed to the type prop", () => {
+        console.warn = jest.fn();
```
nalogiudice: Is there a way of only decalring this once?

chrisdholt:
@janechu From what I read in the semver docs, if you are deprecating part of an API it needs to be a minor version bump rather than a patch.

packages/fast-components-react-base/src/text-field/text-field.spec.tsx
chrisdholt: moved to a new describe.

nalogiudice: Cool thanks

PR #880, opened 2018-09-13
fix(button): add missing super of generateClassNames method when appearance prop is not set

### LINKED ISSUE #879: Styles Button does not take custom className
**Describe the bug**
Attempting to specify className with the MSFT style [Button](https://github.com/Microsoft/fast-dna/blob/master/packages/fast-components-react-msft/src/button/button.tsx) results in a no-op. Likely because it also creates it's own class names which I guess get applied internally?

**To Reproduce**
'''js
import { Button } from "@microsoft/fast-components-react-msft";
...
<Button className={managedClasses.addButton}/>
'''

**Expected behavior**
The externally specified className should probably be respected, or used in combination with any internally used class names instead of ignoring it.

I'm not sure if my expectation above is incorrect. Basically the className works fine for any of the MSFT styled components such as Label, TextField, Dialog, etc. but apparently does not work the same way for Button. We should expose a unified API so that it works in an expected way when devs attempt to use it.

Currently I'm using a raw <button> element to achieve what I need.

chrisdholt:
<body>
fixes #879 
<footer>
 

### DIFF
--- a/packages/fast-components-react-msft/src/button/button.tsx
+++ b/packages/fast-components-react-msft/src/button/button.tsx
@@ -37,7 +37,7 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
      */
     protected generateClassNames(): string {
         if (!this.props.appearance) {
-            return;
+            return super.generateClassNames();
         }
 
         switch (this.props.appearance) {
--- a/packages/fast-components-react-msft/src/button/button.tsx
+++ b/packages/fast-components-react-msft/src/button/button.tsx
@@ -36,10 +36,6 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
      * Generates class names
      */
     protected generateClassNames(): string {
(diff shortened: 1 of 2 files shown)


prabashn:
Holy turn-around time batman! :) Thanks for fixing this so quick Chris - you're awesome.

packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -37,7 +37,7 @@ class Button extends Foundation<IButtonHandledProps & IManagedClasses<IMSFTButto
      */
     protected generateClassNames(): string {
         if (!this.props.appearance) {
-            return;
+            return super.generateClassNames();
```
nalogiudice: Super still gets added by the base button, so I don't think this is necessary

nalogiudice: Super still gets added by the base button, so I don't think this is necessary? If it is, I would like to take care of it in my PR since I am altering button quite a bit.

PR #884, opened 2018-09-13
feat(radio): add radio component

khamudom:
create radio component in base and msft 
#868


packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { get } from "lodash-es";
+import Chroma from "chroma-js";
+
+const styles: ComponentStyles<IRadioClassNameContract, IDesignSystem> = {
+    radio: {},
+    radio_disabled: {},
+    radio_input: {},
+    radio_label: {},
+    radio_stateIndicator: {}
+};
```
janechu: these all appear to be empty?

packages/fast-components-class-name-contracts-base/src/radio/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * The class name contract for the radio component
+ */
+export interface IRadioClassNameContract {
+    radio: string;
+    radio_disabled: string;
```
nalogiudice: Should be `radio__disabled`

packages/fast-components-react-base/src/radio/README.md
```diff
@@ -0,0 +1,8 @@
+# Radio
+An element that can be turned on and off. *Radio* buttons are almost always listed together in groups.
```
nalogiudice: Maybe "A radio component is an input element that can be turned on and off"

packages/fast-components-react-base/src/radio/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+## Usage
+Only one *radio* within a *radio* group may be selected at a time. Limit the *radio* label to a single line.
+
+## Style guidance
+Should not put two radio groups next to each other. When two radio groups are right next to each other, it’s difficult to determine which radios belong to which group. Use group labels to separate them.
```
nalogiudice: "Do not put two radio groups next to each other."

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,39 @@
+import * as React from "react";
+import { IManagedClasses, IRadioClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export enum RadioHTMLTags {
```
nalogiudice: `RadioTag` to match standards

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,39 @@
...
+     */
+    checked?: boolean;
+
+    /**
+     * The disabled state
+     */
+    disabled?: boolean;
+
+    /**
+     * The selected state
```
nalogiudice: Wrong comment

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,39 @@
...
+     */
+    onChange?: boolean;
+
+    /**
+     * The textual content
+     */
+    text?: string;
```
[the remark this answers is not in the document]
nalogiudice: +1

packages/fast-components-react-base/src/radio/radio.schema.json
```diff
@@ -0,0 +1,30 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Radio",
+    "description": "A radio component's schema definition",
+    "type": "object",
+    "id": "radio",
```
nalogiudice: And add "children": true

packages/fast-components-react-base/src/radio/radio.schema.json
```diff
@@ -0,0 +1,30 @@
...
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "text": {
+            "title": "Text",
+            "type": "string"
+        }
+    }
+}
```
nalogiudice: nit: end with newline

packages/fast-components-react-msft/src/radio/README.md
```diff
@@ -0,0 +1,8 @@
+# Radio
+An element that can be turned on and off. *Radio* buttons are almost always listed together in groups.
```
nalogiudice: See other comments

packages/fast-components-react-base/src/radio/README.md
nalogiudice: Also style guidance should only be in MSFT. So move this change there

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface IRadioHandledProps {
+    /**
+     * The HTML tag
+     */
+    tag?: RadioHTMLTags;
+
+    /**
+     * The checked state
+     */
+    checked?: boolean;
```
chrisdholt: @nalogiudice @nicholasrice @janechu:

Per our discussion on props, do we want "checked" for user facing? It maps to the component, but does it map to an intuitive API for radio? Is a radio in a group "checked" or is it "selected"? Per conversations on toggle.

packages/fast-components-react-base/src/radio/radio.props.ts
chrisdholt: *maybe. Perhaps give input on line 18 question

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { get } from "lodash-es";
+
+/**
+ * Radio state interface
+ */
+export interface IRadioState {
+    checked: boolean;
+}
+
+class Radio extends Foundation<IRadioHandledProps & IRadioManagedClasses, IRadioUnhandledProps, IRadioState> {
```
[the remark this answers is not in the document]
khamudom: added getDerivedStateFromProps

packages/fast-components-react-base/src/radio/radio.schema.json
khamudom: added

packages/fast-components-react-base/src/radio/radio.schema.json
khamudom: add new line

packages/fast-components-react-base/src/radio/radio.props.ts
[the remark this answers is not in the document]
khamudom: changed to children

packages/fast-components-react-base/src/radio/radio.schema.json
```diff
@@ -0,0 +1,30 @@
...
+        },
+        "checked": {
+            "title": "Checked",
+            "type": "boolean"
+        },
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "text": {
```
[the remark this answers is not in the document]
khamudom: removed

packages/fast-components-react-base/src/radio/examples.data.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from "react";
+import Radio, { IRadioHandledProps, IRadioManagedClasses, IRadioUnhandledProps, RadioHTMLTags } from "./radio";
+import schema from "./radio.schema.json";
+import Documentation from "./.tmp/documentation";
+import { IComponentFactoryExample } from "@microsoft/fast-development-site-react";
+
+const classes: IRadioManagedClasses = {
+    managedClasses: {
+        radio: "radio",
+        radio_disabled: "radio_disabled",
```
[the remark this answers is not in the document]
khamudom: updated

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,111 @@
...
+     */
+    private get tag(): string {
+        return RadioHTMLTags[this.props.tag] || RadioHTMLTags.label;
+    }
+
+    private renderedLabel(): JSX.Element {
+        if (this.props.text) {
+            return(
+                <span className={get(this.props, "managedClasses.radio_label")}>
+                    {this.props.text}
```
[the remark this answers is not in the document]
khamudom: changed

packages/fast-components-react-base/src/radio/README.md
nicholasrice: +1 for Chris's comment - radio is really a single-selection model from a group of items

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,124 @@
...
+            >
+                <input
+                    className={get(this.props, "managedClasses.radio_input")}
+                    type="radio"
+                    ref={this.inputRef}
+                    onChange={this.handleRadioChange}
+                    disabled={this.props.disabled || null}
+                    checked={this.state.checked}
+                />
+                <span className={get(this.props, "managedClasses.radio_stateIndicator")} />
```
nicholasrice: radio requires a label, so this should always create a label element - I think the best structure for this would be as a sibling to the input element. Then we can make the wrapping element a div and remove the "tag" prop

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,124 @@
...
+                </span>
+            );
+        }
+    }
+
+    /**
+     * Handles onChange as a controlled component
+     */
+    private handleRadioChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+        if (typeof this.props.checked !== "boolean") {
```
nicholasrice: I don't think that you actually need to manage state here in this component. You can use the "defaultChecked" property on the input to assign a checked value if no value prop is passed - that will handled the unhandled state. Then, if value is a boolean, it should be applied as the value of the input.

packages/fast-components-react-base/src/radio/README.md
khamudom: going with: "Only one *radio* button within a group of items can be selected at the same time."

packages/fast-components-react-msft/src/radio/README.md
khamudom: going with: "Only one radio button within a group of items can be selected at the same time."

packages/fast-components-class-name-contracts-base/src/radio/index.ts
khamudom: changed

packages/fast-components-react-base/src/radio/radio.props.ts
khamudom: changed

packages/fast-components-react-base/src/radio/README.md
khamudom: removed style guide from here

packages/fast-components-react-base/src/radio/radio.props.ts
nicholasrice: I think here "checked" is expected because it has a direct mapping to the HTML element - both input's and aria use "checked" for the "checkbox" type/role

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * The onChange state
+     */
+    onChange?: boolean;
+
+    /**
+     * The caption content
+     */
+    children?: React.ReactNode | React.ReactNode[];
```
nicholasrice: nit - can you remove `React.ReactNode[]`? You shouldn't need it

packages/fast-components-react-base/src/radio/radio.props.ts
nalogiudice: +1

packages/fast-components-styles-msft/src/index.ts
```diff
@@ -75,6 +75,12 @@ export { ParagraphStyles };
 import ProgressStyles from "./progress";
 export { ProgressStyles };
 
+/**
+ * Export radio style
```
janechu: nit: *styles

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { IRadioClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { get } from "lodash-es";
+import Chroma from "chroma-js";
+
+const styles: ComponentStyles<IRadioClassNameContract, IDesignSystem> = {
+    radio: {},
+    radio__disabled: {},
+    radio_input: {},
+    radio_label: {},
+    radio_stateIndicator: {}
```
janechu: how is it that these have no styles?

packages/fast-components-styles-msft/src/index.ts
khamudom: fixed

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,124 @@
...
+     * Define constructor
+     */
+    constructor(props: RadioProps) {
+        super(props);
+
+        this.state = {
+            checked: this.props.checked || false
+        };
+
+        this.inputRef = React.createRef();
```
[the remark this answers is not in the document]
khamudom: removed

packages/fast-components-react-base/src/radio/radio.tsx
khamudom: add "slot"

packages/fast-components-react-base/src/radio/radio.tsx
khamudom: add "slot label"

packages/fast-components-styles-msft/src/radio/index.ts
khamudom: added some styles

packages/fast-components-react-base/src/radio/examples.data.tsx
```diff
@@ -0,0 +1,38 @@
...
+    detailData: {
+        ...classes,
+        checked: true
+    },
+    data: [
+        {
+            ...classes,
+            checked: true
+        },
+        {
```
nicholasrice: Can you add an example of `checked: false`, and without a `checked` value at all?

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+        classes = this.props.disabled ? `${classes} ${get(this.props, "managedClasses.radio__disabled")}` : classes;
+
+        return super.generateClassNames(classes);
+    }
+
+    /**
+     * Renders slotted children in the appropriate slot
+     */
+    private renderChildrenBySlot(slot: RadioSlot): React.ReactChild[] {
```
nicholasrice: You can remove this function - a version of it exists on the Foundation class now, so just change all references to this function to `this.withSlot(someSlot)`

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,88 @@
...
+                );
+            }
+        });
+    }
+
+    /**
+     * Handles onChange as a controlled component
+     */
+    private handleRadioChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
+        if (this.props.onChange) {
```
nicholasrice: because you're not doing any handling yourself, you can just drop `this.props.onChange` as a prop directly onto the input - no reason to have a function that calls the callback when React can handle this for us

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { IManagedClasses, IRadioClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface IRadioHandledProps {
+
+    /**
+     * Unique Id
+     */
+    id?: string;
```
nicholasrice: I think the ID should be required for accessability

packages/fast-components-react-msft/src/radio/examples.data.tsx
```diff
@@ -0,0 +1,63 @@
+import * as React from "react";
+import { IComponentFactoryExample } from "@microsoft/fast-development-site-react";
+import Radio from "./index";
+import { IRadioHandledProps } from "@microsoft/fast-components-react-base";
+import schema from "@microsoft/fast-components-react-base/dist/radio/radio.schema.json";
+import Documentation from "./.tmp/documentation";
+import { Label } from "../label";
+
+const slotRadioExample: JSX.Element = (
```
nicholasrice: these need to have an `htmlFor` attribute that points to the ID of the radio.

packages/fast-components-react-base/src/radio/examples.data.tsx
khamudom: examples added

packages/fast-components-react-base/src/radio/radio.props.ts
khamudom: changed to be required

packages/fast-components-react-msft/src/radio/examples.data.tsx
khamudom: added htmlFor

packages/fast-components-react-base/src/radio/radio.schema.json
```diff
@@ -0,0 +1,28 @@
...
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+            
+        }
+    },
+    "reactProperties": {
+        "children": {
+            "title": "Building blocks",
+            "type": "children"
```
janechu: Can you add to this:
```
"type": "children",
"defaults": [
    "text"
]
```

packages/fast-components-react-base/src/radio/radio.schema.json
```diff
@@ -0,0 +1,28 @@
...
+            "type": "string"
+        },
+        "checked": {
+            "title": "Checked",
+            "type": "boolean"
+        },
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+
```
janechu: nit: empty line

packages/fast-components-react-msft/src/radio/examples.data.tsx
```diff
@@ -0,0 +1,64 @@
...
+        },
+        {
+            checked: true,
+            onChange: noop,
+            children: slotCheckedExample,
+            id: "checked",
+        },
+        {
+            disabled: true,
+            children: [slotDisabledExample],
```
janechu: I don't think this needs to be an array

packages/fast-components-react-msft/src/radio/index.ts
```diff
@@ -0,0 +1,8 @@
+import * as React from "react";
+import {
+    Radio
+} from "@microsoft/fast-components-react-base";
+import manageJss, { IJSSManagerProps } from "@microsoft/fast-jss-manager-react";
+import { IDesignSystem, RadioStyles } from "@microsoft/fast-components-styles-msft";
+
+export default manageJss(RadioStyles)(Radio);
```
janechu: just like with base you need to export this from the `/fast-comopnents-react-msft/src/index.ts`

packages/fast-components-react-base/src/radio/radio.schema.json
khamudom: removed empty line

packages/fast-components-react-base/src/radio/radio.schema.json
khamudom: added

packages/fast-components-react-msft/src/radio/examples.data.tsx
khamudom: no longer an array

packages/fast-components-react-msft/src/radio/index.ts
khamudom: exported

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,33 @@
+import { IDesignSystem, withDesignSystemDefaults } from "../design-system";
+import { ComponentStyles, ComponentStyleSheet, ICSSRules } from "@microsoft/fast-jss-manager";
+import { IRadioClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import {
+    disabledContrast,
+    ensureForegroundNormal,
+    ensureNormalContrast,
+    hoverContrast,
+    normalContrast
+} from "../utilities/colors";
```
chrisdholt: I don't see any of these being used. Can we remove?

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,33 @@
...
+            display: "inline-flex",
+            flexDirection: "row",
+            alignItems: "center"
+        },
+        radio__disabled: {
+            cursor: "not-allowed"
+        },
+        radio_input: {
+            background: backgroundColor
+        },
```
chrisdholt: Do we need disabled color here?

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,33 @@
...
+        radio: {
+            display: "inline-flex",
+            flexDirection: "row",
+            alignItems: "center"
+        },
+        radio__disabled: {
+            cursor: "not-allowed"
+        },
+        radio_input: {
+            background: backgroundColor
```
chrisdholt: Is this all that is required for Chrome and Edge to look 100% the same?

khamudom:
This feature requires the disabled label style #980

packages/fast-components-styles-msft/src/radio/index.ts
khamudom: added radio disabled style

packages/fast-components-styles-msft/src/radio/index.ts
khamudom: ended up using them

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,35 @@
...
+     */
+    onChange?: boolean;
+
+    /**
+     * The radio content
+     */
+    children?: React.ReactNode;
+}
+
+export interface IRadioUnhandledProps extends React.AllHTMLAttributes<HTMLElement> {}
```
nalogiudice: @nicholasrice 's PR just went in that changes how this is done a bit. I think you need to update before it can go in

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { get } from "lodash-es";
+
+/**
+ * Radio slot options
+ */
+export enum RadioSlot {
+    label = "label"
+}
+
+class Radio extends Foundation<IRadioHandledProps & IRadioManagedClasses, IRadioUnhandledProps, {}> {
```
nalogiudice: See my other comment about @nicholasrice 's changes

packages/fast-components-react-base/src/radio/radio.props.ts
chrisdholt: @nalogiudice That work went into the v3 components branch. I think most of the work to align will need to be done in that branch or after it's merged in.

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private renderChildrenWithSlot(slot: RadioSlot): React.ReactChild[] {
+        const node: React.ReactNode = this.withSlot(RadioSlot.label);
+
+        return React.Children.map(node, (child: JSX.Element, index: number) => {
+            let labelSlotClassName: string = get(this.props, "managedClasses.radio_label");
+            const classNameKey: string = "className";
+
+            if (child.props[classNameKey] !== undefined) {
```
nalogiudice: @janechu should we use lodash's `isUndefined`?

packages/fast-components-react-msft/src/radio/README.md
```diff
@@ -0,0 +1,8 @@
+# Radio
+*Radio* buttons are almost always listed together in groups.
```
nalogiudice: Perhaps something like "*Radio* buttons are designed for selecting one value out of a set. Therefore, *radio* buttons are almost always listed together in groups."

khamudom: sounds good

packages/fast-components-react-base/src/radio/radio.props.ts
nicholasrice: yep - I will make the relevant updates when I rebase the 3.0.0 branch with this change so the commit history is accurate.

packages/fast-components-react-base/src/radio/radio.tsx
janechu: not necessary

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    const radioColor: string = normalContrast(designSystem.contrast, foregroundColor, backgroundColor);
+    const radioHover: string = hoverContrast(designSystem.contrast, foregroundColor, backgroundColor);
+    const radioDisabled: string = disabledContrast(
+        designSystem.contrast,
+        foregroundColor,
+        backgroundColor
+    );
+
+    return {
+        radio: {
```
janechu: These styles do not match radio styles, they look like checkbox. Also this should not be transparent it should use the background color.

packages/fast-components-react-base/src/radio/radio.props.ts
nalogiudice: @chrisdholt and @nicholasrice sorry guys - thought that went into master -disregard

packages/fast-components-react-base/src/radio/radio.tsx
nalogiudice: Disregard

packages/fast-components-styles-msft/src/radio/index.ts
khamudom: good catch. fixed

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,88 @@
...
+                height: "12px",
+                width: "12px"
+           }
+        },
+        radio__disabled: {
+            cursor: "not-allowed",
+            "& $radio_input": {
+                boxShadow: `inset 0 0 0 1px ${radioDisabled}`
+            },
+            "& $radio_label": {
```
chrisdholt: does `radio_label` only exist for styling of the disabled radio? I thought we were going with a prop value to style the label as "disabled" separately. Can you add some insight into why we reverted back to adding a class to a child? If a user wants to change the disabled color on the label using JssStylesheet, would they pass it to the radio or the label?

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,88 @@
...
+            },
+        },
+        radio_stateIndicator: {
+            position: "relative",
+            borderRadius: "50%",
+            display: "inline-block",
+            width: "20px",
+            height: "20px",
+            flexShrink: "0",
+            marginRight: "5px",
```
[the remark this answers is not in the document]
khamudom: RTL fixed

packages/fast-components-styles-msft/src/radio/index.ts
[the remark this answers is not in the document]
khamudom: added, cursor: "not-allowed"

khamudom: looks like I will be using radio_label for RTL. 
But as for why we ended up coming back to adding a class, I will let @nicholasrice answer that.

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,88 @@
...
+                width: "12px"
+           }
+        },
+        radio__disabled: {
+            cursor: "not-allowed",
+            "& $radio_input": {
+                boxShadow: `inset 0 0 0 1px ${radioDisabled}`
+            },
+            "& $radio_label": {
+                color: radioDisabled
```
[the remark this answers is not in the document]
khamudom: added

packages/fast-components-styles-msft/src/radio/index.ts
chrisdholt: That's fine - just looking for clarity on the change.

nicholasrice: @chrisdholt either is acceptable, but the recommendation is through the radio component's stylesheet, because the radio provides a single point of entry into the component's state.

The crux of the issue IMO is that a text element can never actually be *disabled* - it can only be related to disabled things. If we were to implement a *disabled* prop, it would be for the sole purpose of applying styling. To me, it makes more sense to contextually apply disabled styles to the label when the context is disabled than it does to add a stateful prop to an otherwise stateless component with un-contextual styles mapped to that prop. 

With a disabled prop on the label, it would be possible to create a non-disabled radio with a disabled label, and disabled radio with a non-disabled label. Additionally, the radio stylesheet wouldn't have the ability to style the label - set aside disabled for a moment, what if I want to apply margin-left to the label from the radio stylesheet?

chrisdholt: @nicholasrice for sure - when @janechu and I initially discussed this the idea was to have a prop not that distinguished the label as disabled, but the related input as disabled. So, I think we were in agreement in principle on that, perhaps it wasn't relayed in an offline conversation that way. Consider the benefit to disabled text-fields (no label included in base component), checkbox (where label may not be nested and unrelated to the actual input); those scenarios could benefit as well from adding a class on label to distinguish that the input is disabled.

On the others, I'll take a bit of a "devil's advocate" stance on a few things. Don't want to block (and don't plan to), but it seems like some of the points I brought up previously (also offline) may not have been articulated fully on this. 

>With a disabled prop on the label, it would be possible to create a non-disabled radio with a disabled label, and disabled radio with a non-disabled label. Additionally, the radio stylesheet wouldn't have the ability to style the label 

Correct. But, that aligns well with our principle that things that are *bad* are possible. We don't show it as a pattern, and we don't recommend it, but we haven't built out our components to prevent things like this previously.

>  set aside disabled for a moment, what if I want to apply margin-left to the label from the radio stylesheet?

I think this is probably the *best* argument for adding the label class here, but I would still argue that our principle has been to handle this through composition, jssStylesheet, and the many other ways that we expose. In the initial discussion I had on this, my primary point was, if we didn't go the prop route, we should handle adding the class in the MSFT version rather than the base version as it more closely follows an established pattern and our principles.

nicholasrice: To me, a principle of allowing "bad" things does not justify intentionally exposing more opportunity for failure when solutions that *don't* expose that opportunity exist.

Principally, a component can apply classes arbitrarily - we do this all the time. We use classes to apply styling to component parts, in base classes and out. I don't see how the label - a component part of the radio - is any different.

I understand the argument around text-box. From my perspective, the root of that issue is there is no abstraction to create a composite component with a label and checkbox. If we simply add a disabled prop to label, we've added an additional place an author needs to apply app state, which I think we can both agree is not an ideal and ergonomic solution.

If you feel we should not go this direction, please block the PR and we can figure out the best direction forward.

chrisdholt: Like I said above, I don't plan on blocking, nor am I currently blocking. We add classes all over; we have not modified children in this way up until this point. I wasn't arguing that as a reason not to do it, just trying to work through the implications.

chrisdholt:
@nicholasrice no need to block per my above comment. I do think we might want to align checkbox with this direction (use a slot for label and follow this pattern of applying a class).

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { ManagedClasses, RadioClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface RadioHandledProps {
```
nicholasrice: Can you change this to `export interface RadioHandledProps extends RadioManagedClasses`?

packages/fast-components-react-base/src/radio/radio.props.ts
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * The radio content
+     */
+    children?: React.ReactNode;
+}
+
+export interface RadioUnhandledProps extends React.AllHTMLAttributes<HTMLElement> {}
+export type RadioOnChange = (event?: React.ChangeEvent<HTMLElement>) => void;
+export interface RadioManagedClasses extends ManagedClasses<RadioClassNameContract> {}
+export type RadioProps = RadioHandledProps & RadioUnhandledProps & RadioManagedClasses;
```
nicholasrice: after the above change, you should then change this to `export type RadioProps = RadioHandledProps & RadioUnhandledProps`

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { get } from "lodash-es";
+
+/**
+ * Radio slot options
+ */
+export enum RadioSlot {
+    label = "label"
+}
+
+class Radio extends Foundation<RadioHandledProps & RadioManagedClasses, RadioUnhandledProps, {}> {
```
nicholasrice: You can remove `RadioManagedClasses` after you make changes to the prop interfaces.

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    public static displayName: string = "Radio";
+
+    /**
+     * Handled props instantiation
+     */
+    protected handledProps: HandledProps<RadioHandledProps & ManagedClasses<RadioClassNameContract>> = {
```
nicholasrice: same here: remove `& ManagedClasses<RadioClassNameContract>`

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+        return React.Children.map(node, (child: JSX.Element, index: number) => {
+            let labelSlotClassName: string = get(this.props, "managedClasses.radio_label");
+            const classNameKey: string = "className";
+
+            if (child.props[classNameKey] !== undefined) {
+                labelSlotClassName = `${child.props[classNameKey]} ${labelSlotClassName}`;
+            }
+
+            const childWithClass: JSX.Element = React.cloneElement(child, {className: labelSlotClassName});
```
nicholasrice: I don't think you need a React.Fragment here - you can just return `childWithClass`

packages/fast-components-react-msft/src/radio/examples.data.tsx
```diff
@@ -0,0 +1,64 @@
...
+            children: slotCheckedExample,
+            id: "checked",
+        },
+        {
+            disabled: true,
+            children: slotDisabledExample,
+            id: "disabled"
+        }
+    ]
+} as ComponentFactoryExample<RadioHandledProps>;
```
nicholasrice: this should be `ComponentFactoryExample<RadioProps>`

packages/fast-components-react-msft/src/radio/index.ts
```diff
@@ -0,0 +1,8 @@
+import * as React from "react";
+import {
+    Radio
+} from "@microsoft/fast-components-react-base";
+import manageJss, { ManagedJSSProps } from "@microsoft/fast-jss-manager-react";
+import { DesignSystem, RadioStyles } from "@microsoft/fast-components-styles-msft";
+
+export default manageJss(RadioStyles)(Radio);
```
nicholasrice: You'll need to update the exports here - I'll work with you directly on that.

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -0,0 +1,93 @@
...
+                boxShadow: `inset 0 0 0 2px ${radioColor}`,
+            },
+            "&:checked": {
+                "& + span": {
+                    "&::before": {
+                        position: "absolute",
+                        zIndex: "1",
+                        content: "\"\"",
+                        borderRadius: "50%",
+                        background: radioColor
```
janechu: This is still not applying the radioDisabled color when this is both checked and disabled.

khamudom: set checked to disabled when both check and disabled are selected

Issue #888, opened 2018-09-13
Tab Component is clobbering the role of `button` so it should be a `div`


eljefe223:
Tab Component should render a `div` and not a `button`.

chrisdholt:
@eljefe223 can you add more detail as to *why* this change should be made?

PR #889, opened 2018-09-13
fix(tab): tab should render as a div instead of a button element

eljefe223:
Closes: #888 
 

### DIFF
--- a/packages/fast-components-react-base/src/tabs/tabs.tsx
+++ b/packages/fast-components-react-base/src/tabs/tabs.tsx
@@ -142,7 +142,7 @@ class Tabs extends Foundation<ITabsHandledProps & ITabsManagedClasses, ITabsUnha
                     "aria-selected": this.state.activeId === tabItem.props.id,
                     onClick: this.handleClick,
                     onKeyDown: this.handleKeyDown,
-                    tabIndex: this.state.activeId !== tabItem.props.id ? -1 : 0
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -142,7 +142,7 @@ class Tabs extends Foundation<ITabsHandledProps & ITabsManagedClasses, ITabsUnha
                     "aria-selected": this.state.activeId === tabItem.props.id,
                     onClick: this.handleClick,
                     onKeyDown: this.handleKeyDown,
-                    tabIndex: this.state.activeId !== tabItem.props.id ? -1 : 0
+                    tabIndex: this.state.activeId !== tabItem.props.id ? -1 : 1
```
chrisdholt: Should be 0 instead of 1.

chrisdholt:
Can you update your merge message?

packages/fast-components-react-base/src/tabs/tabs.tsx
janechu: @eljefe223 to add to this it's helpful to look at https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex which is a more accurate source for this kind of information.

janechu:
please change the PR message to conform to our conventional commits

eljefe223:
Whoops sorry I missed the merge message. Will update.

Issue #892, opened 2018-09-14
Update all class names across base and MSFT to align with standards


nalogiudice:
Since we are introducing breaking changes in 3.0.0, now is the time to change any managed class names that do not adhere to our standards. A list was made with @chrisdholt @janechu @nicholasrice and myself with help from @Falkicon

PR #893, opened 2018-09-14
feat(class names): update all class names across base and MSFT to align with naming standards

nalogiudice:
Closes [#892](https://github.com/Microsoft/fast-dna/issues/892)
<body>

<footer>
 

### DIFF
--- a/packages/fast-components-class-name-contracts-base/src/button/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/button/index.ts
@@ -3,7 +3,7 @@
  */
 export interface IButtonClassNameContract {
     /**
-     * The root button component
+     * The root of the button component
      */
     button: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/button/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/button/index.ts
@@ -2,5 +2,8 @@
  * The class name contract for the button component
  */
 export interface IButtonClassNameContract {
+    /**
+     * The root button component
+     */
     button: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/checkbox/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/checkbox/index.ts
@@ -3,7 +3,7 @@
  */
 export interface ICheckboxClassNameContract {
     /**
-     * The root checkbox component
+     * The root of the checkbox component
      */
     checkbox: string;
 
@@ -23,7 +23,7 @@ export interface ICheckboxClassNameContract {
     checkbox_label: string;
 
     /**
-     * The checkbox state indicator (e.g. checked, inditerminant, or not checked)
+     * The checkbox state indicator
      */
     checkbox_stateIndicator: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/checkbox/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/checkbox/index.ts
@@ -2,9 +2,28 @@
  * The class name contract for the checkbox component
  */
 export interface ICheckboxClassNameContract {
+    /**
+     * The root checkbox component
+     */
     checkbox: string;
-    checkbox_disabled: string;
+
+    /**
+     * The disabled state modifier
+     */
+    checkbox__disabled: string;
+
+    /**
+     * The checkbox input element
+     */
     checkbox_input: string;
+
+    /**
+     * The label associated with the checkbox
+     */
     checkbox_label: string;
-    checkbox_span: string;
+
+    /**
+     * The checkbox state indicator (e.g. checked, inditerminant, or not checked)
+     */
+    checkbox_stateIndicator: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/dialog/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/dialog/index.ts
@@ -2,7 +2,18 @@
  * The class name contract for the dialog component
  */
 export interface IDialogClassNameContract {
+    /**
+     * The root dialog component
+     */
     dialog: string;
+
+    /**
+     * The dialog backdrop
+     */
     dialog_modalOverlay: string;
+
+    /**
+     * The foregournd dialog content region
+     */
     dialog_contentRegion: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/dialog/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/dialog/index.ts
@@ -3,17 +3,17 @@
  */
 export interface IDialogClassNameContract {
     /**
-     * The root dialog component
+     * The root of the dialog component
      */
     dialog: string;
 
     /**
-     * The dialog backdrop
+     * The element that overlays the page while the modal is visible
      */
     dialog_modalOverlay: string;
 
     /**
-     * The foregournd dialog content region
+     * The dialog content region
      */
     dialog_contentRegion: string;
 }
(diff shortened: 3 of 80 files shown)


packages/fast-components-class-name-contracts-base/src/checkbox/index.ts
```diff
@@ -2,9 +2,28 @@
...
     checkbox_input: string;
+
+    /**
+     * The label associated with the checkbox
+     */
     checkbox_label: string;
-    checkbox_span: string;
+
+    /**
+     * The checkbox state indicator (e.g. checked, inditerminant, or not checked)
```
nicholasrice: I'd omit the "eg" here - to me that is more confusing because it speaks to the *state* while the sentence speaks to the *element*

packages/fast-components-class-name-contracts-base/src/dialog/index.ts
```diff
@@ -2,7 +2,18 @@
  * The class name contract for the dialog component
  */
 export interface IDialogClassNameContract {
+    /**
+     * The root dialog component
+     */
     dialog: string;
+
+    /**
+     * The dialog backdrop
```
nicholasrice: You're mixing terminology here - maybe change to "The element that overlays the page while the modal is visible" or something to that effect

packages/fast-components-class-name-contracts-base/src/dialog/index.ts
```diff
@@ -2,7 +2,18 @@
...
+     */
     dialog_modalOverlay: string;
+
+    /**
+     * The foregournd dialog content region
```
nicholasrice: I don't think "foreground" is relevant here - "The dialog content region" tells me everything I need to know

packages/fast-components-class-name-contracts-base/src/horizontal-overflow/index.ts
```diff
@@ -2,8 +2,23 @@
  * The class name contract for the horizontal overflow component
  */
 export interface IHorizontalOverflowClassNameContract {
+    /**
+     * The root horizontal overflow component - contains overflow items and previous/next controls
```
nicholasrice: Why mention what this containers here and not other places? I'd suggest removing the second sentence.

packages/fast-components-class-name-contracts-base/src/horizontal-overflow/index.ts
```diff
@@ -2,8 +2,23 @@
...
+     */
     horizontalOverflow: string;
+
+    /**
+     * The overflow items
```
nicholasrice: This is a bit confusing - are these the items that shouldn't be displayed because they are overflowed?

Maybe we should change to "contentRegion" to align? That's really what this is

packages/fast-components-class-name-contracts-base/src/hypertext/index.ts
```diff
@@ -2,5 +2,8 @@
  * The class name contract for the hypertext component
  */
 export interface IHypertextClassNameContract {
+    /**
+     * The root hypertext component
```
nicholasrice: I think we should tweak the description for all of these "root" classes. Its the root *of* the component, not the root component

packages/fast-components-react-base/src/image/image.tsx
```diff
@@ -67,38 +66,24 @@ class Image extends Foundation<IImageHandledProps & IManagedClasses<IImageClassN
             return (
                 <picture
                     {...this.unhandledProps()}
-                    className={this.generateClasses()}
+                    className={super.generateClassNames(get(this.props, "managedClasses.image__picture"))}
```
nicholasrice: this should still recieve the root class - `managedClasses.image`. A modifier should never exist without a block

packages/fast-components-class-name-contracts-base/src/button/index.ts
```diff
@@ -2,5 +2,8 @@
  * The class name contract for the button component
  */
 export interface IButtonClassNameContract {
+    /**
+     * The root button component
```
chrisdholt: This reads a bit weird to me. This is technically the root element of the component. Can we consider an update to make these more clear?

packages/fast-components-class-name-contracts-base/src/checkbox/index.ts
chrisdholt: +1 and spelling of `indeterminate`

packages/fast-components-class-name-contracts-base/src/dialog/index.ts
chrisdholt: Correct; there's not a background content region on the component so this would confuse me

packages/fast-components-react-base/src/typography/typography.tsx
```diff
@@ -49,9 +49,9 @@ class Typography extends Foundation<ITypographyHandledProps & IManagedClasses<IT
      */
     protected generateClassNames(): string {
         const classes: string = this.props.typeLevel ?
-            get(this.props, `managedClasses.typography_${this.props.typeLevel}`) : get(this.props, "managedClasses.typography_1");
+            get(this.props, `managedClasses.typography__${this.props.typeLevel}`) : get(this.props, "managedClasses.typography__1");
 
-        return super.generateClassNames(classes);
+        return super.generateClassNames(`${get(this.props, "managedClasses.typography")} ${classes}`);
```
chrisdholt: for consistency in how this is typically done throughout the repo would prefer the base class be set first...but that's more preference. This actually leads to an interesting question - is level required now that we have a base class for users to hook to?

packages/fast-components-react-msft/src/flipper/flipper.tsx
```diff
@@ -26,7 +26,7 @@ class Flipper extends Foundation<IFlipperHandledProps & IManagedClasses<IFlipper
             <Button
                 {...this.unhandledProps()}
                 {...this.coerceButtonProps()}
-                managedClasses={this.props.managedClasses}
+                managedClasses={null}
```
chrisdholt: I don't think we need this

packages/fast-components-react-base/src/typography/typography.tsx
nalogiudice: It is? `super.generateClassNames(`${get(this.props, "managedClasses.typography")} ${classes}`);`
The typography base class will always be set first... if we decide we don't need a default, this syntax would change.

packages/fast-components-react-msft/src/flipper/flipper.tsx
nalogiudice: Yes, the flipper expects managedClasses from base button, but we no longer extend the class name contract, so this overrides those managed classes.

packages/fast-components-react-base/src/typography/typography.tsx
chrisdholt: Yeah - I see that. Multitasking when I wrote.

I would propose though that even if we do it in a different task, we should not set a default type level on the base component if we have a managed class to target the base.

packages/fast-components-class-name-contracts-base/src/tabs/index.ts
```diff
@@ -2,17 +2,42 @@
...
 export interface ITabsClassNameContract {
-    tabs_tabItems: string;
-    tabs_tabList: string;
+    /**
+     * The root of the tabs component
+     */
     tabs: string;
+
+    /**
+     * The tab oanels
```
chrisdholt: spelling: `panels`

PR #901, opened 2018-09-14
feat(style): add elevation shadow ramp and functions

### LINKED ISSUE #895: Add shadow ramp for "depth/elevation" to MSFT styles package
Depth is part of the current MSFT design system and varies based on context/use case. For instance app panels, flyouts, dialogs and contextual menus all have various levels of "elevation". We should create a ramp that makes it easy to reference these values as well as a function that makes it easy to apply.

chrisdholt:
<body>
closes #895 
<footer>
 

### DIFF
--- a/packages/fast-components-styles-msft/src/dialog/index.ts
+++ b/packages/fast-components-styles-msft/src/dialog/index.ts
@@ -5,7 +5,7 @@ import { toPx } from "@microsoft/fast-jss-utilities";
 import { get } from "lodash-es";
 import Chroma from "chroma-js";
 import { applyAcrylicMaterial } from "../utilities/acrylic";
-import { applyElevation, elevationRamp } from "../utilities/elevation";
+import { elevation, elevationRamp } from "../utilities/elevation";
 
 /* tslint:disable-next-line */
 const styles: ComponentStyles<IDialogClassNameContract, IDesignSystem> = (config: IDesignSystem): ComponentStyleSheet<IDialogClassNameContract, IDesignSystem> => {
@@ -34,7 +34,7 @@ const styles: ComponentStyles<IDialogClassNameContract, IDesignSystem> = (config
             top: "50%",
             transform: "translate(-50%, -50%)",
             background: backgroundColor,
-            ...applyElevation("e14", foregroundColor)(designSystem)
+            ...elevation("e14", foregroundColor)(designSystem)
         }
     };
 };
--- a/packages/fast-components-styles-msft/src/dialog/index.ts
+++ b/packages/fast-components-styles-msft/src/dialog/index.ts
@@ -5,7 +5,7 @@ import { toPx } from "@microsoft/fast-jss-utilities";
 import { get } from "lodash-es";
 import Chroma from "chroma-js";
 import { applyAcrylicMaterial } from "../utilities/acrylic";
-import { elevation, elevationRamp } from "../utilities/elevation";
+import { elevation, ElevationMultiplier } from "../utilities/elevation";
 
 /* tslint:disable-next-line */
 const styles: ComponentStyles<IDialogClassNameContract, IDesignSystem> = (config: IDesignSystem): ComponentStyleSheet<IDialogClassNameContract, IDesignSystem> => {
@@ -34,7 +34,7 @@ const styles: ComponentStyles<IDialogClassNameContract, IDesignSystem> = (config
             top: "50%",
             transform: "translate(-50%, -50%)",
             background: backgroundColor,
-            ...elevation("e14", foregroundColor)(designSystem)
+            ...elevation(ElevationMultiplier.e14, foregroundColor)(designSystem)
         }
     };
 };
--- a/packages/fast-components-styles-msft/src/dialog/index.ts
+++ b/packages/fast-components-styles-msft/src/dialog/index.ts
@@ -5,15 +5,13 @@ import { toPx } from "@microsoft/fast-jss-utilities";
 import { get } from "lodash-es";
 import Chroma from "chroma-js";
 import { applyAcrylicMaterial } from "../utilities/acrylic";
+import { applyElevation, elevationRamp } from "../utilities/elevation";
 
 /* tslint:disable-next-line */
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -0,0 +1,96 @@
...
+    blurMultiplier: 0.225,
+    xOffsetMultiplier: 0,
+    yOffsetMultiplier: 0.075,
+    opacity: .18
+};
+
+/**
+ * Directional shadow config
+ */
+export const directionalShadowConfig: IShadowConfig = {
```
nalogiudice: Would directionality change in RTL/LTR?

chrisdholt: Current implementation, no - simply because we fix the x value to 0.

packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -0,0 +1,96 @@
...
+    xOffsetMultiplier: 0,
+    yOffsetMultiplier: 0.4,
+    opacity: .22
+};
+
+/**
+ * Apply elevation
+ * Used to apply elevation shadow treatment to a component
+ */
+export function applyElevation(elevation: keyof IElevationRamp, color: string): (config: IDesignSystem) => ICSSRules<IDesignSystem> {
```
nicholasrice: for consistency, do we want to simply call this "elevation" ?

packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -0,0 +1,96 @@
...
+            boxShadow: `${directionalShadow}, ${ambientShadow}`
+        };
+    };
+}
+
+/**
+ * Generate Elevation Shadow
+ * Generates a string representing a box shadow value
+ */
+export function generateElevationShadow(elevation: number, color: string, shadowConfig: IShadowConfig): (config: IDesignSystem) => string {
```
nicholasrice: same concept here  - maybe just elevationShadow?

packages/fast-components-styles-msft/src/utilities/elevation.ts
nicholasrice: Should color have a default? Potentially using foreground?

chrisdholt: ah, yes consistency with density. That makes sense.

packages/fast-components-styles-msft/src/utilities/elevation.ts
chrisdholt: agree with this based on update to naming

packages/fast-components-styles-msft/src/utilities/elevation.ts
chrisdholt: Added default value of `designSystemDefaults.foreground` for color param

packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -0,0 +1,97 @@
...
+};
+
+/**
+ * Directional shadow config
+ */
+export const directionalShadowConfig: IShadowConfig = {
+    blurMultiplier: 0.9,
+    xOffsetMultiplier: 0,
+    yOffsetMultiplier: 0.4,
+    opacity: .22
```
janechu: nit: can we be consistent and have 0.22

packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -0,0 +1,97 @@
...
+};
+
+/**
+ * Apply elevation
+ * Used to apply elevation shadow treatment to a component
+ */
+/* tslint:disable-next-line */
+export function elevation(elevationKey: keyof IElevationRamp, color: string = designSystemDefaults.foregroundColor): (config: IDesignSystem) => ICSSRules<IDesignSystem> {
```
janechu: nit: I think this can be formatted and we can remove the disable-next-line comment above

packages/fast-components-styles-msft/src/dialog/index.ts
```diff
@@ -36,7 +34,7 @@ const styles: ComponentStyles<IDialogClassNameContract, IDesignSystem> = (config
             top: "50%",
             transform: "translate(-50%, -50%)",
             background: backgroundColor,
-            boxShadow: dialogShadow
+            ...elevation("e14", foregroundColor)(designSystem)
```
janechu: given how we are using this, should it be an enum? strings seem breakable.

chrisdholt: I don't think an enum allows us to be extensible and support other people creating their own ramp.

janechu: are we allowing people to create their own ramp within MSFT?

chrisdholt: Different teams may require different values. We do this with the typeramp and it follows our principles of extensibility. We give good defaults, but they can change or manipulate to serve their requirements; I think the same should be true here.

janechu: I think we need to find a solution that allows us to have type safety within our own config, if you find a good solution we should also apply this to typography.

chrisdholt: Typography will be another task. Can you create one for it?

chrisdholt: @janechu I think we should explore if it will work for typography, but while these were similar, they are quite different in terms of the functional requirements. 

I *do* think we can make an enum change in typography, just saying that I don't think the change will be 1:1

packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -0,0 +1,98 @@
...
+    blurMultiplier: number;
+    opacity: number;
+    xOffsetMultiplier: number;
+    yOffsetMultiplier: number;
+}
+
+/**
+ * The elevation ramp interface
+ */
+export interface IElevationRamp {
```
janechu: looks like not used anymore

PR #905, opened 2018-09-15
chore(devops): add cross browser testing on Sauce Labs

awentzel:
Include build tools to execute cross-browser testing for msft-docs documentation. This can be executed by each developer during development by simply running a script and reviewing changes in Sauce Labs at https://saucelabs.com/beta/dashboard/builds


build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Assign browser configurations to an array
+let browsers = getConfiguration(process.argv[2]);
+
+/**
+ * Run tests on Sauce Labs
+ * @param {string} branchName 
+ */
+function Test(branchName) {
```
nicholasrice: nit - we typically only capitalize classes - maybe rename to `testBranch` to be more descriptive

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,199 @@
...
+                        await driver.get(domain);
+
+                        // Note: There are many ways to nagivate/find elements on a page
+                        // unfortunately, not all are cross-browser compliant and so
+                        // for example, we can't use xpath. As a result, we're using
+                        // navigation though not necessarily most performant but effective.
+
+                        // After browsers matrix is determined we can fine tune and
+                        // optimize to use other selectors to see what works. If we're
+                        // t8esting mostly on modern browsers this isn't as much a limitation.
```
nicholasrice: nit: "testing"

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,199 @@
...
+ * @param {Phases} phases
+ * @returns {Configuration based on phases}
+ */
+const getConfiguration = (phases) => {
+
+    // Parse command arguments and return valid configs
+    try {
+        switch (phases) {
+            
+            case "alpha":
```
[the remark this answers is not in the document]
awentzel: I agree with your suggestion, however, sentence casing takes more code since there isn't an actual function like there is for UpperCase and LowerCase so I opted for the less code solution.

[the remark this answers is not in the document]
awentzel: 1. We could require the commands that user's pass in to just use sentence casing.  `node test-browsers.js Alpha' but this just didn't feel right.
2. We could instead pass in a char only,  `node test-browsers.js -a` but this didn't feel as intuitive and probably would then require more work to build in `--help` from the CLI on this file. 
3. Maybe there are some other work arounds?  Open to suggestions.

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,199 @@
...
+            case "release":
+                return new Configure(Phases.Release);
+                
+            default:
+                console.log(chalk.red("Invalid Argument : must be 'alpha', 'beta', or 'release'"));
+                return process.exit(1);
+        }
+
+    } catch(err) {
+        console.error(err)
```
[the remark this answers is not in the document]
awentzel: What do you mean "configure"?  Do you mean if we don't handle the error in a meaningful way back to the user?

build/testing/sauce-labs/test-browsers.js
awentzel: Good catch. I will update.

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,199 @@
...
+ * @param {string} branchName 
+ */
+function Test(branchName) {
+    console.log("Testing Git branch:", branchName);
+
+    // Execute Selenium/Appium Web Drivers on Sauce Labs for each browser configuration
+    var flows = browsers.map(function (browser) {
+
+        // Setup capabilities
+        let caps = {
```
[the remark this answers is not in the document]
awentzel: I will update this to `capabilities` since you are correct that it's not a standard acronym and we should follow our own coding standards.

build/testing/sauce-labs/test-browsers.js
awentzel: You don't like 'Tasting' modern browsers?  ;-)  To be fixed.

build/testing/sauce-labs/test-browsers.js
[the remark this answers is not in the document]
nicholasrice: Does all lower case work? I think that would be my expectation.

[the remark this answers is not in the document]
awentzel: converted Enum type from Any to String and eliminated lowercase as it's not needed now.

README.md
```diff
@@ -8,6 +8,7 @@ Teams spend significant time building parts of the design system (colors, variab
 [![CircleCI](https://circleci.com/gh/Microsoft/fast-dna/tree/master.svg?style=shield&circle-token=d159a8b24ccb8046e07138c98717c32cb92589d6)](https://circleci.com/gh/Microsoft/fast-dna/tree/master)
 [![Maintainability](https://api.codeclimate.com/v1/badges/8a74621e634a6e9b9561/maintainability)](https://codeclimate.com/github/Microsoft/fast-dna/maintainability)
 [![Test Coverage](https://api.codeclimate.com/v1/badges/8a74621e634a6e9b9561/test_coverage)](https://codeclimate.com/github/Microsoft/fast-dna/test_coverage)
+[![Build Status](https://saucelabs.com/buildstatus/FAST-DNA)](https://saucelabs.com/beta/builds/997f70860d604d83b3f5c69129216d02)
```
chrisdholt: This is interesting. I'm all for badges if they convey valuable information. When I view this, it shows:

"SauceLabs: complete"

Is this valuable? What does complete mean in the sense of the repository?

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,196 @@
...
+    });
+});
+
+/**
+ * Once Branch name returns start testing
+ */
+getBranchName.then(function(branchName) {
+
+    // slice off '/n' at the end of string
+    branchName = branchName.slice(0, -1);
```
nicholasrice: you could use `branchName.trim()` to do this also - will remove all white space surrounding the text

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,196 @@
...
+ * @see
+ * To understand in detail:
+ * Getting Started: https://help.crossbrowsertesting.com/selenium-testing/getting-started/javascript/
+ * Scripting docs: https://www.seleniumhq.org/docs/03_webdriver.jsp#chapter03-reference 
+ */
+const { Builder } = require("selenium-webdriver");
+const chalk = require('chalk');
+const { Configure, Phases } = require("../config-browsers.js");
+const { spawn } = require('child_process');
+const git = spawn('git', ['rev-parse', '--abbrev-ref', 'HEAD']);
```
chrisdholt: for readability, I would consider moving this away from the modules that are being required in.

build/testing/sauce-labs/test-browsers.js
```diff
@@ -0,0 +1,196 @@
...
+                        await driver.quit();
+                }
+        })
+    });
+}
+
+/**
+ * Get Branch Name using child_process and Git which runs Asynchronously
+ */
+var getBranchName = new Promise(function(resolve, reject) {
```
chrisdholt: this block reads a bit weird to me. Perhaps instead of getBranchName, we rename this to be initializeTest? Then we can chain the "then" directly with the promise and call initializeTest?

```
var initializeTest = new Promise(function(resolve, reject) {
    git.stdout.on('data', (data) => {
        resolve(`${data}`);
    });
}).then(function(branchName) {
     // slice off '/n' at the end of string
    branchName = branchName.slice(0, -1);
     // Run Testing
    test(branchName);
});

initializeTest();
```

README.md
chrisdholt: @nicholasrice or @janechu for thoughts

build/testing/sauce-labs/test-browsers.js
chrisdholt: @nicholasrice or @janechu for thoughts

README.md
awentzel: It is required for Open Sauce (open source) projects using Sauce Labs.  But, I can see your point since we're no longer plugging this directly into our pipeline.  I guess the good thing is it will still show status for the past person that ran it from their dev system.

Sometimes these don't always work until they are actually committed in readme.  If it doesn't I can fix it with another PR.

build/testing/sauce-labs/test-browsers.js
awentzel: Could you clarify what is "this"?  Are you talking about just the git / spawn line?

build/testing/sauce-labs/test-browsers.js
awentzel: But, this isn't initializing.  This is literally getting the current branch name.

build/testing/sauce-labs/test-browsers.js
awentzel: Good call.

build/testing/sauce-labs/test-browsers.js
awentzel: I moved near the rest of the Git work lower in the file.

README.md
chrisdholt: gotcha

build/testing/sauce-labs/test-browsers.js
janechu: I think if you want to go further you could skip the variable assignment so this should work:

```js
/**
* Execute tests against the branch
*/
new Promise(function(resolve, reject) {
    git.stdout.on('data', (data) => {
        resolve(`${data}`.trim());
    });
}).then(function(branchName) {
    test(branchName);
});
```

awentzel:
Closes #918

PR #906, opened 2018-09-15
chore: remove unused changelog script in root package

awentzel:


package.json
```diff
@@ -13,7 +13,6 @@
     "url": "git+https://github.com/Microsoft/fast-dna.git"
   },
   "scripts": {
```
chrisdholt: Are there any other scripts? Can we remove the entire reference to scripts?

awentzel: Yes. I have another PR up now adds three new scripts for Integration-testing.

PR #908, opened 2018-09-17
feat(fast-components-react-msft): adding density input to development site

khamudom:
- density slider is added and hooked up to work #907


packages/fast-components-react-msft/app/app.tsx
```diff
@@ -165,6 +180,13 @@ export default class App extends React.Component<{}, IAppState> {
         );
     }
 
+    private handleDensityUpdate = (e: React.ChangeEvent<HTMLInputElement>): void => {
+        const value: number = +e.target.value;
```
nicholasrice: I think you'll want to change this to `const value: number = parseInt(e.target.value, 10);` - `e.target.value` is a string and this is required to be a number.

awentzel:
Can you update the title to conform to standards defined [here](https://github.com/Microsoft/fast-dna/wiki/pull-request-guidance)

packages/fast-components-react-msft/app/app.tsx
khamudom: made the changed

packages/fast-components-react-msft/app/app.tsx
```diff
@@ -100,12 +102,24 @@ export default class App extends React.Component<{}, IAppState> {
...
                 </SiteCategory>
                 <div slot={ShellSlot.infoBar}>
-                    <ColorPicker
-                        foregroundColor={this.state.foregroundColor}
-                        backgroundColor={this.state.backgroundColor}
-                        accentColor={this.state.accentColor}
-                        onColorUpdate={this.handleColorUpdate}
-                    />
+                    <span style={{display: "flex", alignItems: "center", height: "100%"}}>
+                        <label style={{marginRight: "8px"}}>density</label>
```
chrisdholt: do we want this to be lowercase "d"? Also, any reason we aren't using the Label component here?

khamudom: switched to div and Label

khamudom: I used a lowercase "d" because the theme colors are lowercase.

khamudom:
Abandoning this pull request. Creating a new one for this change.

PR #914, opened 2018-09-18
feat(image): update viewport logic in favor of slots

### LINKED ISSUE #239: Fix "similar-code" issue in packages/fast-components-react-base/src/image/image.tsx
Similar blocks of code found in 2 locations. Consider refactoring.

https://codeclimate.com/github/Microsoft/fast-dna/packages/fast-components-react-base/src/image/image.tsx#issue_5acd43ed4b1b320001000175

### LINKED ISSUE #896: Refactor Image component to use slots to create a picture element
Currently the `<Image />` component includes props that can create a `<picture></picture>` element. In order to do this, the component makes assumptions about the maximum number of potential `srcset` elements as well as hard coded viewport styles. In order to allow the most flexibility out of the component, we should instead use slots to pass srcset elements. In the event that those slots exist, we would create a srcset element.

This would remove some significant assumptions and barriers to using responsive images and allow us to support a more robust implementation of the picture element, as well as support responsive `` elements.

chrisdholt:
<body>
closes #896, closes #239 
<footer>
 


chrisdholt:
Also closes #239

packages/fast-components-react-base/src/image/image.props.ts
```diff
@@ -35,34 +24,9 @@ export interface IImageHandledProps {
     srcSet?: string;
```
nicholasrice: Is `srcSet` still valid?

chrisdholt: it's valid on the image itself as part of the responsive images spec. It gets passed to `` but not `<picture>` so I think it's valid as a handled prop.

packages/fast-components-react-base/src/image/image.tsx
```diff
@@ -5,101 +5,69 @@ import Foundation, { HandledProps } from "../foundation";
...
-        const minWidth: number = vp;
-
-        if (!srcset) {
-            return null;
-        }
-
-        const minWidthString: string = minWidth === 0 ? "0" : `${minWidth}px`;
-
-        return <source srcSet={srcset} media={`(min-width: ${minWidthString})`} />;
+    private renderChildrenBySlot(slot: ImageSlot): Array<React.ReactElement<HTMLSourceElement>> {
```
nalogiudice: I think you'll want to use filter here to be more semantic - thoughts @nicholasrice ?

packages/fast-components-react-base/src/image/image.tsx
```diff
@@ -5,101 +5,69 @@ import Foundation, { HandledProps } from "../foundation";
...
 
-        if (this.props.src) {
+        if (!this.props.children) {
             return (
                 <img
                     {...this.unhandledProps()}
-                    className={super.generateClassNames(get(this.props, "managedClasses.image"))}
+                    className={super.generateClassNames(className)}
                     alt={this.props.alt}
                     sizes={this.props.sizes ? this.props.sizes : null}
```
nalogiudice: I wonder if we want to do something like `{...this.props. sizes && {'sizes' : this.props.sizes}}` so it doesn't get written to the DOM if it doesn't exist

packages/fast-components-react-base/src/image/image.schema.json
```diff
@@ -4,90 +4,28 @@
     "description": "An image component's schema definition.",
     "type": "object",
     "id": "image",
```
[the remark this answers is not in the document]
chrisdholt: Not in this case. I spoke with @janechu about this. We aren't supporting "arbitrary" children here. It's typed to be specific to `<source>` elements. Jane is also working on an updated API for children that will change everything, and we thought it best to keep it out for now.

packages/fast-components-react-base/src/image/image.tsx
chrisdholt: This doesn't get written to the DOM if it doesn't exist.

packages/fast-components-react-base/src/image/image.tsx
nicholasrice: I learned something new today - `React.Children.map` will implicitly filter out any `null` values returned by the callback (not sure if it works for undefined or not but probably?)

Do you really need a key here? Can you not just return the `child` if the condition is met and `null` if it isn't?

packages/fast-components-react-base/src/image/image.tsx
```diff
@@ -5,101 +5,65 @@ import Foundation, { HandledProps } from "../foundation";
...
-
-        if (!srcset) {
-            return null;
-        }
-
-        const minWidthString: string = minWidth === 0 ? "0" : `${minWidth}px`;
-
-        return <source srcSet={srcset} media={`(min-width: ${minWidthString})`} />;
+    private renderChildrenBySlot(slot: ImageSlot): Array<React.ReactElement<HTMLSourceElement>> {
+        return React.Children.map(this.props.children, (child: React.ReactElement<HTMLSourceElement>, index: number) => {
```
nicholasrice: Your callback doesn't have a return type - not sure why it is passing tslint but it shouldn't be

chrisdholt: Added - had to update formatting as well due to line length.

packages/fast-components-react-base/src/image/image.tsx
nalogiudice: I guess I was seeing it in the snapshots

chrisdholt: That's due to our jest version. they pull all null || undefined types in a later version, but we have not updated.

chrisdholt:
changing title again. This is a breaking change and more of a major feature than it is a chore.

PR #924, opened 2018-09-19
chore(devops): update build scripts with better formatting

### LINKED ISSUE #923: Clean up E2E test configurations
Comments:

`build/testing/config-browsers.js`
- `Phases` can be singular `Phase`

`build/testing/sauce-labs/test-browsers.js`
- use backticks for remoteHub string
- file has too many extra lines
- remove the console log if unnecessary
- indenting is off in several places
- navigate misspelling
- use multi-line commenting where necessary
- use backticks where domain is concatted
- remove conventional changelogs from package.json scripts, this is not installed or used

janechu:
Closes #923


build/testing/config-browsers.js
```diff
@@ -34,293 +202,56 @@ exports.Phases = Phases;
...
-                        'platformName': 'Android',
-                        'platformVersion': '6.0',
-                        'deviceOrientation': 'PORTRAIT',
-                        'deviceName': 'Android Emulator',
-                        'appiumVersion': '1.8.1',
-                        'browserName': 'Chrome',
-                        'language': 'en',
-                        'locale': 'en_US'
-                    }
+                    Windows10_MicrosoftEdge,
```
awentzel: I don't see the advantages of this set of changes on the configuration.  These configurations will change quarterly and are provided by platform configurators of the service.  Your solution feels less maintainable and certainly isn't as friendly to read.  Now if I go into the phases to see what's being tested I have to find the const, look at it's properties, and return back to the switch.  It's better in my opinion to keep it as it is.

awentzel: What maybe better is to just create three separate configuration files which get imported into a const object that can hold the entire collection.  I can see where you eliminated some duplication by using constants, but, I think in this case duplication is ok given how this is used.  Otherwise, your const name should be generated by every property of the array or details get lost.  Like resolution, versions, etc.

janechu: There is a lot of duplication between the alpha/beta/releases, that's why. This helps solve that, now if you want to update, you can just remove the named config and the variable assigned to that config. Otherwise it's very messy to try and find the right items. Opening this up for comment thought, @chrisdholt @nicholasrice thoughts?

package.json
```diff
@@ -13,7 +13,6 @@
     "url": "git+https://github.com/Microsoft/fast-dna.git"
   },
   "scripts": {
-    "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s -r 0",
```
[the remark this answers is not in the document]
janechu: The last PR with the E2E changes also added this back in so I am removing it.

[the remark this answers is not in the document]
nicholasrice: might need to rebase

build/testing/config-browsers.js
janechu: That's true, why three configurations? If its one per phase then we have the same duplication problem. I think it might be better to go with your suggestion on the naming convention of the variable to make it more precise.

nicholasrice: If it's duplicated, then abstract it - that's the rule I try to follow. We could also implement a "common" array if they are shared across all phases.

janechu: To that end, if these configurations are provided by a service, can we not pull from that service?

awentzel: I think it's unique in the way it's used.  When building out the matrix you want to see everything together because it's easier to make sure you have the right combination of all the different properties.

One idea may be to separate configuration files.  config-browsers.alpha.js which has min requirements, then config-browsers.beta.js which imports the alpha config and appends to the beta config.  Then the config-browsers.release.js which imports beta config and appends the release ones.  This way it's not duplicate, still maintainable, and I think might check the collective boxes we're trying to achieve?

PR #929, opened 2018-09-20
chore(devops): remove unneeded ts config rules

awentzel:
the tsconfig.js file included rules that don't apply to the root of the Lerna monorepository and so these were removed.  They were also causing lint errors in vscode depending on strictness of type script plugins.


nalogiudice:
Can you update the title? It seems to have been cut off

tsconfig.json
```diff
@@ -4,12 +4,8 @@
         "module": "ES6",
```
janechu: It's actually confusing to me why each individual package has a tslint.json but some of them are referring (in webpack for example in fast-components-react-msft) to the tslint.json at root level, same with the tsconfig which I'm not sure is referred to anywhere so could it be removed? Does anyone know why we are doing this?

chrisdholt: I don't recall a good, intentional reason for this. It *could* be due to the fact that previously there were some packages that required significant tslint updates to conform to the rules of other packages and we didn't have the ability to do so at the time?

awentzel: I don't think NPM packages should be referring to configs from root or other packages.  Certainly, feels like a gap we should cleanup.  But, I'm not aware myself if there was a purpose so we'll need to get others to chime in for sure.  @chrisdholt @nicholasrice

nicholasrice: I don't know why that webpack file is referencing a root tslint.json. It shouldn't be.

we have a tslint package that all packages should inherit from - this may have been holdover from when we didn't have that package in the lerna repo.

I'm not aware of any tsconfigs that reference each other because that is a newly added feature to TypeScript - what are you referring too?

+1 for not depending upon configs form other packages unless there are idiomatic ways to accomplish that - eg the "extends" field of tslint

janechu: Looks like there is an extends for tsconfig, so we should use that.

chrisdholt: Updating with comment - after a sync offline, we should just remove the tsconfig from the root altogether.

PR #935, opened 2018-09-21
chore(fast-form-generator-react): update the api for react children in json schemas

### LINKED ISSUE #807: Refactor the component tracker

janechu:
Resolves #807


packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -290,3 +335,205 @@ export function getBreadcrumbs(
...
+                })
+            && !Array.isArray(get(data, dataLocation))
+        ) {
+            dataLocationsOfChildren.push(dataLocation);
+        }
+    });
+
+    // for every child location get nested data locations of children
+    dataLocationsOfChildren.forEach((dataLocationOfChildren: string) => {
+        const dataLocation: string = `${dataLocationOfChildren}.props`;
```
nicholasrice: why does `getSchemaByDataLocation` not accept your `dataLocation` variable? Is this naming accurate?

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -290,3 +335,205 @@ export function getBreadcrumbs(
...
+/**
+ * Finds the schema using the data location
+ */
+export function getSchemaByDataLocation(currentSchema: any, data: any, dataLocation: string, childOptions: IChildOptionItem[]): any {
+    if (dataLocation === "") {
+        return currentSchema;
+    }
+
+    const subData: any = get(data, dataLocation);
+    const id: string = subData ? subData.id : void(0);
```
nicholasrice: nit: `string | undefined`

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -290,3 +335,205 @@ export function getBreadcrumbs(
...
+    let schema: any = currentSchema;
+
+    childOptions.forEach((childOption: IChildOptionItem) => {
```
nicholasrice: consider refactoring this to use `array.find`

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -290,3 +335,205 @@ export function getBreadcrumbs(
...
+    const childOption: IChildOptionItem = getChildOptionBySchemaId(id, childOptions);
+
+    return childOption ? childOption.component : null;
+}
+
+/**
+ * Finds the child option using the schema id
+ */
+export function getChildOptionBySchemaId(id: string, childOptions: IChildOptionItem[]): IChildOptionItem | null {
+    let childOptionBySchemaId: any = null;
```
nicholasrice: same here - also consider refactoring to use `array.find`

packages/fast-form-generator-react/src/form/form.utilities.ts
```diff
@@ -290,3 +335,205 @@ export function getBreadcrumbs(
...
+    });
+
+    return childOptionBySchemaId;
+}
+
+/**
+ * Finds a subset of locations that are react children
+ */
+export function getReactChildrenLocationsFromSchema(schema: any, schemaLocations: any): string[] {
+    const locationsContainingReactProperties: string[] = schemaLocations.filter((schemaLocation: string): boolean => {
```
nicholasrice: Consider chaining `filter` and `forEach` so it reads as a cohesive statement.

packages/fast-form-generator-react/src/form/form.tsx
```diff
@@ -216,7 +216,7 @@ class Form extends React.Component<IFormProps & IManagedClasses<IFormClassNameCo
         return <FormSection {...sectionProps} />;
     }
 
-    private handleOnChange = (location: string, data: any, isArray: boolean, index: number): void => {
+    private handleOnChange = (location: string, data: any, isArray: boolean, index: number, isChildren?: boolean): void => {
```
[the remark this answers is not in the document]
janechu: The only way to know is by looking at the schema, which would get very complicated because of children nested (the same issue happened with arrays), I would need to pass in a sub-data location instead of the current location and pass up the schema and then look up the JSON schema type, so there would be the exact same number of arguments passed.

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -155,19 +182,23 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
...
+                this.getReactComponents(currentChildrenArray, item),
+                undefined,
+                undefined,
+                true
             );
         } else if (typeof item === "string") {
             this.props.onChange(
-                dataLocation,
-                this.getStringComponents(currentChildrenArray, item)
+                this.props.dataLocation,
```
nicholasrice: What are your thoughts on making `onChange` receive an object instead of a big list of arguments? I think that would be more readable and prevent you from needing to pass `undefined` all the time.

If that seems reasonable but not in scope for this change I can file a separate issue.

packages/fast-form-generator-react/src/form/form-item.children.tsx
```diff
@@ -189,65 +220,56 @@ class FormItemChildren extends React.Component<IFormItemChildrenProps & IManaged
...
     }
 
-    private getChildComponent(item: any): JSX.Element {
-        return (
-            <item.component
-                key={uniqueId()}
-                {...generateExampleData(item.schema, "")}
-            />
-        );
+    private getChildComponent(item: any): IComponentData {
```
nicholasrice: is `item` really any?

packages/fast-form-generator-react/src/form/form.tsx
```diff
@@ -241,8 +241,8 @@ class Form extends React.Component<IFormProps & IManagedClasses<IFormClassNameCo
             }
         }
 
-        if (location.length >= 8 && location.slice(location.length - 8, location.length) === "children") {
-            const children: JSX.Element = get(obj, location);
+        if (isChildren) {
+            const children: any = get(obj, location);
```
chrisdholt: Is this actually "any"?

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -165,7 +165,7 @@ export function applyHeaderStyle(): ICSSRules<{}> {
 
 export function applyGlobalStyle(): ICSSRules<{}> {
     return {
-        "body > li[draggable=\"true\"]": {
+        "body > div > li[draggable=\"true\"]": {
```
chrisdholt: Can't find where the markup changes are, but `<li>` is technically not a valid list unless contained within a `<ul>`.

janechu: it's still contained within a ul when it's not being dragged.

packages/fast-form-generator-react/src/form/form-item.children.tsx
janechu: It could be a config object sure, let's create another issue to consolidate.

packages/fast-form-generator-react/src/form/form-item.children.tsx
janechu: what it is is either a string or an interface with an id which is a string and props which is any so I can type it as that.

packages/fast-form-generator-react/src/form/form.tsx
janechu: see reply to rice, I'll update to be more precise

packages/fast-form-generator-react/src/utilities/form-input.style.ts
chrisdholt: Gotcha - makes sense

packages/fast-components-react-base/app/components/container.tsx
```diff
@@ -0,0 +1,16 @@
...
+import * as React from "react";
+
+export interface IContainerProps {
+    slot: string;
+}
+
+export default class Container extends React.Component<IContainerProps, {}> {
+    public render(): JSX.Element {
+        return (
+            <div>
```
chrisdholt: We need to pass slot somewhere, don't we?

janechu:
@chrisdholt I only pulled those 3 that had issues, there is an open proposal to remove all of them which should be a separate task.

packages/fast-components-react-base/app/components/container.tsx
janechu: You do not need to pass it anywhere actually, since all of our getChildSlot methods do is look for the passed prop, not the prop on an element.

chrisdholt:
> @chrisdholt I only pulled those 3 that had issues, there is an open proposal to remove all of them which should be a separate task.

Sure - I just didn't see them having been removed, but I see them in there now. Disregard.

packages/fast-components-react-base/app/components/container.tsx
```diff
@@ -0,0 +1,16 @@
+
+import * as React from "react";
+
+export interface IContainerProps {
+    slot: string;
```
nicholasrice: should slot be an enum?

PR #937, opened 2018-09-21
feat: add slot-selection support to foundation

### LINKED ISSUE #916: Foundation component should provide a utility for retrieving children by slot
Given some slot, the Foundation component should retrieve all child elements that match that slot. Optionally, a set of children  could be provided when looking for slot items outside of `this.props.children`

nicholasrice:
closes #916

### DIFF
--- a/packages/fast-components-react-base/src/foundation/foundation.spec.tsx
+++ b/packages/fast-components-react-base/src/foundation/foundation.spec.tsx
@@ -286,14 +286,14 @@ describe("generateClassNames", () => {
 });
 
 describe("withSlot", () => {
-    class BySlotTestComponent extends Foundation<{}, {}, {}> {
+    class WithSlotTestComponent extends Foundation<{}, {}, {}> {
         public render(): React.ReactNode {
             return this.props.children || "hello world";
         }
     }
 
     test("should not throw if no props are passed", () => {
-        const Component: BySlotTestComponent = new ReactTestUtils.renderIntoDocument(<BySlotTestComponent />);
+        const Component: WithSlotTestComponent = new ReactTestUtils.renderIntoDocument(<WithSlotTestComponent />);
 
         expect(() => {
             Component["withSlot"]("test");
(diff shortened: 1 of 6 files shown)


packages/fast-components-react-base/src/foundation/foundation.spec.tsx
```diff
@@ -284,3 +284,147 @@ describe("generateClassNames", () => {
...
+
+    test("should not throw if no props are passed", () => {
+        const Component: WithSlotTestComponent = new ReactTestUtils.renderIntoDocument(<WithSlotTestComponent />);
+
+        expect(() => {
+            Component["withSlot"]("test");
+        }).not.toThrow();
+    });
+
+    test("should return an empty array if slot no matches are found", () => {
```
janechu: "should return an empty array if there are no children with a slot prop matching the provided slot are found"

packages/fast-components-react-base/src/foundation/foundation.spec.tsx
```diff
@@ -284,3 +284,147 @@ describe("generateClassNames", () => {
...
+               <div slot="foo" />
+               <div slot="bar" />
+               <div slot="test" />
+            </WithSlotTestComponent>
+        );
+
+        expect(Component["withSlot"]("test")).toHaveLength(2);
+    });
+
+    test("should opperate on input nodes if they are provided", () => {
```
janechu: nit: *operate

packages/fast-components-react-base/src/foundation/foundation.spec.tsx
```diff
@@ -284,3 +284,147 @@ describe("generateClassNames", () => {
...
+
+    test("should not throw if no props are passed", () => {
+        const Component: WithoutSlotTestComponent = new ReactTestUtils.renderIntoDocument(<WithoutSlotTestComponent />);
+
+        expect(() => {
+            Component["withoutSlot"]("test");
+        }).not.toThrow();
+    });
+
+    test("should return all elements if slot no matches are found", () => {
```
janechu: "should return all children if none have a slot prop matching the provided slot"

packages/fast-components-react-base/src/foundation/foundation.spec.tsx
```diff
@@ -284,3 +284,147 @@ describe("generateClassNames", () => {
...
+               <div slot="foo" />
+               <div slot="bar" />
+               <div slot="test" />
+            </WithoutSlotTestComponent>
+        );
+
+        expect(Component["withoutSlot"]("test")).toHaveLength(2);
+    });
+
+    test("should opperate on input nodes if they are provided", () => {
```
janechu: nit: *operate

packages/fast-components-react-base/src/foundation/foundation.ts
```diff
@@ -121,6 +121,43 @@ class Foundation<H, U, S> extends React.Component<H & U & IFoundationProps, S> {
         return componentClasses.concat(` ${this.props.className || ""}`).trim().replace(/(\s){2,}/g, " ") || null;
     }
 
+    /*
+     * Return an array of all nodes who's slot prop matches the provided slot.
+     * If no nodes are provided, `this.props.children` will be used
```
janechu: should we also have the case that an assigned property may also contain children for examples `this.props.headerRegion`

nicholasrice: I don't follow

packages/fast-components-react-base/src/foundation/foundation.spec.tsx
nicholasrice: bah - confirmed. My spell check isn't working

chrisdholt:
@nicholasrice we should add a task to address anything added using slots in the release branch after this goes in and the release branch has been rebased with master.

packages/fast-components-react-base/src/foundation/foundation.ts
nicholasrice: discussed offline - this will accept any `React.ReactNode` - if `this.props.headerRegion` is a react node, then this will work

PR #938, opened 2018-09-21
chore: remove deprecated changelog

awentzel:
After the switch to Lerna --individual mode we now publish changelogs to the packages root not the root of the project.


janechu:
please update the title to include the part that has been cut off and moved (...ackages)

PR #949, opened 2018-09-24
fix: implement slot utility into image

### LINKED ISSUE #945: Update release-branch component's with `withSlot` utility

nicholasrice:
closes #945

### DIFF
--- a/packages/fast-components-react-base/src/image/image.tsx
+++ b/packages/fast-components-react-base/src/image/image.tsx
@@ -49,7 +49,7 @@ class Image extends Foundation<
                     {...this.unhandledProps()}
                     className={super.generateClassNames(className)}
                 >
-                    {this.renderChildrenBySlot(ImageSlot.source)}
+                    {this.withSlot(ImageSlot.source)}
                     <img
                         src={this.props.src}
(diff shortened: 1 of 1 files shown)


awentzel:
should the package name be added into the title?  feat(fast-components-react-base): "... ?

nicholasrice:
@awentzel I have been leaving those off my PRs - I don't think there is clear guidance on what should go here. For instance, should this be "fast-components-react-base" or "image"? My preference is to have descriptive titles and if I *really* want to know what the scope of the changes are I can go look at that in the commit diff.

awentzel:
@nicholasrice then the policy should be updated.  I think we've been using different rules as a team and we should be on the same page so we can provide clear guidance.  See wiki for current guidance and then we all should propose changes to improve it if it's not clear.  Like is the situation now.  The idea is that when reading changelogs it should be easy to spot packages you're working with and they should be grouped I think.

nicholasrice:
@awentzel I've always held that scopes are unnecessary and inherently blurry. Other contributors wanted them, and we landed on them being optional. I'm all for changing the guidance to remove scopes, but this PR currently is in line with our guidance.

PR #959, opened 2018-09-27
feat: update class-name contracts to be entirely optional

### LINKED ISSUE #951: Make managedClasses an optional property for all components
Discussion: #917 

As a developer, I should not be required to supply any class-name to a base-component that is not required for my implementation. This aligns with the opinionated goals of the FAST project.

**Acceptance criteria:**
- the `managedClasses` property should be optional for all components
- all properties of a class-name-contract should be optional

nicholasrice:
The managedClasses prop will be optional, as well as all properties of a class-name contract. It will be up to developer-discretion if class-names are required for their specific implementation.

closes #951

### DIFF
--- a/packages/fast-components-react-msft/src/metatext/metatext.spec.tsx
+++ b/packages/fast-components-react-msft/src/metatext/metatext.spec.tsx
@@ -26,6 +26,14 @@ describe("metatext", (): void => {
         expect((MSFTMetatext as any).name).toBe(MSFTMetatext.displayName);
     });
 
+    test("should not throw if managedClasses are not provided", () => {
+        expect(
+            () => {
+                shallow(<Metatext />);
+            }
+        ).not.toThrow();
+    });
+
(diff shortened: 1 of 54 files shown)


packages/fast-components-react-msft/src/metatext/metatext.spec.tsx
```diff
@@ -26,6 +26,14 @@ describe("metatext", (): void => {
         expect((MSFTMetatext as any).name).toBe(MSFTMetatext.displayName);
     });
 
+    test("should not throw if managedClasses are not provided", () => {
+        expect(
+            () => {
+                shallow(<Metatext />);
+            }
+        ).not.toThrow();
+    });
```
chrisdholt: This is a MSFT instance - are we sure that managed classes are NOT being provided here? We don't currently need to provide them to MSFT instances due to the work of the JSS manager.

nicholasrice: oops, no this should be an unmanaged instance

PR #969, opened 2018-09-29
fix: update label to always apply label class-name

### LINKED ISSUE #957: Label should always apply the "label" class if available
Label currently only applies `managedClasses.label` if the `hidden` prop is false. This class should always be applied if it exists, and the hidden should be applied *in addition* if hidden is true.

khamudom:
- set the class "label" to always show, even when hidden prop is true.

resolve #957

### DIFF
--- a/packages/fast-components-react-base/src/label/label.spec.tsx
+++ b/packages/fast-components-react-base/src/label/label.spec.tsx
@@ -79,4 +79,16 @@ describe("label", (): void => {
         expect(rendered.instance().props.tag).toEqual(LabelTag.legend);
         expect(rendered.type()).toBe(LabelTag.legend);
     });
+
+    test("should rendered the correct className base on hidden prop", () => {
+        const rendered: any = shallow(
+            <Label managedClasses={managedClasses} />
+        );
+        const renderedHidden: any = shallow(
+            <Label managedClasses={managedClasses} hidden={true} />
+        );
+
+        expect(rendered.hasClass("label-class")).toBe(true);
+        expect(renderedHidden.hasClass("label-class label-hidden-class")).toBe(true);
+    });
 });
--- a/packages/fast-components-react-base/src/label/label.spec.tsx
+++ b/packages/fast-components-react-base/src/label/label.spec.tsx
@@ -80,15 +80,19 @@ describe("label", (): void => {
         expect(rendered.type()).toBe(LabelTag.legend);
     });
 
-    test("should rendered the correct className base on hidden prop", () => {
+    test("should add the base className", () => {
         const rendered: any = shallow(
             <Label managedClasses={managedClasses} />
         );
+
+        expect(rendered.hasClass("label-class")).toBe(true);
+    });
+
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-msft/src/label/__snapshots__/label.spec.ts.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`label snapshots Label: 0 1`] = `
 <label
-  className="label-0-1-1"
+  className={null}
```
nalogiudice: @janechu & @nicholasrice should @khamudom remove snapshots?

packages/fast-components-react-base/src/label/label.tsx
```diff
@@ -52,7 +52,13 @@ class Label extends Foundation<
      * Generates class names based on props
      */
     protected generateClassNames(): string {
-        return super.generateClassNames(get(this.props, this.props.hidden ? "managedClasses.label_hidden" : "managedClasses.label"));
+        let className: string = get(this.props, "managedClasses.label");
```
nalogiudice: We should add a jest test to test classnames

packages/fast-components-react-msft/src/label/__snapshots__/label.spec.ts.snap
awentzel: I think there is another task to remove snapshots and It feels that would be the time to remove Label snapshots as well.

awentzel:
It feels like this should be part of the overall snapshots removal and not Khams PR?

packages/fast-components-react-base/src/label/label.tsx
khamudom: added

packages/fast-components-react-msft/src/label/__snapshots__/label.spec.ts.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`label snapshots Label: 0 1`] = `
 <label
-  className="label-0-1-1"
```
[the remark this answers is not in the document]
khamudom: it is no longer "null"

packages/fast-components-react-base/src/label/label.spec.tsx
```diff
@@ -79,4 +79,16 @@ describe("label", (): void => {
         expect(rendered.instance().props.tag).toEqual(LabelTag.legend);
         expect(rendered.type()).toBe(LabelTag.legend);
     });
+
+    test("should rendered the correct className base on hidden prop", () => {
```
janechu: this test naming structure, although we are doing something similar above, is wrong, tests should always pass therefore the word "correct" is redundant, I would rather see two tests, one when the hidden prop is not passed with the description "should add the base className" and when the hidden prop is passed "should add the hidden className and the base className when the hidden prop is passed"

khamudom: updated

PR #970, opened 2018-10-01
chore: use updated API for form generator and adjust small issue regarding recognition of strings as removable children

janechu:
This fixes some dev site issues for the component packages and adds more documentation for JSON schema writing relevant to the recent breaking changes.


packages/fast-components-react-msft/app/components/examples.data.tsx
```diff
@@ -0,0 +1,8 @@
+import schema from "./react-html-element.schema.json";
+import ReactHTMLElement from "./react-html-element";
```
awentzel: Why are some things imported in Hungarian notation and others in Camel casing?  
`import ReactHTMLElement` but in app.tsx it's imported as `import reactHTMLElementExamples` ?  Should these be made consistent?  I think this is a problem in many places, but, we should start cleaning up if others agree this is an issue.

chrisdholt: I think this is likely due to ReactHTMLElement being a class vs reactHTMLElementExamples being an exported constant.

janechu: Correct, classes are PascalCased, variables are camelCased

PR #976, opened 2018-10-01
fix: explicitly apply an additional active prop to Tab and TabPanel via the Tabs component instead of using unhandledProps

### LINKED ISSUE #958: Tab uses unhandledProps as part of it's code logic when it shouldn't
The tab component accesses the `aria-selected` property of unhandled props to descide if it should be active or not. This should instead be handled by a handled "active" (or some similar name) and should not have explicit behavior based on unhandledProps

janechu:
Resolves #958

### DIFF
--- a/packages/fast-components-react-base/src/tabs/tab-panel.tsx
+++ b/packages/fast-components-react-base/src/tabs/tab-panel.tsx
@@ -2,14 +2,14 @@ import * as React from "react";
 import { get } from "lodash-es";
 import Foundation, { HandledProps } from "../foundation";
 import { IManagedClasses, ITabPanelClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
-import { ITabPanelHandledProps, ITabPanelManagedClasses, ITabPanelUnhandledProps } from "./tab-panel.props";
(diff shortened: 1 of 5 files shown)


janechu:
@chrisdholt these tests are already in place.

packages/fast-components-react-base/src/tabs/tab.tsx
```diff
@@ -9,10 +9,15 @@ class Tab extends Foundation<
     ITabUnhandledProps,
     {}
 > {
+    public static defaultProps: Partial<ITabHandledProps> = {
```
nicholasrice: `defaultProps` should be `Partial<TabProps>` - defaultProps are always a partial of the comprehensive prop object.

packages/fast-components-react-base/src/tabs/tab-panel.tsx
```diff
@@ -9,10 +9,15 @@ class TabPanel extends Foundation<
     ITabPanelUnhandledProps,
     {}
 > {
+    public static defaultProps: Partial<ITabPanelHandledProps> = {
```
nicholasrice: `Partial<TabPanelProps>`

PR #984, opened 2018-10-03
feat: add card as a new component

### LINKED ISSUE #992: Add "Card" as a new component
*Cards* are snapshots of content that are typically used in a group to present collections of related information. *Cards* can represent any quickly consumable information.  The *card* component can be either a [div](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div), [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article), or [section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) element.

- Add a card base component that can be styled for various design languages
- The card base component should accept children
- The card base is a compositional element to create various types of cards

In the MSFT package, the base card will include a small set of styles and an elevation/depth based on the current design language.

chrisdholt:
<body>
closes #992 
<footer>
 


packages/fast-components-react-base/src/card/card.tsx
```diff
@@ -0,0 +1,54 @@
...
+    CardHTMLTags,
+    ICardHandledProps,
+    ICardManagedClasses,
+    ICardUnhandledProps
+} from "./card.props";
+import { ICardClassNameContract, IManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+
+/* tslint:disable-next-line */
+class Card extends Foundation<ICardHandledProps & ICardManagedClasses,  ICardUnhandledProps, {}> {
+    public static displayName: string = "Card";
```
awentzel: Should the public property for `displayName` be `readonly` of the class?

packages/fast-components-react-msft/src/card/README.md
```diff
@@ -0,0 +1,10 @@
+# Card
+*Cards* are snapshots of content that are typically used in a group to present collections of related information. *Cards* can represent any quickly consumable information.  The *card* component can be either a [div](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div), [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article), or [section](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) element.
+
+## Usage
+Use *cards* when displaying a feed of content or when a stronger isolation between each content items is needed. *Cards* are also well suited for scenarios where the content is not consistent and requires different fields, image sizes, or content, which is often the case in a feed. *Cards* create an opportunity to present a collection of related content in a rich and dynamic way. *Cards* offer flexible layout options that can be used interchangeably with cards using other options; often a requirement when the content available is not consistent.
```
nalogiudice: "*Cards* offer flexible layout options... " Is this true in this PR? I don't see anything with that functionality. I would say we should take it out if we aren't offering it.

packages/fast-components-react-base/src/card/card.tsx
chrisdholt: I don't think it's necessary, but @nicholasrice can chime in. This matches what we are doing elsewhere, so I would prefer to make that change (if we decide to) in a new PR.

nicholasrice: I don't see a reason for it to be `readonly`. If folks want to re-assign it for some reason it doesn't matter to us. For instance, if they already have a `Card` component they may wan't to rename this `FASTCard` for debugging purposes. Probably unlikely, but I don't see a reason to stop them.

packages/fast-components-react-base/src/card/card.props.ts
```diff
@@ -0,0 +1,27 @@
...
+    section = "section"
+}
+
+export interface ICardUnhandledProps extends React.HTMLAttributes<HTMLDivElement | HTMLElement> {}
+export interface ICardManagedClasses extends IManagedClasses<ICardClassNameContract> {}
+export interface ICardHandledProps {
+    /**
+     * The card children
+     */
+    children?: React.ReactNode;
```
janechu: I believe Foundation already has children specified as a handledProp

packages/fast-components-styles-msft/src/card/index.ts
```diff
@@ -0,0 +1,25 @@
...
+import { get } from "lodash-es";
+import Chroma from "chroma-js";
+import { elevation, ElevationMultiplier } from "../utilities/elevation";
+
+const styles: ComponentStyles<ICardClassNameContract, IDesignSystem> = (
+    config: IDesignSystem
+): ComponentStyleSheet<ICardClassNameContract, IDesignSystem> => {
+    const designSystem: IDesignSystem = withDesignSystemDefaults(config);
+    const foregroundColor: string = designSystem.foregroundColor;
+    const backgroundColor: string = designSystem.backgroundColor;
```
janechu: nit: not sure it's necessary to declare these since they are only used once

packages/fast-components-react-base/src/card/card.props.ts
chrisdholt: We enumerate in the prop contract because it's possible certain components may not take children (text field base component), but we don't have to enumerate as part of the handledProp value inside the class for Card because if children are passed we never want to pass it to the DOM.

packages/fast-components-styles-msft/src/card/index.ts
chrisdholt: will update

packages/fast-components-react-base/src/card/card.tsx
awentzel: That's seems like a valid use case.  disregard.

PR #990, opened 2018-10-04
fix: apply checkbox `indeterminate` property regardless of "checked" state

### LINKED ISSUE #968: Checkbox `indeterminate` logic is incorrect
Checkbox only supports setting of the indeterminate property on *checked* checkboxes. This is incorrect.

From MDN:
> In addition to the checked and unchecked states, there is a third state a checkbox can be in: indeterminate. This is a state in which it's impossible to say whether the item is toggled on or off.

**More from MDN here:** https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

**From the W3C Spec:** https://www.w3.org/TR/html5/sec-forms.html#dom-htmlinputelement-indeterminate

Mrigank11:
As per the issue, we should apply `indeterminate` property irrespective of `checked` state.

Fixes #968

### DIFF
--- a/packages/fast-components-react-base/src/checkbox/checkbox.tsx
+++ b/packages/fast-components-react-base/src/checkbox/checkbox.tsx
@@ -136,7 +136,7 @@ class Checkbox extends Foundation<
      * This method should be called after render because it relies on element references.
      */
     private applyIndeterminateState(): void {
-         if (this.props.indeterminate && this.inputRef.current && this.state.checked) {
+         if (this.props.indeterminate && this.inputRef.current) {
             this.inputRef.current.indeterminate = this.props.indeterminate;
         }
     }


packages/fast-components-react-base/src/checkbox/checkbox.tsx
```diff
@@ -136,7 +136,7 @@ class Checkbox extends Foundation<
      * This method should be called after render because it relies on element references.
      */
     private applyIndeterminateState(): void {
-         if (this.props.indeterminate && this.inputRef.current && this.state.checked) {
+         if (this.props.indeterminate && this.inputRef.current) {
```
chrisdholt: @Mrigank11 thanks for the PR! I think we'll also need some styles added as part of this PR. For instance, if an indeterminate checkbox *is* checked, we'll want to make sure that the indeterminate style is the visual treatment. Let me know if you need a rundown on where to find and make that change.

Mrigank11: That'll be helpful, please point me to the right place.

chrisdholt: In the repo, we have a package called `fast-components-styles-msft` where the checkbox code lives. You can make the changes in that file. You'll need to make your changes in the styles and then run `lerna bootstrap` in order to sync and symlink the packages.

From a "how" this should come together, I would suggest extracting the "indeterminate" styles into a new function so they can be applied in two places (you can see an example in the `hypertext` file). 

If you need any additional help or questions answered, join us on [discord](https://discord.gg/FcSNfg4) and we'll be happy to help you!

chrisdholt: Resolving as this is currently working as expected.

nalogiudice: @chrisdholt I originally created this component and decided on the behavior. After discussions with the team, we decided that the indeterminate prop would be valid only when checked since it is sort of a gray area, and you could really go either way. I'm ok with this change, but I wanted to give some context.

chrisdholt: I added some links in the issues attached here; ultimately indeterminate does not require checked. There are some great examples in the links provided to go deeper into why I think this change is necessary. We can't actually support all valid indeterminate scenarios if we bind it to a state. Indeterminate is simply a visual state and not a functional state, so I think binding to a functional state will prevent us from accomplishing our goals.

nalogiudice: Sure, I support this.

Issue #258, opened 2018-04-13
Refactor React JSS manager to use the latest lifecycle methods.


chrisdholt:
With react 16.3.x comes a few changes to the component lifecycle api. The following API's are now considered "unsafe" and should be swapped for their new safe counterparts:

UNSAFE_componentWillMount()
UNSAFE_componentWillReceiveProps()
UNSAFE_componentWillUpdate()

Issue #175, opened 2018-03-30
Update React JSS manager to use new context API


nicholasrice:
https://reactjs.org/docs/context.html

awentzel:
closes #236 which requests upgrading React package

nicholasrice:
This work should also address #258

Issue #142, opened 2018-03-27
Write additional JSS manager tests


janechu:
Look at the TODO items in JSS Manager (React and Angular) and write tests for these methods.

nicholasrice:
Good tips for this https://medium.com/@stipsan/testing-with-jest-15-awesome-tips-and-tricks-42150ec4c262

PR #993, opened 2018-10-05
feat: update JSSManager context API

### LINKED ISSUE #856: JSS Manager should render with a debug-friendly display-name
If the wrapped component has a display-name, we the display-name of the JSS manager should be `withJssManager(ComponentDisplayName)`

https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

nicholasrice:
Updates JSSManager to use React 16.3 context API and removes deprecated life-cycle hooks.

closes #856 #258 #175 #142

### DIFF
--- a/packages/fast-components-react-msft/src/button/button.spec.tsx
+++ b/packages/fast-components-react-msft/src/button/button.spec.tsx
@@ -58,7 +58,7 @@ describe("button", (): void => {
         ).not.toThrow();
     });
 
-    test("should return an object that includes all valid props which are not enumerated as handledProps", () => {
+    test("should accept unhandledProps", () => {
         const handledProps: IButtonHandledProps = {
             href
         };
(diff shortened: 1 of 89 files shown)


packages/fast-jss-manager-react/src/design-system-provider.tsx
```diff
@@ -6,36 +6,31 @@
...
-    public getChildContext(): IDesignSystemProviderProps<T> {
-        const designSystem: T = this.props.designSystem;
-        const contextualDesignSystem: T = this.context.designSystem;
-        const context: T = Object.assign({}, contextualDesignSystem, designSystem);
-
-        return { designSystem: context };
+export class DesignSystemProvider<T> extends React.Component<IDesignSystemProviderProps<T>, {}> {
+    public render(): React.ReactNode {
+        return (
+            <Consumer>
```
chrisdholt: Can you help me understand this "outer" `<Consumer />`? 

I'm probably missing something, I'm only aware of the pattern from the docs of
```
<Provider>
    <Consumer>
</Provider>
```

Is our React output more similar to below?
```
<Consumer>
    <Provider>
        <Consumer>
               // Child / Children
        </Consumer>
    </Provider>
</Consumer>
```

nicholasrice: The outer consumer is to consume the parent context. As far as I can tell, there is no way for a provider to consume context context values, so in order to inherit context values we need to first render a consumer, which *then* gives the provider the parent context to merge with its own props.

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -186,14 +128,13 @@ describe("button", (): void => {
         };
 
         const rendered: any = mount(
-            <Button {...props} />
+            <MSFTButton {...props} />
```
nalogiudice: Just curious why these are MSFT buttons?

chrisdholt: @nalogiudice I think this was testing the unstyled vs styled instance (of the MSFT package button).

nicholasrice: Test doesn't have anything to do with the managed component. We're testing the capabilities of the Microsoft button, not a *styled* Microsoft button.

packages/fast-jss-manager-react/src/design-system-provider.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    test("should allow partial updates to context values", (): void => {
+        function render(value: { a: string, b: string }): string {
+            return `${value.a}, ${value.b}`;
+        }
+
+        const tree: ShallowWrapper = mount(
+            <DesignSystemProvider designSystem={{a: "a", b: "b"}}>
+                <DesignSystemProvider designSystem={{a: "A"}}>
```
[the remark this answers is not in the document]
nicholasrice: @janechu updated

PR #995, opened 2018-10-05
feat: update checkbox to use slots to create label

nicholasrice:


packages/fast-components-react-base/src/checkbox/checkbox.schema.json
```diff
@@ -6,15 +6,6 @@
     "id": "checkbox",
     "children": true,
     "properties": {
```
nicholasrice: @janechu what does the schema need to look like to enable label's to use slots? I tried to reference use button as a reference but then the label didn't show up at all.

janechu: gotcha, so what you'd need is to add children to `reactProperties` instead of `children: true` which no longer works. Then refer to the ReactHTMLElement component as one of the permittable `ids` and that should have a slot prop for a schema.

PR #999, opened 2018-10-06
chore: fix the flipper examples to use the correct property name

janechu:
Clean up chore for the flipper examples.


nicholasrice:
You probably need to update snapshots

PR #1000, opened 2018-10-06
feat: add a utility that converts hex values to rgb or rgba

### LINKED ISSUE #639: Add a utility to convert hex values to rgb or rgba

RamIyer1998:
Made a new file called `rgb.ts` along with 50 tests to accompany it. I followed the format set by @chrisdholt in [commit b4f5617](https://github.com/Microsoft/fast-dna/pull/640). Depending on the length, the function will return either a string in the format `rgb(r, g, b)` (3 or 6 length hexits) or `rgba(r, g, b, a)` (4 or 8 length hexits). My test suite has ten tests each for 3, 4, 6, and 8 length hexits along with ten additional tests to make sure invalid hexits throw an error. 

Closes #639 
 


packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+  if (g.length === 1) {
+    g += g;
+  }
+  if (b.length === 1) {
+    b += b;
+  }
+
+  if (rgb) {
+    return "rgb(" + parseInt(r, 16) + ", " + parseInt(g, 16) + ", " + parseInt(b, 16) + ")";
```
janechu: for these kinds of concatenated strings we use template syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,57 @@
...
+  } else if (rgba) {
+    if (a.length === 1) {
+      a += a;
+    }
+    return "rgba(" + parseInt(r, 16) + ", " + parseInt(g, 16) + ", " + parseInt(b, 16) + ", " + (parseInt(a, 16) / 255) + ")";
+  }
+
```
janechu: nit: extra line

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,57 @@
...
+export function hexToRgb(hex: string): string {
```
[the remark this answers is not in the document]
janechu: FYI cognitive complexity arises when a single function has a lot of logic in it, here you have a lot of if/else, I would try to separate these out into simpler functions like `getHexLength` as an example.

packages/fast-colors/src/rgb.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+    });
+});
+
+describe("Invalid hexits", (): void => {
+    test("Should throw an error", (): void => {
+        expect((): void => {
+            hexToRgb("");
+        }).toThrow();
+    });
+    test("Should throw an error", (): void => {
```
janechu: I think these could be more specific and collected into groupings eg. "should throw an error if a non-alphanumeric value is used", "should throw if the alphanumeric values provided are out of range", you can add multiple expect to each test where it makes sense.

packages/fast-colors/src/rgb.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+        expect(hexToRgb("#D47C8A")).toBe("rgb(212, 124, 138)");
+    });
+    test("Should return rgb(188, 218, 158)", (): void => {
+        expect(hexToRgb("#BCDA9E")).toBe("rgb(188, 218, 158)");
+    });
+});
+
+describe("Valid 8 digit hexits", (): void => {
+    test("Should return rgba(255, 255, 255, 1)", (): void => {
+        expect(hexToRgb("#FFFFFFFF")).toBe("rgba(255, 255, 255, 1)");
```
janechu: I think you could combine these and try more of a grouping strategy, if it's wrong we will see the expected value is not being met so it's not super necessary to put the expected value in the test stdout string, you could group some like minded tests with "should convert when all alphabetical characters in provided hex are uppercase", "should convert when alphabetical characters in provided hex are a mix of upper and lower case", "should convert if provided hex is missing a '#' symbol"

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,90 @@
...
+    return {
+      r: hex.substring(0, 2),
+      g: hex.substring(2, 4),
+      b: hex.substring(4, 6),
+      a: hex.substring(6, 8),
+      rgb: false,
+      rgba: true
+    };
+  } else {
+    throw new Error("Not a convertible hex value");
```
nicholasrice: What are your thoughts on having the error message contain the input `hex`? I think it would helpful at a glance to know what the input value is.

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,90 @@
...
+interface IRgbJSON {
+  r: string;
+  g: string;
+  b: string;
+  a: string;
+  rgb: boolean;
+  rgba: boolean;
+}
+
+function properRGBValues(hex: string): IRgbJSON {
```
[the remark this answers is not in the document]
nicholasrice: To circumvent this, you could split the definition into two smaller functions: `processRGBHex` and `processRGBA` hex. You know that if length is 3 or 6, you need to process RGB, and than when length is 4 or 8 you need to process RGBA.

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,90 @@
+
+export function hexToRgb(hex: string): string {
```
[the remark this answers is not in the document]
nicholasrice: To fix this issue, you could create a `normalizeRGBValues` function that ensures all rgb values have a length of two.

nicholasrice:
Thanks for the PR @RamIyer1998! I've left a few notes on the code-climate issues that should get this PR passing the build.

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,90 @@
...
+    return `rgb(${parseInt(r, 16)}, ${parseInt(g, 16)}, ${parseInt(b, 16)})`;
+  } else if (rgba) {
+    if (a.length === 1) {
+      a += a;
+    }
+    return `rgba(${parseInt(r, 16)}, ${parseInt(g, 16)}, ${parseInt(b, 16)}, ${(parseInt(a, 16) / 255)})`;
+  }
+}
+
+interface IRgbJSON {
```
nicholasrice: We've also changed our naming convention for interfaces - would you mind removing the `I` from the interface name?

packages/fast-colors/src/rgb.ts
```diff
@@ -0,0 +1,90 @@
+
+export function hexToRgb(hex: string): string {
+  let processedHex: IRgbJSON;
+  processedHex = properRGBValues(hex);
```
chrisdholt: I'd suggest defining `processedHex` as a const and assigning it to the return value of the function here since we are changing it anywhere else. You could likely do the same with everything below.

packages/fast-colors/src/rgb.ts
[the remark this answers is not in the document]
chrisdholt: +1 for splitting this into multiple functions for simplicity

packages/fast-colors/src/rgb.ts
nalogiudice: +1

packages/fast-colors/src/rgb.ts
RamIyer1998: Hi Nicholas. It seems that if I remove the `I` from the name, I get an error. I've synced with upstream. Should I still remove it?

RamIyer1998:
Will open a new PR.

Issue #774, opened 2018-08-10
fast-components-react-msft are added last in the cascading order, causing the need for !important in JSS


nalogiudice:
**Is your feature request related to a problem? Please describe.**
When using fast-components-react-msft (and the full fast JSS stack), the msft-styles are added lastly, causing the need for '!important' to be used to override the base msft styles.

**Describe the solution you'd like**
The msft styles should be added first, allowing the additional styles to be added last and therefore no longer needing '!important'

**Example of issue**
```
import { Heading, Typography } from '@microsoft/fast-components-react-msft';
...
<Heading tag={HeadingTag.h1} level={HeadingLevel._2} children={'foo'} className={this.props.managedClasses.landingPage_heading} />
...
(in JSS - !important needed)
landingPage_heading: {
    fontWeight: '300 !important',
    marginBottom: `${toPx(40)} !important`
}
```

chrisdholt:
I think this may be an issue with the JSS manager. 

An immediate fix to this problem would be to use the jssStylesheet prop for the MSFT styles on the component.

Example:

```jsx
const headingStyles: Partial<IHeadingClassNameContract, IDesignSystem> = {
    heading_4: {
        fontWeight: "300",
        marginBottom: toPx(40)
    }
};

<Heading level={HeadingLevel._4} jssStylesheet={headingStyles}>
    My heading
</Heading>
```

PR #1006, opened 2018-10-09
fix: reintroduce componentDidMount in JSS manager to address incidental regressions

chrisdholt:
<body>

<footer>
 

### DIFF
--- a/packages/fast-jss-manager-react/src/jss-manager.tsx
+++ b/packages/fast-jss-manager-react/src/jss-manager.tsx
@@ -72,13 +72,18 @@ export class JSSManager<S, C> extends React.Component<JSSManagerProps<S, C>, JSS
         if (Boolean(props.styles)) {
             state.styleSheet = this.createStyleSheet();
             state.styleSheet.attach();
-
-            // It appears we need to update the stylesheet for any style properties defined as functions
-            // to work.
-            state.styleSheet.update(props.designSystem);
         }
 
         this.state = state;
+    }
+
+    /**
+     * TODO #774: Remove lifecycle event with fix for managing stylesheet registry
+     */
+    public componentDidMount(): void {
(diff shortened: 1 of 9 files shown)


nalogiudice:
Can you link a task/bug so we can understand why this was a regression?

chrisdholt:
> Can you link a task/bug so we can understand why this was a regression?

@nalogiudice I'll provide what detail I can here:

Previously we had a lifecycle method that was being called in `componentWillMount` which is unsafe. In moving to the new lifecycle events, React suggests that you either use the constructor, or use `componentDidMount`.

We moved the logic to update stylesheets to the constructor. Our goal was to ensure that users didn't have a flash of unstyled content (as opposed to doing this work in `componentDidMount`). While helping some projects update to FAST-DNA v3.0 I noticed that styles were all of a sudden broken. While it worked before, all of a sudden issue #774 was suddenly repro'ing. Previously the bug was not manifesting in the repo where the work was being done, and now it was. 

Upon investigation, the cause of this was moving the logic from the lifecycle event to the constructor. The order of constructor vs mount is different. We don't have all the details, but what will likely fix this is the same thing that will resolve #774. For the time being, we can prevent the regression by adding the lifecycle event back in. This gives us the opportunity to come up with a "registry" to ensure that parent components with style overrides/updates/modifications carry more weight than their children.

packages/fast-jss-manager-react/src/jss-manager.tsx
```diff
@@ -81,6 +81,13 @@ export class JSSManager<S, C> extends React.Component<JSSManagerProps<S, C>, JSS
         this.state = state;
     }
 
+    /**
```
nicholasrice: You can probably comment out line 78 also - that is where the update cycle got moved to.

nalogiudice:
Thanks for the explanation @chrisdholt

PR #1010, opened 2018-10-09
feat: add breakpoint tracker as a new package

### LINKED ISSUE #898: Breakpoint tracker should export as a non-instantiated instance of the class
Currently breakpoint tracker is exported as a singleton. We should simply export the class and allow the user to import it however they want. This would allow for multiple trackers to exist if a user wanted to do so. This also would allow breakpoints to be passed at the class level rather than at the subscribe level as currently exists.

chrisdholt:
<body>
closes #898 
<footer>
 

### DIFF
--- a/packages/fast-breakpoint-tracker-react/package.json
+++ b/packages/fast-breakpoint-tracker-react/package.json
@@ -76,6 +76,7 @@
     "watch": "^1.0.2"
   },
   "peerDependencies": {
+    "exenv-es6": "^1.0.0",
     "lodash-es": "^4.0.0",
     "react": "^16.3.0"
   }
--- a/packages/fast-breakpoint-tracker-react/package.json
+++ b/packages/fast-breakpoint-tracker-react/package.json
@@ -70,7 +70,7 @@
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
     "react-test-renderer": "^16.2.0",
-    "ts-jest": "^22.0.4",
+    "ts-jest": "^23.10.4",
     "tslint": "^5.9.1",
     "typescript": "^3.0.1",
     "watch": "^1.0.2"
--- a/packages/fast-breakpoint-tracker-react/package.json
+++ b/packages/fast-breakpoint-tracker-react/package.json
@@ -2,7 +2,7 @@
   "name": "@microsoft/fast-breakpoint-tracker-react",
   "description": "React HOC for managing breakpoints",
   "sideEffects": false,
-  "version": "2.2.1",
+  "version": "1.0.0",
   "author": {
     "name": "Microsoft",
     "url": "https://discord.gg/FcSNfg4"
@@ -30,10 +30,10 @@
     "collectCoverage": true,
     "coverageThreshold": {
       "global": {
-        "statements": 98,
-        "branches": 66,
-        "functions": 100,
-        "lines": 98
+        "statements": 80,
+        "branches": 50,
+        "functions": 80,
+        "lines": 80
       }
     },
     "testURL": "http://localhost",
@@ -53,7 +53,7 @@
     "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$"
   },
   "devDependencies": {
-    "@microsoft/fast-tslint-rules": "^2.2.0",
+    "@microsoft/fast-tslint-rules": "^3.0.0",
     "@types/jest": "^22.1.2",
     "@types/lodash-es": "^4.17.0",
     "@types/node": "^9.4.6",
@@ -65,7 +65,6 @@
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "exenv-es6": "^1.0.0",
-    "hoist-non-react-statics": "^3.0.1",
     "jest": "^22.3.0",
     "lodash-es": "^4.0.0",
     "react": "^16.3.0",
@@ -77,7 +76,6 @@
     "watch": "^1.0.2"
   },
   "peerDependencies": {
-    "hoist-non-react-statics": "^3.0.1",
     "lodash-es": "^4.0.0",
     "react": "^16.3.0"
   }
--- a/packages/fast-breakpoint-tracker-react/package.json
+++ b/packages/fast-breakpoint-tracker-react/package.json
@@ -54,7 +54,7 @@
   },
   "devDependencies": {
     "@microsoft/fast-tslint-rules": "^3.0.0",
-    "@types/jest": "^22.1.2",
+    "@types/jest": "^23.0.0",
     "@types/lodash-es": "^4.17.0",
     "@types/node": "^9.4.6",
     "@types/react": "^16.3.6",
@@ -65,7 +65,7 @@
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "exenv-es6": "^1.0.0",
-    "jest": "^22.3.0",
+    "jest": "^23.0.0",
     "lodash-es": "^4.0.0",
     "react": "^16.3.0",
     "react-dom": "^16.3.0",
--- a/packages/fast-breakpoint-tracker-react/package.json
+++ b/packages/fast-breakpoint-tracker-react/package.json
@@ -1,0 +1,84 @@
+{
+  "name": "@microsoft/fast-breakpoint-tracker-react",
+  "description": "React HOC for managing breakpoints",
+  "sideEffects": false,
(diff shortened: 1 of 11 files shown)


packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        this.openRequestAnimationFrame = false;
+    }
+
+    /**
+     * Request's an animation frame if there are currently no open animation frame requests
+     */
+    private requestFrame = (): void => {
+        if (this.openRequestAnimationFrame) {
```
nicholasrice: I think you want to execute this `if` block when `this.openRequestAnimationFrame` is `false`, not `true`

packages/fast-breakpoint-tracker-react/src/breakpoints.ts
```diff
@@ -0,0 +1,32 @@
...
+    vp3: 768,
+    vp4: 1084,
+    vp5: 1400,
+    vp6: 1779
+};
+
+/**
+ * Identifies current breakpoint based on window width
+ */
+export function identifyBreakpoint(windowWidth: number, breakpoints: Breakpoints): any {
```
nicholasrice: Can you type the return value? should be `keyof Breakpoints | void`

packages/fast-breakpoint-tracker-react/src/breakpoints.ts
```diff
@@ -0,0 +1,32 @@
...
+    const breakpointArray: any = Object.keys(breakpoints).map(
+        (key: keyof Breakpoints, value: number) => ({key, value: breakpoints[key]})
+    ).sort((a: any, b: any) => {
+        const x: number = a.value;
```
nicholasrice: kinda a nit, but you don't really need to define these variables: 

```
.sort((a: any, b: any) => a.value - b.value)
```

packages/fast-breakpoint-tracker-react/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "babel-preset-react": "^6.24.1",
+    "enzyme": "^3.3.0",
+    "enzyme-adapter-react-16": "^1.1.1",
+    "exenv-es6": "^1.0.0",
+    "jest": "^22.3.0",
+    "lodash-es": "^4.0.0",
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0",
+    "react-test-renderer": "^16.2.0",
+    "ts-jest": "^22.0.4",
```
nicholasrice: can you update jest here? ^23.0.0 get's us type-checking in tests

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
chrisdholt: Yeah - I inverted this logic the other day and must have missed the `!`

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.spec.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+configure({adapter: new Adapter()});
+
+const globalAny: any = global;
+globalAny.window.resizeTo = (width: number, height: number): any => {
+    globalAny.window.innerWidth = width || globalAny.window.innerWidth;
+    globalAny.window.innerHeight = width || globalAny.window.innerHeight;
+    globalAny.window.dispatchEvent(new Event("resize"));
+};
+// globalAny.Promise = require.requireActual("promise");
```
nalogiudice: Should this be commented out still?

packages/fast-breakpoint-tracker-react/src/breakpoints.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+        expect(identifyBreakpoint(768, defaultBreakpoints)).toBe("vp3");
+        expect(identifyBreakpoint(1083, defaultBreakpoints)).toBe("vp3");
+        expect(identifyBreakpoint(1084, defaultBreakpoints)).toBe("vp4");
+        expect(identifyBreakpoint(1399, defaultBreakpoints)).toBe("vp4");
+        expect(identifyBreakpoint(1400, defaultBreakpoints)).toBe("vp5");
+        expect(identifyBreakpoint(1778, defaultBreakpoints)).toBe("vp5");
+        expect(identifyBreakpoint(1779, defaultBreakpoints)).toBe("vp6");
+    });
+
+    test("should return the correct breakpoint values when cust breakpoint values are passed", (): void => {
```
nalogiudice: typo "custom"

janechu:
I see no README.md or export from index file.

packages/fast-breakpoint-tracker-react/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "@types/lodash-es": "^4.17.0",
+    "@types/node": "^9.4.6",
+    "@types/react": "^16.3.6",
+    "babel-core": "^6.26.0",
+    "babel-jest": "^22.4.3",
+    "babel-preset-env": "^1.6.1",
+    "babel-preset-react": "^6.24.1",
+    "enzyme": "^3.3.0",
+    "enzyme-adapter-react-16": "^1.1.1",
+    "exenv-es6": "^1.0.0",
```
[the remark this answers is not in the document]
chrisdholt: good catch.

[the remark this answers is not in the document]
chrisdholt: good catch here again :)

packages/fast-breakpoint-tracker-react/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@microsoft/fast-breakpoint-tracker-react",
+  "description": "React HOC for managing breakpoints",
```
nicholasrice: It's not really a HOC anymore - just a normal component

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
```diff
@@ -0,0 +1,101 @@
...
+        this.state = {
+            activeBreakpoint: null
+        };
+    }
+
+    /**
+     * React life-cycle method
+     */
+    public componentDidMount(): void {
+        if (canUseDOM()) {
```
nicholasrice: Is there a reason you can't do this work in the constructor? It seems like you can prevent an initial re-render if you move this work there.

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+            window.addEventListener("resize", this.requestFrame);
+        }
+    }
+
+    /**
+     * React life-cycle method
+     */
+    public componentDidUpdate(): void {
+        if (canUseDOM()) {
```
nicholasrice: Why does this need to update the breakpoint when the component updates? Shouldn't it just update the breakpoint when the window resizes?

packages/fast-breakpoint-tracker-react/package.json
chrisdholt: Yeah - I'm updating this as well as adding a Readme per @janechu's comment

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
chrisdholt: This doesn't belong here...

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
chrisdholt: Added a comment, to ensure we address this. Added below as well:

We are doing this work in a lifecycle method instead of the constructor to ensure that server rendered instances with conditional DOM rendering by breakpoint will be re-rendered. ReactDOM.hydrate() will call the constructor again, but it does not trigger a re-render. It will only bind event handlers. The only way to ensure the correct DOM is consistently rendered on the client is to perform this work here.

So, if I have a view that conditionally renders DOM based on breakpoint, and I run that on the server, the HTML that is output may be different than what is actually required. When ReactDOM.hydrate() is called, the event listeners will be bound, but it will not re-render, meaning we could have an inaccurate DOM.

nicholasrice: Makes sense, thanks for adding the comment

chrisdholt:
Closing to refactor back into using an array rather than an object.

PR #1011, opened 2018-10-10
chore: ensure TypeScript compile errors cause unit-test failures

nicholasrice:
Utilize ts-jest 23 where tests will fail when there are TypeScript compile errors.


packages/fast-viewer-react/helpers/generateSnapshots.ts
```diff
@@ -2,20 +2,20 @@
...
+export interface Example {
     title: string;
-    props?: IExampleProps;
+    props?: ExampleProps;
     children?: React.ReactNode;
 }
 
-export interface IExampes {
-    data: IExample[];
+export interface Exampes {
```
chrisdholt: How do we want to handle the spelling error here: `Exampes` => `Examples`?

nicholasrice: oh we can correct - this isn't exported.

packages/fast-viewer-react/tsconfig.json
```diff
@@ -1,7 +1,28 @@
...
+            "scripthost"
+        ],
+        "jsx": "react",
+        "module": "CommonJS",
+        "moduleResolution": "node",
+        "outDir": "../dist",
+        "removeComments": true,
+        "rootDir": "./",
         "sourceMap": true,
         "target": "es5",
```
chrisdholt: Unrelated and non-blocking Q for @janechu - Do you know if there is a reason we are targeting ES5 here instead of ES6 like all other packages?

janechu: This is an old library we never got around to updating. I'm actually working on a major update for it now so it uses the foundation.

chrisdholt: party

PR #1038, opened 2018-10-16
chore: update .npmignore files across monorepo

### LINKED ISSUE #1004: Non-distributed folders should be ignored with an .npmignore file
Currently, only some of our packages include an .npmignore file which ignores non-distributed folders and files from being published to npm. Each package should be audited to see if additional folders or files are being unnecessarily published to the npm registry.

Example - The packages below are not currently ignoring their respective www/ folder: 

- fast-layouts-react
- fast-components-react-**

chrisdholt:
<body>
closes #1004 
<footer>
 

### DIFF
--- a/packages/fast-components-react-base/.npmignore
+++ b/packages/fast-components-react-base/.npmignore
@@ -4,7 +4,12 @@ __test__/
 *.test.*
 
 # Source files
+app/
+coverage/
 src/
+www/
 
 # babel config
 .babelrc
+tsconfig.json
+tslint.json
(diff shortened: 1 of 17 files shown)


packages/fast-components-react-base/.npmignore
```diff
@@ -4,7 +4,12 @@ __test__/
...
+app/
+coverage/
 src/
+www/
 
 # babel config
-.babelrc
\ No newline at end of file
+.babelrc
+tsconfig.json
```
nicholasrice: we might also want to add `webpack.config.js` to packages that use webpack

chrisdholt: good idea, updated.

PR #1043, opened 2018-10-16
fix: corrects typings and exports for all of fast-layouts-react

### LINKED ISSUE #1014: Pane component is not excepting className
While working in FARM on downloads.tsx I noticed managedClasses on Pane was not getting resolved. The change I had to make was going from:
<Pane className={this.props.managedClasses.content}>
to
<Pane {...{className: this.props.managedClasses.content}}>

nicholasrice:
closes #1014


packages/fast-layouts-react/src/canvas/index.ts
```diff
@@ -1,6 +1,27 @@
...
+    CanvasHandledProps as BaseCanvasHandledProps,
+    CanvasManagedClasses,
+    CanvasProps as BaseCanvasProps,
+    canvasStyleSheet,
+    CanvasUnhandledProps
+} from "./canvas";
+import { Subtract } from "utility-types";
 
-import Canvas from "./canvas";
+/* tslint:disable-next-line:typedef */
```
nalogiudice: Hey, I'm still a little unclear on what this is doing and when it needs to be done. Do you mind explaining it (possibly again - sorry) thanks

nicholasrice: It's to keep typescript from getting confused, though it may not actually be necessary with the JSSManager refactor. Previously, typescript would get confused when we would return from a function a class defined *in* that function. Now that we're not defining the class inside of a function though, this may not be necessary - I'll investigate separately.

PR #1049, opened 2018-10-18
chore: add "prettier" to format code and pre-commit hook to enforce

### LINKED ISSUE #950: proposal: add "prettier" for code formatting
**What is "Prettier"**
Prettier is an opinionated code formatter with support for many languages, and in our case, support for TypeScript.

**Why do we need "Prettier"? Don't we have linters?**
From prettier's website:
> Linters have two categories of rules:
>
> Formatting rules: eg: max-len, no-mixed-spaces-and-tabs, keyword-spacing, comma-style...
>
> Prettier alleviates the need for this whole category of rules! Prettier is going to reprint the entire program from scratch in a consistent way, so it's not possible for the programmer to make a mistake there anymore :)
>
> Code-quality rules: eg no-unused-vars, no-extra-bind, no-implicit-globals, prefer-promise-reject-errors...
>
> Prettier does nothing to help with those kind of rules. They are also the most important ones provided by linters as they are likely to catch real bugs with your code!

Prettier handles formatting of code to our specification, which will allow us to spend less mental energy when writing code or reviewing PR's.

**Can we use it with tslint?**
Yes, tslint can still be used to catch code-quality errors. From an implementation standpoint, there is a [package](https://github.com/alexjoverm/tslint-config-prettier) that helps prevent conflicts between the two.

**Can we configure prettier?**
Kind of. 
> Prettier is not a kitchen-sink code formatter that attempts to print your code in any way you wish. It is opinionated.

Prettier's primary goal is to stop all the on-going debates over styles. In their own words:

> The more options Prettier has, the further from the above goal it gets. The debates over styles just turn into debates over which Prettier options to use.

Prettier still supports options. Here is their [philosophy](https://prettier.io/docs/en/option-philosophy.html). Here are the [options](https://prettier.io/docs/en/options.html)

**Implementation**
I suggest we implement this using some sort of pre-commit hook. This way users don't need to worry about running it themselves, it will run as part of the normal dev process. https://prettier.io/docs/en/precommit.html

chrisdholt:
<body>
closes #950 
<footer>
 


.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+    "printWidth": 90,
+    "tabWidth": 4,
+    "useTabs": false,
+    "semi": true,
+    "singleQuote": false,
+    "trailingComma": "none",
```
nicholasrice: I actually think trailing commas could be nice - what are your thoughts on setting this to `es5`? It would be nice to avoid having to always to add a comma to an array or object every time somthing new is added.

It would also clean up diffs and help avoid merge conflicts in those areas, because you'd only be editing a single line instead of two.

packages/fast-animation/tslint.json
```diff
@@ -1,6 +1,7 @@
 {
     "extends": [
-        "@microsoft/fast-tslint-rules"
+        "@microsoft/fast-tslint-rules",
+        "tslint-config-prettier"
```
janechu: instead of putting this in each tslint.json, can we not extend the package from @microsoft/fast-tslint-rules instead?

chrisdholt: Synced with @nicholasrice on this. Thoughts below: 

I don't think we would *extend* this from the MSFT fast-tslint-rules because this really doesn't have anything to do with tslint. Instead of taking the dependency on this package, we would likely end up removing all style rules that conflict with prettier's opinion. This would require us to keep that manually updated instead of updating the package. Additionally, this would be breaking for anyone with a dependency on the tslint package because it would remove existing functionality they may be relying on.

If we find that this is difficult to maintain, I think we can revisit moving it.

PR #1061, opened 2018-10-18
feat: add tests to ensure all components are exported in base and MSFT

nalogiudice:
Closes [890](https://github.com/Microsoft/fast-dna/issues/890)
<body>

Returns `true` if tests pass, otherwise throws an error.

jest ignore code _could be added_ for [this issue](https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/#easier-ecmascript-module-interoperability). Imports seem correct but still throws WARN within core/MSFT

<footer>
 


packages/fast-components-react-base/package.json
```diff
@@ -42,6 +42,13 @@
         "lines": 80
       }
     },
+    "globals": {
+      "ts-jest": {
+        "diagnostics": {
+          "ignoreCodes": [151001]
+        }
+      }
+    },
```
chrisdholt: Is this necessary and/or related to this PR?

packages/fast-components-react-base/src/index.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+        const index: any = path.resolve(__dirname, "index.ts");
+
+        srcDir.forEach((entry: string) => {
+            components.push(entry.slice(4).replace(/-/g, ""));
+        });
+
+        const data: any = fs.readFileSync(index, "UTF-8");
+        const found: any = [];
+        const rxp: RegExp = /{([^}]+)}/g;
+        let curMatch: any;
```
chrisdholt: Can we clean up some of these "any" type defs? For instance, line 22 is clearly an array unless you're  changing types...to which I would ask, "why"?

Since we have "typings" in here, can you update typings that are actually known? For those that are truly *unknown* (which I'm not sure there are any) we should use the *unknown* typedef.

packages/fast-components-react-base/src/index.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    test("All components have respective class exports in index.ts", () => {
+        const isDirectory: any = (source: any): any => fs.lstatSync(source).isDirectory();
+        const getDirectories: any = (source: any): any =>
+            fs.readdirSync(source).map((name: any) => path.join(source, name)).filter(isDirectory);
+        const srcDir: string[] = getDirectories("src");
+        const components: string[] = [];
+        const index: any = path.resolve(__dirname, "index.ts");
+
+        srcDir.forEach((entry: string) => {
+            components.push(entry.slice(4).replace(/-/g, ""));
```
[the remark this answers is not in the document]
nalogiudice: Maybe I don't follow, but the issue I was trying to solve is that the exports look like this `export { HorizontalOverflow };` and the folder structure looks like `horizontal-exports` (in base)... so to compare, something has to be transformed.

packages/fast-components-react-base/src/index.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+        let curMatch: any;
+
+        /* tslint:disable-next-line */
+        while (curMatch = rxp.exec(data.split("\n"))) {
+            found.push(curMatch[1].replace(/\s/g, "").toLowerCase());
+        }
+
+        const missingExports: any = components.filter((component: string) => !found.includes(component));
+
+        expect(missingExports).toEqual([]);
```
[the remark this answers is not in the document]
janechu: In that case I think yes, it's better to have an output of missing exports, however as stated previously it should be in a root level helpers file. I'm not sure what issues you might have run into, can you elaborate?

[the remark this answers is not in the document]
nalogiudice: Import/export issues. Let me try extrapolating to the helper and then see if I run into any more issues. I'm thinking `root/build/helpers/*`

packages/fast-components-react-base/src/index.spec.ts
nalogiudice: Will all be updated in new PR

nalogiudice:
Without the ignoreCodes, the following WARN will occur when running unit-tests in core and MSFT
```
ts-jest[config] (WARN) TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
message TS151001: If you have issues related to imports, you should consider setting `esModuleInterop` to `true` in your TypeScript configuration file (usually `tsconfig.json`). See https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/#easier-ecmascript-module-interoperability for more information
```

packages/fast-components-react-base/src/index.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+import * as path from "path";
+import { includesAllSubdirectoriesAsNamedExports } from "../../../build/helpers/file-includes-all-subdirectories-as-named-exports";
+
+configure({ adapter: new Adapter() });
+
+describe("index.ts", (): void => {
+    test("should export all components in the src directory", () => {
+        expect(
+            includesAllSubdirectoriesAsNamedExports(path.resolve(__dirname, "index.ts"))
+        ).toEqual([]);
```
chrisdholt: This is interesting from a name standpoint as I expect a boolean value to be returned when I look at the function.

Does this "includesAllSubdirectoriesAsNamedExports" - Yes. Instead I see, "includesAllSubdirectoriesAsNamedExports" - empty array.

nalogiudice: A boolean would only tell you that it failed, but not which exports are missing. @janechu suggested the function name, but I am open to changing it so the expectation is a listing of missing exports or an empty array.

build/helpers/file-includes-all-subdirectories-as-named-exports.ts
```diff
@@ -0,0 +1,29 @@
+const path = require("path");
+const fs = require("fs");
+
+/**
+ * Verifies all components in source directory are exported
+ */
+export function includesAllSubdirectoriesAsNamedExports(indexFile: string) {
```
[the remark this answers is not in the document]
janechu: Without it we have no way of verifying that this function works. We will be doing more testing on our packages from the root in the future so I see it as inevitable.

packages/fast-components-react-base/src/index.spec.ts
janechu: you could use true when the array is empty and give an actual error when it returns with the appropriate details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

janechu: Then you could simply expect the function to not throw

nalogiudice: I'll look into implementing as described above.

nalogiudice:
@chrisdholt code climate issues are won't fix - not in my work files.

build/helpers/file-includes-all-subdirectories-as-named-exports.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export function includesAllSubdirectoriesAsNamedExports(indexFile: string): any {
+    /**
+     * Get the folders in the indexFile directory to compare to export listing
+     */
+    const directoryPath: string = path.dirname(indexFile);
+    const components: string[] = [];
+    const sourceDirectories: string[] = fs.readdirSync(directoryPath).map(
+        (name: string) => path.join(directoryPath, name)).filter(
+            (source: string) => fs.statSync(source).isDirectory());
```
nicholasrice: I think you should just `require` the path and then test all the keys of that object - that seems less fragile than regexing the contents of the file.

nalogiudice: Can't do this due to the fact with export differently in different repos. I left notes about this in the file.

nicholasrice: Can you describe the issue you're having? Why does differing export syntax prevent us from taking that approach?

nalogiudice: So, here is what the export data looks like across packages:
```
/**
 * Example formatting from fast-components-react-base
 */
import Foo from "./foo";
export { Foo };
export * from "./foo";

import Baz from "./baz";
export { Baz };
export * from "./baz";


/**
 * Example formatting from fast-component-react-msft
 */
export * from "./bar";
```
so, requiring like this:
```
const foo: any = require(indexFile);
console.log(Object.keys(foo));
```
results in `[ 'Foo', 'default', 'Baz' ]`

See the issue?

nalogiudice: Correct - sorry I thought supplying the full detail would be more helpful.

nicholasrice: In a quick prototype it *did* have those keys available. Given the following export file:

```
const Foo = "Foo";
const bar = "bar";

export default Foo;
export { bar };
```

The keys on the require object are "['default', 'bar']".

If you're getting something different than lets work together to get it figured out - I think testing the actual export object will be more reliable long-term.

nalogiudice: @nicholasrice  sure ping me whenever. To be clear, the exports in `fast-react-component-react` looks like this _only_
```
export * from "./button";
export * from "./call-to-action";
export * from "./caption";
```

nalogiudice: Should be all good now - thanks for the help

build/helpers/file-includes-all-subdirectories-as-named-exports.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    sourceDirectories.forEach((entry: string) => {
```
nicholasrice: You can avoid creating both a `sourceDirectories` and `components` variables:

```
const components = fs.readDirSync(directoryPath)
    .map(/*...*/)
    .filter(/*...*/)
    .map((entry: string) => PascalCase(path.basename(entry)))`
```

build/helpers/file-includes-all-subdirectories-as-named-exports.ts
```diff
@@ -0,0 +1,33 @@
+import * as path from "path";
+import * as fs from "fs";
+import { pascalCase } from "../../packages/fast-web-utilities";
+
+/**
+ * Verifies all components in source directory are exported
+ */
+export function includesAllSubdirectoriesAsNamedExports(indexFile: string): boolean | Error {
```
nicholasrice: You're not really returning an error here - you're throwing an error. This should just be boolean.

build/helpers/file-includes-all-subdirectories-as-named-exports.spec.ts
```diff
@@ -0,0 +1,20 @@
+import Adapter from 'enzyme-adapter-react-16'
+import { configure } from "enzyme";
+import * as path from "path";
+import { includesAllSubdirectoriesAsNamedExports } from "./file-includes-all-subdirectories-as-named-exports";
+
```
nicholasrice: I don't think you really need any of this enzyme setup to test exports.

nalogiudice: Removing all enzyme stuff.

build/helpers/file-includes-all-subdirectories-as-named-exports.ts
nalogiudice: Sure thing

build/helpers/file-includes-all-subdirectories-as-named-exports.ts
nalogiudice: Got it.

package.json
```diff
@@ -19,31 +19,85 @@
...
   },
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged"
     }
   },
+  "jest": {
+    "collectCoverage": true,
+    "collectCoverageFrom": [
+      "build/helpers/*.{tsx|ts}"
```
janechu: for now this could just be ts

package.json
```diff
@@ -19,31 +19,85 @@
...
+        "statements": 100,
+        "branches": 100,
+        "functions": 100,
+        "lines": 100
+      }
+    },
+    "testURL": "http://localhost",
+    "transform": {
+      "^.+\\.tsx?$": "ts-jest",
+      "^.+\\.jsx?$": "babel-jest"
```
janechu: I think this could just be ts and js respectively

package.json
```diff
@@ -19,31 +19,85 @@
...
       "lerna run prettier",
       "git add ."
     ]
   },
   "devDependencies": {
+    "@microsoft/fast-tslint-rules": "^3.0.0",
+    "@types/jest": "^23.3.5",
+    "@types/lodash-es": "^4.17.0",
     "@types/node": "^9.4.7",
+    "@types/react": "^16.3.6",
```
janechu: i think you can just get rid of @types/react

tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+    "compilerOptions": {
+        "moduleResolution": "node",
+        "module": "ES6",
+        "target": "ES6",
+        "baseUrl": ".",
+        "declaration": true,
+        "jsx": "react",
```
janechu: you can get rid of "jsx": "react"

package.json
```diff
@@ -19,31 +19,85 @@
...
+      "!<rootDir>/node_modules/lodash-es"
+    ],
+    "testPathIgnorePatterns": [
+      "packages/",
+      "build/helpers/__tests__/*"
+    ],
+    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$",
+    "moduleFileExtensions": [
+      "ts",
+      "tsx",
```
janechu: can remove the tsx asd jsx here

package.json
```diff
@@ -19,31 +19,85 @@
...
+      "^.+\\.jsx?$": "babel-jest"
+    },
+    "transformIgnorePatterns": [
+      "!<rootDir>/node_modules/lodash-es"
+    ],
+    "testPathIgnorePatterns": [
+      "packages/",
+      "build/helpers/__tests__/*"
+    ],
+    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$",
```
janechu: you can remove the x? parts of these

build/helpers/__tests__/src/bar/bar.tsx
```diff
@@ -0,0 +1 @@
+export default class Bar {}
```
janechu: this does not use react, can just be a ts file

build/helpers/__tests__/src/foo/foo.tsx
```diff
@@ -0,0 +1 @@
+export default class Foo {}
```
janechu: this does not use react, can just be a ts file

tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+        "esModuleInterop": true
+    },
+    "exclude": [
+        "node_modules",
+        "**/*.spec.ts",
+        "**/*.spec.tsx"
```
janechu: i think you could remove this line "**/*.spec.tsx"

.npmignore
```diff
@@ -0,0 +1,11 @@
+# Tests
```
janechu: you don't need this file at all since the root level package will never be published

package.json
```diff
@@ -19,31 +19,84 @@
...
+    "moduleFileExtensions": [
+      "ts",
+      "js"
+    ],
+    "moduleDirectories": [
+      "node_modules"
+    ]
+  },
   "lint-staged": {
-    "*.{ts,tsx,js}": [
```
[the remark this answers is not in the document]
nalogiudice: Got it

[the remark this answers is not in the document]
chrisdholt: This is true, with exception being that lint-staged runs at the git root. This will *not* lint .tsx files staged for commit and run prettier. We should add this back in.

package.json
```diff
@@ -19,31 +19,84 @@
     "integration-tests:alpha": "node build/testing/sauce-labs/test-browsers.js alpha",
     "integration-tests:beta": "node build/testing/sauce-labs/test-browsers.js beta",
     "integration-tests:release": "node build/testing/sauce-labs/test-browsers.js release",
-    "prettier:diff": "prettier --config .prettierrc '**/*.{ts,tsx,js}' --list-different"
+    "prettier:diff": "prettier --config .prettierrc '**/*.{ts,tsx,js}' --list-different",
```
chrisdholt: I don't think we're actually running prettier at the root. It's not setup for it currently. I think you can remove this npm script here as it runs via a lerna command across our packages. You can likely remove the tslint-config-prettier as well...unless you're having issues, which would surprise me since we aren't running it here.

nalogiudice: Got it - removing

package.json
```diff
@@ -19,31 +19,82 @@
...
+    "jest": "^23.6.0",
     "lerna": "^3.3.0",
     "lint-staged": "^7.3.0",
     "markdown-it": "^8.4.1",
     "rimraf": "^2.6.2",
     "prettier": "1.14.3",
+    "ts-jest": "^23.10.4",
     "ts-node": "^5.0.1",
+    "tslint": "^5.9.1",
+    "tslint-loader": "^3.6.0",
```
nicholasrice: You're not adding webpack are you? do you need the loader?

nalogiudice: Nope not needed - removed thanks

PR #1064, opened 2018-10-19
fix: remove external margin, padding, and fix RTL micro-interaction on MSFT call to action

### LINKED ISSUE #1048: MSFT call to action component is visually broken

chrisdholt:
<body>
closes #1048 
<footer>
 

### DIFF
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -31,9 +31,11 @@ export const callToActionButtonOverrides: ComponentStyles<
     button__primary: {
         "&:hover": {
             "& $button_contentRegion": {
-                left: (config: DesignSystem): string => {
+                transform: (config: DesignSystem): string => {
                     const designSystem: DesignSystem = withDesignSystemDefaults(config);
-                    return designSystem.direction === Direction.ltr ? "-4px" : "4px";
+                    return designSystem.direction === Direction.ltr
+                        ? "translateX(-4px)"
+                        : "translateX(4px)";
                 },
             },
         },
@@ -76,7 +78,10 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             "&:hover, &:focus": {
                 outline: "none",
                 "& $callToAction_glyph": {
-                    left: direction === Direction.ltr ? "4px" : "-4px",
+                    transform:
+                        direction === Direction.ltr
+                            ? "translateX(4px)"
+                            : "rotate(180deg) translateX(4px)",
                     position: "relative",
                 },
             },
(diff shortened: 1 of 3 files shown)


packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -34,7 +31,10 @@ export const callToActionButtonOverrides: ComponentStyles<
     button__primary: {
         "&:hover": {
             "& $button_contentRegion": {
-                [applyLocalizedProperty("left", "right", Direction.ltr)]: "-4px",
+                left: (config: DesignSystem): string => {
```
nicholasrice: I noticed the animation of these is pretty janky also - switching to a transform makes it much smoother. If you don't want to change in this PR I can file a task for a follow up but thought it might be within scope - your call.

chrisdholt: May as well do with these changes. CTA is completely broken right now as it is, so I'd say it's definitely in scope

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -91,10 +93,8 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             width: "8px",
             [applyLocalizedProperty("marginLeft", "marginRight", direction)]: "6px",
             transform: direction === Direction.ltr ? "none" : "rotate(180deg)",
-            verticalAlign: "top",
             transition: "all 600ms cubic-bezier(0.19, 1, 0.22, 1)",
-            left: "0",
-            marginTop: "4px",
+            marginTop: direction === Direction.ltr ? "4px" : "0",
```
nicholasrice: Why is he `marginTop` value concerned with document direction?

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -65,21 +67,21 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
...
-            marginTop: "16px",
             transition: "all 0.2s ease-in-out",
             "&:hover, &:focus": {
                 outline: "none",
                 "& $callToAction_glyph": {
-                    left: "8px",
+                    transform:
+                        direction === Direction.ltr
+                            ? "translateX(4px)"
+                            : "rotate(180deg) translateX(4px)",
```
nicholasrice: should translate be `-4px`?

chrisdholt: @nicholasrice can you pull down? it seems odd, but this was actually working as expected...

packages/fast-components-styles-msft/src/call-to-action/index.ts
chrisdholt: RTL is weird.

packages/fast-components-styles-msft/src/call-to-action/index.ts
nicholasrice: yep, thats... not what I expected. Looks good!

PR #1068, opened 2018-10-20
feat: add fast-breakpoint-tracker-react as a new package

### LINKED ISSUE #898: Breakpoint tracker should export as a non-instantiated instance of the class

chrisdholt:
<body>
closes #898
<footer>
 


packages/fast-breakpoint-tracker-react/src/breakpoints.ts
```diff
@@ -0,0 +1,23 @@
+export type Breakpoint = number;
+
+export type Breakpoints = Breakpoint[];
+
+export const defaultBreakpoints: Breakpoints = [0, 540, 768, 1084, 1400, 1779];
+
+/**
+ * Identifies current breakpoint based on window width
+ */
+export function identifyBreakpoint(
```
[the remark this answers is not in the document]
chrisdholt: Leaving this - we will deprecate (and possibly remove) the duplicate method from fast-layouts-react in a different PR

packages/fast-breakpoint-tracker-react/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```jsx
+const defaultBreakpoints = [0, 540, 768, 1084, 1400, 1779];
+```
+
+### Set custom breakpoint values
+Custom breakpoint values can be set by setting the `breakpoints` value on the BreakpointTracker class with new key:value pairs. Custom keys are supported, but the values must be numbers.
+
+```jsx
+const customBreakpoints = [0, 800, 1024, 1559];
```
nicholasrice: Maybe just
```
BreakpointTracker.breakpoints = [0, 800, 1024, 1559]
```

From a documentation perspective, there isn't a real reason to assign breakpoints to a const first

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
```diff
@@ -0,0 +1,106 @@
...
+     */
+    activeBreakpoint: Breakpoint | void;
+}
+
+export default class BreakpointTracker extends React.Component<
+    BreakpointTrackerProps,
+    BreakpointTrackerState
+> {
+    /**
+     * The current breakpoint.
```
nicholasrice: This property description doesn't seem quite accurate.

packages/fast-breakpoint-tracker-react/src/breakpoints.ts
```diff
@@ -0,0 +1,23 @@
+export type Breakpoint = number;
+
```
nicholasrice: I don't really think having a "Breakpoint" type is helpful. I think having an abstraction for a simple number just makes it harder to understand. Thoughts?

chrisdholt: Good with this. We can go with Breakpoints = number[];

packages/fast-breakpoint-tracker-react/README.md
```diff
@@ -0,0 +1,40 @@
...
+Custom breakpoint values can be set by setting the `breakpoints` value on the BreakpointTracker class with new key:value pairs. Custom keys are supported, but the values must be numbers.
+
+```jsx
+// Set the public static value to your custom breakpoints
+BreakpointTracker.breakpoints = [0, 800, 1024, 1559];
+```
+
+## Usage
+```jsx
+import React from "react";
```
nalogiudice: nit, don't we usually import like: `import * as React from "react";`?

packages/fast-breakpoint-tracker-react/package.json
```diff
@@ -0,0 +1,87 @@
...
+    "unit-tests": "jest --maxWorkers=4",
+    "unit-tests:watch": "jest --watch && npm-watch",
+    "watch": "npm run build -- -w --preserveWatchOutput"
+  },
+  "jest": {
+    "collectCoverage": true,
+    "coverageThreshold": {
+      "global": {
+        "statements": 80,
+        "branches": 50,
```
nalogiudice: Any way to up this coverage to 80?

chrisdholt: Currently there is no way to valuable way to test the resize event of the window that the component should respond to. While I could create brittle fake tests to up the coverage, they wouldn't be representative of what/how we should test and would serve only to help the branch coverage look good

packages/fast-breakpoint-tracker-react/README.md
```diff
@@ -0,0 +1,40 @@
...
+`fast-breakpoint-tracker-react` is a React component for managing viewport breakpoints. Given a set of breakpoints, this component will monitor the width of the browser window and track the current active breakpoint.
+
+## Installation
+`npm i --save @microsoft/fast-breakpoint-tracker-react`
+
+## Breakpoints
+`Breakpoints` are expected as an array. Six default breakpoint values are provided for convenience:
+
+```jsx
+const defaultBreakpoints = [0, 540, 768, 1084, 1400, 1779];
```
awentzel: Is the first array element `0` a valid breakpoint?  Feels like the first one might be 320 or 540.  Not zero.  Don't we assume zero to some breakpoint?

packages/fast-breakpoint-tracker-react/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+### Set custom breakpoint values
+Custom breakpoint values can be set by setting the `breakpoints` value on the BreakpointTracker class with new key:value pairs. Custom keys are supported, but the values must be numbers.
+
+```jsx
+// Set the public static value to your custom breakpoints
```
awentzel: I would suggest breaking the code comment out to be heading of the actual code block.

packages/fast-breakpoint-tracker-react/src/breakpoint-tracker.tsx
```diff
@@ -0,0 +1,104 @@
...
+        this.state = {
+            activeBreakpoint: null,
+        };
+    }
+
+    /**
+     * React life-cycle method
+     */
+    public componentDidMount(): void {
+        // We are doing this work in a lifecycle method instead of the constructor to ensure that
```
awentzel: Great to see this type of code comments.  Really helpful! Thx

packages/fast-breakpoint-tracker-react/src/breakpoints.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+        expect(identifyBreakpoint(539, defaultBreakpoints)).toBe(0);
+        expect(identifyBreakpoint(540, defaultBreakpoints)).toBe(1);
+        expect(identifyBreakpoint(767, defaultBreakpoints)).toBe(1);
+        expect(identifyBreakpoint(768, defaultBreakpoints)).toBe(2);
+        expect(identifyBreakpoint(1083, defaultBreakpoints)).toBe(2);
+        expect(identifyBreakpoint(1084, defaultBreakpoints)).toBe(3);
+        expect(identifyBreakpoint(1399, defaultBreakpoints)).toBe(3);
+        expect(identifyBreakpoint(1400, defaultBreakpoints)).toBe(4);
+        expect(identifyBreakpoint(1778, defaultBreakpoints)).toBe(4);
+        expect(identifyBreakpoint(1779, defaultBreakpoints)).toBe(5);
```
awentzel: Retina displays go up quite a bit and 4k even more?  Should we extend to include at least up to 4k with a final 4001+?

packages/fast-breakpoint-tracker-react/src/breakpoints.ts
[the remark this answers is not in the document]
awentzel: Would it make sense to add a todo in here referencing the future task?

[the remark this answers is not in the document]
chrisdholt: Not here, we're not going to pull the method from fast-layouts. It'll be part of that task, I don't know that we need to note it in the code as well.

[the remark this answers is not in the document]
chrisdholt: Also - intentionally NOT removing from fast-layouts-react right now as this package addition is independent of it.

packages/fast-breakpoint-tracker-react/src/breakpoints.spec.ts
chrisdholt: I mean, we can add tests but they won't validate anything here. Here we are testing that the method can correctly identify the correct breakpoint. If we want to add additional breakpoints to the defaultBreakpoints array I think that would be another task as I wouldn't want to add a breakpoint at random to cover larger screen sizes or densities.

packages/fast-breakpoint-tracker-react/README.md
chrisdholt: This matches our previous implementations and breakpoints from a mobile first strategy.

packages/fast-breakpoint-tracker-react/README.md
chrisdholt: I can remove this, but I think the existing title makes more sense: "Set custom breakpoint values". This serves to identify that we are setting a public static value to change the breakpoints.

PR #1079, opened 2018-10-23
feat: add a utility that converts hex values to rgb or rgba

ghost:
Made a new file called `rgb.ts` along with test cases to accompany it found in `rgb.spec.ts`. Depending on the length, the function will return either a string in the format `rgb(r, g, b)` (3 or 6 length hexits) or `rgba(r, g, b, a)` (4 or 8 length hexits). To learn more, look at my [previous PR](https://github.com/Microsoft/fast-dna/pull/1000)

Closes [#639](https://github.com/Microsoft/fast-dna/issues/639)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -21,6 +21,9 @@
 # fast-application-utils
 /packages/fast-application-utils/ @janechu @chrisdholt
 
+# fast-breakpoint-tracker-react
+/packages/fast-breakpoint-tracker-react/ @nicholasrice @chrisdholt
+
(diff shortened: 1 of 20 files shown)


nicholasrice:
@RamIyer1998 It looks like prettier needs to get run on your branch. To do this, you should reinstall the root node_modules, then run `lerna clean && lerna bootstrap && lerna run prettier`. Committing those changes should allow you to pass the build.

RamIyer1998:
Thanks! That seemed to do the trick, hopefully CircleCI passes now.

Issue #617, opened 2018-06-29
The detail view should show a message if no documentation has been provided


janechu:
The user currently sees a blank space where documentation should exist.

It should instead show the text "No documentation provided."

PR #1096, opened 2018-10-28
feat: add default message when no documentation is provided in development site

nalogiudice:
Closes [#617](https://github.com/Microsoft/fast-dna/issues/617)

<body>

<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/button/README.md
+++ b/packages/fast-components-react-base/src/button/README.md
@@ -1,5 +1,0 @@
-# Button
-*Buttons* are commonly used for final actions that complete a user work flow and the text should communicate what will happen after interaction. Typical use cases include submitting a form, closing a *doalog*, or saving settings. The *button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element.
-
-## Accessibility
-Consult the current [W3C WAI-ARIA documentation](https://www.w3.org/TR/wai-aria-practices/#button) for best practices.
--- a/packages/fast-components-react-base/src/button/README.md
+++ b/packages/fast-components-react-base/src/button/README.md
@@ -1,5 +1,5 @@
 # Button	
 *Buttons* are commonly used for final actions that complete a user work flow and the text should communicate what will happen after interaction. Typical use cases include submitting a form, closing a *doalog*, or saving settings. The *button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element.
 
- ## Accessibility	
+ ## Accessibility
 Consult the current [W3C WAI-ARIA documentation](https://www.w3.org/TR/wai-aria-practices/#button) for best practices.
--- a/packages/fast-components-react-base/src/button/README.md
+++ b/packages/fast-components-react-base/src/button/README.md
@@ -1,0 +1,5 @@
+# Button	
+*Buttons* are commonly used for final actions that complete a user work flow and the text should communicate what will happen after interaction. Typical use cases include submitting a form, closing a *doalog*, or saving settings. The *button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element.
+
+ ## Accessibility	
+Consult the current [W3C WAI-ARIA documentation](https://www.w3.org/TR/wai-aria-practices/#button) for best practices.
--- a/packages/fast-components-react-base/src/button/README.md
+++ b/packages/fast-components-react-base/src/button/README.md
@@ -1,5 +1,5 @@
-# Button	
+# Button
 *Buttons* are commonly used for final actions that complete a user work flow and the text should communicate what will happen after interaction. Typical use cases include submitting a form, closing a *doalog*, or saving settings. The *button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element.
 
- ## Accessibility
+## Accessibility
 Consult the current [W3C WAI-ARIA documentation](https://www.w3.org/TR/wai-aria-practices/#button) for best practices.
--- a/packages/fast-development-site-react/app/app.tsx
+++ b/packages/fast-development-site-react/app/app.tsx
@@ -1,5 +1,5 @@
-import ButtonSchema from "./components/button/button.schema.json";
 import * as React from "react";
+import * as ReactDOM from "react-dom";
 import { glyphBuildingblocks } from "@microsoft/fast-glyphs-msft";
 import Site, {
     FormChildOption,
@@ -14,10 +14,10 @@ import Site, {
     Theme,
 } from "../src";
 import designSystemDefaults, { Direction } from "./design-system";
-import Button from "./components/button/button";
-import * as ReactDOM from "react-dom";
-import Paragraph from "./components/paragraph/paragraph";
+import { Button, Paragraph, Span } from "./components/index";
+import ButtonSchema from "./components/button/button.schema.json";
 import ParagraphSchema from "./components/paragraph/paragraph.schema.json";
+import SpanSchema from "./components/paragraph/paragraph.schema.json";
 import { SiteCategoryProps, Status } from "../src/components/site/category";
 import ParagraphDocs from "./components/paragraph/.tmp/documentation";
 import ButtonDocs from "./components/button/.tmp/documentation";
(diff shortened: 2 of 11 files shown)


packages/fast-development-site-react/app/components/button/index.ts
```diff
@@ -0,0 +1,13 @@
...
+/*
+ * The type returned by manageJss type is very complicated so we'll let the
+ * compiler infer the type instead of re-declaring just for the package export
+ */
+/* tslint:disable-next-line:typedef */
+const Button = manageJss(ButtonStyles)(BaseButton);
+type Button = typeof Button;
+
+export default Button;
+export * from "./button";
```
chrisdholt: Is this a necessary addition for this PR? This seems like a lot of additional overhead to maintain for components that only exist to test on the dev site.

packages/fast-development-site-react/app/components/index.ts
```diff
@@ -0,0 +1,11 @@
...
+export { Button };
+export * from "./button";
+
+import Paragraph from "./paragraph";
+export { Paragraph };
+export * from "./paragraph";
+
+import Span from "./span";
+export { Span };
+export * from "./span";
```
chrisdholt: Is this a necessary addition for this PR? This seems like more additional overhead to maintain for components that only exist to test on the dev site.

packages/fast-development-site-react/app/components/paragraph/index.ts
```diff
@@ -0,0 +1,13 @@
...
+/*
+ * The type returned by manageJss type is very complicated so we'll let the
+ * compiler infer the type instead of re-declaring just for the package export
+ */
+/* tslint:disable-next-line:typedef */
+const Paragraph = manageJss(ParagraphStyles)(BaseParagraph);
+type Paragraph = typeof Paragraph;
+
+export default Paragraph;
+export * from "./paragraph";
```
chrisdholt: Same question on the value this gives us for components that we aren't exporting and only using for testing

packages/fast-development-site-react/src/components/site/category-documentation.tsx
```diff
@@ -28,11 +28,21 @@ class CategoryDocumentation extends React.Component<
...
+                <div className={this.props.managedClasses.documentationPanel}>
+                    {this.props.children}
+                </div>
+            );
+        } else {
+            return (
+                <div className={this.props.managedClasses.documentationPanel}>
+                    <div style={{ padding: "14px" }}>
+                        <p>No documentation provided.</p>
+                    </div>
```
chrisdholt: Is there a reason we are using the JSS manager for the wrapper here but not the div? Do we actually need the `<div/>`, or could this just be the paragraph?

packages/fast-development-site-react/app/app.tsx
```diff
@@ -31,7 +31,7 @@ export interface AppState {
 const themes: Theme[] = [
     { id: "Foo", displayName: "Foo", background: "#000" },
     { id: "Bar", displayName: "Bar", background: "#FFF" },
-    { id: "Rumple", displayName: "Rumple", background: "#333" },
+    { id: "Rumple", displayName: "Rumple", background: "#222" },
```
chrisdholt: Was this an intentional change?

packages/fast-development-site-react/src/components/site/category-documentation.tsx
nalogiudice: This is how we are formatting the output of the documentation within the temp folders. If you look at the documentation.tsx you can see the div with style, so this matches that.

packages/fast-development-site-react/app/app.tsx
nalogiudice: Yes, it allows the text to be visible on all test themes.

packages/fast-development-site-react/app/components/button/index.ts
nalogiudice: I tend to think of additional overhead as a negative _only_ when it can impact things negatively, i.e. performance or upkeep. This actually aligns the test app in the documentation site more to how we do things "in the real world" (base or MSFT) so I think it has good value.

I added another component to fully show a directory without documentation (a blank README.md in this case, though it would work the same without a README.md as well) as a good proof. As an outside user, I would find value in thorough examples that align with real usage.

packages/fast-development-site-react/app/components/index.ts
nalogiudice: See above comment.

packages/fast-development-site-react/app/components/paragraph/index.ts
nalogiudice: See above comment.

packages/fast-development-site-react/app/app.tsx
awentzel: But, it's hard coded?  Should it be using our color library so it's dynamic?

packages/fast-development-site-react/src/components/site/category-documentation.tsx
awentzel: I think @chrisdholt has a valid point here.  Even though it's how it's been done in other places we should always look to only have minimal required number of elements.  So if `<div/>` is required then great.  But, if it's not then we should eliminate this extra element.

nalogiudice: Maybe @janechu can shed some light on why the `<div />` was added in this way in the first place? I have no issue changing, just trying to understand the intent.

packages/fast-development-site-react/app/app.tsx
nalogiudice: Not for the dev site app - it's just a test app showing how to set up a dev site. I'm sure @janechu set it up this way intentionally.

packages/fast-development-site-react/app/app.tsx
```diff
@@ -14,10 +14,10 @@ import Site, {
     Theme,
 } from "../src";
 import designSystemDefaults, { Direction } from "./design-system";
-import Button from "./components/button/button";
-import * as ReactDOM from "react-dom";
-import Paragraph from "./components/paragraph/paragraph";
+import { Button, Paragraph, Span } from "./components/index";
```
janechu: nit: no need to point at index, the path can just be "./components"

packages/fast-development-site-react/app/app.tsx
janechu: I didn't set this particular part of it up, but yes its just a test app for the development site, it can't take a dependency, dev or otherwise, on any other library in our repository because it would be circular. Having said that I would prefer if we are adding to the array that it follows the current naming convention so the next item should be "Bat"

packages/fast-development-site-react/app/app.tsx
```diff
@@ -123,10 +128,19 @@ export default class App extends React.Component<{}, AppState> {
             component: Paragraph,
             status: Status.alpha,
         };
+        const componentObj2: any[] = [{ text: "uno" }, { text: "due" }, { text: "tre" }];
```
janechu: Can we rename these so that we have more of an indication on what they are, so this could be `spanTestComponents`, and can you change the ones above to `paragraphTestComponents`

packages/fast-development-site-react/app/app.tsx
```diff
@@ -123,10 +128,19 @@ export default class App extends React.Component<{}, AppState> {
...
+        const categoryObj2: SiteCategoryProps = {
```
janechu: can you rename this to `spanTestCategory` and the paragraph version to `paragraphTestCategory`

packages/fast-development-site-react/src/components/site/category-documentation.tsx
janechu: This was done quite a while ago and things have since changed in our dependent packages. I believe we could extrapolate it into its own component. That would require a larger refactor however.

nalogiudice: Prefer to do this in another [task](https://github.com/Microsoft/fast-dna/issues/1118)

chrisdholt: Does removing the div break something in this instance? From what I can tell this is *net-new* work and we are just re-using a pre-existing pattern. We have a className, then inline style, and a nested div, but have we tested just removing it to see if we can lighten the load here? It just seems all over the place. 

I'm fine doing a larger refactor for the component stuff; I don't know that it's a priority. But if this PR is a priority we should try and ensure that the code we are checking in is necessary and helpful. Is there a visual difference that we can't overcome by simplifying this and moving the code to the wrapping div?

nalogiudice: @chrisdholt the `<div></div>` element was there to provide a wrapper for the documentation, but we can use a React.Fragment instead. I've altered the logic to be done in the convert script, as this was working as expected in the dev site, but not on the component sites. Now it just sees if the markdown length is 0 and writes the appropriate message.

packages/fast-components-react-base/src/button/README.md
```diff
@@ -1,5 +1,5 @@
-# Button
+# Button	
 *Buttons* are commonly used for final actions that complete a user work flow and the text should communicate what will happen after interaction. Typical use cases include submitting a form, closing a *doalog*, or saving settings. The *button* component can be either an [anchor](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button) element.
 
-## Accessibility
+ ## Accessibility
```
janechu: extra spaces are getting added

PR #1110, opened 2018-11-05
feat: performance updates to JSS Manager

### LINKED ISSUE #1082: Components do not accept a ref
Adding a ref prop to any component causes a typescript error. All components should accept refs.

nicholasrice:
Implements significant performance improvements for the JSS Manager.  Implement's memoization techniques to avoid re-rendering stylesheets when the resulting stylesheet would be identical to one that has already been created. 

Also fixes issue where refs could not be applied to managed components.

closes #1082


packages/fast-jss-manager-react/src/sheet-manager.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+        expect(manager.get(stylesheet, designSystem)).toBeUndefined();
+    });
+
+    test("should not throw if updated with data that has not been previously added", () => {
+        expect(() => {
+            manager.update(stylesheet, designSystem, {});
+        }).not.toThrow();
+    });
+
+    test("should remove a sheet when it is only added once and is then updated", (): void => {
```
janechu: can you change to "should remove a sheet after it has been updated with a different design system" and add another test "should add a sheet after it has bee updated with a different design system" after the expect another one like `expect(manager.get(stylesheet, nextDesignSystem).not.toBeUndefined();`

packages/fast-jss-manager-react/src/sheet-manager.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    };
+
+    const designSystem: any = {
+        foreground: "blue",
+    };
+
+    const manager: SheetManager = new SheetManager();
+
+    beforeEach(() => {
+        manager.clean();
```
janechu: should there be a test for the clean method?

packages/fast-jss-manager-react/src/sheet-manager.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    test("should remove a sheet when it is only added once and is then updated", (): void => {
+        const nextDesignSystem: any = { foregroundColor: "red" };
+
+        manager.add(stylesheet, designSystem);
+        manager.update(stylesheet, designSystem, nextDesignSystem);
+
+        expect(manager.get(stylesheet, designSystem)).toBeUndefined();
+    });
+
+    test("should not remove a sheet on update when it has been added multiple times", (): void => {
```
janechu: can there be another test "should remove a sheet on update when it has been added and removed the same amount of tiles" and "should throw an error when attempting to update a sheet that does not exist"

packages/fast-jss-manager-react/src/design-system-provider.spec.tsx
```diff
@@ -78,4 +78,21 @@ describe("DesignSystemProvider", (): void => {
         expect(renderTwo.mock.calls).toHaveLength(1);
         expect(renderTwo.mock.calls[0][0]).toEqual({ a: "A", b: "b" });
     });
+
+    test("should not update the designSystem if it's props are unchanged", (): void => {
```
janechu: nit: its

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -60,201 +72,127 @@ const staticAndDynamicStyles: ComponentStyles<any, any> = {
 };
 
 describe("The JSSManager", (): void => {
-    interface TestDesignSystem {
-        color: string;
+    class NoPropsManager extends JSSManager<any, any, any> {
```
janechu: seems like this has props, id when used by SimpleComponent

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -60,201 +72,127 @@ const staticAndDynamicStyles: ComponentStyles<any, any> = {
...
-            />
-        );
-
-        const functionStylesheetComponent: ReactWrapper = mount(
-            <JSSManager
-                render={renderChild}
-                styles={functionStyleSheet}
-                designSystem={testDesignSystem}
-            />
+    test("should not pass the jssStyleSheet prop to managed component", (): void => {
```
janechu: should this be "should not pass the jssStyleSheet prop to a managed component without styles"?

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -60,201 +72,127 @@ const staticAndDynamicStyles: ComponentStyles<any, any> = {
...
 
-    test("should update an object stylesheet when the design-system changes", (): void => {
-        const objectStylesheetComponent: ReactWrapper = mount(
-            <JSSManager
-                designSystem={{ color: "blue" }}
-                styles={stylesheet}
-                render={renderChild}
-            />
-        );
+    test("should not pass the managedClasses prop through to managed component", (): void => {
```
janechu: likewise here "should not pass the managedClasses prop through to a managed component without styles"

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -60,201 +72,127 @@ const staticAndDynamicStyles: ComponentStyles<any, any> = {
...
     });
 
-    test("should remove stylesheets when unmounting", (): void => {
-        const objectStylesheetComponent: ReactWrapper = mount(
-            <JSSManager
-                designSystem={{ color: "red" }}
-                styles={stylesheet}
-                render={renderChild}
-            />
+    test("the default context should share identity between component instances", (): void => {
```
janechu: "should have a default context that shares identity between component instances"

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -60,201 +72,127 @@ const staticAndDynamicStyles: ComponentStyles<any, any> = {
...
         );
 
-        const objectSheet: any = objectStylesheetComponent.state("styleSheet");
-        const functionSheet: any = functionStylesheetComponent.state("styleSheet");
+        expect(renderedOne.instance().context).toBe(renderedTwo.instance().context);
+    });
 
-        expect(objectSheet.attached).toBe(true);
-        expect(functionSheet.attached).toBe(true);
+    test("the default context should share identity between component instances", (): void => {
```
janechu: "should have a default context that increments between component instances"

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
nicholasrice: oh - not this is correct. It means what its saying.

PR #1120, opened 2018-11-07
fix: correct jssStyleSheet style element ordering

nicholasrice:
closes #1109

### DIFF
--- a/packages/fast-jss-manager-react/src/jss-manager.spec.tsx
+++ b/packages/fast-jss-manager-react/src/jss-manager.spec.tsx
@@ -103,6 +103,60 @@ describe("The JSSManager", (): void => {
         expect(rendered.find("SimpleComponent").prop("jssStyleSheet")).toBeUndefined();
     });
 
+    test("should create jssStyleSheet with more specificity than the composed stylesheet", (): void => {
+        const context: any = { color: "red" };
+        const rendered: any = mount(
+            <StyledManager jssStyleSheet={{ foo: { color: "green" } }} />
+        );
+        const manager: any = rendered.find("StyledManager").instance();
+
+        expect(manager.secondaryStyleSheet().options.index).toBeGreaterThan(
+            manager.primaryStyleSheet().options.index
+        );
+    });
+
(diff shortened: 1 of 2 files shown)


packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -103,6 +103,81 @@ describe("The JSSManager", (): void => {
         expect(rendered.find("SimpleComponent").prop("jssStyleSheet")).toBeUndefined();
     });
 
+    test("should create jssStyleSheet with more specificity than the composed stylesheet", (): void => {
+        const context: any = { color: "red" };
```
janechu: doesnt seem like this gets used

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -103,6 +103,81 @@ describe("The JSSManager", (): void => {
...
+        );
+        const manager: any = rendered.find("StyledManager").instance();
+
+        expect(manager.secondaryStyleSheet().options.index).toBeGreaterThan(
+            manager.primaryStyleSheet().options.index
+        );
+    });
+
+    test("should create subsequent jssStyleSheet sheets with more specificity than a composed stylesheet", (): void => {
+        const context: any = { color: "red" };
```
janechu: also doesnt seem like this gets used

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -103,6 +103,81 @@ describe("The JSSManager", (): void => {
...
+
+        expect(managerTwo.secondaryStyleSheet().options.index).toBe(
+            manager.secondaryStyleSheet().options.index
+        );
+        expect(managerThree.secondaryStyleSheet().options.index).toBe(
+            managerTwo.secondaryStyleSheet().options.index
+        );
+    });
+
+    test("should decrease a counter tracking global stylesheet order on mount", (): void => {
```
chrisdholt: nit: Could this cause confusion since it's technically done prior to the component having mounted?

PR #1125, opened 2018-11-08
feat: create breadcrumb as a new component

eljefe223:
Create breadcrumb as a new component


packages/fast-components-class-name-contracts-base/src/breadcrumb-item/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * The class name contract for the breadcrumb-item component
```
nalogiudice: nit: no hyphen

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.props.ts
```diff
@@ -0,0 +1,29 @@
...
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface BreadcrumbItemManagedClasses
+    extends ManagedClasses<BreadcrumbItemClassNameContract> {}
+export interface BreadcrumbItemUnhandledProps
+    extends React.HTMLAttributes<HTMLLIElement> {}
+export interface BreadcrumbItemHandledProps extends BreadcrumbItemManagedClasses {
+    /**
+     * The children of the breadcrumb-item
```
nalogiudice: nit: np hyphen - I know it's just code comments, but we should be consistent

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.spec.tsx
```diff
@@ -0,0 +1,61 @@
...
+import BreadcrumbItem, { BreadcrumbItemUnhandledProps } from "./breadcrumb-item";
+import { BreadcrumbItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import Breadcrumb from "../breadcrumb/breadcrumb";
+
+/*
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("breadcrumb-item", (): void => {
```
nalogiudice: nit: breadcrumb item

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+class BreadcrumbItem extends Foundation<
+    BreadcrumbItemHandledProps,
+    BreadcrumbItemUnhandledProps,
+    {}
+> {
+    public static displayName: string = "BreadcrumbItem";
+
+    public static defaultProps: Partial<BreadcrumbItemHandledProps> = {
+        current: false,
```
nalogiudice: nit: remove comma

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
```diff
@@ -0,0 +1,74 @@
...
+    };
+
+    protected handledProps: HandledProps<BreadcrumbItemHandledProps> = {
+        managedClasses: void 0,
+        children: void 0,
+        current: void 0,
```
nalogiudice: nit: remove comma

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    aria-current={this.props.current ? "page" : undefined}
+                >
+                    {this.props.children}
+                </Hypertext>
+            </li>
+        );
+    }
+
+    /**
+     * Create class-names
```
nalogiudice: nit: class names

packages/fast-components-react-base/src/breadcrumb/README.md
```diff
@@ -0,0 +1,8 @@
+## Breadcrumb
+A generic breadcrumb component following the w3.org [breadcrumb](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/breadcrumb/index.html) pattern. The *breadcrumb* will accept any child elements but is built to work with *breadcrumb-item* component.
```
nalogiudice: nit remove hyphen

nalogiudice: And elsewhere

packages/fast-components-react-base/src/breadcrumb/README.md
```diff
@@ -0,0 +1,8 @@
+## Breadcrumb
+A generic breadcrumb component following the w3.org [breadcrumb](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/breadcrumb/index.html) pattern. The *breadcrumb* will accept any child elements but is built to work with *breadcrumb-item* component.
+
+### Usage
+The *breadcrumb* automatically sets the last *breadcrumb-item* `href` to "undefined" and `current` prop to "true". Addtionally *breadcrumb* accepts an optional `seperator` prop as a JSX.element that will render a JSX element between navigation elements. To support `rtl` in jss you need to target the `ol` in *breadcrumb* at set it to `display: "flex", flexWrap: "wrap"`
```
nalogiudice: Can you provide a why for "...automatically sets the last *breadcrumb-item* `href` to "undefined" and `current` prop to "true"."

nalogiudice: And, do we not support RTL out of the box here because it is base?

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <nav
+                {...this.unhandledProps()}
+                aria-label="Breadcrumb"
```
nalogiudice: I don't think we want to hardcode this string in

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,93 @@
...
+                className={this.generateClassNames()}
+            >
+                <ol>{this.renderChildren()}</ol>
+            </nav>
+        );
+    }
+
+    /**
+     * Create class-names
```
nalogiudice: nit: class names

packages/fast-components-react-base/src/breadcrumb/examples.data.tsx
```diff
@@ -0,0 +1,141 @@
...
+import Breadcrumb, { BreadcrumbManagedClasses, BreadcrumbProps } from "./breadcrumb";
+import { BreadcrumbItemProps } from "../breadcrumb-item";
+import { noop } from "lodash-es";
+import Documentation from "./.tmp/documentation";
+
+function breadcrumbItemPropFactory(): BreadcrumbItemProps {
+    return {
+        managedClasses: {
+            breadcrumbItem: "breadcrumb-item",
+            breadcrumbItem__current: "breadcrumb-item-current",
```
nalogiudice: nit: some unneeded commas in this file

packages/fast-components-react-msft/src/breadcrumb-item/README.md
```diff
@@ -0,0 +1,2 @@
+# Breadcrumb item
+A breadcrumb item component intended to work as a child of the *breadcrumb*.
```
nalogiudice: nit: italicize breadcrumb item

packages/fast-components-styles-msft/src/breadcrumb/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    config: DesignSystem
+): ComponentStyleSheet<BreadcrumbClassNameContract, DesignSystem> => {
+    const designSystem: DesignSystem = withDesignSystemDefaults(config);
+
+    return {
+        breadcrumb: {
+            padding: "4px",
+            "& ol": {
+                listStyle: "none",
+                paddingLeft: "0",
```
nalogiudice: need rtl for paddingLeft

packages/fast-components-styles-msft/src/breadcrumb-item/index.ts
```diff
@@ -0,0 +1,49 @@
...
+        },
+        breadcrumbItem__current: {},
+        breadcrumbItem__hypertext: {
+            outline: "none",
+            textDecoration: "none",
+            color: ensureForegroundNormal,
+            "&[href]": {
+                color: ensureBrandNormal,
+                "&:hover, &:focus": {
+                    ...applyHypertextBorder(2),
```
nalogiudice: You only call this once - why make it a function?

packages/fast-components-class-name-contracts-base/src/breadcrumb-item/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * The class name contract for the breadcrumb-item component
+ */
+export interface BreadcrumbItemClassNameContract {
+    breadcrumbItem?: string;
+    breadcrumbItem__current?: string;
+    breadcrumbItem__hypertext?: string;
```
chrisdholt: I don't think this is always going to be hypertext. We should be allowing "children", so I would suggest a more generic approach.

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
chrisdholt: Prettier will catch this if it's important; this may be added by prettier as well (ES5 trailing comma)

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
```diff
@@ -0,0 +1,74 @@
...
+        current: void 0,
+    };
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLLIElement> {
+        return (
+            <li {...this.unhandledProps()} className={this.generateClassNames()}>
+                <Hypertext
```
chrisdholt: What if someone needed these to be buttons? I don't know that we should hard require hypertext here.

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
```diff
@@ -0,0 +1,74 @@
...
+     */
+    protected generateClassNames(): string {
+        let className: string = get(this.props.managedClasses, "breadcrumbItem") || "";
+
+        if (this.props.current) {
+            className = className.concat(
+                " ",
+                get(this.props.managedClasses, "breadcrumbItem__current")
+            );
+        }
```
chrisdholt: I know this was done in context menu item for readability, I don't know that it's necessary here. This is not in a switch and I think it's actually more readable in this scenario to simply use interpolation.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
chrisdholt: +2

chrisdholt: @eljefe223 unless this is always an accessibility requirement (I don't think it is), we should give guidance and leave this to be passed via unhandledProps

packages/fast-components-styles-msft/src/breadcrumb-item/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    ComponentStyleSheet,
+    CSSRules,
+} from "@microsoft/fast-jss-manager";
+import { ensureBrandNormal, ensureForegroundNormal } from "../utilities/colors";
+import { BreadcrumbItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { applyTypeRampConfig } from "../utilities/typography";
+import { fontWeight } from "../utilities/fonts";
+import { toPx } from "@microsoft/fast-jss-utilities";
+
+function applyHypertextBorder(pixels: number): CSSRules<DesignSystem> {
```
chrisdholt: We shouldn't need these styles.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.props.ts
```diff
@@ -0,0 +1,22 @@
...
+export interface BreadcrumbHandledProps extends BreadcrumbManagedClasses {
+    /**
+     * The breadcrumb children
+     */
+    children?: React.ReactNode;
+
+    /**
+     * The breadcrumb seprator
+     */
+    seperator?: JSX.Element;
```
chrisdholt: I think we wanted to use a render prop here, not just allow a JSX.Element? At worst, we should support React.ReactNode

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
nicholasrice: semantically though, these would always need to be anchors I think

chrisdholt: I feel like we could find an outlier that was valid, but either way - we likely should follow predefined patterns and not hard code our own components here.

I think this should likely just be children.

eljefe223: According to w3 these need to be anchor tags and aria-current needs to be set to page on the last item

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
eljefe223: According to this it is https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/breadcrumb/index.html

packages/fast-components-styles-msft/src/breadcrumb-item/index.ts
eljefe223: This is the way its implemented in hypertext and wanted to keep things consistent.

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
chrisdholt: In MSFT we can put hypertext in as our examples/patterns.

packages/fast-components-styles-msft/src/breadcrumb-item/index.ts
chrisdholt: @eljefe223 I don't know that we need these - but even if we did, I would expect us to import the hypertext styles for instead of duplicating.

packages/fast-components-react-base/src/breadcrumb-item/breadcrumb-item.tsx
nalogiudice: Just wondering why we want trailing commas now (seems messy to me)

packages/fast-components-styles-msft/src/breadcrumb/index.ts
eljefe223: Even if its 0?

nalogiudice: Yep. When it switches it need to know right is 0

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
nalogiudice: It would still have to be localized I would think, which means that the user needs to pass the string value

packages/fast-components-react-base/src/breadcrumb/README.md
```diff
@@ -0,0 +1,8 @@
+## Breadcrumb
+A generic breadcrumb component following the w3.org [breadcrumb](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/breadcrumb/index.html) pattern. The *breadcrumb* will accept any child elements but is built to work with `a` elements.
+
+### Usage
+To denote the current page the *breadcrumb* automatically sets the last child `aria-current` prop to "page". Addtionally *breadcrumb* accepts an optional `separator` prop that will render a ReactNode between navigation elements. To support `rtl` in jss you need to target the `breadcrumb_ol` class for *breadcrumb* and set it to `display: "flex", flexWrap: "wrap"`
```
chrisdholt: > To support `rtl` in jss you need to target the `breadcrumb_ol` class for *breadcrumb* and set it to `display: "flex", flexWrap: "wrap"`

I don't think we need to include this. It's not fully accurate as this is *one way* and very specific to our MSFT implementation. There are a bunch of ways we could address RTL scenarios; for MSFT we are using flexbox, but other users may not choose to leverage that technology.

packages/fast-components-react-base/src/breadcrumb/README.md
```diff
@@ -0,0 +1,8 @@
+## Breadcrumb
+A generic breadcrumb component following the w3.org [breadcrumb](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/breadcrumb/index.html) pattern. The *breadcrumb* will accept any child elements but is built to work with `a` elements.
+
+### Usage
+To denote the current page the *breadcrumb* automatically sets the last child `aria-current` prop to "page". Addtionally *breadcrumb* accepts an optional `separator` prop that will render a ReactNode between navigation elements. To support `rtl` in jss you need to target the `breadcrumb_ol` class for *breadcrumb* and set it to `display: "flex", flexWrap: "wrap"`
+
+### Accessibility
+The *breadcrumb* component consists of a `nav` element with `aria-label` set to a `label` prop. One should set the `label` prop to "Breadcrumb" to identify the structure as a breadcrumb trail and make it a navigation landmark so that it is easy to locate. The last child element `aria-current` prop is marked "page" to indicate that it reprents the current page.
```
chrisdholt: >  One should set the `label` prop to "Breadcrumb"

I don't think we need to be this specific. We should call out why an aria-label is required, not what the content should be

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+        const unhandledProps: BreadcrumbUnhandledProps = {
+            "aria-controls": "controls",
+        };
+
+        const rendered: any = shallow(<Breadcrumb {...unhandledProps} />);
+
+        expect(rendered.first().prop("aria-controls")).toEqual("controls");
+    });
+
+    test("should set current value on last rendered", (): void => {
```
chrisdholt: So, exists will ensure that a current value exists, but it can't ensure that it sets it on the last value. We'd need to drill into the children to do so.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+                <a />
+                <a />
+                <a />
+            </Breadcrumb>
+        );
+
+        expect(rendered.exists("a.breadcrumb-item-current-class")).toBe(true);
+    });
+
+    test("should render a `div` element as second child if the `separator` prop is passed", () => {
```
chrisdholt: Is this an accurate description of something we want to test? It seems more like we are testing that seperator passes the element when the prop is provided.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,108 @@
...
+    };
+
+    /**
+     * Renders the component
+     */
+    public render(): any {
+        return (
+            <nav
+                {...this.unhandledProps()}
+                aria-label={this.props.label}
```
chrisdholt: label is not required, I would suggest passing a null value if it doesn't exist.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,108 @@
...
+     */
+    protected generateClassNames(): string {
+        return super.generateClassNames(get(this.props.managedClasses, "breadcrumb"));
+    }
+
+    /**
+     * Create class names
+     */
+    protected generateOlClassNames(): string {
+        return super.generateClassNames(get(this.props.managedClasses, "breadcrumb_ol"));
```
chrisdholt: I don't think we want to super here. This isn't a component, so this is going to pass a className intended for the root element if provided.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,108 @@
...
+                              this.props.managedClasses.breadcrumb_item
+                          }`
+                        : this.props.managedClasses.breadcrumb_item,
+            };
+
+            if (childCount - 1 === index) {
+                props.className = `${props.className} ${
+                    this.props.managedClasses.breadcrumb_item__current
+                }`;
+                props["aria-current"] = "page";
```
chrisdholt: we should simply use `true` as a value here if acceptable. "Page" will not localize and we should not manipulate strings. If we want to handle `aria-current`, we should do so with true/false values.

packages/fast-components-styles-msft/src/breadcrumb/index.ts
```diff
@@ -0,0 +1,45 @@
...
+    return {
+        breadcrumb: {
+            padding: "4px",
+        },
+        breadcrumb_item: {
+            fontWeight: `${fontWeight.bold}`,
+            display: "inline",
+            ...applyTypeRampConfig("t7"),
+        },
+        breadcrumb_item__current: {},
```
chrisdholt: What happens if a hypertext is passed with an href value? I would assume to ensure consistency we would want to use this within the MSFT package. Otherwise, what purpose is it serving if we expect that users style their current item to meet our requirements?

packages/fast-components-react-base/src/breadcrumb/README.md
chrisdholt: > To denote the current page the *breadcrumb* automatically sets the last child `aria-current` prop to "page".

The above does not localize. We should go with `true` as a value instead.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
eljefe223: But isn't page an appropriate value for aria-current? Its not just a boolean. https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/current/ Or do we map it to a prop? In the case of breadcrumb it should be set to "page"

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+        expect(rendered.first().prop("aria-controls")).toEqual("controls");
+    });
+
+    test("should set current class name on last rendered", (): void => {
+        const rendered: any = shallow(
+            <Breadcrumb managedClasses={managedClasses}>
+                <a />
+                <a />
+                <a />
+                <a key="blah" />
```
nalogiudice: Lets use `foo`

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+            </Breadcrumb>
+        );
+
+        /* tslint:disable-next-line */
+        console.log(rendered.instance().props.children[3].props.class);
```
nalogiudice: I don't think we should have a console.log in our tests.

eljefe223: This is just a product of me testing something at home. But didn't want to loose my progress so I pushed my changes. This is now longer part of the review.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
eljefe223: This is just a product of me testing something at home. But didn't want to loose my progress so I pushed my changes. This is now longer part of the review.

packages/fast-components-react-base/src/breadcrumb/examples.data.tsx
```diff
@@ -0,0 +1,142 @@
...
+            {
+                id: "hypertext",
+                props: {
+                    children: "breadcrumb item 3",
+                },
+            },
+        ],
+    },
+    data: [
+        {
```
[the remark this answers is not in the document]
mateuszkoga: Sam już niewiem

packages/fast-components-react-base/src/breadcrumb/examples.data.tsx
```diff
@@ -0,0 +1,142 @@
...
+};
+
+const examples: ComponentFactoryExample<BreadcrumbProps> = {
+    name: "Breadcrumb",
+    component: Breadcrumb,
+    schema: schema as any,
+    documentation: <Documentation />,
+    detailData: {
+        ...managedClasses,
+        label: "breadcrumb",
```
chrisdholt: Since label doesn't change at all in the examples, I would recommend setting this at the top of the file as a constant and passing it using the spread operator.

packages/fast-components-react-base/src/breadcrumb/examples.data.tsx
```diff
@@ -0,0 +1,142 @@
...
+        /* tslint:disable-next-line */
+        separator: (className?: string): React.ReactNode => {
+            return <div className={className}>\</div>;
+        },
```
chrisdholt: Can we assign this function to a variable or at minimum break it out so we can reduce the duplication?

packages/fast-components-react-msft/src/breadcrumb/examples.data.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { ComponentFactoryExample } from "@microsoft/fast-development-site-react";
+import schema from "@microsoft/fast-components-react-base/dist/breadcrumb/breadcrumb.schema.json";
+import Documentation from "./.tmp/documentation";
+import ReactHTMLElementSchema from "../../app/components/react-html-element.schema.json";
+
+export default {
+    name: "Breadcrumb",
+    component: Breadcrumb,
+    schema: schema as any,
+    documentation: <Documentation />,
```
chrisdholt: Same suggestions as the other example file to mitigate unnecessary duplication.

packages/fast-components-class-name-contracts-base/src/breadcrumb/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * The class name contract for the breadcrumb component
+ */
+export interface BreadcrumbClassNameContract {
+    breadcrumb?: string;
+    breadcrumb_item?: string;
+    breadcrumb_item__current?: string;
+    breadcrumb_ol?: string;
```
chrisdholt: We tend to shy away from naming something after the element; is there something we can use for this instead of `ol`?

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+        );
+
+        const lastItem: any = rendered.findWhere((element: any) => {
+            return element.props()["aria-current"] !== undefined;
+        });
+        /* tslint:disable-next-line */
+        expect(lastItem.props()["className"]).toContain("breadcrumb-item-current-class");
+    });
+
+    test("should not clobber custom class name on last rendered", (): void => {
```
chrisdholt: Thanks for this test case!

packages/fast-components-react-msft/src/breadcrumb/examples.data.tsx
```diff
@@ -0,0 +1,125 @@
...
+            {
+                id: "hypertext",
+                props: {
+                    children: "breadcrumb item 3",
+                },
+            },
+        ],
+    },
+    data: [
+        {
```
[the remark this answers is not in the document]
mateuszkoga: Hi

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -62,6 +62,11 @@ class Breadcrumb extends Foundation<
     private isClonableElement(node: React.ReactNode): node is React.ReactElement<any> {
         return React.isValidElement(node);
     }
+
+    private generateCurrentClassName(): string {
+        const className: string = this.props.managedClasses.breadcrumb_item__current;
```
chrisdholt: I think this will work:
```
get(this.props, "managedClasses.breadcrumb_item__current") || ""
```

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,111 @@
...
+    ): React.ReactFragment => {
+        const childCount: number = React.Children.count(this.props.children);
+        let augmentedChild: React.ReactNode = child;
+        let notLastItem: boolean = true;
+        if (this.isClonableElement(child)) {
+            const props: any = {
+                className:
+                    child.props && typeof child.props.className === "string"
+                        ? `${child.props.className} ${
+                              this.props.managedClasses.breadcrumb_item
```
chrisdholt: you should likely use `get` for this and all other managedClasses; I'd suggest setting an empty string if unavailable.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.props.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * The aria-label to provide an accessible name for the breadcrumb
+     */
+    label?: string;
+
+    /**
+     * The breadcrumb seprator
+     */
+    separator?: (className: string) => React.ReactNode;
```
nicholasrice: This could theoretically be optional if no class is provided to the stylesheet

chrisdholt: +1 good call

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,111 @@
...
+                props["aria-current"] = "page";
+                notLastItem = false;
+            }
+
+            augmentedChild = React.cloneElement(child, props);
+        }
+        return (
+            <li>
+                {augmentedChild}
+                {typeof this.props.separator === "function" && notLastItem
```
nicholasrice: You should use `get(this.props, "managedClasses.breadcrumb_separator"` so that it doesn't throw if managedClasses are not provided

eljefe223: already fixed.

packages/fast-components-react-base/src/breadcrumb/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,97 @@
...
+    breadcrumb_itemsContainer: "breadcrumb-items-container-class",
+    breadcrumb_separator: "breadcrumb-separator-class",
+};
+
+describe("breadcrumb", (): void => {
+    test("should have a displayName that matches the component name", () => {
+        expect((Breadcrumb as any).name).toBe(Breadcrumb.displayName);
+    });
+
+    test("should not throw if managedClasses are not provided", () => {
```
nicholasrice: you should add an expect statement with a breadcrumb that implements children and a separator also - with the current implementation it would throw

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,132 @@
...
+            </nav>
+        );
+    }
+
+    /**
+     * Create class names
+     */
+    protected generateClassNames(): string {
+        return super.generateClassNames(
+            get(this.props.managedClasses, "breadcrumb") || ""
```
nicholasrice: just FYI - `super.generateClassName` will handle an undefined value so you don't need the `|| ""` here

packages/fast-components-react-base/src/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,132 @@
...
+        );
+    }
+
+    /**
+     * Create items container class names
+     */
+    protected generateItemsContainerClassNames(): string {
+        return get(this.props.managedClasses, "breadcrumb_itemsContainer") || "";
```
nicholasrice: same for all of these really - you only need `|| ""` when you're doing string operations on the value. React will handle `undefined` just fine.

packages/fast-components-react-msft/app/components/react-html-element.tsx
```diff
@@ -15,6 +15,12 @@ export default class ReactHTMLElement extends React.Component<ReactHTMLElementPr
     };
 
     public render(): JSX.Element {
-        return <this.props.tag>{this.props.children}</this.props.tag>;
+        const props: any = Object.assign({}, this.props);
+
+        delete props.tag;
+        delete props.slot;
```
nicholasrice: slot is a valid attribute - why are we deleting it again?

packages/fast-components-react-base/app/components/react-html-element.tsx
```diff
@@ -15,6 +15,11 @@ export default class ReactHTMLElement extends React.Component<ReactHTMLElementPr
     };
 
     public render(): JSX.Element {
-        return <this.props.tag>{this.props.children}</this.props.tag>;
+        const props: any = Object.assign({}, this.props);
+
+        delete props.tag;
```
nalogiudice: Not holding up the PR, but what's going on here?

packages/fast-components-styles-msft/src/breadcrumb/index.ts
```diff
@@ -0,0 +1,54 @@
...
+): ComponentStyleSheet<BreadcrumbClassNameContract, DesignSystem> => {
+    const designSystem: DesignSystem = withDesignSystemDefaults(config);
+    const direction: Direction = designSystem.direction;
+
+    return {
+        breadcrumb: {
+            padding: "4px",
+        },
+        breadcrumb_item: {
+            fontWeight: `${fontWeight.bold}`,
```
nalogiudice: Unless I am missing something, can't you just do `fontWeight: fontWeight.bold`?

packages/fast-components-styles-msft/src/breadcrumb/index.ts
```diff
@@ -0,0 +1,54 @@
...
+            display: "inline",
+            outline: "none",
+            textDecoration: "none",
+            color: ensureForegroundNormal,
+            ...applyTypeRampConfig("t7"),
+            "&:link, &:visited": {
+                color: ensureBrandNormal,
```
chrisdholt: Do we not need to pass config similar to what we are doing below?

packages/fast-components-styles-msft/src/breadcrumb/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const styles: ComponentStyles<BreadcrumbClassNameContract, DesignSystem> = (
+    config: DesignSystem
+): ComponentStyleSheet<BreadcrumbClassNameContract, DesignSystem> => {
+    const designSystem: DesignSystem = withDesignSystemDefaults(config);
+    const direction: Direction = designSystem.direction;
+
+    return {
+        breadcrumb: {
+            padding: "4px",
```
chrisdholt: As a principle - we typically don't include padding or margin on the root of components; that way, we don't have to override existing values every time someone wants to use those values to position something contextually.

packages/fast-components-styles-msft/src/breadcrumb/index.ts
eljefe223: Yeah... Copied form elsewhere

packages/fast-components-react-base/app/components/react-html-element.tsx
eljefe223: this allows managed classes to pass down to the child rendered by the component... Which it was not doing before

Issue #1081, opened 2018-10-24
Add overflow callback in horizontal overflow


chrisdholt:
Horizontal overflow does not provide a callback for when the scroll position is at the start, and items do not overflow.

Should return true, true in this scenario so a user knows there is nothing to scroll.

chrisdholt:
Current thinking is to add a *new* callback for when the horizontal overflow initializes and provides:
- Same values that currently exist in the onScrollChange callback
- An additional value for if there are not enough items to overflow

PR #1134, opened 2018-11-13
feat: add optional callback for overflow to horizontal overflow

nalogiudice:
Closes [#1081](https://github.com/Microsoft/fast-dna/issues/1081)

I left comments to explain rationale. The intent is that the overflow callback also provides the onScroll callback data.

<body>

<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -2,16 +2,7 @@
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
 ### Usage guidance
-The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL. 
-
-### Callbacks
-Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplies the `onScrollChange` data.) The `onHorizontalOverflowChange` callback uses the new `ResizeObserver` (available Chrome) so it is necessary to apply a polyfill for `ResizeObserver` if functionality is needed in Safari and/or Firefox. Window resize is used as a fallback.
-
-Usage example:
-
-`onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange data: ", obj))`
-
-`onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange data: ", obj))`
+The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL. Several callbacks are supplied to provide the consumer with overflow and scroll data. The not fully supported [ResizeObserver](https://developers.google.com/web/updates/2016/10/resizeobserver) is used (supported in Chrome), so it is necessary to apply a polyfill if more thorough browser support is needed (Safari and Firefox). Full adaptation is expected soon.
 
 ## Accessibility
 The 'next' and 'previous' buttons are only useful for sighted users, since overflow is a purely visual construct, so they should be hidden from screen readers.
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -2,7 +2,16 @@
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
 ### Usage guidance
-The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL.
+The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL. 
+
+### Callbacks
+Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplied the `onScrollChange` data.)
+
+Usage example:
+
+`onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange data: ", obj))`
+
+`onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange data: ", obj))`
 
 ## Accessibility
 The 'next' and 'previous' buttons are only useful for sighted users, since overflow is a purely visual construct, so they should be hidden from screen readers.
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -5,7 +5,7 @@ The *horizontal overflow* component accepts items as children. A child can have 
 The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL. 
 
 ### Callbacks
-Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplied the `onScrollChange` data.)
+Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplies the `onScrollChange` data.)
 
 Usage example:
 
--- a/packages/fast-components-react-base/src/horizontal-overflow/README.md
+++ b/packages/fast-components-react-base/src/horizontal-overflow/README.md
@@ -5,7 +5,7 @@ The *horizontal overflow* component accepts items as children. A child can have 
 The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL. 
 
 ### Callbacks
-Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplies the `onScrollChange` data.)
+Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplies the `onScrollChange` data.) The `onHorizontalOverflowChange` callback uses the new `ResizeObserver` (available Chrome) so it is necessary to apply a polyfill for `ResizeObserver` if functionality is needed in Safari and/or Firefox. Window resize is used as a fallback.
 
 Usage example:
 
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -184,12 +184,6 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
             },
             ...images,
         ],
-        // Example onOverflowChange callback usage
-        // onOverflowChange: (obj: OverflowChange): any =>
-        //     console.log("onOverflowChange data: ", obj),
-        // Example onScrollChange callabck usage
-        // onScrollChange: (obj: ScrollChange): any =>
-        //    console.log("onScrollChange data: ", obj)
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -1,9 +1,9 @@
 import * as React from "react";
 import HorizontalOverflow, {
-    HorizontalOverflowChange,
     HorizontalOverflowHandledProps,
     HorizontalOverflowManagedClasses,
     HorizontalOverflowUnhandledProps,
+    OverflowChange,
     ScrollChange,
 } from "./horizontal-overflow";
 import schema from "./horizontal-overflow.schema.json";
@@ -182,8 +182,19 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
                     children: "next",
                 },
             },
-            ...images,
+            // ...images,
+            images[0],
+            images[0],
+            images[0],
+            images[0],
+            images[0],
+            images[0],
         ],
+        // Example onOverflowChange callback usage
+        onOverflowChange: (obj: OverflowChange): any =>
+            console.log("onOverflowChange data: ", obj),
+        // Example onScrollChange callabck usage
+        // onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange data: ", obj))
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -2,9 +2,6 @@ import * as React from "react";
 import HorizontalOverflow, {
     HorizontalOverflowHandledProps,
     HorizontalOverflowManagedClasses,
-    HorizontalOverflowUnhandledProps,
-    OverflowChange,
-    ScrollChange,
 } from "./horizontal-overflow";
 import schema from "./horizontal-overflow.schema.json";
 import Documentation from "./.tmp/documentation";
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -1,8 +1,10 @@
 import * as React from "react";
 import HorizontalOverflow, {
+    HorizontalOverflowChange,
     HorizontalOverflowHandledProps,
     HorizontalOverflowManagedClasses,
     HorizontalOverflowUnhandledProps,
+    ScrollChange,
 } from "./horizontal-overflow";
 import schema from "./horizontal-overflow.schema.json";
 import Documentation from "./.tmp/documentation";
@@ -159,6 +161,8 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
     documentation: <Documentation />,
     detailData: {
         ...classes,
+        // Example of onScrollChange usage
+        // onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange: ", obj)),
         children: [
             {
                 id: ButtonSchema.id,
@@ -182,6 +186,9 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
             },
             ...images,
         ],
+        ...classes,
+        // Example of onHorizontalOverflowChange usage
+        // onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange: ", obj)),
     },
     data: [
         {
@@ -245,9 +252,6 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
         {
             ...classes,
             children: [...images],
-        },
-        {
-            ...classes,
         },
     ],
 };
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -182,16 +182,8 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
                     children: "next",
                 },
             },
-            // ...images,
-            images[0],
-            images[0],
-            images[0],
-            images[0],
-            images[0],
-            images[0],
+            ...images,
         ],
-        onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any =>
-            console.log("onHorizontalOverflowChange data: ", obj),
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -179,15 +179,8 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
                     children: "next",
                 },
             },
-            // ...images,
-            images[0],
-            images[0],
-            images[0],
-            images[0],
-            images[0],
-            images[0],
+            ...images,
         ],
-        onOverflowChange: (overflowObject: any): any => console.log(overflowObject),
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -182,8 +182,15 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
                     children: "next",
                 },
             },
-            ...images,
+            // ...images,
+            images[0],
+            images[0],
+            images[0],
+            images[0],
+            images[0],
+            images[0],
         ],
+        // onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange data: ", obj))
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -179,8 +179,15 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
                     children: "next",
                 },
             },
-            ...images,
+            // ...images,
+            images[0],
+            images[0],
+            images[0],
+            images[0],
+            images[0],
+            images[0],
         ],
+        onOverflowChange: (overflowObject: any): any => console.log(overflowObject),
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -184,10 +184,6 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
             },
             ...images,
         ],
-        // Example of onHorizontalOverflowChange usage
-        // onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange: ", obj)),
-        // Example of onScrollChange usage
-        // onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange: ", obj)),
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -190,7 +190,8 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
             images[0],
             images[0],
         ],
-        // onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange data: ", obj))
+        onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any =>
+            console.log("onHorizontalOverflowChange data: ", obj),
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -188,7 +188,8 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
         // onOverflowChange: (obj: OverflowChange): any =>
         //     console.log("onOverflowChange data: ", obj),
         // Example onScrollChange callabck usage
-        // onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange data: ", obj))
+        // onScrollChange: (obj: ScrollChange): any =>
+        //    console.log("onScrollChange data: ", obj)
     },
     data: [
         {
--- a/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
@@ -182,17 +182,11 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
                     children: "next",
                 },
             },
-            // ...images,
-            images[0],
-            images[0],
-            images[0],
-            images[0],
-            images[0],
-            images[0],
+            ...images,
         ],
         // Example onOverflowChange callback usage
-        onOverflowChange: (obj: OverflowChange): any =>
-            console.log("onOverflowChange data: ", obj),
+        // onOverflowChange: (obj: OverflowChange): any =>
+        //     console.log("onOverflowChange data: ", obj),
         // Example onScrollChange callabck usage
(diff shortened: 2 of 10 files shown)


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -36,14 +36,25 @@ class HorizontalOverflow extends Foundation<
...
+
+    /**
+     * Throttle resize request animation frame usage
+     */
+    private throttledResize: throttle;
+
+    /**
+     * The scroll change object (so it can be accessed in other callbacks)
+     */
+    private scrollChangeObject: ScrollChange = {} as ScrollChange;
```
chrisdholt: Why are you setting this as an empty object and casting it? Do we need to set it as an empty object here?

nalogiudice:
Great feedback @chrisdholt - I've updated most of the functions to align with your comments.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +212,50 @@ class HorizontalOverflow extends Foundation<
...
+    /**
+     * Checks if overflow is occuring
+     */
+    private checkOverflow(): boolean {
+        const regionWidthInformation: any = this.getItemsWidthAndTotalWidth();
+        const availableWidth: number = regionWidthInformation.availableWidth;
+        const itemWidths: number[] = regionWidthInformation.itemWidths;
+        const totalItemWidth: number = itemWidths.reduce((a: number, b: number) => a + b);
+
+        return totalItemWidth < availableWidth ? false : true;
```
chrisdholt: @nalogiudice this statement should return bool:

```
return totalItemWidth < availableWidth;
```

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +212,50 @@ class HorizontalOverflow extends Foundation<
...
+    }
+
+    /**
+     * Gets the overflow change data
+     */
+    private getOnHorizontalOverflowChangeData = (): HorizontalOverflowChange => {
```
chrisdholt: Could we lose this abstraction and just return this below in the onHorizontalOverflowChange?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: Duh yeah.

packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
```diff
@@ -182,6 +184,10 @@ const examples: ComponentFactoryExample<HorizontalOverflowHandledProps> = {
             },
             ...images,
         ],
+        // Example of onHorizontalOverflowChange usage
+        // onHorizontalOverflowChange: (obj: HorizontalOverflowChange): any => (console.log("onHorizontalOverflowChange: ", obj)),
+        // Example of onScrollChange usage
+        // onScrollChange: (obj: ScrollChange): any => (console.log("onScrollChange: ", obj)),
```
janechu: can this be in the README.md instead, noone is going to come to the example files for this information.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +212,41 @@ class HorizontalOverflow extends Foundation<
...
+        const regionWidthInformation: any = this.getItemsWidthAndTotalWidth();
+        const availableWidth: number = regionWidthInformation.availableWidth;
+        const itemWidths: number[] = regionWidthInformation.itemWidths;
+        const totalItemWidth: number = itemWidths.reduce((a: number, b: number) => a + b);
+
+        return totalItemWidth > availableWidth;
+    }
+
+    /**
+     * Callback for the horizontal overflow change
```
janechu: This is not a callback, it is a method that fires a callback. This should be called "handleHorizontalOverflowChange"

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -406,9 +459,12 @@ class HorizontalOverflow extends Foundation<
     };
 
     /**
-     * Handler for the click event fired after next or previous has been clicked
+     * Returns the items width and total width of the scroll region
      */
-    private handleClick(direction: ButtonDirection): void {
+    private getItemsWidthAndTotalWidth(): {
```
janechu: doesn't seem to specify a return type

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -420,6 +476,16 @@ class HorizontalOverflow extends Foundation<
...
             itemWidths.push(getClientRectWithMargin(item).width);
         }
+        return { availableWidth, itemWidths };
+    }
+
+    /**
+     * Handler for the click event fired after next or previous has been clicked
+     */
+    private handleClick(direction: ButtonDirection): void {
+        const regionWidthInformation: any = this.getItemsWidthAndTotalWidth();
```
janechu: this should not be any.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +212,41 @@ class HorizontalOverflow extends Foundation<
...
+    }
+
+    /**
+     * Callback for the horizontal overflow change
+     */
+    private onHorizontalOverflowChange = (): void => {
+        if (typeof this.props.onHorizontalOverflowChange === "function") {
+            this.props.onHorizontalOverflowChange({
+                ...this.getScrollChangeData(),
+                overflow: this.checkOverflow(),
```
janechu: I'm actually a little lost on why we are providing a whole new callback to add a value to the scroll change object, does anyone know what the context of this decision was?

chrisdholt: + @nicholasrice who may be able to keep me honest

I'll try and capture the conversation we had a couple weeks ago that led to this additional callback:
- On initialization, a user may need to know if the container has enough items to overflow (perhaps I want to hide flippers)
- A user need to know if the window has been resized and the container is now overflowing (I now want to show flippers)
- The onScrollChange event should only fire in the event of a scroll change; this is not a scroll change.
- The onHorizontalOverflowChange callback includes when there is a scroll change, but we landed that they are both useful and different enough to necessitate separate callbacks.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -403,5 +403,55 @@ describe("horizontal overflow", (): void => {
...
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+
+        rendered.unmount();
+
+        map.resize();
+
+        expect(resizeCallback).toHaveBeenCalledTimes(1);
+    });
```
chrisdholt: @nalogiudice can we add a test here to ensure that `onHorizontalOverflowChange` is fired on load? I think this should be doable given the limited scope of the test

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: +1 @janechu you were part of this discussion as well and agreed on the above points

packages/fast-components-react-base/src/horizontal-overflow/examples.data.tsx
nalogiudice: > noone is going to come to the example files for this information.

I actually strongly _disagree_ with this, but I will move to the README for the sake of not contending. The first place I look when trying to learn implementation is the examples.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: It does? 
```
private getItemsWidthAndTotalWidth(): {
        availableWidth: number;
        itemWidths: number[];
    } {...}

Comment (nalogiudice) [reply packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx]: I went ahead and made this an interface to solve the below issue and clear this up.

File: packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx (hunk 7 above)
Comment (nalogiudice) [reply packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx]: See above.

File: packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx (hunk 6 above)
Comment (chrisdholt) [reply packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx]: Ah - I think this is prettier showing up here because an object is being specified as the return object and it looks like the function...

_Should we create an interface here?_ Thoughts?

File: packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx (hunk 7 above)
Comment (chrisdholt) [reply packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx]: +1 @nalogiudice if we create an interface for the return type above we can use it here as well

File: packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx (hunk 9 above)
Comment (nalogiudice) [reply packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx]: Sure

File: packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx [hunk 10]
```diff
@@ -367,6 +367,22 @@ describe("horizontal overflow", (): void => {
 
         expect(renderedWithImages.instance().state.itemsHeight).toBe(0);
     });
+    test("should call `onHorizontalOverflowChange` in the `onLoad` method if prop exists", () => {
```
chrisdholt: nit: we can remove reference to the method itself. 

"should call `onHorizontalOverflowChange` when initialized when a `onHorizontalOverflowChange` prop is passed" or something like it. 

I don't think you need to simulate load either. You should be able to just render it with the prop and get the values. 

Simulate is somewhat of a hack in this sense. Does this not work without `simulate`?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: Sounds good, thanks for the breakdown.

janechu:
Please implement the [Resize Observer](https://github.com/WICG/ResizeObserver) a helpful resource on this: https://developers.google.com/web/updates/2016/10/resizeobserver. Ideally this should replace the resize listener for window but we should keep it because Resize Observer is only implemented by Chrome. Also add in the readme that is this is necessary a user should apply a polyfill for Resize Observer so that it can work in safari and firefox, at least until they have implemented the Resize Observer. In the code where Resize Observer is added link the relevant issues for firefox and safari:

- https://bugzilla.mozilla.org/show_bug.cgi?id=1272409
- https://bugs.webkit.org/show_bug.cgi?id=157743

And create an issue to revisit this once Firefox and Safari have implemented.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -403,5 +419,55 @@ describe("horizontal overflow", (): void => {
             renderedWithImagesAndNextAndPrevious.instance()["handlePreviousClick"]()
         ).toBe(undefined);
     });
+    test("should add resize event listener to the window", (): void => {
```
janechu: It took me a while to understand what it was these tests were actually doing, I would suggest instead you do:

```ts
test("should add a resize event listener to the window", (): void => {
        let addResizeListenerCount: number = 0;

        /**
         * Mock window.addEventListener to count the number of times
         * a resize event has been added
         */
        window.addEventListener = jest.fn((event: string, callback: any) => {
            if (event === "resize") {
                addResizeListenerCount = addResizeListenerCount ? addResizeListenerCount + 1 : 1;
            }
        });

        expect(addResizeListenerCount).toBe(0);

        const rendered: any = mount(
            <HorizontalOverflow managedClasses={managedClasses}>
                {imageSet1}
            </HorizontalOverflow>
        );

        expect(addResizeListenerCount).toBe(1);
    });
    test("should remove a resize event listener from the window when component unmounts", (): void => {
        let removeResizeListenerCount: number = 0;

        /**
         * Mock window.removeEventListener to count the number of times
         * a resize event has been removed
         */
        window.removeEventListener = jest.fn((event: string, callback: any) => {
            if (event === "resize") {
                removeResizeListenerCount = removeResizeListenerCount ? removeResizeListenerCount + 1 : 1;
            }
        });

        const rendered: any = mount(
            <HorizontalOverflow managedClasses={managedClasses}>
                {imageSet1}
            </HorizontalOverflow>
        );

        expect(removeResizeListenerCount).toBe(0);

        rendered.unmount();

        expect(removeResizeListenerCount).toBe(1);
    });
```

This way you can test that the resize event listener has been added once, and removed once. In this way you are testing that there are 0 resize listeners before mounting, 1 after mounting, similarly the remove event listeners are 0 before mounting, and 1 after unmounting.

packages/fast-components-react-base/src/horizontal-overflow/index.ts
```diff
@@ -2,3 +2,4 @@ import HorizontalOverflow from "./horizontal-overflow";
 
 export default HorizontalOverflow;
 export * from "./horizontal-overflow";
+export * from "./resize-observer";
```
janechu: I would not export this outside of our package

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -367,6 +367,22 @@ describe("horizontal overflow", (): void => {
...
+                className="foo"
+                managedClasses={managedClasses}
+            >
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+
+        renderedWithImages.find("div.foo").simulate("load", {});
+
+        expect(onHorizontalOverflowChangeFunction).toHaveBeenCalledTimes(1);
```
janechu: So I think this is actually the wrong way to do things. The `onHorizontalOverflowChange` should fire when the overflow has changed, same as when the `onScrollChange` is fired. However it is arbitrarily getting hit whenever the overflow *might* have changed. I think this is something that should be kept as a private variable to be updated and checked against, so when event like load etc may trigger a change in overflow, it checks first before firing the callback.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -406,9 +486,9 @@ class HorizontalOverflow extends Foundation<
     };
 
     /**
-     * Handler for the click event fired after next or previous has been clicked
+     * Returns the items width and total width of the scroll region
      */
-    private handleClick(direction: ButtonDirection): void {
+    private getItemsWidthAndTotalWidth(): ItemsWidthAndTotalWidth {
```
janechu: This seems hard to understand, can create two methods `getItemWidths` and `getAvailableWidth`

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -25,6 +28,11 @@ export interface HorizontalOverflowState {
     itemsHeight: number;
 }
 
+export interface ItemsWidthAndTotalWidth {
```
janechu: As per my comment below, I think this is unnecessary and we should not need to have an interface if we just split the method for getting the item widths and available width.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: sure

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -150,26 +168,50 @@ class HorizontalOverflow extends Foundation<
...
+    /**
+     * Observe resize of parent element
+     */
+    private onResizeObserver = (): void => {
+        // TODO #1142 https://github.com/Microsoft/fast-dna/issues/1142
+        // Full browser support imminent
+        // Revisit usage once Safari and Firefox adapt
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1272409
+        // https://bugs.webkit.org/show_bug.cgi?id=157743
+        if (ResizeObserver) {
```
nicholasrice: how is this `if` check working? isn't `ResizeObserver` just a type definition of a class?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -14,6 +14,15 @@ export interface ScrollChange {
     end: boolean;
 }
 
+/**
+ * Horizontal overflow interface for consumers
+ * Includes all properties found in ScrollChange
+ * 'overflow' is true when there are enough items to cause overflow
+ */
+export interface HorizontalOverflowChange extends ScrollChange {
```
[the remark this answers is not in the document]
nalogiudice: It extends the ScrollChange so it includes the start and end and anything else provided by that interface

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -29,6 +38,11 @@ export interface HorizontalOverflowHandledProps extends HorizontalOverflowManage
      */
     scrollDuration?: number;
 
+    /**
+     * Callback for on horizontal overflow change
+     */
+    onHorizontalOverflowChange?: (changeObject: HorizontalOverflowChange) => void;
```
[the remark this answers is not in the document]
chrisdholt: If we have a vertical is the interface the same or different? One might include "left/right" properties and one might not?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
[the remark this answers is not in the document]
nicholasrice: I see, missed that.

Does "overflow" add value? I'm not sure it's worth adding a third property when the first two already communicate that information.

It's probably okay now - but I'm not sure this should extend scroll. scroll might have information like distance that would not be present in an overflow change event. You might consider extending the other direction.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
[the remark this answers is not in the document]
nalogiudice: Sure - it matched onScrollChange better

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: @janechu I realized after removing this check, the solution you supplied was not actually working. The typings allow it to build, but the console outputs a type error `TypeError: _resize_observer__WEBPACK_IMPORTED_MODULE_7__.ResizeObserver is not a constructor`
I've been digging trying to figure out how to solve this, but I feel like I've hit a wall.

Any guidance from you or @nicholasrice would be helpful

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +239,41 @@ class HorizontalOverflow extends Foundation<
...
+     * Handles the resize event
+     */
+    private onWindowResize = (): void => {
+        this.handleHorizontalOverflowChange();
+    };
+
+    /**
+     * Checks if overflow is occuring
+     */
+    private checkOverflow(): boolean {
```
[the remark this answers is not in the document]
nalogiudice: Sure.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +239,41 @@ class HorizontalOverflow extends Foundation<
                 itemsHeight,
             });
         }
+
+        if (this.props.onHorizontalOverflowChange) {
```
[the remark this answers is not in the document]
nalogiudice: Right.

packages/fast-components-react-base/src/horizontal-overflow/README.md
```diff
@@ -2,7 +2,16 @@
 The *horizontal overflow* component accepts items as children. A child can have an optional *slot* property which is either "next" or "previous" which can be used to allow the next/previous click actions. *Horizontal overflow* uses native scrolling behavior.
 
 ### Usage guidance
-The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL.
+The *horizontal overflow* component always initializes at the start of the overflow content &mdash; left side in LTR and right side in RTL. 
+
+### Callbacks
+Several callbacks are supplied to aid the consumer. Use `onScrollChange` to receive if scroll is at the start or end of the overflow set. Use `onHorizontalOverflowChange` to know if there are enough items to cause overflow (this callback also supplies the `onScrollChange` data.) The `onHorizontalOverflowChange` callback uses the new `ResizeObserver` (available Chrome) so it is necessary to apply a polyfill for `ResizeObserver` if functionality is needed in Safari and/or Firefox. Window resize is used as a fallback.
```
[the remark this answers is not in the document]
nalogiudice: Ok. Originally @janechu asked to move this to the README.md. I've added back to the props.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
nalogiudice: I've updated.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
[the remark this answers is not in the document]
nalogiudice: @chrisdholt for backup - we had several conversations about how the overflow callback would include scroll information as well. I thought we all agreed this data is important to the consumer on an overflow change.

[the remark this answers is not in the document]
nicholasrice: Yes. We had several conversations about how these handlers might (and do) have common properties. 

I'm just saying the two events are inherently different and that scrolling is likely (in the future) to contain information about a scroll event that cant be known in an overflow event because overflow events are not necessarily the product of scrolling , so the relationship seems backwards to me.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nicholasrice: yea - thats cause the `ResizeObserver` you're importing is just a typings for the DOM resize observer.

You would check for it's existence with `window.ResizeObserver`.

nalogiudice: Right, but even to add that check I need to add it to the global namespace, because `window.ResizeObserver` is unknown.

The issue with the typings still remains and I get the same error. I am still unsure how to solve.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -29,8 +38,16 @@ export interface HorizontalOverflowHandledProps extends HorizontalOverflowManage
      */
     scrollDuration?: number;
 
+    /**
+     * Callback for on overflow change
+     * Use `onOverflowChange` to know if there are enough items to cause overflow
+     *
```
janechu: nit: extra line

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -29,8 +38,16 @@ export interface HorizontalOverflowHandledProps extends HorizontalOverflowManage
...
+     */
+    onOverflowChange?: (changeObject: OverflowChange) => void;
```
janechu: nit: overflowObject

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -123,8 +144,12 @@ class HorizontalOverflow extends Foundation<
         if (canUseDOM() && this.horizontalOverflowItemsRef.current) {
             this.horizontalOverflowItemsRef.current.addEventListener(
                 "scroll",
-                this.throttled
+                this.throttledScroll
             );
+            window.addEventListener("resize", this.throttledResize);
+            if (this.overflow !== this.isOverflow()) {
```
janechu: This will:
A. Always fire because this.overflow has not been set
B. Does not indicate a change in overflow since no children will have been loaded during the constructor phase

Please remove this if... else block and initially set this.overflow = false.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -150,26 +176,50 @@ class HorizontalOverflow extends Foundation<
...
-            return;
+        if (typeof this.props.onScrollChange === "function") {
+            this.props.onScrollChange(this.getScrollChangeData());
         }
+    };
 
+    /**
+     * Observe resize of parent element
+     */
+    private onResizeObserver = (): void => {
```
janechu: This should not be its own method, I would instead place this in componentDidMount as once the component is mounted you will have access to the ref.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +247,42 @@ class HorizontalOverflow extends Foundation<
                 itemsHeight,
             });
         }
+
+        this.handleOverflowChange();
```
janechu: This assumes that the overflow is changing everytime an item is loaded, there should be a check to see if the overflow has actually changed.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -197,6 +247,42 @@ class HorizontalOverflow extends Foundation<
...
+    /**
+     * Callback for the horizontal overflow change
+     */
+    private handleOverflowChange = (): void => {
+        this.overflow = this.isOverflow();
+
+        if (typeof this.props.onOverflowChange === "function") {
+            this.props.onOverflowChange({
+                ...this.getScrollChangeData(),
+                overflow: this.isOverflow(),
```
janechu: you're calling this.isOverflow twice, you can just use this.overflow here.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: You can try extending the window type and add ResizeObserver to it the explicitly try (window as WindowWithResizeObserver).ResizeObserver

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -403,5 +419,53 @@ describe("horizontal overflow", (): void => {
...
     });
+    test("should add a resize event listener to the window", (): void => {
+        let addResizeListenerCount: number = 0;
+
+        /**
+         * Mock window.addEventListener to count the number of times
+         * a resize event has been added
+         */
+        window.addEventListener = jest.fn((event: string, callback: any) => {
+            if (event === "resize") {
```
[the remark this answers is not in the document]
nalogiudice: @nicholasrice  Originally, I had something closer to what you describe and @janechu  asked me to change to the above code. I _think_ I've found an in-between. Please let me know.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nalogiudice: Should all be resolved now thanks @janechu

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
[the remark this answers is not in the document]
nalogiudice: Resolving for now...

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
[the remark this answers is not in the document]
nicholasrice: Hmm, I'm not sure - using spy's is pretty standard practice. This looks good.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -14,6 +14,15 @@ export interface ScrollChange {
     end: boolean;
 }
 
+/**
+ * Overflow interface for consumers
+ * Includes all properties found in ScrollChange
+ * 'overflow' is true when there are enough items to cause overflow
+ */
+export interface OverflowChange extends ScrollChange {
```
nicholasrice: I don't feel a single overflow value is the most intuitive API:

I would prefer to see "overflowStart" and "overflowEnd" values explicitly. With this approach, my code to hide/show flippers might look something like:

```
let showStartFlipper = true;
let showEndFlipper = true;

if (overflowChange.start) {
    showStartFlipper = false;
}

if (overflowChange.end || !overflowChange.overflow) {
    showFlipperEnd = false;
}
```

What is the reason we would land this instead of simply:

```
let showStartFlipper = overflowChange.overflowStart;
let showEndFlipper = overflowChange.overflowEnd;
```

Apologies if I'm being thick - we can sync on a call if that is preferred.

nalogiudice: Updated per our convo.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -150,26 +202,44 @@ class HorizontalOverflow extends Foundation<
...
+            ) {
+                this.handleOverflowChange();
+            }
         }
+    };
 
+    /**
+     * Get the scroll change data
+     */
+    private getScrollChangeData = (): ScrollChange => {
```
janechu: Can we change this name since it is being used by both the scroll and overflow callbacks to reflect the data it's actually getting back, I think `getItemsPositionWithinContainer` unless there is a better name we can come up with.

nalogiudice: Yeah I had the same thought on changing this name. @nicholasrice for input as well.

nalogiudice: Agreed on `getPositionData`

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.props.ts
```diff
@@ -8,12 +8,25 @@ import {
  * Scroll interface for consumers
  * 'start' is when the horizontal overflow scroll is all the way left in LTR (all the way right in RTL)
  * 'end' is when the horizontal overflow scroll is all the right in LTR (all the way left in RTL)
+ * TODO #1177: Change interface property names to be more meaningful
  */
-export interface ScrollChange {
```
nicholasrice: You need to keep the interface name the same, otherwise anyone import this will get a module resolution error.

nalogiudice: I mentioned this to @janechu but she said that interface name weren't contractual, but only the properties themselves...

chrisdholt: +1 on deprecate and export a new one.

nalogiudice: Added

PR #1138, opened 2018-11-14
Users/scomea/action trigger

scomea:
Adding action trigger to MSFT React components.  

Also fixed a typo in button component.


packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
```diff
@@ -0,0 +1,98 @@
...
+                    ButtonAppearance[ActionTriggerAppearance[this.props.appearance]]
+                }
+                jssStyleSheet={actionTriggerButtonOverrides}
+            >
+                <div
+                    slot="before"
+                    className={get(this.props, "managedClasses.actionTrigger_glyph")}
+                >
+                    {this.props.glyph}
+                </div>
```
chrisdholt: Instead of a `<div>`, this should really be a `span`. However, I don't think it's necessary if we use a render prop. With a render prop, we can inform people that we will be passing a className to style the component; that removes the need for a generic outer wrapper like this. If they opt-in to the className, we can apply the class and set styles to whatever "thing" they pass without having to wrap this.

With a render prop, this might look like:
```
{ this.props.glyph(get(this.props, "managedClasses.actionTrigger_glyph")) }
```

packages/fast-components-react-msft/src/action-trigger/examples.data.tsx
```diff
@@ -0,0 +1,56 @@
...
+        glyph: glyphExample,
+    },
+    data: [
+        {
+            appearance: ActionTriggerAppearance.primary,
+            children: "Primary action trigger",
+            href: testDestination,
+            glyph: glyphExample,
+            "data-sketch-symbol": "Action trigger - primary",
+        } as any, // TODO https://github.com/Microsoft/TypeScript/issues/6579
```
chrisdholt: This reads like the TODO is on the object, but that doesn't make sense to me as the linked issue is about a regex validated string issue. Can you elaborate on the TODO for me?

packages/fast-components-styles-msft/src/action-trigger/index.ts
```diff
@@ -0,0 +1,156 @@
...
+        actionTrigger_glyph: {
+            display: "inline-block",
+            position: "relative",
+            width: "8px",
+            [applyLocalizedProperty("marginRight", "marginLeft", direction)]: "6px",
+            marginTop: direction === Direction.ltr ? "4px" : "0",
+        },
+        actionTrigger__primary: {
+            "& $actionTrigger_glyph": {
+                fill: color,
```
chrisdholt: Question: What happens with these fill property values if the element is not an SVG?

packages/fast-components-react-msft/src/action-trigger/examples.data.tsx
scomea: I copied it over from call to action's example data file figuring that if it applied there it would also apply to the same code in action trigger.

chrisdholt: Interesting - I'll see if I can track down who added this. Casting the same thing in call to action as `CallToActionProps` also eliminated the type error. All that aside, we should ensure the issue is valid for this scenario as well (but I totally get the logic)

packages/fast-components-styles-msft/src/action-trigger/index.ts
scomea: It's inert as far as I can tell, and elements that use "color" inherit the color property that button applies to it's content so seems to do the right thing.

packages/fast-components-react-msft/src/action-trigger/action-trigger.schema.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    "reactProperties": {
+        "children": {
+            "title": "Building blocks",
+            "type": "children",
+            "defaults": [
+                "text"
+            ]
+        },
+        "glyph": {
```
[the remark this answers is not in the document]
scomea: Glyph is no longer included in that doc, created #1143 "Update action-trigger schema to include 'glyph' render prop" to track.

packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
scomea: So I changed the wrapping element to a span but kept it in order to assign the glyph to the "before" slot.  Lumping it in the the children messed up the vertical alignment.  (and also if we aren't going to use slots we should be consistent between similar controls, like call to action in this case).

chrisdholt: How is this being lumped in with children if we are using a render prop? This is a net-new component and so the API should reflect the long-term API we would like to see. We are moving to render props; a user should not be concerned with the internals (we use the render prop and pass to another component "slot" property). 

It's a correct statement that we are moving away from slots, but in my opinion, we should not:

1. Use an API we intend to deprecate (slots)
2. Not do the right thing until we change all other component API's

scomea: Got rid of slots, override button base styles to make content region a flex container so items can vertically align properly.

packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+                />
+            );
+        }).not.toThrow();
+    });
+
+    test("should implement unhandledProps", () => {
+        const handledProps: ActionTriggerProps = {
+            managedClasses,
+            href,
+            glyph: null,
```
chrisdholt: interesting - isn't glyph required?

packages/fast-components-react-msft/src/action-trigger/README.md
```diff
@@ -0,0 +1,11 @@
+# Action trigger
+Action trigger performs an operation that is positioned inline on the page. Use action trigger for taking an action or for interacting with something like a file or media resource. Examples include “Save,” “Edit,” or “Play.” Using an action trigger keeps the user within the current context rather than forcing them to navigate to a different page. Thus, use an action trigger as part of an interactive in-page experience.  Action trigger is built using a host element such as a button or an anchor, and can also be used in a group of elements. Typically, action triggers are used inside interactive tables or related to other interactive content. This means that the trigger should be co-located with the positioning of the material it will be acting upon. There is no provision for displaying alternate labels to indicate a change of state — use the action toggle if you need this capability.
```
nalogiudice: Please _italicize_ each usage of _action trigger_ in the documentation.

packages/fast-components-react-msft/src/action-trigger/action-trigger.schema.json
```diff
@@ -0,0 +1,37 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Action trigger",
+    "description": "An action trigger component's schema definition.",
+    "type": "object",
+    "id": "action-trigger",
+    "properties": {
+        "disabled": {
```
nalogiudice: Is there no way to alter the glyph in the side pane?

packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
nalogiudice: I feel like we should have at least one test that includes a glyph...

packages/fast-components-react-msft/src/action-trigger/action-trigger.schema.json
scomea: Not at the moment as we don't yet have an implementation for render props here.  Tracked by https://github.com/Microsoft/fast-dna/issues/1143

packages/fast-components-react-msft/src/action-trigger/examples.data.tsx
scomea: Since nothing is specifically broken I think we should remove it.

packages/fast-components-react-msft/src/action-trigger/action-trigger.props.ts
```diff
@@ -0,0 +1,49 @@
...
+        ActionTriggerManagedClasses {
+    /**
+     * The action trigger appearance
+     */
+    appearance?: ActionTriggerAppearance;
+
+    /**
+     * The action trigger glyph render prop
+     */
+    glyph?: (className: string) => React.ReactNode;
```
chrisdholt: @scomea weren't we going to require glyph as this is part of the design pattern?

scomea: Yes. Could have sworn I'd changed that already.

packages/fast-components-react-msft/src/action-trigger/action-trigger.props.ts
```diff
@@ -0,0 +1,49 @@
...
+        ActionTriggerManagedClasses {
+    /**
+     * The action trigger appearance
+     */
+    appearance?: ActionTriggerAppearance;
+
+    /**
+     * The action trigger glyph render prop
+     */
+    glyph: (className: string) => React.ReactNode;
```
nicholasrice: `className` should be optional here if it is coming from JSS

nicholasrice: nm, this isn't a base component. This will always be here.

nalogiudice: Should glyph be optional?

packages/fast-components-class-name-contracts-msft/src/action-trigger/index.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * The class name contract for the action trigger component
+ */
+export interface ActionTriggerClassNameContract {
+    /**
+     * The root of the action trigger component
+     */
+    actionTrigger?: string;
```
nalogiudice: Should this be optional?

packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
```diff
@@ -0,0 +1,209 @@
...
+                />
+            );
+        }).not.toThrow();
+    });
+
+    test("should throw if glyph prop is not provided", () => {
+        expect(() => {
+            shallow(
+                <MSFTActionTrigger
+                    /* tslint:disable-next-line */
```
nalogiudice: Do you need all of the `tslint:disables`?

packages/fast-components-styles-msft/src/action-trigger/index.ts
```diff
@@ -0,0 +1,153 @@
...
+        actionTrigger__outline: {
+            [applyLocalizedProperty("marginLeft", "marginRight", direction)]: "-10px",
+            "& $actionTrigger_glyph": {
+                fill: outlineColor,
+            },
+            "&$actionTrigger__disabled $actionTrigger_glyph": {
+                fill: outlineDisabledColor,
+            },
+        },
+        actionTrigger__disabled: {},
```
nalogiudice: Should the `fill: outlineDisabledColor,` actually be here?

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -254,15 +254,15 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = (
...
                 color: outlineColor,
                 borderColor: outlineBorderColor,
-                ...applyTransaprentBackground(),
+                ...applyTransparentBackground(),
             },
             "&:focus": {
-                ...applyTransaprentBackground(),
+                ...applyTransparentBackground(),
                 borderColor: outlineBorderColor,
                 boxShadow: `inset 0 0 0 1px ${outlineBorderColor}`,
```
nalogiudice: @chrisdholt should we be using `ShadowConfig` here?

packages/fast-components-react-msft/src/action-trigger/action-trigger.props.ts
scomea: By design.

packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
scomea: Got rid of most

packages/fast-components-styles-msft/src/action-trigger/index.ts
scomea: Yes, otherwise an svg glyph won't pick up the disabled color.

packages/fast-components-styles-msft/src/button/index.ts
scomea: Note: this was a "drive by" fixing of a typo in the button component.  Please open bugs for any other button issues. :-)

packages/fast-components-class-name-contracts-msft/src/action-trigger/index.ts
nalogiudice: Yeah, I see that... hey @nicholasrice why do we do it that way? Won't the root MSFT class _always_ be there in a styled component?

chrisdholt: I mean - technically, this only becomes styled once we style it. The component as it exists as a class in the .tsx file can technically be used without style.

PR #1145, opened 2018-11-16
fix: update CSS editor components to extend the Foundation component

### LINKED ISSUE #1144: Update the CSS Editor to extend Foundation
Update the CSS Editor components to extend the Foundation component.
This will ensure future enhancements to the Foundation will to be passed to these components.

janechu:
Resolves #1144

This now includes a breaking change because of the inferred use of an onChange prop, it is now specific to what CSS values it is changing, in this case position so the rename will change it to onPositionUpdate.


packages/fast-css-editor-react/src/editor.props.ts
```diff
@@ -0,0 +1,12 @@
+import { ManagedClasses } from "@microsoft/fast-jss-manager";
+import { CSSEditorClassNameContract } from "./editor.style";
+import { CSSPositionHandledProps } from "./";
+
+export interface CSSEditorUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+export type CSSEditorComponentsHandledProps = CSSPositionHandledProps &
```
nicholasrice: do you actually need this type definition? Can't you just define `CSSEditorHandledProps` as these two types?

packages/fast-css-editor-react/src/position/position.props.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface CSSPositionUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+export interface CSSPositionHandledProps
+    extends ManagedClasses<CSSPositionClassNameContract> {
+    position?: PositionValue;
+    left?: string;
+    right?: string;
+    top?: string;
+    bottom?: string;
+    onChange?: (positionValues: any) => void;
```
nicholasrice: Can you type this more accurately?

packages/fast-css-editor-react/src/editor.props.ts
janechu: this is going to be extended for every component that is added to the editor, so for example one we will be adding after this will be CSSMarginHandledProps.

packages/fast-css-editor-react/src/position/position.props.ts
janechu: Done, a lot of other updates happened as well so please take another look!

packages/fast-css-editor-react/src/position/position.tsx
```diff
@@ -1,50 +1,43 @@
...
+        top: void 0,
+        bottom: void 0,
+        left: void 0,
+        right: void 0,
+        onPositionUpdate: void 0,
+        managedClasses: void 0,
+    };
 
-    public render(): JSX.Element {
+    public render(): React.ReactElement<HTMLDivElement> {
```
nicholasrice: We were running into some issues with returning React.ReactElement types - react.d.ts types this as `React.ReactNode` so we should probably align

PR #1148, opened 2018-11-17
fix: correct misspelled button enum value

### LINKED ISSUE #1147: Default button renders when `justified` is selected as part of configuration on MSFT docs site
On the MSFT documentation site (https://msft-docs.azurewebsites.net/components/button/) changing a button appearance value to `justified` instead renders a default button. 

Initially, the correct value shows up, but upon changing this in the editor, the correct class is not being applied and a default button is rendered.

chrisdholt:
<body>
closes #1147 
<footer>
 

### DIFF
--- a/packages/fast-components-react-msft/src/button/button.schema.json
+++ b/packages/fast-components-react-msft/src/button/button.schema.json
@@ -17,7 +17,7 @@
             "title": "Appearance",
             "type": "string",
             "enum": [
-                "justfied",
+                "justified",
                 "lightweight",
                 "outline",
                 "primary"


packages/fast-components-react-msft/src/button/button.schema.json
```diff
@@ -17,7 +17,7 @@
             "title": "Appearance",
             "type": "string",
             "enum": [
```
nicholasrice: This is spelled correctly in the component though, right?

chrisdholt: Correct - the reason it showed up correctly the first time was because it was referencing the Enum and the schema was an independent string. Did a search as well and checked similar components (CTA)

PR #1151, opened 2018-11-20
fix: add a check for a theme, if it is not available use an empty object

### LINKED ISSUE #1150: The development site errors when no theme is provided
A runtime error occurs when no theme is provided to the development site.

janechu:
Closes #1150


packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -612,7 +612,10 @@ class Site extends React.Component<
                         <DesignSystemProvider
                             designSystem={
                                 this.state.componentView === ComponentViewTypes.examples
-                                    ? { backgroundColor: this.state.theme.background }
+                                    ? {
```
nicholasrice: we could also always initialize the state with an empty theme object

PR #1153, opened 2018-11-20
fix: JSSManager should create stylesheets prior to initial render

### LINKED ISSUE #1149: JSS manager causes components to render twice
Because stylesheets are not created until `componentDidMount`, each styled component has to go through two render cycles before classes are passed to the wrapped component. We should factor the manager in a way where styles are rendered prior to the first render call.

nicholasrice:
fixes #1149

### DIFF
--- a/packages/fast-jss-manager-react/src/jss-manager.spec.tsx
+++ b/packages/fast-jss-manager-react/src/jss-manager.spec.tsx
@@ -304,7 +304,7 @@ describe("mergeClassNames", (): void => {
         expect(mergeClassNames(undefined, "bar")).toBe("bar");
     });
 
-    test("should return a space-seperated string containing both arguments if both arguments are strings", (): void => {
+    test("should return a space-separated string containing both arguments if both arguments are strings", (): void => {
         expect(mergeClassNames("foo", "bar")).toBe("foo bar");
     });
 });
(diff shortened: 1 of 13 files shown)


packages/fast-jss-manager-react/src/jss-manager.tsx
```diff
@@ -38,6 +38,16 @@ export type ManagedJSSProps<T, S, C> = Pick<
...
 
+export function mergeClassNames(a: string | void, b: string | void): string {
+    if (typeof a === "string" && typeof b === "string") {
+        return a.concat(" ", b);
+    } else if (typeof a === "string") {
+        return a;
+    } else if (typeof b === "string") {
+        return b;
+    }
+}
```
janechu: if this function always has to return a string, should it, outside of the if...else blocks `return "";`? Otherwise the return value is not string its potentially `string | void`

packages/fast-jss-manager-react/src/jss-manager.spec.tsx
```diff
@@ -265,7 +268,45 @@ describe("The JSSManager", (): void => {
...
+    test("should return undefined if neither param is a string", (): void => {
+        expect(mergeClassNames(undefined, undefined)).toBe(undefined);
+    });
+
+    test("should return a string if either argument is a string", (): void => {
+        expect(mergeClassNames("foo", undefined)).toBe("foo");
+        expect(mergeClassNames(undefined, "bar")).toBe("bar");
+    });
+
+    test("should return a space-seperated string containing both arguments if both arguments are strings", (): void => {
```
nalogiudice: nit: separated

PR #1159, opened 2018-11-21
fix: hypertext hover and focus do not work in Edge

eljefe223:
Hypertext hover and focus do not work in Edge

closes: #1159


packages/fast-components-styles-msft/src/hypertext/index.ts
```diff
@@ -1,31 +1,29 @@
...
-            color: ensureBrandNormal,
-            "&:hover, &:focus": {
-                ...applyHypertextBorder(2),
+        hypertext: {
+            outline: "none",
+            textDecoration: "none",
+            color: ensureForegroundNormal,
+            "&:link, &:visited": {
+                borderBottom: `1px solid ${ensureBrandNormal(config)}`,
+                color: ensureBrandNormal,
```
chrisdholt: @eljefe223 do we not need to pass config to this instance of `ensureBrandNormal`?

PR #1162, opened 2018-11-26
feat: add action toggle component

### LINKED ISSUE #1131: Create "action toggle" as a new composed MSFT component
Action toggle is a special-use button in the MSFT design language that allows users to switch between two related states, and is often used for activate/deactivate functions. Action toggle can be grouped together for related functions. Action toggle’s intended use is for interacting with a particular element or region on the page. Examples are switching between “Play” and “Pause,” “Start” and “Stop,” or “Connect” and “Disconnect.” Action toggle keeps the user within a functional context rather than navigating somewhere else.

**Visually:**
Action toggle is ultimately different "subset" visual states of either the MSFT action trigger or button.

*Untoggled*
action_toggle_untoggled

*Toggled*
action_toggle_toggled

**Props:**
Action toggle should expose first class props to manage an aria-label as well as the aria-pressed state.

Outstanding questions:
- Should this be a *controlled* component only? (No "uncontrolled" behavior)

scomea:
feat: add action toggle to MSFT React components.
<body>
Closes #1131 - Create "action toggle" as a new composed MSFT component
<footer>

### DIFF
--- a/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
@@ -1,0 +1,52 @@
+/**
+ * The class name contract for the action toggle component
+ */
+export interface ActionToggleClassNameContract {
+    /**
+     * The root of the action toggle component
+     */
+    actionToggle?: string;
+    /**
+     * The selected modifier
+     */
+    actionToggle__selected?: string;
+    /**
+     * The action toggle selected glyph
+     */
+    actionToggle_selectedGlyph?: string;
+
+    /**
+     * The action toggle unselected glyph
+     */
+    actionToggle_unselectedGlyph?: string;
+
+    /**
+     * The primary appearance modifier
+     */
+    actionToggle__primary?: string;
+
+    /**
+     * The outline appearance modifier
+     */
+    actionToggle__outline?: string;
+
+    /**
+     * The lightweight appearance modifier
+     */
+    actionToggle__lightweight?: string;
+
+    /**
+     * The justified appearance modifier
+     */
+    actionToggle__justified?: string;
+
+    /**
+     * The disabled modifier
+     */
+    actionToggle__disabled?: string;
+
+    /**
+     * The sngle element modifier
+     */
+    actionToggle__singleElement?: string;
+}
--- a/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
@@ -26,19 +26,9 @@ export interface ActionToggleClassNameContract {
     actionToggle__primary?: string;
 
     /**
-     * The outline appearance modifier
-     */
-    actionToggle__outline?: string;
-
-    /**
      * The lightweight appearance modifier
      */
     actionToggle__lightweight?: string;
-
-    /**
-     * The justified appearance modifier
-     */
-    actionToggle__justified?: string;
 
     /**
      * The disabled modifier
@@ -48,5 +38,5 @@ export interface ActionToggleClassNameContract {
     /**
      * The single element modifier
      */
-    actionToggle__singleElement?: string;
+    actionToggle__hasGlyphAndContent?: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
@@ -5,7 +5,7 @@ export interface ActionToggleClassNameContract {
     /**
      * The root of the action toggle component
      */
-    actionToggle?: string;
+    actionToggle: string;
     /**
      * The selected modifier
      */
--- a/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
@@ -46,7 +46,7 @@ export interface ActionToggleClassNameContract {
     actionToggle__disabled?: string;
 
     /**
-     * The sngle element modifier
+     * The single element modifier
      */
     actionToggle__singleElement?: string;
 }
--- a/packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts
+++ b/packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts
@@ -1,0 +1,69 @@
+import * as React from "react";
+import { Omit } from "utility-types";
+import {
+    ActionToggleClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-msft";
+import {
+    ButtonAppearance,
+    ButtonHandledProps,
+    ButtonManagedClasses,
+    ButtonUnhandledProps,
+} from "../button/button.props";
+
+export enum ActionToggleAppearance {
+    justified = ButtonAppearance.justified,
+    lightweight = ButtonAppearance.lightweight,
+    outline = ButtonAppearance.outline,
+    primary = ButtonAppearance.primary,
+}
+
+export interface ActionToggleManagedClasses
+    extends ManagedClasses<ActionToggleClassNameContract> {}
+export interface ActionToggleHandledProps
+    extends Omit<ButtonHandledProps, "appearance" | keyof ButtonManagedClasses>,
+        ActionToggleManagedClasses {
+    /**
+     * The action toggle appearance
+     */
+    appearance?: ActionToggleAppearance;
+
+    /**
+     * The action toggle selected glyph render prop
+     */
+    selectedGlyph: (className: string) => React.ReactNode;
+
+    /**
+     * The action toggle usselected glyph render prop
+     */
+    unselectedGlyph: (className: string) => React.ReactNode;
+
+    /**
+     * The action toggle selected text
+     */
+    selectedText?: string;
+
+    /**
+     * The action toggle unselected text
+     */
+    unselectedText?: string;
+
+    /**
+     * The onChange event handler
+     */
+    onChange?: (event: React.MouseEvent<HTMLElement>) => void;
+
+    /**
+     * The action toggle disabled property
+     */
+    disabled?: boolean;
+
+    /**
+     * The action toggle selected state
+     */
+    selected?: boolean;
+}
+
+/* tslint:disable-next-line:no-empty-interface */
+export interface ActionToggleUnhandledProps extends ButtonUnhandledProps {}
+export type ActionToggleProps = ActionToggleHandledProps & ActionToggleUnhandledProps;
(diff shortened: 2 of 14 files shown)


packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
```diff
@@ -0,0 +1,53 @@
...
+        },
+        "selectedText": {
+            "title": "Selected text",
+            "type": "string"
+        },
+        "unselectedText": {
+            "title": "Unselected text",
+            "type": "string"
+        },
+        "selectedARIALabel": {
```
nalogiudice: I would just remove `ARIA` all together from the prop names. If we left it in it would be `Aria` but I think we should remove

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,246 @@
...
+        };
+
+        const rendered: any = mount(
+            <ActionToggle className={"custom-class-name"} {...props} />
+        );
+
+        expect(rendered.find("button").prop("className")).toContain("custom-class-name");
+    });
+
+    // tslint:disable-next-line:max-line-length
```
nalogiudice: I think all of these `// tslint:disable-next-line:max-line-length` are unneeded

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
scomea: So there are two text fields for "selected" - the one that populates the actual label and the one used for accessibility.  We still need the accessibility string if we have a glyph only version of this component.  How should we differentiate between the two name wise?

awentzel: I'm not sure about this one.  Feels like there are specific props for accessibility and calling this out could be useful.  Otherwise, context could be lost for "selectedLabel" vs "selectedARIALabel". Also, this can be left in as "ARIA" if we think it's helpful since I would consider it a well known industry acronym.

packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts
```diff
@@ -0,0 +1,79 @@
...
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-msft";
+import {
+    ButtonAppearance,
+    ButtonHandledProps,
+    ButtonManagedClasses,
+    ButtonUnhandledProps,
+} from "../button/button.props";
+
+export enum ActionToggleAppearance {
```
chrisdholt: I don't believe appearance would be exposed in this scenario. I'm struggling to understand what purpose it would serve considering these appearances change based on state.

This component has two states; selected/toggled and unselected/untoggled. So, we either need to provide selected and unselected styles for each appearance, which would be completely new designs - or we need to base this off the current design in the task.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,239 @@
...
+        } else {
+            if (
+                isNullOrWhiteSpace(this.props.unselectedText) ||
+                isNullOrUndefined(this.props.unselectedGlyph)
+            ) {
+                return true;
+            } else {
+                return false;
+            }
+        }
```
chrisdholt: Can you provide some insight on the usefulness of this? Looking to understand as it's not clear to me.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,239 @@
...
+                {this.renderLabel()}
+                {this.props.children}
+            </Button>
+        );
+    }
+
+    /**
+     * Returns the appropriate ARIA label
+     */
+    public renderARIALabel(): string {
```
chrisdholt: Curious as to @nicholasrice and @janechu's thoughts on this, but I wonder if it doesn't make more sense to simply manage this as a completely controlled component. On change we may would expect new values for glyph and children. Does it makes sense for this to support controlled behavior only?

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,239 @@
...
+                disabled={this.props.disabled}
+                onClick={this.handleToggleChange}
+                aria-label={this.renderARIALabel()}
+                appearance={
+                    ButtonAppearance[ActionToggleAppearance[this.props.appearance]]
+                }
+                jssStyleSheet={actionToggleButtonOverrides}
+            >
+                {this.renderGlyph()}
+                {this.renderLabel()}
```
chrisdholt: What is the difference between label and children? I think this should really just support children, which leads to the question as to whether this should be a controlled component only

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,239 @@
...
+import Foundation, { HandledProps } from "@microsoft/fast-components-foundation-react";
+import { Button, ButtonAppearance } from "../button";
+import {
+    ActionToggleAppearance,
+    ActionToggleHandledProps,
+    ActionToggleProps,
+    ActionToggleUnhandledProps,
+} from "./action-toggle.props";
+import { actionToggleButtonOverrides } from "@microsoft/fast-components-styles-msft";
+import { isNullOrWhiteSpace } from "../../../fast-web-utilities/dist";
```
nicholasrice: importing from a dist directory will fail in the published package - you need to import this from a named import: `@microsoft/fast-web-utilities`

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
nicholasrice: my thoughts on this in your other thread

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
nicholasrice: I think I'm on board with it being controlled.

However, I don't think that means we shouldn't have selected/unselected props (I think that is what you're proposing above with only accepting children). Just taking children, I now need to manage when what child text is applied which at some level defeats the point of the component. Take that further to glyphs and aria-labels, and the component is pretty much useless.

I think the only "controlled" prop should be `selected` - the other info is provided and the component determines what to display based on that state.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,239 @@
...
+                break;
+        }
+
+        return super.generateClassNames(classNames);
+    }
+
+    /**
+     * Checks to see if the toggle is displaying both glyph and text or not
+     */
+    private isSingleElement(): boolean {
```
nicholasrice: maybe name this "`hasGlyphAndText`"?

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,239 @@
...
+        if (this.state.selected) {
+            if (
```
nicholasrice: not 100% sure but it might fix your complexity issue to do this instead of an if statement:

```
return this.state.selected
?  isNullOrWhiteSpace(this.props.selectedText) || isNullOrUndefined(this.props.selectedGlyph)
: isNullOrWhiteSpace(this.props.unselectedText) || isNullOrUndefined(this.props.unselectedGlyph)

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx (hunk 5 above)
Comment (chrisdholt) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx]: Sounds fair. 

If we are limiting this to selected/unselected text (strings) I think we convolute by passing children as well. If this is a pattern/subset/more restricted design system component - does it need "children" AND selected/unselected text? If I want to do something special outside of this pattern I can use the button component to do so. 

If we want to support passing components here as well and that is why we still support children, I think we would be better served somehow making the props selectedContent and unselectedContent and accepting a generic React node. Thoughts?

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json (hunk 1 above)
Comment (nalogiudice) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json]: In the past we have just labeled it like `selectedLabel` - we have never used Aria. So we would have `text` and `label`, and it was inferred that `text` was visual and `label` was not. 

I could see this being confusing so I am open to renaming BUT we have other props in the library that deal with Aria that are _not_ labeled this way, and that in turn is confusing.

Comment (chrisdholt) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json]: We should align with the current props we have in existence for this because we talked over those for hours.

In my opinion we can document what and when this is useful for. For instance, the only reason this existed in previous implementations was when there was just a glyph. Because we have selected/unselected, we need to enumerate this, but it should really *only* be used if there is just a glyph.

Comment (scomea) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json]: We ended up going with "label" for the accessibility stuff and "content" for the regular content (which is now a react node and not just a string).

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts (hunk 3 above)
Comment (scomea) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts]: Changed to be hard-coded "primary" for selected state and "lightweight" for unselected.

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx (hunk 5 above)
Comment (scomea) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx]: No longer support children.  Content is a react node rather than just a string.

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx (hunk 6 above)
Comment (scomea) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx]: after discussion we ended up not supporting children and content can be a react node rather than just a string

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx (hunk 8 above)
Comment (scomea) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx]: "hasGlyphandContent", since content can be something other than text now (react node).

File: packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx (hunk 4 above)
Comment (scomea) [reply packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx]: We need to add a margin between glyph and content when both are present, but not when there is only one as stuff won't be centered.  This function evaluates that so we can add the right classes to add that margin.

File: packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts [hunk 10]
```diff
@@ -0,0 +1,42 @@
+/**
+ * The class name contract for the action toggle component
+ */
+export interface ActionToggleClassNameContract {
+    /**
+     * The root of the action toggle component
+     */
+    actionToggle: string;
```
nicholasrice: I think this should be optional

chrisdholt: yes, 100%. Technically all of these can still exist as *unstyled* components per our pattern

scomea: I got the opposite feedback on the previous Action Trigger PR on the same issue:
image

PR #1163, opened 2018-11-27
chore: split documentation site for msft components into Foundation and Pattern categories

janechu:


packages/fast-components-react-msft/app/examples.ts
```diff
@@ -1,68 +1,74 @@
...
-export { Dialog };
+Foundation.Dialog = Dialog;
 
 import DividerExample from "../src/divider/examples.data";
-export { DividerExample };
+Foundation.DividerExample = DividerExample;
 
 import FlipperExample from "../src/flipper/examples.data";
-export { FlipperExample };
+Pattern.FlipperExample = FlipperExample;
```
chrisdholt: @janechu I think this is up for discussion in my mind. It's built off base button but in the sense of pattern vs foundation in the MSFT world, this is more in line with an MSFT foundational component...it is the exception, not the rule. But I think this one requires some discussion potentially.

nicholasrice: I can see how this could be a "pattern". An action-trigger is a foundation component - it can be configured in a variety of ways for a variety of purposes. Flipper is specific, it is for a specific thing and has a specific purpose.

chrisdholt: Action trigger is not a foundation component - did you mean "pattern"?

chrisdholt: Signing off for now, but we should revisit this because I don't know that the current buckets are extensible

PR #1169, opened 2018-11-28
feat: ensure all utilities are exported in the package distribution

### LINKED ISSUE #1168: fast-components-styles-msftdoes not export all utilities
We should export all of our utilities so that ingesting projects can build experiences using the same toolkit that the style sheets themselves use.

nicholasrice:
closes #1168

### DIFF
--- a/packages/fast-components-styles-msft/src/index.ts
+++ b/packages/fast-components-styles-msft/src/index.ts
@@ -122,7 +122,10 @@ export { TypographyStyles };
  */
 export * from "./utilities/acrylic";
 export * from "./utilities/breakpoints";
+export * from "./utilities/colors";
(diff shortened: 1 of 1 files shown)


packages/fast-components-styles-msft/src/index.ts
```diff
@@ -122,7 +122,10 @@ export { TypographyStyles };
  */
 export * from "./utilities/acrylic";
 export * from "./utilities/breakpoints";
+export * from "./utilities/colors";
```
nalogiudice: Do we want to use the export checker utility for this?

nicholasrice: I don't think it will work because these are all named exports.

PR #1197, opened 2018-12-09
fix: update components to handle undefined className

fmamic:
<body>

<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/image/image.tsx
+++ b/packages/fast-components-react-base/src/image/image.tsx
@@ -31,7 +31,7 @@ class Image extends Foundation<ImageHandledProps, ImageUnhandledProps, {}> {
      * Renders the component
      */
     public render(): React.ReactElement<HTMLImageElement | HTMLPictureElement> {
-        let className: string = get(this.props, "managedClasses.image");
+        let className: string = get(this.props, "managedClasses.image", "");
 
         if (!this.props.children) {
             return (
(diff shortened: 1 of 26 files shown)


fmamic:
Duplication was existing in the generateClassNames() and now codeclimte is reporting it because I edited duplicated code. 
I have few ideas how to resolve it but I would like to check with you @janechu  and get your input before refactoring.

chrisdholt:
@fmamic I’d prefer to have the duplication you see in the test files extracted to a new PR. If for some reason we need to revert, both changes will be pulled.

fmamic:
Hi @chrisdholt, I updated changes as per your request. PR contains only required changes and tests for the bug fix.

chrisdholt:
@fmamic thanks for updating! Can you clean up the additional calls to .trim()? Aside from that everything looks good. While I would love a typedef for the className tests, I don't know where that would go and I don't think it needs to block the PR. Can you pull the extraneous `trim()` methods from throughout and we can get this in!

packages/fast-components-react-base/src/image/image.tsx
```diff
@@ -47,8 +47,9 @@ class Image extends Foundation<ImageHandledProps, ImageUnhandledProps, {}> {
         } else {
             className = `${className} ${get(
                 this.props,
-                "managedClasses.image__picture"
-            )}`;
+                "managedClasses.image__picture",
+                ""
+            )}`.trim();
```
janechu: As noted by @nicholasrice and @chrisdholt you can remove all `trim()` from strings going into `super.generatClassNames` because it will trim these strings for you

fmamic: I see now that we are doing trim in foundation class. Thanks!

PR #1198, opened 2018-12-10
fix: style cleanup and consolidation

nicholasrice:


packages/fast-components-class-name-contracts-base/src/radio/index.ts
```diff
@@ -2,9 +2,10 @@
  * The class name contract for the radio component
  */
 export interface RadioClassNameContract {
-    radio: string;
-    radio__disabled: string;
-    radio_input: string;
-    radio_label: string;
-    radio_stateIndicator: string;
+    radio?: string;
```
awentzel: Would it be more maintainable if properties on the interface were alphabetically ordered?

nicholasrice: I tend to order these by type, not alphabetically.

packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -53,28 +59,18 @@ const styles: ComponentStyles<RadioClassNameContract, DesignSystem> = (
...
             borderRadius: "50%",
-            boxSizing: "content-box",
             margin: "0",
             zIndex: "1",
-            background: backgroundColor,
-            boxShadow: `inset 0 0 0 1px ${radioColor}`,
+            ...outlinePattern.rest,
             "&:hover": {
-                boxShadow: `inset 0 0 0 1px ${radioHover}`,
+                ...outlinePattern.hover,
```
eljefe223: image
Little bit of strange aliasing in Dark theme with the focus outline.

packages/fast-components-react-base/src/radio/radio.tsx
```diff
@@ -35,16 +55,24 @@ class Radio extends Foundation<RadioHandledProps, RadioUnhandledProps, {}> {
         children: void 0,
     };
 
+    constructor(props: RadioProps) {
+        super(props);
+
+        this.state = {
+            checked: this.props.checked || false,
```
janechu: should this be set also in a `getDerivedStateFromProps`?

Issue #1166, opened 2018-11-27
Add support for Render Props to Tab


eljefe223:
**What is your feature request? Please describe.**
Currently Tabs relies on slots to render. Slots are prone to falling out of sync with component updates because they are loosely types. Moving to render props will make the the tab component more intuitive.

**Describe the solution you'd like**
Add render props to tabs component to make tabs API more intuitive and easier to implement.

**Additional context**
More discussion on slots vs render props can be read [here](https://github.com/Microsoft/fast-dna/issues/1135)

nicholasrice:
In the "new way" the TabItems still have slot attributes - is that intentional?

eljefe223:
@nicholasrice no! Thanks for catching that

nicholasrice:
yep - this looks good to me. we should still preserve the slot functionality if possible - if it's not we'll need to make a breaking change.

PR #1199, opened 2018-12-10
feat: add support for render props to tabs 

eljefe223:
Add support for Render Props to Tabs

Closes: #1166


packages/fast-components-react-base/src/tabs/examples.data.tsx
```diff
@@ -252,29 +276,43 @@ const examples: ComponentFactoryExample<TabsHandledProps> = {
...
             orientation: Orientation.horizontal,
-            children: exampleChildren1,
+            tabItems: exampleTabItemData1,
         },
         {
+            /**
+             * @Deprecated 3.4.0
+             */
             ...tabsManagedClasses,
-            activeId: "tab03",
```
janechu: why are we removing activeId?

packages/fast-components-react-base/src/tabs/README.md
```diff
@@ -1,5 +1,8 @@
 # Tabs
 Use *tabs* when there is a paged interface where one controlling tab element corresponds to one tab panel.
 
+## Usage
+The *tabs* component can be passed a collection of *tabitems* via the `tabitems` prop. `tabitems` expects an array of the interface `TabItems`. `Tabitems` consists of `tab`, `content` and `id`. `tab` and `content` will render a ReactNode and `id` is a string that represents its unique identifier.
```
janechu: The casing is inconsistent, when referring to the prop it should be `tabItems`, when referring to the component it should be `TabItem`.

packages/fast-components-react-base/src/tabs/examples.data.tsx
```diff
@@ -244,6 +223,51 @@ const exampleChildren4: any[] = [
...
+    id: "tab03",
+};
+
+const tabItem4: TabItems = {
+    tab: renderTab("tab four"),
+    content: renderTabContent("tab four content"),
+    id: "tab04",
+};
+
+const detailTabItemData: TabItems[] = [tabItem1, tabItem2, tabItem3];
```
janechu: prefer that the testing naming does not include "Data", this is all data.

packages/fast-components-react-base/src/tabs/tabs.props.ts
```diff
@@ -5,6 +5,22 @@ import {
...
+    className: string;
+    key: string;
+    "aria-controls": string;
+    active: boolean;
+    onClick: (e: React.MouseEvent) => void;
+    onKeyDown: (e: React.KeyboardEvent) => void;
+    tabIndex: number;
+}
+
+export interface TabItems {
```
janechu: this is the interface for a single TabItem, there should be no "s"

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -142,14 +162,59 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
...
+                            {...tabManagedClasses}
+                            slot={TabsSlot.tab}
+                            key={tabItem.id}
+                            aria-controls={tabItem.id}
+                            active={this.setActive(tabItem, index)}
+                            onClick={this.handleClick}
+                            onKeyDown={this.handleKeyDown}
+                            tabIndex={this.setActive(tabItem, index) ? 0 : -1}
+                        >
+                            {tabItem.tab("")}
```
janechu: should this not be passing a classname not empty string?

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -142,14 +162,59 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
...
+                        </Tab>
+                        <TabPanel
+                            {...tabPanelManagedClasses}
+                            slot={TabsSlot.tabPanel}
+                            key={tabItem.id}
+                            id={tabItem.id}
+                            aria-labelledby={tabItem.id}
+                            active={this.setActive(tabItem, index)}
+                        >
+                            {tabItem.content("")}
```
janechu: same issue here, shouldn't this pass a string for the class name?

packages/fast-components-react-base/src/tabs/examples.data.tsx
eljefe223: I can add the activeId back but it's not breaking when I do.

packages/fast-components-react-base/src/tabs/tabs.props.ts
eljefe223: Then we have a name conflict with TabItem which already exists. Remembering now... I think this is why I choose the name tabItemData

packages/fast-components-react-base/src/tabs/tabs.tsx
eljefe223: Yes it should!

packages/fast-components-react-base/src/tabs/tabs.tsx
eljefe223: Yup!

packages/fast-components-react-base/src/tabs/tabs.props.ts
```diff
@@ -5,6 +5,22 @@ import {
 } from "@microsoft/fast-components-class-name-contracts-base";
 import { Orientation } from "@microsoft/fast-web-utilities";
 
+export interface TabConfig {
```
nicholasrice: What is this? It doesn't seem like it is actually used.

packages/fast-components-react-base/src/tabs/tabs.props.ts
```diff
@@ -5,6 +5,22 @@ import {
...
+    className: string;
+    key: string;
+    "aria-controls": string;
+    active: boolean;
+    onClick: (e: React.MouseEvent) => void;
+    onKeyDown: (e: React.KeyboardEvent) => void;
+    tabIndex: number;
+}
+
+export interface Item {
```
nicholasrice: Can we arrive at a more illustrative name for this?

nalogiudice: +1

eljefe223: @janechu @chrisdholt and I chatted about this pretty extensively. `TabItem` is already being used. I originally had `tabItemData` but @janechu asked me to remove data. You Can see  Jane's prevuous comment.

packages/fast-components-react-base/src/tabs/tabs.props.ts
eljefe223: Not sure honestly. This branch has existed for awhile but I don't remember adding that. At any rate its been removed.

packages/fast-components-react-base/src/tabs/examples.data.tsx
```diff
@@ -189,29 +192,6 @@ function getTabItemFour(): any {
...
-                    children: "tab one - missing panel",
-                },
-            },
-        },
-    },
-    getTabItemTwo(),
-    getTabItemThree(),
-    getTabItemFour(),
-];
-
```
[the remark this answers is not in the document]
janechu: I would like to see the original examples as well as the new ones so that the new ones are represented and in the detailData and are the ones at the top of the data array, if that makes sense. It's OK there is duplication, the important thing to see in the PR for validation purposes is that all examples previously are represented, plus a new set of examples for your update.

PR #1201, opened 2018-12-11
fix: handle context menu item keydown and click events

### LINKED ISSUE #1195: Keyboard enter does not fire on click on contentMenuItem
**Describe the bug**
Keyboard enter does not fire on click on `<contentMenuItem/>`.

**To Reproduce**
1. Implement an `onClick()` on `<contentMenuItem/>`
2. Focus `<contentMenuItem/>` with keyboard
3. Hit `Enter` on keyboard

**Expected behavior**
`Enter` on keyboard should fire `onClick()`

chrisdholt:
<body>
closes #1195 
<footer>
 

### DIFF
--- a/packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
+++ b/packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
@@ -25,6 +25,12 @@ export interface ContextMenuItemHandledProps extends ContextMenuItemManagedClass
      * If the menu item is disabled
      */
     disabled?: boolean;
+
+    /**
+     * Callback for when an item is invoked
+     * Returns the prop contract for the invoked context menu item
+     */
+    onInvoke?: (props: ContextMenuItemProps) => void;
 }
 
 export type ContextMenuItemProps = ContextMenuItemHandledProps &
--- a/packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
+++ b/packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
@@ -30,7 +30,10 @@ export interface ContextMenuItemHandledProps extends ContextMenuItemManagedClass
      * Callback for when an item is invoked
      * Returns the prop contract for the invoked context menu item
      */
-    onInvoke?: (props: ContextMenuItemProps) => void;
+    onInvoke?: (
+        e: React.SyntheticEvent<HTMLDivElement>,
+        props?: ContextMenuItemProps
+    ) => void;
 }
 
 export type ContextMenuItemProps = ContextMenuItemHandledProps &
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
```diff
@@ -25,6 +25,12 @@ export interface ContextMenuItemHandledProps extends ContextMenuItemManagedClass
      * If the menu item is disabled
      */
     disabled?: boolean;
+
+    /**
+     * Callback for when an item is invoked
+     * Returns the prop contract for the invoked context menu item
+     */
+    onInvoke?: (props: ContextMenuItemProps) => void;
```
janechu: why would you want all of the props for that context menu item?

chrisdholt: Great question @janechu. Pinging @nicholasrice for awareness as he was part of this investigation. 

The idea was to keep this from being a breaking change. By passing back the full prop object, we can pass back the unhandledProps which include an id and other identifiers to know which menu item was invoked. If we pass back an "id" we need to add it to the prop contract and likely require it, which breaks the API

janechu: In that case what we have been doing is passing a React synthetic event which should also have all of this information, plus the event information.

chrisdholt: For transparency, synced offline and we'll pass back both the event object AND the props for the item

packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
```diff
@@ -25,6 +25,15 @@ export interface ContextMenuItemHandledProps extends ContextMenuItemManagedClass
      * If the menu item is disabled
      */
     disabled?: boolean;
+
+    /**
+     * Callback for when an item is invoked
+     * Returns the prop contract for the invoked context menu item
+     */
```
nicholasrice: I think you want props to not be optional, otherwise you're allowing us to internally call the function without the props argument.

packages/fast-components-react-base/src/context-menu-item/context-menu-item.props.ts
```diff
@@ -25,6 +25,15 @@ export interface ContextMenuItemHandledProps extends ContextMenuItemManagedClass
      * If the menu item is disabled
      */
     disabled?: boolean;
+
+    /**
+     * Callback for when an item is invoked
+     * Returns the prop contract for the invoked context menu item
+     */
+    onInvoke?: (
+        e: React.SyntheticEvent<HTMLDivElement>,
```
nicholasrice: Do we want to be more explicit here? MouseEvent or KeyboardEvent?

PR #1203, opened 2018-12-12
feat: add ability to remove data from optional properties

### LINKED ISSUE #560: Add to form generator a way to reset optional properties
The user should be able to reset optional properties, currently there is no way to do so. For example, a component may have an optional boolean which shows up as a checkbox, once this is pressed there is no way to get rid of that property.

If a property is not required, add it to a UI list that will reset it.

janechu:
Closes #560


packages/fast-form-generator-react/src/class-name-contracts/form-item.select.class-name-contract.ts
```diff
@@ -3,7 +3,10 @@
...
 export default interface FormItemSelectClassNameContract {
     formItemSelect: string;
-    formItemSelect_label: string;
-    formItemSelect_span: string;
-    formItemSelect_input: string;
+    formItemSelect_control: string;
+    formItemSelect_control_label: string;
+    formItemSelect_control_span: string;
+    formItemSelect_control_input: string;
+    formItemSelect_softRemove: string;
```
awentzel: What is a "softRemove"

packages/fast-form-generator-react/src/form/form-item.base.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import SoftRemove from "./soft-remove";
```
awentzel: This isn't common terminology I'm familiar with on the web?  Maybe a comment where this is first used may help explain what soft remove is?

awentzel:
It would be great to see a little more context in the description.  The title is clear.  But, more context on why this is necessary and important I feel could be useful.

packages/fast-form-generator-react/src/class-name-contracts/form-item.select.class-name-contract.ts
janechu: It will remove the data, but its a soft remove so if you click it you will add back in the data you just removed.

packages/fast-form-generator-react/src/form/form-item.base.tsx
janechu: Sounds good, added to the component description.

Issue #1130, opened 2018-11-09
Update progress to include updated style and circular loader


janechu:
Update the existing progress to the new style:
https://codepen.io/anon/pen/aQeVOm

This will also include work to add the circular version of progress to the component.

Ultimatly, there there should be four archetypes; indeterminate  horizontal, determinate horizontal, indeterminate circular, determinate circular. Each of the archetypes can be rendered in multiple sizes.

PR #1204, opened 2018-12-12
fix (progress): to include updated indeterminate animation

eljefe223:
This PR updates the style of the horizontal progress. We will add the circular style at a later time.
Partial fix for: #1130

### DIFF
--- a/packages/fast-components-class-name-contracts-msft/src/progress/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/progress/index.ts
@@ -22,15 +22,30 @@ export interface ProgressClassNameContract extends BaseProgressClassNameContract
     /**
      * The indeterminate progress indicator
      */
-    progress_indicator__indeterminate?: string;
+    progress_dot?: string;
 
     /**
      * The indeterminate progress indicator 1 modifier
      */
-    progress_indicator__indeterminate__1?: string;
+    progress_dot__1?: string;
 
     /**
      * The indeterminate progress indicator 2 modifier
      */
-    progress_indicator__indeterminate__2?: string;
+    progress_dot__2?: string;
+
+    /**
+     * @Deprecated 3.4.0
+     */
+    progress_dot__3?: string;
+
+    /**
+     * @Deprecated 3.4.0
+     */
+    progress_dot__4?: string;
+
+    /**
+     * @Deprecated 3.4.0
+     */
+    progress_dot__5?: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/progress/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/progress/index.ts
@@ -22,30 +22,15 @@ export interface ProgressClassNameContract extends BaseProgressClassNameContract
     /**
      * The indeterminate progress indicator
      */
-    progress_dot?: string;
+    progress_indicator__indeterminate?: string;
 
     /**
      * The indeterminate progress indicator 1 modifier
      */
-    progress_dot__1?: string;
+    progress_indicator__indeterminate__1?: string;
 
     /**
      * The indeterminate progress indicator 2 modifier
      */
-    progress_dot__2?: string;
-
-    /**
-     * The indeterminate progress indicator 3 modifier
-     */
-    progress_dot__3?: string;
-
-    /**
-     * The indeterminate progress indicator 4 modifier
-     */
-    progress_dot__4?: string;
-
-    /**
-     * The indeterminate progress indicator 5 modifier
-     */
-    progress_dot__5?: string;
+    progress_indicator__indeterminate__2?: string;
 }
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-msft/src/progress/README.md
```diff
@@ -5,4 +5,7 @@ The *progress* component is an [div](https://developer.mozilla.org/en-US/docs/We
 The *progress* component should be used for an element that displays the progress status for a task that takes a long time or consists of several steps. The *progress* component indicates that the user's request has been received and the application is making progress toward completing the requested action. If the `value` of the *progress* component can't be determined then the indeterminate visuals will be shown until the `value` is determined.
 
 ## Style guidance
-The determinate visuals should indicate to the user how much is remaining of a given task. The indeterminate should be a repeating animation to indicate to the user that their request is being processed and the application is responsive.
\ No newline at end of file
+The determinate visuals should indicate to the user how much is remaining of a given task. The indeterminate should be a repeating animation to indicate to the user that their request is being processed and the application is responsive.
```
nalogiudice: Maybe "visual display" instead of "visuals" in the first sentence would be clearer?
And if you change that, please add "visual display" to "The indeterminate should be a repeating...", like "The indeterminate visual display should be a repeating...".

packages/fast-components-react-msft/src/progress/README.md
```diff
@@ -5,4 +5,7 @@ The *progress* component is an [div](https://developer.mozilla.org/en-US/docs/We
...
 
 ## Style guidance
-The determinate visuals should indicate to the user how much is remaining of a given task. The indeterminate should be a repeating animation to indicate to the user that their request is being processed and the application is responsive.
\ No newline at end of file
+The determinate visuals should indicate to the user how much is remaining of a given task. The indeterminate should be a repeating animation to indicate to the user that their request is being processed and the application is responsive.
+
+## Browser requirements
+Known issue: The indeterminate animation relies on percentage-based values. In Edge there is known issue where the percentage values do not get recalculated when the container is resized. If properly rendering in Edge is a top priority, we recommend hardcoding the width of the progress bar to prevent glitchy animations.
```
nalogiudice: I would rewrite this to be:
"In Edge, there is known issue where percentage-based values (as found in the indeterminate animation) do not get recalculated when the container is resized. As a workaround for Edge, we recommend hardcoding the width of the progress bar to prevent animation issues."

packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -29,83 +29,85 @@ const styles: ComponentStyles<ProgressClassNameContract, DesignSystem> = (
...
         progress_indicator: {
             position: "relative",
             display: "flex",
             alignItems: "center",
             width: "100%",
+            overflow: "hidden",
+            borderRadius: "100px",
+            height: toPx(designSystem.designUnit),
+            background: backgroundColor,
+            maskImage: "-webkit-radial-gradient(white, black)",
```
nalogiudice: Interesting - did you verify in all browsers?

eljefe223: The issue is that if a parent element has a `border-radius` property set and its `overflow` set to `hidden` Safari does not properly mask the child elements with the set corner radius... you just end up with a square clip. This is a fix for Safari only (Edge and Chrome seem to clip properly) and I verified it fixed the issue in a codepen sample on @janechu machine. If you don't mind, would be good for you to smoke this review and verify it fixes the issue in Safari (Since I'm not on a Mac)

packages/fast-components-class-name-contracts-msft/src/progress/index.ts
```diff
@@ -22,30 +22,15 @@ export interface ProgressClassNameContract extends BaseProgressClassNameContract
     /**
      * The indeterminate progress indicator
      */
-    progress_dot?: string;
+    progress_indicator__indeterminate?: string;
 
     /**
      * The indeterminate progress indicator 1 modifier
```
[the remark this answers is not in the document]
eljefe223: Thats not a problem. Its just a terrible name for these new elements. Also only 2 now should instead of 5... I assume we still want to keep all the progress_dot__X even though they no longer have elements?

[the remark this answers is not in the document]
nicholasrice: You should keep them and mark them deprecated

packages/fast-components-styles-msft/src/progress/index.ts
nalogiudice: Tested on Safari - seems the same as Chrome.

packages/fast-components-class-name-contracts-msft/src/progress/index.ts
```diff
@@ -35,17 +35,17 @@ export interface ProgressClassNameContract extends BaseProgressClassNameContract
     progress_dot__2?: string;
 
     /**
-     * The indeterminate progress indicator 3 modifier
+     * @Deprecated 3.4.0
```
janechu: Can you create a task for changing the names of `progress__dot`, `progress__dot__1`, `progress__dot__2` to be more semantically correct and link the task? As you mentioned in your other comment, the names don't make sense anymore so it would be fine to track that as an issue.

eljefe223: Good Idea.

New Issue here:
https://github.com/Microsoft/fast-dna/issues/1215

PR #1207, opened 2018-12-13
fix: adds better high-contrast support

nicholasrice:


packages/fast-components-styles-msft/src/progress/index.ts
```diff
@@ -58,6 +65,11 @@ const styles: ComponentStyles<ProgressClassNameContract, DesignSystem> = (
                 width: toPx(designSystem.designUnit),
                 height: toPx(designSystem.designUnit),
                 borderRadius: "50%",
+
+                "@media (-ms-high-contrast:active)": {
+                    background: "ButtonShadow",
+                    opacity: "1 !important",
```
chrisdholt: `!important` - Gross ;). Is this due to a bug or are we just ensuring that a dev cannot change or address this with their own style changes?

nicholasrice: Its mostly just to avoid creating a new animation for that doesn't change opacity. Because this is only for high-contrast I think the danger is pretty low here.

Issue #1135, opened 2018-11-13
Create `text action` as a new MSFT component


chrisdholt:
Text action allows the user to apply an action to a specific text entry. For instance, typing in an email address and then adding that to a list of other addresses. The glyph informs the user what the expected action will be.

chrisdholt:
@nalogiudice instead of "slots", let's look at using render props for these: https://reactjs.org/docs/render-props.html

**Example:**
```
button?: (className?: string) => React.ReactNode;
```

This allows us to tell a user:
1.  That we expect a function that returns a react node
2. That we will be passing data to them

**Implementation:**
```
this.props.button(this.props.managedClasses.textAction_button)
```

nalogiudice:
The slot implementation proposal was for the `glyph` usage - like we currently do in MSFT [button](https://github.com/Microsoft/fast-dna/blob/master/packages/fast-components-react-msft/src/button/button.tsx). Are you suggesting to _not_ do this?

chrisdholt:
> The slot implementation proposal was for the `glyph` usage - like we currently do in MSFT [button](https://github.com/Microsoft/fast-dna/blob/master/packages/fast-components-react-msft/src/button/button.tsx). Are you suggesting to _not_ do this?

Yes; suggest we move to render props. It's a more transparent and "react" way of implementing things like this. These will show up in `action trigger` and `breadcrumb` which are currently in process.

nalogiudice:
Got it - I'll reference those components for usage.

nalogiudice:
@chrisdholt it doesn't look like they are using this yet, and reading the documentation, it seems like there are some caveats with this approach and pure components (which I believe we will be moving to eventually). 

Where is this direction coming from? Do @nicholasrice or/and @janechu also agree with this new approach?

chrisdholt:
@nalogiudice these are a common pattern and @nicholasrice and @janechu are on board. Where are you seeing that "they are not using them"? This has been a pattern in the community for awhile, and now has first class documentation.

nalogiudice:
Not talking about the react community, talking about our current PRs (breadcrumb and action trigger). Action trigger uses slots and it looks like breadcrumb has not been updated yet.

The caveat still stands (see documentation). I just don't want to start using a pattern we will have to rip out anyway - just making sure we want to go this route.

Currently, it looks like the only place with a similar pattern is within `breakpoint tracker`

chrisdholt:
[...]

PR #1208, opened 2018-12-14
feat: add text action as a component

nalogiudice:
Closes [#1135](https://github.com/Microsoft/fast-dna/issues/1135)
<body>

<footer>
 


packages/fast-components-react-msft/src/text-action/text-action.spec.tsx
```diff
@@ -0,0 +1,323 @@
...
+        ).toBe(false);
+        expect(
+            rendered
+                .find(`.${managedClasses.textAction}`)
+                .find("button")
+                .exists()
+        ).toBe(true);
+    });
+
+    test("should render a button only if a glyph and button are passed in the same (before) position", () => {
```
connorjsmith: nit: I read this as a button should be rendered only in the case that both a glyph and button are passed. Consider `should render only a button if...`?

packages/fast-components-react-msft/src/text-action/text-action.spec.tsx
```diff
@@ -0,0 +1,323 @@
...
+        expect(
+            rendered
+                .find(`.${managedClasses.textAction}`)
+                .first()
+                .childAt(1)
+                .type()
+        ).toEqual("svg");
+    });
+
+    test("should render a button only if a glyph and button are passed in the same (after) position", () => {
```
connorjsmith: same wording comment here

packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -0,0 +1,121 @@
...
+        return (
+            <div className={this.generateClassNames()}>
+                {this.getButton() && this.props.buttonPosition === ButtonPosition.before
+                    ? this.props.button(
+                          get(this.props, "managedClasses.textAction_button"),
+                          this.props.disabled
+                      )
+                    : null}
+                {this.generateBeforeGlyph()}
+                <TextField
```
connorjsmith: If we type into a TextAction with a button and hit `enter`, does it use the button action or the TextField's?

nalogiudice: As it is, it uses TextField's

connorjsmith: Not sure if this is what I'd expect as a developer, but I guess its easy enough to code in with `preventDefault`

nalogiudice: Typically this would be wrapped in a `form` that would handle the action - or other props could be passed to the button

nalogiudice: Open to a discussion on best implementation. I know we are trying to get away from prescribing to heavily in situations like this and allowing the developers to wire things up how they need in their instance.

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    };
+
+    return {
+        textAction: {
+            position: "relative",
+            height: density(defaultHeight)(designSystem),
+            minHeight: toPx(minHeight),
+            maxHeight: toPx(maxHeight),
+            margin: `${density(designSystem.designUnit * 3)(designSystem)} 0`,
+            minWidth: "92px",
```
nicholasrice: Can you adjust the use the outline pattern instead? You can use text-field's stylesheet for reference.

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,147 @@
...
+            border: `1px solid ${foregroundNormal(designSystem)}`,
+            background: ensuresBackgroundNormal,
+            borderRadius: "2px",
+            display: "flex",
+            flexDirection: "row",
+            transition: "all 0.2s ease-in-out",
+            "&:hover": {
+                borderColor: hoverColor(designSystem),
+            },
+            "&:focus-within": {
```
nicholasrice: The focus states of this component feel wrong because there are two tab-stops with the same focus treatment, one for the button and one for the input. I think we need to address focus differently so that I can clearly identify which element is focused

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,147 @@
...
+        textAction: {
+            position: "relative",
+            height: density(defaultHeight)(designSystem),
+            minHeight: toPx(minHeight),
+            maxHeight: toPx(maxHeight),
+            margin: `${density(designSystem.designUnit * 3)(designSystem)} 0`,
+            minWidth: "92px",
+            border: `1px solid ${foregroundNormal(designSystem)}`,
+            background: ensuresBackgroundNormal,
+            borderRadius: "2px",
```
nicholasrice: I believe this radius was added to the desginSystem - you should reference it instead.

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,147 @@
...
+            maxHeight: toPx(maxHeight),
+            minWidth: "fit-content",
+            margin: "2px",
+            padding: "0 16px",
+            [applyLocalizedProperty("right", "left", direction)]: "0",
+            top: "0",
+            transition: "color .1s, background-color .1s, border-color 0.2s ease-in-out",
+            flex: "0 0 auto",
+            cursor: "pointer",
+            "&:focus-visible": {
```
nicholasrice: You'll want to use the `focusVisible` helper when using focus-visible

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/* tslint:disable:max-line-length */
+const styles: ComponentStyles<TextActionClassNameContract, DesignSystem> = (
+    config: DesignSystem
+): ComponentStyleSheet<TextActionClassNameContract, DesignSystem> => {
+    const designSystem: DesignSystem = withDesignSystemDefaults(config);
+    const backgroundColor: string = designSystem.backgroundColor;
+    const direction: Direction = designSystem.direction;
+    const foregroundColor: string = ensureNormalContrast(
```
nicholasrice: there already exists a `ensureForegroundNormal` helper you can import instead of defining one here.

packages/fast-components-styles-msft/src/text-action/index.ts
nalogiudice: Agreed - there is no design for this, so I am just going by the seat of my pants as it were. Any suggestions?

packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -0,0 +1,125 @@
...
+            )}`;
+        }
+
+        return super.generateClassNames(classNames);
+    }
+
+    /**
+     * Returns truthy if button exist
+     */
+    private getButton(): boolean {
```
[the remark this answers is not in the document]
nalogiudice: Sure

packages/fast-components-react-msft/src/text-action/text-action.props.ts
```diff
@@ -0,0 +1,50 @@
...
+import { ManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+import { TextActionClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import {
+    TextFieldHandledProps,
+    TextFieldManagedClasses,
+} from "@microsoft/fast-components-react-base";
+import { Omit, Subtract } from "utility-types";
+import { ButtonAppearance } from "../button/button.props";
+
+export enum ButtonPosition {
```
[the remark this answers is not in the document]
nalogiudice: Well, it's the optional button that can be added to the text-action. `TextActionButtonPosition` seems unnecessarily verbose, but that's the only other name that comes to mind...

packages/fast-components-styles-msft/src/text-action/index.ts
nalogiudice: Yeah, I don't think this pattern existed when I first started this work - all adjusted.

packages/fast-components-react-msft/src/text-action/text-action.props.ts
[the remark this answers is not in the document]
nicholasrice: It might seem unnecessarily verbose but remember these are global names. What is to differentiate this enum from an enum exported to be used directly by the button component?

packages/fast-components-styles-msft/src/text-action/index.ts
nicholasrice: I would start with not removing the focus state on the button itself.

packages/fast-components-react-msft/src/text-action/text-action.props.ts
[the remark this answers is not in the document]
nalogiudice: Totally - good point. Will change.

packages/fast-components-styles-msft/src/text-action/index.ts
nalogiudice: So after debating on the best way to solve this, I decided that instead of using `focus-within`, I would just use a focus class to control the focus UI. So when the inner `input` gets focus, an outer focus class is added to the wrapper (the `<TextAction />`. On blur, the focus class is removed. Therefore, we are not mixing focus states.

nalogiudice:
@Falkicon 

1. We need to file a task to target the `::placeholder` pseudo-element in text-field and that would fix here as well. Without a design, I was just aligning it with text-field for consistency. 

2.-4. Right now there is no way to edit **rendered** props in the editor. Because all glyphs and buttons are _optional_ and passed via rendered props, you cannot (at this time) access them or change them. This is why you cannot change the button text. I assume this is being tracked, but @janechu or @nicholasrice may have more information on this.

5. This is tricky... these glyphs are passed as react nodes via rendered props. Do we want to enforce colors on these? I thought about this a bit I decided against it because these glyphs could, in fact, be any color. For the objective of our dev site, we could enforce glyph color via JSS. I am just not sure if that is "correct".  I'll push an update with the fixed, but I would like others to chime in (@nicholasrice )

6. I used the `12px` margin inherited from `text-field`as a basis. I can remove, but we would need to align text-field for consistency.

7. This was due to min-width on the text-field - I have fixed.

8. I'm not quite sure I understand the theme issues. Is there a design I can compare to, as I was basing this off of text-field.

9. When you say pattern do you mean design or example? There are examples of every configuration under the examples tab. I wanted to allow the most flexibility with this component but also explain the mechanics/decision making behind the glyphs and buttons. So I was trying to be clear that the component allows a before and after glyph, and a before and after button, but they cannot inhabit the same area (i.e. a before glyph and button cannot coexist), and that buttons take precedence over glyphs (in the same location). The original task thread had discussions with the team about these decisions.

10. Will remove.

packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Generates before glyph based on props
+     */
+    private generateBeforeGlyph(): React.ReactNode {
+        if (typeof this.props.beforeGlyph === "function") {
+            if (
+                (this.buttonExists() &&
+                    this.props.buttonPosition !== TextActionButtonPosition.before) ||
+                !this.buttonExists()
+            ) {
```
chrisdholt: I think we can lose some of the verbosity here if we check if there is no button first, right? If that check is true, we can perform the next check under that assumption.

```
if (!this.buttonExists() || this.props.buttonPosition !== TextActionButtonPosition.before) {}
```

packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Generates after glyph based on props
+     */
+    private generateAfterGlyph(): React.ReactNode {
+        if (typeof this.props.afterGlyph === "function") {
+            if (
+                (this.buttonExists() &&
+                    this.props.buttonPosition !== TextActionButtonPosition.after) ||
+                !this.buttonExists()
+            ) {
```
chrisdholt: I think we can lose some of the verbosity here if we check if there is no button first, right? If that check is true, we can perform the next check under that assumption.

```
if (!this.buttonExists() || this.props.buttonPosition !== TextActionButtonPosition.before) {}
```

packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -0,0 +1,167 @@
...
+     */
+    private generateBeforeGlyph(): React.ReactNode {
+        if (typeof this.props.beforeGlyph === "function") {
+            if (
+                (this.buttonExists() &&
+                    this.props.buttonPosition !== TextActionButtonPosition.before) ||
+                !this.buttonExists()
+            ) {
+                return this.props.beforeGlyph(
+                    get(this.props, "managedClasses.textAction_beforeGlyph")
```
chrisdholt: Can you include a default value of an empty string?

```
get(this.props, "managedClasses.textAction_beforeGlyph", "")
```

packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -0,0 +1,167 @@
...
+     */
+    private generateAfterGlyph(): React.ReactNode {
+        if (typeof this.props.afterGlyph === "function") {
+            if (
+                (this.buttonExists() &&
+                    this.props.buttonPosition !== TextActionButtonPosition.after) ||
+                !this.buttonExists()
+            ) {
+                return this.props.afterGlyph(
+                    get(this.props, "managedClasses.textAction_afterGlyph")
```
chrisdholt: Can you include a default value of an empty string?

```
get(this.props, "managedClasses.textAction_afterGlyph", "")
```

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,146 @@
...
+import outlinePattern from "../patterns/outline";
+
+// Since MSFT button is already styled, we need to override in this way to alter button classes
+export const textFieldOverrides: ComponentStyles<
+    Partial<TextFieldClassNameContract>,
+    DesignSystem
+> = {
+    textField: {
+        height: "100%",
+        marginTop: "0",
```
chrisdholt: margin: "0"

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,146 @@
...
+        fill: designSystem.foregroundColor,
+    };
+
+    return {
+        textAction: {
+            position: "relative",
+            height: density(defaultHeight)(designSystem),
+            minHeight: toPx(minHeight),
+            maxHeight: toPx(maxHeight),
+            margin: `${density(designSystem.designUnit * 3)(designSystem)} 0`,
```
chrisdholt: `margin: "0"`

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,146 @@
...
+            margin: "2px",
+            padding: "0 16px",
+            [applyLocalizedProperty("right", "left", direction)]: "0",
+            top: "0",
+            transition: "color .1s, background-color .1s, border-color 0.2s ease-in-out",
+            flex: "0 0 auto",
+            cursor: "pointer",
+            // Because button is passed as a rendered prop, there is no way to access internal
+            // selectors and pass style via jssStyleSheet prop
+            "& svg": {
```
chrisdholt: I don't think we want to target all `svg` elements here

packages/fast-components-styles-msft/src/text-action/index.ts
nalogiudice: This can be removed all-together once the other PR is in

packages/fast-components-styles-msft/src/text-action/index.ts
nalogiudice: Yeah, this will also be removed per my comment in the meeting. Will update once the other PR is in.

packages/fast-components-react-msft/src/text-action/text-action.tsx
nalogiudice: Good call.

packages/fast-components-react-msft/package.json
```diff
@@ -88,6 +88,7 @@
     "babel-jest": "^23.6.0",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "exenv-es6": "^1.0.0",
```
nicholasrice: I don't see this used elsewhere - is this necessary?

nalogiudice: Not any more - at one point I had to check the DOM. Will remove.

packages/fast-components-react-msft/src/text-action/text-action.tsx
nalogiudice: Resolving for now since there has been no follow-up.

packages/fast-components-react-msft/src/text-action/text-action.tsx
chrisdholt: @nalogiudice FYI - I think my mistake was that the {} in my example would contain an empty return statement. You'll likely want to triple equals check that position IS before (and after in that instance)

packages/fast-components-class-name-contracts-msft/src/text-action/index.ts
```diff
@@ -0,0 +1,31 @@
...
+     * The root of the text action component
+     */
+    textAction: string;
+
+    /**
+     * The disabled state
+     */
+    textAction__disabled: string;
+
+    textAction__focus: string;
```
chrisdholt: nit, missing typedoc

packages/fast-components-react-msft/src/text-action/text-action.tsx
chrisdholt: and this may need to be an && if you are executing within the block.

otherwise I might just suggest an empty return.

```
if (typeof this.props.beforeGlyph === "function") {
             if (
                 !this.buttonExists() ||
                 this.props.buttonPosition !== TextActionButtonPosition.before
             ) {
                 return;
             }
```

chrisdholt: or - just swap to check button position is not "after"

chrisdholt: disregard; I'm a dummy.

nalogiudice: Resolved offline - this is correct.

packages/fast-components-class-name-contracts-msft/src/text-action/index.ts
nalogiudice: Fixed

PR #1217, opened 2018-12-19
feat: update the form objects to use the same eye for soft removing them

### LINKED ISSUE #1216: Match the method for resetting optional properties
The form generator currently uses toggles to add/remove optional objects, it should use the eye glyph as used by other form items instead.

janechu:
Closes #1216


packages/fast-form-generator-react/src/form/form-item.section-link.tsx
```diff
@@ -0,0 +1,59 @@
...
+     * The update section callback
+     */
+    onUpdateSection: (schemaLocation: string, dataLocation: string) => void;
+}
+
+/**
+ * Schema form component definition
+ * @extends React.Component
+ */
+/* tslint:disable-next-line */
```
nicholasrice: what is this disabling?

packages/fast-form-generator-react/src/form/form-section.tsx
```diff
@@ -168,11 +169,42 @@ class FormSection extends React.Component<
...
 
+    /**
+     * Generates a section link for properties
+     * that are objects
+     */
+    private generateFormItemSectionLink(
+        location: string,
+        property: any,
+        formItemProps: FormItemCommon
+    ): any {
```
chrisdholt: @janechu is this actually *any*? Looks like you're returning JSX?

janechu: good catch

PR #1225, opened 2018-12-23
Update issue templates

Falkicon:
I have simplified the templates and made them more consistent. Also deleted a placeholder template we were not using.


Falkicon:
@awentzel, confirmed the new templates are working.

PR #1232, opened 2018-12-24
fix: updated component readme.md files to be more consistent 

Falkicon:
Aligned heading sizes, element, role and attribute links, and some minor content edits. Overall, I did not focus on the actual content; I will do that in another pass.


packages/fast-components-react-msft/src/text-field/README.md
```diff
@@ -1,8 +1,11 @@
...
+
+It is recommended to have hint text inside or a *label* above a *text field* to indicate what kind of text input the user should type. When reasonable, automatically set focus to the primary *text field* or the first *text field* in a *form*. Limit the length of allowable input text when possible. Since *text field* input may vary from one use case to another, the length can be limited to a maximum or minimum number of characters. A *text field* with the [`readonly`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#readonly) attribute is focusable and should have [`aria-label="Read only"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) attribute.
```
chrisdholt: Just an observation - looks like we recommend hit text or a label, but I think a label is required for accessibility. I don’t think placeholder text is considered a replacement to a label. Should we move this and address in an accessibility section for consistency?

Issue #1239, opened 2018-12-27
Margin and placeholder style issues


nalogiudice:
Components should **not** include their own margin. The text-field and divider top and bottom margins should be removed.

The placeholder text in the text-field disabled state is the wrong color - it should match the disabled state text color. Update.

PR #1240, opened 2018-12-27
fix: remove margins from components and fix text-field placeholder color in disabled state

nalogiudice:
Closes [#1239](https://github.com/Microsoft/fast-dna/issues/1239)

## Description
Removes margin from components and addresses placeholder text color for the disabled state in text-field.

## Motivation & context
Noticed margin and placeholder issue when creating `text action` component.

### DIFF
--- a/packages/fast-components-styles-msft/src/divider/index.ts
+++ b/packages/fast-components-styles-msft/src/divider/index.ts
@@ -18,7 +18,7 @@ const styles: ComponentStyles<DividerClassNameContract, DesignSystem> = (
         divider: {
             boxSizing: "content-box",
             height: "0",
-            margin: `0`,
+            margin: "0",
             border: "none",
             borderTop: `1px solid ${normalContrast(
                 designSystem.contrast,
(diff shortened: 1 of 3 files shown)


chrisdholt:
@nalogiudice can you update your PR title?

packages/fast-components-styles-msft/src/divider/index.ts
```diff
@@ -18,7 +18,7 @@ const styles: ComponentStyles<DividerClassNameContract, DesignSystem> = (
         divider: {
             boxSizing: "content-box",
             height: "0",
-            margin: `${density(designSystem.designUnit * 3)(designSystem)} 0`,
+            margin: `0`,
```
chrisdholt: nit: I know this works, but we don't actually need back-ticks here and would prefer to use quotes in these cases instead.

packages/fast-components-styles-msft/src/text-field/index.ts
```diff
@@ -34,7 +34,7 @@ const styles: ComponentStyles<TextFieldClassNameContract, DesignSystem> = (
             boxSizing: "border-box",
             borderRadius: toPx(designSystem.cornerRadius),
             padding: "10px",
-            margin: `${density(designSystem.designUnit * 3)(designSystem)} 0`,
+            margin: `0`,
```
chrisdholt: nit: I know this works, but we don't actually need back-ticks here and would prefer to use quotes in these cases instead.

PR #1250, opened 2018-12-29
feat: simplify array UI by removing the dropdown menu

janechu:
# Description

This change updates the handling of the array data type in the `fast-form-generator-react` package to use a simplified method of controlling it and removes the drop down menu.

This change will go into a feature branch which also addresses a UI change for the handling of the React children data type.

## Motivation & context

This simplifies the UI for adding and removing array items.

## Issue type checklist

[ ] This change causes current functionality to break.

## Process & policy checklist

### DIFF
--- a/packages/fast-form-generator-react/src/form/form-item.array.style.ts
+++ b/packages/fast-form-generator-react/src/form/form-item.array.style.ts
@@ -1,4 +1,4 @@
-import { ellipsis, toPx } from "@microsoft/fast-jss-utilities";
+import { ellipsis, focusVisible, toPx } from "@microsoft/fast-jss-utilities";
 import {
     applyCleanListStyle,
     applyControl,
@@ -39,7 +39,7 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
         height: "16px",
         zIndex: "1",
         borderRadius: "2px",
-        "&:focus": {
+        [`&${focusVisible()}`]: {
             ...insetStrongBoxShadow(colors.pink),
             outline: "none",
         },
@@ -63,7 +63,6 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
     },
     formItemArray_existingItemList: {
         ...applyCleanListStyle(),
-        "& li": {},
     },
     formItemArray_existingItemListItem: {
         position: "relative",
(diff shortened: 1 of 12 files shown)


packages/fast-form-generator-react/src/form/form-item.array.style.ts
```diff
@@ -20,59 +18,81 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
...
+            content: "''",
+            opacity: ".6",
+            pointerEvents: "none",
+            top: toPx(11),
+            width: toPx(16),
+            height: toPx(16),
+            background: lines,
+            left: "0",
+        },
+        "& a": {
```
chrisdholt: @janechu do we not have a class for this? Would be *ideal* if we didn't have to target elements wholesale.

packages/fast-form-generator-react/src/form/form-item.children.style.ts
```diff
@@ -46,9 +45,36 @@ const styles: ComponentStyles<FormItemChildrenClassNameContract, {}> = {
...
+            pointerEvents: "none",
+            top: "calc(50% - 8px)",
+            width: toPx(16),
+            height: toPx(16),
+            background: lines,
+            left: "0",
+        },
+        "& a": {
+            width: "calc(100% - 36px)",
+            "& span, i": {
```
chrisdholt: @janechu `li`? I don't think that we are using `i` (nor should we as that's a deprecated element).

janechu: Deprecated, where are you reading that? There is of course `<i>` and `<em>` which is similar but the text used here is not for emphasis, it's italicized to set it apart from the other text in the `<li>`.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i

packages/fast-form-generator-react/src/form/form-item.array.style.ts
janechu: We are very precisely targeting the `<a>` inside of the `<li>` which has the class `formItemArray_existingItemListItem`, there should only be a single `<a>` inside.

I can add a class and use that instead, however I would argue the reasoning is that it is more maintainable, I don't think it's a good argument that the targeting is "wholesale" in this case.

chrisdholt: Cool - primary consideration was the reasoning. Thx!

packages/fast-form-generator-react/src/form/form-item.children.style.ts
chrisdholt: I wasn’t seeing the ‘i’ which is why I was curious. If we are considering this as a semantic representation for “italicized” text, the spec has changed: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-i-element

I’d suggest if the existing spec doesn’t align and there is no semantic meaning - just presentational italics - we should just style with a span.

janechu: For context, the `<span>` is used as the name of the child component, the `<i>` is used if there is content for it, they should not be treated the same or in the "same voice" therefore the use of `<i>` to differentiate.

An example would be a component named "Button" with the inner text being "Hello world". You would then read "Button hello world" in the same voice if both used the `<span>` element.

chrisdholt: Fantastic, thanks for the context!

PR #1257, opened 2019-01-02
feat: add action toggle component 

### LINKED ISSUE #1131: Create "action toggle" as a new composed MSFT component

scomea:
# Description
Add an action toggle styled component.  
(This pr went stale over the holidays and this is the revived version - https://github.com/Microsoft/fast-dna/pull/1162)

## Motivation & context
Closes #1131 - Create "action toggle" as a new composed MSFT component.

## Issue type checklist

## Process & policy checklist
- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://microsoft.github.io/fast-dna/docs/en/contributing/standards) for this project


packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * The class name contract for the action toggle component
+ */
+export interface ActionToggleClassNameContract {
+    /**
+     * The root of the action toggle component
+     */
+    actionToggle: string;
+    /**
```
nalogiudice: nit: can you normalize the line breaks between comments and declarations?

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,13 @@
+# Action Toggle
+*Action toggle* should be used as control on or around an interactive region of the page that affects the region or items within the region. It usually represents a switch between one of two states.
+
+## Usage
+Use *action toggle* for specific control around a functional state. Labeling can be done using a glyph and text, but could also contain just a glyph or just text. Make sure the text and glyph clearly represent the expected function. When only a glyph is used to label the button, include a tooltip in a span element following the button element.
```
nalogiudice: "When only a glyph is used to label the button, include a tooltip in a span element following the button element."
I'm not sure we should include this right now since tooltip doesn't exist.

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+The "selectedGlyph" and "unselectedGlyph" props take a function that returns a ReactNode to render the glyph (ie. a [render prop](https://reactjs.org/docs/render-props.html)).  Note that the function should accept a string parameter that is applied as a classname to the top level element of the resulting node.  This classname  will be populated by the *action toggle* component to use an internally generated class ("actionToggle_selectedGlyph" & "actionToggle_unselectedGlyph") to apply to the glyphs for styling.
```
nalogiudice: class name should be two words

nalogiudice: I'm also not sure we should be including this level of detail about something that is basically baked-in.

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+The "selectedGlyph" and "unselectedGlyph" props take a function that returns a ReactNode to render the glyph (ie. a [render prop](https://reactjs.org/docs/render-props.html)).  Note that the function should accept a string parameter that is applied as a classname to the top level element of the resulting node.  This classname  will be populated by the *action toggle* component to use an internally generated class ("actionToggle_selectedGlyph" & "actionToggle_unselectedGlyph") to apply to the glyphs for styling. 
+
+## Style guidance
+Multiple related *action toggle* controls should be grouped together visually to reinforce their associated functionality. Place the toggle where it is not hidden or obscured in the clutter of the page.
+
+## Accessibility
+The *action toggle* utilizes the text found in the "selectedARIALabel" and "unselectedARIALabel" properties to populate an aria-label with the toggled state. This is important when the *action toggle* is configured as glyph-only. Otherwise, a non-sighted user would not have context to the toggle change. If a more verbose description is needed, tooltip is always an option to provide additional context.
```
nalogiudice: Same thing about tooltip here too

packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * The action toggle unselected content
+     */
+    unselectedContent?: React.ReactNode;
+
+    /**
+     * The action toggle unselected ARIA text
+     */
+    unselectedLabel: string;
```
nalogiudice: In the README, you call this `selectedARIALabel` but here it is `unselectedLabel`. These should be consistent

packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts
```diff
@@ -0,0 +1,72 @@
...
+    primary = ButtonAppearance.primary,
+}
+
+export interface ActionToggleManagedClasses
+    extends ManagedClasses<ActionToggleClassNameContract> {}
+export interface ActionToggleHandledProps
+    extends Omit<ButtonHandledProps, "appearance" | keyof ButtonManagedClasses>,
+        ActionToggleManagedClasses {
+    /**
+     * The action toggle selected glyph render prop
```
nalogiudice: Since these are actually more about a UI state, I would rewrite the comment to something more like:
"The action toggle selected state glyph"
Same for the unselectedGlyph

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
```diff
@@ -0,0 +1,35 @@
...
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "selected": {
+            "title": "Selected",
+            "type": "boolean"
+        },
+        "selectedContent": {
+            "title": "Selected text",
+            "type": "object"
```
nalogiudice: What does this look like in the editor?

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
```diff
@@ -0,0 +1,35 @@
...
+        },
+        "unselectedContent": {
+            "title": "Unselected text",
+            "type": "object"
+        },
+        "selectedLabel": {
+            "title": "Selected label",
```
nalogiudice: If these are aria, the title should say "Selected HTML aria-label attribute". If these are just the display string, then this title works.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+/*
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("action toggle", (): void => {
+    const managedClasses: ActionToggleClassNameContract = {
+        actionToggle: "action-toggle",
+        actionToggle_selectedGlyph: "selectedGlyph",
```
nalogiudice: nit: `action-toggle-selected-glyph` would align with our standards better

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+    test("should not throw if managedClasses are not provided", () => {
+        expect(() => {
+            shallow(
+                <MSFTActionToggle selectedLabel="selected" unselectedLabel="unselected" />
+            );
+        }).not.toThrow();
+    });
+
+    test("should create unselectedglyph element passed in", () => {
```
nalogiudice: typo I think "...element _if_ passed in" 
This occurrence is elsewhere in the file.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+        const props: ActionToggleHandledProps = {
+            selectedLabel: "selected",
+            unselectedLabel: "unselected",
+            disabled: true,
+        };
+
+        const rendered: any = mount(<ActionToggle {...props} />);
+
+        expect(rendered.find("button").prop("className")).toContain(
+            "actionToggle__disabled"
```
nalogiudice: I would prefer to interpolate this class name in case it changes i.e. `${managedClasses.actionToggle__disabled}`

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+        };
+
+        const rendered: any = mount(
+            <ActionToggle className={"custom-class-name"} {...props} />
+        );
+
+        expect(rendered.find("button").prop("className")).toContain("custom-class-name");
+    });
+
+    test("should apply the custom selected ARIA label", () => {
```
nalogiudice: This test description is wrong. And there is some copy/paste errors in the render.

packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * The class name contract for the action toggle component
+ */
+export interface ActionToggleClassNameContract {
+    /**
+     * The root of the action toggle component
+     */
+    actionToggle: string;
```
janechu: I believe this should be optional, all class names should be optional.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+import Foundation, { HandledProps } from "@microsoft/fast-components-foundation-react";
+import { Button, ButtonAppearance } from "../button";
+import {
+    ActionToggleAppearance,
+    ActionToggleHandledProps,
+    ActionToggleProps,
+    ActionToggleUnhandledProps,
+} from "./action-toggle.props";
+import { actionToggleButtonOverrides } from "@microsoft/fast-components-styles-msft";
+import { isNullOrWhiteSpace } from "@microsoft/fast-web-utilities";
```
janechu: This is not used, please remove.

packages/fast-components-react-msft/package.json
```diff
@@ -80,6 +80,7 @@
     "@microsoft/fast-jss-manager": "^3.0.7",
     "@microsoft/fast-sketch-library": "^3.0.3",
     "@microsoft/fast-tslint-rules": "^3.0.1",
+    "@microsoft/fast-web-utilities": "^3.0.3",
```
janechu: Please remove, see my comment on the file, the import is not actually used

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+            "actionToggle__hasGlyphAndContent"
+        );
+
+        rendered.find("button").simulate("click");
+        expect(rendered.find("button").prop("className")).not.toContain(
+            "actionToggle__hasGlyphAndContent"
+        );
+    });
+
+    test("should not apply a hasGlyphandContent  class when display is glyph only", () => {
```
nalogiudice: Some extra spacing in the description

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
```diff
@@ -0,0 +1,35 @@
...
+            "title": "Selected",
+            "type": "boolean"
+        },
+        "selectedContent": {
+            "title": "Selected text",
+            "type": "object"
+        },
+        "unselectedContent": {
+            "title": "Unselected text",
+            "type": "object"
```
janechu: These are not objects in the sense that you are not enumerating properties on a regular object, they are instead ReactNodes, now if they are not react renderprops you should use place these in "reactProperties" with the type set as "children" see: https://github.com/Microsoft/fast-dna/tree/master/packages/fast-form-generator-react#react-children

Looks like you added the property below, just need to place these there with the appropriate type set.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        } else {
+            classNames = `${classNames} ${this.getAppearanceString(
+                ActionToggleAppearance.lightweight
+            )}`;
+        }
+
+        if (this.hasGlyphAndContent()) {
+            classNames = `${classNames} ${get(
+                this.props,
+                "managedClasses.actionToggle__hasGlyphAndContent"
```
nalogiudice: I think you need the trailing `,""` for all these get class name calls

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+                "managedClasses.actionToggle__disabled"
+            )}`;
+        }
+
+        if (this.state.selected) {
+            classNames = `${classNames} ${get(
+                this.props,
+                "managedClasses.actionToggle__selected"
+            )}`;
+            classNames = `${classNames} ${this.getAppearanceString(
```
nalogiudice: Why do you need the `getAppearanceString` function at all? Why can't you just get the class name here?

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+        };
+
+        const rendered: any = mount(<ActionToggle {...props} />);
+
+        expect(rendered.find("button").prop("className")).toContain(
+            "actionToggle__disabled"
+        );
+    });
+
+    test("should apply a custom class-name", () => {
```
janechu: nit: prefer "should add a custom class name if a class name has been provided"

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+                  !isNullOrUndefined(this.props.unselectedContent);
+    }
+
+    /**
+     * Handles onClick
+     */
+    private handleToggleChange = (e: React.MouseEvent<HTMLElement>): void => {
+        if (typeof this.props.selected !== "boolean") {
+            this.setState({ selected: !this.state.selected });
+        }
```
nalogiudice: nit: line breaks between function calls

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        if (this.props.onChange) {
```
nalogiudice: I think you'll need to check if this is a function

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+                }
+                jssStyleSheet={actionToggleButtonOverrides}
+            >
+                {this.renderGlyph()}
+                {this.renderLabel()}
+            </Button>
+        );
+    }
+
+    public getAppearance(): ActionToggleAppearance {
```
janechu: should be private

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        if (this.state.selected) {
+            return ActionToggleAppearance.primary;
+        } else {
```
janechu: no need for else wrapper

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    /**
+     * Returns the appropriate ARIA label
+     */
+    public renderARIALabel(): string {
+        if (this.state.selected) {
+            return this.props.selectedLabel;
+        } else {
```
janechu: no need for else wrapper

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+            return ActionToggleAppearance.primary;
+        } else {
+            return ActionToggleAppearance.lightweight;
+        }
+    }
+
+    /**
+     * Returns the appropriate ARIA label
+     */
+    public renderARIALabel(): string {
```
janechu: should be private, also I believe the casing should be renderAriaLabel

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    /**
+     * Returns the appropriate text label
+     */
+    public renderLabel(): React.ReactNode {
+        if (this.state.selected) {
+            return this.props.selectedContent;
+        } else {
```
janechu: no need for else wrapper

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+            return this.props.selectedLabel;
+        } else {
+            return this.props.unselectedLabel;
+        }
+    }
+
+    /**
+     * Returns the appropriate text label
+     */
+    public renderLabel(): React.ReactNode {
```
janechu: should be private

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+            return this.props.selectedContent;
+        } else {
+            return this.props.unselectedContent;
+        }
+    }
+
+    /**
+     * Render Glyphs
+     */
+    public renderGlyph(): React.ReactNode {
```
janechu: should be private

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        if (this.state.selected) {
+            return this.renderSelectedGlyph();
+        } else {
```
janechu: no need for else wrapper

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+            return this.renderUnselectedGlyph();
+        }
+    }
+
+    public renderSelectedGlyph(): React.ReactNode {
```
janechu: should be private

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+        if (typeof this.props.selectedGlyph === "function") {
+            return this.props.selectedGlyph(
+                get(this.props, "managedClasses.actionToggle_selectedGlyph")
+            );
+        }
+        return null;
+    }
+
+    public renderUnselectedGlyph(): React.ReactNode {
```
janechu: should be private

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
+import * as React from "react";
```
janechu: I'm not sure what's going on here but is the indenting off?

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+            );
+        }
+        return null;
+    }
+
+    /**
+     * Generates class names
+     */
+    protected generateClassNames(): string {
+        let classNames: string = get(this.props, "managedClasses.actionToggle") || "";
```
janechu: you can add the "" as the third argument in the get instead

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
```diff
@@ -0,0 +1,220 @@
...
+            classNames = `${classNames} ${get(
+                this.props,
+                "managedClasses.actionToggle__hasGlyphAndContent"
+            )}`;
+        }
+
+        return super.generateClassNames(classNames);
+    }
+
+    private getAppearanceString(appearance: ActionToggleAppearance): string {
```
janechu: can you rename to "getAppearanceClassName"?

packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
nalogiudice: I thought we decided the opposite? In a styled component, the base class name should always be there. That is what makes it styled.

@nicholasrice I feel like we have been around about this more than once...

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: +1 also class name is two words

packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
chrisdholt: I think Jane is correct here, all class names should be optional. In a previous PR I think this was incorrectly implied.

nalogiudice: Can we have a formal statement on why this is the case, so we don't run into this back-and-forth again?

packages/fast-components-react-msft/src/action-toggle/README.md
Falkicon: Tooltip is coming fairly soon. If we remove this we should file a task to re-add it once tooltip is completed.

packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
Falkicon: Maybe a good topic to note in our contributing documentation if we are getting confused.

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,13 @@
+# Action Toggle
```
Falkicon: Please add one line break between headings and paragraphs in markdown. (this rule applies to all markdown 'elements').

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,13 @@
...
+*Action toggle* should be used as control on or around an interactive region of the page that affects the region or items within the region. It usually represents a switch between one of two states.
```
Falkicon: "should be used as a control" not "should be used as control".

packages/fast-components-react-msft/src/action-toggle/README.md
Falkicon: You can remove the word "clearly" in "text and glyph clearly represent"

packages/fast-components-react-msft/src/action-toggle/README.md
Falkicon: "string parameter applied as a class name"

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Usage
+Use *action toggle* for specific control around a functional state. Labeling can be done using a glyph and text, but could also contain just a glyph or just text. Make sure the text and glyph clearly represent the expected function. When only a glyph is used to label the button, include a tooltip in a span element following the button element.
+
+The "selectedGlyph" and "unselectedGlyph" props take a function that returns a ReactNode to render the glyph (ie. a [render prop](https://reactjs.org/docs/render-props.html)).  Note that the function should accept a string parameter that is applied as a classname to the top level element of the resulting node.  This classname  will be populated by the *action toggle* component to use an internally generated class ("actionToggle_selectedGlyph" & "actionToggle_unselectedGlyph") to apply to the glyphs for styling. 
+
+## Style guidance
+Multiple related *action toggle* controls should be grouped together visually to reinforce their associated functionality. Place the toggle where it is not hidden or obscured in the clutter of the page.
```
Falkicon: You can remove the word "together". e.g. "controls should be grouped visually to reinforce"

packages/fast-components-react-msft/src/action-toggle/README.md
Falkicon: "The aria-label is important when ..."

Falkicon: "Otherwise, a non-sighted user would not get the context of the toggle change."

Falkicon: If the tooltip sentence stays: "If a more verbose description is needed, a tooltip is an option to provide additional context."

packages/fast-components-react-msft/src/action-toggle/README.md
scomea: Seems like unnecessary churn to me to remove tooltip references only to restore them in the short term?

packages/fast-components-react-msft/src/action-toggle/README.md
scomea: Applied Jason's edits.  Same notion about tooltip as above, seems like unnecessary churn.

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
scomea: Changed to string which works and is editable but does not enable the full range of options for this prop which takes a ReactNode

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
scomea: changed

packages/fast-components-react-msft/src/action-toggle/README.md
chrisdholt: I like keeping the tooltip so long as we are not referencing it as a component. It's a nit, but I would suggest ensuring that we call it out as "a tooltip" so that we aren't referencing something internal ("tooltip" as a thing we have), but rather the accessible pattern/control in general.

packages/fast-components-react-msft/src/action-toggle/README.md
chrisdholt: I like keeping tooltip so long as we reference it in terms of "a tooltip" (generic pattern), rather than "tooltip" (component)

packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
nalogiudice: Not seeing this resolved, though it is marked resolved.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
scomea: That does not actually return the string we're checking for.  Hardcoding the string being checked for seems to be a consistent pattern - search the repo for 'prop("className")).toContain('  and you get a bunch of hits.  Is there an example of a better way?  (I'm also realizing that the managedClasses declaration at the top might not serve any useful purpose)

chrisdholt: Managed classes declaration at the top is typically useful to remove duplication (not having to declare that object over and over and over for individual tests). As far as checking via accessing the managed classes const - can we write an issue to address this? I think it warrants discussion regarding cost/benefit and any potential gotcha's. I don't think it's necessary to block this PR on that change unless there is a clear problem being solved in this PR

packages/fast-components-class-name-contracts-msft/src/action-toggle/index.ts
scomea: Made optional for this and action-trigger as well (it was also incorrect)

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
scomea: worked, thx

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: Actually don't need that function at all and can put the logic here.  I guess I thought I might need to use it elsewhere at some point.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: yep

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+        const rendered: any = mount(
+            <ActionToggle
+                className={"custom-class-name"}
+                {...props}
+                onChange={onChange}
+            />
+        );
+
+        expect(rendered.find("button").prop("className")).not.toContain(
+            "actionToggle__selected"
```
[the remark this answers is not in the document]
scomea: see above

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: I forgot the `.` at the beginning, so that is why it probably didn't work. This can be done in another task to clean up usage, but I do feel like it is the more correct way, as per my reasoning.

packages/fast-components-react-msft/src/action-toggle/README.md
nalogiudice: I am seeing a lot of "resolved" comments without changes. For instance, above class name is still one word, yet this  was marked as resolved. If it is resolved, I would expect the change OR if there is a disagreement, there should be a follow-up comment regarding it.

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
nalogiudice: So shouldn't it be removed here now?

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: Not seeing resolution or reasoning for not resolving

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: Not seeing resolution or reasoning for not resolving

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: Not seeing resolution or reasoning for not resolving

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: Not seeing resolution or reasoning for not resolving

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+    test("should apply the custom selected ARIA label", () => {
+        const props: ActionToggleHandledProps = {
+            selectedLabel: "selected",
+            unselectedLabel: "unselected",
+            selected: true,
+        };
+
+        const rendered: any = mount(
+            <ActionToggle className={"custom-class-name"} {...props} />
```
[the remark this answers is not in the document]
nalogiudice: Not seeing resolution or reasoning for not resolving

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
nalogiudice: Not seeing resolution or reasoning for not resolving

packages/fast-components-react-msft/src/action-toggle/README.md
scomea: Because I have not yet pushed up an updated changelist yet?

packages/fast-components-react-msft/src/action-toggle/README.md
scomea: This? -"When only a glyph is used to label the button other elements, such as a tooltip, may be required to make the toggle's function clear."  I think the previous blurb was overly prescriptive.

packages/fast-components-react-msft/src/action-toggle/README.md
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/action-toggle.schema.json
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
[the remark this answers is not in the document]
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
scomea: Updated with changes.

packages/fast-components-react-msft/src/action-toggle/README.md
scomea: Yep, tooltips exist even if we don't have a component yet.  How about "If a more verbose description is needed, a tooltip is an option to provide additional context."?

packages/fast-components-react-msft/src/action-toggle/action-toggle.spec.tsx
scomea: changes uploaded

packages/fast-components-react-msft/src/action-toggle/README.md
```diff
@@ -0,0 +1,17 @@
+# Action Toggle
+
+*Action toggle* should be used as a control on or around an interactive region of the page that affects the region or items within the region. It usually represents a switch between one of two states.
+
+## Usage
+
+Use *action toggle* for specific control around a functional state. Labeling can be done using a glyph and text, but could also contain just a glyph or just text. Make sure the text and glyph represent the expected function. When only a glyph is used to label the button, include a tooltip in a span element following the button element.
+
+The "selectedGlyph" and "unselectedGlyph" props take a function that returns a ReactNode to render the glyph (ie. a [render prop](https://reactjs.org/docs/render-props.html)).  Note that the function should accept a string parameter that is applied as a class name to the top level element of the resulting node.  This class name  will be populated by the *action toggle* component to use an internally generated class ("actionToggle_selectedGlyph" & "actionToggle_unselectedGlyph") to apply to the glyphs for styling.
```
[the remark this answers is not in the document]
scomea: Good eye.  Fix queued.

packages/fast-components-react-msft/src/action-toggle/README.md
chrisdholt: @scomea yep - exactly. I think the only thing missing from the last sentance above is "a" to denote, "a tooltip". Agree with this 100%

packages/fast-components-react-msft/src/action-toggle/action-toggle.props.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * The action toggle selected ARIA text
+     */
+    selectedLabel: string;
+
+    /**
+     * The onChange event handler
+     */
+    onChange?: (event: React.MouseEvent<HTMLElement>) => void;
```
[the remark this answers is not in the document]
scomea: I changed it to onToggle and added the props to the return value as we do in Context Menu Item.

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: Yep

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: yep

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: yep

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: yep

packages/fast-components-react-msft/src/action-toggle/action-toggle.tsx
scomea: yep

Issue #1265, opened 2019-01-04
Add render props for before and after slots to button


nalogiudice:
We plan on deprecating the idea of slots in the future, so button should accept render props for the before and after positions.

PR #1266, opened 2019-01-04
feat: add render props for before and after slots in button

nalogiudice:
Closes [#1265](https://github.com/Microsoft/fast-dna/issues/1265)

# Description
Adds render props for the before and after button positions. Users can now pass a react node to populate the space before or after the button content. Other components can now access these nodes via a class name.

Added `beforeSlot` and `afterSlot` props. Added functions to render props and add class names. Added appropriate tests.

## Motivation & context
Will allow other components (such as text action) to target the slot.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/button/button.props.ts
```diff
@@ -25,6 +25,16 @@ export interface ButtonHandledProps
      * The Button appearance
      */
     appearance?: ButtonAppearance;
+
+    /**
+     * The preceding slot
+     */
+    beforeSlot?: (classname?: string) => React.ReactNode;
```
chrisdholt: is there a reason not to go with something unrelated to "slots"? I think that may be confusing; what about `beforeContent` an `afterContent`?

nalogiudice: Yeah, I could see the usage of slots being confusing. I'm ok with `beforeContent` and `afterContent` if everyone agrees.

nalogiudice:
@chrisdholt I have updated the call to action usage. Action trigger does not use slots and action toggle doesn't exist yet. I am unable to find other usages.

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -113,6 +113,35 @@ describe("button", (): void => {
         expect(rendered.find("div.slotBefore").length).toBe(1);
     });
 
+    test("should render prop into the before content location if prop exist", () => {
```
chrisdholt: nittiest nit: "exists"

packages/fast-components-react-msft/src/button/button.spec.tsx
```diff
@@ -113,6 +113,35 @@ describe("button", (): void => {
...
+        const props: ButtonHandledProps = {
+            children: "Foo",
+            beforeContent: (classname?: string): React.ReactNode => {
+                return (
+                    <svg
+                        width="16"
```
chrisdholt: do we know if these values should be pixels? Not blocking...genuinely curious as they seem to differ place to place.

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
```diff
@@ -89,6 +85,17 @@ class CallToAction extends Foundation<
 
         return super.generateClassNames(classNames);
     }
+
+    private renderAfterContent(): any {
+        return (): React.ReactNode => {
+            return (
+                <div
```
chrisdholt: nit: inline within button should be a span.

nalogiudice: @chrisdholt  Hm this would be a shape change... I guess it doesn't really matter because it is internal?

packages/fast-components-react-msft/src/button/button.spec.tsx
nalogiudice: These are correct being unitless. Where are you seeing px afterwards?

packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
chrisdholt: It's internal and not exposed by any part of the API, so this isn't breaking whatsoever in my opinion.

PR #1271, opened 2019-01-07
chore: remove component buckets from documentation site

chrisdholt:


nalogiudice:
Guess we changed our mind?

Falkicon:
Yes Nic, we landed a strat that won't require the buckets (at least, not on the doc site).

Issue #1191, opened 2018-12-06
Buttons should apply default colors to svg elements in button content


scomea:
Child elements of our buttons inherit the "color" attribute of the button so that is typically inherited by text elements added as children of the button.  SVG glyphs are commonly used in buttons but are not affected by the color property and use fill instead, therefore they do not inherit the button color by default.  We should treat svg elements as first class citizens of our system and set the fill as well as the color attributes in our default button styles.

**To Reproduce**
Add an svg glyph and text to a MSFT styled button.  Note the text is appropriately colored and the svg is not.

**Expected behavior**
Glyph and text inherit button content colors.

PR #1272, opened 2019-01-07
fix: update button to properly target fill color based on appearance

nalogiudice:
Closes [#1191](https://github.com/microsoft/fast-dna/issues/1191)

# Description

Adds proper fill colors for all cases of glyphs found within button and updates examples and usage in button and text action to match.

## Motivation & context

Issue was noted when creating text action - there was no way to target the inner glyph when it was passed as content to the button. This also targets the newly added `beforeContent` and `afterContent` prop areas.

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -77,7 +85,37 @@ function applyTransaprentBackplateStyles(): CSSRules<DesignSystem> {
                 return disabledContrast(
                     designSystem.contrast,
                     designSystem.foregroundColor,
-                    designSystem.brandColor
+                    designSystem.backgroundColor
+                );
+            },
+            fill: (config: DesignSystem): string => {
+                const designSystem: DesignSystem = withDesignSystemDefaults(config);
```
chrisdholt: @nalogiudice it seems like we're ending up with a fairly significant amount of duplication for `disabledContrast` (3) and `ensureNormalContrast` (4) and use for defining the design system here (7 times). Can we do the work to pass those values into this function so we can define these once instead of redeclaring this multiple times in this function?

packages/fast-components-react-msft/src/button/examples.data.tsx
```diff
@@ -91,17 +88,49 @@ export default {
...
-            children: "Before render props slots",
+            children: "Before render props",
         },
         {
-            children: "After render props slots",
+            children: "After render props",
             afterContent: testGlyph,
         },
+        {
+            children: "Disabled fter render props",
```
scomea: "after"

packages/fast-components-styles-msft/src/button/index.ts
nalogiudice: @chrisdholt ok I did a lot of clean up. There were a bunch of other things imported or created that weren't being used. A lot cleaner now.

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -6,62 +6,42 @@ import {
...
-    normalContrast,
     scaleContrastNormal,
 } from "../utilities/colors";
 import Chroma from "chroma-js";
 import { density } from "../utilities/density";
 import { defaultHeight, maxHeight, minHeight } from "../utilities/height";
 import outlinePattern from "../patterns/outline";
 
-function applyTransaprentBackplateStyles(): CSSRules<DesignSystem> {
+function applyTransaprentBackplateStyles(
```
janechu: nit: can we take a moment to fix the spelling on this

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -71,19 +51,34 @@ function applyTransaprentBackplateStyles(): CSSRules<DesignSystem> {
...
+                fill: disabledForegroundContrast(designSystem),
             },
         },
+        "& $button_beforeContent, & $button_afterContent": {
+            fill: ensureNormalBrandContrast(designSystem),
+        },
     };
 }
 
+function ensureNormalBrandContrast(designSystem: DesignSystem): string {
```
nicholasrice: This function already exists in color utilities

packages/fast-components-react-msft/src/button/examples.data.tsx
```diff
@@ -91,17 +88,49 @@ export default {
...
             afterContent: testGlyph,
         },
         {
             beforeContent: testGlyph,
-            children: "Before render props slots",
+            children: "Before render props",
         },
         {
-            children: "After render props slots",
+            children: "After render props",
```
nicholasrice: minor nit, but this the "after" prop is singular

packages/fast-components-styles-msft/src/button/index.ts
nalogiudice: Resolved

PR #1278, opened 2019-01-09
feat: add badge as an MSFT component

nalogiudice:
Closes [#1278](https://github.com/Microsoft/fast-dna/pull/1278)

# Description
Adds the badge component.

## Motivation & context
Provides the badge component to users.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/badge/badge.props.ts
+++ b/packages/fast-components-react-msft/src/badge/badge.props.ts
@@ -17,8 +17,7 @@ export enum BadgeSize {
 }
 
 export interface BadgeManagedClasses extends ManagedClasses<BadgeClassNameContract> {}
-/* tslint:disable-next-line:no-empty-interface */
-export interface BadgeUnhandledProps {}
+export interface BadgeUnhandledProps extends React.HTMLAttributes<HTMLSpanElement> {}
 export interface BadgeHandledProps extends BadgeManagedClasses {
     /**
      * The badge appearance
--- a/packages/fast-components-react-msft/src/badge/badge.props.ts
+++ b/packages/fast-components-react-msft/src/badge/badge.props.ts
@@ -8,7 +8,6 @@ export enum BadgeAppearance {
     accent = "accent",
     highlight = "highlight",
     lowlight = "lowlight",
-    lightweight = "lightweight",
 }
 
 export enum BadgeSize {
--- a/packages/fast-components-react-msft/src/badge/badge.props.ts
+++ b/packages/fast-components-react-msft/src/badge/badge.props.ts
@@ -1,0 +1,34 @@
+import * as React from "react";
+import {
+    BadgeClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-msft";
+
+export enum BadgeAppearance {
+    accent = "accent",
+    highlight = "highlight",
+    lowlight = "lowlight",
+    lightweight = "lightweight",
+}
+
+export enum BadgeSize {
+    small = "small",
+    large = "large",
+}
+
+export interface BadgeManagedClasses extends ManagedClasses<BadgeClassNameContract> {}
+/* tslint:disable-next-line:no-empty-interface */
+export interface BadgeUnhandledProps {}
+export interface BadgeHandledProps extends BadgeManagedClasses {
+    /**
+     * The badge appearance
+     */
+    appearance?: BadgeAppearance;
+
+    /**
+     * The badge size
+     */
+    size?: BadgeSize;
+}
+
+export type BadgeProps = BadgeHandledProps & BadgeUnhandledProps;
--- a/packages/fast-components-react-msft/src/badge/badge.spec.tsx
+++ b/packages/fast-components-react-msft/src/badge/badge.spec.tsx
@@ -96,4 +96,10 @@ describe("badge", (): void => {
             managedClasses.badge__large
         );
     });
+
+    test("should apply only a 'badge' html class when no appearance is passed", () => {
+        const rendered: any = mount(<Badge />);
+
+        expect(rendered.find("span").prop("className")).toContain(managedClasses.badge);
+    });
 });
--- a/packages/fast-components-react-msft/src/badge/badge.spec.tsx
+++ b/packages/fast-components-react-msft/src/badge/badge.spec.tsx
@@ -36,7 +36,6 @@ describe("badge", (): void => {
             shallow(<MSFTBadge />);
             shallow(<MSFTBadge appearance={BadgeAppearance.lowlight} />);
             shallow(<MSFTBadge appearance={BadgeAppearance.highlight} />);
-            shallow(<MSFTBadge appearance={BadgeAppearance.lightweight} />);
             shallow(<MSFTBadge appearance={BadgeAppearance.accent} />);
         }).not.toThrow();
     });
--- a/packages/fast-components-react-msft/src/badge/badge.spec.tsx
+++ b/packages/fast-components-react-msft/src/badge/badge.spec.tsx
@@ -1,0 +1,100 @@
+import * as React from "react";
+import * as Adapter from "enzyme-adapter-react-16";
+import { configure, mount, shallow } from "enzyme";
+import { BadgeClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import MSFTBadge from "./badge";
+import {
+    Badge,
+    BadgeAppearance,
+    BadgeHandledProps,
+    BadgeProps,
+    BadgeSize,
+    BadgeUnhandledProps,
+} from "./index";
+
+/*
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("badge", (): void => {
+    const managedClasses: BadgeClassNameContract = {
+        badge: "badge",
+        badge__small: "badge__small",
+        badge__large: "badge__large",
+        badge__lowlight: "badge__lowlight",
+        badge__highlight: "badge__highlight",
+        badge__accent: "badge__accent",
+    };
+
+    test("should have a displayName that matches the component name", () => {
+        expect((MSFTBadge as any).name).toBe(MSFTBadge.displayName);
+    });
+
+    test("should not throw if managedClasses are not provided", () => {
+        expect(() => {
+            shallow(<MSFTBadge />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.lowlight} />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.highlight} />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.lightweight} />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.accent} />);
+        }).not.toThrow();
+    });
+
+    test("should accept unhandledProps", () => {
+        const handledProps: BadgeHandledProps = {
+            size: BadgeSize.small,
+            appearance: BadgeAppearance.lowlight,
+        };
+
+        const unhandledProps: BadgeUnhandledProps = {
+            "aria-hidden": true,
+        };
+
+        const props: BadgeProps = { ...handledProps, ...unhandledProps };
+
+        const rendered: any = mount(<Badge {...props} />);
+
+        expect(rendered.find("span").prop("aria-hidden")).toEqual(true);
+    });
+
+    test("should apply a 'lowlight' html class when appearance is lowlight", () => {
+        const rendered: any = mount(<Badge appearance={BadgeAppearance.lowlight} />);
+
+        expect(rendered.find("span").prop("className")).toContain(
+            managedClasses.badge__lowlight
+        );
+    });
+
+    test("should apply a 'highlight' html class when appearance is highlight", () => {
+        const rendered: any = mount(<Badge appearance={BadgeAppearance.highlight} />);
+
+        expect(rendered.find("span").prop("className")).toContain(
+            managedClasses.badge__highlight
+        );
+    });
+
+    test("should apply an 'accent' html class when appearance is accent", () => {
+        const rendered: any = mount(<Badge appearance={BadgeAppearance.accent} />);
+
+        expect(rendered.find("span").prop("className")).toContain(
+            managedClasses.badge__accent
+        );
+    });
+
+    test("should apply a 'small' html class when size is small", () => {
+        const rendered: any = mount(<Badge size={BadgeSize.small} />);
+
+        expect(rendered.find("span").prop("className")).toContain(
+            managedClasses.badge__small
+        );
+    });
+
+    test("should apply a 'large' html class when size is large", () => {
+        const rendered: any = mount(<Badge size={BadgeSize.large} />);
+
+        expect(rendered.find("span").prop("className")).toContain(
+            managedClasses.badge__large
+        );
+    });
+});
--- a/packages/fast-components-react-msft/src/badge/badge.tsx
+++ b/packages/fast-components-react-msft/src/badge/badge.tsx
@@ -14,7 +14,6 @@ class Badge extends Foundation<BadgeHandledProps, BadgeUnhandledProps, {}> {
     public static displayName: string = "Badge";
 
     public static defaultProps: Partial<BadgeProps> = {
-        appearance: BadgeAppearance.lowlight,
         size: BadgeSize.small,
     };
 
@@ -66,6 +65,8 @@ class Badge extends Foundation<BadgeHandledProps, BadgeUnhandledProps, {}> {
                     "managedClasses.badge__accent",
                     ""
                 )}`;
+                break;
+            default:
                 break;
         }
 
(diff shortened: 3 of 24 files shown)


packages/fast-components-react-msft/src/badge/badge.props.ts
```diff
@@ -0,0 +1,34 @@
...
+    BadgeClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-msft";
+
+export enum BadgeAppearance {
+    accent = "accent",
+    highlight = "highlight",
+    lowlight = "lowlight",
+    lightweight = "lightweight",
+}
```
chrisdholt: Even though this is an MSFT concept/pattern, I don't think we want to set these from enums. We'll want a more extensible solution here. Another scenario I foresee is wanting the "lightweight" to be a different color than just matching the "foreground" color. Perhaps backplate is a prop that get's passed/set; will need to consider the scenario of the text and a11y as well. But don't think we want to limit the options to just these enums.

chrisdholt: Confirmed with @Falkicon that we'll want a more extensible solution

packages/fast-components-react-msft/src/badge/badge.props.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export enum BadgeSize {
+    small = "small",
+    large = "large",
+}
+
+export interface BadgeManagedClasses extends ManagedClasses<BadgeClassNameContract> {}
+/* tslint:disable-next-line:no-empty-interface */
+export interface BadgeUnhandledProps {}
```
chrisdholt: This should extend the correct React interface

packages/fast-components-styles-msft/src/index.ts
```diff
@@ -5,6 +5,10 @@ import ActionTriggerStyles from "./action-trigger";
 export { ActionTriggerStyles };
 export * from "./action-trigger";
 
+import BadgeStyles from "./badge";
+export { BadgeStyles };
+export * from "./badge";
```
chrisdholt: I don't think that the export * is exporting anything additional here.

packages/fast-components-styles-msft/src/badge/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    const lowlightForegroundColor: string = ensureNormalContrast(
+        contrast,
+        backgroundColor,
+        lowlightBackground
+    );
+
+    return {
+        badge: {
+            fontSize: "13px",
+            lineHeight: "20px",
```
chrisdholt: everything should be on the new type ramp.

packages/fast-components-styles-msft/src/badge/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    ellipsis,
+    toPx,
+} from "@microsoft/fast-jss-utilities";
+import { DesignSystem, withDesignSystemDefaults } from "../design-system/index";
+import { ensureNormalContrast } from "../utilities/colors";
+import { fontWeight } from "../utilities/fonts";
+
+function smallBadgeStyle(direction: Direction): CSSRules<DesignSystem> {
+    return {
+        padding: direction === Direction.ltr ? "0 7px 2px 0" : "0 0 2px 7px",
```
nicholasrice: pretty sure we have a spacing utility to do this

packages/fast-components-styles-msft/src/badge/index.ts
chrisdholt: 14 or 12; I would ask @Falkicon, but this would create issues by being the only thing off the typeramp. It would be one thing if the delta was a line-height only; but the font being different and off doesn't really seem to make sense aside from the fact that we're porting this over from a previous implementation. Everything added should really be considered net-new work.

chrisdholt: I also noticed that this is different than "smallBadge" above.

packages/fast-components-styles-msft/src/badge/index.ts
```diff
@@ -0,0 +1,108 @@
...
+        contrast,
+        backgroundColor,
+        lowlightBackground
+    );
+
+    return {
+        badge: {
+            fontSize: "13px",
+            lineHeight: "20px",
+            ...ellipsis(),
```
chrisdholt: I see that we're adding ellipsis; assuming we need maxlines to ensure a single line of text as well?

packages/fast-components-react-msft/src/badge/badge.props.ts
nalogiudice: So, trying to understand the API... backplate would be prop that would take what exactly?

packages/fast-components-styles-msft/src/index.ts
chrisdholt: In the MSFT styles package? I'm not seeing those as being imported.

packages/fast-components-react-msft/src/badge/badge.props.ts
nalogiudice: Fixed.

packages/fast-components-styles-msft/src/index.ts
nalogiudice: Sorry, wrong package. Fixed.

packages/fast-components-styles-msft/src/badge/index.ts
nalogiudice: No, it will never go over one line, but without a max-width, the ellipsis will never show... this will just be one line indefinitely. The old version did not have a max-width. Perhaps this one should.

packages/fast-components-styles-msft/src/badge/index.ts
nalogiudice: Yeah, whoops. Changed.

packages/fast-components-styles-msft/src/badge/index.ts
nalogiudice: Have changed to `14px`. The small version's height had to be updated from `18px` to `20px` There is just not enough room inside to include the text (especially since it won't be all caps now).

packages/fast-components-react-msft/src/badge/badge.tsx
```diff
@@ -0,0 +1,89 @@
...
+                )}`;
+                break;
+            case BadgeAppearance.lowlight:
+                classNames = `${classNames} ${get(
+                    this.props,
+                    "managedClasses.badge__lowlight",
+                    ""
+                )}`;
+                break;
+            case BadgeAppearance.accent:
```
janechu: This switch never checks for lightweight, why do we have it as part of BadgeAppearance?

packages/fast-components-react-msft/src/badge/badge.props.ts
chrisdholt: I think we need to tease out the API and land something extensible. I'm happy to propose in a solution, but I won't be able to jump on it immediately. The enum approach doesn't represent the type of flexibility that I think we want from this component considering all the variations of backplates, and even variations within various MSFT usages of the text color when no backplate exists.

packages/fast-components-react-msft/src/badge/badge.tsx
nalogiudice: "Lightweight" is just a badge, with no additional class name added. In the editor, you would switch to lightweight to see that appearance, hence it being in appearance.

I think this will change though based on @chrisdholt 's comments

packages/fast-components-react-msft/src/badge/badge.props.ts
nalogiudice: Perhaps we could accept an object that takes `backplateColor` and `textColor`, and also accepts the enum above to provide some "default" settings?

Falkicon:
Why do we think this is a MSFT only pattern, should this have a base? Badges are quite universal...

packages/fast-components-react-msft/src/badge/badge.props.ts
nicholasrice: Is this over complicating it? If I want to change the CSS I can do that through JSSManager overrides like any other component.

chrisdholt: I don't think we need a complex object - the component is flat. I think the issue with overrides is that there hasn't been a discussion on what the default states should be here. We simply have a port from a previous internal library.

 I've seen this have 3 or 4 different treatments from one MSFT team alone. The size and weight of the text is consistent as well is the size of backplates, but the text and backplate color regularly varies based on context. I think we need a more flexible solution or at least more thought as to what the defaults should be.

nicholasrice: @chrisdholt I think not knowing a the defaults is a different issue than how we describe those defaults. At a base level, we have named options of known patterns, and outliers can be created using `jssStyleSheet` props.

If we don't know the defaults then we obviously need to figure those out, but I'm not sure that means named enum values is a bad approach. What are the 3 or 4 that you have seen?

chrisdholt: synced offline. I'm open to using enums but we need to validate the default states and values first and confirm with @Falkicon.

chrisdholt:
> Why do we think this is a MSFT only pattern, should this have a base? Badges are quite universal...

> @Falkicon base components have no sense of style - as in they are completely unstyled. The only other route would be to build `badge` in base (the shape) and then add the styles in MSFT.

@nalogiudice, @Falkicon - I would be interested in seeing if someone could make the case for a base component. It would be helpful to see if MSFT has an cases of these being actionable (something VERY similar to "lightweight" has been proposed as actionable), as well as how these manifest in other design systems.

nalogiudice:
I was just having a discussion on the base components - not saying any one person is saying something or not. And I was taking a look at the libraries and noticed that the only two base components not also found in MSFT were those two, at the time I thought it was relevant, but I suppose it is not - just ignore.

>  I would be interested in seeing if someone could make the case for a base component

nalogiudice:
@chrisdholt @Falkicon  And to be clear, I'm fine with putting this as a base component. If users want to override in MSFT that is fine - if they want to take only the base component and do all the styles, that is fine too.

nalogiudice:
Yeah @janechu - I am awaiting feedback from @Falkicon or @chrisdholt

chrisdholt:
> Why do we think this is a MSFT only pattern, should this have a base? Badges are quite universal...

@nalogiudice @Falkicon I propose that we make this a base component (sorry @nalogiudice...). I did a quick search for three major design systems and came up with the following badge instances:

* [material-ui](https://material-ui.com/demos/badges/)
* [lightning design](https://www.lightningdesignsystem.com/components/badges/)
* [bootstrap](https://getbootstrap.com/docs/4.0/components/badge/)
* [foundation](https://foundation.zurb.com/sites/docs/badge.html)

I think we'll want to include the ability to have this be "actionable" in the scenario where it shows categories...or similar.

packages/fast-components-react-msft/src/badge/badge.props.ts
```diff
@@ -0,0 +1,33 @@
...
+import {
+    BadgeClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-msft";
+
+export enum BadgeAppearance {
+    accent = "accent",
+    highlight = "highlight",
+    lowlight = "lowlight",
+    lightweight = "lightweight",
```
chrisdholt: @nalogiudice I don't see lightweight being used or included anywhere else - can we remove this?

packages/fast-components-react-msft/src/badge/badge.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+    test("should have a displayName that matches the component name", () => {
+        expect((MSFTBadge as any).name).toBe(MSFTBadge.displayName);
+    });
+
+    test("should not throw if managedClasses are not provided", () => {
+        expect(() => {
+            shallow(<MSFTBadge />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.lowlight} />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.highlight} />);
+            shallow(<MSFTBadge appearance={BadgeAppearance.lightweight} />);
```
chrisdholt: Looks like we can remove lightweight from here

packages/fast-components-styles-msft/src/badge/index.ts
```diff
@@ -0,0 +1,109 @@
...
+} from "@microsoft/fast-jss-utilities";
+import { DesignSystem, withDesignSystemDefaults } from "../design-system/index";
+import { applyTypeRampConfig } from "../utilities/typography";
+import { ensureNormalContrast } from "../utilities/colors";
+import { fontWeight } from "../utilities/fonts";
+
+function smallBadgeStyle(direction: Direction): CSSRules<DesignSystem> {
+    return {
+        padding: localizeSpacing(Direction.ltr)(`0 ${toPx(7)} ${toPx(5)} 0`),
+        height: "13px",
```
chrisdholt: @nalogiudice would be good to get design feedback on this height being 13px - seems off considering new type ramp values.

nalogiudice: The final height is not 13px - small comes out to 20px and large to 26px when rendered (as shown in the comp)

packages/fast-components-react-msft/src/badge/badge.props.ts
nalogiudice: It is, in fact, being used. When switching in the editor, lightweight is the non-backplate badge. Instead of adding a class, it is just plain `badge`. The other backplate classes then get applied.

packages/fast-components-react-msft/src/badge/badge.spec.tsx
nalogiudice: See other comment.

packages/fast-components-styles-msft/src/badge/index.ts
nalogiudice: I put in a max width of `350px` for now... design can weigh-in on final thoughts.

packages/fast-components-styles-msft/src/badge/index.ts
bheston: I've updated the docs to make the sizes more apparent, but for large the text should be 14/20 with a padding of 3px 12px, with an overall height of 26px. Small should be 12/16 with a padding of 2px 8px, overall height of 20px. The letter spacing should all be default.

packages/fast-components-styles-msft/src/badge/index.ts
bheston: 350px is pretty wide, but that's fine. I can't think of a good use for this component with lots of text though, so we should never hit this case.

packages/fast-components-styles-msft/src/badge/index.ts
bheston: 20px and 26px are correct. If the top and bottom padding have to be different to align the text correctly that's fine, but the type ramp line heights are 16px and 20px respectively.

packages/fast-components-react-msft/src/badge/badge.tsx
chrisdholt: I think we should either add the class here or make the enum optional. In my opinion the default is just the default and the enums are modifiers. This is really just tricking the system. I think if we make this optional then things work as expected. Default badge with no modifier has a lightweight appearance.

packages/fast-components-styles-msft/src/badge/index.ts
nalogiudice: Since this is just style, it wouldn't be a breaking change to alter it in the future. Perhaps we could go a bit smaller and see how it works out. The entire point of the having the ellipses is that they actually come into play. Maybe 200 or 225?

packages/fast-components-react-msft/src/badge/badge.props.ts
nalogiudice: @bheston verifying that these are good default styles
screen shot 2019-02-07 at 11 05 18 am

packages/fast-components-react-msft/src/badge/badge.tsx
nalogiudice: Removing lightweight

packages/fast-components-react-msft/src/badge/badge.spec.tsx
nalogiudice: Lightweight removed

packages/fast-components-react-msft/src/badge/badge.props.ts
nalogiudice: Lightweight removed

packages/fast-components-styles-msft/src/badge/index.ts
nalogiudice: Went with `215px` - seems pretty good for now.

packages/fast-components-react-msft/src/badge/badge.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+        );
+    });
+
+    test("should apply a 'small' html class when size is small", () => {
+        const rendered: any = mount(<Badge size={BadgeSize.small} />);
+
+        expect(rendered.find("span").prop("className")).toContain(
+            managedClasses.badge__small
+        );
+    });
```
chrisdholt: should we write a test to check that a default is passed?

nalogiudice: Added

PR #1283, opened 2019-01-10
feat: add dark theme and set as default to fast-development-site

eljefe223:
# Description

Updated styling for fast form and dev site to new Dark Theme.

## Motivation & context

This is the ground work for the new fast creator app.

## Issue type checklist

## Process & policy checklist


packages/fast-development-site-react/src/components/site/configuration-panel.tsx
```diff
@@ -57,10 +68,11 @@ const style: ComponentStyles<ConfigurationPanelManagedClasses, DevSiteDesignSyst
             padding: `${toPx(6)} ${toPx(8)}`,
             margin: toPx(2),
             minWidth: toPx(25),
+            color: "white",
```
chrisdholt: HEX please.

packages/fast-development-site-react/src/components/site/configuration-panel.tsx
```diff
@@ -57,10 +68,11 @@ const style: ComponentStyles<ConfigurationPanelManagedClasses, DevSiteDesignSyst
...
             backgroundPosition: "center",
             "&:hover": {
                 cursor: "pointer",
-                background: "rgba(0, 0, 0, 0.04)",
+                background: "rgba(255, 255, 255, 0.04)",
```
chrisdholt: Assuming this doesn't connect to the design system provider

packages/fast-development-site-react/src/components/site/dev-tools.tsx
```diff
@@ -112,7 +119,7 @@ const style: ComponentStyles<DevToolsManagedClasses, DevSiteDesignSystem> = {
         border: "none",
         "&:after": {
             position: "absolute",
-            background: "#000000",
+            background: "#F2F2F2",
```
chrisdholt: No design system value?

packages/fast-development-site-react/src/components/site/dev-tools.tsx
```diff
@@ -122,7 +129,7 @@ const style: ComponentStyles<DevToolsManagedClasses, DevSiteDesignSystem> = {
         },
         "&:before": {
             position: "absolute",
-            background: "#000000",
+            background: "#F2F2F2",
```
chrisdholt: No design system value?

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -173,7 +173,7 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
             borderRadius: toPx(2),
             width: toPx(1),
             height: toPx(10),
-            background: "#000000",
+            background: "#F2F2F2",
```
chrisdholt: No design system values?

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -202,13 +202,8 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
...
             display: "none",
         },
-        "&:hover": {
-            boxShadow: `inset 0 0 ${toPx(2)} 0 rgba(0,0,0, .3)`,
-        },
         "&:focus": {
-            boxShadow: (config: DevSiteDesignSystem): string => {
-                return `inset 0 0 0 1 ${config.brandColor}`;
-            },
+            boxShadow: "inset 0 0 0 1px #FB356D",
```
chrisdholt: Wondering why we have colors defined in the design system but we aren't using them?

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -223,6 +218,16 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
...
+        "&::-webkit-scrollbar": {
+            width: "6px",
+            height: "6px",
+            overflow: "auto",
+            float: "left",
+        },
+        "&::-webkit-scrollbar-thumb": {
+            background: "#303030",
+            borderRadius: "2px",
+        },
```
chrisdholt: This seems to be duplicated throughout this package; can we create a utility for this instead of redefining it over and over? Maintenance will be painful if we want to change these.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -223,6 +218,16 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
...
         overflow: "auto",
+        "&::-webkit-scrollbar": {
+            width: "6px",
+            height: "6px",
+            overflow: "auto",
+            float: "left",
+        },
+        "&::-webkit-scrollbar-thumb": {
+            background: "#303030",
+            borderRadius: "2px",
```
chrisdholt: I'd suggest wiring up a design system value for this.

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -245,6 +250,9 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
         alignSelf: "flex-start",
     },
     site_paneToggleButtonIcon: {
+        "& svg": {
```
chrisdholt: Is there a reason we are targeting all SVG elements?

packages/fast-development-site-react/src/components/design-system/index.ts
```diff
@@ -2,7 +2,12 @@ export interface DevSiteDesignSystem {
     foregroundColor: string;
     backgroundColor: string;
     brandColor: string;
-    lightGray: string;
+    background000: string;
+    background100: string;
+    background200: string;
+    background300: string;
+    background800: string;
+    foreground300: string;
```
chrisdholt: Since we are setting defaults, we should make these optional so that we don't break anyone.

packages/fast-development-site-react/src/components/site/configuration-panel.tsx
```diff
@@ -39,6 +39,17 @@ const style: ComponentStyles<ConfigurationPanelManagedClasses, DevSiteDesignSyst
...
         overflowX: "auto",
+        color: "white",
+        "&::-webkit-scrollbar": {
+            width: "6px",
+            height: "6px",
+            overflow: "auto",
+            float: "left",
+        },
+        "&::-webkit-scrollbar-thumb": {
+            background: "#303030",
```
nalogiudice: Isn't this `background800`?

packages/fast-development-site-react/src/components/site/configuration-panel.tsx
```diff
@@ -57,10 +68,11 @@ const style: ComponentStyles<ConfigurationPanelManagedClasses, DevSiteDesignSyst
             padding: `${toPx(6)} ${toPx(8)}`,
```
nalogiudice: nit: none of these need to be `toPx()`

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: Site level design system values are not working. The design system is null. so any design system declarations on this page were not working.

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: Same issue design system is null at this level.

packages/fast-development-site-react/src/components/site/configuration-panel.tsx
eljefe223: I think we should make a seperate task to clean these up. I removed them all from Fast Form but since this package will be rebuilt soon I figured I would save that work for that task.

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: Keeping as is until the design system issue is fixed.

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: Removed and added fill directly on `site_paneToggleButtonIcon`.

packages/fast-development-site-react/src/components/site/configuration-panel.tsx
eljefe223: We should do a pass on these later when we hook up the themes on the dev site

chrisdholt: @eljefe223 I'm curious if we should target a feature branch then so we can push the feature as "complete". I wouldn't be comfortable with the possibility of this deploying without a user being able to opt-out of dark theme and opt-in to light (or their own). Let's discuss further.

eljefe223: @chrisdholt sounds like a good plan to me. @janechu @nicholasrice thoughts?

nicholasrice: The review title is "add dark theme and set as default" - can we *not* set it as default?

chrisdholt: This will set it as "the way it is" with no way to change

janechu:
You've marked this as a bug fix in your PR description, but it is a feature, please update.

packages/fast-development-site-react/app/components/button/README.md
```diff
@@ -1,2 +1,2 @@
 # Button
-The button test component.
+The button test component. [testURL](https://www.microsoft.com)
```
janechu: I've realized this is confusing to anyone looking at the contents of the `app` folder so while you are here, can you add a `README.md` to the root of the `app` folder with the following content:

# Testing application

This folder is for a testing the output of the development site and which mocks up some of the scenarios we expect the development site to encounter. This is not an exportable part of this package.

packages/fast-development-site-react/src/components/breadcrumb/breadcrumb-item.tsx
```diff
@@ -23,7 +23,7 @@ const style: ComponentStyles<BreadcrumbItemManagedClasses, DevSiteDesignSystem>
         "&::after": {
             content: "'\\002F'",
             padding: `0 0 0 ${toPx(10)}`,
-            color: (config: DevSiteDesignSystem): string => config.foregroundColor,
+            color: (config: DevSiteDesignSystem): string => config.foreground300,
```
janechu: can you add a fallback to the devSiteDesignSystemDefaults here

packages/fast-development-site-react/src/components/shell/info-bar.tsx
```diff
@@ -17,7 +17,10 @@ export interface ShellInfoBarManagedClasses {
 const style: ComponentStyles<ShellInfoBarManagedClasses, DevSiteDesignSystem> = {
     shellInfoBar: {
         display: "flex",
-        background: "#E9ECEC",
+        color: (config: DevSiteDesignSystem): string => {
+            return config.foreground300 || devSiteDesignSystemDefaults.foreground300;
+        },
+        background: "#111",
```
janechu: should this be using a value from devSiteDesignSystemDefaults?

packages/fast-development-site-react/src/components/site/category-documentation.tsx
```diff
@@ -20,6 +20,9 @@ const style: ComponentStyles<CategoryDocumentationManagedClasses, DevSiteDesignS
     documentationPanel: {
         maxWidth: "1000px",
         padding: "14px",
+        "& a": {
+            color: (config: DevSiteDesignSystem): string => config.brandColor,
```
janechu: should this have a fallback to devSiteDesignSystemDefaults?

packages/fast-development-site-react/src/components/site/component-view.tsx
```diff
@@ -34,10 +34,14 @@ const style: ComponentStyles<ComponentViewManagedClasses, DevSiteDesignSystem> =
         display: "grid",
         gridTemplateColumns: "1fr 1fr 1fr",
         backgroundColor: (config: DevSiteDesignSystem): string => config.backgroundColor,
```
janechu: should this use a fallback to devSiteDesignSystemDefaults?

packages/fast-development-site-react/src/components/site/component-view.tsx
```diff
@@ -34,10 +34,14 @@ const style: ComponentStyles<ComponentViewManagedClasses, DevSiteDesignSystem> =
...
+        ...applyScrollbarStyle(),
     },
     componentDetailView: {
         overflow: "auto",
         flexGrow: "1",
+        background: (config: DevSiteDesignSystem): string => config.background100,
+        color: (config: DevSiteDesignSystem): string => config.foreground300,
```
janechu: same comments here, fallback to devSiteDesignSystemDefaults?

packages/fast-development-site-react/src/components/site/component-wrapper.tsx
```diff
@@ -36,7 +36,7 @@ const checker: string =
     "url('data:image/gif;base64,R0lGODlhEAAQAPABANfX1////yH5BAAAAAAALAAAAAAQABAAAAIfhG+hq4jM3IFLJhoswNly/XkcBpIiVaInlLJr9FZWAQAh/wtYTVAgRGF0YVhNUDw/eHBhY2tldCBiZWdpbj0i77u/IiBpZD0iVzVNME1wQ2VoaUh6cmVTek5UY3prYzlkIj8+Cjx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8iIHg6eG1wdGs9IkFkb2JlIFhNUCBDb3JlIDUuNi1jMTM4IDc5LjE1OTgyNCwgMjAxNi8wOS8xNC0wMTowOTowMSAgICAgICAgIj4KIDxyZGY6UkRGIHhtbG5zOnJkZj0iaHR0cDovL3d3dy53My5vcmcvMTk5OS8wMi8yMi1yZGYtc3ludGF4LW5zIyI+CiAgPHJkZjpEZXNjcmlwdGlvbiByZGY6YWJvdXQ9IiIvPgogPC9yZGY6UkRGPgo8L3g6eG1wbWV0YT4KPD94cGFja2V0IGVuZD0iciI/PgH//v38+/r5+Pf29fTz8vHw7+7t7Ovq6ejn5uXk4+Lh4N/e3dzb2tnY19bV1NPS0dDPzs3My8rJyMfGxcTDwsHAv769vLu6ubi3trW0s7KxsK+urayrqqmop6alpKOioaCfnp2cm5qZmJeWlZSTkpGQj46NjIuKiYiHhoWEg4KBgH9+fXx7enl4d3Z1dHNycXBvbm1sa2ppaGdmZWRjYmFgX15dXFtaWVhXVlVUU1JRUE9OTUxLSklIR0ZFRENCQUA/Pj08Ozo5ODc2NTQzMjEwLy4tLCsqKSgnJiUkIyIhIB8eHRwbGhkYFxYVFBMSERAPDg0MCwoJCAcGBQQDAgEAADs=')";
 
 function componentWrapperBorder(config: DevSiteDesignSystem): string {
-    return `1px solid ${contrast(1.3, "#FFF", config.backgroundColor)}`;
+    return `1px solid ${config.background300}`;
```
janechu: fallback?

packages/fast-form-generator-react/app/configs/align-vertical/align-vertical.schema.json
```diff
@@ -12,8 +12,7 @@
                 "top",
                 "bottom",
                 "center"
-            ],
-            "disabled": true
```
janechu: why remove this?

packages/fast-form-generator-react/src/utilities/form-input.colors.ts
```diff
@@ -0,0 +1,12 @@
...
+    background100: "#1A1A1A",
+    background200: "#1E1E1E",
+    background300: "#212121",
+    background800: "#303030",
+    foreground000: "#FFF",
+    foreground200: "#F6F6F6",
+    foreground300: "#F2F2F2",
+    foreground800: "#D5D5D5",
+    pink: "#FB356D",
+};
```
janechu: I would prefer that this is in the `form` folder and that this is for color constants so that the name of the file is `form.constants.style.ts` and the internals are:

```
export const background000: string: "#000";
...
```

packages/fast-form-generator-react/src/utilities/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./form-input.colors";
+export * from "./form-input.style";
```
janechu: If you make the above change for the colors file I think you can remove this file and the utilities folder. Ideally form should contain all the necessary items for its rendering

packages/fast-form-generator-react/src/utilities/form-input.style.ts
```diff
@@ -5,25 +5,16 @@ import {
     localizeSpacing,
```
janechu: please move the contents of this file to the `form` folder and rename the file to `form.utilities.style.ts`

packages/fast-form-generator-react/app/configs/align-vertical/align-vertical.schema.json
eljefe223: just for testing I'll put back

packages/fast-form-generator-react/src/utilities/index.ts
eljefe223: By moving these into the form folder its causing an issue that @nicholasrice and I investigated for awhile. I think we should save this for another PR.

packages/fast-form-generator-react/src/utilities/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./form.constants.style";
+export * from "./form.utilities.style";
```
janechu: can you move both of these files into the form folder and remove the utilities folder?

packages/fast-form-generator-react/src/utilities/form.constants.style.ts
```diff
@@ -0,0 +1,10 @@
+export const background000: string = "#000";
+export const background100: string = "#1A1A1A";
+export const background200: string = "#1E1E1E";
+export const background300: string = "#212121";
+export const background800: string = "#303030";
+export const foreground000: string = "#FFF";
+export const foreground200: string = "#F6F6F6";
+export const foreground300: string = "#F2F2F2";
+export const foreground800: string = "#D5D5D5";
+export const pink: string = "#FB356D";
```
janechu: can you rename this to "accent"?

packages/fast-form-generator-react/src/utilities/form.utilities.style.ts
```diff
@@ -0,0 +1,293 @@
...
+export interface BoxShadowConfig {
+    offsetX: number;
+    offsetY: number;
+    blurRadius: number;
+    spreadRadius: number;
+    color: string;
+    inset?: boolean;
+}
+
+export const DISABLED_OPACITY: number = 0.3;
```
janechu: this should be exported from the constants file and should follow the same naming convention used there.

packages/fast-form-generator-react/src/form/form-item.select.tsx
```diff
@@ -70,6 +68,21 @@ class FormItemSelect extends FormItemBase<
         );
     }
 
+    /**
+     * Generates class names
+     */
+    protected generateClassNames(): string {
```
janechu: can you be more specific on this method naming, for example `generateControlSpanClassNames`

packages/fast-form-generator-react/src/form/form-item.children.style.ts
```diff
@@ -24,36 +24,29 @@ const styles: ComponentStyles<FormItemChildrenClassNameContract, {}> = {
...
-        display: "block",
-        marginTop: toPx(12),
     },
     formItemChildren_existingChildren: {
         ...applyCleanListStyle(),
     },
+    formItemChildren_existingChildren__sorting: {
+        fontSize: "11px",
+        color: foreground300,
+        backgroundColor: `${background100} !important`,
```
janechu: you should never have to use !important here, instead inside the `formItemChildren_existingChildren` key you can add:

```
"&.$formItemChildren_existingChildren__sorting" : {
      // your override requiring more specificity
}
```

packages/fast-form-generator-react/src/form/form-item.array.style.ts
```diff
@@ -18,76 +19,77 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
...
-        lineHeight: "32px",
     },
     formItemArray_existingItemList: {
         ...applyCleanListStyle(),
+        fontSize: "11px",
+    },
+    formItemArray_existingItemList__sorting: {
+        fontSize: "11px",
+        color: foreground300,
+        backgroundColor: `${background100} !important`,
```
janechu: you should be able to increase specificity in the `formItemArray_existingItemList`, see my other comment on form-item.children.style.ts

packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -570,7 +583,9 @@ class Site extends React.Component<
         > = {
             pane: {
                 backgroundColor: (config: DevSiteDesignSystem): string => {
-                    return config.lightGray;
+                    return (
+                        config.background300 || devSiteDesignSystemDefaults.background300
+                    );
```
janechu: this shouldnt be in parens

packages/fast-form-generator-react/src/form/form-category.style.ts
```diff
@@ -1,49 +1,77 @@
...
+        borderBottom: `1px solid ${background800}`,
         position: "relative",
-        width: "100%",
+        width: "calc(100% + 10px)",
         fontFamily: "inherit",
-        fontSize: toPx(14),
-        padding: `${toPx(10)} 0`,
+        fontSize: "11px",
+        padding: "0 0 0 10px",
+        marginLeft: "-10px",
```
chrisdholt: negative margin is typically a code smell. I think the issue here is on line 18 where you are setting "width". I think you want,

`calc(100% - 10px)`;

With this, you should also be able to remove the padding here.

packages/fast-form-generator-react/src/form/form-category.style.ts
```diff
@@ -1,49 +1,77 @@
...
+        display: "flex",
+        alignItems: "center",
+        textTransform: "uppercase",
+        minHeight: "30px",
+        margin: "0",
+        fontWeight: "600",
+        boxSizing: "border-box",
+        fontSize: "11px",
+        padding: "0 0 0 10px",
+        marginLeft: "-10px",
```
chrisdholt: I think here we will also end up wanting to set the width to the same that you have for button and remove padding and marginLeft. A div higher up the chain keeps this off the edge of the screen.

packages/fast-form-generator-react/src/form/form-item.array.style.ts
```diff
@@ -18,76 +19,77 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
...
-            opacity: ".6",
-            pointerEvents: "none",
-            top: toPx(11),
-            width: toPx(16),
-            height: toPx(16),
-            background: lines,
-            left: "0",
-        },
+        paddingLeft: "10px",
+        marginLeft: "-10px",
```
chrisdholt: It seems like this repeats quite a bit and seems caused by padding being created several layers outside of the actual form itself. We should either remove the padding outside or fix the instances where we are simultaneously adding adding padding and removing margin on the same side of this.

PR #1284, opened 2019-01-11
fix: ensure onClick and onKeyDown unhandledProps on ContextMenuItem are triggered as expected

### LINKED ISSUE #1260: onClick is getting eaten in context-menu-item
**Description**
When using the ContextMenuItem component if a dev specifies onClick that event handler is never invoked

**Repro steps**
create an instance of ContextMenuItem and attempt to set an onClick handler:
                 
```
<ContextMenuItem
                        className={this.props.managedClasses.contextMenuItem}
                        onClick={(event: React.MouseEvent<HTMLDivElement>): void => {
                            console.log("test");
                        }}
>Test
</ContextMenuItem>
```

In the code below from context-menu-item.tsx the dev's onClick is passed in as unhandledProps and then an onClick is specified on the div effectively eating the dev's onClick.

```
        return (
            <div
                {...this.unhandledProps()}
                role={this.props.role}
                aria-disabled={this.props.disabled || undefined}
                onKeyDown={this.handleMenuItemKeyDown}
                onClick={this.handleMenuItemClick}
            >
                {this.props.children}
            </div>
        );
    }
```

nicholasrice:
# Description

`onClick` and `onKeyDown` unhandled props were being clobbered by props applied internally by the `ContextMenuItem`, which caused these event handlers to never get called. This PR ensures those unhandled prop callbacks work as epxcted

fixes #1260
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
+++ b/packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
@@ -87,11 +87,7 @@ class ContextMenuItem extends Foundation<
      * Handle the keydown event of the item
      */
     private handleMenuItemKeyDown = (e: React.KeyboardEvent<HTMLDivElement>): void => {
-        if (this.props.disabled) {
-            return;
-        }
-
-        if (typeof this.props.onInvoke === "function") {
+        if (typeof this.props.onInvoke === "function" && !this.props.disabled) {
             switch (e.keyCode) {
                 case KeyCodes.enter:
                 case KeyCodes.space:
@@ -99,18 +95,22 @@ class ContextMenuItem extends Foundation<
                     break;
             }
         }
+
+        if (typeof this.props.onKeyDown === "function") {
+            this.props.onKeyDown(e);
+        }
     };
 
     /**
      * Handle the keydown event of the item
      */
     private handleMenuItemClick = (e: React.MouseEvent<HTMLDivElement>): void => {
-        if (this.props.disabled) {
-            return;
+        if (typeof this.props.onInvoke === "function" && !this.props.disabled) {
+            this.props.onInvoke(e, this.props);
         }
 
-        if (typeof this.props.onInvoke === "function") {
-            this.props.onInvoke(e, this.props);
+        if (typeof this.props.onClick === "function") {
+            this.props.onClick(e);
         }
     };
 }
--- a/packages/fast-components-react-base/src/context-menu-item/context-menu-item.spec.tsx
+++ b/packages/fast-components-react-base/src/context-menu-item/context-menu-item.spec.tsx
@@ -110,6 +110,22 @@ describe("context menu item", (): void => {
         expect(onInvoke).toHaveBeenCalledTimes(0);
     });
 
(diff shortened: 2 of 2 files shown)


packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
```diff
@@ -87,30 +87,30 @@ class ContextMenuItem extends Foundation<
...
+        if (typeof this.props.onInvoke === "function" && !this.props.disabled) {
             switch (e.keyCode) {
                 case KeyCodes.enter:
                 case KeyCodes.space:
                     this.props.onInvoke(e, this.props);
                     break;
             }
         }
+
+        if (typeof this.props.onKeyDown === "function") {
```
chrisdholt: Do we want to execute this if the item is disabled?

packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
```diff
@@ -87,30 +87,30 @@ class ContextMenuItem extends Foundation<
...
     private handleMenuItemClick = (e: React.MouseEvent<HTMLDivElement>): void => {
-        if (this.props.disabled) {
-            return;
+        if (typeof this.props.onInvoke === "function" && !this.props.disabled) {
+            this.props.onInvoke(e, this.props);
         }
 
-        if (typeof this.props.onInvoke === "function") {
-            this.props.onInvoke(e, this.props);
+        if (typeof this.props.onClick === "function") {
```
chrisdholt: Do we want to execute this if disabled? I could potentially see a reason, but wonder if it's counterintuitive.

packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
```diff
@@ -87,30 +87,30 @@ class ContextMenuItem extends Foundation<
...
             switch (e.keyCode) {
                 case KeyCodes.enter:
                 case KeyCodes.space:
                     this.props.onInvoke(e, this.props);
                     break;
             }
         }
+
+        if (typeof this.props.onKeyDown === "function") {
+            this.props.onKeyDown(e);
```
chrisdholt: Assuming we are only sending the event because it's implicit and we aren't defining a contract to enumerate what we are sending back?

packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
nicholasrice: yes - this should work as an unhandledProp.

packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
nicholasrice: yes - this should work as an unhandledProp.

packages/fast-components-react-base/src/context-menu-item/context-menu-item.tsx
nicholasrice: The contract is defined in `unhandledProps`

PR #1291, opened 2019-01-12
chore: add prerequisites to install markdown

Falkicon:
# Description

Added npm as prerequisite to install.md

## Motivation & context

We have had feedback from two people now that they missed installing npm first. This should help avoid that misstep. 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -8,10 +8,7 @@ sidebar_label: Install
 
 ## Prerequisites
 
-Before setting up FAST-DNA, install Git and npm:
-
-* [Download Git](https://git-scm.com/downloads)
-* [Download npm](https://www.npmjs.com/get-npm)
+Before setting up FAST-DNA, install [npm](https://www.npmjs.com/get-npm).
 
 ## Setup the source repository
 
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -6,6 +6,13 @@ sidebar_label: Install
 
 # Install
 
+## Prerequisites
+
+Before setting up FAST-DNA, install Git and npm:
+
+* [Download Git](https://git-scm.com/downloads)
+* [Download npm](https://www.npmjs.com/get-npm)
+
 ## Setup the source repository
 
 Clone the repository, `cd` into the project, install [Lerna](https://github.com/lerna/lerna) and install dependencies:
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -8,7 +8,7 @@ sidebar_label: Install
 
 ## Prerequisites
 
-Before setting up FAST-DNA, install [npm](https://www.npmjs.com/get-npm).
+Before setting up FAST-DNA, install Git and [npm](https://www.npmjs.com/get-npm).
 
 ## Setup the source repository
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -4,6 +4,8 @@ title: FAST-DNA
 sidebar_label: What is FAST-DNA
 ---
 
+# FAST-DNA
+
 An unopinionated system of components, development tools, and utilities used à la carte or as a suite to build enterprise-grade websites and applications.
 
 ## Key features & benefits


awentzel:
For consistency, if we provide a link for npm, should we do the same for git?

docs/en/contributing/install.md
```diff
@@ -6,6 +6,10 @@ sidebar_label: Install
 
 # Install
 
+## Prerequisites
+
+Before setting up FAST-DNA, install Git and [npm](https://www.npmjs.com/get-npm).
```
nalogiudice: Should we link to git installation site as well?

chrisdholt: Since this lives on github, I don't think so; we're being fairly generous in calling out that git is a prerequisite here IMO. There are several different ways to install and use git, including using the github app. If someone does not know how to use git or how to install it, we should not point them to a specific way or flavor; it is easy enough to search how to do so and find a solution that works.

I think it makes sense to call out and link to NPM because there is now a competitor in `yarn` and it is important to ensure that you have the correct package manager installed. Additionally, `npm` has a single source of truth which handles all scenarios for ensuring you have it installed and describing what it is.

chrisdholt: cc: @awentzel as I see you asked the same question as well

docs/readme.md
```diff
@@ -4,6 +4,8 @@ title: FAST-DNA
 sidebar_label: What is FAST-DNA
 ---
 
+# FAST-DNA
```
janechu: I'm a little confused as to where this is linked to, currently we seem to be targetting the files in the `en` folder.

docs/en/contributing/install.md
awentzel: I don't think it costs anything to add a link to [here](https://git-scm.com/doc) (I do not agree we should provide link to any specific type of installation or config).  It makes it faster for those that need the link because it's their first time on git and doesn't hurt our content at all since it's only a link and doesn't change the content otherwise.  I am sending MIS/CS students to our repository and 50% of them have not used git and it's useful to them among others.

chrisdholt: I think we need to delineate between contributing to FAST and consuming FAST. Git not required to install our packages, or to use them in a project. NPM is required. 

As an aside, non-collaborators *should not* directly clone our package - they will not be able to push. They should fork and clone their own instance. If this documentation is because we think people are trying to "use" our stuff, our documentation is confusing. I would not recommend us pushing people to the repo in order to "play" with the packages. That's likely better done using something like "create-react-app" and adding the packages as a dependency.

If this is specific to *contributing*, git is required and while I want to encourage anyone to contribute (I also shared this with the students at Code Fellows on Friday), there is a certain expectation that you can clone/fork the repo yourself (via terminal, or the github app, etc) and you have a way to manage that. If you can't there are significant resources online (including in the repo itself via a Github button) to help someone get up and running.

My primary problem with the link you provided is that there is a very clear "downloads" link that exists which drives people to specific flavors and ways to use git.

Either way, I'd be fine pointing to documentation if there were a place that either didn't include specific flavors, or documented/aggregated all the various ways you can use git.

chrisdholt: Possibly, although I think that works in the context of a FULL tutorial. It's specific to what they are trying to accomplish there that they are including that. Here we are directly within the context of contributing to the source repository, and there is an expectation that you are responsible for your development environment.

Falkicon: Leaving out the link for now given that this is the contributors section. If we come up with a decent generic solution we can re-visit but I would agree that the reminder should be good enough for anyone who fits the profile of this section. I asked similar questions last night and I think the big issue is that we don't have a regular getting started section yet so folks are thinking about this like this is getting started, and its not.

docs/readme.md
Falkicon: It is not linked at the moment, but wanted to correct the file anyway so it is valid (I was already there). This file will probably turn into a top level readme for the docs section but that is not part of this PR.

Issue #1269, opened 2019-01-07
Make sure all components check for their base elements


janechu:
All components in `fast-components-react-base` should have test(s) to check for the HTML elements they should be using.

For example, the `Checkbox` component should have a test that checks for an `input` with the attribute `type` set to `checkbox`.

fmamic:
I will take this one (can't assign to myself)

janechu:
Done :)

PR #1296, opened 2019-01-13
fix: Make sure all components check for their base elements

fmamic:
Adding unit tests to check base elements for all components in the fast-components-react-base.

# Description

Adding unit tests to check base elements for all components in the fast-components-react-base.
Every component spec updated to check base element, base element type or base element role attribute.

## Motivation & context

Ensures that the base component element will stay unchanged.

## Issue type checklist

## Process & policy checklist

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://microsoft.github.io/fast-dna/docs/en/contributing/standards) for this project.


packages/fast-components-react-base/src/dialog/dialog.spec.tsx
```diff
@@ -29,6 +29,14 @@ describe("dialog", (): void => {
         expect((Dialog as any).name).toBe(Dialog.displayName);
     });
 
+    test("should have correct element type and correct role attribute value", () => {
```
chrisdholt: Since these are two different DOM nodes (one is the root and one is nested), I would prefer to have these within different test blocks.

packages/fast-components-react-base/src/radio/radio.spec.tsx
```diff
@@ -30,6 +30,12 @@ describe("radio", (): void => {
         expect((Radio as any).name).toBe(Radio.displayName);
     });
 
+    test("should have correct input radio type", () => {
```
chrisdholt: Should we test for the root as well here, which is div?

packages/fast-components-react-base/src/checkbox/checkbox.spec.tsx
```diff
@@ -29,6 +29,12 @@ describe("checkbox", (): void => {
 
     const inputSelector: string = `.${managedClasses.checkbox_input}`;
 
+    test("should have correct element type checkbox", () => {
```
chrisdholt: Should we test for the root as well here, which is div?

chrisdholt: Suggest within a different test block if we do.

packages/fast-components-react-base/src/toggle/toggle.spec.tsx
```diff
@@ -39,6 +39,12 @@ describe("toggle", (): void => {
         expect((Toggle as any).name).toBe(Toggle.displayName);
     });
 
+    test("should have correct input type checkbox", () => {
```
chrisdholt: Should we test for the root in an additional test block as well as the nested element and type?

packages/fast-components-react-base/src/context-menu-item/context-menu-item.spec.tsx
```diff
@@ -19,6 +19,11 @@ describe("context menu item", (): void => {
         expect((ContextMenuItem as any).name).toBe(ContextMenuItem.displayName);
     });
 
+    test("should have root element type div", () => {
```
janechu: I don't actually think we need this test, it's not an accessibility check like the ones we have for "role" and "aria-label".

packages/fast-components-react-base/src/context-menu/context-menu.spec.tsx
```diff
@@ -15,6 +15,12 @@ describe("context menu", (): void => {
         expect((ContextMenu as any).name).toBe(ContextMenu.displayName);
     });
 
+    test("should be type div and contain role menu", () => {
+        const rendered: ShallowWrapper = shallow(<ContextMenu />);
+        expect(rendered.type()).toBe("div");
```
janechu: you can remove the type check and just look for "role"

packages/fast-components-react-base/src/dialog/dialog.spec.tsx
```diff
@@ -29,6 +29,14 @@ describe("dialog", (): void => {
...
+        const rendered: any = shallow(<Dialog managedClasses={managedClasses} />);
+        expect(rendered.type()).toBe("div");
```
janechu: also here, you can remove the type check since this is just a div, in which case it would also solve @chrisdholt concern.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -49,6 +49,11 @@ describe("horizontal overflow", (): void => {
         }).not.toThrow();
     });
 
+    test("should have correct root element type div", () => {
+        const rendered: ShallowWrapper = shallow(<HorizontalOverflow />);
+        expect(rendered.type()).toBe("div");
+    });
```
janechu: can remove this test

packages/fast-components-react-base/src/progress/progress.spec.tsx
```diff
@@ -17,6 +17,12 @@ describe("progress", (): void => {
         expect((Progress as any).name).toBe(Progress.displayName);
     });
 
+    test("should have correct root element type div and correct role progressbar", () => {
+        const rendered: ShallowWrapper = shallow(<Progress />);
+        expect(rendered.type()).toBe("div");
```
janechu: can remove just the type check

packages/fast-components-react-base/src/radio/radio.spec.tsx
janechu: Same comment as before, we should not test for things not related to accessibility. The component will not be any functionally different if we decide one day to make it a span etc.

packages/fast-components-react-base/src/tabs/tab-panel.spec.tsx
```diff
@@ -12,6 +12,12 @@ import TabPanel from "./tab-panel";
 configure({ adapter: new Adapter() });
 
 describe("tab-panel", (): void => {
+    test("should have correct root element type div and correct role tabpanel", () => {
+        const rendered: ShallowWrapper = shallow(<TabPanel />);
+        expect(rendered.type()).toBe("div");
```
janechu: can remove the type check here

packages/fast-components-react-base/src/tabs/tab.spec.tsx
```diff
@@ -11,7 +11,13 @@ import Tab from "./tab";
  */
 configure({ adapter: new Adapter() });
 
-describe("tab-panel", (): void => {
+describe("tab", (): void => {
+    test("should have correct root element type div and correct role tab", () => {
+        const rendered: ShallowWrapper = shallow(<Tab />);
+        expect(rendered.type()).toBe("div");
```
janechu: can remove type check here

packages/fast-components-react-base/src/tabs/tabs.spec.tsx
```diff
@@ -132,6 +132,19 @@ describe("tabs", (): void => {
         </TabItem>,
     ];
 
+    test("should have correct root element type div and correct role tablist", () => {
+        const rendered: ShallowWrapper = shallow(
+            <Tabs managedClasses={tabsManagedClasses} label="label" />
+        );
+        expect(rendered.type()).toBe("div");
```
janechu: can remove type check here

packages/fast-components-react-base/src/checkbox/checkbox.spec.tsx
chrisdholt: To a point that makes sense. However, certain other wrappers around this would break accessibility. `div` => `span` is fine, but what if someone changed it to something else with a semantic role? I would say that *does* impact accessibility. Perhaps we don't want to or feel the need to check for it, but I do think that changing the root element to anything but a non-semantic element would break accessibility.

packages/fast-components-react-base/src/toggle/toggle.spec.tsx
janechu: Again, I would say it is unnecessary.

packages/fast-components-react-base/src/checkbox/checkbox.spec.tsx
chrisdholt: If we do a test for this, let's do it separately because the goal is different. Closing this out

packages/fast-components-react-base/src/dialog/dialog.spec.tsx
chrisdholt: This will solve it

packages/fast-components-react-base/src/radio/radio.spec.tsx
chrisdholt: same comments as before as well :) but resolving to solve further down the road

janechu:
@fmamic I have added it for you in your description "Closes #1269", what this does it automatically closes the associated issue due to the keyword. See https://help.github.com/articles/closing-issues-using-keywords/

packages/fast-components-react-base/src/dialog/dialog.spec.tsx
fmamic: Thanks for comments @chrisdholt @janechu, PR updated.

packages/fast-components-react-base/src/tabs/tab-panel.spec.tsx
```diff
@@ -12,6 +12,11 @@ import TabPanel from "./tab-panel";
 configure({ adapter: new Adapter() });
 
 describe("tab-panel", (): void => {
+    test("should have correct root element type div and correct role tabpanel", () => {
```
janechu: can rename this test to "should have correct role tabpanel"

packages/fast-components-react-base/src/tabs/tab.spec.tsx
```diff
@@ -11,7 +11,12 @@ import Tab from "./tab";
  */
 configure({ adapter: new Adapter() });
 
-describe("tab-panel", (): void => {
+describe("tab", (): void => {
+    test("should have correct root element type div and correct role tab", () => {
```
janechu: can rename test to "should have correct role tab"

PR #1297, opened 2019-01-14
feat: add a plugin system for use in data utilities and subsequently JSON schemas

### LINKED ISSUE #1234: Add the ability to have custom interpretation of schema properties
The current issue: The form generator currently can show React children as default behavior, to add to this there is a common pattern where render props are used instead. We need a way to reinterpret static data into functional data.

There should be a plugin system flexible enough to allow:

1. A custom type to be specified
2. A type to be associated with an existing data type ("string", "boolean", "number", "children" etc.)
3. A plugin which converts it into the necessary data for rendering

Example schema:
```
{
    ...
    "type": "object",
    "customProperties": {
        "example": {
            "type": "react-prop-passing-classname-argument",
        }
    }
}
```

Example data:
```
{
    "custom": {
        "id": "my-callback",
        "props": {
             "text": "Hello world"
        }
    }
}
```
 Example interpretation:

```
const plugins = [
    new MapReactPropToClassName({
        "type": "react-prop-passing-classname-argument"
    });
]

mapDataToComponent(schema, data, childOptions, plugins)
```

In this example the instance of MapReactPropToClassName should have access to:
1. Any additional data that is passed along with that type
2. Any additional data that is included in the schema for that type (this may include any specific restrictions for that plugin to be concerned with as a configuration detail)

janechu:
This will add an optional set of plugins to the `mapDataToComponent` function in the `fast-data-utilities-react` package. The plugins allow a custom manipulation of data when mapped to schemas.

# Description

## Motivation & context

This will solve an issue where the form items being shown must be manipulated before being mapped to a component. An example of this is the react render props. In this way instead of mapping the props directly to a component, a plugin can be used to return a function which correctly maps the data to the component.

Closes #1234 

## Issue type checklist

## Process & policy checklist


packages/fast-data-utilities-react/src/plugin.ts
```diff
@@ -0,0 +1,34 @@
+import { ChildOptionItem } from "./";
+
+export interface PluginProps {
+    /**
```
nicholasrice: Can this be more descriptive?

packages/fast-data-utilities-react/src/plugin.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.config.id = Array.isArray(this.config.id)
+            ? this.config.id
+            : [this.config.id];
+    }
+
+    /**
+     * Determines if there is a match for the IDs set for the plugin
+     * and a provided ID
+     */
+    public resolvesForId(id: string): boolean {
```
nicholasrice: should this method name just be "matches"?

packages/fast-data-utilities-react/src/index.ts
```diff
@@ -278,6 +325,95 @@ function getLocationsFromObject(data: any, location: string = ""): string[] {
     return updatedLocations;
 }
 
+/**
+ * Callback to determine if a string is found within an array of plugin locations
+ */
+function pluginPartialFindIndexCallback(location: string): any {
```
nicholasrice: I think you have enough info to accuratly type the return value of this function

packages/fast-data-utilities-react/src/index.ts
```diff
@@ -373,6 +516,58 @@ export function mapDataToComponent(
...
+    const pluginModifiedDataLocations: PluginLocation[] = getDataLocationsOfPlugins(
+        schema,
+        mappedData,
+        childOptions
+    );
+
+    // remove any children data locations from plugin modified locations
+    reactChildrenDataLocations = reactChildrenDataLocations.map(
+        (reactChildrenDataLocation: string): string => {
+            if (
```
nicholasrice: this doesn't always return a string - is that a problem? If not, can you update the type signature?

PR #1304, opened 2019-01-15
feat: add listbox base component

scomea:
# Description
Listbox will be a building block for the upcoming Select component.  It is quite similar to the Context Menu component with the addition of type ahead functionality and the different roles.  I expect we'll want to further refine how type ahead works, like taking into account current focus in finding the next match, but this implementation seems adequate to get the base component in.

## Motivation & context
We need listbox functionality in our component system, listbox is needed in order to build a select component.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/listbox/README.md
```diff
@@ -0,0 +1,10 @@
+## Listbox
+A generic menu component implementing the [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) role and the keyboard navigation aligning to that role. The *listbox* will accept any child elements but only children with role of `"option"` will receive focus.
+
+### Usage
+To manage document focus the *listbox* assumes that all rendered children that have the role `"option"` and not decorated with "aria-disabled" are focusable html elements. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *listbox*. Additionally, the *listbox* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
```
nalogiudice: nit: prefer `aria-disabled` instead of "aria-disabled"

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+    public componentWillUnmount(): void {
+        clearTimeout(this.typeAheadTimer);
+    }
+
+    /**
+     * Create class names
+     */
+    protected generateClassNames(): string {
+        return super.generateClassNames(get(this.props.managedClasses, "listbox"));
```
nalogiudice: need `,""` after "listbox"

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
```diff
@@ -0,0 +1,219 @@
...
+import { configure, mount, shallow } from "enzyme";
+import Listbox, { ListboxUnhandledProps } from "./listbox";
+import { KeyCodes } from "@microsoft/fast-web-utilities";
+
+/*
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("context menu", (): void => {
```
janechu: context menu?

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
```diff
@@ -0,0 +1,219 @@
...
+        const rendered: any = mount(
+            <Listbox>
+                <div role="option">one</div>
+                <div role="option">two</div>
+            </Listbox>
+        );
+
+        rendered.childAt(0).simulate("keydown", { keyCode: KeyCodes.arrowDown });
+
+        expect(rendered.state("focusIndex")).toBe(1);
```
janechu: I would check before simulation that the focusIndex is 0

packages/fast-components-react-base/src/listbox/listbox.props.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface ListboxHandledProps extends ListboxManagedClasses {
+    /**
+     * The listbox children
+     */
+    children?: React.ReactNode;
+
+    /**
+     * The name of the prop on child components that is used as the data for type ahead focus (default is 'value')
+     */
+    typeAheadPropName?: string;
```
janechu: I would prefer `typeAheadOptionPropertyKey`, does anyone else have any thoughts on this?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+        return (
+            <div
+                {...this.unhandledProps()}
+                ref={this.rootElement}
+                role="listbox"
+                className={this.generateClassNames()}
+                onKeyDown={this.handleMenuKeyDown}
+            >
+                {this.renderChildren()}
+            </div>
```
janechu: what was the reasoning behind this being a `div` and not a `ul`, with every child being an `li` with the `role="option"` applied?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+                return false;
+            }
+        );
+
+        if (matchIndex !== -1) {
+            this.typeAheadTimer = setTimeout((): void => {
+                this.typeAheadTimerExpired();
+            }, 1000);
+            this.setFocus(matchIndex, 1);
```
janechu: can you explain why we need a timer? should we not just check to see if there is a new element to focus on every keystroke?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+            default:
+                if (!e.ctrlKey) {
+                    this.processTypeAhead(e);
+                }
+        }
+    };
+
+    private processTypeAhead = (e: React.KeyboardEvent<HTMLDivElement>): void => {
+        const acceptedChars: string[] = [
```
janechu: why do we restrict to just these characters?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+    ListboxHandledProps,
+    ListboxUnhandledProps,
+    ListboxState
+> {
+    public static displayName: string = "Listbox";
+
+    protected handledProps: HandledProps<ListboxHandledProps> = {
+        children: void 0,
+        managedClasses: void 0,
+        typeAheadPropName: void 0,
```
chrisdholt: `Prop` feels weird within the context of the prop contract; is there a better name that doesn't contain "prop"?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+    private renderChildren(): React.ReactChild[] {
+        this.typeAheadData = [];
+        return React.Children.map(this.props.children, this.renderChild);
+    }
+
+    /**
+     * Render a single child
+     */
+    private renderChild = (
+        child: React.ReactElement<any>,
```
chrisdholt: I wonder if this is better typed as `unknown` instead of `any`

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * Determines if a given element is disabled
+     */
+    private isDisabledElement = (element: Element): element is HTMLElement => {
+        return (
+            element instanceof HTMLElement &&
+            element.getAttribute("aria-disabled") === "true"
+        );
+    };
```
chrisdholt: The two methods above expect/anticipate that:
- something is being passed with a non-native role of option
- something is being passed with a non-natively supported "disabled" attribute.

Are we planning on having a list-box-item with a distinct shape that supports these internal expectations? I don't see it a part of this PR

packages/fast-components-react-base/src/listbox/listbox.tsx
chrisdholt: would it help to check valid characters against a given regex?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,314 @@
...
+     */
+    focusIndex: number;
+}
+
+export interface TypeAheadDataItem {
+    index: number;
+    compareString: string;
+}
+
+class Listbox extends Foundation<
```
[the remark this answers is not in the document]
nalogiudice: +1

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: That appears to be how native select works now.  If I have an option that gets focus when I rapidly type "a-b-c" it does not get focus if I type "a-b" pause for a few secs, then "c".  It looks to me like the keystroke buffer expires after time has passed.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: I will add the activeDescendant bit, seems straightforward.  As for multi-select -  this component does not know about selection at all in the current implementation, it essentially is a focus manager for a menu.  Managing selection state is the job of whatever component owns the listbox, most likely a Select component, and the options themselves.  (And yes, I have multi-select working in my working branch)

chrisdholt: I think the difficult thing here is that in the listbox is an accessible and functional pattern itself; not a partial pattern - and part of the pattern is that it includes the selection as part of its functional spec (https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox). I think it would be odd for us to implement this partially when an expectation of a user would likely be that this supports the functionality of a listbox

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: "Type-ahead is recommended for all listboxes, especially those with more than seven options:
Type a character: focus moves to the next item with a name that starts with the typed character.
Type multiple characters in **rapid succession**: focus moves to the next item with a name that starts with the string of characters typed."  
 https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox

packages/fast-components-react-base/src/listbox/README.md
scomea: Sure, queued for next iteration

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Sure, queued for next iteration

chrisdholt: Why does this require a new PR?

packages/fast-components-react-base/src/listbox/README.md
chrisdholt: Why does this require a new PR?

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
scomea: Copy paste error, fix queued for next iteration

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
scomea: yep, fix queued for next iteration

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Same reasoning as for context-menu.  ie.  ul's only accept li elements as children, we are overriding role and other aria props so it is not necessary for accessiblity.

packages/fast-components-react-base/src/listbox/listbox.tsx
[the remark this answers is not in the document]
scomea: I am totally indifferent so happy to go with consensus.  I note that "listbox" exists as official markup as in:

role="listbox"

So that tells me that our lowercase title should be "listbox" and not "listBox".  given that I thought the capitalized version should be "Listbox" over "ListBox".   That being said I considered it a best guess and am happy to switch.

packages/fast-components-react-base/src/listbox/listbox.props.ts
scomea: Fine with me, queued for next iteration

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Discussed name change, we decided to with Jane's suggestion of "typeAheadPropertyKey".  That and activeDescendant changes queued for next iteration.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: That listbox depends on interactable children to have a role of  "option" seems reasonable to me because the aria spec (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role) says "One or more nested options are required." under related roles.  We use roles similarly to identify valid children in context-menu.

The usage of  "aria-disabled" is exactly what we are doing in context-menu.  Although from reading here I wonder if we should be using just "disabled" instead in both places. (https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme)

The "select-option" component is not in this PR but it would fulfill these requirements.   The expectation would be that other components that fulfill these requirements could be valid items as well.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: I get errors when make that change.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: I'm trying this, seems to work (note that I'm not a regex wiz):
        if  (!/^[a-z0-9\s]+$/i.test(newChar)) {
            return;
        }

packages/fast-components-react-base/src/listbox/README.md
scomea: I meant the next iteration of this PR.

chrisdholt: Thanks for clarifying!

packages/fast-components-react-base/src/listbox/listbox.tsx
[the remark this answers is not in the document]
scomea: So after some office conversation yesterday it seems there is no definitely right way to do this.  Current consensus is to stick with "Listbox" mostly because it is how we do "Checkbox".  Anyone who wants to argue otherwise please pile on.

packages/fast-components-react-base/src/listbox/listbox.tsx
chrisdholt: @scomea I see the select-option specific thing as a likely implementation of the `listbox-item`. With this, if we publish, I only have part of what is required to create a functioning listbox. Additionally, the expectation for a user to be passing a `div` with `aria-disabled` is fine and valid per context menu; however context menu can clearly be constructed using `context-menu-items` where that assumption is true. Without a `listbox-item` I don't know that we can safely assume the above without either documenting it or providing something that meets those requirements.

I think the definition of done of this PR needs to be that a user can adequately create a functioning listbox per the spec so that this can be used in isolation outside of the context of our implementation of MSFT select.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,284 @@
...
+            case KeyCodes.home:
+                e.preventDefault();
+                this.setFocus(0, 1);
+
+                break;
+
+            default:
+                if (!e.ctrlKey) {
+                    this.processTypeAhead(e);
+                }
```
chrisdholt: With multi-select there are quite a few additional expectations for keyboard functionality which are different than the expected interaction model for single select: https://www.w3.org/TR/wai-aria-practices-1.1/#listbox_kbd_interaction

I don't think we need to manage the alternative selection model, but we should support the recommend model.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,284 @@
...
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <div
+                {...this.unhandledProps()}
+                ref={this.rootElement}
+                role="listbox"
+                aria-multiselectable={this.props.multiselectible}
+                aria-activedescendant={this.state.focussedItemId}
```
chrisdholt: On a second look from the spec, it looks like aria-activedescendant is optional. I'm not sure if we should handle it or not...open to thoughts from @nicholasrice who may have some insight on whether we should provide this as opt-in or not at all if we are managing focus.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,284 @@
...
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <div
+                {...this.unhandledProps()}
+                ref={this.rootElement}
+                role="listbox"
+                aria-multiselectable={this.props.multiselectible}
```
chrisdholt: I think we likely only need to pass this value if it's true; I don't think it needs to be enumerated in the DOM if false

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <div
+                {...this.unhandledProps()}
+                ref={this.rootElement}
```
chrisdholt: I think we should consider adding a first-class prop for aria-labelled by. Per the spec:

> If the listbox is not part of another widget, then it has a visible label referenced by aria-labelledby on the element with role listbox.

packages/fast-components-react-base/src/listbox/README.md
```diff
@@ -0,0 +1,10 @@
+## Listbox
+A generic menu component implementing the [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) role and the keyboard navigation aligning to that role. The *listbox* will accept any child elements but only children with role of `"option"` will receive focus.
+
+### Usage
+To manage document focus the *listbox* assumes that all rendered children that have the role `"option"` and not decorated with aria-disabled are focusable html elements. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *listbox*. Additionally, the *listbox* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
+
+The *listbox* "typeAheadPropName" prop indicates which prop the *listbox* examines on children to manage focus when text entry is used to control focus.  The referenced property must be a string.
```
chrisdholt: I think we've changed this prop name for "typeAheadPropName", correct?

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,284 @@
...
+        managedClasses: void 0,
+        typeAheadPropertyKey: void 0,
+    };
+
+    private rootElement: React.RefObject<HTMLDivElement> = React.createRef<
+        HTMLDivElement
+    >();
+
+    private typeAheadString: string = "";
+    private typeAheadTimer: any;
```
chrisdholt: Is this actually "any".

packages/fast-components-react-base/src/listbox/README.md
```diff
@@ -0,0 +1,10 @@
+## Listbox
+A generic menu component implementing the [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) role and the keyboard navigation aligning to that role. The *listbox* will accept any child elements but only children with role of `"option"` will receive focus.
+
+### Usage
+To manage document focus the *listbox* assumes that all rendered children that have the role `"option"` and not decorated with aria-disabled are focusable html elements. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *listbox*. Additionally, the *listbox* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
```
chrisdholt: > Additionally, the *listbox* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.

We are setting tabIndex and managing focus; is this actually accurate?

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Selection isn't managed by this component, but rather the parent component (likely a select, but we should leave the door open to others).  I think type ahead essentially works the same in both cases.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: It seems straightforward and I think I have it working right.  Not sure why we'd rip it out.

packages/fast-components-react-base/src/listbox/listbox.tsx
chrisdholt: Listbox is valid without a parent - how do I implement a listbox and manage selection? I think we need to document this. It seems like we've built this primarily to support our scenario of select

packages/fast-components-react-base/src/listbox/README.md
scomea: Same wording and underlying functionality as context menu.

chrisdholt: OK - Is it accurate? If it is, that's fine; it seems incorrect as we are handling it. If this is incorrect, should we still commit this?

packages/fast-components-react-base/src/listbox/README.md
scomea: Yes we did.  Corrected.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Added that.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: NodeJS.Timer seems to work, changed.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Agreed, same with labelled by.  Done

packages/fast-components-react-base/src/listbox/listbox.props.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * The name of the prop on child components that is used as the data for type ahead focus
+     */
+    typeAheadPropertyKey?: string;
+
+    /**
+     * Whether this listbox supports multi-selection (default is 'false')
+     */
+    multiselectible?: boolean;
```
janechu: selectable*

I think though you could just remove this all-together as unhandledProps will allow this to fall through

packages/fast-components-react-base/src/listbox/listbox.props.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * The aria-labelledby attribute to link the listbox to an existing
+     * element that provides it an accessible name
+     */
+    labelledBy?: string;
```
janechu: Nowhere in here is a callback for when something has actually been selected.

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
```diff
@@ -0,0 +1,246 @@
...
+            <Listbox>
+                <div>not a focusable element</div>
+                <div role="option">focusable element</div>
+            </Listbox>
+        );
+
+        expect(rendered.state("focusIndex")).toBe(1);
+    });
+
+    test("should move focus down when the down arrow is pressed", (): void => {
```
janechu: please add a test for a single child, a cursory view looks like because you are using the index you assume children are in an array, however a single react child is usually an object.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,286 @@
...
+import { KeyCodes } from "@microsoft/fast-web-utilities";
+import { get, inRange, invert } from "lodash-es";
+import { canUseDOM } from "exenv-es6";
+
+export interface ListboxState {
+    /**
+     * The index of the focusable child
+     */
+    focusIndex: number;
+    focussedItemId: string;
```
janechu: Our standard for things like this is `active`, the idea of focus is an implementation detail. I would recommend that this is `activeIndex`, `activeItemId`

packages/fast-components-react-base/src/listbox/listbox.tsx
janechu: Makes sense.

packages/fast-components-react-base/src/listbox/listbox.tsx
janechu: Yes, but you can use the li yourself and wrap the given children, you do not need to rely on the user to pass children that are li's

packages/fast-components-react-base/src/listbox/listbox.tsx
janechu: But why would there ever be a case where we would not want for example, "@"?

packages/fast-components-react-base/src/listbox/listbox.props.ts
chrisdholt: Eh - I'm not sure we want to leave this to unhandled. There is what seems to be significantly different keyboard interaction/behavior with multi-select. If we support it (I think we should at base), I think we should handle it; otherwise how do they manage their focus for just that scenario?

scomea: Fixed the typo.  But I agree with Chris.  I realized I need to refactor to move selection logic that I hoped could live in the parent "Select" control into the listbox itself so this prop will do more than just flag whether we need to aria label the listbox as multselect.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: This is exactly what we do in context menu.  I don't have strong feelings either way myself tho.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: What benefit does this give?  Should we not support any arbitrary non-focusable elements as children?

packages/fast-components-react-base/src/listbox/listbox.props.ts
scomea: Revisit after I've refactored selection logic out of the parent Select into the Listbox itself.

scomea: Refactored listbox to include selection

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Pulled out validation in the current iteration so we process all keys.  Chris and I talked about maybe adding a prop where a developper could provide a validation function.  Thoughts?

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Listbox-item added and Listbox has been refactored to handle selection.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Listbox has been refactored to manage selection and supports the recommended kbd interaction you pointed to above.

packages/fast-components-react-base/src/listbox/listbox.props.ts
scomea: Multi-selectable does much more in "listbox manages selection" refactor.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,605 @@
...
+                    child.props[Listbox.idPropertyKey] === undefined ||
+                    child.props[Listbox.idPropertyKey] !== itemId
+                ) {
+                    return false;
+                }
+                return true;
+            }
+        );
+
+        return matchNode;
```
janechu: same comment as above, you can just do:

```
return React.Children.toArray(children).find(
    // etc
};
```

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,605 @@
...
+
+        children.some(
+            (child: React.ReactElement<any>, index: number): boolean => {
+                if (child.props[this.props.typeAheadPropertyKey] === undefined) {
+                    return false;
+                }
+                if (
+                    child.props[this.props.typeAheadPropertyKey]
+                        .toLowerCase()
+                        .startsWith(this.typeAheadString)
```
janechu: startsWith does not work with IE, any version https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -0,0 +1,172 @@
...
+            return;
+        }
+
+        const itemData: ListboxItemData = {
+            id: this.props.id,
+            value: this.props.value,
+            displayString: this.props.displayString,
+        };
+
+        if (typeof (this.context as ListboxContextType).itemFocused === "function") {
```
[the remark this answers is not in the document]
scomea: If what you are saying is that different context types can collide then we definitely should prefix with component name.  How about:
    listboxItemInvoked
    listboxItemInvoked
    listboxSelectedItems
    listboxMultiselectable

packages/fast-components-react-base/src/listbox/README.md
```diff
@@ -0,0 +1,21 @@
+## Listbox
+A generic listbox component implementing the [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) role and the keyboard navigation aligning to that role.
+
+### Usage
+The *listbox* will accept any child elements but the only children that should receive focus are either *listbox-item* or custom components that have the role of '"option"' and appropriately call the '"itemInvoked"' and '"itemFocused"' functions on the ListboxContext provided by the parent *listbox* component.  It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *listbox*. Additionally, the *listbox* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
```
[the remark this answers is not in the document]
scomea: Hmmm....I thought I was copying an existing style but apparently I was confused.  Fixing

packages/fast-components-react-base/src/listbox/listbox.schema.json
```diff
@@ -0,0 +1,30 @@
...
+            "type": "string"
+        },
+        "labelledBy": {
+            "title": "Labelled by",
+            "type": "string"
+        },
+        "multiselectable": {
+            "title": "Multi-selectable",
+            "type": "boolean"
+        }
```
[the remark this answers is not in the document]
scomea: fixed

packages/fast-components-react-base/src/listbox-item/README.md
```diff
@@ -0,0 +1,8 @@
+## Listbox item
+A component designed to work as a child of the *listbox*.
+
+### Usage
+The component required '"id"' prop must be unique and is applied as an id to to the resulting element in the dom.  Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the '"onSelectedItemsChanged"' callback on the parent listbox.
```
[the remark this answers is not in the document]
scomea: fixed, not really sure where I got that....

packages/fast-components-react-base/src/listbox-item/listbox-item.props.ts
```diff
@@ -0,0 +1,46 @@
...
+     */
+    children?: React.ReactNode;
+
+    /**
+     * If the item is disabled
+     */
+    disabled?: boolean;
+
+    /**
+     * the value of the item
```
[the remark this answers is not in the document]
scomea: Fixed

packages/fast-components-react-base/src/listbox-item/listbox-item.props.ts
```diff
@@ -0,0 +1,46 @@
...
+     */
+    value: string;
+
+    /**
+     * The unique id for the item
+     */
+    id: string;
+
+    /**
+     * Friendly string that may be used in ui display
```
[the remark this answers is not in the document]
scomea: fixed

packages/fast-components-react-base/src/listbox-item/listbox-item.props.ts
```diff
@@ -0,0 +1,46 @@
...
+     */
+    displayString?: string;
```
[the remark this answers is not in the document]
scomea: Yes, displayString is different from children.  It is the human friendly version of value and gets passed around with value should parents (ie. the select box) want to show selection in string format.

packages/fast-components-react-base/src/listbox-item/listbox-item.schema.json
```diff
@@ -0,0 +1,35 @@
...
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "displayString": {
+            "title": "Display string",
+            "type": "string"
+        },
+        "id": {
+            "title": "Id",
```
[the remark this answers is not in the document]
scomea: fixed

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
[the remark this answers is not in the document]
scomea: If what you are saying is that different context types can collide then we definitely should prefix with component name. How about:
listboxItemInvoked
listboxItemInvoked
listboxSelectedItems
listboxMultiselectable

packages/fast-components-react-base/src/listbox/README.md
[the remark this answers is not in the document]
scomea: Fixed, not sure where I got that...

packages/fast-components-react-base/src/listbox-item/README.md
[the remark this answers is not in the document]
scomea: fixed, not sure where I got that

packages/fast-components-react-base/src/listbox-item/listbox-item.props.ts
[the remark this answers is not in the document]
scomea: Fixed

packages/fast-components-react-base/src/listbox-item/listbox-item.schema.json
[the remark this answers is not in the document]
scomea: fixed

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,605 @@
...
+    private static asItemData(items: string[]): ListboxItemData[] {
+        const itemsToReturn: ListboxItemData[] = items.map((item: string) => {
+            const newItem: ListboxItemData = {
+                id: item,
+                displayString: "",
+                value: "",
+            };
+            return newItem;
+        });
+        return itemsToReturn;
```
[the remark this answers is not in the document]
scomea: fixed

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,605 @@
...
+                    child.props[Listbox.idPropertyKey] === undefined ||
+                    child.props[Listbox.idPropertyKey] !== itemId
+                ) {
+                    return false;
+                }
+                return true;
+            }
+        );
+
+        return matchNode;
```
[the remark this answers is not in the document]
scomea: I skipped declaring itemdata, but I do need matchnode as I copy values from it.

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: yep

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Presumably the one in @microsoft/fast-web-utilities is good
                 startsWith(
                        child.props[this.props.typeAheadPropertyKey].toLowerCase(),
                        this.typeAheadString
                    )

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
scomea: Added and passed without code changes.  I think that because I always run children through one of the React node functions (map, filter or toArray) as a first action and I expect they handle the single child case appropriately.

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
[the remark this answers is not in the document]
janechu: Since it is a method I think we normally prefix it with "on" on this case, because you would be calling it correct? That's why my suggestion would be `onListboxItemFocus`, `onListboxItemInvoke`, same with the others if they are methods

packages/fast-components-react-base/src/listbox/README.md
```diff
@@ -0,0 +1,21 @@
+## Listbox
+A generic listbox component implementing the [listbox](https://www.w3.org/TR/wai-aria-1.1/#listbox) role and the keyboard navigation aligning to that role.
+
+### Usage
+The *listbox* will accept any child elements but the only children that should receive focus are either *listbox-item* or custom components that have the role of "option" and appropriately call the "itemInvoked"' and "itemFocused" functions on the ListboxContext provided by the parent *listbox* component.  It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *listbox*. Additionally, the *listbox* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
```
janechu: looks like you have an extra ' in itemInvoked

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -0,0 +1,600 @@
...
+        if (event.defaultPrevented) {
+            return;
+        }
+
+        let focusItemId: string;
+
+        switch (event.keyCode) {
+            case KeyCodes.arrowDown:
+            case KeyCodes.arrowRight:
+                event.preventDefault();
```
janechu: nit: you keep calling event.preventDefault() in every switch case, you can just put this before the switch.

packages/fast-components-react-base/src/listbox/examples.data.tsx
```diff
@@ -0,0 +1,82 @@
...
+            },
+            {
+                id: "listbox-item",
+                props: {
+                    ...listboxItemPropFactory("c"),
+                },
+            },
+        ],
+    },
+    data: [
```
janechu: I'd like to see an example where you are using something other than "displayString" as a typeahead property so we can validate that this works

packages/fast-components-react-base/src/listbox/README.md
scomea: fixed

packages/fast-components-react-base/src/listbox/examples.data.tsx
scomea: Added such a test with "value" instead

packages/fast-components-react-base/src/listbox/README.md
scomea: I see. What we are actually saying is that the component manages the tabindex of children so children should not interfere with that. Yes?

scomea: "Additionally, the *listbox* component sets the `tabIndex` property of its children as it manages focus, child components should not set tabindex independently."?

packages/fast-components-react-base/src/listbox-item/listbox-item.schema.json
```diff
@@ -0,0 +1,35 @@
...
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "displayString": {
+            "title": "Display string",
+            "type": "string"
+        },
+        "id": {
+            "title": "ID",
```
chrisdholt: We have a few different ways we enumerate "ID" as a title - @nalogiudice do you recall what our decided format was here?

packages/fast-components-react-base/src/listbox-item/listbox-item.props.ts
```diff
@@ -0,0 +1,46 @@
+import * as React from "react";
+import {
+    ListboxItemClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface ListboxItemManagedClasses
+    extends ManagedClasses<ListboxItemClassNameContract> {}
+export interface ListboxItemUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
```
chrisdholt: @nicholasrice to confirm, but I think because we are enumerating "id", we'll want to subtract that value from the HTMLAttribute interface we are extending here. We can use utility-types (existing dependency) to do this

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -0,0 +1,174 @@
...
+    public render(): React.ReactElement<HTMLDivElement> {
+        let isSelected: boolean = false;
+        if ((this.context as ListboxContextType).listboxSelectedItems !== undefined) {
+            isSelected =
+                (this.context as ListboxContextType).listboxSelectedItems.filter(
+                    (item: ListboxItemData) => {
+                        return item.id === this.props.id;
+                    }
+                ).length === 1;
+        }
```
chrisdholt: nit/suggestion - it may be easier to grok this if we extract this to a function and just assign it to the aria-selected value as that is the only place this gets used. As it stands now, I expected the above selection logic to be used in multiple places.

packages/fast-components-react-base/src/listbox/listbox-context.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+
+export interface ListboxItemData {
+    id: string;
+    value: string;
+    displayString: string;
+}
```
chrisdholt: Curious on the reason to create a new interface here for this data rather than expecting the prop contract itself. At a glance it seems like passing ListboxItemProps would give us a more extensible solution long-term, but I may be missing an obvious reason for a different interface

packages/fast-components-react-base/src/listbox/listbox.schema.json
```diff
@@ -0,0 +1,38 @@
...
+            "type":"string"
+        }
+    },
+    "reactProperties": {
+        "children": {
+            "title": "Building blocks",
+            "type": "children",
+            "defaults": [
+                "text"
+            ]
```
chrisdholt: I think here we want our defaults to be `ListboxItem` correct?

packages/fast-components-react-base/src/listbox-item/listbox-item.props.ts
nicholasrice: I think this is okay since it is the same type but we should validate

packages/fast-components-react-base/src/listbox/listbox-context.tsx
chrisdholt: Synced with @nicholasrice on his thoughts here and I don't think we'll run into perf issues by passing the props.

packages/fast-components-react-base/src/listbox-item/listbox-item.schema.json
nalogiudice: HTML id attribute

packages/fast-components-react-base/src/listbox/listbox.schema.json
scomea: Yep

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
scomea: Yep, cleaner

packages/fast-components-react-base/src/listbox-item/listbox-item.schema.json
scomea: "Unique ID" seems most common, although there are other variants.

scomea: (Gonna go with that, if we want to standardize we should open a bug for someone to do a consistency pass)

chrisdholt: I'm fine with this ^^^. I checked three different components and somehow the schemas were all different (although I think we did align these at some point?)

packages/fast-components-react-base/src/listbox/listbox-context.tsx
scomea: Pushed an update that uses props.  Seems to work as expected.

Issue #1312, opened 2019-01-17
add npm package documentation to existing documentation


awentzel:
This has been discussed with leadership and approved as an interim step moving towards our ideal solution.  The ideal solution will take much longer and will be described in future proposal task.  This gets us by for the foreseeable future.

**What feature are you requesting?**
We want to combine package documentation into the Docusaurus documentation.

**What would be the benefit of adding this feature?**
A single place where all documentation exists guide + api + marketing site

**What solution would be ideal for you?**
Described below

**What alternatives have you considered?**
Catalog, Gatsby, GitBook, Custom, etc

**Is there any additional context for your request?**
Yes. See below

**Attach screenshots or images add detail to your request:**
Included below in the requirements

**Requirements**
Write a package script that will execute a new process to copy each package readme.md into `./docs/en/packages/[package-name]/readme.md` where Docusaurus expects markdown content files.

This image shows how things come together in Docusaurus land.
![image](https://user-images.githubusercontent.com/16669785/51350832-0f73b500-1a5e-11e9-8974-c69d5d58d40a.png)

This should all be scripted with exception to the Docusaurus sidebars.json file.

1. Script should be located as `./build/documentation/copy-package-readme.js`
2. Script will copy every package root readme. To `./docs/en/packages/[package-name]/`
	a. Conditionally create folder if nonexistent matching the package name.
	b. Copy and replace if already existing from the `./packages/**/readme.md` to `./docs/en/packages/[package-name]/`
3. Add the Docusaurus header as shown in the image lines 1-4. This will need to be created in code. Id will always be index.  Title should be the package name as indicated by the package folder name.  First word "FAST" all caps.  Other words should be readable without hyphens.  Sidebar_label: Should be the name without hyphens and without FAST.
![image](https://user-images.githubusercontent.com/16669785/51350860-25817580-1a5e-11e9-8ae1-2952ea7d5b71.png)
4. Update Docusaurus navigation file located in /website/sidebars.json
![image](https://user-images.githubusercontent.com/16669785/51350874-2b775680-1a5e-11e9-8e18-2ce21ecd062a.png)
5. Compile Docusaurus and ensure that these files are compiled along with our other documentation markdown files along with the appropriate navigation being rendered to the page. `npm run build` to compile any structural changes.  `npm run start` to run the website.

PR #1315, opened 2019-01-17
chore: add script for copying package readme files to docs/en/packages

jkmdev:
# Description

Adds a script to `build/documentation` that will copy all package `readme.md` files to `docs/en/packages`.
Run `node build/documentation/copy-package-readme.js` to see its results.

## Motivation & context

Incremental addition to #1312.
Implements points 1 and 2.

## Issue type checklist

## Process & policy checklist


awentzel:
I tried running this on my system that has two folders already existing and it rewrote those which is great (though I did lost the header of the files in my case since I had templated out steps you have not yet gotten too.  But, the script didn't create any of the folders or files for all the other packages?  

Is there special documentation notes we should add?  How did you test?  Should we add a parameter for --dry-run to pretest since this will ultimately be ran as part of a another script, for example, building the Docusaurus site. 

```bash
Error: ENOENT: no such file or directory, mkdir 'D:\projects\fast-dna\docs/en/packages\fast-breakpoint-tracker-react\README.md'
```

jkmdev:
Oh boy. :/
It's because I wrote this in Linux, my use of forward-slashes in some places created an error or Windows.

I made the fix and tested it in Windows, it should work now after running `node build/documentation/copy-package-readme.js`.

To test I've just been running  `node build/documentation/copy-package-readme.js` and checking if it has the desired output.

`--dry-run`? So for this I'd print to the console what the script is doing without creating folders or files? Seems simple, I can add it if you want. I'm not 100% sure about its utility, but it seems like a good idea since it's a process that will overwrite all related files.

Falkicon:
@jkmdev, this was something I really wanted to see done for the docs but we had not gotten to it yet. Thank you for the contribution!

jkmdev:
@Falkicon Happy to help!
@awentzel Alright, will do.

build/documentation/copy-package-readme.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+function copyReadmeFiles() {
+
+    const resolvedSrcReadmePaths = path.resolve(rootDir, srcReadmePaths);
+
+    glob(resolvedSrcReadmePaths, {realpath:true}, function(error, files) {
+        
+        files.forEach((filePath) => {
+            const destReadmePath = filePath.replace(/(\bpackages\b)(?!.*\1)/, destDir);
+            const dirPath = destReadmePath.replace(/(\b\\README.md\b)(?!.*\1)/, '');
```
nicholasrice: Instead of replacing `"README.md"` I would instead utilize Node's API - `const dirPath = path.dirname(destReadmePath)` - I've found using the supported API's for path manipulation really helps minimize platform bugs.

build/documentation/copy-package-readme.js
```diff
@@ -0,0 +1,70 @@
...
+ * Utility for copying readme to .docs/en/packages/[package-name]/readme.md.
+ * Usage: node build/documentation/copy-package-readme.js
+ */
+const path = require("path");
+const fs = require("fs");
+const glob = require("glob");
+
+const rootDir = path.resolve(process.cwd());
+const srcReadmePaths = "packages/*/README.md";
+const destDir = "docs\\en\\packages";
```
nicholasrice: this `destDir` is now causing the script to throw in my environment.

I would recommend using `path.join("docs", "en", "packages")`. Alternatively,  you can construct the path using `path.sep` to easily allow Node to pick the path separator based on runtime environment.

jkmdev: Ahhh this feedback was very helpful, tyty

nicholasrice: Of course! Thanks for making the update :)

PR #1319, opened 2019-01-18
feat: increase density of development site

eljefe223:
# Description
Now that the density has increased in the fast form package we want to align the development site.

## Motivation & context
increasing the density better matches fast from and removes design inconsistencies. 

image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-development-site-react/src/components/shell/index.tsx
+++ b/packages/fast-development-site-react/src/components/shell/index.tsx
@@ -22,7 +22,7 @@ class Shell extends React.Component<{}, {}> {
     > = {
         container: {
             fontFamily: "Segoe UI, SegoeUI, Helvetica Neue, Helvetica, Arial, sans-serif",
-            fontSize: toPx(14),
+            fontSize: "11px",
             height: "100vh",
         },
     };
--- a/packages/fast-development-site-react/src/components/site/index.tsx
+++ b/packages/fast-development-site-react/src/components/site/index.tsx
@@ -144,13 +144,13 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
     "@global": {
         "body, html": {
             fontFamily: "Segoe UI, SegoeUI, Helvetica Neue, Helvetica, Arial, sans-serif",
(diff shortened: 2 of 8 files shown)


packages/fast-development-site-react/src/components/site/index.tsx
```diff
@@ -162,48 +162,38 @@ const styles: ComponentStyles<SiteManagedClasses, DevSiteDesignSystem> = {
...
             return config.foreground300 || devSiteDesignSystemDefaults.foreground300;
         },
+        fontFamily: "inherit",
+        fontSize: "inherit",
         borderRadius: toPx(2),
         appearance: "none",
-        padding: localizeSpacing(Direction.ltr)(
-            `${toPx(8)} ${toPx(36)} ${toPx(8)} ${toPx(10)}`
-        ),
+        padding: localizeSpacing(Direction.ltr)("2px 36px 3px 10px"),
```
nalogiudice: I noticed in some places we are localizing and in others we are not...

packages/fast-development-site-react/src/components/shell/index.tsx
```diff
@@ -22,7 +22,7 @@ class Shell extends React.Component<{}, {}> {
     > = {
         container: {
             fontFamily: "Segoe UI, SegoeUI, Helvetica Neue, Helvetica, Arial, sans-serif",
-            fontSize: toPx(14),
+            fontSize: "11px",
```
awentzel: Why not use `toPx(11)` ?  Is there a difference?

eljefe223: ToPX is more expensive. It’s runs a functions that converts a number to a string and appends “pix” to the end. It should only be used when you want to calculate some value. I think people just assumed they should use it because it was used everywhere. So really we should be ripping it out when we can.

packages/fast-development-site-react/src/components/site/index.tsx
eljefe223: If there is an issue we can address it in another PR. As long as none of them were removed.

PR #1323, opened 2019-01-21
chore: add plugins and update example data in test applications

janechu:
# Description

Updates the test sites in the component libraries to use plugins.

## Motivation & context

This uses the new plugin option for the `fast-data-utilities-react` `mapDataToComponent` and resolves the React render props in `fast-components-react-base` and `fast-components-react-msft`.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/breadcrumb/examples.data.tsx
```diff
@@ -19,10 +19,6 @@ const label: any = {
     label: "breadcrumb",
 };
 
-function renderSeparator(): (className?: string) => React.ReactNode {
-    return (className?: string): React.ReactNode => <div className={className}>\</div>;
```
nicholasrice: Maybe just change the return value of this function instead of duplicating the object?

packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
```diff
@@ -0,0 +1,30 @@
...
+    public resolver(data: any, childOption?: ChildOptionItem): any {
+        return (
+            className: string,
+            disabled?: boolean,
+            appearance?: string
+        ): React.ReactNode => {
+            const callbackArguments: any = {
+                className,
+            };
+            callbackArguments.disabled = disabled !== undefined ? disabled : undefined;
```
nicholasrice: If you're expecting a `boolean` should you just check for a boolean instead of undefined?

packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
```diff
@@ -0,0 +1,30 @@
...
+            callbackArguments.appearance =
+                appearance !== undefined ? appearance : undefined;
```
nicholasrice: Same here - if `appearance` needs to be a string, why not check for `typeof string`?

PR #1330, opened 2019-01-22
feat: add direction and localization helpers to fast-web-utilities

chrisdholt:
## Issue type checklist

## Process & policy checklist


packages/fast-web-utilities/src/localization.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Expose ltr and rtl strings
+ */
+export enum Direction {
+    ltr = "ltr",
+    rtl = "rtl",
+}
+
+export const localeDirectionMapping: { [key: string]: Direction } = {
+    en: Direction.ltr,
+    "en-rtl": Direction.rtl,
```
nicholasrice: so `en-rtl` isn't really a locale - do you know what we're using this for?

chrisdholt: @nicholasrice I don't - this was part of Jane's work. We can fix/update this. Do you have any suggestions for an alternative? I think it's that you could have en as ltr and rtl?

janechu: It was patched in for the development site to allow for checking rtl since there is no translation story, this should be only in the development site and not exposed here.

chrisdholt: Thanks @janechu! I'll remove here.

chrisdholt: I've removed this here; I had to force push due to rebase considerations.

The larger change to scope this to the dev site should align with the deprecation of the `fast-application-utilities` as that package is the one still being used.

packages/fast-web-utilities/src/localization.ts
```diff
@@ -0,0 +1,20 @@
...
+};
+
+export default function isRTL(locale: string): boolean {
```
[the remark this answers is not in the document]
chrisdholt: I've confirmed this and noted it will be resolved with the full deprecation of the application utils package.

packages/fast-web-utilities/src/localization.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Expose ltr and rtl strings
+ */
+export enum Direction {
+    ltr = "ltr",
+    rtl = "rtl",
+}
+
+export const localeDirectionMapping: { [key: string]: Direction } = {
```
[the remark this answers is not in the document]
chrisdholt: I've confirmed this and noted it will be resolved with the full deprecation of the application utils package.

PR #1331, opened 2019-01-22
feat: deprecate fast-application-utilities package

chrisdholt:
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-application-utilities/README.md
+++ b/packages/fast-application-utilities/README.md
@@ -1,5 +1,5 @@
 # FAST application utilities
-**FAST application utilities has been deprecated. Use `@microsoft/fast-web-utilities` instead.**
+**FAST application utilities has been deprecated. Use [FAST web utilities](https://www.npmjs.com/package/@microsoft/fast-web-utilities) instead.**
 
 A list of utilities used by FAST applications.
 
--- a/packages/fast-application-utilities/README.md
+++ b/packages/fast-application-utilities/README.md
@@ -1,4 +1,5 @@
 # FAST application utilities
+**FAST application utilities has been deprecated. Use `@microsoft/fast-web-utilities` instead.**
 
 A list of utilities used by FAST applications.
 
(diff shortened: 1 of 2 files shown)


packages/fast-application-utilities/src/localization.ts
```diff
@@ -1,16 +1,23 @@
 /**
  * Expose ltr and rtl strings
+ * @deprecated - use @microsoft/fast-web-utilties instead
  */
 export enum Direction {
     ltr = "ltr",
     rtl = "rtl",
 }
 
+/**
```
nicholasrice: Are we migrating these in a separate PR?

chrisdholt: Yeah - I have a seperate PR to add to fast-web-utilities.

janechu: I feel like this is related and could be in one PR, otherwise we have a moment in time where we can't publish, plus it would bump both of them as it's a feature for both packages.

janechu: @chrisdholt then in that case I would say this PR is premature, the feature in the other package that allows for the deprecation needs to go in before this PR can be considered.

chrisdholt: @janechu agree - that's why the PR was pushed prior and is up for review prior to this. Feel free to hold until that goes in if you like.

packages/fast-application-utilities/README.md
```diff
@@ -1,4 +1,5 @@
 # FAST application utilities
+**FAST application utilities has been deprecated. Use `@microsoft/fast-web-utilities` instead.**
```
awentzel: Could we also include a link to the recommended package readme.md?

chrisdholt: @awentzel how about the npm package? That makes more sense to me.

Issue #1128, opened 2018-11-09
Create "carousel" as a new MSFT component


chrisdholt:
Carousel allows the user to page through content blocks one at a time, watch an animated slide show sequence, or tab through a sequence of smaller content blocks. From a structural standpoint, this control is a tabpanel which includes flippers for moving between each panel (for sighted users only).

Examples of use in the MSFT design language:
<img width="654" alt="carousel_example_1" src="https://user-images.githubusercontent.com/13071055/48280027-3e8afc00-e407-11e8-83c1-c821ae7dc5e5.png">
<img width="1674" alt="carousel_example_2" src="https://user-images.githubusercontent.com/13071055/48280052-4ba7eb00-e407-11e8-8f44-7625efe86c16.png">

Outstanding question:
- Should we support features such as "autoplay" to automatically rotate through the carousel items. This has significant accessibility considerations. If we *don't* handle this ourselves, we should consider how we can easily enable it to be hooked up by users.

chrisdholt:
Additional bit of info - I would expect that the actual carousel would simply grow based on the content passed within it. IE, depending on the content passed, it would size appropriately.

PR #1338, opened 2019-01-24
feat: add carousel as a MSFT component

nalogiudice:
Closes [#1128](https://github.com/Microsoft/fast-dna/issues/1128)

** This is an initial PR - many changes will come with the review. Please pull down. A style guide for the sequence indicators does not exist as of yet.

# Description

Adds the basic MSFT carousel:

- uses `<Tabs />`
- cycles to the appropriate tab on arrowing left or right
- cycles to the appropriate tab by clicking a sequence indicator
- cycles to next or previous slide on `<Flipper /> click
- no autoplay
- flippers are always visible (but not to screen readers)
- no transitions or animation

screen shot 2019-01-23 at 7 28 13 pm
screen shot 2019-01-23 at 7 26 56 pm

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
@@ -61,10 +61,10 @@ export interface CarouselClassNameContract {
     /**
      * The light slide theme
      */
-    theme__light?: string;
+    carousel__themeLight?: string;
 
     /**
      * The dark slide theme
      */
-    theme__dark?: string;
+    carousel__themeDark?: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
@@ -57,9 +57,4 @@ export interface CarouselClassNameContract {
      * The dark slide theme
      */
     theme__dark?: string;
-
-    /**
-     * Autoplay toggle
-     */
-    carousel_autoplay?: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
@@ -1,0 +1,65 @@
+/**
+ * The class name contract for the carousel component
+ */
+
+export interface CarouselClassNameContract {
+    /**
+     * The root of the carousel component
+     */
+    carousel?: string;
+
+    /**
+     * The carousel tabs
+     */
+    carousel_tabs?: string;
+
+    /**
+     * The carousel tab items
+     */
+    carousel_tabsTabItem?: string;
+
+    /**
+     * The carousel tab item tab
+     */
+    carousel_tabsTabItemTab?: string;
+
+    /**
+     * The carousel tab item tab panel
+     */
+    carousel_tabsTabItemTabPanel?: string;
+
+    /**
+     * The carousel tab item tab content
+     */
+    tabsTabItemTabContent?: string;
+
+    /**
+     * The carousel tab item tab panel content
+     */
+    carousel_tabsTabItemTabPanelContent?: string;
+
+    /**
+     * The carousels previous flipper
+     */
+    carousel_flipperPrevious?: string;
+
+    /**
+     * The carousels next flipper
+     */
+    carousel_flipperNext?: string;
+
+    /**
+     * The light slide theme
+     */
+    theme__light?: string;
+
+    /**
+     * The dark slide theme
+     */
+    theme__dark?: string;
+
+    /**
+     * Autoplay toggle
+     */
+    carousel_autoplay?: string;
+}
--- a/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
@@ -9,42 +9,52 @@ export interface CarouselClassNameContract {
     carousel?: string;
 
     /**
-     * The carousel tabs
+     * The collection of tabs and tab panels
      */
-    carousel_tabs?: string;
+    carousel_slides?: string;
 
     /**
-     * The carousel tab items
+     * The collection of sequence indicators
      */
-    carousel_tabsTabItem?: string;
+    carousel_sequenceIndicators?: string;
 
     /**
-     * The carousel tab item tab
+     * The sequence indicator
      */
-    carousel_tabsTabItemTab?: string;
+    carousel_sequenceIndicator?: string;
 
     /**
-     * The carousel tab item tab panel
+     * The active carousel sequence indicator
      */
-    carousel_tabsTabItemTabPanel?: string;
+    carousel_sequenceIndicator__active?: string;
 
     /**
-     * The carousel tab item tab content
+     * The collection of tab panels
      */
-    tabsTabItemTabContent?: string;
+    carousel_tabPanels?: string;
 
     /**
-     * The carousel tab item tab panel content
+     * The tab panel
      */
-    carousel_tabsTabItemTabPanelContent?: string;
+    carousel_tabPanel?: string;
 
     /**
-     * The carousels previous flipper
+     * The hidden tab panel
+     */
+    carousel_tabPanel__hidden?: string;
+
+    /**
+     * The tab panel content
+     */
+    carousel_tabPanelContent?: string;
+
+    /**
+     * The carousel's previous flipper
      */
     carousel_flipperPrevious?: string;
 
     /**
-     * The carousels next flipper
+     * The carousel's next flipper
      */
     carousel_flipperNext?: string;
 
--- a/packages/fast-components-react-msft/src/carousel/carousel.schema.json
+++ b/packages/fast-components-react-msft/src/carousel/carousel.schema.json
@@ -21,11 +21,11 @@
                 "type": "object",
                 "properties": {
                     "id": {
-                        "title": "ID",
+                        "title": "HTML id attribute",
                         "type": "string"
                     },
                     "theme": {
-                        "title": "Slide theme",
+                        "title": "Theme",
                         "type": "string",
                         "enum": [
                             "dark",
@@ -34,7 +34,7 @@
                     },
                     "content": {
                         "title": "Slide content",
-                        "type": "string",
+                        "type": "object",
                         "pluginId": "@microsoft/fast-components-react-msft/carousel/content"
                     }
                 }
--- a/packages/fast-components-react-msft/src/carousel/carousel.schema.json
+++ b/packages/fast-components-react-msft/src/carousel/carousel.schema.json
@@ -31,10 +31,12 @@
                             "dark",
                             "light"
                         ]
-                    },
+                    }
+                },
+                "reactProperties": {
                     "content": {
                         "title": "Slide content",
-                        "type": "object",
+                        "type": "children",
                         "pluginId": "@microsoft/fast-components-react-msft/carousel/content"
                     }
                 }
--- a/packages/fast-components-react-msft/src/carousel/carousel.schema.json
+++ b/packages/fast-components-react-msft/src/carousel/carousel.schema.json
@@ -33,14 +33,20 @@
                         ]
                     }
                 },
+                "required": [
+                    "id"
+                ],
                 "reactProperties": {
                     "content": {
                         "title": "Slide content",
                         "type": "children",
-                        "pluginId": "@microsoft/fast-components-react-msft/carousel/content"
+                        "pluginId": "@microsoft/fast-components-react-msft/carousel/items/content"
                     }
                 }
             }
         }
-    }
+    },
+    "required": [
+        "content"
+    ]
 }
--- a/packages/fast-components-react-msft/src/carousel/carousel.schema.json
+++ b/packages/fast-components-react-msft/src/carousel/carousel.schema.json
@@ -1,0 +1,42 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Carousel",
+    "description": "A carousel component's schema definition.",
+    "type": "object",
+    "id": "carousel",
+    "properties": {
+        "label": {
+            "title": "Carousel accessible label",
+            "type": "string"
+        },
+        "activeId": {
+            "title": "Set active slide by HTML id attribute",
+            "type": "string"
+        },
+        "items": {
+            "title": "Slides",
+            "type": "array",
+            "items": {
+                "title": "Slide",
+                "type": "object",
+                "properties": {
+                    "tabItemSlot": {
+                        "title": "Tab item slot",
+                        "type": "string",
+                        "default": "tab-item"
+                    },
+                    "tabPanelSlot": {
+                        "title": "Tab panel slot",
+                        "type": "string",
+                        "default": "tab-panel"
+                    },
+                    "tabSlot": {
+                        "title": "Tab slot",
+                        "type": "string",
+                        "default": "tab"
+                    }
+                }
+            }
+        }
+    }
+}
--- a/packages/fast-components-react-msft/src/carousel/carousel.schema.json
+++ b/packages/fast-components-react-msft/src/carousel/carousel.schema.json
@@ -20,20 +20,22 @@
                 "title": "Slide",
                 "type": "object",
                 "properties": {
-                    "tabItemSlot": {
-                        "title": "Tab item slot",
-                        "type": "string",
-                        "default": "tab-item"
+                    "id": {
+                        "title": "ID",
+                        "type": "string"
                     },
-                    "tabPanelSlot": {
-                        "title": "Tab panel slot",
+                    "theme": {
+                        "title": "Slide theme",
                         "type": "string",
-                        "default": "tab-panel"
+                        "enum": [
+                            "dark",
+                            "light"
+                        ]
                     },
-                    "tabSlot": {
-                        "title": "Tab slot",
+                    "content": {
+                        "title": "Slide content",
                         "type": "string",
-                        "default": "tab"
+                        "pluginId": "@microsoft/fast-components-react-msft/carousel/content"
                     }
                 }
             }
--- a/packages/fast-components-react-msft/src/carousel/carousel.tsx
+++ b/packages/fast-components-react-msft/src/carousel/carousel.tsx
@@ -29,6 +29,25 @@ class Carousel extends Foundation<
 > {
     public static displayName: string = "Carousel";
 
+    /**
+     * React life-cycle method
+     */
+    public static getDerivedStateFromProps(
+        nextProps: CarouselProps,
+        prevState: CarouselState
+    ): null | Partial<CarouselState> {
+        if (
+            typeof nextProps.activeId === "string" &&
+            nextProps.activeId !== prevState.activeId
+        ) {
+            return {
+                activeId: nextProps.activeId,
+            };
+        }
+
+        return null;
+    }
+
     protected handledProps: HandledProps<CarouselHandledProps> = {
         managedClasses: void 0,
         label: void 0,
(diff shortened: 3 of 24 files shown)


packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * The carousel tab item tab panel
+     */
+    carousel_tabsTabItemTabPanel?: string;
+
+    /**
+     * The carousel tab item tab content
+     */
+    tabsTabItemTabContent?: string;
```
eljefe223: this is missing carousel_ prefix

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Renders the tab item
+     */
+    private renderItems = (): JSX.Element[] => {
+        if (!this.slides || this.slides.length <= 0) {
+            return;
+        }
+
+        return this.slides.map(
```
eljefe223: I could be wrong here but i don't think you need to map and return a `<TabItem/>` we are already doing that in the tabs base component. Really all you need to do is style it to layout the way you want. So you could just map the classNames to your carousel classNames like:

```
private tabsManagedClasses(): TabsClassNameContract {
        return {
            tabs: this.props.managedClasses.carousel_tabs,
            tabs_tabPanels: this.props.managedClasses.carousel_tabs_tabPanels,
            tabs_tabList: this.props.managedClasses.carousel_tabs_tabList,
        };
    }
```

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -0,0 +1,150 @@
...
+import {
+    applyMixedColor,
+    ensureNormalContrast,
+    hoverContrast,
+    normalContrast,
+} from "../utilities/colors";
+import Chroma from "chroma-js";
+import { CarouselClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+
+function flipperStyles(): CSSRules<{}> {
```
scomea: Isn't this effectively hard coding this styling?  I'm thinking the vertical positioning of the flippers might be something that designers might want to modify based on slide content.

nalogiudice: Yeah, but I believe this is the correct way we want devs to change styles. They can simply override with `jssStyleSheet`. We want the MSFT shape/style to cover the widest ground, and I think flippers in the middle covers this. Happy to make design updates either way.

packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * The carousel tab item tab panel content
+     */
+    carousel_tabsTabItemTabPanelContent?: string;
+
+    /**
+     * The carousels previous flipper
```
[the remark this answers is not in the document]
awentzel: I'm not positive, so had to look a little deeper, @scomea, are you talking about line 42 the comment plural?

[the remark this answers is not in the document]
scomea: Possesive, not plural -  as in "The carousel's previous flipper"?

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Yep yep. After the fix went in yesterday to Tabs, this is working right here.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+            ? this.slides[this.getActiveIndex()].theme
+            : SlideTheme.dark;
+    }
+
+    /**
+     * Updates slides
+     */
+    private updateSlides(): void {
+        if (this.props.items.length > 0) {
+            this.slides = this.props.items.map((item: Slide): Slide => item);
```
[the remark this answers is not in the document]
nalogiudice: Good call - I will update

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <div {...this.unhandledProps()} className={this.generateClassNames()}>
+                <Flipper
+                    direction={FlipperDirection.previous}
+                    onClick={this.handlePreviousInvoke}
+                    className={get(
+                        this.props,
+                        "managedClasses.carousel_flipperPrevious",
+                        ""
+                    )}
+                />
```
chrisdholt: I pulled this down and am noticing that these flippers always exist and they are also always visible. As a sighted keyboard user, I am confused as to why I can't access the flippers. I suggest that we consider defaulting these to only show on hover. The question becomes, "what happens on mobile?" We can consider adding swipe out of the box, or we could also consider exposing a prop to have the flippers always be visible.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+                <Tabs
+                    label={this.props.label}
+                    activeId={this.state.activeId}
+                    onUpdate={this.handleUpdate}
+                    items={this.slides}
+                    managedClasses={this.generateTabsClassNames()}
+                />
+                <Flipper
+                    direction={FlipperDirection.next}
+                    onClick={this.handleNextInvoke}
```
chrisdholt: Regarding these flipper click behaviors; since these flippers are never exposed to screen readers, they would never be invoked on anything but a click. Since this is primarily click related, I would suggest naming as such.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+     */
+    private getActiveIndex(): number {
+        return this.slides.map((slide: Slide) => slide.id).indexOf(this.state.activeId);
+    }
+
+    /**
+     * Get the current slide theme (needed for UI overlay contrast)
+     */
+    private getSlideTheme(): string {
+        if (!this.slides || this.slides.length <= 0) {
```
chrisdholt: do we know that slides is an array here?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+        return this.slides[this.getActiveIndex()].theme
+            ? this.slides[this.getActiveIndex()].theme
+            : SlideTheme.dark;
+    }
+
+    /**
+     * Updates slides
+     */
+    private updateSlides(): void {
+        if (this.props.items.length > 0) {
```
chrisdholt: do we know that this is for sure an array?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+            this.updateSlides();
+        }
+    }
+
+    /**
+     * React life-cycle method
+     */
+    public componentDidUpdate(): void {
+        if (this.slides !== this.props.items) {
+            this.updateSlides();
```
chrisdholt: Curious what happens to local state of the active ID in the case that a new item(s) is pushed in?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+    public static displayName: string = "Carousel";
+
+    protected handledProps: HandledProps<CarouselHandledProps> = {
+        managedClasses: void 0,
+        label: void 0,
+        activeId: void 0,
+        items: void 0,
+    };
+
+    private slides: Slide[];
```
chrisdholt: Can you elaborate on the decision to store the slides on the `this` object of the class?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    /**
+     * Define constructor
+     */
+    constructor(props: CarouselProps) {
+        super(props);
+
+        if (this.props.items) {
```
chrisdholt: What happens if this is just one item? Perhaps we can perform that check here instead of in the state object:

```
if (Array.isArray(this.props.items) && this.props.items.length > 0) {}
```

chrisdholt:
How difficult will it be to:

- extend this to include autoplay (can it be done in a non-breaking way?)
- add animations (if this is just a matter of having those I think we can get them ASAP)

Regarding the design side of things, I notice that the carousel takes up 100% width of its parent element; however, this means that if the items inside are *smaller* the flippers and the carousel region extend beyond the items

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: If you review the notes, I left details on these decisions. Since there is no design or spec as of yet, I started with the most simple version of the carousel. So I can obviously change the functionality of how the flippers act, but I wanted to open up the conversation.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,227 @@
...
+            this.state = {
+                activeId:
+                    typeof this.props.activeId === "string"
+                        ? this.props.activeId
+                        : this.props.items.length > 0
+                            ? get(this.props.items[0], "id")
+                            : "",
```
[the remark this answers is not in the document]
nalogiudice: I can change, but to be clear this is how it was done in Tabs, and it is basically the same check.

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: Sure - I wanted to start that conversation here; I likely could have led with that caveat. 

In the scenario of a tabpanel, I think flippers function as an ease-of-use for mouse users. Keyboard users should likely never be exposed to these as they act as duplicate navigation. With that in mind, we should consider only showing them on hover, which is a typical solution on hover. 

Open to suggestions on how we handle mobile scenarios for this.

nalogiudice:
@chrisdholt I spoke with @janechu regarding autoplay. She agreed that it would be good to do in another PR in the future. The issue is that `tabs` has to be refactored to allow a before and/or after slot (and possibly a div wrapper). Right now we would not be able to place the autoplay inline (as it was in older designs).
Animations should be relatively easy once I get the information on what we want.
I will look into the sizing issue.

chrisdholt:
> @chrisdholt I spoke with @janechu regarding autoplay. She agreed that it would be good to do in another PR in the future. The issue is that `tabs` has to be refactored to allow a before and/or after slot (and possibly a div wrapper). Right now we would not be able to place the autoplay inline (as it was in older designs).
> Animations should be relatively easy once I get the information on what we want.
> I will look into the sizing issue.

Thanks @nalogiudice - no request here to do in this PR; you answered my question in terms of ease of addition in the future.

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Sure, that would result in this:
```
if (Array.isArray(this.props.items) && this.props.items.length > 0) {
            this.state = {
                activeId:
                    typeof this.props.activeId === "string"
                        ? this.props.activeId
                        : get(this.props.items[0], "id")
                    };
        } else {
            this.state = {
                activeId: ""
        }
```

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: May want to lean on design for some input on this... would they even be open to flippers on mobile? If not, then swipe is probably our only option.

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Sure

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: @chrisdholt  I actually think it was correct before, if you follow the logic again. First we want to check to make sure items exists at all. Then we want to set the id if it is passed OR get the id from the first slide OR set it to empty (because there is no first slide)
Changing the initial if parameters to check for length as well convolutes this logic and would actually cause this if/else block to be longer (even longer than I put above - it's missing another block). I think the correct solution would be this:
```
if (Array.isArray(this.props.items)) {
            this.state = {
                activeId:
                    typeof this.props.activeId === "string"
                        ? this.props.activeId
                        : this.props.items.length > 0
                            ? get(this.props.items[0], "id")
                            : "",
            };

            this.updateSlides();
        }
```
@janechu can verify the logic from `tabs`

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Having a slides object allows several things.
- It enforces an array of type Slide
- It means we don't have to check for an array or type each time we manipulate the slides

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: See slide comment

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: See slide comment

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Sure

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Changed to visible on hover for now. Not sure how we want to handle mobile-side. May want weigh-in from design.

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: State of active ID does not change

chrisdholt: @nalogiudice what happens if I pass a different active Id? The internal state is one thing and my prop reference is different. I think that may be a gap here.

packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * The light slide theme
+     */
+    theme__light?: string;
+
+    /**
+     * The dark slide theme
+     */
+    theme__dark?: string;
```
chrisdholt: Should these theme values be namespaced with carousel to follow our conventions?

packages/fast-components-react-msft/src/carousel/carousel.tsx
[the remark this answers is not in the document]
chrisdholt: @nalogiudice could we leverage more of `get` from lodash here to trim this down perhaps? I think below might remove the need for chaining here.

```
typeof this.props.activeId === "string
    ? this.props.activeId
    : get(this.props.items[0], "id", "")
```

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: I think we might be able to simplify; noted below.

packages/fast-components-react-msft/src/carousel/carousel.tsx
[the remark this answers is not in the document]
nalogiudice: Good call

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+            this.updateSlides();
+        }
+    }
+
+    /**
+     * React life-cycle method
+     */
+    public componentDidUpdate(): void {
+        if (this.slides !== this.props.items) {
```
nicholasrice: The way you have this set up, `this.slides` will never be strictly equal to `this.props.items` because `slice` returns a *new* array. I don't really think this is necessary though, if you make the changes I've discussed in the `updateSlides` method

packages/fast-components-react-msft/src/carousel/carousel.tsx
nicholasrice: Left a few comments related to this below.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,242 @@
...
+        return this.slides[this.getActiveIndex()].theme
+            ? this.slides[this.getActiveIndex()].theme
+            : SlideTheme.dark;
+    }
+
+    /**
+     * Updates slides
+     */
+    private updateSlides(): void {
+        this.slides = this.props.items.slice();
```
nicholasrice: This will throw if `this.props.items` is not an array. I think you should remove this and change your `slides` prop to the following:

```
private get slides(): Slide[] {
    return Array.isArray(this.props.items) ? this.props.items : [];
}
```

That will allow you to safely access slides without checking to ensure it's an array everywhere you need to access slides, and also means you don't have to keep a private member in sync with props.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -0,0 +1,242 @@
...
+     */
+    private getActiveIndex(): number {
+        return this.slides.map((slide: Slide) => slide.id).indexOf(this.state.activeId);
+    }
+
+    /**
+     * Get the current slide theme (needed for UI overlay contrast)
+     */
+    private getSlideTheme(): string {
+        if (!this.slides || this.slides.length <= 0) {
```
nicholasrice: If you make my proposed change, you can change this to just check length because it will *always* be an array

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Awesome. Makes sense - updated.

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -0,0 +1,144 @@
...
+            top: "calc(50% - 20px)",
+            zIndex: "100",
+            display: "none",
+        };
+    }
+
+    return {
+        carousel: {
+            position: "relative",
+            width: "fit-content",
```
chrisdholt: @nalogiudice do we want to use `fit-content` here with the significant lack of support? If this is to support growing to fit the size of the contents, I think we can get the same scenario with a display property of `inline-block`

nalogiudice: Once I fix the rendering issue that @janechu is helping me with, I will go back to this and come up with a better solution

packages/fast-components-react-msft/src/carousel/carousel.schema.json
```diff
@@ -0,0 +1,44 @@
...
+                        "type": "string",
+                        "enum": [
+                            "dark",
+                            "light"
+                        ]
+                    },
+                    "content": {
+                        "title": "Slide content",
+                        "type": "object",
+                        "pluginId": "@microsoft/fast-components-react-msft/carousel/content"
```
janechu: if this is meant to be a react child, it should not be in properties but in `reactProperties`

packages/fast-components-styles-msft/src/carousel/index.ts
nalogiudice: Fixed

PR #1351, opened 2019-01-30
feat: update styling to match the FAST form generator

janechu:
# Description

Updates to the CSS editor components to match the new `fast-form-generator-react` package styling.

## Motivation & context

The tooling must match, therefore this package has been updated to reflect the new styling direction.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-css-editor-react/src/position/position.style.ts
+++ b/packages/fast-css-editor-react/src/position/position.style.ts
@@ -1,89 +1,107 @@
 import { ComponentStyles } from "@microsoft/fast-jss-manager";
-import { toPx } from "@microsoft/fast-jss-utilities";
+import { background800, foreground800 } from "../editor.constants.style";
+import {
+    applyInputStyle,
+    applySelectInputStyles,
+    applySelectSpanStyles,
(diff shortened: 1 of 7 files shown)


packages/fast-css-editor-react/src/position/position.style.ts
```diff
@@ -1,89 +1,107 @@
...
+    absoluteInput_centerRow__activeLeft?: string;
+    absoluteInput_centerRow__activeRight?: string;
+    absoluteInput_centerRowContent?: string;
 }
 
 const styles: ComponentStyles<CSSPositionClassNameContract, {}> = {
-    cssPosition: {},
-    cssPosition_selectContainer: {
+    cssPosition: {
+        fontSize: "11px",
```
chrisdholt: I wonder if this actually should be 10 or 12? Since this is matching the form generator, this is fine here; more curious from @Falkicon if 11px is accurate as the value or if it should be 10 or 12? 

@janechu no need to make the change here, question is more from a design standpoint so I can file a bug.

Issue #1365, opened 2019-02-08
Create new base and MSFT number field component


nalogiudice:
**What feature are you requesting?**
Create a number field component.

**What would be the benefit of adding this feature?**
Number field is a common web component.

Number field should be first built as a base component and then styled as an MSFT component.

PR #1371, opened 2019-02-12
feat: create number field as base and MSFT component

nalogiudice:
Closes [#1365](https://github.com/Microsoft/fast-dna/issues/1365)
# Description
Basic `<input type="number" />` component.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/number-field/number-field.tsx
```diff
@@ -0,0 +1,103 @@
...
+            attributes.readOnly = true;
+        }
+
+        if (typeof this.props.required === "boolean") {
+            attributes.required = true;
+        }
+
+        if (typeof this.props.step === "number") {
+            attributes.step = this.props.step;
+        }
```
janechu: I question the need to enumerate these, they should already be specified by the unhandledProps, since this should be typed as an HTMLInputElement instead of HTMLDivElement.

packages/fast-components-react-base/src/number-field/number-field.schema.json
```diff
@@ -0,0 +1,41 @@
...
+            "type": "boolean"
+        },
+        "readOnly": {
+            "title": "HTML readOnly attribute",
+            "type": "boolean"
+        },
+        "required": {
+            "title": "HTML required attribute",
+            "type": "boolean"
+        }
```
janechu: please also include "value"

packages/fast-components-react-base/src/number-field/number-field.tsx
```diff
@@ -0,0 +1,103 @@
...
+            attributes.disabled = true;
+        }
+
+        if (typeof this.props.max === "number") {
+            attributes.max = this.props.max;
+        }
+
+        if (typeof this.props.min === "number") {
+            attributes.min = this.props.min;
+        }
```
eljefe223: Should this be able to handle negative values? It currently doesn't. I can manually type negative values in the input but cannot in the Configuration panel

packages/fast-components-styles-msft/src/utilities/fields.ts
```diff
@@ -0,0 +1,41 @@
...
+            cursor: "not-allowed",
+            "&::placeholder": {
+                ...typographyPattern.disabled,
+            },
+        },
+        "&::placeholder": {
+            color: foregroundNormal,
+        },
+    };
+}
```
chrisdholt: Please break this refactor out into a different PR. Changing text field and text area don't align with this feature's intent.

packages/fast-components-react-base/src/number-field/number-field.tsx
nalogiudice: It should... I will look into it.

packages/fast-components-react-base/src/number-field/number-field.tsx
nalogiudice: `HTMLDivElement` to `HTMLInputElement` sure - I missed that.

> I question the need to enumerate **these**

What are these specifically @janechu ? My thought was to expose the common attributes users would expect with a number field. Or else you get a configuration pane without much configuration.

packages/fast-components-styles-msft/src/utilities/fields.ts
nalogiudice: Code climate would have complained if I had repeated the same pattern for a third time. I suppose I could pull this out into a small PR to get in first

packages/fast-components-react-base/src/number-field/number-field.schema.json
nalogiudice: Sure

packages/fast-components-styles-msft/src/utilities/fields.ts
nalogiudice: @chrisdholt pulled out into another [PR](https://github.com/Microsoft/fast-dna/pull/1375)

chrisdholt: Thanks - Unrelated to code climate, this is just all around different work than the purpose of this PR. I'll take a look over there.

packages/fast-components-react-base/src/number-field/number-field.tsx
nalogiudice: Everything seems to be working as expected. I can set a negative min in the configurator and it functions as expected. Can you look into your issue again?

packages/fast-components-react-base/src/number-field/number-field.tsx
janechu: All of the props, the fact that this is typed as an HTMLInputElement means this should not need you to enumerate disabled, step, min, max etc.

janechu: I think it's fine to leave them in the JSON schema because we want to be able to configure those values in the form generator, but you shouldn't need these in the handledProps

packages/fast-components-react-base/src/number-field/number-field.tsx
eljefe223: I still can't set a negative value in the configuration panel... Its probably not a blocker or even an issue with your code. Probably an issue in fast form

nalogiudice: What browser are you using? I am able to set negative values in the form generator no problem:

screen shot 2019-02-18 at 3 48 43 pm

eljefe223: So weird... I'm using Chrome

nalogiudice: Anyone else seeing this?

packages/fast-components-react-msft/src/number-field/examples.data.tsx
```diff
@@ -0,0 +1,44 @@
...
+        },
+        {
+            required: true,
+        },
+        {
+            placeholder: "Placeholder",
+        },
+        {
+            name: "Name",
+        },
```
chrisdholt: can we add an example which includes `value`?

packages/fast-components-react-base/src/number-field/number-field.tsx
```diff
@@ -0,0 +1,64 @@
...
+    protected generateClassNames(): string {
+        return super.generateClassNames(get(this.props, "managedClasses.numberField"));
+    }
+
+    /**
+     * Generates attributes based on props
+     */
+    private assignAttributes(): Partial<NumberFieldUnhandledProps> {
+        if (typeof this.props.value === "number") {
+            return { defaultValue: `${this.props.value}` };
```
janechu: I would not re-assign value to defaultValue, these are two separate ideas, one which results in a controlled component and one which results in uncontrolled, see: https://reactjs.org/docs/uncontrolled-components.html#default-values

I don't think it's necessary to specify either of these in the component or interface, only in the JSON schema.

packages/fast-components-react-base/src/number-field/number-field.tsx
```diff
@@ -0,0 +1,64 @@
...
+    NumberFieldClassNameContract,
+} from "@microsoft/fast-components-class-name-contracts-base";
+import { TextFieldType } from "../text-field/index";
+
+class NumberField extends Foundation<
+    NumberFieldHandledProps,
+    NumberFieldUnhandledProps,
+    {}
+> {
+    public static defaultProps: Partial<NumberFieldProps> = {};
```
chrisdholt: If we aren't using this we should be able to remove this.

packages/fast-components-react-base/src/number-field/number-field.props.ts
```diff
@@ -0,0 +1,18 @@
+import * as React from "react";
+import {
+    ManagedClasses,
+    NumberFieldClassNameContract,
+} from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface NumberFieldManagedClasses
+    extends ManagedClasses<NumberFieldClassNameContract> {}
+export interface NumberFieldUnhandledProps
+    extends React.HTMLAttributes<HTMLInputElement> {}
```
chrisdholt: @nalogiudice we should remove "type" here using utility-types

packages/fast-components-react-base/src/number-field/number-field.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    public static displayName: string = "NumberField";
+
+    protected handledProps: HandledProps<NumberFieldHandledProps> = {
+        managedClasses: void 0,
+        value: void 0,
```
chrisdholt: @janechu and @nalogiudice can we jump on discord and sync up sometime today about the handled prop contract here?

packages/fast-components-react-base/src/number-field/number-field.tsx
nalogiudice: I believe I have resolved this.

packages/fast-components-react-base/src/number-field/number-field.props.ts
nalogiudice: Yep good catch

packages/fast-components-react-base/src/number-field/number-field.tsx
nalogiudice: I think it will only be `managedClasses`. @janechu has specified all props can be passed via unhandled props. If we need a chat that's fine, or you may just be able to talk to Jane.

packages/fast-components-react-msft/src/number-field/examples.data.tsx
nalogiudice: Added

packages/fast-components-react-base/src/number-field/number-field.tsx
nalogiudice: Going to resolve for now, but if someone else runs into this issue, I will reopen.

PR #1381, opened 2019-02-14
feat: add additional functionality to listbox base

scomea:
# Description
- add and implement autofocus and disabled props to listbox, these were found to needed to implement Select component
- exposed some utility static functions.  These are used by the Select component when it wants to manipulate selected items when the listbox is not actually instanciated.

## Motivation & context
Towards a working select component

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/listbox/listbox.props.ts
+++ b/packages/fast-components-react-base/src/listbox/listbox.props.ts
@@ -43,6 +43,16 @@ export interface ListboxHandledProps extends ListboxManagedClasses {
      * The  onSelectedItemsChanged event handler
      */
     onSelectedItemsChanged?: (selectedItems: ListboxItemProps[]) => void;
+
+    /**
+     * Whether a listitem should automatically get focus when this component is mounted
+     */
+    autoFocus?: boolean;
+
+    /**
+     * Specifies that the control is disabled
+     */
+    disabled?: boolean;
 }
 
 export type ListboxProps = ListboxHandledProps & ListboxUnhandledProps;
--- a/packages/fast-components-react-base/src/listbox/listbox.props.ts
+++ b/packages/fast-components-react-base/src/listbox/listbox.props.ts
@@ -47,7 +47,7 @@ export interface ListboxHandledProps extends ListboxManagedClasses {
     /**
      * Whether a listitem should automatically get focus when this component is mounted
      */
-    autoFocus?: boolean;
+    focusItemOnMount?: boolean;
 
     /**
      * Specifies that the control is disabled
--- a/packages/fast-components-react-base/src/listbox/listbox.spec.tsx
+++ b/packages/fast-components-react-base/src/listbox/listbox.spec.tsx
@@ -48,15 +48,39 @@ describe("listbox", (): void => {
         expect(rendered.state("focusIndex")).toBe(-1);
     });
 
-    test("should set focusIndex to the first focusable element on mount", (): void => {
+    test("should set focusIndex to the first focusable element on mount but not set focus to it", (): void => {
         const rendered: any = mount(
             <Listbox>
                 <div>not a focusable element</div>
                 {itemA}
-            </Listbox>
+            </Listbox>,
+            { attachTo: document.body }
         );
 
         expect(rendered.state("focusIndex")).toBe(1);
+        expect(document.activeElement.id).toBe("");
+    });
+
+    test("should set focus to listitem when autofocus is set to true", (): void => {
+        const rendered: any = mount(
+            <Listbox autoFocus={true}>
+                <div>not a focusable element</div>
+                {itemA}
+            </Listbox>,
+            { attachTo: document.body }
+        );
+
+        expect(rendered.state("focusIndex")).toBe(1);
+        expect(document.activeElement.id).toBe("a");
+    });
+
+    test("should not set focus to listitem when disabled", (): void => {
+        const rendered: any = mount(<Listbox disabled={true}>{itemA}</Listbox>, {
+            attachTo: document.body,
+        });
+
+        expect(rendered.state("focusIndex")).toBe(-1);
+        expect(document.activeElement.id).toBe("");
     });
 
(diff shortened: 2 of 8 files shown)


packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -211,10 +265,21 @@ class Listbox extends Foundation<
...
 
+        if (this.state.selectedItems.length > 0) {
+            focusIndex = Listbox.getItemIndexById(
+                this.state.selectedItems[0].id,
+                this.props.children
+            );
+        } else {
+            const children: Element[] = this.domChildren();
+            focusIndex = children.findIndex(this.isFocusableElement);
+        }
```
janechu: nit: we usually use a ternary here and since there is no need to store children you could instead do:

```
const focusIndex: number = this.state.selectedItems.length > 0
    ? this.getItemIndexById(
           this.state.selectedItems[0].id,
           this.props.children
       );
    : this.domChildren().findIndex(this.isFocusableElement);
```

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -229,7 +294,16 @@ class Listbox extends Foundation<
      * Create class names
      */
     protected generateClassNames(): string {
-        return super.generateClassNames(get(this.props.managedClasses, "listbox", ""));
+        let className: string = get(this.props.managedClasses, "listbox", "");
+
+        if (this.props.disabled) {
+            className = className.concat(
+                " ",
+                get(this.props.managedClasses, "listbox__disabled")
```
janechu: please check that this managedClass exists before attempting to concat.

scomea: Doing our more standard approach instead: 
       if (this.props.disabled) {
            className = `${className} ${get(
                this.props,
                "managedClasses.listbox__disabled",
                ""
            )}`;
        }

packages/fast-components-react-base/src/listbox/listbox.tsx
scomea: Done

packages/fast-components-react-base/src/listbox/listbox.spec.tsx
```diff
@@ -360,8 +384,8 @@ describe("listbox", (): void => {
             .childAt(0)
             .simulate("keydown", { keyCode: KeyCodes.arrowUp, shiftKey: true });
         expect(rendered.state("selectedItems").length).toBe(2);
-        expect(rendered.state("selectedItems")[0].id).toBe("a");
-        expect(rendered.state("selectedItems")[1].id).toBe("b");
+        expect(rendered.state("selectedItems")[0].id).toBe("b");
+        expect(rendered.state("selectedItems")[1].id).toBe("a");
```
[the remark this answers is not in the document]
scomea: This is because I changed the toggleItem function to append newly selected items to the list of selected items rather than prepend it as it had done previously.  Seemed like a more expected behavior.  So this test now shows items added to the list in the order they were added (vs the inverse).  Do we need to be more deliberate about how the list of selected items is sorted selection order, match order in child list, or something else?  We are kinda hand wavey on that at the moment.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -50,11 +51,65 @@ class Listbox extends Foundation<
         return selectedItems;
     }
 
+    /**
+     * returns the the first selectable item in the provided array of children
+     */
+    public static getFirstValidOptionInRange = (
```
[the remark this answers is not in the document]
scomea: The three public static functions (getListboxItemDataFromIds(), getFirstValidOptionInRange() and getItemIndexById()) are used by select component as well.  The select component has need to process the children to figure out selection at times where there is not even a listbox instanciated.  Making these public seemed like a better idea than duplicating them.

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -337,18 +418,23 @@ class Listbox extends Foundation<
...
+            case KeyCodes.tab:
+                return;
+
             case KeyCodes.arrowDown:
             case KeyCodes.arrowRight:
                 focusItemId = this.setFocus(this.state.focusIndex + 1, 1);
-
+                event.preventDefault();
                 if (this.props.multiselectable && event.shiftKey && focusItemId !== "") {
                     const itemProps: ListboxItemProps = Listbox.getItemPropsById(
```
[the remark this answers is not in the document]
scomea: see above

packages/fast-components-react-base/src/listbox/listbox.tsx
```diff
@@ -50,11 +51,65 @@ class Listbox extends Foundation<
...
+                return thisOption;
+            }
+            return null;
+        }
+    };
+
+    /**
+     * Gets the index of an item from it's id by examining children props
+     */
+    public static getItemIndexById(itemId: string, children: React.ReactNode): number {
```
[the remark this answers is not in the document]
scomea: see above

packages/fast-components-react-base/src/listbox/listbox.props.ts
```diff
@@ -43,6 +43,16 @@ export interface ListboxHandledProps extends ListboxManagedClasses {
      * The  onSelectedItemsChanged event handler
      */
     onSelectedItemsChanged?: (selectedItems: ListboxItemProps[]) => void;
+
+    /**
+     * Whether a listitem should automatically get focus when this component is mounted
+     */
+    autoFocus?: boolean;
```
chrisdholt: I think this may just be the terminology, but I believe autoFocus on a `select` would actually focus the control: https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/autofocus

scomea: Renamed to "focusItemOnMount"

PR #1385, opened 2019-02-14
feat: add pivot msft component

eljefe223:
# Description
created MSFT `<Pivot />` component that is based on base `<Tabs />` component. 

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/package.json
```diff
@@ -41,8 +41,8 @@
     "coverageThreshold": {
       "global": {
         "statements": 95,
-        "branches": 90,
-        "functions": 95,
+        "branches": 89,
```
scomea: Should this be changing?

packages/fast-components-react-msft/src/pivot/README.md
```diff
@@ -0,0 +1,8 @@
+# Pivot
+Use *pivot* when there is a paged interface where one controlling tpivot element corresponds to one pivot panel.
```
scomea: tpivot

packages/fast-components-react-msft/package.json
eljefe223: Only if we cant get coverage any higher. Most of this component uses queries to find elements in the DOM. As far as I know there is no way to test against that.

packages/fast-components-react-msft/src/pivot/pivot.spec.tsx
```diff
@@ -0,0 +1,185 @@
...
+            ...handledProps,
+            ...unhandledProps,
+        };
+
+        const rendered: any = mount(<MSFTPivot {...props} />);
+
+        expect(rendered.prop("aria-hidden")).toEqual(true);
+    });
+
+    test("should set offsetX state to when offsetX state is set", () => {
```
scomea: confusing

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,287 @@
...
+                this.props,
+                "managedClasses.pivot_tabPanelContent",
+                ""
+            ),
+            tab: get(this.props, "managedClasses.pivot_item", ""),
+            tab__active: get(this.props, "managedClasses.pivot_item__active", ""),
+            tabPanel: get(this.props, "managedClasses.pivot_tabPanel", ""),
+            tabPanel__hidden: get(
+                this.props,
+                "managedClasses.pivot_tabPanel__hidden",
```
scomea: nit: should probably format the get functions consistently in the block (either all wrap or not)

eljefe223: That's prettier doing its thing. I might be able to line break the other lines but it may just format it back.

packages/fast-components-react-msft/package.json
chrisdholt: This only changes if we manually change it. 

I don't think this should be going down.

packages/fast-components-react-msft/src/pivot/pivot.tsx
chrisdholt: I'd just let prettier do it's thing.

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,287 @@
...
+     * The constructor
+     */
+    constructor(props: PivotProps) {
+        super(props);
+
+        const activeId: string = !this.props.activeId
+            ? this.props.items
+                ? this.props.items[0].id
+                : ""
+            : this.props.activeId;
```
chrisdholt: We should ensure that this is an array before treating it as one. I'd suggest taking a look at carousel - we can use get from lodash and limit how much we need to actually execute here.

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,287 @@
...
+            return mytabsArray.findIndex(this.isSelected);
+        }
+    }
+
+    private generateTabPanelsClassNames(): string {
+        let className: string = get(this.props, "managedClasses.pivot_tabPanels", "");
+        if (this.props.items) {
+            if (this.tabPanelIndex === this.updateTabPanelIndex()) {
+                className = className;
+            } else if (this.tabPanelIndex < this.updateTabPanelIndex()) {
```
chrisdholt: what is the initial "if" doing? Do we need it or we can rewrite this in a way that removes that complexity?

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,287 @@
...
+        return null;
+    }
+
+    protected handledProps: HandledProps<PivotHandledProps> = {
+        label: void 0,
+        items: void 0,
+        managedClasses: void 0,
+    };
+
+    private tabsRef: React.RefObject<any>;
```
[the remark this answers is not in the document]
eljefe223: I believe we need any when putting a ref onto one of our components

packages/fast-components-react-msft/src/pivot/pivot.props.tsx
```diff
@@ -0,0 +1,31 @@
...
+        Subtract<BaseTabsHandledProps, BaseTabsManagedClasses> {
+    /**
+     * The aria-label applied to the tablist for the pivot items
+     */
+    label: string;
+
+    /**
+     * The pivots' pivot item
+     */
+    items?: TabsItem[];
```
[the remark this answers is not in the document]
eljefe223: Yup this can be removed!

packages/fast-components-react-msft/src/pivot/pivot.props.tsx
```diff
@@ -0,0 +1,31 @@
...
+} from "@microsoft/fast-components-class-name-contracts-msft";
+
+export interface PivotManagedClasses extends ManagedClasses<PivotClassNameContract> {}
+export interface PivotHandledProps
+    extends PivotManagedClasses,
+        Subtract<BaseTabsHandledProps, BaseTabsManagedClasses> {
+    /**
+     * The aria-label applied to the tablist for the pivot items
+     */
+    label: string;
```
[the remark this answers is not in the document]
eljefe223: Yup this can be removed!

packages/fast-components-react-msft/src/pivot/pivot.tsx
eljefe223: Carousel assumes items is a required prop... It is not in Base tabs. so if activeId is not set and items is not passsed it fails

eljefe223: I'll add items back to pivot props and make it required.

chrisdholt: @eljefe223 That's not the issue here - that's a typescript assumption. I'm speaking purely in terms of JS. We should ensure that items is an array before we treat it as such.

packages/fast-components-react-msft/src/pivot/pivot.tsx
eljefe223: can be removed since I've now made items a required prop

packages/fast-components-react-msft/src/pivot/pivot.tsx
eljefe223: Agreed that we need that check. I'm just saying that I think its valid change to make items a required props here

packages/fast-components-react-msft/package.json
eljefe223: we are not able to test against queryselectors for grabbing elements in the DOM. I was able to get the test coverage up though by testing some other areas.

packages/fast-components-class-name-contracts-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,61 @@
+import { TabsClassNameContract as BaseTabsClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+/**
+ * The class name contract for the button component
```
nalogiudice: copy/paste typo

packages/fast-components-react-msft/src/pivot/examples.data.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { TabsItem } from "@microsoft/fast-components-react-base";
+import schema from "./pivot.schema.json";
+import Documentation from "./.tmp/documentation";
+import { ComponentFactoryExample } from "@microsoft/fast-development-site-react";
+
+const pivotManagedClasses: PivotManagedClasses = {
+    managedClasses: {
+        pivot: "pivot-class",
+        pivot_tabList: "pivot_tabList-class",
+        pivot_tab: "pivot_tabclass",
```
nalogiudice: nit: missed hyphen `tab-class`

packages/fast-components-react-msft/src/pivot/examples.data.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+function renderPivot(pivotTitle: string): (className?: string) => React.ReactNode {
+    return (className?: string): React.ReactNode => (
+        <div className={className}>{pivotTitle}</div>
+    );
+}
+
+function renderPivotContent(
+    pivotContent: string
+): (className?: string) => React.ReactNode {
```
nalogiudice: I believe we no longer should have functional render props code within examples. Take a look at how carousel does this. @janechu for guidance

packages/fast-components-react-msft/src/pivot/pivot.schema.json
```diff
@@ -0,0 +1,37 @@
...
+        "children": {
+            "title": "Building blocks",
+            "type": "children",
+            "ids": [
+                "tab-item"
+            ]
+        }
+    },
+    "required": [
+        "label"
```
nalogiudice: `id` isn't required?

eljefe223: label is the only required prop of tabs

eljefe223: there should be items though

packages/fast-components-styles-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            overflow: "hidden",
+        },
+        pivot_tabList: {
+            display: "flex",
+            boxSizing: "border-box",
+        },
+        pivot_tab: {
+            minHeight: "32px",
+            padding: "0 12px",
+            fontSize: "14px",
```
nalogiudice: Can we use typography?

packages/fast-components-styles-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            whiteSpace: "nowrap",
+            display: "flex",
+            border: "2px solid",
```
nalogiudice: Why not just `border: "2px solid transparent", instead of bother `border` and `borderColor`?

packages/fast-components-styles-msft/src/pivot/index.ts
chrisdholt: We should at the very least use a helper to map to the font ramp so that if it changes this gets updated for free.

packages/fast-components-styles-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            ...applyFocusVisible({
+                borderColor: designSystem.foregroundColor,
+            }),
+        },
+        pivot_tab__active: {},
+        pivot_activeIndicator: {
+            position: "absolute",
+            borderRadius: toPx(designSystem.cornerRadius),
+            top: "27px",
+            left: "-10px",
```
nalogiudice: LTR?

packages/fast-components-styles-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,106 @@
...
+            display: "block",
+        },
+        pivot_tabPanel__hidden: {
+            display: "none",
+        },
+        pivot_tabPanels: {
+            animationTimingFunction: "cubic-bezier(0.4, 0.0, 0.6, 1.0)",
+        },
+        pivot_tabPanels__fromLeft: {
+            animation: "fromLeft 0.2s",
```
nalogiudice: We tend to use `previous` and `next` instead of `left` and `right`, as to not confuse in LTR/RTL

nalogiudice: I think we should rename all

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    private setActiveIndicatorOffset(): void {
+        if (canUseDOM() && this.tabsRef.current && Array.isArray(this.props.items)) {
+            const tabElement: HTMLElement = ReactDOM.findDOMNode(
+                this.tabsRef.current
+            ) as HTMLElement;
+
+            const mytabs: NodeListOf<Element> = tabElement.querySelectorAll(
+                "[aria-selected='true']"
```
nalogiudice: Do we have no other way of accomplishing this? Using `querySelectorAll` in react is usually a no-no

eljefe223: This is the only way if we derive form the current version of Tabs

packages/fast-components-styles-msft/src/pivot/index.ts
eljefe223: this isn't effected by LTR is simply is for centering the sequence indicator

packages/fast-components-styles-msft/src/pivot/index.ts
eljefe223: We probably should just remove this from here and add the fontsize to the class that gets attached to the renderprop

packages/fast-components-styles-msft/src/pivot/index.ts
eljefe223: Because we are updating the borderColor in the focus state and you can run into weirdness if you use short hand and then target an individual property.

packages/fast-components-react-msft/src/pivot/pivot.tsx
nalogiudice: Hmm. This would be the first occurrence of using `querySelectorAll` in a component... I just want to make sure this is the best route.

packages/fast-components-styles-msft/src/pivot/index.ts
eljefe223: Agree on the classNames animation names I will keep as is and swap those based on direction

packages/fast-components-react-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/*tslint:disable-next-line:typedef */
+const Pivot = manageJss(PivotStyles)(MSFTPivot);
+type Pivot = InstanceType<typeof Pivot>;
+
+interface PivotHandledProps
+    extends Subtract<MSFTPivotHandledProps, PivotManagedClasses> {}
+type PivotProps = ManagedJSSProps<MSFTPivotProps, PivotClassNameContract, DesignSystem>;
+
+export * from "./pivot";
```
nicholasrice: I don't think you want this `export * from "./pivot"` statement.

With that statement, you are actually exporting two interfaces named `PivotHandledProps` with conflicting types. I think we want to be more intentional/explicit about what we're exporting from this index file.

eljefe223: Got it, yeah I think all I need to do is moved PivotManagedClasses into the named export at the bottom of this file.

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+    private generateTabPanelsClassNames(): string {
+        let className: string = get(this.props, "managedClasses.pivot_tabPanels", "");
+        if (this.state.tabPanelIndex === this.prevTabPanelIndex) {
+            className = className;
+        } else if (this.state.tabPanelIndex < this.prevTabPanelIndex) {
+            className = `${className} ${get(
+                this.props,
+                "managedClasses.pivot_tabPanels__previous",
+                "Error"
```
scomea: If we want show error messages in the dom, and I'm not sure we should, it should probably be consistent across our components.

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+     * Gets the direction of the element
+     */
+    private getLTR(): Direction {
+        if (canUseDOM()) {
+            const tabElement: HTMLElement = ReactDOM.findDOMNode(
+                this.tabsRef.current
+            ) as HTMLElement;
+
+            return !tabElement
+                ? Direction.ltr
```
nicholasrice: `getComputedStyle` is a pretty heavy function to call every render cycle because it can usually forces style recalculation and sometimes layout reflow (leading to janky animations) - https://gist.github.com/paulirish/5d52fb081b3570c81e3a. 

One solution would be to default this to the *page* direction, which is assigned by adding the `dir` attribute to the `html` elmeent: https://www.w3.org/International/questions/qa-html-dir 

If necessary, would could also add a `dir` prop to the component to provide the opportunity to override the direction on a per-instance basis, but I'm not confident anyone will actually need that capability.

packages/fast-components-react-msft/src/pivot/pivot.tsx
eljefe223: that is left over thanks for catching that

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    public componentDidMount(): void {
+        this.ltr = this.getLTR();
+        this.setActiveIndicatorOffset();
+        this.prevTabPanelIndex = this.state.tabPanelIndex;
+    }
+
+    public componentDidUpdate(prevProps: PivotProps, prevState: PivotState): void {
+        if (this.ltr !== this.getLTR()) {
```
nicholasrice: You should store the result of `this.getLTR()` as a variable to avoid calling the method twice.

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+        if (this.state.activeId !== prevState.activeId) {
+            this.setActiveIndicatorOffset();
+            this.updateTabPanelIndex();
+        }
+    }
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLButtonElement | HTMLAnchorElement> {
```
nicholasrice: I don't think the function signature is accurate here - does this actually render a `<button />` or `<a />`?

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    private handleTabsUpdate = (activeTabId: string): void => {
+        this.setState({
+            activeId: activeTabId,
+        });
+    };
+
+    private isSelected(element: HTMLElement): boolean {
+        return element.className.includes("active") === true;
```
nicholasrice: checking that a class-name includes "active" will fail in production builds when these turn into gibberish. Is there an aria attribute you can use to determine selection?

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+            );
+            this.setState({
+                tabPanelIndex: mytabsArray.findIndex(this.isSelected),
+            });
+        }
+    }
+
+    private generateTabPanelsClassNames(): string {
+        let className: string = get(this.props, "managedClasses.pivot_tabPanels", "");
+        if (this.state.tabPanelIndex === this.prevTabPanelIndex) {
```
nicholasrice: What is the purpose for assigning `className` to itself? I'm 98% sure you can refactor this to avoid that check

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,240 @@
...
+        return className;
+    }
+
+    private setActiveIndicatorOffset(): void {
+        if (canUseDOM() && this.tabsRef.current && Array.isArray(this.props.items)) {
+            const tabElement: HTMLElement = ReactDOM.findDOMNode(
+                this.tabsRef.current
+            ) as HTMLElement;
+
+            const mytabs: NodeListOf<Element> = tabElement.querySelectorAll(
```
nicholasrice: It looks like you only ever want a single element here - if so, you should use `tabElement.querySelector` instead of `tabElement.querySelectorAll` - `querySelector` will return the first matching element.

packages/fast-components-react-msft/src/pivot/pivot.tsx
eljefe223: chatted offline, we will keep this as is for now.

packages/fast-components-react-msft/src/pivot/pivot.tsx
eljefe223: chatted offline we will keep this as is for now. Once [this](https://github.com/Microsoft/fast-dna/issues/1438) work gets done we will update

packages/fast-components-react-base/src/tabs/tab-panel.tsx
```diff
@@ -59,3 +59,4 @@ class TabPanel extends Foundation<TabPanelHandledProps, TabPanelUnhandledProps,
 
 export default TabPanel;
 export * from "./tab-panel.props";
+export { TabPanelClassNameContract };
```
janechu: why are we exporting these again here?

janechu: why are we exporting these again here?

packages/fast-components-react-base/src/tabs/tab.tsx
```diff
@@ -59,3 +59,4 @@ class Tab extends Foundation<TabHandledProps, TabUnhandledProps, {}> {
 
 export default Tab;
 export * from "./tab.props";
+export { TabClassNameContract };
```
janechu: same issue here, we have only been exporting from base components the handledProps and unhandledProps

packages/fast-components-react-msft/package.json
```diff
@@ -90,6 +90,7 @@
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "eyes.images": "3.6.2",
+    "exenv-es6": "^1.0.0",
```
janechu: You've added this as a devDependency but you are using it as an export, you should add it to the peerDependencies list as well. I believe we have other packages that use it as a peerDependency but we shouldn't rely on that.

packages/fast-components-react-msft/src/pivot/pivot.props.tsx
```diff
@@ -0,0 +1,26 @@
...
+    ManagedClasses,
+    PivotClassNameContract,
+} from "@microsoft/fast-components-class-name-contracts-msft";
+
+export interface PivotManagedClasses extends ManagedClasses<PivotClassNameContract> {}
+export interface PivotHandledProps
+    extends PivotManagedClasses,
+        Subtract<BaseTabsHandledProps, BaseTabsManagedClasses> {
+    /**
+     * The pivots' pivot item
```
janechu: I think this could just be "Items that will make up the pivot and pivot content"

packages/fast-components-styles-msft/src/index.ts
```diff
@@ -101,6 +101,12 @@ export { NumberFieldStyles };
 import ParagraphStyles from "./paragraph";
 export { ParagraphStyles };
 
+/**
+ * Export paragraph styles
```
janechu: paragraph?

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,244 @@
...
+                : getComputedStyle(tabElement).direction === Direction.rtl
+                    ? Direction.rtl
+                    : Direction.ltr;
+        }
+    }
+}
+
+export default Pivot;
+export * from "./pivot.props";
+export { PivotClassNameContract };
```
janechu: again, I would remove this.

packages/fast-components-react-msft/src/pivot/pivot.tsx
```diff
@@ -0,0 +1,244 @@
...
+            tabs: get(this.props, "managedClasses.pivot", ""),
+            tabs_tabPanels: this.generateTabPanelsClassNames(),
+            tabs_tabList: get(this.props, "managedClasses.pivot_tabList", ""),
+            tabs_tabContent: get(this.props, "managedClasses.pivot_tabContent", ""),
+            tabs_tabPanelContent: get(
+                this.props,
+                "managedClasses.pivot_tabPanelContent",
+                ""
+            ),
+            tab: get(this.props, "managedClasses.pivot_tab", ""),
```
janechu: have you tested what these will do if you only defined some of them? (eg. will they result in "undefined" in the class attribute that is rendered)

packages/fast-components-react-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/*tslint:disable-next-line:typedef */
+const Pivot = manageJss(PivotStyles)(MSFTPivot);
+type Pivot = InstanceType<typeof Pivot>;
+
+interface PivotHandledProps
+    extends Subtract<MSFTPivotHandledProps, PivotManagedClasses> {}
+type PivotProps = ManagedJSSProps<MSFTPivotProps, PivotClassNameContract, DesignSystem>;
+
+export { Pivot, PivotProps, PivotHandledProps, PivotManagedClasses, PivotUnhandledProps };
```
janechu: So this is where we would want to export the `PivotClassNameContract`, we do not want to export `PivotManagedClasses`.

What the user needs from this export is:
- The component (Pivot)
- The full prop contract (PivotProps)
- The class name contract (PivotClassNameContract)
- The unhandledProps (PivotUnhandledProps)
- The handledProps (PivotHandledProps)
- any other type definition inside of the props that would be useful, such as enums, this component does not appear to have any

packages/fast-components-react-msft/src/pivot/pivot.props.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
```
janechu: This file can just be a ts file, no need for it to be tsx unless there is actually jsx templating syntax, it theoretically makes testing slower as it needs to pass through two transpiling stages if you use .tsx vs .ts.

packages/fast-components-react-msft/src/pivot/pivot.schema.json
```diff
@@ -0,0 +1,58 @@
...
+                        "pluginId": "@microsoft/fast-components-react-msft/pivot/items/content"
+                    }
+                }
+            }
+        }
+    },
+    "required": [
+        "tab",
+        "content"
+    ]
```
janechu: This is in the wrong location, it should be on the same level as id in items, so you could add it to that array. If it errors when you do that (because the data utilities may not know what to do to generate a required child) then I would remove it altogether.

packages/fast-components-react-msft/src/pivot/pivot.schema.json
```diff
@@ -0,0 +1,56 @@
...
+            "items": {
+                "title": "Item",
+                "type": "object",
+                "properties": {
+                    "id": {
+                        "title": "HTML id attribute",
+                        "type": "string"
+                    }
+                },
+                "required": [
```
nalogiudice: This required is referencing properties not found in items?

eljefe223: Items is of type TabsItem which consists of:
```
export interface TabsItem {
    tab: (className: string) => React.ReactNode;
    content: (className: string) => React.ReactNode;
    id: string;
}
```

packages/fast-components-class-name-contracts-msft/src/pivot/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * The collection of tab panels
+     */
+    pivot_tabPanels?: string;
+
+    /**
+     * The previous panel from left
+     */
+    pivot_tabPanels__previous?: string;
```
chrisdholt: NIT: is this for the animation? If so, perhaps we should be more specific

chrisdholt: *same with `next` below

packages/fast-components-react-msft/src/pivot/examples.data.tsx
```diff
@@ -0,0 +1,115 @@
...
+        props: {
+            children: "pivot one",
+        },
+    } as any,
+    content: {
+        id: pivotItemContentSchema.id,
+        props: {
+            children: "pivot one content",
+        },
+    } as any,
```
chrisdholt: @eljefe223 is there a reason we are coercing prop types here? Last time I saw this it was due to an actual bug.

chrisdholt: Looks like this is due to how we are manipulating children

PR #1400, opened 2019-02-20
feat: add PureFoundation component to foundation package

### LINKED ISSUE #766: PureComponents: Please write ShouldComponentUpdate in Foundation
**Describe the solution you'd like**
I am part of Microsoft News team and inheriting from Foundation class. We would like to suggest a lifecycle event addition in Foundation class. 

Kindly add ShouldComponentUpdate method by checking old and new props and determine if lifecycle should proceed with rendering or skip.

chrisdholt:
closes #766 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-foundation-react/README.md
+++ b/packages/fast-components-foundation-react/README.md
@@ -1,6 +1,6 @@
 # FAST Components foundation React
 
-The foundation component for FAST component packages in React. The default export is an extension from the `React.Component` and includes additional functionality for extension by React component libraries.
+The foundation component for FAST component packages in React. The default export is an extension from the `React.Component` and includes additional functionality for extension by React component libraries. There is also a named export for `PureFoundation` which is an extension from `React.PureComponent`.
 
 ## Installation
 
@@ -94,9 +94,9 @@ Rendered result:
 <div class="my-class-name"></div>
 ```
 
-### `withSlot`
+### `withSlot` (Foundation only)
 
-Gets all children with the `slot` prop that matches a given string, example: `this.withSlot("after")`.
+Gets all children with the `slot` prop that matches a given string, example: `this.withSlot("after")`. 
 
 Component example:
 
@@ -132,7 +132,7 @@ Rendered result:
 </div>
 ```
 
-### `withoutSlot`
+### `withoutSlot` (Foundation only)
 
 Gets all children with the `slot` prop that does not match a given string, example: `this.withoutSlot("after")`.
 
--- a/packages/fast-components-foundation-react/src/pure-foundation/pure-foundation.ts
+++ b/packages/fast-components-foundation-react/src/pure-foundation/pure-foundation.ts
@@ -1,0 +1,138 @@
+import * as React from "react";
+import { get, isPlainObject, pick, set } from "lodash-es";
+import { PureFoundationProps } from "./pure-foundation.props";
+import {
+    HandledProps,
+    ReferenceResolver,
+    ReferenceResolverStore,
+    ReferenceStore,
+} from "../foundation";
+
+/**
+ * The foundation component is the component that all fast base components are built on top of. It provides a common
+ * set of utilities that each component inherits.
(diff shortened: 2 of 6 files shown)


packages/fast-components-foundation-react/README.md
```diff
@@ -94,9 +94,9 @@ Rendered result:
 <div class="my-class-name"></div>
 ```
 
-### `withSlot`
+### `withSlot` (Foundation only)
```
janechu: I feel like this is ambiguous since Foundation is not a named export. Can you place a heading for the shared default and PureFoundation exports, then at the end another heading for the default export only methods?

packages/fast-components-foundation-react/src/pure-foundation/pure-foundation.ts
```diff
@@ -0,0 +1,138 @@
...
+} from "../foundation";
+
+/**
+ * The foundation component is the component that all fast base components are built on top of. It provides a common
+ * set of utilities that each component inherits.
+ * @param H - These are the props that are "handled". "handled" props are not mapped automatically to the root element
+ * returned by the render function. Use handled props to expose inputs that will not map directly to DOM attributes
+ * (eg a custom callback) or where the DOM attribute would be required.
+ * @param U - These are "unhandled" props. Any props from this interface will be mapped onto the root DOM node of the
+ * render function as-is. It is advised that these props map to valid HTML attributes - otherwise you will likely have HTML errors.
```
nicholasrice: I think we should see if there is a way to do this without essentially duplicating Foundation - I believe Mixins give us this functionality: https://mariusschulz.com/blog/typescript-2-2-mixin-classes

chrisdholt: @NicholasRice I agree we should aim to accomplish this without the duplication; however, I don't think mixins are the way to go for two primary reasons. ES6 was launched without support for mixins and React moved away from supporting mixins (16) in favor of HOC's. 

I'm going to consider a few options on how to address this.

nicholasrice: My primary goal is to avoid duplication, so if you have another approach then great. I don't think it is accurate that ES6 doesn't support TypeScript mixins though - after all, TypeScript is simply a superset of JavaScript, and this doesn't have anything inherently to do with React - we're simply trying to abstract class creation.

If you can find another method to abstract this then great, but I believe the above is a viable solution.

chrisdholt: Not trying to be dense or argumentative here, the above comment wasn't intended to diminish the idea of using mixins. After my initial work to read through the TS link you provided, I went to track down other examples such as in React.

I ran into [this](https://reactjs.org/docs/react-without-es6.html#mixins) from their documentation site and [this blog](https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html) as well. I'm not saying that React is correct or that their problems are accurate, only that my initial consideration was to explore additional solutions to reduce duplication rather than work against their existing system (similar to our conversation around renderProp vs slot) which doesn't seem to support ES6 classes.

I'm not able to track down any confirmation of mixins not being supported in ES6 as of now, so you are likely correct in your statement above - all I have are that ES6 launched without support.

Totally happy to go implement that solution and see if we run into any issues. The initial alternative in my head was to extend Foundation and write a shallow compare against props/state in shouldComponentUpdate. Downside there is we have to manage that ourselves and if React changes how they manage PureComponent we won't reap the benefits. I think that in itself is worth trying the mixin route, but wasn't sure if you were aware of the limitations above by React.

nicholasrice: Yea really what we're after is a way extend an input class - any way you can do that I'll be happy with.

chrisdholt: :thumbsup:

Going to close this down and explore a more viable solution.

PR #1405, opened 2019-02-20
feat: add listbox & listbox item to msft components

scomea:
# Description
Adding styled versions of listbox and listbox item to our component library

## Motivation & context
We need a listbox.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/listbox-item/README.md
+++ b/packages/fast-components-react-msft/src/listbox-item/README.md
@@ -1,0 +1,3 @@
+# Listbox item
+
+A item component intended to work as a child of the *listbox*.
--- a/packages/fast-components-react-msft/src/listbox-item/README.md
+++ b/packages/fast-components-react-msft/src/listbox-item/README.md
@@ -1,3 +1,8 @@
-# Listbox item
+## Listbox item
+A component designed to work as a child of the *listbox*.
 
-A item component intended to work as a child of the *listbox*.
+### Usage
+The component required "id" prop must be unique and is applied as an id to to the resulting element in the dom.  Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.
+
+### Accessibility
+The *listbox-item* generates an html element with role="option" as well as "aria-selected" and "aria-disabled" attributes. 
--- a/packages/fast-components-react-msft/src/listbox-item/listbox-item.schema.json
+++ b/packages/fast-components-react-msft/src/listbox-item/listbox-item.schema.json
@@ -1,0 +1,40 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Listbox item",
+    "description": "A listbox item component's schema definition.",
+    "type": "object",
+    "id": "listbox-item",
+    "properties": {
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "displayString": {
+            "title": "Display string",
+            "type": "string"
+        },
+        "id": {
+            "title": "Unique ID",
+            "type": "string"
+        },
+        "value": {
+            "title": "Value",
+            "type": "string"
+        }
+    },
+    "reactProperties": {
+        "children": {
+            "title": "Children",
+            "type": "children",
+            "defaults": [
+                "text"
+            ]
+        },
+        "glyph": {
+            "title": "Glyph",
+            "type": "children",
+            "pluginId": "@microsoft/fast-components-react-msft/listbox-item/glyph"
+        }
+    }
+}
+
--- a/packages/fast-components-styles-msft/src/listbox-item/index.ts
+++ b/packages/fast-components-styles-msft/src/listbox-item/index.ts
@@ -1,0 +1,113 @@
+import designSystemDefaults, {
+    DesignSystem,
+    withDesignSystemDefaults,
+} from "../design-system";
+import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
+import { ListboxItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
+import { density } from "../utilities/density";
+import { defaultHeight, maxHeight, minHeight } from "../utilities/height";
+import {
+    disabledContrast,
+    ensureForegroundNormal,
+    ensureNormalContrast,
+    hoverContrast,
+} from "../utilities/colors";
+import { applyFocusVisible } from "@microsoft/fast-jss-utilities";
+import { applyTypeRampConfig } from "../utilities/typography";
+import typographyPattern from "../patterns/typography";
+import { contrast, toPx } from "@microsoft/fast-jss-utilities";
+import { curry } from "lodash-es";
+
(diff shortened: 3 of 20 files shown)


packages/fast-components-react-msft/src/listbox-item/README.md
```diff
@@ -0,0 +1,3 @@
+# Listbox item
+
+A item component intended to work as a child of the *listbox*.
```
janechu: nit: An*, but I would actually prefer "A component..." item means nothing here.

packages/fast-components-styles-msft/src/listbox/index.ts
```diff
@@ -0,0 +1,25 @@
+import designSystemDefaults, {
+    DesignSystem,
+    withDesignSystemDefaults,
+} from "../design-system";
+import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
+import { ListboxClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+import { foregroundLarge } from "../utilities/colors";
+import { elevation, ElevationMultiplier } from "../utilities/elevation";
+import { toPx } from "@microsoft/fast-jss-utilities";
```
janechu: no need to import this as it is not needed, same with elevation and foreground from what I can see.

packages/fast-components-react-msft/src/listbox-item/listbox-item.schema.json
```diff
@@ -0,0 +1,40 @@
...
+            "type": "string"
+        },
+        "id": {
+            "title": "Unique ID",
+            "type": "string"
+        },
+        "value": {
+            "title": "Value",
+            "type": "string"
+        }
```
janechu: arent id and value required?

packages/fast-components-styles-msft/src/listbox-item/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    const primarySelectedBackground: string = contrast(
+        1.7,
+        designSystem.foregroundColor,
+        designSystem.backgroundColor
+    );
+    return {
+        listboxItem: {
+            listStyleType: "none",
+            height: density(32),
+            display: "grid",
```
janechu: @chrisdholt do you have an opinion on display grid vs flexbox?

packages/fast-components-styles-msft/src/listbox-item/index.ts
```diff
@@ -0,0 +1,113 @@
+import designSystemDefaults, {
```
janechu: @chrisdholt @nicholasrice do you think this should be a msft style component? @scomea what is the use case for this not just being a base component?

packages/fast-components-styles-msft/src/listbox-item/index.ts
scomea: We use grid in context menu item for this also.  (Personally I find grids much, much easier to control than flexbox)

packages/fast-components-styles-msft/src/listbox-item/index.ts
scomea: I think we want to maximize the flexibility of what a listbox item is.  Our version uses displaystring, some other might not and choose to base visuals on children and not use displaystring at all, for example.

chrisdholt: I think I'm hesitant here because I don't know that we have a MSFT listbox (non-select) which is styled. Is this primarily to support select? If so, it seems like an odd abstraction to me.

scomea: How is this different from context menu item?  This is essentially the same pattern.

chrisdholt: We have a design and style for context menu item. I don't know that we have a design for *just* as listbox. @bheston can advise on this from a design standpoint. We could have a "select item" here I think - as we have a design for select which happens to be composed of listbox items. I just don't know if we have a default style for a functional listbox. While listboxes can be styled but I don't think we have a default style or a listbox design system component. We have select which is styled and functionally made up of listbox, but those are different things I think.

scomea: I had considered having an "option" componentb (and context) instead of "listbox item" to try to be more generic.

packages/fast-components-styles-msft/src/listbox-item/index.ts
janechu: That may be true, but flexbox is more performant

packages/fast-components-styles-msft/src/listbox-item/index.ts
janechu: I'm still not seeing the use case or the request from design for this to be a component, let's put it this way, just because we have a tabs component does not mean that there is an msft styled tabs component (we do, pivot, but that's not the point). Things are not necessarily 1:1 for base and msft.

Consider that the reason tabs was added as a base component was because it is an established functional pattern. If there was no need for a tabs component styled for msft, we simply would never add it. However tabs can still be used as a pattern for an msft component like carousel. Or there could simply be the case where we never need to use tabs in msft because there is no corresponding component, that's OK too.

I believe, and correct me if I'm wrong, the only requirement we have right now is for a msft select.

packages/fast-components-styles-msft/src/listbox-item/index.ts
scomea: That doesn't seem to always be the case (and I'm not sure the difference is significant) - https://blogs.igalia.com/jfernandez/2015/06/24/performance-on-grid-layout/

packages/fast-components-react-msft/src/listbox-item/README.md
scomea: Actually I had better text in the base component that I copied over.  

## Listbox item
A component designed to work as a child of the *listbox*.

### Usage
The component required "id" prop must be unique and is applied as an id to to the resulting element in the dom.  Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.

### Accessibility
The *listbox-item* generates an html element with role="option" as well as "aria-selected" and "aria-disabled" attributes.

packages/fast-components-styles-msft/src/listbox/index.ts
scomea: and topx...deleted

packages/fast-components-styles-msft/src/listbox-item/index.ts
scomea: So perhaps we don't have listbox as a styled component and the styled "select" component uses listbox base, and then instead of a styled listbox item we have a styled component called "select option" which wraps the base listbox item?

scomea: See the "Select Base" PR for what the above proposal looks like.

packages/fast-components-react-msft/src/listbox-item/listbox-item.schema.json
scomea: yep

scomea: added required items in the base select pr before pulling this one.

Issue #1410, opened 2019-02-21
Add carousel functionality and styles


nalogiudice:
Update carousel in several places:

- new [design updates](https://www.figma.com/file/FnigSVQnGOXbEW8HQuEu1ZMo/UIS-%2F-Components-(light-theme)?node-id=0%3A1) (to sequence indicator and flippers) 
- adds form generator editor functionality for slide clontent (can now update test slide data)
- adds the functionality to add class names based on slide transition direction change

PR #1411, opened 2019-02-21
feat: update carousel functionality and design

nalogiudice:
Closes [#1410](https://github.com/Microsoft/fast-dna/issues/1410)
# Description

Update carousel in several places:

- new [design updates](https://www.figma.com/file/FnigSVQnGOXbEW8HQuEu1ZMo/UIS-%2F-Components-(light-theme)?node-id=0%3A1) (to sequence indicator and flippers) 
- adds form generator editor functionality for slide clontent (can now update test slide data)
- adds the functionality to add class names based on slide transition direction change

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
@@ -67,4 +67,14 @@ export interface CarouselClassNameContract {
      * The dark slide theme
      */
     carousel__themeDark?: string;
+
+    /**
+     * The previous slide transition
+     */
+    carousel__slideTransitionDirectionPrevious?: string;
+
+    /**
+     * The next slide transition
+     */
+    carousel__slideTransitionDirectionNext?: string;
 }
--- a/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
@@ -71,10 +71,10 @@ export interface CarouselClassNameContract {
     /**
      * The previous slide transition
      */
-    carousel__slideTransitionDirectionPrevious?: string;
+    carousel__slideAnimatePrevious?: string;
 
     /**
      * The next slide transition
      */
-    carousel__slideTransitionDirectionNext?: string;
+    carousel__slideAnimateNext?: string;
 }
--- a/packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
+++ b/packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
@@ -4,5 +4,14 @@
     "description": "A carousel dark image content component's schema definition.",
     "id": "carousel-dark-image-content",
     "type": "object",
-    "properties": {}
+    "properties": {
+        "src": {
+            "title": "HTML src attribute",
+            "type": "string"
+        },
+        "alt": {
+            "title": "HTML alt attribute",
+            "type": "string"
+        }
+    }
 }
--- a/packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
+++ b/packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
@@ -11,13 +11,17 @@
             "properties": {
                 "src": {
                     "title": "HTML src attribute",
-                    "type": "string"
+                    "type": "string",
+                    "default": "http://placehold.it/1399x600/2F2F2F/171717"
                 },
                 "alt": {
                     "title": "HTML alt attribute",
                     "type": "string"
                 }
-            }
+            },
+            "required": [
+                "src"
+            ]
         }
     }
 }
--- a/packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
+++ b/packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
@@ -5,13 +5,19 @@
     "id": "carousel-dark-image-content",
     "type": "object",
     "properties": {
-        "src": {
-            "title": "HTML src attribute",
-            "type": "string"
-        },
-        "alt": {
-            "title": "HTML alt attribute",
-            "type": "string"
+        "image": {
+            "title": "Image",
+            "type": "object",
+            "properties": {
+                "src": {
+                    "title": "HTML src attribute",
+                    "type": "string"
+                },
+                "alt": {
+                    "title": "HTML alt attribute",
+                    "type": "string"
+                }
+            }
         }
     }
 }
--- a/packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
+++ b/packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
@@ -17,8 +17,8 @@ import {
 export interface CarouselHeroContentProps {
     heading?: HeadingProps;
     paragraph?: ParagraphProps;
-    callToAction: CallToActionProps;
-    image: ImageProps;
+    callToAction?: CallToActionProps;
+    image?: ImageProps;
     className?: string;
 }
 
--- a/packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
+++ b/packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
@@ -2,21 +2,54 @@ import * as React from "react";
 import {
     CallToAction,
     CallToActionAppearance,
+    CallToActionProps,
     Heading,
+    HeadingProps,
     HeadingSize,
     HeadingTag,
     Image,
+    ImageProps,
     Paragraph,
+    ParagraphProps,
     ParagraphSize,
 } from "../../src/index";
+
+export interface CarouselHeroContentProps {
+    heading?: HeadingProps;
+    paragraph?: ParagraphProps;
+    callToAction: CallToActionProps;
+    image: ImageProps;
+    className?: string;
+}
 
 /**
  * A stand-in a mock hero
  */
-export default class CarouselHeroContent extends React.Component<{}, {}> {
+export default class CarouselHeroContent extends React.Component<
+    CarouselHeroContentProps,
+    {}
+> {
+    public static defaultProps: CarouselHeroContentProps = {
+        heading: {
+            children: "Heading text",
+        },
+        paragraph: {
+            children: "Hero paragraph test test",
+        },
+        image: {
+            src: "http://placehold.it/1399x600/2F2F2F/171717",
+            alt: "Placeholder image",
+        },
+        callToAction: {
+            children: "Call to action",
+            href: "#",
+            appearance: CallToActionAppearance.primary,
+        },
+    };
+
     public render(): React.ReactNode {
         return (
-            <div {...this.props}>
+            <div className={this.props.className}>
                 <div
                     style={{
                         justifyContent: "center",
@@ -44,7 +77,7 @@ export default class CarouselHeroContent extends React.Component<{}, {}> {
                         <Heading
                             tag={HeadingTag.h2}
                             size={HeadingSize._3}
-                            children={"Hero title"}
+                            children={this.props.heading.children}
                         />
                         <Paragraph
                             size={ParagraphSize._1}
@@ -52,16 +85,13 @@ export default class CarouselHeroContent extends React.Component<{}, {}> {
                         />
                         <CallToAction
                             style={{ marginTop: "16px" }}
-                            appearance={CallToActionAppearance.primary}
-                            children={"Call to action"}
-                            href={"#"}
+                            appearance={this.props.callToAction.appearance}
+                            children={this.props.callToAction.children}
+                            href={this.props.callToAction.href}
                         />
                     </div>
                 </div>
-                <Image
-                    src={"http://placehold.it/1399x600/2F2F2F/171717"}
-                    alt={"Placeholder image"}
-                />
+                <Image src={this.props.image.src} alt={this.props.image.alt} />
             </div>
         );
     }
--- a/packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
+++ b/packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
@@ -34,7 +34,7 @@ export default class CarouselHeroContent extends React.Component<
             children: "Heading text",
         },
         paragraph: {
-            children: "Hero paragraph test test",
+            children: "Hero paragraph test text",
         },
         image: {
             src: "http://placehold.it/1399x600/2F2F2F/171717",
--- a/packages/fast-components-react-msft/src/carousel/README.md
+++ b/packages/fast-components-react-msft/src/carousel/README.md
@@ -1,3 +1,7 @@
 # Carousel
 
 *Carousel* allows the user to page through content blocks one at a time or arrow through a sequence of smaller content blocks.
+
+## Style guidance
+
+Class names have been provided and applied to capture the transition of the slides, and allow animation or transformation depending on slide entry position.
--- a/packages/fast-components-react-msft/src/carousel/README.md
+++ b/packages/fast-components-react-msft/src/carousel/README.md
@@ -1,7 +1,7 @@
 # Carousel
 
-*Carousel* allows the user to page through content blocks one at a time or arrow through a sequence of smaller content blocks.
+*Carousel* allows the user to move through a series of slides.
 
 ## Style guidance
 
-Class names have been provided and applied to capture the transition of the slides, and allow animation or transformation depending on slide entry position.
+Class names have been provided and applied to capture the transition of the slides, and allow animation or transformation depending on incoming slide entry position.
--- a/packages/fast-components-react-msft/src/carousel/carousel.tsx
+++ b/packages/fast-components-react-msft/src/carousel/carousel.tsx
@@ -22,36 +22,12 @@ export interface CarouselState {
     activeId: string;
 }
 
-enum SlideTransitionDirection {
-    previous = "previous",
-    next = "next",
-}
-
 class Carousel extends Foundation<
     CarouselHandledProps,
     CarouselUnhandledProps,
     CarouselState
 > {
     public static displayName: string = "Carousel";
-
-    /**
-     * React life-cycle method
-     */
-    public static getDerivedStateFromProps(
-        nextProps: CarouselProps,
-        prevState: CarouselState
-    ): null | Partial<CarouselState> {
-        if (
-            typeof nextProps.activeId === "string" &&
-            nextProps.activeId !== prevState.activeId
-        ) {
-            return {
-                activeId: nextProps.activeId,
-            };
-        }
-
-        return null;
-    }
 
     /**
      * Handled props
@@ -66,7 +42,7 @@ class Carousel extends Foundation<
     /**
      * Initial slide transition direction is none (on carousel load)
      */
-    private slideTransitionDirection: SlideTransitionDirection | "" = "";
+    private slideTransitionDirection: FlipperDirection = null;
 
     /**
      * Define constructor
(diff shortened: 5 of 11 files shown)


packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -82,62 +88,62 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
...
-                },
-            },
         },
         carousel__themeLight: {
             "& $carousel_flipperPrevious, & $carousel_flipperNext": {
                 color: "white",
                 fill: "white",
-                background: "black",
+                background: "rgba(0, 0, 0, 0.6)",
+                border: "1px solid #CECECE",
```
chrisdholt: @bheston can you advise if this is part of a larger change to flipper? If it is - @nalogiudice can we do the flipper change in a new PR and we can get this in and updated?

nalogiudice: We would still have to override here as regular flipper uses foreground and background colors, and this uses hard-coded color values

bheston: Realizing that we still need to take in all the recent color work, what aspect of this requires the hard-coded colors? Is this to do with the light and dark theme support only? Perhaps that's something we can fix when we take in the new color work, or perhaps that indicates that we actually want the light/dark support in Flipper itself. I think any use we can think of for Flipper would benefit from the opacity, like for instance in something like PDF or other document navigation.
Also, if we do leave them hard coded here, the border should be 60% as well, so we should probably switch that to opacity on the backplate as a whole.

chrisdholt: Thanks both - sounds like we need to just tackle flipper in a new task unrelated to this? I assume the carousel work will get validated more when we do a pass with the new color stuff. I think the only action is ensuring that @bheston validates the final design changes for carousel specifically.

nalogiudice: > ... what aspect of this requires the hard-coded colors? Is this to do with the light and dark theme support only?

Correct. 

> or perhaps that indicates that we actually want the light/dark support in Flipper itself

Interesting thought - could be useful.

> I think the only action is ensuring that @bheston validates the final design changes for carousel specifically. - @chrisdholt 

Ok, so for this PR I will just validate the overrides are correct. When we update flipper, we can circle back and change what we need to here. Sound good?

chrisdholt: Good

bheston: Works for me.

nalogiudice: @bheston `rgba(206, 206, 206, 0.6)` for the `#CECECE` border change?

nalogiudice: Looks about right to me with the rgba. Went ahead and made the change.

nalogiudice:
I believed I have resolved all of the relevant code climate issue - @chrisdholt

packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
```diff
@@ -67,4 +67,14 @@ export interface CarouselClassNameContract {
      * The dark slide theme
      */
     carousel__themeDark?: string;
+
+    /**
+     * The previous slide transition
+     */
+    carousel__slideTransitionDirectionPrevious?: string;
```
awentzel: I haven't looked at how contracts are done closely so this comment may seem out of left field.  Can you describe how you came to this name?  Should there be another property for the type of animation/transition?  It's already a CarouselClass so why does Carousel need to be in the name of the properties as well.  Feels redundant.  Consider PreviousSlide or NextSlide.

packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
```diff
@@ -2,21 +2,54 @@ import * as React from "react";
...
+export default class CarouselHeroContent extends React.Component<
+    CarouselHeroContentProps,
+    {}
+> {
+    public static defaultProps: CarouselHeroContentProps = {
+        heading: {
+            children: "Heading text",
+        },
+        paragraph: {
+            children: "Hero paragraph test test",
```
awentzel: test test or test te**x**t

awentzel: Also, consider adding long text at times to capture overflow and localization issues.

packages/fast-components-react-msft/src/carousel/README.md
```diff
@@ -1,3 +1,7 @@
 # Carousel
 
 *Carousel* allows the user to page through content blocks one at a time or arrow through a sequence of smaller content blocks.
```
awentzel: Would arrow key also work for paging through content blocks?  Is "page" the most accurate way of describing this?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -22,6 +22,12 @@ export interface CarouselState {
     activeId: string;
 }
 
+enum SlideTransitionDirection {
+    previous = "previous",
+    next = "next",
+    initial = "",
```
awentzel: Curious if initial to null or maybe if this might be an index and then set to 0 as default. That way it's possible to set to any slide as the start.  Just making stuff up on my end.  But, initial isn't really in the language i've used for slides before.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -58,6 +64,9 @@ class Carousel extends Foundation<
         items: void 0,
     };
 
+    private slideTransitionDirection: Partial<SlideTransitionDirection> =
```
awentzel: Are there things that might TransitionDirection other than slides?  It might be nice to extract this out one level.  Then have TransitionDirection or TransitionTiming etc that could be applied to many different things besides just slides.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -80,27 +89,35 @@ class Carousel extends Foundation<
...
-                    )}
-                />
+                {this.isSingleSlide && (
+                    <Flipper
+                        direction={FlipperDirection.previous}
+                        onClick={this.handlePreviousClick}
+                        className={get(
+                            this.props,
+                            "managedClasses.carousel_flipperPrevious",
+                            ""
```
awentzel: I don't think i've seen an answer yet, but, still confused why we have `""` on classNames.  I've seen this used a lot.  What's the case "" would not be necessary or why couldn't the get method support 2 or 3 input parameters using function overloading?

chrisdholt: `get` is a utility function from lodash which takes 3 params. The first is an object, the second is a path, and the third is a default value. Get goes to retrieve the value of the object's path - if it's undefined we return the default value. In our scenario, if the className doesn't exist, we pass empty strings.

We use `get` here because managed classes are a prop value and not a hard coded value. They are essentially a variable which could be undefined. In this file, the class names don't actually exist. They ultimately don't exist until they are passed by the HOC (manageJSS) to this component as props. In the case this is undefined, we would rather pass an empty string (no class) rather than a non-string `undefined` value.

packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
nalogiudice: @awentzel 
All class name contracts include their root element. This is common across all MSFT components.

This is simply a class that a dev can key off of the add any animations or transitions when switching slides. The name was chosen to be explicit.

I came to this naming convention by this rationale: 
`[main element]_[element][action][action type]`
`[carousel]_[slide][transition][direction previous]`

packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
nalogiudice: Normally yes, but this is a mock component (not a real hero in any way) so it doesn't have any constraints. I will fix the text though.

packages/fast-components-react-msft/src/carousel/README.md
nalogiudice: This is carousel text from older documentation... I could see this as a bit convoluted though.

"Carousel allows the user to move through a series of slides" could be a better way of saying this.

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: When a carousel loads fresh, it has no sense of "previous slide direction" because nothing has slid. I am simply setting the starting variable to an empty string in a slightly more fancy way.

I could just do 
`private slideTransitionDirection: Partial<SlideTransitionDirection> = "";`
instead of
`private slideTransitionDirection: Partial<SlideTransitionDirection> = SlideTransitionDirection.initial;`

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: `Slide` is just the internal language of content blocks. Once the direction is known, the appropriate class is added. Then the user can add any animation or transition needed.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -110,15 +127,16 @@ class Carousel extends Foundation<
...
-                this.getSlideTheme() === SlideTheme.light
-                    ? get(this.props, "managedClasses.carousel__themeLight", "")
-                    : get(this.props, "managedClasses.carousel__themeDark", "");
+            className += this.assignSlideThemeClassName();
+        }
 
-            className += ` ${theme}`;
+        if (
+            this.slideTransitionDirection === SlideTransitionDirection.next ||
+            this.slideTransitionDirection === SlideTransitionDirection.previous
```
chrisdholt: Do we need to check both of these because "internal" exists? Is it possible we could simplify this and just check for this.slideTransitionDirection?

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: since the enum isn't exported and is just internal, can we simply remove it?

packages/fast-components-react-msft/src/carousel/README.md
awentzel: I think that would be an improvement. please change

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: changed

packages/fast-components-react-msft/src/carousel/README.md
nalogiudice: Changed

packages/fast-components-react-msft/app/components/carousel-dark-image-content.schema.json
```diff
@@ -4,5 +4,20 @@
...
     "type": "object",
-    "properties": {}
+    "properties": {
+        "image": {
+            "title": "Image",
+            "type": "object",
+            "properties": {
+                "src": {
+                    "title": "HTML src attribute",
+                    "type": "string"
```
janechu: can you give this a default value and make the src required?

packages/fast-components-react-msft/app/components/carousel-hero-content.tsx
```diff
@@ -2,21 +2,54 @@ import * as React from "react";
...
     Paragraph,
+    ParagraphProps,
     ParagraphSize,
 } from "../../src/index";
 
+export interface CarouselHeroContentProps {
+    heading?: HeadingProps;
+    paragraph?: ParagraphProps;
+    callToAction: CallToActionProps;
+    image: ImageProps;
```
janechu: can you either make these optional with ? or add these to required in the JSON schema.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -22,6 +22,11 @@ export interface CarouselState {
     activeId: string;
 }
 
+enum SlideTransitionDirection {
+    previous = "previous",
+    next = "next",
+}
```
janechu: doesnt the FlipperDirection enum have the same values as this enum?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -58,6 +63,11 @@ class Carousel extends Foundation<
         items: void 0,
     };
 
+    /**
+     * Initial slide transition direction is none (on carousel load)
+     */
+    private slideTransitionDirection: SlideTransitionDirection | "" = "";
```
janechu: Should this be null? Empty string seems odd.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -80,27 +90,35 @@ class Carousel extends Foundation<
...
-                <Flipper
-                    direction={FlipperDirection.previous}
-                    onClick={this.handlePreviousClick}
-                    className={get(
-                        this.props,
-                        "managedClasses.carousel_flipperPrevious",
-                        ""
-                    )}
-                />
+                {this.isSingleSlide && (
```
janechu: we typically will not do any inline functionality within a rendering template like this, I would prefer to see a this.renderSingleSlide() method

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -110,15 +128,13 @@ class Carousel extends Foundation<
...
         if (this.getSlideTheme()) {
-            theme =
-                this.getSlideTheme() === SlideTheme.light
-                    ? get(this.props, "managedClasses.carousel__themeLight", "")
-                    : get(this.props, "managedClasses.carousel__themeDark", "");
+            className += this.assignSlideThemeClassName();
+        }
 
-            className += ` ${theme}`;
+        if (this.slideTransitionDirection) {
```
janechu: See previous note, I would like this to check for null explicitly, empty string is a weird value to purposely assign

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -42,23 +44,30 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
...
+                opacity: "0.2",
+                border: "1px solid transparent",
+                borderRadius: "40px",
                 content: "''",
                 display: "block",
-                height: "100%",
-                width: "100%",
+                height: "4px",
+                width: "32px",
+                transition: "all 0.2s ease-in-out",
```
janechu: just checking, but are none of these values based on the design system?

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -82,62 +91,62 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
             "& $carousel_flipperPrevious, & $carousel_flipperNext": {
                 color: "black",
                 fill: "black",
-                background: "white",
+                background: "rgba(255, 255, 255, 0.6)",
+                border: "1px solid rgba(206, 206, 206, 0.6)",
```
janechu: same here, are none of these values based on the design system?

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -82,62 +91,62 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
...
                 "& span::before": {
                     borderColor: "black",
                 },
                 "&:hover": {
+                    background: "white",
```
janechu: should this not be foreground/background?

nalogiudice: No, color should not change on theme change. These are hard-coded on purpose. A dark or light slide always has the same styles, regardless of design system theme change.

packages/fast-components-styles-msft/src/carousel/index.ts
nalogiudice: See other response.

packages/fast-components-styles-msft/src/carousel/index.ts
nalogiudice: See other response.

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Sure, I can use that

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Changed

packages/fast-components-react-msft/src/carousel/carousel.tsx
nalogiudice: Sure, I'll switch to null

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -82,62 +91,62 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
             "& $carousel_flipperPrevious, & $carousel_flipperNext": {
                 color: "black",
                 fill: "black",
```
eljefe223: Should this be pure black? Most cases we don't use black we use `#101010`

nalogiudice: @bheston for guidance

bheston: Is that the foreground color for the chevron icons? If so, it should be #101010. I know flippers might be a little different because of opacity, but I think most of the fills and outlines should be some other non-black color on the neutral color ramp. The only thing that is currently full black is text over light-colored accent buttons.

nalogiudice: Ok, changed all black to `#101010`

packages/fast-components-react-msft/src/carousel/examples.data.tsx
```diff
@@ -72,12 +108,13 @@ export default {
...
+            activeId: detailTabItem[2].id,
         },
-        {
-            label: "A set of example text content",
-            items: singleTabItem,
-        } as any,
+        // {
+        //     label: "A single slide carousel",
+        //     items: singleTabItem,
+        // } as any,
```
janechu: still needed?

packages/fast-components-styles-msft/src/carousel/index.ts
janechu: In that case can you make these constants that are referred to as they are used multiple times.

chrisdholt: FYI - these exist in the vNext release branch as constants.

nalogiudice: Updated to consts for now

packages/fast-components-class-name-contracts-msft/src/carousel/index.ts
chrisdholt: In pivot, we have [block]_[el]__animatePrevious

Can we go less verbose and not lose any of the semantics here?

nalogiudice: The class is added to the overall carousel element, so there's no inner element per se. It could be `carousel__slideAnimatePrevious` I suppose. I think having `slide` enumerated is important, or else the class name becomes too vague.

chrisdholt: I'm good adding slide into it. Distinction makes sense

nalogiudice: Will change to `carousel__slideAnimatePrevious` and `carousel__slideAnimateNext` then

Issue #1399, opened 2019-02-20
Document handled vs unhandled prop guidance 


nicholasrice:
We need clear documentation on when a prop is handled vs unhandled. Documentation should clearly communicate the criteria that makes a prop handled:
- Any HTML attribute name that is required (in general we should avoid HTML attribute names)
  - id, title, etc 
- Any prop name that is not a valid HTML attribute
- Any HTML attribute not applied to the HTML element populated with `unhandledProps()` - for instance, `value` of checkbox where the root of the `Checkbox` component is actually a `div`

We should also clearly document the unhandledProps()` mechanism and it's intended use.

chrisdholt:
closed with #3517

PR #1422, opened 2019-02-21
feat: add base select component

scomea:
# Description
Adds a select component to the base react components.

## Motivation & context
We need a select component.

## Issue type checklist

## Process & policy checklist

- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://microsoft.github.io/fast-dna/docs/en/contributing/standards) for this project


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+            if (
+                this.props.multiselectable === false &&
+                this.state.selectedItems.length > 1
+            ) {
+                this.updateSelection([this.state.selectedItems[0]]);
+            }
+            this.toggleMenu(this.checkPropsForMenuState());
+        }
+        if (prevProps.isMenuOpen !== this.props.isMenuOpen) {
+            this.toggleMenu(this.checkPropsForMenuState());
```
janechu: seems like there is nothing preventing this from being toggled multiple times even though the result should be the same.

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+    /**
+     * This exists only to suppress a build warning
+     */
+    private onSelectValueChange = (event: React.ChangeEvent): void => {
+        return null;
+    };
+
+    /**
+     * Deternmines which function to use to render content display (ie. the part of the control that shows when the menu isn't open)
```
janechu: nit: Determines* - you might want to run this file through spellcheck, I see this word spelt wrong a few times

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+    private static idPropertyKey: string = "id";
+
+    /**
+     * Handled props instantiation
+     */
+    protected handledProps: HandledProps<SelectHandledProps> = {
+        isMenuOpen: void 0,
+        disabled: void 0,
+        displayStringFormatterFunction: void 0,
```
janechu: "Function" seems redundant, it's like calling something myIndexVariable.

I would prefer:

displayStringFormatterFunction => displayStringFormatter
contentDisplayRenderFunction => contentFormatter (is this what it is doing?)
menuRenderFunction => menuFormatter

packages/fast-components-react-base/src/select/select.schema.json
```diff
@@ -0,0 +1,54 @@
...
+        "form": {
+            "title": "Form",
+            "type": "string"
+        },
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
+        },
+        "isMenuOpen": {
+            "title": "IsMenuOpen",
```
janechu: This title is not readable

packages/fast-components-react-base/src/select/select.schema.json
```diff
@@ -0,0 +1,54 @@
...
+        "placeholder": {
+            "title": "Placeholder",
+            "type": "string"
+        },
+        "name": {
+            "title": "Name",
+            "type": "string"
+        },
+        "form": {
+            "title": "Form",
```
janechu: Form... ID?

scomea:
Note: I have added a styled select component to this PR for discussion purposes (ie. still needs some cleanup) regarding how we structure select & listbox across base and styled component libs

packages/fast-components-class-name-contracts-base/src/select/index.ts
```diff
@@ -0,0 +1,29 @@
...
+export interface SelectClassNameContract {
+    /**
+     * The root of the select component
+     */
+    select?: string;
+
+    /**
+     * The root of the select component
+     */
+    orange?: string;
```
janechu: orange?

packages/fast-components-react-base/src/select/select.props.ts
```diff
@@ -0,0 +1,111 @@
...
+import { SelectState } from "./select";
+
+export interface SelectManagedClasses extends ManagedClasses<SelectClassNameContract> {}
+export interface SelectUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+export interface SelectHandledProps extends SelectManagedClasses {
+    /**
+     * The children populate the select menu, any SelectOption components in the
+     */
+    children?: React.ReactNode | React.ReactNode[];
```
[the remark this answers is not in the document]
scomea: removed

packages/fast-components-react-base/src/select/select.props.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Specifies that the control is disabled
+     */
+    disabled?: boolean;
+
+    /**
+     * Defines one or more forms the select field belongs to
+     */
+    form?: string;
```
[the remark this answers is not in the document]
scomea: "form" is the attribute name used on native select so I thought I'd keep that.  I don't object to changing to "formId" as it is more descriptive if folks prefer that over matching the native attributes.

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+    ): string | string[] => {
+        if (this.props.multiselectable) {
+            return newSelection.map((thisItem: ListboxItemProps) => {
+                return thisItem.value;
+            });
+        } else {
+            if (newSelection.length === 0) {
+                return "";
+            } else {
+                return newSelection[0].value;
```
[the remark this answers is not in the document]
scomea: removed

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+    };
+
+    /**
+     * Handles clicks
+     */
+    private selectClicked = (e: React.MouseEvent): void => {
+        if (this.props.disabled || e.defaultPrevented) {
+            return;
+        }
+        this.state.isMenuOpen ? this.toggleMenu(false) : this.toggleMenu(true);
```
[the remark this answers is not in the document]
scomea: Done

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+        } else {
+            returnValue = placeholder;
+        }
+        return returnValue;
+    };
+
+    /**
+     * Handles clicks
+     */
+    private selectClicked = (e: React.MouseEvent): void => {
```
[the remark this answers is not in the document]
scomea: done

packages/fast-components-react-base/src/select/select.tsx
scomea: I added a check to toggleMenu() so that we don't setState unnecessarily.

        if (shouldOpenMenu !== this.state.isMenuOpen) {
            this.setState({
                isMenuOpen: shouldOpenMenu,
            });
        }

packages/fast-components-react-base/src/select/select.tsx
scomea: corrected

packages/fast-components-react-base/src/select/select.schema.json
scomea: fixed

packages/fast-components-react-base/src/select/select.schema.json
scomea: Sure, that makes for a more descriptive title.

packages/fast-components-class-name-contracts-base/src/select/index.ts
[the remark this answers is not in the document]
scomea: yep, test string

packages/fast-components-class-name-contracts-msft/src/select-option/index.ts
```diff
@@ -0,0 +1,11 @@
+import { ListboxItemClassNameContract as BaseListboxItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+/**
+ * The class name contract for the context menu item component
+ */
+export interface SelectOptionClassNameContract extends BaseListboxItemClassNameContract {
+    /**
+     * Select option content region
```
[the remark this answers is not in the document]
scomea: I don't think so.  The idea is that the base listbox item component encapsulates functionality that can be reused accross multiple other components but when it gets styled it gets a name appropriate to the context it is being used in. ie. Select -> SelectOption, AutoSelect -> AutoSelectItem, etc....

packages/fast-components-react-base/src/select/select.props.ts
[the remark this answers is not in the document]
chrisdholt: if we're using the native attribute name, we should use something like `formId` per @janechu's advice per #1399

packages/fast-components-react-base/src/select/select.props.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Defines a name for the drop-down list
+     */
+    name?: string;
+
+    /**
+     * String displayed when there is no selected value
+     */
+    placeholder?: string;
```
chrisdholt: I think placeholder maps to an unhandledProp on HTMLDivElement

packages/fast-components-react-msft/src/select/select.props.ts
```diff
@@ -0,0 +1,25 @@
...
+} from "@microsoft/fast-components-class-name-contracts-msft";
+
+export interface SelectManagedClasses extends ManagedClasses<SelectClassNameContract> {}
+export interface SelectHandledProps
+    extends SelectManagedClasses,
+        Subtract<BaseSelectHandledProps, BaseSelectManagedClasses> {
+    /**
+     * The trailing content
+     */
+    placeholdertemp?: string;
```
chrisdholt: `placeholderTemp`?

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,470 @@
...
+     * constructor
+     */
+    constructor(props: SelectProps) {
+        super(props);
+
+        this.state = {
+            selectedItems: [],
+            value: "",
+            displayString: "",
+            isMenuOpen: false,
```
chrisdholt: I think that since `isMenuOpen` is a handled prop, we actually want state to be something where we set the prop value or default to false, `isMenuopen: this.props.isMenuOpen || false`

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,470 @@
...
+    /**
+     * constructor
+     */
+    constructor(props: SelectProps) {
+        super(props);
+
+        this.state = {
+            selectedItems: [],
+            value: "",
+            displayString: "",
```
chrisdholt: Assuming for the initial state here, we want to default to a prop value if it exists

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,470 @@
...
+            shouldOpenMenu = this.props.isMenuOpen;
+        } else if (this.props.multiselectable === true) {
+            shouldOpenMenu = true;
+        } else {
+            if (shouldOpenMenu) {
+                window.addEventListener("click", this.handleWindowClick);
+            } else {
+                window.removeEventListener("click", this.handleWindowClick);
+            }
+        }
```
chrisdholt: Do we want to be adding and removing listeners for this over and over? That seems like something we're over-managing. Why not just add this when the component mounts and remove when the component unmounts?

packages/fast-components-class-name-contracts-base/src/select/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * The menu of the select component
+     */
+    select_menu?: string;
+
+    /**
+     * The menu open modifier
+     */
+    select__menuOpen?: string;
```
nalogiudice: Since `select_menu` exists, in the sense of `block_element`, shouldn't this actually be `select_menu__open`? Same with below

packages/fast-components-react-base/src/select/select.tsx
scomea: Agreed.  Changed it.

packages/fast-components-react-base/src/select/select.props.ts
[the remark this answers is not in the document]
scomea: formId it is

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+    };
+
+    /**
+     * Increment selection
+     */
+    private incrementSelectedOption = (increment: number): void => {
+        const childrenAsArray: React.ReactNode[] = React.Children.toArray(
+            this.props.children
+        );
+        const iterationIncrement: number = increment > -1 ? 1 : -1;
```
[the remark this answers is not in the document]
scomea: yep, removed variable

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+                childrenAsArray,
+                increment
+            );
+            if (validOption !== null) {
+                this.updateSelection([(validOption as React.ReactElement<any>).props]);
+            }
+        }
+
+        this.toggleMenu(true);
+        return;
```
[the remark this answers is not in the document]
scomea: Yep

packages/fast-components-class-name-contracts-msft/src/select-option/index.ts
```diff
@@ -0,0 +1,26 @@
+import { ListboxItemClassNameContract as BaseListboxItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
+
+/**
+ * The class name contract for the context menu item component
+ */
+export interface SelectOptionClassNameContract extends BaseListboxItemClassNameContract {
+    /**
+     * The root of the select option component
+     */
+    SelectOption?: string;
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-class-name-contracts-msft/src/select-option/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * The disabled modifier
+     */
+    SelectOption__disabled?: string;
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-class-name-contracts-msft/src/select-option/index.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * The selected modifier
+     */
+    SelectOption__selected?: string;
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/select/select.props.ts
scomea: What do you suggest as an alternative?

packages/fast-components-react-base/src/select/select.tsx
scomea: Yep, added

packages/fast-components-class-name-contracts-base/src/select/index.ts
scomea: sure, done

packages/fast-components-react-base/src/select/select.tsx
scomea: Added similar logic (but with a function since we reuse this logic)

packages/fast-components-react-base/src/select/select.props.ts
```diff
@@ -0,0 +1,102 @@
...
+    contentDisplayRenderFunction?: (
+        props: SelectProps,
+        state: SelectState
+    ) => React.ReactNode;
+
+    /**
+     * Function which converts and formats the string may be displayed in the ui
+     * in the UI
+     */
+    displayStringFormatterFunction?: (
```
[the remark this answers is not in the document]
scomea: sure, changed

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -125,10 +133,9 @@ class ListboxItem extends Foundation<
      * Handle the keydown event of the item
      */
     private handleClick = (e: React.MouseEvent<HTMLDivElement>): void => {
-        if (this.props.disabled) {
+        if (this.props.disabled || e.defaultPrevented) {
```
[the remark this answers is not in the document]
scomea: Because we provide a render function for the display component it is possible a developer could do something we don't expect there, and should there be interactions there a developer could preventDefault on clicks and key presses they want the parent control to ignore.  Same thing with the children of the list.

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+                    if (validOption !== null) {
+                        this.updateSelection([
+                            (validOption as React.ReactElement<any>).props,
+                        ]);
+                    }
+                }
+            }
+        } else {
+            const startIndex: number = increment > -1 ? 0 : childrenAsArray.length - 1;
+            const endIndex: number = increment > -1 ? childrenAsArray.length - 1 : 0;
```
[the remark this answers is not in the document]
scomea: done

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+            const validOption: React.ReactNode = Listbox.getFirstValidOptionInRange(
+                startIndex,
+                endIndex,
+                childrenAsArray,
+                increment
+            );
+            if (validOption !== null) {
+                this.updateSelection([(validOption as React.ReactElement<any>).props]);
```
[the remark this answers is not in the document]
scomea: done

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,479 @@
...
+    };
+
+    /**
+     * Toggles the menu
+     */
+    private toggleMenu = (isMenuOpen: boolean): void => {
+        let shouldOpenMenu: boolean = isMenuOpen;
+        if (this.props.isMenuOpen !== undefined) {
+            shouldOpenMenu = this.props.isMenuOpen;
+        } else if (this.props.multiselectable === true) {
```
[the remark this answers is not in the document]
scomea: Yes

packages/fast-components-react-base/src/select/select.tsx
scomea: Did this:
displayStringFormatterFunction => displayStringFormatter
contentDisplayRenderFunction =>trigger
menuRenderFunction => no longer exists

packages/fast-components-react-msft/src/select/select.props.ts
scomea: Thought I needed something, then I didn't...  removed

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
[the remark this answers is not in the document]
scomea: Modified click and keydown functions to not invoke if default prevented, but otherwise calls user provided functions (ie. they can check for default prevented themselves)

packages/fast-components-react-base/src/select/select.props.ts
scomea: Discussed w. Jane, decided we should keep the prop names that match the html select element props.

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+            if (
+                this.props.multiselectable === false &&
+                this.state.selectedItems.length > 1
+            ) {
+                this.updateSelection([this.state.selectedItems[0]]);
+            }
+            this.toggleMenu(this.checkPropsForMenuState());
+        }
+        if (prevProps.isMenuOpen !== this.props.isMenuOpen) {
+            this.toggleMenu(this.checkPropsForMenuState());
```
janechu: The issue for me here is that it's very hard to discern all the logical paths that might or might not happen when you are firing function multiple times with unrelated criteria. I would prefer something more like:

```
let updatedSelection: ListboxItemProps[];
let updatedMenuVisibility: boolean;

if (prevProps.selectedItems !== this.props.selectedItems) {
    updatedSelection = Listbox.getListboxItemDataFromIds(
        this.props.selectedItems,
        this.props.children
    );
}

if (prevProps.multiselectable !== this.props.multiselectable) {
    // in the case that the multiselect is turned off but there are multiple items selected,
    // choose the first item in the list of selected items
    if (
        this.props.multiselectable === false &&
        this.state.selectedItems.length > 1
    ) {
        updatedSelection = [this.state.selectedItems[0]];
    }

    updatedMenuVisibility = this.checkPropsForMenuState();
}

if (prevProps.isMenuOpen !== this.props.isMenuOpen) {
    updatedMenuVisibility = this.checkPropsForMenuState();
}

if (updatedSelection !== undefined) {
    this.updateSelection(updatedSelection);
}

if (typeof toggledMenu === "boolean") {
    this.toggleMenu(updatedMenuVisibility);
}
```

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+            initialSelection = Listbox.getListboxItemDataFromIds(
+                this.props.selectedItems,
+                this.props.children
+            );
+        } else {
+            initialSelection = Listbox.getListboxItemDataFromIds(
+                this.props.defaultSelection,
+                this.props.children
+            );
+        }
```
janechu: nit: I would prefer this reads the same way I suggested above:

```
initialSelection = Listbox.getListboxItemDataFromIds(
    this.props.selectedItems !== undefined ? this.props.selectedItems : this.props.defaultSelection,
    this.props.children
);
```

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+    private renderHiddenSelectElement(): React.ReactNode {
+        return (
+            <select
+                required={this.props.required || null}
+                name={this.props.name || null}
+                form={this.props.form || null}
+                value={this.state.value}
+                multiple={this.props.multiselectable || null}
+                disabled={this.props.disabled || null}
+                onChange={this.onSelectValueChange}
```
janechu: It bears asking the question @chrisdholt asked earlier, should the unhandledProps be on the hidden select element and removed from the handled props entirely?

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+     */
+    private updateSelection = (newSelection: ListboxItemProps[]): void => {
+        const newValue: string | string[] = this.getValueFromSelection(newSelection);
+        const newDisplayString: string = this.getFormattedDisplayString(newSelection);
+        this.setState({
+            selectedItems: newSelection,
+            value: newValue,
+            displayString: newDisplayString,
+        });
+        if (this.props.onValueChange) {
```
janechu: I would prefer if this checked for an actual type, eg. `typeof this.props.onValueChange === "function"`

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+            selectedOptions.forEach((thisOption: ListboxItemProps) => {
+                if (returnValue.length > 0) {
+                    returnValue = returnValue + separator;
+                }
+                returnValue =
+                    returnValue +
+                    (thisOption.displayString === undefined
+                        ? thisOption.value
+                        : thisOption.displayString);
+            });
```
janechu: for this method I believe you can just do:

```
const optionValues: string[] = selectedOptions.map(
    (selectedOption: ListboxItemProps): string => {
        return selectedOption.displayString || selectedOption.value;
    }
);

return selectOptions.length > 0 ? optionValues.join(", ") : placeholder;
```

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+        if (validOption !== null) {
+            this.updateSelection([(validOption as React.ReactElement<any>).props]);
+        }
+    };
+
+    /**
+     * Toggles the menu
+     */
+    private toggleMenu = (desiredMenuState: boolean): void => {
+        const shouldOpenMenu: boolean = this.validateMenuState(desiredMenuState);
```
janechu: I feel like the const should be renamed to `updatedIsMenuOpen` or something because its not exactly determining if a menu should open, but whether it should be opened OR closed

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -0,0 +1,499 @@
...
+            initialSelection = Listbox.getListboxItemDataFromIds(
+                this.props.selectedItems,
+                this.props.children
+            );
+        } else {
+            initialSelection = Listbox.getListboxItemDataFromIds(
+                this.props.defaultSelection,
+                this.props.children
+            );
+        }
```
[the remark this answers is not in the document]
scomea: Yep

packages/fast-components-react-base/src/select/select.tsx

packages/fast-components-react-base/src/select/select.tsx
scomea: Some of these are actually used by the select component code so they are technically "handled".  My general take is that since we're dealing with a finite list of props for native select and this is very much a "component internal" I'd be more worried that arbitrarily passing props down is more likely to cause weird bugs than provide a useful api.

packages/fast-components-react-base/src/select/select.tsx
scomea: yep

packages/fast-components-react-base/src/select/select.tsx
scomea: yep, cleaner

packages/fast-components-react-base/src/select/select.tsx
scomea: sure,  I circled around what to call this so happy to go with your suggestion

packages/fast-components-react-base/src/select/select.tsx
scomea: yep

packages/fast-components-react-msft/src/select/select.tsx
```diff
@@ -0,0 +1,80 @@
...
+                afterContent={this.generateGlyph}
+            >
+                {state.displayString}
+            </Button>
+        );
+    };
+
+    /**
+     * Gets the dropdown glyph
+     * TODO: scomea - replace with ref to a glyph resource
```
nicholasrice: Can you write an issue to reference this against?

scomea: Done - https://github.com/Microsoft/fast-dna/issues/1488

packages/fast-components-react-base/src/select/select.schema.json
```diff
@@ -0,0 +1,62 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Select",
+    "description": "A select component's schema definition.",
+    "type": "object",
+    "id": "select",
```
janechu: This will need to be brought inline with our new ID policy, per @chrisdholt work in the v4 PR, this should be `@microsoft/fast-components-react-msft/select`

packages/fast-components-react-msft/src/select-option/select-option.schema.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Select option",
+    "description": "A select option component's schema definition.",
+    "type": "object",
+    "id": "select-option",
```
janechu: This ID must be changed to `@microsoft/fast-components-react-msft/select-option`

packages/fast-components-react-msft/src/select/select.schema.json
```diff
@@ -0,0 +1,62 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Select",
+    "description": "A select component's schema definition.",
+    "type": "object",
+    "id": "select",
```
janechu: This ID must be changed to `@microsoft/fast-components-react-msft/select`

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        select__multiSelectable: {
+            "& $select_menu": {
+                position: "static",
+                boxShadow: "none",
+                border: "1px solid",
+                borderColor: foregroundColor,
+            },
+        },
+
```
janechu: nit: we don't typically add all these extra empty lines between properties.

packages/fast-components-react-base/src/select/select.tsx
chrisdholt: This is fine

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    );
+
+    return {
+        select: {
+            minWidth: "276px",
+            maxWidth: "374px",
+        },
+
+        select_toggleGlyph: {
+            fill: { foregroundColor },
```
janechu: This shouldn't be in an object

packages/fast-components-styles-msft/src/select-option/index.ts
```diff
@@ -0,0 +1,122 @@
...
+            gridColumnStart: "3",
+            overflow: "hidden",
+            textOverflow: "ellipsis",
+        },
+        selectOption_glyph: {
+            gridColumnStart: `${applyLocalizedProperty("2", "4", direction)}`,
+            display: "inline-block",
+            position: "relative",
+            maxWidth: "16px",
+            margin: "0 12px 0 0",
```
janechu: per https://developer.mozilla.org/en-US/docs/Web/CSS/margin this is adding a margin-right value, please add RTL

packages/fast-components-styles-msft/src/select-option/index.ts
```diff
@@ -0,0 +1,122 @@
...
+                background: hoverContrast(
+                    designSystem.contrast,
+                    designSystem.backgroundColor
+                ),
+            },
+        },
+        selectOption_contentRegion: {
+            gridColumnStart: "3",
+            overflow: "hidden",
+            textOverflow: "ellipsis",
```
janechu: I believe we have an ellipses utility

packages/fast-components-react-msft/src/select/select.tsx
janechu: Can you add that issue number to the TODO?

packages/fast-components-react-msft/src/select/select.schema.json
```diff
@@ -0,0 +1,62 @@
...
+            "title":"Labelled by",
+            "type":"string"
+        }
+    },
+    "reactProperties": {
+        "children": {
+            "title": "Building blocks",
+            "type": "children",
+            "defaults": [
+                "text"
```
janechu: I don't think this should have a text default, shouldn't it have select options for defaults? you can add an ids array with the select option JSON schema id.

packages/fast-components-react-msft/src/select-option/select-option.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+    SelectOptionProps,
+    SelectOptionUnhandledProps,
+} from "./index";
+
+/*
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("listbox menu item", (): void => {
```
janechu: listbox?

packages/fast-components-react-base/src/select/select.schema.json
```diff
@@ -0,0 +1,62 @@
...
+            "title":"Labelled by",
+            "type":"string"
+        }
+    },
+    "reactProperties": {
+        "children": {
+            "title": "Building blocks",
+            "type": "children",
+            "defaults": [
+                "text"
```
janechu: per my other comment, would this not be list item?

packages/fast-components-react-base/src/select/select.schema.json
scomea: done

packages/fast-components-react-msft/src/select-option/select-option.schema.json
scomea: done

packages/fast-components-react-msft/src/select/select.schema.json
scomea: done

packages/fast-components-styles-msft/src/select/index.ts
scomea: ok

packages/fast-components-styles-msft/src/select/index.ts
scomea: ok

packages/fast-components-react-msft/src/select-option/select-option.spec.tsx
scomea: fixed

packages/fast-components-styles-msft/src/select-option/index.ts
scomea: yep

packages/fast-components-styles-msft/src/select-option/index.ts
scomea: used it

packages/fast-components-react-msft/src/select/select.schema.json
scomea: yep

packages/fast-components-react-base/src/select/select.schema.json
scomea: yep

packages/fast-components-react-msft/src/select/select.tsx
scomea: done

packages/fast-components-react-msft/src/select-option/select-option.schema.json
```diff
@@ -0,0 +1,44 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Select option",
+    "description": "A select option component's schema definition.",
+    "type": "object",
+    "id": "@microsoft/fast-components-react-base/select-option",
```
janechu: should be msft, not base.

packages/fast-components-react-msft/src/select/examples.data.tsx
```diff
@@ -0,0 +1,76 @@
...
+const examples: ComponentFactoryExample<SelectProps> = {
+    name: "Select",
+    component: Select,
+    schema: schema as any,
+    documentation: <Documentation />,
+    detailData: {
+        placeholder: "Select an option",
+        children: [
+            {
+                id: "@microsoft/fast-components-react-base/select-option",
```
janechu: this will need to be updated after you make the schema change from base to msft

packages/fast-components-react-msft/src/select-option/select-option.schema.json
scomea: doh, fixed

packages/fast-components-react-msft/src/select/examples.data.tsx
scomea: fixed

PR #1424, opened 2019-02-22
chore: update number field documentation

nalogiudice:
# Description

Copy/paste error in MSFT number field readme

Copy base number field readme to MSFT number field readme

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/number-field/README.md
```diff
@@ -1,11 +1,7 @@
...
 A *number field* is a form component which allows users to input and edit numerical data.
 
 ## Usage
 
-Use sentence case (e.g. "Last name"), but avoid punctuation such as colons. *Label* text should be short and succinct, not a full sentence. When labeling components, write the *label* text as a noun or a concise noun phrase, without using full sentences. Using commas to separate list items is an exception to this rule.
-
-## Behavioral guidance
-
-Use *number field* when numeric values of input are required. Use step to control number accuracy. It is recommended to have a label above the number field to indicate its usage.
+The *number field* is not appropriate for input that happens to only consist of numbers but isn’t strictly speaking a number. For example, it would be inappropriate for credit card numbers or US postal codes. A simple way of determining whether to use the *number field* is to consider whether it would make sense for the input control to have a spinbox interface (e.g., with “up” and “down” arrows). When a number field is not appropriate, use text field instead.
```
awentzel: I feel like this might be more clear if we focus first on when to use it as opposed to when not to use it.  Then maybe provide a couple actual examples if at all possible.  What is a spinbox interface?  I'm not actually clear when to use it by reading the documentation.

chrisdholt: I think this is good feedback. I always have appreciated when we've stated things from a "Use `component` when..." and then followed with, "don't use this for _____". It just flows better and gives me context for the acceptable use case right away

nalogiudice: @chrisdholt & @awentzel how does this read now?

chrisdholt: Much more clear and has a more positive feel to it (I get it, that sounds weird)...but I think the recent push is much better and gives me positive context on when and how to use appropriately.

packages/fast-components-react-base/src/number-field/README.md
```diff
@@ -1,7 +1,9 @@
...
 
 ## Usage
 
-The *number field* is not appropriate for input that happens to only consist of numbers but isn’t strictly speaking a number. For example, it would be inappropriate for credit card numbers or US postal codes. A simple way of determining whether to use the *number field* is to consider whether it would make sense for the input control to have a spinbox interface (e.g., with “up” and “down” arrows). When a number field is not appropriate, use text field instead.
\ No newline at end of file
+As an `<input>` element of `type=number`, *number field* is used to let the user enter purely numerical values, such as quantity. Inputs of `type=number` include built-in validation to reject non-numerical entries. The browser provides default stepper arrows (i.e. the up and down arrows found within the input control) to let the user increase and decrease the value using either mouse or keyboard. *Number field* is not appropriate for input that happens to only consist of numbers but isn’t strictly speaking a number. For example, it would be inappropriate for credit card numbers or US postal codes. When a number field is not appropriate, use text field instead.
```
nicholasrice: Regarding "*Number field* is not appropriate for input that happens to only consist of numbers but isn’t strictly speaking a number":

These are (or can be) strictly a number, so I think we could rephrase this part. To me, the distinction is really whether increment and decrement are useful operations on the value - typically, those operations on numbers used as identifiers are not useful.

nalogiudice: Ok, how does this read now?

PR #1425, opened 2019-02-22
feat: add neutral fill input recipe

nicholasrice:
adds neutral fill input color recipe

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/color/neutral-fill-input.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/neutral-fill-input.ts
@@ -1,0 +1,81 @@
+import {
+    DesignSystem,
+    DesignSystemResolver,
+    ensureDesignSystemDefaults,
+    withDesignSystemDefaults,
+} from "../../design-system";
+import { clamp, memoize } from "lodash-es";
+import { findClosestSwatchIndex, palette, Palette, PaletteType, Swatch } from "./palette";
+import {
+    ColorRecipe,
+    FillSwatch,
+    StatefulSwatchToColorRecipeFactory,
+    SwatchStates,
+} from "./common";
+
+/**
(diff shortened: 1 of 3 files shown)


packages/fast-components-styles-msft/src/utilities/color/neutral-fill-input.ts
```diff
@@ -0,0 +1,81 @@
...
+    (designSystem: DesignSystem): string => {
+        return designSystem.backgroundColor;
+    }
+);
+
+export function neutralFillInput(designSystem: DesignSystem): FillSwatch;
+export function neutralFillInput(
+    backgroundResolver: (designSystem: DesignSystem) => Swatch
+): (designSystem: DesignSystem) => FillSwatch;
+export function neutralFillInput(arg: any): any {
```
nalogiudice: total nit: don't we usually stay away from abbreviations (i.e. `arg`)

nicholasrice: I think "arg" is an industry-standard acronym - which is our guideline. Happy to change if folks don't think it is though.

packages/fast-components-styles-msft/src/utilities/color/neutral-fill-input.ts
```diff
@@ -0,0 +1,81 @@
...
+            (designSystem: DesignSystem): FillSwatch => {
+                return neutralFillInputAlgorithm(
+                    Object.assign({}, designSystem, {
+                        backgroundColor: arg(designSystem),
+                    })
+                );
+            }
+        );
+    } else {
+        return neutralFillInputAlgorithm(withDesignSystemDefaults(arg));
```
janechu: nit: don't actually need to wrap this in an else

PR #1426, opened 2019-02-22
fix: remove defaultValue from number field schema and examples

chrisdholt:
# Description
The `defaultValue` attribute for inputs are expected to be string. I noticed that we were typing our examples as "any" in several places to overcome type errors. I think these errors were actually valid. Default value is expected as a string, our examples and schema showed it as a number. I don't believe this is a breaking change as the components actual contract is not changing.

## Issue type checklist

- [] This change causes current functionality to break.

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/number-field/number-field.schema.json
+++ b/packages/fast-components-react-base/src/number-field/number-field.schema.json
@@ -37,10 +37,6 @@
             "title": "HTML required attribute",
             "type": "boolean"
         },
-        "defaultValue": {
-            "title": "Initial value",
-            "type": "number"
-        },
         "value": {
             "title": "HTML value attribute",
             "type": "number"
--- a/packages/fast-components-react-base/src/number-field/examples.data.tsx
+++ b/packages/fast-components-react-base/src/number-field/examples.data.tsx
@@ -15,7 +15,7 @@ const classes: NumberFieldManagedClasses = {
     },
 };
 
-const examples: ComponentFactoryExample<NumberFieldHandledProps> = {
+const examples: ComponentFactoryExample<NumberFieldProps> = {
     name: "Number field",
     component: NumberField,
     schema: schema as any,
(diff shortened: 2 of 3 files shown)


nalogiudice:
> Default value is expected as a string, our examples and schema showed it as a number.

While this is true (it being a string), we would not want to allow users to enter strings for the defaultValue, when it should in fact end up being a number (strings are not valid in type=number)

Since we pass everything as an unhandledProp, we can't perform any parsing on the defultValue (i.e. turn it into a number from a string)

Shouldn't we still show defaultValue? If you remove it, any value shown will be uncontrolled (unable to change).

chrisdholt:
> > Default value is expected as a string, our examples and schema showed it as a number.
> 
> While this is true (it being a string), we would not want to allow users to enter strings for the defaultValue, when it should in fact end up being a number (strings are not valid in type=number)
I may be reading into this - but I don't think we want to coerce values here. This is being removed based on the documentation you've written.
> Since we pass everything as an unhandledProp, we can't perform any parsing on the defultValue (i.e. turn it into a number from a string)
Nor do I think we should, thus removing it from the schema and examples.
> Shouldn't we still show defaultValue? If you remove it, any value shown will be uncontrolled (unable to change).
I think that's more of a tooling issue. While we want the tooling to work and to be able to address scenarios like this, I don't think we should hack the component do so. 

I initially thought we may want to pull defaultValue from unhandled props, but I don't think we should be that prescriptive; we can let users solve for that scenario themselves.

packages/fast-components-react-base/src/number-field/number-field.schema.json
```diff
@@ -37,10 +37,6 @@
             "title": "HTML required attribute",
             "type": "boolean"
         },
-        "defaultValue": {
-            "title": "Initial value",
-            "type": "number"
```
janechu: remove this or just type it as a string?

chrisdholt: I think we should remove for now.

PR #1431, opened 2019-02-22
fix: add additional wrapper to dialog to prevent subpixel aliasing due to css transforms

### LINKED ISSUE #1339: Blurry content in Dialog component
In some cases the content of a dialog, text in particular, can be blurry.  This is because of issues with using transform: translate to position the dialog.

We noticed the issue with the print dialog in a prototype which had a window positioned with a transform hosting a dialog, also positioned with a transform.  The problem was fixed by overriding styles and using css styling other than transform to position the dialog's content region.  

We should investigate whether we can improve the dialog component so that this is not an issue developers have to deal with.

### LINKED ISSUE #1399: Document handled vs unhandled prop guidance

chrisdholt:
closes #1339 (not 99...)

I don't believe this is a breaking change as the API is not affected. It does add a new div wrapper and class name contract, but the styled component remains centered.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-class-name-contracts-base/src/dialog/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/dialog/index.ts
@@ -8,6 +8,11 @@ export interface DialogClassNameContract {
     dialog?: string;
 
     /**
+     * The positioning region for the dialog
+     */
+    dialog_positioningRegion?: string;
+
+    /**
      * The element that overlays the page while the modal is visible
      */
     dialog_modalOverlay?: string;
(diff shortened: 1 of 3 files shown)


packages/fast-components-class-name-contracts-base/src/dialog/index.ts
```diff
@@ -7,6 +7,11 @@ export interface DialogClassNameContract {
      */
     dialog?: string;
 
+    /**
+     * The wrapper of the dialog contents
+     */
+    dialog_wrapper?: string;
```
janechu: should we be more specific to what is being wrapped? for example "dialog_contentRegionWrapper"

PR #1432, opened 2019-02-22
feat: add color library with color space conversions (rpg, hsl, hsv, xyz, lab, lch), blending functions, and palette generation

eigenvector32:
add typescript color library with color space conversions for rpg, hsl, hsv, xyz, lab, lch, color blending functions, color palette generation and unit tests and sample app

This code is intended to eventually replace the use of chroma. 

## Issue type checklist

## Process & policy checklist


packages/fast-colors/app/index.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react";
```
chrisdholt: Since this is going into the v-next branch, I believe we're updating all instances of `import * as ____` to `import ____ from "";` as we are leveraging esModuleInterop

eigenvector32: Those should all be updated now

packages/fast-colors/app/src/PaletteEditor.tsx
```diff
@@ -0,0 +1,199 @@
...
+> {
+  /*-------------
+    CONSTRUCTOR
+  --------------*/
+
+  constructor(props: IPaletteEditorProps) {
+    super(props);
+    this.state = {
+      paletteList: [
+        parseColor("#e74856")!,
```
nicholasrice: I'm not familiar with what the `!` is doing here - can you explain that? I also see this below when accessing properties of a `ColorRGBA64`

packages/fast-colors/src/colorlib/colorRGBA64.ts
```diff
@@ -0,0 +1,131 @@
+// tslint:disable:member-ordering
+import {
+    clamp,
+    denormalize,
+    getHexStringForByte,
+    roundToPrecisionSmall,
+} from "./mathUtils";
+
+export class ColorRGBA64 {
+    public static fromObject(data: any): ColorRGBA64 | null {
```
nicholasrice: should `data: any` be `data: {r: number, g: number, b: number}`?

packages/fast-colors/src/colorlib/parseColor.ts
```diff
@@ -0,0 +1,219 @@
...
+    violet: new ColorRGBA64(0.933333, 0.509804, 0.933333, 1),
+    wheat: new ColorRGBA64(0.960784, 0.870588, 0.701961, 1),
+    white: new ColorRGBA64(1, 1, 1, 1),
+    whitesmoke: new ColorRGBA64(0.960784, 0.960784, 0.960784, 1),
+    yellow: new ColorRGBA64(1, 1, 0, 1),
+    yellowgreen: new ColorRGBA64(0.603922, 0.803922, 0.196078, 1),
+};
+
+// accepts formats
+// #RRGGBB
```
nicholasrice: I think we should split up the parsers into smaller functions, with `parseColor` invoking those functions as necessary:
`parseRGB`, `parseRGBA`, and `parseHEX` or something similar.

I expect almost every consumer of this library to depend on color parsing - however I think a much less likely scenario is a user needs to parse the color "yellow" (or any other named color). As this is architected today, I cannot depend on that function without *also* depending on the 147 named colors above.

Splitting out these functions would give app authors the proper tools to remove code they don't need.

packages/fast-colors/src/colorlib/colorRGBA64.ts
```diff
@@ -0,0 +1,131 @@
...
+            "#" +
+            getHexStringForByte(denormalize(this.r, 0.0, 255.0)) +
+            getHexStringForByte(denormalize(this.g, 0.0, 255.0)) +
+            getHexStringForByte(denormalize(this.b, 0.0, 255.0))
+        );
+    }
+
+    // #RRGGBBAA
+    public toStringHexRGBA(): string {
+        return (
```
nicholasrice: you can simplify this to `return this.toStringHexRGB() + getHexStringForByte(denormalize(this.a, 0.0, 255.0))`

packages/fast-colors/app/src/PaletteEditor.tsx
eigenvector32: In Typescript, the ! operator can be used to assert that an object is not undefined. EG: in this case the ! asserts that the returned value is not undefined.

packages/fast-colors/__tests__/colorBlendingTest.test.ts
```diff
@@ -0,0 +1,241 @@
+// tslint:disable:no-string-literal
+// tslint:disable:no-console
```
nalogiudice: Do you still need this no-console tslint?

packages/fast-colors/app/src/PaletteEditor.tsx
chrisdholt: Interesting, I've never seen it used _after_ an object before

packages/fast-colors/app/src/PaletteDisplay/ColorSwatch.tsx
```diff
@@ -0,0 +1,78 @@
+/*---------
```
nalogiudice: We typically do not leave these types of comments (i.e. "IMPORTS" and "JSS STYLES")

If we do have comments above an interface or function, it would be formatted like:
```
/**
 * Comment text
 */
interface foo {}
```

packages/fast-colors/src/colorlib/colorLAB.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    public readonly l: number;
+    public readonly a: number;
+    public readonly b: number;
+
+    public equalValue(rhs: ColorLAB): boolean {
+        return this.l === rhs.l && this.a === rhs.a && this.b === rhs.b;
+    }
+
+    public roundToPrecision(precision: number): ColorLAB {
```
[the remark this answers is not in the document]
eigenvector32: roundToPrecisionSmall does the same thing as Math.Round(int precision) in C#. That is, it gives you "precision" number of digits to the right of the decimal point (which differs from javascript toFixed which returns a string with the specified number of digits combined across both left and right of the decimal point). The reason the function is named roundToPrecisionSmall is because the function will fail for very large numbers (ones where i*10^(precision+1) is larger than the max value of number). But in the color types we know the numbers will not be anywhere near that max.

packages/fast-colors/app/src/PaletteEditor.tsx
nicholasrice: Super cool - learn something new every day.

packages/fast-colors/src/colorlib/colorLAB.ts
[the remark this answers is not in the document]
nicholasrice: Got it - okay cool, thanks.

packages/fast-colors/src/colorlib/colorConverters.ts
```diff
@@ -0,0 +1,384 @@
...
+    let sat: number = 0;
+    if (delta !== 0) {
+        sat = delta / (1 - Math.abs(2 * lum - 1));
+    }
+
+    return new ColorHSL(hue, sat, lum);
+}
+
+export function hslToRGB(hsl: ColorHSL, alpha: number = 1): ColorRGBA64 {
+    const c: number = (1 - Math.abs(2 * hsl.l - 1)) * hsl.s;
```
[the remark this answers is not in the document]
eigenvector32: In this particular case, the numbers c, x and m are very abstract intermediate values in the calculation so don't have physically relevant names. Will look through the file to see if there are other cases where names might be improved though.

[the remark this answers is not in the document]
nicholasrice: I thought that might be the case - no worries.

nicholasrice:
@eigenvector32 I just realized that adding the app as-is is going to cause circular package dependencies - perhaps it would be better to not have the app in this package. Otherwise, you'll need to refactor so that the app doesn't make any dependency that also depends on `@microsoft/fast-colors`

packages/fast-colors/src/colorlib/colorBlending.ts
```diff
@@ -0,0 +1,214 @@
...
+        blendMultiplyChannel(bottom.g, top.g),
+        blendMultiplyChannel(bottom.b, top.b),
+        1
+    );
+}
+
+export function blendOverlayChannel(bottom: number, top: number): number {
+    if (bottom < 0.5) {
+        return clamp(2.0 * top * bottom, 0, 1);
+    } else {
```
janechu: nit: since its a return value you don't need the else

packages/fast-colors/src/colorlib/colorConverters.ts
```diff
@@ -0,0 +1,385 @@
...
+
+// All hue values are in degrees rather than radians or normalized
+// All conversions use the D65 2 degree white point for XYZ
+// Info on conversions and constants used can be found in the following:
+// https://en.wikipedia.org/wiki/CIELAB_color_space
+// https://en.wikipedia.org/wiki/Illuminant_D65
+// https://ninedegreesbelow.com/photography/xyz-rgb.html
+// http://user.engineering.uiowa.edu/~aip/Misc/ColorFAQ.html
+// https://web.stanford.edu/~sujason/ColorBalancing/adaptation.html
+// http://brucelindbloom.com/index.html
```
janechu: thank you for including these references

packages/fast-colors/src/colorlib/colorConverters.ts
```diff
@@ -0,0 +1,385 @@
...
+
+// The alpha channel of the input is ignored
+export function rgbToLuminance(rgb: ColorRGBA64): number {
+    function luminanceHelper(i: number): number {
+        if (i <= 0.03928) {
+            return i / 12.92;
+        } else {
+            return Math.pow((i + 0.055) / 1.055, 2.4);
```
janechu: nit: same comment as earlier, all else values with no qualifiers and a return can just return without the else wrapper.

packages/fast-colors/src/colorlib/colorConverters.ts
```diff
@@ -0,0 +1,385 @@
...
+}
+
+export function temperatureToRGB(tempKelvin: number, alpha: number = 1): ColorRGBA64 {
+    // The constants I could find assumed a decimal range of [0,255] for each channel. Just going to put a /255.0 at the end
+    let r: number = 0;
+    let g: number = 0;
+    let b: number = 0;
+
+    if (tempKelvin <= 1000) {
+        tempKelvin = 1000;
```
janechu: We've been avoiding the reassigning values from arguments passed, it almost feels like a side effect from running this function that tempKelvin is normalized to 1000 or 40000. Whats the opinion on this and if we go forward I suggest we document what this will do to the passed argument as it's not what I would have expected.

packages/fast-colors/src/colorlib/colorLAB.ts
```diff
@@ -0,0 +1,45 @@
...
+import { roundToPrecisionSmall } from "./mathUtils";
+
+// CIELAB color space
+// https://en.wikipedia.org/wiki/CIELAB_color_space
+// This implementation uses the D65 constants for 2 degrees. That determines the constants used for the pure white point of the XYZ space of 0.95047, 1.0, 1.08883.
+// https://en.wikipedia.org/wiki/Illuminant_D65
+// These constants determine how the XYZ, LCH and LAB colors convert to/from RGB.
+export class ColorLAB {
+    public static readonly epsilon: number = 216 / 24389;
+    public static readonly kappa: number = 24389 / 27;
```
janechu: Not a big deal since the impact would be small, but why not just have these as `0.00885645168` and `903.296296`?

packages/fast-colors/src/colorlib/colorPalette.ts
```diff
@@ -0,0 +1,271 @@
...
+// tslint:disable:interface-name
+
+import { blendMultiply, blendOverlay, saturateViaLCH } from "./colorBlending";
+import { rgbToHSL } from "./colorConverters";
+import { ColorHSL } from "./colorHSL";
+import { ColorInterpolationSpace, interpolateByColorSpace } from "./colorInterpolation";
+import { ColorRGBA64 } from "./colorRGBA64";
+import { ColorScale } from "./colorScale";
+
+export interface IColorPaletteConfig {
```
janechu: nit: can we remove the "I" prefix? It's suggested by typescript that interfaces do not start with "I" as the name of the interface should be axiomatic (and appears to be in this case).

packages/fast-colors/src/colorlib/colorPalette.ts
```diff
@@ -0,0 +1,271 @@
...
+        overlayDark: 0.25,
+        multiplyLight: 0,
+        multiplyDark: 0,
+    };
+
+    constructor(config: IColorPaletteConfig) {
+        this.config = {
+            baseColor: config.baseColor
+                ? config.baseColor
+                : ColorPalette.defaultPaletteConfig.baseColor,
```
janechu: nit: a slightly shorter way of doing all of these would just be to use `||` so `baseColor: config.baseColor || ColorPalette.defaultPaletteConfig.baseColor`, etc.

packages/fast-colors/src/colorlib/colorPalette.ts
```diff
@@ -0,0 +1,271 @@
...
+            this.config.overlayDark = newConfig.overlayDark;
+            changed = true;
+        }
+        if (
+            newConfig.multiplyLight &&
+            newConfig.multiplyLight !== this.config.multiplyLight
+        ) {
+            this.config.multiplyLight = newConfig.multiplyLight;
+            changed = true;
+        }
```
janechu: I wonder if these if conditionals could be consolidated into something like `checkColorPaletteConfigValue(ColorPaletteConfigPropertyKey.multiplyLight)` where ColorPaletteConfigPropertyKey is an enum like:
```
enum ColorPaletteConfigPropertyKey {
    multiplyLight = "multiplyLight"
    ...etc.
}
```

packages/fast-colors/src/colorlib/colorScale.ts
```diff
@@ -0,0 +1,203 @@
+// tslint:disable:member-ordering
+// tslint:disable:prefer-for-of
+// tslint:disable:interface-name
+
+import { contrastRatio } from "./colorConverters";
+import { ColorInterpolationSpace, interpolateByColorSpace } from "./colorInterpolation";
+import { ColorRGBA64 } from "./colorRGBA64";
+
+export interface IColorScaleStop {
```
janechu: nit: same here, think we could do away with the "I"

packages/fast-colors/src/colorlib/colorConverters.ts
janechu: @nicholasrice @eigenvector32 The other comments I've made are just nits, but this one I would like to address.

packages/fast-colors/src/colorlib/colorPalette.ts
nicholasrice: I think you can just do `this.config = Object.assign({}. ColorPalette.defaultPaletteConfig, config)` to make this even more simple. That will apply the static defaults and then clobber thosse values with config values if they exist.

packages/fast-colors/src/colorlib/colorScale.ts
nicholasrice: Yea good call - the reason we do this is because Interfaces and Type Aliases can be used interchangibly, but changing from one to the other can cause a breaking change if the naming convention is to use an `I` for all interface names. Not using an I to differentiate interfaces means we can change from an interface to a type alias as necessary without introducing naming changes.

packages/fast-colors/src/colorlib/colorLAB.ts
eigenvector32: I ended up using the rational values based on what I read here: http://brucelindbloom.com/index.html (his site does something odd with navigation that makes html links not go to sub pages, go to Info -> A (dis)continuity study of the CIE L* Function).

janechu: I see the section has "The CIE's decision to represent these constants as decimal values was an unnecessary approximation that introduces both function and slope discontinuities.", so looks good to me.

PR #1444, opened 2019-02-26
feat: update namespace component schema id's to prevent conflicts

### LINKED ISSUE #1441: Component schemas should be namespaced to prevent conflicts
Currently we simply identify schemas by id in multiple packages as follows:

Base button = `id: button`

MSFT button = `id: button`

This is going to create conflicts down the road somewhere - and it technically is creating issues within my own explorations. We should namespace the schema id's, possibly by matching our package names:

`@microsoft/fast-components-react-base/button`

`@microsoft/fast-components-react-msft/button`

chrisdholt:
closes #1441 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/card/card.schema.json
+++ b/packages/fast-components-react-base/src/card/card.schema.json
@@ -3,7 +3,7 @@
     "title": "Card",
     "description": "A card component's schema definition.",
     "type": "object",
-    "id": "card",
+    "id": "@microsoft/fast-components-react-base/card",
     "properties": {
         "tag": {
             "title": "HTML tag",
--- a/packages/fast-components-react-msft/src/number-field/number-field.schema.json
+++ b/packages/fast-components-react-msft/src/number-field/number-field.schema.json
@@ -1,0 +1,45 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Number field",
+    "description": "A number field component's schema definition.",
+    "type": "object",
+    "id": "@microsoft/fast-components-react-msft/number-field",
+    "properties": {
+        "min": {
+            "title": "HTML min attribute",
+            "type": "number"
+        },
+        "max": {
+            "title": "HTML max attribute",
+            "type": "number"
+        },
+        "step": {
+            "title": "HTML step attribute",
+            "type": "number"
+        },
+        "name": {
+            "title": "HTML name attribute",
+            "type": "string"
+        },
+        "placeholder": {
+            "title": "HTML placeholder attribute",
+            "type": "string"
+        },
+        "disabled": {
+            "title": "HTML disabled attribute",
+            "type": "boolean"
+        },
+        "readOnly": {
+            "title": "HTML readOnly attribute",
+            "type": "boolean"
+        },
+        "required": {
+            "title": "HTML required attribute",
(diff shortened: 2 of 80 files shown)


packages/fast-components-react-msft/src/text-action/examples.data.tsx
```diff
@@ -15,7 +16,7 @@ const examples: ComponentFactoryExample<TextActionProps> = {
     detailData: {
         placeholder: "Placeholder",
         button: {
-            id: "button",
+            id: buttonSchema.id,
```
chrisdholt: @janechu I'm having an issue with the plugin here; I've added the plugin for button in text action, but I'm finding that the children for the button is throwing an invariant violation. Additionally, even if I change this to a different render prop, it doesn't render as expected. Would love some help figuring this out. As it is - this is blocking this work from going in for v4.0

packages/fast-components-react-base/src/card/card.schema.json
```diff
@@ -3,7 +3,7 @@
     "title": "Card",
     "description": "A card component's schema definition.",
     "type": "object",
-    "id": "card",
```
nicholasrice: @chrisdholt @janechu For components that do not have MSFT versions (simply wrapped in the JSSManager), we re-use the "base" schema - is that going to be an issue?

chrisdholt: Ah - if we are trying to reference both a styled and unstyled instance within a tooling site, I think this would be an issue. This would likely pull in an unstyled base instance.

@janechu @nicholasrice what are your thoughts on managing duplicate schemas?

janechu: I think we should copy schemas over and provide duplicates with the precise ids for the package and we should run our examples through tv4 to validate our example data against the schemas, in case it ever goes out of date. Which incidentally, we should be doing regardless but this would help validate our APIs.

packages/fast-components-react-msft/src/text-action/examples.data.tsx
janechu: There is a PR up to resolve this issue see https://github.com/Microsoft/fast-dna/pull/1464

packages/fast-components-react-msft/src/number-field/number-field.schema.json
```diff
@@ -0,0 +1,45 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Number field",
+    "description": "A number field component's schema definition.",
+    "type": "object",
+    "id": "@microsoft/fast-components-react-base/number-field",
```
janechu: should be msft not base

packages/fast-components-react-msft/src/typography/typography.schema.json
```diff
@@ -0,0 +1,49 @@
+{
```
janechu: unrelated to this file, but there should be a copy schema task we can remove.

chrisdholt: Doesn't look like we were ever copying the schemas to the MSFT package from Base. Looks like we are all good on this - appears to just copy the files in src to dist.

PR #1446, opened 2019-02-27
feat: add focus features to ContextMenu

### LINKED ISSUE #1398: Context Menu should facilitate initial focus of items
Context menu currently provides no ease-of-use features to facilitate applying initial focus to a child context menu item.

**Given two core scenarios:**
1. context menu  is launched via right-click on an arbitrary element/context
2. context menu is launched via click on an element

How is document focus brought to the context menu?

**Current behavior:**
Context menu and context-menu-item provide no mechanism for setting initial focus when a context-menu is created. To achieve desired and accessible behavior, a reference to the menu-item must be obtained imperatively and through DOM API's, and the element focus manually via *HTMLElement.focus*

**Desired behavior:**
Context menu should provide a mechanism for moving focus to the context-menu-item when the menu is launched.

**Success Criteria:**
- Should bring focus to the first item when instructed
- Should function when visibility of the component is managed through CSS and the component is not unmounted
- Should function when visibility of the component is managed through conditional rendering (component is mounted and unmounted)
- ?

**Possible solutions:**
- public method on the component to focus the item
- boolean prop instructing the component to manage focus on mount
- ?

We should spend a little time to investigate potential solutions that meet the above success criteria and draft a proposal.

nicholasrice:
# Description
Adds a public function the `ContextMenu` to bring focus to the appropriate menu item. Also adds an opt-in prop (`enableFocusVisible`) to bring focus to invoke the focus method when the component mounts, which is generally desired. Making `enableFocusVisible` false by default ensures we don't change behavior of existing implementations of `ContextMenu`.

## Process & policy checklist

closes #1398

### DIFF
--- a/packages/fast-components-react-base/src/context-menu/context-menu.props.ts
+++ b/packages/fast-components-react-base/src/context-menu/context-menu.props.ts
@@ -12,6 +12,11 @@ export interface ContextMenuHandledProps extends ContextMenuManagedClasses {
      * The context menu children
      */
     children?: React.ReactNode;
+
+    /**
+     * When true, the context menu will immediately place focus on the appropriate menu-item when the context-menu mounts
+     */
+    enableAutoFocus?: boolean;
 }
 
 export type ContextMenuProps = ContextMenuHandledProps & ContextMenuUnhandledProps;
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-base/src/context-menu/context-menu.props.ts
```diff
@@ -12,6 +12,11 @@ export interface ContextMenuHandledProps extends ContextMenuManagedClasses {
      * The context menu children
      */
     children?: React.ReactNode;
+
+    /**
+     * When true, the context menu will immediately place focus on the appropriate menu-item when the context-menu mounts
```
nalogiudice: nit: if you update this, can you remove the hyphens between menu item and content menu, so it is consistent with the way we write it elsewhere?

PR #1448, opened 2019-02-27
feat: add class name for horizontal overflow items

### LINKED ISSUE #1428: Add managed JSS class to horizontal overflow content items
Add a JSS class to the content items generated by horizontal overflow's `getItems()`
https://github.com/Microsoft/fast-dna/blob/master/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx#L376

Easier/less implementation dependent styling of content items (e.g. margin between items).

`getItems()` adds a class (e.g. `horizontalOverflow_contentItem`) to the elements it generates.

Current workaround:
```js
{
    horizontalOverflow_contentRegion: {
        '& li': { /* ... */ }
    }
}
```

chrisdholt:
closes #1428 

# Description

Adds a className to provide more fine grain control over the list item being render on the wrapper of horizontal overflow children. 

*Note*: If the className is undefined, add style for the list item to ensure it renders with `display: inline-block`

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
@@ -377,7 +377,18 @@ class HorizontalOverflow extends Foundation<
         return React.Children.map(
             this.withoutSlot([ButtonDirection.previous, ButtonDirection.next]),
             (child: React.ReactNode): React.ReactElement<HTMLLIElement> => {
-                return <li style={{ display: "inline-block" }}>{child}</li>;
+                return (
+                    <li
+                        className={get(
+                            this.props.managedClasses,
+                            "horizontalOverflow_item",
+                            ""
+                        )}
+                        style={{ display: "inline-block" }}
+                    >
+                        {child}
+                    </li>
+                );
             }
         );
     }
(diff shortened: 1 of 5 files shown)


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -377,7 +377,20 @@ class HorizontalOverflow extends Foundation<
...
+                const itemClass: string = get(
+                    this.props.managedClasses,
+                    "horizontalOverflow_item",
+                    ""
+                );
+
+                return (
+                    <li
+                        className={itemClass}
+                        style={itemClass === "" ? { display: "inline-block" } : null}
```
nicholasrice: Presumably we would still want these to display inline-block though, wouldn't' we? Does the component work without setting this?

My guess is you're making this conditional because you don't want specificity issues, right?

chrisdholt: Exactly - There's an assumption here that would need to be documented that you are "taking active control" by passing a class. With inline-block here, you must include "!important" to change this.

nalogiudice: > that would need to be documented

Can we update the documentation then? It's not apparent why we are doing it this way from glancing at the code.

chrisdholt: The comment was from the initial draft PR form, which is why it was there and not as part of the PR. Regardless, this is outdated; ppl can use important to override the position.

PR #1455, opened 2019-02-27
feat: remove fast-application-utilities package

### LINKED ISSUE #1116: Deprecate fast-application-utilities package in favor of fast-web-utilities
Please deprecate `fast-application-utilities` package and move the functions into the `fast-web-utilities` package.

chrisdholt:
closes #1116 
# Description

## Issue type checklist

## Process & policy checklist


packages/fast-web-utilities/src/localization.ts
```diff
@@ -10,7 +10,7 @@ export const localeDirectionMapping: { [key: string]: Direction } = {
     en: Direction.ltr,
 };
 
-export default function isRTL(locale: string): boolean {
```
nicholasrice: are we using this anywhere? I'm not sure how useful this really is with this implementation, because `isRTL` doesn't actually tell us anything about the environment we're in - which is usually what we want to know

chrisdholt: no clue - @janechu can you advise on this?

janechu: We were using it in the dev-site way back, but I'm not sure its being used anywhere since then.

chrisdholt: @janechu @nicholasrice that's the only place I can find us using it; should we just move it in there only?

chrisdholt: I've moved this out

PR #1458, opened 2019-02-27
chore: update to improve generate className function of MSFT button

chrisdholt:
# Description

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/button/button.tsx
```diff
@@ -56,26 +56,14 @@ class Button extends Foundation<ButtonHandledProps, ButtonUnhandledProps, {}> {
...
-                );
-            default:
-                return super.generateClassNames();
-        }
+        const className: string = this.props.appearance
+            ? get(
+                  this.props,
+                  `managedClasses.button__${ButtonAppearance[this.props.appearance]}`
+              )
+            : "";
```
janechu: nit: It would be more accurate to use set this empty string as `undefined` and the typing can be `string | void`.

PR #1464, opened 2019-02-28
fix: update to allow nesting to occur within plugins and components when resolving executable data structures and resolves an issue with passing all children back to a plugin instead of resolving each child

janechu:
# Description

This allows any plugin or child to be resolved in any order for any data nesting configuration.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-data-utilities-react/src/index.ts
+++ b/packages/fast-data-utilities-react/src/index.ts
@@ -1,13 +1,13 @@
 import * as React from "react";
-import { cloneDeep, get, isPlainObject, set } from "lodash-es";
+import { cloneDeep, get, isPlainObject, set, uniqueId } from "lodash-es";
 import * as tv4 from "tv4";
 import Plugin, { PluginProps } from "./plugin";
 
 const pluginIdKeyword: string = "pluginId";
 const typeKeyword: string = "type";
-const childrenKeyword: string = "children";
 const squareBracketsRegex: RegExp = /\[(\d+?)\]/g;
 const propsKeyword: string = "props";
+const idKeyword: string = "id";
 
 export enum DataResolverType {
     plugin = "plugin",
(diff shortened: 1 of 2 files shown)


packages/fast-data-utilities-react/src/index.ts
```diff
@@ -675,110 +720,161 @@ export function mapDataToComponent(
...
+                        absoluteDataLocation: childDataLocation,
+                    };
                 }
+            ),
+            pluginModifiedDataLocations
+        )
+        .sort(orderMappedDataByDataLocation);
+
+    // going from the longest length to the shortest, resolve the data
+    mappedDataLocations.forEach(
```
nicholasrice: This was initially a little hard to follow for me because I wasn't sure why you were clobbering the same variable over and over.

A few things that might make this more clear:
1. chain this off of the `sort` method
2. use `reduce` instead of `forEach` - reduce coveys that you're "reducing" to a single value:

```
.reduce(
  (mappedData: any, mappedDataLocation: MappedDataLocation): any => {
      return resolveData(mappedDataLocation, mappedData, plugins, childOptions);
  },
  cloneDeep(data)
)
```

packages/fast-data-utilities-react/src/index.ts
```diff
@@ -675,110 +720,161 @@ export function mapDataToComponent(
...
+        const pluginResolverMapping: PluginResolverDataMap[] = getPluginResolverDataMap({
+            isReactChildren: pluginModifiedDataLocation.type === dataType.children,
+            pluginData,
+            pluginResolver,
+            childOptions,
+            dataLocation: pluginModifiedDataLocation.absoluteDataLocation,
+        });
 
-                set(mappedData, reactChildrenDataLocation, value);
+        pluginResolverMapping.forEach(
```
nicholasrice: You can also chain this off of `sort` - then you don't need to declare a variable.

Issue #1461, opened 2019-02-28
Create "focus" color recipe


nicholasrice:
Create color utility for focus color

PR #1468, opened 2019-03-01
feat: add neutral-focus recipe

nicholasrice:
# Description

Adds recipe for generating the neutral-focus color

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
@@ -1,0 +1,30 @@
+import { neutralForegroundRest } from "./neutral-foreground";
+import { palette, Palette, PaletteType, Swatch } from "./palette";
+import designSystemDefaults, { DesignSystem } from "../../design-system";
+import { neutralFocus } from "./neutral-focus";
+
+describe("neutralForeground", (): void => {
+    test("should return a string when invoked with an object", (): void => {
+        expect(typeof neutralFocus(designSystemDefaults)).toBe("string");
+    });
+
+    test("should return a function when invoked with a function", (): void => {
+        expect(typeof neutralFocus(() => "#FFF")).toBe("function");
+    });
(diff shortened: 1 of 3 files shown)


packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { neutralForegroundRest } from "./neutral-foreground";
+import { palette, Palette, PaletteType, Swatch } from "./palette";
+import designSystemDefaults, { DesignSystem } from "../../design-system";
+import { neutralFocus } from "./neutral-focus";
+
+describe("neutralForeground", (): void => {
```
chrisdholt: @nicholasrice assuming you mean "neutralFocus" here?

packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+    test("should return a string when invoked with an object", (): void => {
+        expect(typeof neutralFocus(designSystemDefaults)).toBe("string");
+    });
+
+    test("should return a function when invoked with a function", (): void => {
+        expect(typeof neutralFocus(() => "#FFF")).toBe("function");
+    });
+
+    test("should operate on default design system if no design system is supplied", (): void => {
+        expect(neutralFocus({} as DesignSystem)).toBe("#101010");
```
chrisdholt: Do we want this test to fail if we change the default design system value? Should we expect the empty object to match the same function with defaults passed?

nicholasrice: I think we want this to fail if we change the default design-system to something that makes this change, yea -IMO that would be a very significant change that we would want to know about if done inadvertently.

nicholasrice:
closes #1461

Issue #1454, opened 2019-02-27
Call to action schema is missing disabled definition


eljefe223:
**Describe the bug; what happened?**
Call to action schema is missing disabled definition

**What are the steps to reproduce the issue?**
Go to dev site. Select the `<CallToAction/>` component. There should be a way to set disabled in the config panel

Example:

1. Go to Dev Site
2. Select the `<CallToAction/>` 
3. There should be a way to set disabled in the config panel

PR #1477, opened 2019-03-01
fix: update call to action to handle disabled state correctly

nalogiudice:
Closes [#1454](https://github.com/Microsoft/fast-dna/issues/1454)

# Description
Fixes a few issues with the disabled CTA

- now shows up in editor
- now does not transform on hover
- removes href when disabled so it actually gets the `disabled` attribute is no longer actionable (before it was setting `aria-disabled` and was still actionable)

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
+++ b/packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
@@ -33,7 +33,7 @@ class CallToAction extends Foundation<
                 {...this.unhandledProps()}
                 className={this.generateClassNames()}
                 disabled={this.props.disabled}
-                href={this.props.disabled ? null : this.props.href}
+                href={this.props.href}
                 appearance={
                     ButtonAppearance[CallToActionAppearance[this.props.appearance]]
                 }
--- a/packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
+++ b/packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
@@ -33,7 +33,7 @@ class CallToAction extends Foundation<
                 {...this.unhandledProps()}
                 className={this.generateClassNames()}
                 disabled={this.props.disabled}
-                href={this.props.href}
+                href={this.props.disabled ? null : this.props.href}
                 appearance={
                     ButtonAppearance[CallToActionAppearance[this.props.appearance]]
                 }
--- a/packages/fast-components-react-msft/src/call-to-action/call-to-action.schema.json
+++ b/packages/fast-components-react-msft/src/call-to-action/call-to-action.schema.json
@@ -18,6 +18,10 @@
                 "justified",
                 "lightweight"
             ]
+        },
+        "disabled": {
+            "title": "Disabled",
+            "type": "boolean"
         }
     },
     "reactProperties": {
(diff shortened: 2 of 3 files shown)


packages/fast-components-react-msft/src/call-to-action/call-to-action.tsx
```diff
@@ -33,7 +33,7 @@ class CallToAction extends Foundation<
                 {...this.unhandledProps()}
                 className={this.generateClassNames()}
                 disabled={this.props.disabled}
-                href={this.props.href}
+                href={this.props.disabled ? null : this.props.href}
```
nicholasrice: I think we should always pass `href` to `Button` - behavior between this component and `Button` should be consistent (regarding disabled), so I would just let `Button` handle the semantics/accessibility requirements.

chrisdholt: I agree with this - I don't think this additional logic is necessary.

nalogiudice: @nicholasrice Ok, so the button should handle the logic of removing the href if disabled is true?

Currently, if you set an href on a button and then change it to disabled, it adds `aria-disabled=true` and styles the button as disabled, BUT, it is still actionable. This was the issue I was addressing in CTA.

So, if I apply the `href={this.props.disabled ? null : this.props.href}` to button, it would solve this.

chrisdholt: A disabled anchor is essentially one without an href. If someone passes an href and disabled, I think we want to pass it. Aria by itself doesn't necessarily prevent interaction. This seems to be an issue with styling; we can choose to style it appropriately (cursor behavior) to prevent it being actionable. 

CTA is essentially a style abstraction here; I'm not sure we want to start black boxing behavior such as this to solve an issue that arises from prop combinations someone chooses to pass.

nalogiudice: > This seems to be an issue with styling; we can choose to style it appropriately (cursor behavior) to prevent it being actionable.

Can you explain this a little more? Currently, button _does_ have the correct styling when disabled and an anchor - the shape is `<a href="#" aria-disabled="true">foo</a>`, BUT it is still actionable (as in, even though it looks like you cannot click it, you can).

I'm fine with saying if a user passes both `disabled` AND an` href`, we will write the shape `<a href="#" aria-disabled="true">foo</a>`, and the user can choose to do whatever. But, out of the box, interaction is not blocked in this case.

So if we are ok with that, I will revert the logic to remove href, and we will just say things that get both disabled and an href will be styled appropriately, but not actually "disabled"

chrisdholt: I show these both as being actionable and navigating to a page. If you change the actual destination on button from "#" to something like www.microsoft.com, that will navigate you even when disabled

nalogiudice: Right, that's exactly what I am saying. Because, if either button or CTA has an 'href' prop, they will get `aria-disabled`, but not `disabled` (which actually disables it).

**So**, we are saying the same thing, I just want to make sure this is the _expected_ behavior - in which case, the PR has been updated appropriately.

chrisdholt: Yeah, I think this is expected. `disabled` is valid for a button, not an href (thus aria-disabled). I think here we are apply correct attribute and semantics for the element (aria-disabled vs disabled) and applying a disabled style.

nalogiudice: Ok - this is good to go then. Thanks.

Issue #1382, opened 2019-02-14
add and integrate typedoc with docusaurus


awentzel:
**What feature are you requesting?**
Creation and integration of TypeDoc generated API's with Docusaurus site. 

**What would be the benefit of adding this feature?**
[Documentation site](https://microsoft.github.io/fast-dna/) to seamlessly integrate Marketing + Development + API documentation all under one domain.

**What solution would be ideal for you?**
When looking at any package page, [for example](https://microsoft.github.io/fast-dna/docs/en/packages/fast-components-react-msft/), there would be an additional section displayed as a link named API `/api` which would contain the complete API documentation subsite generated by TypeDoc and be accessible as a link from the existing readme.md file.  As shown:
![screen shot 2019-02-14 at 11 57 07 am](https://user-images.githubusercontent.com/16669785/52810385-adec3980-304f-11e9-8b93-727744489047.png)

**Is there any additional context for your request?**
A couple challenges exist:
1. Every time docusuaurus publishes all contents are regenerated in `website/build/fast-dna` so anything placed in there from earlier scripted routines will be deleted.
2. What is required from GH Pages to ensure that all .html files can be linked and executed as a single site even though it's really a bunch of subsites.  There may be a solution [here](https://docusaurus.io/docs/en/custom-pages)
3. Getting the proper status result after spawning TypeDoc commands.

**Attach screenshots or images add detail to your request:**

**Requirements**
-  Default TypeDoc theme is acceptable
- TypeDoc should compile to html
- TypeDoc can copy files into the `./website/build` folder which is gitignored
- Website/build/fast-dna/docs/en/packages/
  - <del>Fast-animation</del>
  - Fast-application-utilities
  - Fast-components-*
    - Api/*
    - Index.html
    - etc
- Package readme.md should reference a link to the API documentation index.html page
- API site should render on gh-pages as seamless integration with Docusaurus generated pages
- Site can look and feel independently of the Guide and Marketing pages.
- Several packages are ignored in script and this is ok since they require refactor later to bring into 
- The package.json script section should include a new script "docs:publish" which will execute the following:
  - Copy-package-readme.js
  - Generate-typedocs.js
  - Integrate typedocs with docusaurus

**Deployment**
At which point `./website/build/fast-dna` would contain the entire set of documentation (Marketing and Technical) and can be bundled and deployed to a single Azure App Service for the domain https://fastdna.net.

PR #1481, opened 2019-03-02
chore: add script for typedoc docusaurus integration

jkmdev:
# Description

Implements #1382
Creates a script that copies package `readme.md` files into `docs/en/packages`. Then uses typedoc to generate Typedoc API documentation that are also added to `docs/en/packages`. Converts these api files into markdown using the `markdown` theme from [typedoc-plugin-markdown](https://github.com/tgreyuk/typedoc-plugin-markdown). Also contains a dry-run version for the script.

To run the dry run:

Run `npm run docs:build:dry-run` in root

To run the regular process:

Run `npm install` to install the new `typedoc` and `typedoc-plugin-markdown` packages
Run `npm run docs:build:api` in root
`cd website`
Run `npm run start`
Navigate to the relevant package readme pages, scroll to the bottom, there should be a link to the related typeoc api doc there

When `npm run build` is run in the `website` folder these api docs will be in the `./website/build/fast-dna/docs/en/packages` folder. 

## Motivation & context

This script automates the documentation generation process so that `./website/build/fast-dna` contains the entire set of documentation. 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/build/documentation/generate-typedocs.js
+++ b/build/documentation/generate-typedocs.js
@@ -1,0 +1,202 @@
+/**
+ * Utility for generating Typedoc API documentation and placing it in the docs/en/packages folder so it can be hosted by
+ * Github pages and viewed through Docusaurus
+ * 
+ * Usage :run node build/documentation/generate-typedocs.js OR
+ * Usage (dry-run): run node build/documentation/generate-typedocs.js --dry-run
+ */
+
+const path = require("path");
+const fs = require("fs");
+const glob = require("glob");
+const os = require('os');
+const chalk = require('chalk');
+
+const { exec } = require('child_process');
+const { spawn } = require("child_process");
+
+
+const rootDir = path.resolve(process.cwd());
+const srcDir = "packages/*";
+const destDir = path.join("docs", "en", "packages");
+
+var copyReadmeScript = 'copy-package-readme.js';
+var copyReadmeScriptPath = path.join('build', 'documentation', copyReadmeScript);
+var dryRun = false;
+
+const excludedPackages = [
+    "fast-browser-extensions",
+    "fast-development-site-react",
+    "fast-permutator",
+    "fast-tslint-rules"
+];
+
+/**
+ * Determine if a dry run will be executed based off --dry-run argument being present
+ * if an invalid third parameter is entered, the application will exit
+ */
+process.argv.forEach(function (val, index) {
+
+    var validArg = true;
+
+    if (index === 2) {
+        val === "--dry-run" ? dryRun = true : validArg = false;
+    }
+
+    if (!validArg) {
+        console.log(chalk.red(`Invalid argument used. To perform a dry-run use --dry-run`));
+        process.exit(1);
+    }
+
+});
+
+/**
+ * Run the copy readme script, then this one if successful
+ * To prevent new links constantly being appended to the readme every 
+ * time the script is run, the copy readme script is run to regenerate 
+ * files to their original form with the needed header
+ */
+
+var proc = dryRun ? exec(`node ${copyReadmeScriptPath} --dry-run`) : exec(`node ${copyReadmeScriptPath}`);
+
+proc.stdout.on('data', function(data) {
+    process.stdout.write(data);
+});
+
+proc.on('error', (err) => {
+    console.log(chalk.red(err));
+});
+
+proc.on('close', function(code) {
+    console.log(chalk.green(`${copyReadmeScript} ran successfully.`));
+    execute();
+});
+
+/**
+ * Generate TypeDocs for each package
+ */
+function execute() {
+
+    if (dryRun) {
+        console.log(`In ${destDir}, this script would...`);
+    } else {
+        console.log(`Generating API documentation using TypeDoc...`);
+    }
+
+    const packages = path.resolve(rootDir, srcDir);
+
+    glob(packages, {realpath:true}, function(error, srcFiles) {
+       
+        srcFiles.forEach((srcPath) => {    
+            
+            var valid = true;
+
+            var packageName = srcPath
+                .split(path.sep)
+                .pop();
+
+            excludedPackages.forEach((excludedPackageName) => {
+                if (packageName === excludedPackageName) {
+                    console.log(chalk.yellow(`...skip generating Typedoc API files for ${packageName}`));
+                    valid = false;
+                }
+            });
+
+            if(valid) {
+                dryRun ? console.log(`...generate Typedoc API files for ${packageName}`) : createAPIFiles(packageName, srcPath);
+                dryRun ? console.log(`...add API link to ${packageName}`) : addAPILinkToReadme(packageName);
+            }
+
+        });
+
+    });
+}
+
+/**
+ * Uses Typedoc process to generate API docs for a given package
+ * Stores them in the docs/en/packages folder 
+ */
+function createAPIFiles(packageName, srcPath) {
+    
+    var config = path.join(srcPath, '/tsconfig.json');
+    var output = path.join(destDir, packageName, 'api');
+    var file = (packageName === "fast-animation") ? path.join(srcPath, '/lib/index.ts') : path.join(srcPath, '/src/index.ts');
+
+    var typedocCmd = os.platform() == 'win32' ? 'typedoc.cmd' : 'typedoc';
+    var typedocPath = path.join('node_modules', '.bin', typedocCmd);
+
+    const typedoc = spawn(typedocPath, // the local typedoc installation is needed for the markdown theme to work
+        [
+            '--tsconfig', config,
+            '--excludeNotExported',
+            '--out', output, file,
+            '--theme', "markdown"
+        ]);
+
+    typedoc.on('close', code => {
+        if (code) {
+            console.log(chalk.red(`${packageName} - typedoc API docs not generated, error code: ${code}`));
+        } else {
+            console.log(`${packageName} - typedoc API docs generated`)
+            addHeaderToReadme(packageName);
+        }
+    });
+
+    // sometimes important errors will get sent to stdout for this process
+    // leaving this here because it's useful for debugging, but generates a lot
+    // of output, makes the terminal output harder to read so it's commented out
+
+    // typedoc.stdout.on('data', (data) => {
+    //     console.error(`${data}`);
+    // });
+
+    typedoc.on('error', (err) => {
+        console.log(chalk.red(err));
+    });
+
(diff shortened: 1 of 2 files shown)


build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
+/**
+ * Utility for generating Typedoc API documentation and placing it in the docs/en/packages folder so it can be hosted by
+ * Guthub pages and viewed through Docusaurus
```
awentzel: Spelling "Guthub"

package.json
```diff
@@ -18,6 +18,8 @@
   "scripts": {
     "docs:build": "node build/documentation/copy-package-readme.js",
     "docs:dry-run": "node build/documentation/copy-package-readme.js --dry-run",
+    "docs:publish": "node build/documentation/generate-typedocs.js",
```
awentzel: "build" generates readme's and "publish" generates typedocs.  I think what i'd like to see is "docs:build" that compiled all of the documentation (readmes, apis, etc).  Each of the parts also have scripts of their own when wanting to just test readmes or api for example. Publish we should add later for the actual uploading to gh-pages.

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
...
+    var readmePath = path.join(destDir, packageName, 'README.md');
+    var apiLink = "api";
+
+    var usageText = 
+        "\n" + 
+        `[API Reference](${apiLink})`;
+
+    fs.appendFile(readmePath, usageText, function (err) {
+        if (err) {
+            console.log('\x1b[31m%s\x1b[0m', err);
```
awentzel: For all errors, could you use make them red when output to CLI?

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
...
+    var docusaurusHeader = 
+    `---\n` + 
+    `id: index\n` +
+    `---\n\n`;
+
+    try {
+        fs.writeFileSync(readmePath, docusaurusHeader);
+        fs.appendFileSync(readmePath, readmeText);
+    } catch (err) {
+        console.log('\x1b[31m%s\x1b[0m', err);
```
awentzel: maybe add a small comment on what the different regular expressions are doing?

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
...
+            console.log(`${packageName} - typedoc API docs generated`)
+            addHeaderToReadme(packageName);
+        }
+    });
+
+    // sometimes important errors will get sent to stdout for this process
+    // leaving this here because it's useful for debugging, but generates a lot
+    // of output, makes the terminal output harder to read so it's commented out
+
+    // typedoc.stdout.on('data', (data) => {
```
awentzel: I wonder if we want to add a future task to add an additional flag like `--verbose` that would then run this extra output useful for debugging?  Thoughts?

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
...
+proc.stdout.on('data', function(data) {
+    process.stdout.write(data);
+});
+
+proc.on('error', (err) => {
+    console.log(`\x1b[31m%s\x1b[0m`, err);
+});
+
+proc.on('close', function(code) {
+    console.log(`\x1b[32m%s\x1b[0m`, `copy-package-readme.js completed successfully.`);
```
awentzel: nit: What happens if this file name changes?  could this be replaced with more generic notation?

build/documentation/generate-typedocs.js
jkmdev: Yah I get what you mean, it's close to impossible to see what's happening there. I was thinking of maybe using [chalk](https://github.com/chalk/chalk), since it would add color to the terminal output in a much easier to read manner, but I'm not sure what people would think about me adding another package for something fairly small. Does it matter, or should I just add a small comment describing what's happening?

package.json
jkmdev: So in that case I should maybe do something like:

`"docs:build": build/documentation/generate-typedocs.js` -> runs `generate-typedocs` script (which generates all docs)
`"docs:build:readme": node build/documentation/copy-package-readme.js`-> only build readme files with `copy-package-readme` script
`"docs:build:typedoc": node build/documentation/generate-typedocs.js` -> meant to only build api docs, however `generate-typedocs` currently builds all documentation (it rebuilds the existing `doc/en/package` readme files to prevent appending multiple versions of the api doc link them). So this one might be misleading unless the readme and api doc process were better separated.

I have an new idea for how we could separate the typedoc and readme process while avoiding this appending issue, gonna message you in the discord about it though.

build/documentation/generate-typedocs.js
jkmdev: I think it's a good idea, but I also think it'd have to be made clear that this option is for making it easier to debug the typedoc process when it generates an error. So it could maybe be mentioned 
like this:

```
if (verbose) { // use --verbose if you're having trouble debugging the typedoc process
    typedoc.stdout.on('data', (data) => {
        console.error(`typedoc stdout:\n${data}`);
    });
}
```

package.json
awentzel: Just in case we swap out typedoc one day I may call it "docs:build:api" to limit impact if we change tech in the future which then would require CI/CD changes.

At might be good to separate out typedocs vs readmes. @nicholasrice thoughts on what he thinks.

build/documentation/generate-typedocs.js
awentzel: I'm personally ok with taking this small dependency at the root of the project for sake of better logging.

build/documentation/generate-typedocs.js
awentzel: This works for me. What is the likely hood of there being issues?  Is this a fairly delicate task?

jkmdev: I think its should be alright, and fairly easy to implement. My only concern is that if there's a lot of places in the code where we're using this `if (verbose)` conditional, it would make the code a bit more difficult to read. But if it's just one or two spots, I think it should be fine.

build/documentation/generate-typedocs.js
jkmdev: Okay, gonna make use of it in that case.

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
...
+
+const rootDir = path.resolve(process.cwd());
+const srcDir = "packages/*";
+const destDir = path.join("docs", "en", "packages");
+
+var copyReadmeScriptPath = path.join('build', 'documentation', 'copy-package-readme.js');
+var dryRun = false;
+
+const excludedPackages = [
+    "fast-animation",
```
chrisdholt: Is there a reason we're excluding fast-animation? I think we absolutely want to generate typedocs for the animation package.

build/documentation/generate-typedocs.js
chrisdholt: you could also just interpolate the file name here

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,198 @@
...
+        srcFiles.forEach((srcPath) => {    
+            
+            var valid = true;
+
+            var packageName = srcPath
+                .split(path.sep)
+                .pop();
+
+            excludedPackages.forEach((excludedPackageName) => {
+                if (packageName == excludedPackageName) {
```
chrisdholt: Curious - are you intentionally not using strict equality here?

build/documentation/generate-typedocs.js
awentzel: It was a mistake in the requirements which I just updated.

awentzel: @chrisdholt are there any other packages we should exclude that you can think of?

chrisdholt: not initially - I think the four below this one make sense to keep undocumented for now.

build/documentation/generate-typedocs.js
awentzel: You have experience running and troubleshooting this script so I'll leave it up to you if you think this is useful or not and where it should be added. I'm also, ok with breaking this task out into a new issue and wiring it up later.

build/documentation/generate-typedocs.js
jkmdev: No, I'll have to fix it

build/documentation/generate-typedocs.js
jkmdev: Alright, I decided to leave it as is for now to make these changes easier to test. I'll add it in a consecutive PR.

package.json
```diff
@@ -16,8 +16,10 @@
     "url": "https://github.com/Microsoft/fast-dna/issues/new/choose"
   },
   "scripts": {
-    "docs:build": "node build/documentation/copy-package-readme.js",
-    "docs:dry-run": "node build/documentation/copy-package-readme.js --dry-run",
+    "docs:build": "node build/documentation/generate-typedocs.js",
```
chrisdholt: @jkmdev @awentzel my assumption is that build should do the full build for all documentation. Is that reflective of this update?

jkmdev: In terms of the package readme files and typedoc api docs (Marketing and Technical documentation), that line will generate both those things. I'm fairly sure that's everything, but it might be best to see what awenzel says just in case.

build/documentation/generate-typedocs.js
```diff
@@ -0,0 +1,201 @@
...
+    var config = path.join(srcPath, '/tsconfig.json');
+    var output = path.join(destDir, packageName, 'api');
+    var file = (packageName === "fast-animation") ? path.join(srcPath, '/lib/index.ts') : path.join(srcPath, '/src/index.ts');
+
+    var typedocCmd = os.platform() == 'win32' ? 'typedoc.cmd' : 'typedoc';
+    var typedocPath = path.join('node_modules', '.bin', typedocCmd);
+
+    const typedoc = spawn(typedocPath, // the local typedoc installation is needed for the markdown theme to work
+        [
+            '--tsconfig', config,
```
nicholasrice: Can you add the `--excludeNotExported` flag also?

jkmdev: Alright, I've added it

nicholasrice: thanks!

awentzel:
Talked with @chrisdholt he's ok with these changes so I'm squash/merging.

PR #1484, opened 2019-03-04
feat: add new data generator and consolidate some files into more meaningful naming structures

janechu:
# Description

This adds the example data generation from a JSON schema to the new tooling package.

## Motivation & context

This consolidates more data manipulation that was coming from the permutator which will be deprecated/replaced at a later date. It also adds in the functionality to generate example data for react children.

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/data-utilities/generate.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+        expect(getDataFromSchema(schema)).toBe(null);
+    });
+    test("should return an enum specified value", () => {
+        const schema: any = {
+            enum: ["foo", "bar", "bat"],
+        };
+
+        expect(getDataFromSchema(schema)).toBe("foo");
+    });
+    test("should return default even if enums are specified", () => {
```
nicholasrice: What happens if "default" is not in the enum?

packages/fast-tooling-react/src/data-utilities/generate.ts
```diff
@@ -0,0 +1,137 @@
...
+                    case PropertyKeyword.properties:
+                        exampleData[requiredItem] = getDataFromSchema(
+                            schema[propertyType][requiredItem]
+                        );
+                        break;
+                    case PropertyKeyword.reactProperties:
+                        exampleData[requiredItem] = getReactChildrenFromSchema(
+                            schema[propertyType][requiredItem],
+                            childOptions
+                        );
```
nicholasrice: did you not add a `break;` here intentionally?

packages/fast-tooling-react/src/data-utilities/generate.spec.ts
janechu: Theoretically then the default value would not validate against the schema, which is a little ambiguous in the way since it "should" valudate. It is indicated that the default value should be used to generate a key/value pair per https://json-schema.org/understanding-json-schema/reference/generic.html

packages/fast-tooling-react/src/data-utilities/generate.ts
janechu: Nope, thanks for catching that!

PR #1485, opened 2019-03-05
feat: remove appearance prop and add filled prop to control backplate usage

nalogiudice:
Closes [#1462](https://github.com/Microsoft/fast-dna/issues/1452)

# Description
Introduces `filled` prop. If `filled` is false, the badge will take the non-backplate styling. If `filled` is true, the yellow backplate will be applied with a text color of accent-foreground-cut.

Removes old `appearance` API.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/badge/README.md
+++ b/packages/fast-components-react-msft/src/badge/README.md
@@ -8,4 +8,4 @@ Keep the *badge* label short and succinct &ndash; it should not be a sentence or
 
 ## Style guidance
 
-Use the small or large size best suited for the specific *badge* usage. The `filled` prop adds a backplate to the badge with default styling. Additional styling can be modified by applying CSS `color` and `background` using the `jssStyleSheet` override.
+Use the small or large size best suited for the specific *badge* usage. The `filled` prop adds a backplate to the badge with default styling. Additional styling can be modified by applying CSS `color` and `background` using the `jssStyleSheet` prop.
--- a/packages/fast-components-react-msft/src/badge/README.md
+++ b/packages/fast-components-react-msft/src/badge/README.md
@@ -8,4 +8,4 @@ Keep the *badge* label short and succinct &ndash; it should not be a sentence or
 
 ## Style guidance
 
-Use the appearance and size best suited for the specific *badge* usage.
+Use the small or large size best suited for the specific *badge* usage. The `filled` prop adds a backplate to the badge with default styling. Additional styling can be modified by applying CSS `color` and `background` using the `jssStyleSheet` override.
--- a/packages/fast-components-react-msft/src/badge/badge.schema.json
+++ b/packages/fast-components-react-msft/src/badge/badge.schema.json
@@ -5,15 +5,9 @@
     "type": "object",
     "id": "@microsoft/fast-components-react-msft/badge",
     "properties": {
-        "appearance": {
-            "title": "Appearance",
-            "type": "string",
-            "default": "lowlight",
-            "enum": [
-                "lowlight",
-                "highlight",
-                "accent"
-            ]
+        "filled": {
+            "title": "Backplate",
+            "type": "boolean"
         },
         "size": {
             "title": "Size",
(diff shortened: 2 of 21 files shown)


chrisdholt:
@nalogiudice can we be more explicit about *removing* the appearance prop here. We don't have a PR to deprecate those values, so ideally I think we should call that out in the PR title

nalogiudice:
@chrisdholt is that clearer?

packages/fast-components-react-msft/src/badge/README.md
```diff
@@ -8,4 +8,4 @@ Keep the *badge* label short and succinct &ndash; it should not be a sentence or
 
 ## Style guidance
 
-Use the appearance and size best suited for the specific *badge* usage.
+Use the small or large size best suited for the specific *badge* usage. The `filled` prop adds a backplate to the badge with default styling. Additional styling can be modified by applying CSS `color` and `background` using the `jssStyleSheet` override.
```
chrisdholt: nit: instead of "override", can we say, "jssStyleSheet prop" as it's technically a prop.

packages/fast-components-react-msft/src/badge/badge.schema.json
```diff
@@ -5,15 +5,9 @@
...
-            "title": "Appearance",
-            "type": "string",
-            "default": "lowlight",
-            "enum": [
-                "lowlight",
-                "highlight",
-                "accent"
-            ]
+        "filled": {
+            "title": "Backplate",
```
chrisdholt: I have a bit of a disconnect between having a prop named "filled" and a title of "backplate". Do we care that the API is not accurately represented for design and developers here? If a designer mentions the backplate prop, will that cause confusion or require "special knowledge" that these map 1:1?

packages/fast-components-react-msft/src/badge/examples.data.tsx
```diff
@@ -11,47 +11,26 @@ export default {
...
         {
-            children: "Badge default small",
+            children: "Filled badge small",
+            filled: true,
             size: BadgeSize.small,
-        },
-        {
-            children: "Badge default large",
-            size: BadgeSize.large,
         } as any,
```
chrisdholt: We don't have any render props, do we need to coerce as any here?

packages/fast-components-react-msft/src/badge/badge.schema.json
nalogiudice: @nicholasrice suggested `filled`, but in the editor view `backplate` made more sense to me. I understand the 1:1 concern.

My vote would be for `backplate` over `filled` - `filled` makes me ask "Filled what?"

chrisdholt: Pinging @nicholasrice who proposed `filled` - my assumption is that it maps to the v4 color work being done, but he can advise on the prop naming.

nicholasrice: v4 color work uses the term "fill" to describe recipes that should create component "backplates".

I think our use of the terms are relativly interchangable, but "fill" does have a long history in design vernacular to describe filling a shape with color. I'm uncertain how broadly "backplate" is used, but I have personally never seen it in design tools.

nicholasrice: @bheston @Falkicon My have thoughts here - I'm kinda impartial aside from that this name maps to *current* color recipe terms (which could change if we want)

bheston: I don't think we have a precedence for something quite like this yet, but it's similar to the appearance property we have on button. One potential option would be to flip it. I think the default state for a badge should have a backplate, but understand that News has a particular usage of it where it's essentially just a single line of short text. Close to a typical badge, but questionable. Between the two I would suggest "filled" as well since we internally refer to things as backplates, but don't publish that anywhere. Another option that might be more descriptive would be something like "Contained" or "Framed", which adds the implication of the padding adjustment that's a key portion of the filled background. Conceivably you could have a "contained" badge with transparent fill, which would be like our lightweight button.

So, my suggestion would be "Filled" (default to true) or "Contained".

nalogiudice: I'm good with both changing to `filled` and defaulting to true. Will update.

PR #1499, opened 2019-03-07
chore: update website title

awentzel:
# Description
Updates title to reflect the actual intended title.
 
## Motivation & context
The title that displays on browser tab inaccurately described what the site is.  This is the "Component explorer".

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/webpack.config.js
+++ b/packages/fast-components-react-msft/webpack.config.js
@@ -62,7 +62,7 @@ module.exports = (env, args) => {
         },
         plugins: [
             new HtmlWebpackPlugin({
-                title: "FAST documentation",
+                title: "FAST Component explorer",
                 contentBase: outDir,
             }),
             new WebpackShellPlugin({


packages/fast-components-react-msft/webpack.config.js
```diff
@@ -62,7 +62,7 @@ module.exports = (env, args) => {
         },
         plugins: [
             new HtmlWebpackPlugin({
-                title: "FAST documentation",
+                title: "FAST Component explorer",
```
chrisdholt: Should we update this on the docs site as well. To get here I simply click "components". Perhaps we either update the button to say "component explorer", or we update this to just say, "FAST Components"?

awentzel: When we talk about it, it's the "FAST Component explorer" so I think this makes sense.  But, I can update the documentation site on https://microsoft.github.io/fast-dna to say "Component explorer" because on that site it's already in the context of FAST since it's a documentation site.

Issue #111, opened 2018-03-20
Rename fast-development-site to fast-react-development-site


chrisdholt:
### Expected behavior 
The package should denote that it is for react projects.

### Actual behavior
The package appears to work in all environments

### Specifications (hardware, software, configurations)
* Include device make, model, OS version: 
* Include browser and version: 
* Include screen orientation: 
* Include screen size: 
* Include page URL: 
* Include localization: 

*Please attach screenshots or video of the problem for quickest resolution*

PR #1501, opened 2019-03-07
feat: apply new color system to carousel

eljefe223:
# Description

Add new color system to `<Carousel />` component. Also updated flipper to have a 80% opacity on the background element only. This is now the design for `<Flipper />` moving forward. 

Shown in these screenshots below:

darktheme

lighttheme

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/carousel/index.ts
+++ b/packages/fast-components-styles-msft/src/carousel/index.ts
@@ -5,7 +5,12 @@ import {
     CSSRules,
 } from "@microsoft/fast-jss-manager";
 import { applyLocalizedProperty, Direction } from "@microsoft/fast-jss-utilities";
-import { hoverContrast } from "../utilities/colors";
+import {
+    neutralFillStealthHover,
+    neutralFillStealthRest,
+    neutralForegroundRest,
+    neutralOutlineRest,
+} from "../utilities/color";
 import { CarouselClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 
 const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
@@ -13,10 +18,8 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
 ): ComponentStyleSheet<CarouselClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
     const direction: Direction = designSystem.direction;
-    const white: string = "white";
-    const darkGray: string = "#101010";
-    const lightGray: string = "rgb(206, 206, 206)";
-    const lightGrayRest: string = "rgba(206, 206, 206, 0.6";
+    const white: string = "#FFF";
+    const black: string = "#111";
 
     function flipperStyles(): CSSRules<{}> {
         return {
@@ -66,15 +69,21 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
                 display: "block",
                 height: "4px",
                 width: "32px",
-                transition: "all 0.2s ease-in-out",
+                transition: "all 0.05s ease-in-out",
             },
-            "&:hover": {
-                "&::before": {
-                    opacity: "0.5",
+            "&:not($carousel_sequenceIndicator__active)": {
+                "&:hover": {
+                    "&::before": {
+                        opacity: "0.5",
+                    },
                 },
             },
         },
-        carousel_sequenceIndicator__active: {},
+        carousel_sequenceIndicator__active: {
+            "&::before": {
+                opacity: "1",
+            },
+        },
         carousel_tabPanel: {
             display: "block",
         },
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -5,18 +5,21 @@ import {
...
     config: DesignSystem
 ): ComponentStyleSheet<CarouselClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
     const direction: Direction = designSystem.direction;
-    const white: string = "white";
-    const darkGray: string = "#101010";
-    const lightGray: string = "rgb(206, 206, 206)";
-    const lightGrayRest: string = "rgba(206, 206, 206, 0.6";
+    const white: string = "#FFF";
+    const black: string = "#111";
```
nicholasrice: Why `#111`? Is this tied to foreground?

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -93,59 +102,65 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
...
             "& $carousel_flipperPrevious, & $carousel_flipperNext": {
-                color: darkGray,
-                fill: darkGray,
-                background: "rgba(255, 255, 255, 0.6)",
-                border: `1px solid ${lightGrayRest}`,
+                "&::before": {
+                    color: neutralForegroundRest((): string => black),
+                    fill: neutralForegroundRest((): string => black),
+                    background: neutralFillStealthRest((): string => black),
+                    border: `1px solid ${neutralOutlineRest((): string => black)}`,
```
bheston: I'm wondering about these default values. Is there a case where the design system may not provide values? If so, seems like we would want the fallback universally, not in each case. Also, everything can't be the same color or it would just be a black circle.

packages/fast-components-styles-msft/src/flipper/index.ts
```diff
@@ -34,19 +34,34 @@ const styles: ComponentStyles<FlipperClassNameContract, DesignSystem> = (
...
             padding: "0",
-            "&:hover": {
-                background: neutralFillStealthHover,
-                color: neutralForegroundHover,
+            "&::before": {
+                content: "''",
+                top: "-1px",
+                right: "-1px",
+                bottom: "-1px",
+                left: "-1px",
```
bheston: Can't quite tell without looking at the dom, but does this assume the border is 1px? Should it use the design system value, or any way to get away without needing this offset?

packages/fast-components-styles-msft/src/carousel/index.ts
bheston: Seems like we'd want to avoid hard-coding this, or maybe this is what Chris mentioned the other day about needing black and white constants globally. If we must, please set it to #101010 for now.

chrisdholt: @nicholasrice didn't we add constants for black and white in the v4 color work?

eljefe223: Filed an issue here: https://github.com/Microsoft/fast-dna/issues/1502

packages/fast-components-styles-msft/src/flipper/index.ts
chrisdholt: this does seem odd - what's this accomplishing?

eljefe223: its aligning the Pseudo element with the button container. Without this the button is 40x40 and the Pseudo is 38x38 even with changing the box model. We need to isolate the background so we can apply an 80% transparency and not effect the other elements in `<Flipper />`

chrisdholt: As Brian mentioned though, I think this has to do with the border which makes assumptions since that maps to a design system value. Can we explore a better solution here? I think using a pseudo element is fine here, but the code itself fails if we change the outlinePatternOutlineWidth.

packages/fast-components-styles-msft/src/carousel/index.ts
eljefe223: I'm not following. Black is passed a reference color to keep the Dark Themed slide flippers always dark. We are three different recipes here `neutralForegroundRest` `neutralFillStealthRest` and `neutralOutlineRest`

nicholasrice: This code merges the provided `DesignSystem` with a default `DesignSystem`, so these values will always exist.

packages/fast-components-styles-msft/src/flipper/index.ts
eljefe223: yeah I can map these values to outlinePatternOutlineWidth

packages/fast-components-styles-msft/src/carousel/index.ts
nicholasrice: @bheston Would these values align to the L{1-4} swatches you have been exploring? What is the logical source of these colors?

PR #1505, opened 2019-03-07
fix: update focus events on keyboard close of menu

scomea:
Select was focusing on next item rather than select trigger on menu close.  This forces focus back to the trigger.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/select/select.tsx
+++ b/packages/fast-components-react-base/src/select/select.tsx
@@ -121,10 +121,7 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
             !this.state.isMenuOpen &&
             !this.props.multiselectable
         ) {
-            const triggerButton: HTMLButtonElement = this.getTriggerButton();
-            if (triggerButton !== null) {
-                triggerButton.focus();
-            }
+            this.focusOnTrigger();
         }
     }
 
@@ -364,13 +361,19 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (this.props.disabled || e.defaultPrevented) {
             return;
         }
+
         switch (e.keyCode) {
             case KeyCodes.enter:
             case KeyCodes.space:
+                e.preventDefault();
                 this.toggleMenu(!this.state.isMenuOpen);
+                if (this.state.isMenuOpen && this.props.isMenuOpen !== true) {
+                    this.focusOnTrigger();
+                }
                 break;
             case KeyCodes.escape:
                 this.toggleMenu(false);
+                this.focusOnTrigger();
                 break;
             case KeyCodes.arrowDown:
             case KeyCodes.arrowRight:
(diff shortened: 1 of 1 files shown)


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -364,13 +361,19 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (this.props.disabled || e.defaultPrevented) {
             return;
         }
+
         switch (e.keyCode) {
             case KeyCodes.enter:
             case KeyCodes.space:
+                e.preventDefault();
                 this.toggleMenu(!this.state.isMenuOpen);
+                if (this.state.isMenuOpen && this.props.isMenuOpen !== true) {
```
chrisdholt: This may be a case where we actually want to use `getDerivedStateFromProps`. IE, if we get new props where it says the menu should close, we should update the state of our component to illustrate that. This would allow us to simply update the state and respond to that change here:

```
if (this.state.isMenuOpen !== true) {
    this.focusOnTrigger();
}
```
Thoughts?

scomea: This change isn't for controlled mode, although we may need to put some thought into how we move focus in that circumstance (frankly not sure how we'd know where to move focus in that situation though).  This just focuses on the trigger in uncontrolled mode if the menu was dismissed from a keyboard event.

nicholasrice: Should this `if (this.state.isMenuOpen && this.props.isMenuOpen !== true)` be using the `validateMenuState ` method instead?

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -364,13 +361,19 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (this.props.disabled || e.defaultPrevented) {
             return;
         }
+
         switch (e.keyCode) {
             case KeyCodes.enter:
             case KeyCodes.space:
+                e.preventDefault();
```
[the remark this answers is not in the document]
scomea: It turns out that if you capture keydown on an element, and that key is "Enter" and during that function call transfer focus to a button element, that button will also emit a keydown event.  PreventDefault prevents the second event from immediately re-opening the menu (and yes, the event emitted by the button is the same one handled the first time around).  Yes, seems weird to me too.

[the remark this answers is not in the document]
scomea: I will add a comment

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -512,6 +514,16 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
             element.getAttribute("aria-disabled") !== "true"
         );
     };
+
+    /**
+     * focus on the trigger button
+     */
+    private focusOnTrigger = (): void => {
```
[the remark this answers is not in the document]
scomea: focusTriggerElement it is

packages/fast-components-react-base/src/select/select.tsx
scomea: Yep, that worked.

PR #1508, opened 2019-03-07
fix: increase specificity on MSFT styled components which are derived from other MSFT components

### LINKED ISSUE #1498: ensure MSFT components are not suseptible to specificity issues
We need to make sure component's we're exporting are not suseptible to the issues outlined in #1279
.

To do this, styled-compnents should never rely exclusivly on document order for CSS property specificity. If one component is overriding the style properties of another component, the overriding component should either:
a. Use jssStyleSheet props to override the styles on the component
b. Bolister specificity by adding seleectors (`.heading.heading__1` vs `.heading__1`)

chrisdholt:
closes #1498 
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/paragraph/index.ts
+++ b/packages/fast-components-styles-msft/src/paragraph/index.ts
@@ -4,16 +4,20 @@ import { ParagraphClassNameContract } from "@microsoft/fast-components-class-nam
 import { fontWeight } from "../utilities/fonts";
 
 const styles: ComponentStyles<ParagraphClassNameContract, DesignSystem> = {
-    paragraph: {},
-    paragraph__1: {
-        fontWeight: `${fontWeight.semilight}`,
+    paragraph: {
+        "&$paragraph__1": {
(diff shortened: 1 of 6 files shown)


packages/fast-components-styles-msft/src/paragraph/index.ts
```diff
@@ -4,16 +4,20 @@ import { ParagraphClassNameContract } from "@microsoft/fast-components-class-nam
 import { fontWeight } from "../utilities/fonts";
 
 const styles: ComponentStyles<ParagraphClassNameContract, DesignSystem> = {
-    paragraph: {},
-    paragraph__1: {
-        fontWeight: `${fontWeight.semilight}`,
```
nicholasrice: I think you should actually leave these as-is, because otherwise they will have the oppertunity to override Heading

nicholasrice: Sorry thought this was Typography - you're good.

Issue #1416, opened 2019-02-21
Unhandled props are being passed to an element which is not the root in text action


chrisdholt:
Unhandled props are expected as part of the API to be passed to the root of a given component. In text action, this is currently not the case as it is being applied to the text field. This complicates things as this is not documented as being "special" or "different" and the expectation for unhandledProps are that they always pass to the root DOM node. 

I get why someone may want to pass unhandled props to the text field, but I don't know that it's reason enough to break the expected API for a user. I think we need to remove unhandled props there and figure out a way to solve the issue of wanting/needing to pass things to the text field component itself.

nalogiudice:
> Unhandled props are expected as part of the API to be passed to the root of a given component

Is this explicitly stated in our documentation?

In this case, it is more important that the user be able to pass unhandled props to the input than the wrapper. Whatever the result of this is, that still needs to be able to happen.

chrisdholt:
> > Unhandled props are expected as part of the API to be passed to the root of a given component
> 
> Is this explicitly stated in our documentation?
> 
> In this case, it is more important that the user be able to pass unhandled props to the input than the wrapper. Whatever the result of this is, that still needs to be able to happen.

It is currently documented as part of the unhandled props detail on the foundation class in `fast-components-foundation-react`. Documentation is being added with #1399 and we should include that clearly there as well.

As I mentioned - I **get** why this is important. I don't think this is the right solution, or we need to solve that in a different way. I think there is a larger story here, and we can create a feature request off that. I don't think the solution is to keep this as "special" though.

PR #1514, opened 2019-03-08
feat: update to move text area unhandled props location to outer div

nalogiudice:
Closes [#1416](https://github.com/Microsoft/fast-dna/issues/1416)

# Description
Changes `{...this.unhandledProps()}` location to outer div.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/text-action/text-action.tsx
```diff
@@ -48,14 +48,13 @@ class TextAction extends Foundation<
      */
     public render(): JSX.Element {
         return (
-            <div className={this.generateClassNames()}>
+            <div className={this.generateClassNames()} {...this.unhandledProps()}>
```
janechu: Unhandled props should be before className or your overrides will not work.

nalogiudice: Makes perfect sense - good to know. We should include this in our documentation if it is not already there.

PR #1524, opened 2019-03-12
fix: remove memoization from color utilities

nicholasrice:
# Description

The current memoization techniques do not work. We can optomize these functions with functional memoization at a later point, but we should remove the bug for the time being. Tracking future work in #1524

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/utilities/color/accent-foreground-cut.ts
```diff
@@ -11,14 +11,9 @@ import { ColorRecipe, contrast } from "./common";
 const accentForegroundCutAlgorithm: (
     backgroundColor: Swatch,
     targetContrast: number
-) => Swatch = memoize(
```
janechu: can you remove the memoize import from this file?

Issue #1515, opened 2019-03-08
ResizeObserver is never disengaged in Horizontal Overflow


hchokshi:
**Describe the bug; what happened?**
In Chrome, console logs show resize events being fired, and access to members of `undefined` once a horizontal overflow is unmounted. This is due to the resize observer not being removed in `componentWillUnmount`.

**What are the steps to reproduce the issue?**
1. Create a `HorizontalOverflow` and a button that removes it.
2. Click the button to remove the overflow component in Chrome.

**What behavior did you expect?**
Observer is cleaned up on component dismount.

**If applicable, provide screenshots:**
![image](https://user-images.githubusercontent.com/10136407/54054624-8e07eb80-419f-11e9-9ecc-7b38a027600d.png)

**In what environment did you see the issue?**

* OS & Device
Windows 10 Enterprise 1809

* Browser [e.g. Edge, Chrome, Safari, FireFox]
Chrome 72

* Version [e.g. 1.8.0]
fast-components-react-base 3.9.1

**Is there any additional context?**
Resize observer object reference should be kept at init time, and `.disconnect()` called on `componentWillUnmount`. 
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#Methods

nicholasrice:
Thanks @hchokshi - looks like a bug, thanks for filing the issue.

shauntc:
@nicholasrice I think this can be closed now

chrisdholt:
closed by #1529

PR #1529, opened 2019-03-12
fix: update to disconnect the resize observer on component will unmount

shauntc:
# Description
Added a private reference to the resize observer (resizeObserver) in horizontal-overflow. Assign the observer if it is created on component did mount and disconnect it on component will unmount

## Motivation & context
Bugfix for #1515

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -6,6 +6,7 @@ import HorizontalOverflow, {
     HorizontalOverflowClassNameContract,
 } from "./";
 import "raf/polyfill";
+import { ConstructableResizeObserver } from "./resize-observer";
 
 /*
  * Configure Enzyme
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -527,5 +528,56 @@ describe("horizontal overflow", (): void => {
...
+        const ActualObserver: ConstructableResizeObserver = (window as WindowWithResizeObserver)
+            .ResizeObserver;
+        const disconnect: jest.Mock<any, any> = jest.fn();
+        // Mock the resize observer
+        // tslint:disable-next-line:max-classes-per-file
+        class MockObserver {
+            public observe: jest.Mock<any, any> = jest.fn();
+            public unobserve: jest.Mock<any, any> = jest.fn();
+            public disconnect: jest.Mock<any, any> = disconnect;
+        }
```
chrisdholt: Is there a particular reason you're creating two instances of the class here instead of defining it outside the scope of the tests and reusing?

shauntc: not really, it's just a simple way of keeping track of the correct mock functions and keeping the tests independant. Each class only keeps a reference to the mock functions that it needs

nicholasrice:
Thanks @shauntc. Looks like your build is failing because of an issue discovered earlier. Rebasing against master should fix the issue.

PR #1543, opened 2019-03-14
feat: update and refactor the CSS editor

janechu:
# Description

Updates the CSS editor to have a similar interface to the `Form` and other tooling, consolidates some logic to allow addition of other components to be easier.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/css-editor/editor.props.ts
+++ b/packages/fast-tooling-react/src/css-editor/editor.props.ts
@@ -1,25 +1,25 @@
 import { ManagedClasses } from "@microsoft/fast-jss-manager";
 import { CSSEditorClassNameContract } from "./editor.style";
-import { Subtract } from "utility-types";
-import { CSSPositionHandledProps, CSSSpacingHandledProps } from "./";
-import { CSSSpacingClassNameContract } from "./spacing/spacing.style";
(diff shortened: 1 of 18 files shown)


packages/fast-tooling-react/src/css-editor/editor.props.ts
```diff
@@ -1,25 +1,25 @@
...
 
 export interface CSSEditorHandledProps
-    extends Subtract<
-            CSSPositionHandledProps,
-            ManagedClasses<CSSPositionClassNameContract>
-        >,
-        Subtract<CSSSpacingHandledProps, ManagedClasses<CSSSpacingClassNameContract>>,
-        ManagedClasses<CSSEditorClassNameContract> {
+    extends ManagedClasses<CSSEditorClassNameContract> {
+    /**
```
nicholasrice: minor nit, but I think `onChange` may be more accurate here. Correct me if I'm wrong, but my understanding is that the callback gets called when the user has changed the inputs right? The author still needs to update the input data with the reflected changes for the UI to to reflect the changes, right?

janechu: That's true, I'll change it back

PR #1548, opened 2019-03-15
feat: update schemas to include form plugin ids

janechu:
# Description

Add the keyword `formPluginId` to the root of schemas and to their children properties.

## Motivation & context

This will allow the creation of plugins to use these hooks to make children required and give them specific defaults per a separate API, for example changing the heading component children to the text "Lorem ipsum".

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/badge/badge.schema.json
+++ b/packages/fast-components-react-base/src/badge/badge.schema.json
@@ -4,11 +4,13 @@
     "description": "A badge component's schema definition.",
     "type": "object",
     "id": "@microsoft/fast-components-react-base/badge",
+    "formPluginId": "@microsoft/fast-components-react-base/badge",
     "properties": {},
     "reactProperties": {
         "children": {
             "title": "Children",
             "type": "children",
+            "formPluginId": "@microsoft/fast-components-react-base/badge/children",
             "defaults": [
                 "text"
             ]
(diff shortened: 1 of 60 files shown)


packages/fast-components-react-msft/src/action-trigger/action-trigger.schema.json
```diff
@@ -28,12 +29,14 @@
...
             "title": "Children",
-            "type": "children"
+            "type": "children",
+            "formPluginId": "@microsoft/fast-components-react-msft/action-trigger/children"
         },
         "glyph": {
             "title": "Glyph",
             "type": "children",
-            "pluginId": "@microsoft/fast-components-react-msft/action-trigger/glyph"
+            "pluginId": "@microsoft/fast-components-react-msft/action-trigger/glyph",
```
chrisdholt: Is the pluginId here different then the new formPluginId?

packages/fast-components-react-base/src/badge/badge.schema.json
```diff
@@ -4,11 +4,13 @@
     "description": "A badge component's schema definition.",
     "type": "object",
     "id": "@microsoft/fast-components-react-base/badge",
+    "formPluginId": "@microsoft/fast-components-react-base/badge",
```
chrisdholt: I’m assuming we need this to be different than the actual ID above, correct?

janechu: It does not, as it is at root level. This merely adds a hook for the entire schema.

packages/fast-components-react-msft/src/action-trigger/action-trigger.schema.json
janechu: It is, the `pluginId` is for the data mapping and allow us to manipulate data blobs, the `formPluginId` is for manipulating the schema that goes into the form generator.

PR #1551, opened 2019-03-18
feat: add auto suggest component

scomea:
Description
Add a base and styled version of auto-suggest to our component library

## Motivation & context
Our partners need this functionality.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+/*
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+const itemA: JSX.Element = <ListboxItem id="a" value="a" displayString="a" />;
+const itemB: JSX.Element = <ListboxItem id="b" value="b" displayString="ab" />;
+const itemC: JSX.Element = <ListboxItem id="c" value="c" displayString="abc" />;
+
+describe("auto select", (): void => {
```
chrisdholt: autosuggest?

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+        rendered
+            .find('[displayString="a"]')
+            .simulate("keydown", { keyCode: KeyCodes.arrowDown });
+        input = rendered.find("input");
+        expect(input.prop("aria-owns")).toEqual("listboxId");
+        expect(input.prop("aria-controls")).toEqual("listboxId");
+        expect(input.prop("aria-activedescendant")).toEqual("b");
+
+        document.body.removeChild(container);
```
chrisdholt: 👍 - As an FYI, you can also call an enzyme method here for "detach()"

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
scomea: yep

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
scomea: ok, thx

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+        };
+
+        const rendered: any = shallow(
+            <AutoSuggest listboxId="listboxId" {...unhandledProps} />
+        );
+
+        expect(rendered.first().prop("aria-label")).toEqual("label");
+    });
+
+    test("initial default state is correct", (): void => {
```
chrisdholt: I think we may want to be more specific in terms of what "correct" is here. Perhaps we can have a test for each of the initial default states below.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+                {itemC}
+            </AutoSuggest>
+        );
+
+        expect(rendered.state("value")).toBe("");
+        expect(rendered.state("focusedItem")).toBe(null);
+        expect(rendered.state("isMenuOpen")).toBe(false);
+    });
+
+    test("default input region attributes are set correctly", (): void => {
```
chrisdholt: Same thought on "correctly" here. Let's try and further define this so we know what we're testing against here

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+        expect(input.prop("aria-label")).toEqual("test-label");
+        expect(input.prop("aria-autocomplete")).toEqual("both");
+        expect(input.prop("aria-activedescendant")).toEqual(null);
+        expect(input.prop("aria-owns")).toEqual(null);
+        expect(input.prop("aria-controls")).toEqual(null);
+        expect(input.prop("disabled")).toBe(true);
+        expect(input.prop("role")).toEqual("combobox");
+    });
+
+    test("input region attributes change correctly as options list navigated", (): void => {
```
chrisdholt: I think we want more detail here, perhaps in this case since attributes are changing during list navigation we should consider breaking this up based on behavior

packages/fast-components-react-msft/src/auto-suggest-option/auto-suggest-option.props.ts
```diff
@@ -0,0 +1,33 @@
...
+        Subtract<BaseListboxItemHandledProps, BaseListboxItemManagedClasses> {
+    /**
+     * The function that formats the display based on current search string
+     */
+    displayFormatter?: (displayString: string, searchString: string) => React.ReactNode;
+
+    /**
+     * Flag that enables/disables display formatting, default is true
+     */
+
```
chrisdholt: nit: extra space

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+        }
+        return shouldOpenMenu;
+    };
+
+    /**
+     * Close the menu when when there are clicks outside
+     */
+    private handleWindowClick = (event: MouseEvent): void => {
+        if (
+            this.state.isMenuOpen &&
```
nicholasrice: are you manually keeping `this.state.isMenuOpen` in sync with `this.props.isMenuOpen`? 

This would probably be a good candidate for `getDerivedStateFromProps` - I think that would make this whole component more declarative.

Your whole `validateMenuState` could be moved to that function, and then you could just `setState` where needed and not have to validate the state prior to use.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+     */
+    private handleMenuKeydown = (e: React.KeyboardEvent<Element>): void => {
+        if (this.props.disabled || e.defaultPrevented) {
+            return;
+        }
+        switch (e.keyCode) {
+            case KeyCodes.escape:
+                this.toggleMenu(false);
+                break;
+
```
nicholasrice: what type of element is focused here? If it is a div, you will likely want to call `e.preventDefault()`, otherwise when this component is on pages that overflow the window vertically the down/up arrow will also scroll the page.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
scomea: made tests more "atomic"

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
scomea: made tests more "atomic"

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
scomea: made tests more "atomic"

packages/fast-components-react-msft/src/auto-suggest-option/auto-suggest-option.props.ts
scomea: ?...not seeing it...

packages/fast-components-react-base/src/auto-suggest/auto-suggest-context.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from "react";
+
+export interface AutoSuggestContextType {
+    currentValue: string;
```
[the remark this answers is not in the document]
scomea: yes

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+            className = className.concat(
+                " ",
+                get(this.props.managedClasses, "autoSuggest__disabled", "")
+            );
+        }
+
+        if (this.state.isMenuOpen) {
+            className = className.concat(
+                " ",
+                get(this.props.managedClasses, "autoSuggest_menu__open", "")
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+            );
+        }
+
+        if (this.state.value === "") {
+            className = className.concat(
+                " ",
+                get(this.props.managedClasses, "autoSuggest_showPlaceholder", "")
```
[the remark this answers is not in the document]
scomea: The actual bug is that the class is not being used so this code should be deleted (logic to show placeholder or not lives in a subcomponent - TextAction).

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+        }
+    };
+
+    /**
+     * Focus on the input element
+     */
+    private focusOnInput = (): void => {
+        const inputElements: HTMLCollectionOf<
+            HTMLInputElement
+        > = this.rootElement.current.getElementsByTagName("input");
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+            inputElements[0].focus();
+        }
+        this.updateFocusedItem([]);
+    };
+
+    /**
+     * Determine menu state by comparing desired state to props
+     */
+    private validateMenuState = (desiredMenuState: boolean): boolean => {
+        let shouldOpenMenu: boolean = desiredMenuState;
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -0,0 +1,523 @@
...
+
+        return super.generateClassNames(className);
+    }
+
+    /**
+     * Determine which function to use to render content display (ie. the part of the control that shows when the menu isn't open)
+     * and invokes it
+     */
+    private renderInputRegion(): React.ReactNode {
+        if (this.props.inputRegion !== undefined) {
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: ok

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: Maybe, but for the moment this is the pattern we've been following for select as well and there are places where we figure out what to do with focus based on validating menu state that may not translate well into the other appoach.  I'd take a bug to investigate making the change, but would rather not delay this component on that now (and it would be non-breaking, in theory).

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: That revealed some issues with my end of list code, fixed.  This could be an issue in other places as I'm trying to only preventDefault where I absolutely have to.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: Looked at this a bit more. Since we don't have access to the component instance in the 'getDerivedState' call I'm not sure where/how we would do the focus management that we currently do in 'togglemenu'.

PR #1553, opened 2019-03-18
feat: update styles and design system density.

bheston:
# Description

- Updated component style to match latest designs as necessary.
- Migrated common values like padding, border widths, and corner radius into the design system.
- Added density support to the design system to adjust font size, padding, and height as configured.

## Motivation & context

Gets components in alignment with the current design and adds support for higher-density uses through design system configuration.

## Issue type checklist

This change removes the "height" utility constants for default, min, and max height.

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -11,13 +11,10 @@ import {
     applyFocusVisible,
     applyLocalizedProperty,
     Direction,
+    toPx,
 } from "@microsoft/fast-jss-utilities";
 import { DesignSystem, withDesignSystemDefaults } from "../design-system/index";
-import {
-    accentFillRest,
-    accentForegroundCut,
-    neutralForegroundRest,
-} from "../utilities/color";
+import { accentFillRest, accentForegroundCut } from "../utilities/color";
 
 // Since MSFT button is already styled, we need to override in this way to alter button classes
 export const callToActionButtonOverrides: ComponentStyles<
@@ -36,9 +33,10 @@ export const callToActionButtonOverrides: ComponentStyles<
             "& $button_contentRegion": {
                 transform: (config: DesignSystem): string => {
                     const designSystem: DesignSystem = withDesignSystemDefaults(config);
+                    const xTranslatePx: string = toPx(designSystem.designUnit);
                     return designSystem.direction === Direction.ltr
-                        ? "translateX(-4px)"
-                        : "translateX(4px)";
+                        ? `translateX(-${xTranslatePx})`
+                        : `translateX(${xTranslatePx})`;
                 },
             },
         },
@@ -57,13 +55,13 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
 ): ComponentStyleSheet<CallToActionClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
     const direction: Direction = designSystem.direction;
+    const xTranslatePx: string = toPx(designSystem.designUnit);
 
     return {
         callToAction: {
             transition: "all 0.2s ease-in-out",
             display: "inline-flex",
             maxWidth: "100%",
-            border: "2px solid transparent",
             lineHeight: "1",
             textDecoration: "none",
             whiteSpace: "nowrap",
@@ -71,16 +69,16 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
                 "& $callToAction_glyph": {
                     transform:
                         direction === Direction.ltr
-                            ? "translateX(4px)"
-                            : "rotate(180deg) translateX(4px)",
+                            ? `translateX(${xTranslatePx})`
+                            : `rotate(180deg) translateX(${xTranslatePx})`,
                     position: "relative",
                 },
             },
             ...applyFocusVisible("& $callToAction_glyph", {
                 transform:
                     direction === Direction.ltr
-                        ? "translateX(4px)"
-                        : "rotate(180deg) translateX(4px)",
+                        ? `translateX(${xTranslatePx})`
+                        : `rotate(180deg) translateX(${xTranslatePx})`,
                 position: "relative",
             }),
         },
@@ -92,7 +90,7 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             [applyLocalizedProperty("marginLeft", "marginRight", direction)]: "6px",
             transform: direction === Direction.ltr ? "none" : "rotate(180deg)",
             transition: "all 600ms cubic-bezier(0.19, 1, 0.22, 1)",
-            marginTop: direction === Direction.ltr ? "4px" : "0",
+            marginTop: direction === Direction.ltr ? xTranslatePx : "0",
         },
         callToAction__primary: {
             "& $callToAction_glyph": {
@@ -105,6 +103,7 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             },
         },
         callToAction__justified: {
+            paddingRight: xTranslatePx,
             "& $callToAction_glyph": {
                 fill: accentFillRest,
             },
--- a/packages/fast-components-styles-msft/src/carousel/index.ts
+++ b/packages/fast-components-styles-msft/src/carousel/index.ts
@@ -62,7 +62,7 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
                 outline: "none",
             },
             "&::before": {
-                opacity: "0.2",
+                opacity: "0.45",
                 border: "1px solid transparent",
                 borderRadius: "40px",
                 content: "''",
@@ -74,7 +74,7 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
             "&:not($carousel_sequenceIndicator__active)": {
                 "&:hover": {
                     "&::before": {
-                        opacity: "0.5",
+                        opacity: "0.9",
                     },
                 },
             },
--- a/packages/fast-components-styles-msft/src/checkbox/index.ts
+++ b/packages/fast-components-styles-msft/src/checkbox/index.ts
@@ -1,11 +1,11 @@
-import { DesignSystem, withDesignSystemDefaults } from "../design-system";
 import {
-    ComponentStyles,
-    ComponentStyleSheet,
-    CSSRules,
-} from "@microsoft/fast-jss-manager";
+    applyCornerRadius,
+    applyDisabledState,
+    DesignSystem,
+    withDesignSystemDefaults,
+} from "../design-system";
+import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
 import { CheckboxClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
-import { applyTypeRampConfig } from "../utilities/typography";
 import {
     applyFocusVisible,
     applyLocalizedProperty,
@@ -22,12 +22,24 @@ import {
     neutralOutlineHover,
     neutralOutlineRest,
 } from "../utilities/color";
+import { applyFontSize, densityToTypeOffset, padding } from "../utilities/density";
 
 const styles: ComponentStyles<CheckboxClassNameContract, DesignSystem> = (
     config: DesignSystem
 ): ComponentStyleSheet<CheckboxClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
     const direction: Direction = designSystem.direction;
+    const size: number =
+        (designSystem.defaultHeightMultiplier + designSystem.density) *
+            (designSystem.designUnit / 2) +
+        designSystem.designUnit;
+    const indicatorMargin: string = toPx(
+        designSystem.designUnit + densityToTypeOffset(designSystem)
+    );
+    const indicatorSvg: string = `<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path fill="${encodeURIComponent(
+        neutralForegroundRest(designSystem)
+    )}" fill-rule="evenodd" clip-rule="evenodd" d="M8.143 12.6697L15.235 4.5L16.8 5.90363L8.23812 15.7667L3.80005 11.2556L5.27591 9.7555L8.143 12.6697Z"/></svg>`;
+    const indicatorSvgHc: string = `<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path fill="ButtonHighlight" fill-rule="evenodd" clip-rule="evenodd" d="M8.143 12.6697L15.235 4.5L16.8 5.90363L8.23812 15.7667L3.80005 11.2556L5.27591 9.7555L8.143 12.6697Z"/></svg>`;
 
     return {
         checkbox: {
@@ -37,15 +49,19 @@ const styles: ComponentStyles<CheckboxClassNameContract, DesignSystem> = (
             alignItems: "center",
             transition: "all 0.2s ease-in-out",
             "& $checkbox_label": {
-                [applyLocalizedProperty("paddingLeft", "paddingRight", direction)]: "8px",
+                [applyLocalizedProperty(
+                    "paddingLeft",
+                    "paddingRight",
+                    direction
+                )]: padding(2)(designSystem),
             },
         },
         checkbox_input: {
             position: "absolute",
-            width: "20px",
-            height: "20px",
+            width: toPx(size),
+            height: toPx(size),
             appearance: "none",
-            borderRadius: "2px",
+            ...applyCornerRadius(designSystem),
             boxSizing: "border-box",
             margin: "0",
             zIndex: "1",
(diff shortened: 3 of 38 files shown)


packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
+        const value: number = paddingNumber(minusBorder)(designSystem);
+        return typeof unit === "string" ? `${value}${unit}` : toPx(value);
+    };
+}
+
+/**
+ * Returns the standard horizontal padding for text and icons as a number.
+ *
+ * @param minusBorder Any border that should be removed from the overall content padding.
+ * @param unit The unit of measurement; px by default.
```
nicholasrice: this doesn't accept a unit

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
 
+/**
+ * Returns the type ramp adjustment from the default setting. Collapses
+ * the fact that a single type size is used for multiple densities.
+ *
+ * @param designSystem The design system config
+ */
+export function densityToTypeOffset(designSystem: DesignSystem): number {
+    const offset: number =
+        designSystem.density >= 2 ? 1 : designSystem.density <= -2 ? -1 : 0;
```
nicholasrice: Is it intended that if density is `2` then the offset would be one, but if the density was `1.9` then the offset would be `1.9`?

Should these always be integers?

packages/fast-components-styles-msft/src/typography/index.ts
```diff
@@ -3,9 +3,11 @@ import { ComponentStyles } from "@microsoft/fast-jss-manager";
 import { TypographyClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 import { applyTypeRampConfig } from "../utilities/typography";
 import { neutralForegroundRest } from "../utilities/color";
+import { applyCursorDefault } from "../utilities/cursor";
 
 const styles: ComponentStyles<TypographyClassNameContract, DesignSystem> = {
     typography: {
+        ...applyCursorDefault(),
```
nicholasrice: We don't want the text cursor?

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
+        return value;
+    };
+}
+
+/**
+ * Adds the font size and line height attributes for the type ramp according to the design system config.
+ *
+ * @param config The design system config.
+ */
+export function applyFontSize(
```
nicholasrice: I kind of think this could be more generally usefull. Really what we're doing is scaling typography with density - I think a more intuitive API might be to pass in a "default" typeramp ID that, with no density modifier, would be exaclty that type ID. but, given a density, it would scale accordingly.

Passing an "offset" requires you to know what the function internals are doing - to achieve a scaled "t5" I would need to do:  `applyFontSize(designSystem, -2)` as opposed to just sying `scaleTypeRamp("t5")(designSystem)`

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
+        const value: number =
+            (designSystem.defaultHeightMultiplier + designSystem.density) *
+                designSystem.designUnit *
+                lines -
+            minusBorder;
+        return value;
+    };
+}
+
+export function minHeight(
```
nicholasrice: These two seem like pretty generic names - can you elboarate on how these should be used?

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -109,6 +141,37 @@ export function ensureDesignSystemDefaults<T>(
     };
 }
 
+export function applyCornerRadius(
```
nicholasrice: Can you make seperate files for these utilities? I wouldn't expect these to live in the `designSystem` file

packages/fast-components-styles-msft/src/flipper/index.ts
```diff
@@ -31,52 +29,49 @@ const styles: ComponentStyles<FlipperClassNameContract, DesignSystem> = (
...
             position: "relative",
             color: neutralForegroundRest,
             background: "transparent",
-            border: `${toPx(designSystem.outlinePatternOutlineWidth)} solid transparent`,
-            borderRadius: "50%",
+            border: "none",
             padding: "0",
             "&::before": {
-                transition: "all 0.2s ease-in-out",
+                transition: "all 0.1s ease-in-out",
```
nicholasrice: yessssss

packages/fast-components-styles-msft/src/number-field/index.ts
```diff
@@ -16,12 +15,12 @@ const styles: ComponentStyles<NumberFieldClassNameContract, DesignSystem> = (
     return {
         numberField: {
             ...inputFieldStyles(designSystem),
-            height: density(defaultHeight)(designSystem),
-            minHeight: toPx(minHeight),
-            maxHeight: toPx(maxHeight),
+            height: height()(designSystem),
+            minHeight: minHeight()(designSystem),
+            maxHeight: maxHeight()(designSystem),
```
nicholasrice: FYI there is a utility called `localizeSpacing` that will split the string and re-arrange it automatically.

Also, do these actually need to be different from LTR to RTL? see `padding()` vs `padding(2)`

packages/fast-components-styles-msft/src/paragraph/index.ts
```diff
@@ -3,16 +3,22 @@ import { ComponentStyles, CSSRules } from "@microsoft/fast-jss-manager";
...
+    return {
+        fontWeight: `${fontWeight.normal}`,
+    };
+}
+
 const styles: ComponentStyles<ParagraphClassNameContract, DesignSystem> = {
     paragraph: {
         "&$paragraph__1": {
-            fontWeight: `${fontWeight.semilight}`,
+            ...applyParagraphStyles(),
```
nicholasrice: if this is just `font-weight: normal` you can probably just apply this to `paragraph` and not apply it to each level *of* paragraph.

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
+/**
+ * Returns the standard horizontal padding for text and icons as a number.
+ *
+ * @param minusBorder Any border that should be removed from the overall content padding.
+ * @param unit The unit of measurement; px by default.
+ */
+export function paddingNumber(minusBorder: number = 0): (config: DesignSystem) => number {
+    return (config: DesignSystem): number => {
+        const designSystem: DesignSystem = withDesignSystemDefaults(config);
+        const offset: number = densityToTypeOffset(designSystem);
```
nicholasrice: it seems a bit odd to me to be using a type offset scale to derive this -  how is this related to typography?

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -44,6 +46,21 @@ export interface DesignSystem {
...
+
+    /**
+     * The number of designUnits used for horizontal padding at the base density (0).
+     */
+    defaultPaddingMultiplier: number;
+
+    /**
+     * The type ramp size to use at the base density (0).
+     */
+    defaultTypeRampSize: number;
```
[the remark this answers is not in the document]
bheston: It's here so we can adjust it by adding or subtracting for density. Since the type ramps are keys, that sounds harder to deal with, unless perhaps in the type ramp they should just be numbers.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: Yes, whole numbers only. I updated the comments on density in design system. Was looking for a stronger-typed method of specifying that, but wasn't sure if it was necessary or not.

packages/fast-components-styles-msft/src/typography/index.ts
bheston: Based on the though that we could use cursors for a11y cues, it seems like there is less value in having the cursor change over static text. Many web apps turn that and other things like right-click and selection off, so I considered this as our default. We should consider if that's not right though.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: I was thinking in terms of offsets because I think it actually requires you to know less. Your t5 example is only because you happen to know that t7 is the default. I've added a design system setting for the default, so we could conceivably change it to 8, in which case a -2 offset would produce t6 text. So default exists, it's just centralized.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: Admittedly they are carryovers from the previous height utility, but turned into a formula based on the design system instead of fixed values. Currently the only place they are used are in actual min/max height css attributes. There's not much point to that though, as those places are not variable-sized components. They are always used with an explicit height. It's possible we don't need this notion and should just remove any setting of min/max based on density.

packages/fast-components-styles-msft/src/design-system/index.ts
bheston: I was thinking of it as convenience methods on things from `designSystem`, but that makes sense. I'll move them out.

packages/fast-components-styles-msft/src/number-field/index.ts
bheston: Updated to `localizeSpacing`. The difference is because of the up and down arrows. It's somewhat temporary considering we want to stylize this component, but it's so they don't push all the way over.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: Because the horizontal padding is related to how large the text is. We only have three text sizes, and only adjust the padding when the text size changes.

packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -101,12 +100,13 @@ export function elevationShadow(
...
-            config
-        );
-        const opacity: number = shadowConfig.opacity;
+        const xOffset: string = toPx(xValue);
+        const yOffset: string = toPx(yValue);
+        const blur: string = toPx(shadowConfig.blurMultiplier * elevationValue);
+        const opacity: number =
+            elevationValue > 24
+                ? shadowConfig.opacity
+                : Math.round(shadowConfig.opacity * 100 * 0.6) / 100;
```
janechu: you could just make this `shadowConfig.opacity * 60`

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
+    return (config: DesignSystem): string => {
+        const designSystem: DesignSystem = withDesignSystemDefaults(config);
+        const value: number =
+            (designSystem.defaultHeightMultiplier - 3) * designSystem.designUnit * lines -
+            minusBorder;
+        return toPx(value);
+    };
+}
+
+export function maxHeight(
```
janechu: I'd like to see some comments on these functions

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -36,9 +33,10 @@ export const callToActionButtonOverrides: ComponentStyles<
             "& $button_contentRegion": {
                 transform: (config: DesignSystem): string => {
                     const designSystem: DesignSystem = withDesignSystemDefaults(config);
+                    const xTranslatePx: string = toPx(designSystem.designUnit);
```
janechu: nit: i would prefer `translateXValue`

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -105,6 +103,7 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             },
         },
         callToAction__justified: {
+            paddingRight: xTranslatePx,
```
janechu: RTL?

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -74,7 +74,7 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = (
             "&:not($carousel_sequenceIndicator__active)": {
                 "&:hover": {
                     "&::before": {
-                        opacity: "0.5",
+                        opacity: "0.9",
```
janechu: should these opacities be based on anything?

packages/fast-components-styles-msft/src/checkbox/index.ts
```diff
@@ -56,90 +72,69 @@ const styles: ComponentStyles<CheckboxClassNameContract, DesignSystem> = (
...
+                    ...applyCornerRadius(designSystem),
                     transform: "none",
-                    [applyLocalizedProperty("left", "right", direction)]: "5px",
-                    top: "5px",
-                    height: "10px",
-                    width: "10px",
+                    [applyLocalizedProperty("left", "right", direction)]: indicatorMargin,
+                    top: indicatorMargin,
+                    bottom: indicatorMargin,
+                    right: indicatorMargin,
```
janechu: should you be setting this here? seems like the applyLocalizedProperty above should set left/right

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -35,7 +37,7 @@ export interface DesignSystem {
     contrast: number;
 
     /**
-     * The density multiplier
+     * The density offset. Plus or minus 1 - 3.
      */
     density: number;
```
janechu: should this be set to `densityOffset`? Also I'm finding the wording in the comment confusing, do you mean to say "The density offset which can be an integer in a range from -3 to 3"? Also you could more strongly type this by changing this to `densityOffset: -3 | -2 | -1 | 0 | 1 | 2 | 3;`

packages/fast-components-styles-msft/src/utilities/density.ts
chrisdholt: I agree with Nick in terms of supporting more of a more intuitive API. Right now this seems like a tightly coupled black box

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -60,22 +75,37 @@ export interface DesignSystem {
...
 }
 
 const designSystemDefaults: DesignSystem = {
     backgroundColor: white,
     contrast: 0,
-    density: 1,
+    density: 0,
     designUnit: 4,
+    defaultHeightMultiplier: 8,
+    defaultPaddingMultiplier: 3,
```
chrisdholt: Is this always padding? Similar to @janechu's question on the function being applied to more than padding

packages/fast-components-styles-msft/src/checkbox/index.ts
```diff
@@ -22,12 +17,26 @@ import {
...
 
 const styles: ComponentStyles<CheckboxClassNameContract, DesignSystem> = (
     config: DesignSystem
 ): ComponentStyleSheet<CheckboxClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
     const direction: Direction = designSystem.direction;
+    const size: number =
+        (designSystem.defaultHeightMultiplier + designSystem.density) *
+            (designSystem.designUnit / 2) +
+        designSystem.designUnit;
```
chrisdholt: This feels a bit arbitrary in order to achieve a certain value(outcome) for radio/checkbox.  Primarily confused by what the principle is here:
```
(designSystem.designUnit / 2) +
        designSystem.designUnit;
```

packages/fast-components-styles-msft/src/utilities/elevation.ts
bheston: I wanted to keep it at whole percentages, like opacity 0.11 instead of 0.108.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: I removed these.

packages/fast-components-styles-msft/src/design-system/index.ts
bheston: Changed to "horizontalSpacing"

packages/fast-components-styles-msft/src/design-system/index.ts
bheston: Made a DensityOffset type.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: Removed these functions.

packages/fast-components-styles-msft/src/utilities/density.ts
nicholasrice: Yea you can strongly type it - we also need to ensure that in code though if that is the rule. While TypeScript is awesome, it is easily possible to circumvent the type system and not everyone that uses these libraries will be using TypeScript. 

You can change this statement to:
`return clamp(Math.floor(designSystem.density)-1, 1)

packages/fast-components-styles-msft/src/checkbox/index.ts
bheston: Updated to tie to height better.

packages/fast-components-styles-msft/src/carousel/index.ts
bheston: Possibly. I don't recognize a pattern that's in place yet though. This change was because the change was not perceivable, but the initial values came from experimentation with this particular usage over an image. I'm not sure what I would call this from a design systems perspective yet, but it's worth watching and considering making a setting.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: Modified per our conversation.

packages/fast-components-styles-msft/src/utilities/density.ts
bheston: Modified per our conversation.

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -1,6 +1,146 @@
...
+}
+
+/**
+ * Returns the standard horizontal padding for text and icons formatted in the provided unit or px by default.
+ *
+ * @param minusBorder Any border that should be removed from the overall content padding.
+ * @param unit The unit of measurement; px by default.
+ */
+export function padding(
+    minusBorder: number = 0,
```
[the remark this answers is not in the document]
bheston: I went with "adjustment" since there are other density related things referred to as offset.

bheston:
Closing this PR as I need to submit a new one on a different branch. Everything we've discussed should be resolved.

PR #1565, opened 2019-03-20
fix: fast layouts react cleanup

mxtopher:
chore(fast-layouts-react)

# Description

Remove some unnecessary imports, simplify some statements.

## Motivation & context

Improving readability, leveraging typescript's spread operator

## Issue type checklist

## Process & policy checklist

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://microsoft.github.io/fast-dna/docs/en/contributing/standards) for this project.


packages/fast-layouts-react/src/page/page.tsx
```diff
@@ -48,27 +41,18 @@ export class Page extends Foundation<PageHandledProps, PageUnhandledProps, {}> {
...
-        attributes.style = {
-            gridTemplateColumns: columns,
-            msGridColumns: columns,
+        return {
+            ...this.unhandledProps(),
+            className: super.generateClassNames(this.props.managedClasses.page),
+            style: {
+                gridTemplateColumns: columns,
+                msGridColumns: columns,
+            },
```
chrisdholt: Will pre-existing style attributes here return a combined style object, or is this going to blow away the entire style object due to shallow compare of the spread operator?

mxtopher: No, it won't merge the style object, but the current code doesn't do that either.
Note how it always overwrites `attributes.style` with the new object (line 66), throwing away what was there before.
That effectively means that the if block on line 62 is useless.

If we wanted to merge the properties, we would have to make another spread on style.

Is that the intended behavior?

chrisdholt: Ah - yes, looks like that was likely a bug; I think that would be the intended behavior, otherwise unhandled style props wouldn't be honored. Do you want to include that change as part of this?

mxtopher: I can.
What should take precedence on style? The unhandled props or this?
I mean, should we allow styles.gridTemplateColumns to be overridden by unhandledProps?
I assume no.

chrisdholt: Yep - I agree with that assessment

nicholasrice:
@mxtopher build re-queued

PR #1568, opened 2019-03-21
feat: add badges

### LINKED ISSUE #1522: Create a icons next to labels
Currently we have no way to express additional information on a field, in order to make this a more freeform experience we should have icons to the right of the field label for indication that there is something different about this field.

Icons to include:
- Info (more information is available for this field)
- Warning (there is an issue with this field the user may wish to know about)
- Locked (this field is locked and cannot be edited)

There may also be an instance where a custom badge may be added, this should require some additional investigation.

In addition to the badges, information should be provided to them that will show text when hovered.

In total two new fields (at least) will need to be added to the JSON schema, `badgeTitle` and `badgeDescription`.

Example:
```json
{
    "title": "Example",
    "description": "This is an example",
    "badge": "info",
    "badgeDescription": "This is further information about this field",
    "type": "string"
}
```

janechu:
# Description

This adds the ability to have badges "info", "warning" and "locked" with native tooltips in the form tooling.

## Motivation & context

Resolves #1522

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/form/badge-control.tsx
```diff
@@ -0,0 +1,36 @@
...
+                        className={this.props.className}
+                        description={this.props.description}
+                    />
+                );
+            case BadgeType.locked:
+                return (
+                    <LockedBadge
+                        className={this.props.className}
+                        description={this.props.description}
+                    />
```
nicholasrice: These three badges are very similar and expect the same data - seems a bit heavy to me to create 3 badges and this `BadgeControl` when you can have just one component. Is there a reason we need four?

janechu: That's true, I can consolidate them

PR #1571, opened 2019-03-21
chore: remove requirement to start pr descriptions with explicit words

chrisdholt:
# Description
This PR updates the pull request template to remove explicit word requirements in `description` details. I believe that the primary requirements are listed below. I believe the removed text which explicitly requires the use of "add", "remove", "update" limits a contributors ability to provide a detailed explanation. While those words may typically fit many scenarios, trying to ensure all descriptions use a specific set of words is tedious and limits the ability to be concise and avoid redundancy.

Description is ultimately secondary to the required types of `fix`, `feat`, `chore`.

```
    <description> is required and speaks to what the user gets from this PR:

        - Be concise.
        - Use all lowercase characters. 
        - Use imperative, present tense (e.g. `add` not `adds`.)
        - Do not end your description with a period.
        - Avoid redundant words.
 ```

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -48,14 +48,13 @@ Example titles:
         - fix: A change which fixes an issue.
         - feat: A that adds functionality.
 
-    <description> is required and speaks to what the user gets from this PR:
+    <description> is required for the CHANGELOG and speaks to what the user gets from this PR:
 
         - Be concise.
         - Use all lowercase characters. 
         - Use imperative, present tense (e.g. `add` not `adds`.)
         - Do not end your description with a period.
         - Avoid redundant words.
-        - Start your description with `add`, `update`, or `remove.`
 
 For additional information regarding working on FAST-DNA, check out our documentation site:
 https://microsoft.github.io/fast-dna/docs/en/contributing/working


chrisdholt:
> I'm opposed to changing this as I believe it's no extra overhead and provides better long term flexability in the experience we design for presenting changelogs to our users. The ultimate goal was to have a single all up changelog that could be organized, filtered, and easy to scan and look beautiful. I'd much rather read a change log where I go view by what features have been added, removed, or updated. I don't think other OSS does this, but, I think they should.

I think there is a fundamental delta between restricting language used to begin a description and long term flexibility. 

> How is a contributor "limited" in their ability to provide a detailed explanation by using add, update, or remove? The first bullet says to "be concise" which contradicts provided a detailed explanation in the title.

If there is a better and more concise word to describe something more robustly rather than "add", "update", or "fix", should we not use that? Only requiring three words to begin a sentance / fragment is by its very nature "limited". This is specifically useful in chore's.
> 
> Do you have an example of where it limits ones ability to explain the work?

While refactor is not a proper title, it is verbose and descriptive when describing large scale changes. There recently was a contribution where most of the work was really just cleaning things up. We should be able to have a PR description such as:

`chore: clean up ________ to be more readable` 
or
`chore: replace Object.assign with spread operator in ___________`.

> Shouldn't a detailed explanation be added added within the PR itself?
> If something were to change I think we should add additional bullets that make it clear that the title should be written to describe what the user gets, not, what is broken.

Title is not description. Title per our docs is "feat", "fix", "chore" because we use conventional commits. Title has functional requirements. The description doesn't provide any additional value besides preference. Getting more nitty with semantics, but technically, any change to existing code is an update. That in and of itself provides no descriptive value and may be considered redundant when followed by other descriptors.

nalogiudice:
I agree with @chrisdholt 's overall assessment - I always through the enforcement of beginning the statement with certain words was always too strict and unnecessary. We already enforce the text before the colon, which seems like enough to me. Being able to start the statements with other words would help them be more descriptive.

janechu:
I am OK with this change, the example given by @chrisdholt shows what trying to reduce messaging to use certain words can lead. The functionally important part of this is the type for determining semver. Also, I haven't come across very many open source projects that have to over-complicate the commit messages. As long as the messaging is clear and concise for the `CHANGELOG.md`, which we can comment on at the time a PR is reviewed.

awentzel:
@chrisdholt, this is ok to merge.

PR #1572, opened 2019-03-21
fix: give components unique display names

scomea:
# Description
We were giving identical display names to both base components and the styled MSFT components.  This change prepends the display names of components according to which package they come from.

I also added the ability to exclude subdirectories by name from "includesAllSubdirectoriesAsNamedExports" tests.

## Motivation & context
Having identical display names for different components made some tasks difficult.  Using display name as a query selector in tests, for example, is not nearly as useful if the display name is not unique per component.

## Issue type checklist

Yes, any code that references display name will break as the value has changed. Probably uncommon outside of tests.

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/utilities/index.ts
+++ b/packages/fast-components-react-msft/src/utilities/index.ts
@@ -1,0 +1,3 @@
+const DisplayNamePrefix: string = "MSFT";
+
+export { DisplayNamePrefix };
--- a/build/helpers/file-includes-all-subdirectories-as-named-exports.ts
+++ b/build/helpers/file-includes-all-subdirectories-as-named-exports.ts
@@ -3,9 +3,9 @@ import fs from "fs";
 import { pascalCase } from "../../packages/fast-web-utilities";
 
 /**
- * Verifies all components in source directory are exported
+ * Verifies all components in source directory are exported, directories listed in "excludes" parameter are ignored
  */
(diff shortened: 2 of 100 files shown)


packages/fast-components-react-msft/src/utilities/index.ts
```diff
@@ -0,0 +1,4 @@
+const DisplayNamePrefix: string = "Msft";
```
nalogiudice: Do we want `Msft` or `MSFT`? (the later seems more correct to me)

packages/fast-components-react-msft/src/utilities/index.ts
```diff
@@ -0,0 +1,4 @@
...
+
+export default DisplayNamePrefix;
```
chrisdholt: Do we need this default export? I think just the named export is fine.

packages/fast-components-react-msft/src/utilities/index.ts
chrisdholt: MSFT seems best here IMO.

scomea: MSFT it is

packages/fast-components-react-msft/src/utilities/index.ts
scomea: removed

PR #1573, opened 2019-03-21
feat: update styles and design system density.

bheston:
# Description

- Updated component style to match latest designs as necessary.
- Migrated common values like padding, border widths, and corner radius into the design system.
- Added density support to the design system to adjust font size, padding, and height as configured.

## Motivation & context

Gets components in alignment with the current design and adds support for higher-density uses through design system configuration.

## Issue type checklist

This change removes the "height" utility constants for default, min, and max height.

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -55,7 +55,7 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
 ): ComponentStyleSheet<CallToActionClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
     const direction: Direction = designSystem.direction;
-    const xTranslatePx: string = toPx(designSystem.designUnit);
+    const translateXValue: string = toPx(designSystem.designUnit);
 
     return {
         callToAction: {
@@ -69,16 +69,16 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
                 "& $callToAction_glyph": {
                     transform:
                         direction === Direction.ltr
-                            ? `translateX(${xTranslatePx})`
-                            : `rotate(180deg) translateX(${xTranslatePx})`,
+                            ? `translateX(${translateXValue})`
+                            : `rotate(180deg) translateX(${translateXValue})`,
                     position: "relative",
                 },
             },
             ...applyFocusVisible("& $callToAction_glyph", {
                 transform:
                     direction === Direction.ltr
-                        ? `translateX(${xTranslatePx})`
-                        : `rotate(180deg) translateX(${xTranslatePx})`,
+                        ? `translateX(${translateXValue})`
+                        : `rotate(180deg) translateX(${translateXValue})`,
                 position: "relative",
             }),
         },
@@ -90,7 +90,7 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             [applyLocalizedProperty("marginLeft", "marginRight", direction)]: "6px",
             transform: direction === Direction.ltr ? "none" : "rotate(180deg)",
             transition: "all 600ms cubic-bezier(0.19, 1, 0.22, 1)",
-            marginTop: direction === Direction.ltr ? xTranslatePx : "0",
+            marginTop: direction === Direction.ltr ? translateXValue : "0",
         },
         callToAction__primary: {
             "& $callToAction_glyph": {
@@ -103,7 +103,11 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = (
             },
         },
         callToAction__justified: {
-            paddingRight: xTranslatePx,
+            [applyLocalizedProperty(
+                "paddingRight",
+                "paddingLeft",
+                direction
+            )]: translateXValue,
             "& $callToAction_glyph": {
                 fill: accentFillRest,
             },
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -11,13 +11,10 @@ import {
     applyFocusVisible,
     applyLocalizedProperty,
     Direction,
+    toPx,
 } from "@microsoft/fast-jss-utilities";
 import { DesignSystem, withDesignSystemDefaults } from "../design-system/index";
-import {
-    accentFillRest,
-    accentForegroundCut,
-    neutralForegroundRest,
-} from "../utilities/color";
+import { accentFillRest, accentForegroundCut } from "../utilities/color";
 
 // Since MSFT button is already styled, we need to override in this way to alter button classes
 export const callToActionButtonOverrides: ComponentStyles<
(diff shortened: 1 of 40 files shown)


packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -36,9 +33,10 @@ export const callToActionButtonOverrides: ComponentStyles<
             "& $button_contentRegion": {
                 transform: (config: DesignSystem): string => {
                     const designSystem: DesignSystem = withDesignSystemDefaults(config);
+                    const xTranslatePx: string = toPx(designSystem.designUnit);
```
chrisdholt: From a principled standpoint, is this always relative to the design unit, or is it mapped because they're the same number currently?

packages/fast-components-styles-msft/src/context-menu-item/index.ts
```diff
@@ -13,52 +10,62 @@ import {
...
-        ...applyTypeRampConfig("t7"),
-        borderRadius: (config: DesignSystem): string => {
-            const designSystem: DesignSystem = withDesignSystemDefaults(config);
+const styles: ComponentStyles<ContextMenuItemClassNameContract, DesignSystem> = (
+    config: DesignSystem
+): ComponentStyleSheet<ContextMenuItemClassNameContract, DesignSystem> => {
+    const designSystem: DesignSystem = withDesignSystemDefaults(config);
+    const padding: number =
+        horizontalSpacingNumber(-2)(designSystem) +
+        16 +
```
chrisdholt: should we define this number as a const so its purpose is understood? It seems like an arbitrary number sandwiched between these two.

packages/fast-components-styles-msft/src/utilities/typography.ts
```diff
@@ -105,3 +110,32 @@ export function applyTypeRampConfig(typeConfig: keyof TypeRamp): CSSRules<Design
...
     };
 }
+
+/**
+ * Adds the font size and line height attributes for the type ramp, scaled according to the design system density config.
+ *
+ * @param config The design system config.
+ * @param typeConfig The default type ramp config at base density.
+ */
+export function scaleApplyTypeRampConfigWithDensity(
```
chrisdholt: I know you guys talked about this a bit, but I think something like `scaleTypography` may be better to use in development. I also think we want something like this without passing the design system:
```
scaleTypography("t7")
```

should return:
```
{
   fontSize: (designSystem: DesignSystem): string => {},
   lineHeight: (designSystem: DesignSystem): string => {},
}
```

If need be, we can do this as a follow-up - although, I think we should land the name...that way we're just changing internals.

packages/fast-components-styles-msft/src/context-menu-item/index.ts
```diff
@@ -13,52 +10,62 @@ import {
...
+            listStyleType: "none",
+            boxSizing: "border-box",
+            height: height(),
+            display: "grid",
+            gridTemplateColumns: `${toPx(padding)} auto ${toPx(padding)}`,
+            gridTemplateRows: "auto",
+            justifyItems: "center",
+            alignItems: "center",
+            padding: "0",
+            margin: "0 4px",
```
chrisdholt: should this be `designSystem.designUnit` similar to the translate for CTA? 

What about horizontalSpacing utility? This technically seems to be providing horizontal spacing, is there a reason why it is not affected by density?

packages/fast-components-styles-msft/src/context-menu-item/index.ts
```diff
@@ -13,52 +10,62 @@ import {
...
+            color: neutralForegroundRest,
+            fill: neutralForegroundRest,
+            whiteSpace: "nowrap",
+            overflow: "hidden",
+            cursor: "default",
```
chrisdholt: `...applyCursorDefault()`?

packages/fast-components-styles-msft/src/flipper/index.ts
```diff
@@ -31,52 +29,49 @@ const styles: ComponentStyles<FlipperClassNameContract, DesignSystem> = (
...
     return {
         flipper: {
-            width: "40px",
-            height: "40px",
+            width: toPx(
+                (designSystem.baseHeightMultiplier + 2) * designSystem.designUnit
+            ),
+            height: toPx(
+                (designSystem.baseHeightMultiplier + 2) * designSystem.designUnit
+            ),
```
chrisdholt: While these are currently set at 40px, the ( + 2) seems like an arbitrary number to get us to the current size of 40px. I think our choice of `40px` was initially to hit a good middle ground that would work for touch. If we are driving a big part of the touch story through density, could we remove the (+2) for the default here and just have 36px be the new width/height?

packages/fast-components-styles-msft/src/pivot/index.ts
```diff
@@ -63,12 +63,14 @@ const styles: ComponentStyles<PivotClassNameContract, DesignSystem> = (
...
+            ...scaleApplyTypeRampConfigWithDensity(designSystem, "t7"),
+            position: "relative",
+            top: "-2px",
         },
         pivot_activeIndicator: {
             position: "absolute",
-            borderRadius: toPx(designSystem.cornerRadius),
-            top: density(27),
+            ...applyCornerRadius(designSystem),
+            top: toPx(heightNumber(1)(designSystem) - 3 - 2) /* height & padding */,
```
chrisdholt: are 3 and 2 significant here and representative of something? I'd recommend either setting as a const so it's clear, or just writing "5".

packages/fast-components-styles-msft/src/call-to-action/index.ts
bheston: Always. It's one unit of space from the text, similar to the two or three units we use for other horizontal spacing depending on density. I don't think this needs to adjust based on density though, even as the font size increases or decreases.

packages/fast-components-styles-msft/src/context-menu-item/index.ts
bheston: Sure. It represents the icon width. I considered pulling that out, but we don't have a story around scaling icons. I'll keep it local to this file, and maybe someday we'll address whether those scale with density or not.

packages/fast-components-styles-msft/src/context-menu-item/index.ts
bheston: Yes, I think it is designUnit. I've  updated this and a few other locations. With horizontalSpacing we're adjusting the number of designUnits with density, not the value. In the case of 4px I think it will always be only one unit as it's intended to be small, essentially just a little ridge margin in this and many other cases.

packages/fast-components-styles-msft/src/flipper/index.ts
bheston: It would end up being 32px. I do agree that we're addressing touch through density, but it seems common that this control is usually larger than a standard control. Everywhere else we use it (microsoft.com, xbox.com, News) it's already 44px. I'd like to take this an an exception, similar to how badge is always a little smaller.

packages/fast-components-styles-msft/src/pivot/index.ts
bheston: Updated to use declarative constants.

packages/fast-components-styles-msft/src/flipper/index.ts
chrisdholt: I believe those are 44px though again because they don't have a way to manage density. Our system does - should we consider that? Won't these get significantly larger when density is enabled to support touch?

chrisdholt: @bheston I think this is still unresolved for me, but I'm not going to block on it as changing the size shouldn't be a breaking change. I would like to have a follow-up on this though

PR #1602, opened 2019-03-29
fix: added support for glyphs to resize based on density

### LINKED ISSUE #1589: Icons do not scale with density in FAST components MSFT
In many components users can pass icons, SVG's, or before/after content. When moving the density slider, while the component and text scale, the SVG's do not. We should ensure that we  scale the icons as well.

- Action Toggle
- Action Trigger
- Call To Action
- Text action
- others?

Example:

1. Go to https://explore.fastdna.net/components/action-toggle/ or https://explore.fastdna.net/components/call-to-action/
2. Scale the density down
3. See that the icon is not scaled in proportion with the text

The icon should scale

### LINKED ISSUE #1590: Align flipper [MSFT components] size to other controls and support density
Flippers are currently sized at 40px, which is an outlier from other controls. The initial reasoning to go with 40px was to hit a good middle ground that would work for touch. Initially we were looking at having two different sizes, and opted for one in the middle.

Since we are driving the touch story through density, it makes sense to have flipper align with other controls and respond to the density story in a similar fashion.

bheston:
# Description

Added density helpers for determining glyph size and updated all components with included glyphs.

## Motivation & context

Resolves #1589 
Resolves #1590

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/flipper/flipper.tsx
```diff
@@ -75,6 +80,34 @@ class Flipper extends Foundation<FlipperHandledProps, FlipperUnhandledProps, {}>
 
         return coercedProps;
     }
+
+    private renderNextGlyph(): React.ReactNode {
+        if (this.props.direction === FlipperDirection.next) {
```
chrisdholt: what if we generated the path based on direction instead of the entire SVG?

chrisdholt: Then we could simply have a single function to generate the glyph

packages/fast-components-react-msft/src/flipper/flipper.tsx
```diff
@@ -14,10 +14,15 @@ import {
     ManagedClasses,
 } from "@microsoft/fast-components-class-name-contracts-msft";
 import { DisplayNamePrefix } from "../utilities";
+import { FlipperProps } from ".";
```
chrisdholt: FYI - I've seen this create pathing issues. I would path this directly to the props file.

packages/fast-components-react-msft/src/flipper/flipper.tsx
bheston: Sounds good.

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -15,6 +15,32 @@ import {
 } from "@microsoft/fast-jss-utilities";
 import { DesignSystem, withDesignSystemDefaults } from "../design-system/index";
 import { accentFillRest, accentForegroundCut } from "../utilities/color";
+import { glyphSize } from "../utilities/density";
+
+function applyContentRegionTransform(): CSSRules<DesignSystem> {
+    return {
+        transform: (config: DesignSystem): string => {
```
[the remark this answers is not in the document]
bheston: These blocks are different, but I cleaned up the code a little bit.

packages/fast-components-styles-msft/src/auto-suggest/index.ts
```diff
@@ -1,18 +1,17 @@
...
 import { neutralFillStealthRest } from "../utilities/color";
-import { curry } from "lodash-es";
+import { heightNumber } from "../utilities/density";
+import { elevation, ElevationMultiplier } from "../utilities/elevation";
 
 const styles: ComponentStyles<AutoSuggestClassNameContract, DesignSystem> = (
     config: DesignSystem
 ): ComponentStyleSheet<AutoSuggestClassNameContract, DesignSystem> => {
     const designSystem: DesignSystem = withDesignSystemDefaults(config);
+    const visibleOptionCount: number = 10;
```
janechu: Can we give this a better name, perhaps `visibleLines` if that's what these are

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -105,6 +103,41 @@ export function horizontalSpacingNumber(
...
+            const value: number = glyphSizeNumber()(designSystem);
+            return typeof unit === "string" ? `${value}${unit}` : toPx(value);
+        }
+    );
+}
+
+/**
+ * Returns the width and height for an icon as a number.
+ */
+export function glyphSizeNumber(): DesignSystemResolver<number> {
```
janechu: I think there would be a better name for this maybe `getGlyphPixelWidthAsNumber`

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -105,6 +103,41 @@ export function horizontalSpacingNumber(
     );
 }
 
+/**
+ * Returns the width and height for an icon formatted in the provided unit or px by default.
+ *
+ * @param unit The unit of measurement; px by default.
+ */
+export function glyphSize(unit?: string): DesignSystemResolver<string> {
```
janechu: Can we name this something more obvious, like  `getGlyphPixelWidthAsCSSStringValue`

packages/fast-components-styles-msft/src/auto-suggest/index.ts
```diff
@@ -1,18 +1,17 @@
...
-} from "../design-system";
+import { DesignSystem, withDesignSystemDefaults } from "../design-system";
 import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
 import { AutoSuggestClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
-import { elevation, ElevationMultiplier } from "../utilities/elevation";
 import { toPx } from "@microsoft/fast-jss-utilities";
+import { applyFloatingCornerRadius } from "../utilities/border";
 import { neutralFillStealthRest } from "../utilities/color";
-import { curry } from "lodash-es";
+import { heightNumber } from "../utilities/density";
```
janechu: just a note, we really need to change this to be more descriptive

packages/fast-components-styles-msft/src/utilities/density.ts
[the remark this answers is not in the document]
bheston: It follows the pattern I started with the other functions in this file. height and horizontalSpacing all work the same. Technically they all "scale", but were written before we named the type ramp function. They are paired though, so the normal one returns the string, but the other returns the number. Do we have a convention for specifying return type?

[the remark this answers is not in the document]
nicholasrice: No - the function name doesn't need to speak to the return type - we use TypeScript for that.

We can keep alignent here if the rest of these are named like this.

packages/fast-components-react-msft/src/flipper/flipper.tsx
```diff
@@ -14,10 +14,15 @@ import {
     ManagedClasses,
 } from "@microsoft/fast-components-class-name-contracts-msft";
 import { DisplayNamePrefix } from "../utilities";
+import { FlipperProps } from "./flipper.props";
```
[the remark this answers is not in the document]
bheston: Doh!

packages/fast-components-styles-msft/src/auto-suggest/index.ts
bheston: The child object is "auto suggest option". Maybe visibleChildCount?

packages/fast-components-react-msft/src/flipper/flipper.tsx
```diff
@@ -34,9 +39,8 @@ class Flipper extends Foundation<FlipperHandledProps, FlipperUnhandledProps, {}>
                 {...this.coerceButtonProps()}
                 className={this.generateClassNames()}
             >
-                <span className={get(this.props, "managedClasses.flipper_glyph")}>
-                    {this.props.children}
-                </span>
+                {this.renderGlyph()}
+                {this.props.children}
```
janechu: wait, did we want to render children in here? I remember this being a bug, which we may was well fix here.

packages/fast-components-styles-msft/src/auto-suggest/index.ts
janechu: I think that works

packages/fast-components-styles-msft/src/auto-suggest/index.ts
bheston: Leaving as is per our conversation.

packages/fast-components-react-msft/src/flipper/flipper.tsx
bheston: The bug was that it (say text) was rotated with the fake glyph. The styling isn't correct as-is, but rendering children would allow someone to override the styling as necessary. It's probably not common, but I think it should still be supported.

PR #1603, opened 2019-03-29
feat: add Background component

nicholasrice:
# Description
Adds a Background component to make coordinating CSS & design-system backgrounds easier

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/background/README.md
```diff
@@ -0,0 +1,44 @@
+# Background
+
+The *background* component is used to create regions of a page that has a background color. The component will coordinate applying the supplied background color to both the created element *and* adjusting the `DesignSystem` for all child components.
```
nalogiudice: should be "...regions of a page that **_have_** "

packages/fast-components-react-msft/src/background/Background.props.ts
```diff
@@ -0,0 +1,45 @@
+import React from "react";
```
janechu: this file has a capital letter B, prefer lowercase.

packages/fast-components-react-msft/src/background/Background.tsx
```diff
@@ -0,0 +1,67 @@
...
+    public static defaultProps: Partial<BackgroundProps> = {
+        tag: "div",
+        value: LightModeBackgrounds.L1,
+    };
+    protected handledProps: HandledProps<Required<BackgroundHandledProps>> = {
+        tag: void 0,
+        value: void 0,
+    };
+
+    private getDesignSystemOverrides: (color: string) => Partial<DesignSystem> = memoize(
```
janechu: I think this should be after render

packages/fast-components-react-msft/src/background/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+#Usage
```
janechu: nit: usage should have two # because you can only have one h1, also needs a space after # or it will not render correctly.

packages/fast-components-react-msft/src/background/Background.tsx
nicholasrice: I think that because it is being assigned a value (not simply declared) the linter sees this as a property and not a method - leaving here because it yells if I don't

PR #1607, opened 2019-04-01
feat: add design system support for font weights

### LINKED ISSUE #306: Create an API or utility to apply variable font properties to components
We need an easy way to apply font properties to typographic components and potentially all components. Create an API or utility that allows for a global set of font properties to be set and easily manipulated from a component level.

Upon completion, the TODO found below should be fixed:

https://codeclimate.com/github/Microsoft/fast-dna/packages/fast-components-styles-msft/src/heading/index.ts#issue_5adfa15298a8bf00010001d9

chrisdholt:
closes #306 

This is a transparent change to how we manage font weights in order to provide support for variable font use. Instead of using internal constants, we expose a way to support user declared values. This allows users to pass different design system values for fontWeight depending on if they want to / can support variable fonts in a given experience. Additionally, this is an important move as we've been operating under an internal assumption that font weights are constant and defined but font family is not. Technically while certain font families support similar weights, our previous implementation was tightly coupled to the MSFT SegoeUI font family.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/badge/index.ts
+++ b/packages/fast-components-styles-msft/src/badge/index.ts
@@ -11,8 +11,8 @@ import { accentForegroundCut, neutralForegroundRest } from "../utilities/color";
 import { Swatch } from "../utilities/color/common";
 import { applyCursorDefault } from "../utilities/cursor";
 import { horizontalSpacing } from "../utilities/density";
-import { fontWeight } from "../utilities/fonts";
 import { applyScaledTypeRamp } from "../utilities/typography";
+import { applyFontWeightNormal, applyFontWeightSemiBold } from "../utilities/fonts";
 
 const styles: ComponentStyles<BadgeClassNameContract, DesignSystem> = (
     config: DesignSystem
@@ -27,11 +27,11 @@ const styles: ComponentStyles<BadgeClassNameContract, DesignSystem> = (
     return {
         badge: {
             ...applyScaledTypeRamp("t7"),
+            ...applyFontWeightSemiBold(),
             ...ellipsis(),
             overflow: "hidden",
             ...applyCursorDefault(),
             boxSizing: "border-box",
-            fontWeight: `${fontWeight.semibold}`,
             display: "inline-block",
             maxWidth: "215px",
             color: neutralForegroundRest,
@@ -39,7 +39,7 @@ const styles: ComponentStyles<BadgeClassNameContract, DesignSystem> = (
         },
         badge__filled: {
             ...applyCornerRadius(),
-            fontWeight: `${fontWeight.normal}`,
+            ...applyFontWeightNormal(),
             backgroundColor: filledBackground,
             color: accentForegroundCut((): Swatch => filledBackground),
         },
--- a/packages/fast-components-styles-msft/src/heading/index.ts
+++ b/packages/fast-components-styles-msft/src/heading/index.ts
@@ -1,36 +1,27 @@
-import { DesignSystem } from "../design-system";
+import { DesignSystem, withDesignSystemDefaults } from "../design-system";
 import { ComponentStyles, CSSRules } from "@microsoft/fast-jss-manager";
 import { HeadingClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
-import { fontWeight } from "../utilities/fonts";
+import { applyFontWeightSemiBold } from "../utilities/fonts";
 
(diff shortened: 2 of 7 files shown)


packages/fast-components-styles-msft/src/heading/index.ts
```diff
@@ -1,36 +1,27 @@
...
-}
-
-/**
- * TODO #306: Pull font weight styles when we have an API for font/variable font properties
- */
 const styles: ComponentStyles<HeadingClassNameContract, DesignSystem> = {
     heading: {
         "&$heading__1": {
-            ...applyHeadingStyles(),
+            ...applyFontWeight("bold"),
```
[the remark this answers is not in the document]
chrisdholt: yes - updated

packages/fast-components-styles-msft/src/utilities/fonts.ts
```diff
@@ -6,10 +13,27 @@ export interface FontWeight {
...
+export const defaultFontWeights: FontWeight = {
     light: 100,
     semilight: 200,
     normal: 400,
     semibold: 600,
     bold: 700,
 };
+
+/**
+ * @deprecated
```
nicholasrice: Can you note the replacement?

```
@deprecated - use defaultFontWeights
```

packages/fast-components-styles-msft/src/utilities/fonts.ts
```diff
@@ -6,10 +13,27 @@ export interface FontWeight {
...
+ */
+export const fontWeight: FontWeight = defaultFontWeights;
+
+function weight(index: keyof FontWeight): DesignSystemResolver<string> {
```
nicholasrice: nitty I know  - but its really more of a key than an index

chrisdholt: yeesh

packages/fast-components-styles-msft/src/badge/index.ts
```diff
@@ -27,19 +27,19 @@ const styles: ComponentStyles<BadgeClassNameContract, DesignSystem> = (
...
-            fontWeight: `${fontWeight.semibold}`,
             display: "inline-block",
             maxWidth: "215px",
             color: neutralForegroundRest,
             transition: "all 0.2s ease-in-out",
         },
         badge__filled: {
             ...applyCornerRadius(),
-            fontWeight: `${fontWeight.normal}`,
+            ...applyFontWeight("normal"),
```
janechu: I think this should be `fontWeight.normal` and not a string

packages/fast-components-styles-msft/src/heading/index.ts
```diff
@@ -1,36 +1,27 @@
...
-}
-
-/**
- * TODO #306: Pull font weight styles when we have an API for font/variable font properties
- */
 const styles: ComponentStyles<HeadingClassNameContract, DesignSystem> = {
     heading: {
         "&$heading__1": {
-            ...applyHeadingStyles(),
+            ...applyFontWeight("semibold"),
```
janechu: I think these should all be enums

packages/fast-components-styles-msft/src/subheading/index.ts
```diff
@@ -1,33 +1,27 @@
...
-        fontWeight: `${fontWeight.normal}`,
-    };
-}
+import { applyFontWeight } from "../utilities/fonts";
 
 const styles: ComponentStyles<SubheadingClassNameContract, DesignSystem> = {
     subheading: {
         "&$subheading__1": {
-            ...applySubheadingStyles(),
+            ...applyFontWeight("normal"),
```
janechu: same here, prefer to use enums

packages/fast-components-styles-msft/src/heading/index.ts
chrisdholt: Adding @nicholasrice for his thoughts; most of the ways we are referencing interface keys is with strings. I'm open to using enums, but then we need an enum that maps to the interface, yes?

nicholasrice: Hmmm - yea that would be a relationship we would need to keep in sync - though I don't think that would be particularly challenging. Enums *are* nice, but you would also then need to import two dependencies to make the utility work which seems a *little* overkill.

A middle-ground might be to export named functions - `applyFontWeightBold` vs `applyFontWeightNormal` - at least then we're not passing the same string around everywhere.

Issue #1598, opened 2019-03-28
Add "stealth" appearance to MSFT button


eljefe223:
**What feature are you requesting?**
We should add a `stealth` option to the `<Button/>` `appearance` prop.

**What would be the benefit of adding this feature?**
Stealth buttons are a common design pattern throughout Microsoft's UI. Currently we need to implement CSS properties to mimic this pattern every time we want a stealth button.

nicholasrice:
I'm wondering if we want to separate button appearances into separate controls.

The `Button` stylesheet is already pretty large, and adding another appearance option only adds to it. In general, I don't think changing between appearances is a common use-case, so in most cases we're compiling style rules that don't ever get used (for the other appearances)

chrisdholt:
@nicholasrice I love that idea. I think it may require us revisiting the need for our button derivatives, which would be good as well.

PR #1616, opened 2019-04-03
feat: add stealth appearance to button appearance enum

eljefe223:
# Description
Added `stealth` appearance to `<Button/>` `appearance` props.
Closes: #1598 

## Motivation & context
Stealth is a reoccurring pattern we see throughout Microsoft UI. This allows us to quickly repeat the pattern without having to copy styles over and over.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-class-name-contracts-msft/src/button/index.ts
+++ b/packages/fast-components-class-name-contracts-msft/src/button/index.ts
@@ -25,6 +25,11 @@ export interface ButtonClassNameContract extends BaseButtonClassNameContract {
     button__justified?: string;
 
     /**
+     * The stealth appearance modifier
+     */
+    button__stealth?: string;
+
+    /**
      * The button content region
      */
     button_contentRegion?: string;
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -183,6 +183,9 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = (
             borderWidth: "0",
             justifyContent: "flex-start",
         },
+        button__stealth: {
+            background: "purple",
+        },
         button_contentRegion: {
             position: "relative",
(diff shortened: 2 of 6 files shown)


packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -177,6 +177,21 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = (
             borderWidth: "0",
             justifyContent: "flex-start",
         },
+        button__stealth: {
+            minWidth: "32px",
+            maxWidth: "unset",
+            padding: "0",
```
bheston: We shouldn't need to change anything other than colors here. They layout should be the same as the primary, secondary, outline, and lightweight styles.

eljefe223: I think its open for discussion. The most common scenario for stealth button is it only displaying an Icon. The idea behind creating this appearance was to not have to override the width, padding and other properties every time we wanted this type of button. If someone wants a stealth button that displays like the other buttons they can just as easily add those overrides but that is not a common scenario.

packages/fast-components-class-name-contracts-msft/src/button/index.ts
```diff
@@ -24,6 +24,11 @@ export interface ButtonClassNameContract extends BaseButtonClassNameContract {
      */
     button__justified?: string;
 
+    /**
+     * The stealth appearance modifier
```
nalogiudice: I wonder if we can add what "stealth" actually means here? Coming from the outside and not being a part of this design conversation, I'm not quite sure what stealth really means.

eljefe223: @nalogiudice I guess one could argue the same for the other appearances. This aligns with our color system. Maybe we need to add something in the Read Me that describes when to use each kind of appearance.

PR #1617, opened 2019-04-03
feat: add public function to breakpoint tracker to get current breakpoint

chrisdholt:
# Description
Currently there is no way to gather the current breakpoint value without an update. This PR exposes a public function which provides access to the current breakpoint at any given time without needing to trigger an update.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
+++ b/packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
@@ -92,6 +92,13 @@ class BreakpointTracker {
     };
 
     /**
+     * Returns the current breakpoint
+     */
+    public currentBreakpoint = (): Breakpoint => {
+        return this.breakpoint;
+    };
+
+    /**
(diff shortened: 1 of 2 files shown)


packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -91,6 +91,13 @@ class BreakpointTracker {
         this.openRequestAnimationFrame = false;
     };
 
+    /**
+     * Returns the current breakpoint
+     */
+    public getCurrentBreakpoint = (): Breakpoint => {
```
nicholasrice: naming nit - not sure you need the 'get' part of the method - I think `currentBreakpoint` makes it pretty clear it is returning the current breakpoint

chrisdholt: 👍

PR #1620, opened 2019-04-04
feat: add DesignSystemResolver formatting utility

nicholasrice:
# Description
The formatter allows us to format an input string with designSystemResolvers, meaning we don't need access to a DesignSystem value to format properties like border. This will help enable us to move away from function stylesheets, which are less performant than object stylesheets when design-system values change at runtime.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/context-menu-item/index.ts
+++ b/packages/fast-components-styles-msft/src/context-menu-item/index.ts
@@ -1,5 +1,9 @@
-import { DesignSystem, withDesignSystemDefaults } from "../design-system";
-import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
+import { applyCornerRadius, applyFocusPlaceholderBorder } from "../utilities/border";
+import {
+    DesignSystem,
+    DesignSystemResolver,
+    withDesignSystemDefaults,
+} from "../design-system";
 import { ContextMenuItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 import { height, horizontalSpacingNumber } from "../utilities/density";
 import {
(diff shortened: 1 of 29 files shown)


packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -60,9 +62,14 @@ const styles: ComponentStyles<TextActionClassNameContract, DesignSystem> = (
...
-                designSystem
-            )}`,
+            border: format(
+                "{0} solid {1}",
+                toPx<DesignSystem>(outlineWidth),
+                neutralOutlineRest
+            ),
+            // border: `${toPx(designSystem.outlineWidth)} solid ${neutralOutlineRest(
+            //     designSystem
+            // )}`,
```
janechu: still needed?

packages/fast-components-styles-msft/src/utilities/design-system.ts
```diff
@@ -0,0 +1,45 @@
...
+                return designSystem.outlineWidth * arg;
+            }
+        );
+    }
+
+    return withDesignSystemDefaults(arg).outlineWidth;
+}
+
+/**
+ * Retrieve the outlineWidth when invoked with a DesignSystem
```
janechu: outlineWidth? Think you might mean designUnit here

chrisdholt: Can we update this to reflect the function below?

packages/fast-components-styles-msft/src/utilities/format.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            formatter(designSystemDefaults) + formatter(designSystemDefaults)
+        );
+    });
+    test("should not format a variable when not supplied a formatter", (): void => {
+        expect(format("{0}{1}", formatter)(designSystemDefaults)).toBe(
+            formatter(designSystemDefaults) + "{1}"
+        );
+    });
+    test("should format multiple variables when multiple formatters are passed", (): void => {
+        expect(format("{0}{1}", formatter, formatter)(designSystemDefaults)).toBe(
```
janechu: I would suggest using a different formatter for the second formatter, just to make sure there isn't an issue where the wrong formatter gets used

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -60,9 +62,14 @@ const styles: ComponentStyles<TextActionClassNameContract, DesignSystem> = (
...
             minWidth: "92px",
-            border: `${toPx(designSystem.outlineWidth)} solid ${neutralOutlineRest(
-                designSystem
-            )}`,
+            border: format(
+                "{0} solid {1}",
+                toPx<DesignSystem>(outlineWidth),
+                neutralOutlineRest
+            ),
+            // border: `${toPx(designSystem.outlineWidth)} solid ${neutralOutlineRest(
```
chrisdholt: I think we can remove this

packages/fast-components-styles-msft/src/context-menu-item/index.ts
```diff
@@ -11,63 +15,64 @@ import {
...
+            color: neutralForegroundActive,
+            background: neutralFillStealthActive,
         },
-    };
+    },
+    contextMenuItem_contentRegion: {
+        gridColumnStart: "2",
+        justifySelf: "start",
+        overflow: "hidden",
+        textOverflow: "ellipsis",
```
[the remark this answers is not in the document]
nicholasrice: this is a copy+pasta so I'll leave this as-is

PR #1662, opened 2019-04-16
feat: allow defaults to be set as the current value

### LINKED ISSUE #1651: Form should be able to be set value to the current default
There should be the ability to set a value to the current default. This could be a button, as an action item it should not be in the same space as badges and labels.

Questions to address:
- Will this be optional?

janechu:
# Description

Allows the setting of a default value as the current value.

## Motivation & context

Closes #1651

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/style/utilities.ts
```diff
@@ -51,6 +51,31 @@ export function applyTriggerStyle(color: string): CSSRules<{}> {
     };
 }
 
+export function applyInteractiveFormItemIndicator(): CSSRules<{}> {
```
nicholasrice: Can this just be an object?

janechu: Actually, good point, these can all just be objects, I filed a task to convert all of them here: https://github.com/Microsoft/fast-dna/issues/1692

PR #1666, opened 2019-04-16
chore: added utility functions for density category and type sizes

bheston:
# Description

Added getOffsetForDensityCategory to simplify sizing calculations based on density category like compact or spacious.
Added ability to get type ramp font size and line height as an unformatted number for use in calculations.

## Motivation & context

Found repeated need for density category adjustments and a new need for determining sizing based on type parameters.

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -61,6 +61,33 @@ export function getDensityCategory(designSystem: DesignSystem): DensityCategory
...
+/**
+ * Returns an offset based on the higher-level category for the density setting.
+ * Used to adjust things like type size and sizing that is based on the category rather than individual density.
+ *
+ * @param compactOffset The adjustment when the category is "compact"
+ * @param spaciousOffset The adjustment when the category is "spacious"
+ * @param normalOffset The adjustment when the category is "normal"
+ */
+export function getOffsetForDensityCategory(
+    compactOffset: number,
```
nicholasrice: Can you write some tests for this to validate that it's doing what it should?

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -52,15 +52,41 @@ export function heightNumber(lines: number = 1): DesignSystemResolver<number> {
...
 
+/**
+ * Returns an offset based on the higher-level category for the density setting.
+ * Used to adjust things like type size and sizing that is based on the category rather than individual density.
+ *
+ * @param compactOffset The adjustment when the category is "compact"
+ * @param spaciousOffset The adjustment when the category is "spacious"
+ * @param normalOffset The adjustment when the category is "normal"
+ */
+export function getOffsetForDensityCategory(
```
nicholasrice: maybe `densityCategorySwitch`  - then this can accept any generic value and return that value depending on the density category

packages/fast-components-styles-msft/src/utilities/density.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+        expect(getOffsetForDensityCategory(3, 2, 1)({} as DesignSystem)).toBe(1);
+        expect(horizontalSpacing()({} as DesignSystem)).toBe("12px");
+        expect(horizontalSpacing(2)({} as DesignSystem)).toBe("10px");
+        expect(horizontalSpacing(2, "em")({} as DesignSystem)).toBe("10em");
+        expect(glyphSize({} as DesignSystem)).toBe("16px");
+        expect(glyphSize("em")({} as DesignSystem)).toBe("16em");
+    });
+
+    test("should operate on spacious design system", (): void => {
+        expect(height()(spaciousDesignSystem)).toBe("55px");
```
janechu: Should we be checking for exact values? Any change in the design system values will result in having to update all of these tests. I would think that a more flexible way to go is to make sure that adding the spacious design system, compact design system etc would produce a different result than the default design system.

bheston: I thought about this and used hard coded values because otherwise the tests would just end up duplicating the same calculations in the methods. I suppose if I rewrote the calculation it would probably be a sufficient test though.

bheston: Updated

packages/fast-components-styles-msft/src/utilities/density.ts
```diff
@@ -52,15 +53,41 @@ export function heightNumber(lines: number = 1): DesignSystemResolver<number> {
...
+ * Returns a value based on the higher-level category for the density setting.
+ * Used to adjust things like type size and sizing that is based on the category rather than individual density.
+ *
+ * @param compactValue The adjustment when the category is "compact"
+ * @param spaciousValue The adjustment when the category is "spacious"
+ * @param normalValue The adjustment when the category is "normal"
+ */
+export function densityCategorySwitch<T = number>(
+    compactValue: T | DesignSystemResolver<T>,
+    spaciousValue: T | DesignSystemResolver<T>,
```
nicholasrice: I think that, because these are all required now, the argument order should be:
1. compact
2. normal
3. spacious

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -209,13 +209,24 @@ export const withDesignSystemDefaults: (config: Partial<DesignSystem>) => Design
...
  * Safely retrieves a single property from a design system
+ * @deprecated Use getDesignSystemValue instead.
  */
 export function getDesignSystemProperty(key: string): DesignSystemResolver<string> {
     return function(config: DesignSystem): string {
         return withDesignSystemDefaults(config)[key];
     };
 }
 
+/**
```
nicholasrice: I actually think these should live in `utilities/design-system` - they are utility functions after all

PR #1675, opened 2019-04-17
feat: add image palette extractor and advanced palette generation routines

eigenvector32:
# Description
add code which can analyze an image and extract a color palette from it
also add some functions that provide more advanced methods of generating color palettes from single colors

## Motivation & context

histogram, quantization, pixel-blob, pixel-box and palette-extractor are all used in extracting a palette of colors from a bitmap
the additions to color-palette allow for palettes to be generated where the base color is not in the exact middle of the palette and for building a larger palette out of a smaller one

## Issue type checklist

## Process & policy checklist


packages/fast-colors/src/color-quantization.ts
```diff
@@ -0,0 +1,197 @@
...
+import { ColorRGBA64 } from "./color-rgba-64";
+
+export interface QuantizedColor {
+    color: ColorRGBA64;
+    pixelCount: number;
+    colorVolume: number;
+}
+
+export interface QuantizeConfig {
+    // must be in the range [1,8]. Memory use increases as 4*2^(3*signifigantBits).
```
nicholasrice: I imagine this info in comments would be usefull for app authors - if you use JSDOC comments the comment will survive the typescript compile process and can be made availible to editors like VSCode.

```
/*
 * my comment
 */
```

It would be awesome if you could update the comments for publicly exported objects to be the above syntax

packages/fast-colors/src/color-quantization.ts
```diff
@@ -0,0 +1,197 @@
...
+    // sort order is population * colorVolume. This helps highly contrasting colors in a small area to show
+    // up in some of the final output.
+    fractionByPopulation: number;
+    // This predicate can be used to screen out undesirable colors from the final output
+    // eg: excluding colors with a pixelCount below a min value
+    isBoxValid: ((box: PixelBox) => boolean) | null;
+    // This predicate can be used to exlude pixels from the histogram.
+    // It is passed numbers in the range [0,255] in rgba order
+    // EG: excluding colors too close to pure white
+    isHistoPixelValid: ((pixel: number[]) => boolean) | null;
```
nicholasrice: Minor naming nit - we try to use the full word when there isn't really a standardized abbreviation - can we make this `isHistogramPixelValid`?

packages/fast-colors/src/color-palette.ts
```diff
@@ -158,10 +181,227 @@ export class ColorPalette {
...
+// 10:
+// 11: input 3
+// 12:
+// 13:
+// 14: input 4
+// 15:
+// 16: scaleColorDark
+export function centeredRescale(
+    input: ColorRGBA64[],
+    targetSize: number,
```
nicholasrice: Can we make this accept an object instead? That will allow us to add to this over time if we want without making a breaking change. It also makes it easier to use if I want to only use a dark scale color.

packages/fast-colors/src/pixel-blob.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private image: ImageData;
+
+    public readonly width: number;
+    public readonly height: number;
+    public readonly totalPixels: number;
+
+    public getPixel = (x: number, y: number): ColorRGBA64 => {
+        if (x < 0 || x >= this.width || y < 0 || y >= this.height) {
```
nicholasrice: I think this could use `getPixelRGBA`

```
const data = this.getPixelRGBA(x, y).map((value: number) => value / 255);

return new ColorRGBA64(data[0], data[1], data[2], data[3])

```

packages/fast-colors/src/histogram.ts
```diff
@@ -0,0 +1,105 @@
...
+// EG: for 5 signifigant bits the histogram is 128K while for 8 it is 64 megs.
+// CPU time increases linearly as pixelSkipping is reduced
+// isHistoPixelValid is an optional predicate which can screen out unwanted pixels from the
+// source data. EG: ignoring transparent pixels.
+// If the image source has more than 2^32 pixels (eg: a square image 65536x65536 in size) of the same color this code will break
+
+export class Histogram {
+    constructor(
+        source: PixelBlob,
+        signifigantBits: number = 5,
```
janechu: significant?

PR #1679, opened 2019-04-18
fix: improved examples for select option

scomea:
# Description
Examples didn't make clear the most common expected implementation

## Motivation & context
A developer was confused that the example was not the most common use case

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/select-option/examples.data.tsx
```diff
@@ -47,5 +47,8 @@ export default {
                 },
             } as any,
         },
+        {
+            children: "html children",
```
janechu: this seems confusing, wouldn't it just show the string "html children"? in which case its just a string child.

scomea: I wanted to simply show that children passed to the component would render, you mean I should have some other element here like a div?

janechu: I suppose it would be less confusing if it said something like "Hello world" or "Foo" because I'm trying to take some kind of meaning away from "html children" and it just reads as a confusing string.

scomea: "Hello world" it is.

PR #1688, opened 2019-04-18
feat: show a red underline for objects and arrays if they contain invalid data

### LINKED ISSUE #1663: Show errors for objects or arrays that contain errors
The user should be able to see errors that are happening downstream from an upstream location, that means there should be some sort of indication (such as a red outline) as well as inline text for the inline error prop option.

janechu:
# Description

Shows a red underline on objects and arrays if they are invalid or if they contain invalid data.

## Motivation & context

Closes #1663 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/form/form/form-item.array.style.ts
+++ b/packages/fast-tooling-react/src/form/form/form-item.array.style.ts
@@ -81,6 +81,10 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
     },
     formItemArray_controlLabel: {
         ...applyLabelStyle(),
+        borderBottom: "1px solid transparent",
+    },
+    formItemArray_controlLabel__invalid: {
+        borderColor: "red",
     },
     formItemArray_controlLabelRegion: {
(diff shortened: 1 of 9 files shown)


packages/fast-tooling-react/src/form/form/form-item.array.style.ts
```diff
@@ -80,6 +80,10 @@ const styles: ComponentStyles<FormItemArrayClassNameContract, {}> = {
     },
     formItemArray_controlLabel: {
         ...applyLabelStyle(),
+        borderBottom: "1px solid transparent",
```
eljefe223: No squiggly line? 😄 
https://stackoverflow.com/questions/17285514/is-it-possible-to-make-a-squiggly-line

janechu: Hahaha, that's actually pretty cool. We can update it if we feel it needs a little more "oomph" and the inferred meaning of a red squiggly isn't only regulated to spelling errors.

PR #1700, opened 2019-04-19
fix: handle selected items correctly in controlled mode

scomea:
# Description
In controlled mode (ie. selected items provided in props) the select control should:

- never actually update the selected items in state from user interaction, but should throw the selection changed event to allow parents to update and pass new props based on interaction
- state changes from updates to selected items props should not trigger selection changed events

The previous implementation wasn't doing this correctly.

## Motivation & context
Engineering encountered issues in implementation.
https://github.com/Microsoft/fast-dna/issues/1690

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -77,40 +79,34 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
...
 
-        if (updatedSelection !== undefined) {
-            this.updateSelection(updatedSelection);
-        }
-
         if (updatedMenuVisibility !== this.state.isMenuOpen) {
             this.toggleMenu(updatedMenuVisibility);
         }
+
+        if (prevProps.selectedItems !== this.props.selectedItems) {
```
nicholasrice: Should this return here? It kinda doesn't it make the effects of the previous checks moot?

PR #1716, opened 2019-04-25
fix: base select trigger on input field styles and not button styles

scomea:
# Description
This change should make it easier to keep select styles in line with the input element styles rather than worrying about correctly overriding button styles.

## Motivation & context
This should improve maintainability.  Styles were getting out of sync.

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -13,44 +15,53 @@ import {
...
-        width: "100%",
-        display: "inline-flex",
-    },
-};
+import { inputFieldStyles } from "../patterns/input-field";
+import { applyDisabledState } from "../utilities/disabled";
+import { focusOutlineWidth } from "../utilities/design-system";
 
 const styles: ComponentStyles<SelectClassNameContract, DesignSystem> = (
     config: DesignSystem
```
bheston: Since you don't need the designSystem anymore, you can clean up this nested return. Essentially get rid of the ComponentStyleSheet and the "return {". @nicholasrice will be happy :)

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -13,44 +15,53 @@ import {
...
+        select_button_contentRegion: {
+            display: "grid",
+            gridTemplateColumns: "1fr auto",
+            alignItems: "center",
+            justifyItems: "start",
+        },
+
+        select_button_displayText: {
+            whiteSpace: "nowrap",
+            overflow: "hidden",
```
bheston: I think there's a helper for this. We should probably use ellipsis.

packages/fast-components-styles-msft/src/select/index.ts
scomea: leaving for now per our discussion

packages/fast-components-styles-msft/src/select/index.ts
scomea: done

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -13,44 +15,53 @@ import {
...
+            textAlign: "left",
+            width: "100%",
+        },
+
         select_toggleGlyph: {
+            margin: "0 0 0 10px",
```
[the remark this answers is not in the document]
scomea: done

packages/fast-components-class-name-contracts-msft/src/select/index.ts
```diff
@@ -8,4 +8,19 @@ export interface SelectClassNameContract extends BaseSelectClassNameContract {
...
+
+    /**
+     * The select toggle button
+     */
+    select_button?: string;
+
+    /**
+     * The select toggle button content region
+     */
+    select_button_contentRegion?: string;
```
nicholasrice: I think this would be select_buttonContentRegion

packages/fast-components-class-name-contracts-msft/src/select/index.ts
```diff
@@ -8,4 +8,19 @@ export interface SelectClassNameContract extends BaseSelectClassNameContract {
...
+
+    /**
+     * The display text container
+     */
+    select_button_displayText?: string;
```
nicholasrice: Same pattern here: select_buttonDisplayText

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -13,44 +17,57 @@ import {
...
-        select__disabled: {
-            opacity: ".3",
-        },
+        select__disabled: {},
 
         select_menu: {
             background: neutralFillStealthRest,
             ...applyFloatingCornerRadius(),
-            ...elevation(ElevationMultiplier.e11)(designSystem),
+            ...elevation(ElevationMultiplier.e11)(withDesignSystemDefaults(config)),
```
nicholasrice: I think you can use the new `...applyElevation` here and avoid direct references to the `designSystem`

scomea: Yes, got that from the merge with Brian's changes, but still need designSystem to get direction.

packages/fast-components-class-name-contracts-msft/src/select/index.ts
scomea: yep

packages/fast-components-class-name-contracts-msft/src/select/index.ts
scomea: yep

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -13,59 +17,78 @@ import {
...
-        position: "absolute",
-        width: "100%",
-        margin: "0",
-        padding: format("{0} 0", toPx<DesignSystem>(designUnit)),
-        maxWidth: "374px",
-        minWidth: "276px",
-        maxHeight: "328px",
-        overflow: "auto",
-    },
+        select__disabled: {},
```
janechu: are there no disabled styles?

packages/fast-components-styles-msft/src/select/index.ts
scomea: (ie - we discovered we still need designsystem to get direction at this time)

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -13,59 +17,78 @@ import {
...
+            "& $select_menu": {
+                position: "static",
+                boxShadow: "none",
+                border: "1px solid",
+                borderColor: neutralOutlineRest,
+            },
         },
-    },
-    select_menu__open: {},
+        select_menu__open: {},
```
janechu: What was the reason for this class again?

packages/fast-components-styles-msft/src/select/index.ts
janechu: I'm re-opening this as the discussion was not had on an open thread, if the reasoning is that we need any part of the design system this is possible without making this a function, an example can be found here: https://github.com/Microsoft/fast-dna/blob/master/packages/fast-components-styles-msft/src/button/index.ts#L206

packages/fast-components-styles-msft/src/select/index.ts
scomea: It is intended for implementations where the trigger might look differently when the menu is open, ours doesn't, but doing things like rotating the icon when the menu is open are common.  So this implementation doesn't require this in the stylesheet and I'll remove it.

packages/fast-components-styles-msft/src/select/index.ts
scomea: We get those from ...inputFieldStyles() under "select_button".  So this implementation doesn't require this in the stylesheet so removed.

PR #1717, opened 2019-04-26
fix: updated elevation function to use the correct design system accessor structure

bheston:
# Description

Adds a new function to apply elevation in style sheets so we don't have to have a reference to the design system.

## Motivation & context

Fixes one of the few remaining places we still needed to reference the design system in style sheet code by changing to a function that is a DesignSystemResolver.

Added a new function to match the 'apply' naming convention of other similar utility functions, as well to be able to continue support and deprecate the previous function.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/select/index.ts
+++ b/packages/fast-components-styles-msft/src/select/index.ts
@@ -1,11 +1,11 @@
-import { DesignSystem, withDesignSystemDefaults } from "../design-system";
-import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
+import { DesignSystem } from "../design-system";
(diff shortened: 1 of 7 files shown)


packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -28,50 +28,44 @@ export const selectDisplayButtonOverrides: ComponentStyles<
...
-        },
+    select_toggleGlyph: {
+        fill: neutralForegroundRest,
+    },
 
-        select__disabled: {
-            opacity: ".3",
-        },
+    select__disabled: {
+        opacity: ".3",
```
nicholasrice: not you, but can you update this to `...applyDisabledState()`?

PR #1721, opened 2019-04-29
fix: deprecate id prop in listbox items

scomea:
# Description
Listbox items use an "id" prop to identify themselves. Because id attributes must be unique application wide this adds an extra burden to developers.  Since we only need an id that is unique within the scope of a particular list we will switch to use another property name - "itemId", which we will not write to the component's "id" attribute.

The "id" prop will continue to be supported until the next major version of the component library at which time it will be removed as a breaking change.

Additionally we fallback to use "value" as an identifier if no other id is provided. I'm admittedly on the fence about this bit as it makes the 90% case easier as most lists are composed of items with unique values anyway, but does it introduce unnecessary complexity? 

## Motivation & context
Developer feedback that unique id's were unnecessarily burdensome.
https://github.com/Microsoft/fast-dna/issues/1713

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/listbox-item/README.md
+++ b/packages/fast-components-react-base/src/listbox-item/README.md
@@ -2,7 +2,11 @@
 A component designed to work as a child of the *listbox*.
 
 ### Usage
-The component required "id" prop must be unique and is applied as an id to to the resulting element in the dom.  Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.
+All listbox items should have a itemId unique to the scope of the listbox.  If no itemId is provided the item is identified by its value.  If multiple list items could share the same value authors should provide itemId's or unexpected behavior may occur.
+
+Note: The "id" prop is being depracated, please use "itemId" instead.  Authors who have some need for the item's html "id" attribute to be populated should continue to do so, but will need to provide an itemId as well in a future update.  
+
+Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.
 
 ### Accessibility
 The *listbox-item* generates an html element with role="option" as well as "aria-selected" and "aria-disabled" attributes. 
--- a/packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
+++ b/packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import Foundation, { HandledProps } from "@microsoft/fast-components-foundation-react";
-import { get } from "lodash-es";
+import { get, isNil } from "lodash-es";
 import { ListboxItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 import {
     ListboxItemHandledProps,
@@ -34,9 +34,9 @@ class ListboxItem extends Foundation<
      * The "id" prop is being depracated and should not be used.
      */
     public static getItemId = (props: ListboxItemProps): string => {
-        return !ListboxItem.isNullOrUndefined(props.itemId)
+        return !isNil(props.itemId)
             ? props.itemId
-            : !ListboxItem.isNullOrUndefined(props.id)
+            : !isNil(props.id)
                 ? props.id
                 : props.value;
     };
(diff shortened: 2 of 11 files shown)


packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -24,25 +25,67 @@ class ListboxItem extends Foundation<
...
+
+        idToUse = element.getAttribute("id");
+        if (!ListboxItem.isNullOrUndefined(idToUse)) {
+            return idToUse;
+        }
+
+        return element.getAttribute("value");
+    };
+
+    public static isNullOrUndefined = (value: string): boolean => {
```
chrisdholt: Does this need to be a public static method? Can we just use `isNil` from lodash-es? https://lodash.com/docs#isNil

janechu:
Please use a lowercase "f" for "fix" in your merge message

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
scomea: Yes, got rid of that function entirely.

packages/fast-components-react-base/src/listbox-item/README.md
```diff
@@ -2,7 +2,11 @@
 A component designed to work as a child of the *listbox*.
 
 ### Usage
-The component required "id" prop must be unique and is applied as an id to to the resulting element in the dom.  Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.
+All listbox items should have a itemId unique to the scope of the listbox.  If no itemId is provided the item is identified by its value.  If multiple list items could share the same value authors should provide itemId's or unexpected behavior may occur.
+
+Note: The "id" prop is being depracated, please use "itemId" instead.  Authors who have some need for the item's html "id" attribute to be populated should continue to do so, but will need to provide an itemId as well in a future update.
```
janechu: spelling, "deprecated", "HTML", please use backticks instead of quotes for properties

packages/fast-components-react-base/src/listbox-item/README.md
```diff
@@ -2,7 +2,11 @@
 A component designed to work as a child of the *listbox*.
 
 ### Usage
-The component required "id" prop must be unique and is applied as an id to to the resulting element in the dom.  Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.
+All listbox items should have a itemId unique to the scope of the listbox.  If no itemId is provided the item is identified by its value.  If multiple list items could share the same value authors should provide itemId's or unexpected behavior may occur.
+
+Note: The "id" prop is being depracated, please use "itemId" instead.  Authors who have some need for the item's html "id" attribute to be populated should continue to do so, but will need to provide an itemId as well in a future update.  
+
+Developers should not examine *listbox-item* components to determine their selected state, but rather hook up to the "onSelectedItemsChanged" callback on the parent listbox.
```
janechu: I would use backticks for the callback and if the consensus is to bold components, be consistent and bolx listbox

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -24,25 +25,60 @@ class ListboxItem extends Foundation<
         disabled: false,
     };
 
+    /**
+     * Gets the id to use for the item
+     * Authors who's items all have unique values can choose to
+     * not provide an itemId and the values of the items will be used as id's,
+     * otherwise use itemId.  All listbox items of a particular listbox should use the same
+     * approach with respect to using values or itemId or results could be unexpected.
+     * The "id" prop is being depracated and should not be used.
```
janechu: spelling, deprecated

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -24,25 +25,60 @@ class ListboxItem extends Foundation<
...
+            : !isNil(props.id)
+                ? props.id
+                : props.value;
+    };
+
+    /**
+     * Gets the id to use for an listbox item by examining the associated Element
+     */
+    public static getItemIdFromElement = (element: Element): string => {
+        let idToUse: string = element.getAttribute("itemId");
```
janechu: This shouldn't actually work, https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute

The `itemId` is enumerated in handled props and `getAttribute` only works on attributes that are placed on the element. In that case `itemId`, even if it weren't removed by including it in `handledProps` shouldn't get placed on the element since it is not valid HTML.

Can you add tests for this?

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
```diff
@@ -24,25 +25,60 @@ class ListboxItem extends Foundation<
...
      */
+
     public render(): React.ReactElement<HTMLDivElement> {
         return (
             <div
                 {...this.unhandledProps()}
                 className={this.generateClassNames()}
                 role="option"
-                id={this.props.id}
+                id={this.props.id || null}
```
janechu: You shouldn't need to do this, if `this.props.id` is undefined I believe it just wont be added.

packages/fast-components-react-base/src/listbox-item/README.md
scomea: yep

packages/fast-components-react-base/src/listbox-item/README.md
scomea: yep

packages/fast-components-react-base/src/listbox-item/listbox-item.tsx
scomea: "deprecated" is clearly my spelling bee weakness.

PR #1728, opened 2019-05-01
fix: addresses data issues with input button in text action

nalogiudice:
# Description
Fixes issue with the input button not being able to be correctly controlled via the form generator

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
+++ b/packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
@@ -13,10 +13,14 @@ export default class MapChildrenPropToCallbackPassingClassName extends Plugin<
             const callbackArguments: any = {
                 className,
             };
-            callbackArguments.disabled =
-                typeof disabled === "boolean" ? disabled : undefined;
-            callbackArguments.appearance =
-                typeof appearance === "string" ? appearance : undefined;
+
+            if (callbackArguments.disabled) {
+                callbackArguments.disabled = disabled;
+            }
+
+            if (callbackArguments.appearance) {
+                callbackArguments.appearance = appearance;
+            }
 
             return React.createElement(
                 childOption.component,
--- a/packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
+++ b/packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
@@ -14,11 +14,11 @@ export default class MapChildrenPropToCallbackPassingClassName extends Plugin<
                 className,
             };
 
-            if (callbackArguments.disabled) {
+            if (disabled) {
                 callbackArguments.disabled = disabled;
             }
 
-            if (callbackArguments.appearance) {
+            if (appearance) {
                 callbackArguments.appearance = appearance;
             }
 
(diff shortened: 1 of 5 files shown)


packages/fast-components-react-msft/app/utilities/additional-props.plugin.ts
```diff
@@ -13,10 +13,14 @@ export default class MapChildrenPropToCallbackPassingClassName extends Plugin<
             const callbackArguments: any = {
                 className,
             };
-            callbackArguments.disabled =
-                typeof disabled === "boolean" ? disabled : undefined;
-            callbackArguments.appearance =
-                typeof appearance === "string" ? appearance : undefined;
+
+            if (callbackArguments.disabled) {
```
nicholasrice: Is this really what we want? These conditionals will never evaluate to true because `callbackArguments` is created directly above and never assigned either a `disabled` or `appearance` property - meaning they will always be `undefined`

nalogiudice: Nope! Sorry, that was a logic error on my part. I've updated.

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -108,8 +108,6 @@ const styles: ComponentStyles<TextActionClassNameContract, DesignSystem> = (
         },
         textAction_button: {
             borderColor: "transparent",
```
bheston: If we're moving button styling into the button, it seems like the border should move as well. Is it possible to use an outline styled button in a text area now?

packages/fast-components-react-msft/src/text-action/text-action.props.ts
```diff
@@ -24,11 +24,7 @@ export interface TextActionHandledProps
     /**
      * The text action button
      */
-    button?: (
-        classname?: string,
-        disabled?: boolean,
-        appearance?: ButtonAppearance
-    ) => React.ReactNode;
+    button?: (classname?: string) => React.ReactNode;
```
janechu: so do you not want a disabled state set on the root level to be passed here to also disable the button?

nalogiudice: I accidentally removed this - back in now and working as expected.

packages/fast-components-styles-msft/src/text-action/index.ts
nalogiudice: I think we want to open it up and allow any button to be displayed. Turns out the borderColor wasn't even doing anything so I just removed.

packages/fast-components-react-msft/src/text-action/text-action.props.ts
```diff
@@ -56,6 +52,11 @@ export interface TextActionHandledProps
      * Callback for when text action input is blurred
      */
     onBlur?: (e?: React.FocusEvent<HTMLInputElement>) => void;
+
+    /**
```
nicholasrice: Does this need to be handled? I think unhandledProps covers this as it is

nalogiudice: I believe it does because this is an overall disabled prop that controls both the input _and_ button disabled states.

nicholasrice: Can you make that clear in the code comment?

nalogiudice: Sure 👍

PR #1729, opened 2019-05-02
feat: add an ad-hoc css property editor

janechu:
# Description

Preliminary work to add a CSS property editor.

## Motivation & context

This adds a CSS property editor which allows for advanced CSS property editing.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/app/pages/css-property-editor.tsx
+++ b/packages/fast-tooling-react/app/pages/css-property-editor.tsx
@@ -3,7 +3,7 @@ import React from "react";
 import { CSSPropertyEditor } from "../../src/css-property-editor";
 
 interface CSSPropertyEditorTestPageState {
-    css: any;
+    css: { [key: string]: string };
 }
 
 class CSSPropertyEditorTestPage extends React.Component<
--- a/packages/fast-tooling-react/app/pages/css-property-editor.tsx
+++ b/packages/fast-tooling-react/app/pages/css-property-editor.tsx
@@ -1,0 +1,41 @@
+import React from "react";
+// TODO: Update this to "../../src" when the validation task is complete #1725
+import { CSSPropertyEditor } from "../../src/css-property-editor";
+
+interface CSSPropertyEditorTestPageState {
+    css: any;
+}
+
+class CSSPropertyEditorTestPage extends React.Component<
+    {},
+    CSSPropertyEditorTestPageState
+> {
+    constructor(props: {}) {
+        super(props);
+
+        this.state = {
+            css: {
+                margin: "10px",
+                padding: "20px",
+            },
+        };
+    }
+
+    public render(): React.ReactNode {
+        return (
+            <React.Fragment>
+                <CSSPropertyEditor data={this.state.css} onChange={this.handleOnChange} />
+                <pre>{JSON.stringify(this.state.css, null, 2)}</pre>
+                <div style={this.state.css}>Hello world</div>
+            </React.Fragment>
+        );
+    }
+
+    private handleOnChange = (css: any): void => {
+        this.setState({
+            css,
+        });
+    };
+}
+
+export { CSSPropertyEditorTestPage };
--- a/packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
+++ b/packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
@@ -1,0 +1,270 @@
+import React from "react";
+import { get, pick } from "lodash-es";
+import Foundation, {
+    FoundationProps,
+    HandledProps,
+} from "@microsoft/fast-components-foundation-react";
+import {
+    CSSProperties,
+    CSSPropertiesConfig,
+    CSSPropertyConfig,
+    CSSPropertyEditorHandledProps,
+    CSSPropertyEditorProps,
+    CSSPropertyEditorState,
+    CSSPropertyEditorUnhandledProps,
+} from "./property-editor.props";
+import { KeyCodes } from "@microsoft/fast-web-utilities";
+
+export default class CSSPropertyEditor extends Foundation<
+    CSSPropertyEditorHandledProps,
+    CSSPropertyEditorUnhandledProps,
+    CSSPropertyEditorState
+> {
+    public static displayName: string = "CSSPropertyEditor";
+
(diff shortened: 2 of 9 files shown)


packages/fast-tooling-react/app/pages/css-property-editor.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+// TODO: Update this to "../../src" when the validation task is complete #1725
+import { CSSPropertyEditor } from "../../src/css-property-editor";
+
+interface CSSPropertyEditorTestPageState {
+    css: any;
```
nicholasrice: is this really any?

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+                >
+                    {this.renderCurrentProperties()}
+                    {this.renderNewProperty()}
+                </div>
+            </pre>
+        );
+    }
+
+    public componentDidUpdate(prevProps: CSSPropertyEditorProps): void {
+        if (prevProps.data !== this.props.data) {
```
nicholasrice: I think you can use `getDerivedStateFromProps` and avoid multiple renders

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+        key: string,
+        value: string,
+        keyWidth: number,
+        valueWidth: number,
+        index: number
+    ): React.ReactNode {
+        return (
+            <div key={index}>
+                <input
+                    type={"text"}
```
nicholasrice: You could make a factory for this to appease the code-climate gods

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+            this.handleCSSUpdate(newData);
+        };
+    }
+
+    private handlePropertyValueChange(
+        key: string
+    ): (e: React.ChangeEvent<HTMLInputElement>) => void {
+        return (e: React.ChangeEvent<HTMLInputElement>): void => {
+            const newData: CSSProperties = Object.assign({}, this.props.data);
```
nicholasrice: you can simplify: `return {...this.props.data, ...{ [key]: e.target.value } }`

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+        const dataConfig: CSSPropertiesConfig = {};
+
+        if (!data) {
+            return dataConfig;
+        }
+
+        Object.keys(data).map(
+            (dataKey: string): void => {
+                dataConfig[dataKey] = {
+                    value: data[dataKey],
```
nicholasrice: You should make `6.6` a property so it doesn't read as a magic number

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+        return dataConfig;
+    }
+
+    private handleClick = (e: React.MouseEvent<HTMLPreElement>): void => {
+        if ((e.target as HTMLElement).nodeName !== "INPUT") {
+            this.keyInputRef.current.focus();
+        }
+
+        e.stopPropagation();
```
nicholasrice: Why do you need to stop propagation?

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+        return (e: React.ChangeEvent<HTMLInputElement>): void => {
+            const newData: CSSProperties = {};
+
+            Object.keys(this.props.data).forEach((key: string) => {
+                if (key === oldKey) {
+                    newData[e.target.value] = this.props.data[key];
+                } else {
+                    newData[key] = this.props.data[key];
+                }
+            });
```
nicholasrice: Sorry if I'm not reading this right, but I don't think you really need a loop here - for the most part you're just copying the object and conditionally changing a single key. I think you can simplify:

```
return this.handleCssUpdate({
  ...this.props.data,
  ...{[e.target.value]: this.props.data[oldKey]}
})
```

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,270 @@
...
+        });
+
+        return this.renderNewPropertyInput(
+            newPropertyConfig[this.state.key].keyWidth,
+            newPropertyConfig[this.state.key].valueWidth
+        );
+    }
+
+    private renderPropertyInput(
+        key: string,
```
nicholasrice: This function would be more simple if it accepted the config object and not all the data from the config object as individual arguments

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
janechu: I am doing this to preserve the location of the keys in the object, this is important because any update to a key would re-order the list and force the current editable key to jump to the bottom. I'm testing for this but I'll add a comment here to clarify.

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
janechu: I don't, it's a hold over from a previous iteration, I'll remove.

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -0,0 +1,235 @@
...
+    };
+
+    private handlePropertyKeyChange(
+        oldKey: string
+    ): (e: React.ChangeEvent<HTMLInputElement>) => void {
+        return (e: React.ChangeEvent<HTMLInputElement>): void => {
+            const newData: CSSProperties = {};
+
+            // The reason this is iterated over in this manner is to preserve
+            // the location of keys in the CSS object
```
nicholasrice: minor nit - you're preserving the *order* of the keys

PR #1740, opened 2019-05-06
feat: add directionSwitch for better localization composition

### LINKED ISSUE #1738: Add direction switch to JSS utilities
Add a function to help facilitate one of two values depending on a design-system direction. Add to `@microsoft/fast-jss-utilities` package - should conform to the following type signature:

``` 
<T extends { direction: Direction }>(
  ltr: string | (T) => string,
  rtl: string | (T) => string
) => (T) => string
```

nicholasrice:
# Description
Add `directionSwitch` utility to pick a value based on a designSystem.direction value.

closes #1738

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -86,68 +79,59 @@ export const callToActionButtonOverrides: ComponentStyles<
...
+            ...applyGlyphTransform(),
+        }),
+    },
+    callToAction_glyph: {
+        fill: accentForegroundCut,
+        display: "inline-block",
+        position: "relative",
+        width: glyphSize,
+        height: glyphSize,
+        marginLeft: directionSwitch("6px", "unset"),
```
nalogiudice: Just curious about this new pattern, as we are writing additional style code that's not needed.
`applyLocalizedProperty` would only output the single property, while now with `directionSwitch` we output both `marginRight` and `marginLeft`

nicholasrice: Yea I realized we can make these empty strings and no value will be written - I'll update.

In general though, yes. The benefit outweighs the cost

nicholasrice: To elaborate further - this continues on the work to reduce a style-sheets direct dependency on the `DesginSystem` object. There are performance benefits associated with object stylesheets in contrast to function stylesheets (that return stylesheet objects), because we don't need to tear down the `<style>` element when the `designSystem` updates when the stylesheet is an object, like we do when the stylesheet is a function.

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -22,10 +8,20 @@ import {
...
+    ComponentStyles,
+    ComponentStyleSheet,
+    CSSRules,
+} from "@microsoft/fast-jss-manager";
 import { applyCornerRadius } from "../utilities/border";
 import { glyphSize, height, horizontalSpacing } from "../utilities/density";
 import { applyDisabledState } from "../utilities/disabled";
-import { outlineWidth } from "../utilities/design-system";
+import { focusOutlineWidth, outlineWidth } from "../utilities/design-system";
+import {} from "lodash-es";
```
janechu: nit: you can remove this line

packages/fast-components-styles-msft/src/call-to-action/index.ts
nalogiudice: Awesome, thanks for the explanation.

PR #1742, opened 2019-05-07
feat: add a CSS property component to the configuration pane

janechu:
# Description

Adds the ad hoc CSS property editor to the dev site and allows it to be opt-in

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/app/index.tsx
+++ b/packages/fast-components-react-base/app/index.tsx
@@ -64,6 +64,7 @@ function render(): void {
             formPlugins={formPlugins}
             showTransparencyToggle={true}
             styleEditing={true}
+            stylePropertyEditing={true}
         >
             <SiteTitle slot={"title"}>
                 <SiteTitleBrand>FAST</SiteTitleBrand> base component documentation
--- a/packages/fast-components-react-base/app/index.tsx
+++ b/packages/fast-components-react-base/app/index.tsx
@@ -64,7 +64,6 @@ function render(): void {
             formPlugins={formPlugins}
             showTransparencyToggle={true}
             styleEditing={true}
-            stylePropertyEditing={true}
         >
             <SiteTitle slot={"title"}>
                 <SiteTitleBrand>FAST</SiteTitleBrand> base component documentation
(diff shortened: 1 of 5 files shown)


packages/fast-components-react-base/app/index.tsx
```diff
@@ -64,6 +64,7 @@ function render(): void {
             formPlugins={formPlugins}
             showTransparencyToggle={true}
             styleEditing={true}
```
nicholasrice: Whats the difference between `styleEditing` and `stylePropertyEditing`?

janechu: It's the ability to edit styles from the CSS editor, and the ability to edit the individual style properties. I *could* just have the one and you get both, it's very "internal use only" right now.

janechu: You know what, because this is internal use, I'll forgo adding another property since both of these will always be true.

packages/fast-development-site-react/src/components/site/configuration-panel.tsx
```diff
@@ -215,27 +77,50 @@ class ConfigurationPanel extends React.Component<
...
 
-    private getTabClassNames(tabItem: TabType): string {
-        if (tabItem === this.state.activeTab) {
-            return `${this.props.managedClasses.configurationPanel_tab} ${
-                this.props.managedClasses.configurationPanel_tab__active
-            }`;
+    private renderCSSPropertyEditor(): React.ReactNode {
+        if (!!this.props.styleEditing) {
+            return (
+                <div style={{ padding: "0 10px" }}>
```
nalogiudice: Is there a reason you aren't applying styles via classes and doing it inline instead?

janechu: Because it's a single property, not based on the design system. I don't see measurable a benefit of doing it one way over another.

janechu: Having said that, I can convert it if others feel strongly about it.

nalogiudice: Got it. I definitely do not have strong feelings - I was just curious.

janechu: I took action on it anyway.

PR #1753, opened 2019-05-14
feat: [DRAFT] add viewport aware positioner component

scomea:
# Description
Still adding tests and figuring out how to gracefully deal with cases where resize and intersection observer are not available.  But seems in a good enough state for folks to take a look and provide feedback.

Essentially, this component enables an author to create a container that will position itself relative to another "anchor" element in the same viewport based on available space. 

## Motivation & context
Necessary to implement ui elements that take into account the available space in the viewport and the positioning of the anchor element to determine their own position.  Examples might include menus that expand in a different direction based on available space, flyouts, tooltips, context menus, etc...  

## Issue type checklist

## Process & policy checklist

- [x ] I have updated the project documentation to reflect my changes.
- [x ] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://microsoft.github.io/fast-dna/docs/en/contributing/standards) for this project

### DIFF
--- a/packages/fast-components-class-name-contracts-base/src/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/index.ts
@@ -26,3 +26,4 @@ export * from "./text-area";
 export * from "./text-field";
 export * from "./toggle";
 export * from "./typography";
+export * from "./viewport-positioner";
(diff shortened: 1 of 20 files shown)


nicholasrice:
@scomea is this needed anymore? Seems like you have a non draft PR up for this feature.

PR #1757, opened 2019-05-15
fix: remove margin on glyph when there is no label

khamudom:
# Description

Created a function that checks if a glyph and content is null or undefined, and appends a class name to set or remove a margin, if the content is provided or not.

## Motivation & context

This will make the glyph look even inside ActionTrigger button.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
+++ b/packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
@@ -29,6 +29,7 @@ describe("action trigger", (): void => {
         actionTrigger__primary: "action-trigger-primary",
         actionTrigger__stealth: "action-trigger-stealth",
         actionTrigger__disabled: "action-trigger-disabled",
+        actionTrigger__hasGlyphAndContent: "action-trigger-has-glyph-and-content",
     };
     const href: string = "#";
 
@@ -211,5 +212,34 @@ describe("action trigger", (): void => {
         );
 
         expect(rendered.find("button").prop("className")).toContain("custom-class-name");
+    });
+
+    test("should apply 'hasGlyphandContent' class when display includes content and glyph", () => {
+        const props: ActionTriggerHandledProps = {
+            glyph: (className?: string): React.ReactNode => {
+                return <div>X</div>;
+            },
+            children: "text",
+        };
+
+        const rendered: any = mount(<ActionTrigger {...props} />);
+
+        expect(rendered.find("button").prop("className")).toContain(
+            "actionTrigger__hasGlyphAndContent"
+        );
+    });
+
+    test("should not apply 'hasGlyphandContent' class when displaying only glyph", () => {
+        const props: ActionTriggerHandledProps = {
+            glyph: (className?: string): React.ReactNode => {
+                return <div>X</div>;
+            },
+        };
+
+        const rendered: any = mount(<ActionTrigger {...props} />);
+
+        expect(rendered.find("button").prop("className")).not.toContain(
+            "actionTrigger__hasGlyphAndContent"
+        );
     });
 });
--- a/packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
+++ b/packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
@@ -8,6 +8,7 @@ import {
     ActionTriggerUnhandledProps,
 } from "./action-trigger.props";
 import { actionTriggerButtonOverrides } from "@microsoft/fast-components-styles-msft";
+import { isNullOrUndefined } from "util";
 import { DisplayNamePrefix } from "../utilities";
 
 class ActionTrigger extends Foundation<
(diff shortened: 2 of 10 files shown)


packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
```diff
@@ -68,12 +69,30 @@ class ActionTrigger extends Foundation<
             )}`;
         }
 
+        if (this.hasGlyphAndContent()) {
```
chrisdholt: can you add a test for this please?

khamudom: test is added

packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
```diff
@@ -212,4 +213,33 @@ describe("action trigger", (): void => {
 
         expect(rendered.find("button").prop("className")).toContain("custom-class-name");
     });
+
+    test("should apply 'hasGlyphandContent' class when display includes content and glyph", () => {
```
bheston: nit: capital A in "and"

packages/fast-components-react-msft/src/action-trigger/action-trigger.spec.tsx
```diff
@@ -212,4 +213,33 @@ describe("action trigger", (): void => {
...
+        };
+
+        const rendered: any = mount(<ActionTrigger {...props} />);
+
+        expect(rendered.find("button").prop("className")).toContain(
+            "actionTrigger__hasGlyphAndContent"
+        );
+    });
+
+    test("should not apply 'hasGlyphandContent' class when displaying only glyph", () => {
```
bheston: nit: here too

packages/fast-components-react-msft/src/action-trigger/examples.data.tsx
```diff
@@ -81,6 +81,15 @@ const examples: ComponentFactoryExample<ActionTriggerProps> = {
...
             "data-sketch-symbol": "Action trigger - stealth",
         } as any,
+        {
+            appearance: ActionTriggerAppearance.stealth,
+            href: testDestination,
+            glyph: {
+                id: svgSchema.id,
+                props: {},
+            },
+            "data-sketch-symbol": "Action trigger - stealth",
```
bheston: We don't really seem to use/support this anymore, but the Sketch symbol name should be different than above. Maybe something like "Action trigger - stealth - icon-only".

packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
```diff
@@ -8,6 +8,7 @@ import {
     ActionTriggerUnhandledProps,
 } from "./action-trigger.props";
 import { actionTriggerButtonOverrides } from "@microsoft/fast-components-styles-msft";
+import { isNullOrUndefined } from "util";
```
nicholasrice: I don't see this in `package.json` - what is this import?

chrisdholt: +1 it looks like this is getting auto-added from node, which we *do not* want.

nalogiudice:
You're not able to do this by simply keying off of the JSS (using adjacent selectors)?

packages/fast-components-react-msft/src/action-trigger/action-trigger.tsx
khamudom: Yes, this was deprecated in node.
They suggested to use `value === undefined || value === null`
https://nodejs.org/api/util.html#util_util_isnullorundefined_object

I created a function, with the same name, in the utilities directory.
Also updated the reference in action-toggle and auto-suggest-option

packages/fast-components-react-msft/src/utilities/index.ts
```diff
@@ -1,3 +1,7 @@
 const DisplayNamePrefix: string = "MSFT";
 
 export { DisplayNamePrefix };
+
+export function isNullOrUndefined(value: any): boolean {
```
chrisdholt: instead of adding this, can we just us `isNil` from lodash? We already bring it in as a peer dependency.

https://lodash.com/docs#isNil

khamudom: removed `isNullOrUndefined` and used `isNil` from lodash

packages/fast-components-react-msft/src/auto-suggest-option/auto-suggest-option.tsx
```diff
@@ -100,8 +99,8 @@ class AutoSuggestOption extends Foundation<
         searchString: string
     ): React.ReactNode => {
         if (
-            !isNullOrUndefined(displayString) &&
-            !isNullOrUndefined(searchString) &&
+            !isNil(displayString) &&
+            !isNil(searchString) &&
```
chrisdholt: While this does need to change, it is not related to this PR. Your PR title is about a stylistic fix; this code change here is in relation to fixing an import statement to not rely on nodeJs internal utilities. I would prefer not to conflate these two as the same issue. It's easy enough to fix this in a different PR.

khamudom: Then my suggestion is, we abandon this PR. First thing, is to create a new PR that is fixes the import issue. Then once the import fix is committed, tackle the style change. Will that work?

chrisdholt: Alternatively; I think the issue you are solving here is not the import. Is there a reason to not just remove that from this branch and do that in a different PR? It seems like that way you don't have to abandon the PR and instead you can just move forward with both.

khamudom: The PR is now only the margin fixes in action trigger component

Issue #1759, opened 2019-05-15
Add scalable prop to typography


eljefe223:
**What feature are you requesting?**
Add an optional `scalable` prop to typography that adds our density based functions (`...applyScaledTypeRamp()`) via className.

**What would be the benefit of adding this feature?**
This would allow `<Typograhy />` to respond to density changes from the design system if `scalable` is set too `true`.

This also allows us to continue to use `TypographyProps` as props for components and opt into text scaling.

chrisdholt:
I've added a comment on this PR which was raised, but I am leaning towards just applying the scaled font by default: https://github.com/microsoft/fast-dna/pull/1760#pullrequestreview-238121370

It seems like density is something you apply into from a design system standpoint, so you can control where and when it gets applied.

PR #1760, opened 2019-05-15
feat: add scalable prop to typography

eljefe223:
# Description
Add an optional scalable prop to typography that adds our density based functions (...applyScaledTypeRamp()) via className.

Closes: #1759 

## Motivation & context
https://github.com/microsoft/fast-dna/issues/1759

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/typography/typography.tsx
+++ b/packages/fast-components-react-base/src/typography/typography.tsx
@@ -23,6 +23,7 @@ class Typography extends Foundation<
 > {
     public static defaultProps: Partial<TypographyProps> = {
         tag: TypographyTag.p,
+        scalable: false,
     };
 
     public static displayName: string = `${DisplayNamePrefix}Typography`;
@@ -31,6 +32,7 @@ class Typography extends Foundation<
         managedClasses: void 0,
         tag: void 0,
         size: void 0,
+        scalable: void 0,
     };
 
     /**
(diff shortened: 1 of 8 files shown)


janechu:
@chrisdholt let's get some opinion from @bheston

chrisdholt:
@janechu - sure, I think simply from a systematic standpoint though I would assume if we want to opt certain regions out of density we would do that via context.

eljefe223:
> I'm for this, but I do wonder if a prop is needed or if we should just scale the typography component by default.
> 
> Here's the thought I'm working through:
> 
> From a design system standpoint, density is something that you opt into all up. If you want parts of the UI to respond to density, but not others, we have our `<DesignSystemProvider />` component that can reset the context and density for given elements or parts of the UI.
> 
> I'm struggling to know why we don't just wire this component for density with that in mind. I either change the density, or I don't. If I want to exempt a portion of UI, I can reset the multiplier for that context.

@bheston How do you feel about this?

packages/fast-components-react-base/src/typography/typography.tsx
```diff
@@ -23,6 +23,7 @@ class Typography extends Foundation<
 > {
     public static defaultProps: Partial<TypographyProps> = {
         tag: TypographyTag.p,
+        scalable: false,
```
nicholasrice: I prefer the name "scaled" - like you used in the class-name. To me, "scalable" implies I can take some action to scale this - not that the property itself causes scaling.

nicholasrice:
I second @chrisdholt - I'll also add the we're effectively doubling the size of this stylesheet for the sake of a single opt-in prop with this approach.

eljefe223:
Closing pull request in favor of just implementing `...applyScaledTypeRamp()` by default.

New PR can be found here: https://github.com/microsoft/fast-dna/pull/1777

PR #1771, opened 2019-05-18
fix: prevent non-children from showing the style editing in the configuration panel

janechu:
# Description

Prevent the any direct style editing from occuring when the navigation is not on a component or child.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-development-site-react/src/components/site/configuration-panel.tsx
+++ b/packages/fast-development-site-react/src/components/site/configuration-panel.tsx
@@ -174,11 +174,8 @@ class ConfigurationPanel extends React.Component<
     }
 
     private isChildComponent(): boolean {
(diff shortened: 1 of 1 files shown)


packages/fast-development-site-react/src/components/site/configuration-panel.tsx
```diff
@@ -173,6 +173,15 @@ class ConfigurationPanel extends React.Component<
         );
     }
 
+    private isChildComponent(): boolean {
+        const dataLocationSegments: string[] = this.props.dataLocation.split(".");
+
+        return (
+            this.props.dataLocation === "" ||
```
nicholasrice: could also use `this.props.dataLocation.endsWith("props")` and avoid splitting the string.

PR #1775, opened 2019-05-19
fix: select menu should match base select width

scomea:
# Description
Simple change - select menu should be positioned "relative" in order to inherit width.

## Motivation & context.
Engineering reported bug
https://github.com/microsoft/fast-dna/issues/1766

## Issue type checklist

## Process & policy checklist
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://microsoft.github.io/fast-dna/docs/en/contributing/standards) for this project


packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -67,7 +68,7 @@ const styles: ComponentStyles<SelectClassNameContract, DesignSystem> = {
         ...applyFloatingCornerRadius(),
         ...applyElevation(ElevationMultiplier.e11),
         zIndex: "1",
-        position: "absolute",
+        position: "relative",
         width: "100%",
         margin: "0",
         padding: format("{0} 0", toPx<DesignSystem>(designUnit)),
```
bheston: Can we get rid of min/max width now? It seems like we don't need them, and there is no easy way to override them without styling the component.

PR #1781, opened 2019-05-21
feat: add viewport positioner component

scomea:
## Motivation & context
Viewport positioner is a container that positions itself relative to another html element "anchor" while taking into account available space in a parent viewport.  This container enables menus that flip to avoid the edge of the viewport, flyouts, etc... 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
@@ -1,0 +1,49 @@
+/**
+ * The class name contract for the viewport positioner component
+ */
+export interface ViewportPositionerClassNameContract {
+    /**
+     * The root of the viewport positioner component
+     */
+    viewportPositioner?: string;
+
+    /**
+     * left positioner horizontal state
+     */
+    viewportPositioner__left?: string;
+
+    /**
+     * center left positioner horizontal state
+     */
+    viewportPositioner__centerLeft?: string;
+
+    /**
+     * right positioner horizontal state
+     */
+    viewportPositioner__right?: string;
+
+    /**
+     * center right positioner horizontal state
+     */
+    viewportPositioner__centerRight?: string;
+
+    /**
+     * top positioner vertical state
+     */
+    viewportPositioner__top?: string;
+
+    /**
+     * middle top positioner vertical state
+     */
+    viewportPositioner__middleTop?: string;
+
+    /**
+     * bottom positioner vertical state
+     */
+    viewportPositioner__bottom?: string;
+
+    /**
+     * middle bottom positioner vertical state
+     */
+    viewportPositioner__middleBottom?: string;
+}
--- a/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
@@ -28,22 +28,12 @@ export interface ViewportPositionerClassNameContract {
     viewportPositioner__bottom?: string;
 
     /**
-     * horizontal flip outward state
-     */
-    viewportPositioner__horizontalFlipOutward?: string;
-
-    /**
      * horizontal flip inward state
      */
-    viewportPositioner__horizontalFlipInward?: string;
+    viewportPositioner__horizontalInset?: string;
 
     /**
-     * vertical flip outward state
+     * vertical flip inward state
      */
-    viewportPositioner__verticalFlipOutward?: string;
-
-    /**
-     * vertical flip inwarward state
-     */
-    viewportPositioner__verticalFlipInward?: string;
+    viewportPositioner__verticalInset?: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
@@ -33,9 +33,9 @@ export interface ViewportPositionerClassNameContract {
     viewportPositioner__top?: string;
 
     /**
-     * middle top positioner vertical state
+     * center top positioner vertical state
      */
-    viewportPositioner__middleTop?: string;
+    viewportPositioner__centerTop?: string;
 
     /**
      * bottom positioner vertical state
@@ -43,7 +43,7 @@ export interface ViewportPositionerClassNameContract {
     viewportPositioner__bottom?: string;
 
     /**
-     * middle bottom positioner vertical state
+     * center bottom positioner vertical state
      */
-    viewportPositioner__middleBottom?: string;
+    viewportPositioner__centerBottom?: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
@@ -13,19 +13,9 @@ export interface ViewportPositionerClassNameContract {
     viewportPositioner__left?: string;
 
     /**
-     * center left positioner horizontal state
-     */
-    viewportPositioner__centerLeft?: string;
-
-    /**
      * right positioner horizontal state
      */
     viewportPositioner__right?: string;
-
-    /**
-     * center right positioner horizontal state
-     */
-    viewportPositioner__centerRight?: string;
 
     /**
      * top positioner vertical state
@@ -33,17 +23,27 @@ export interface ViewportPositionerClassNameContract {
     viewportPositioner__top?: string;
 
     /**
-     * center top positioner vertical state
-     */
-    viewportPositioner__centerTop?: string;
-
-    /**
      * bottom positioner vertical state
      */
     viewportPositioner__bottom?: string;
 
     /**
-     * center bottom positioner vertical state
+     * horizontal flip outward state
      */
-    viewportPositioner__centerBottom?: string;
+    viewportPositioner__horizontalFlipOutward?: string;
+
+    /**
+     * horizontal flip inward state
+     */
+    viewportPositioner__horizontalFlipInward?: string;
+
+    /**
+     * vertical flip outward state
+     */
+    viewportPositioner__verticalFlipOutward?: string;
+
+    /**
+     * vertical flip inwarward state
+     */
+    viewportPositioner__verticalFlipInward?: string;
 }
--- a/packages/fast-components-react-base/src/viewport-positioner/README.md
+++ b/packages/fast-components-react-base/src/viewport-positioner/README.md
@@ -1,0 +1,53 @@
+## Viewport positioner
+
+The *viewport positioner* component is a container that positions itself relative to an anchor element while trying to take maximum advantage of the space available in the viewport.  For example it could enable a menu to open above or below a control depending on the size and scroll positon of the viewport.
+
+### Usage
+
+The *viewport positioner* and the anchor element it is attached to must both be children of the specified viewport element. There should also be no other scrollable elements in the DOM hierarchy between these elements and the viewport.  The *viewport positioner* element will have its position attribute programmatically set to "relative" and the component also writes in values for the top, right, bottom, left, transform-origin and tranform attributes. 
+
+If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
+
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (ie. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "flipInward" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
+
+FlipOutward:
+
+Anchor
+       |Positioner
+
+ or
+
+             Anchor
+ Positioner|
+
+ FlipInward:
+
+ Anchor
+ |Positioner
+
+ or 
+
+      Anchor
+ Positioner|
+
+ The positioning mode chosen determines the positions that can be chosen.  The "flipOutward" mode allows the "left" and "right" positions only on the horizontal axis, while the "flipInward" mode allows the "centerLeft" and "centerRight" positions. The vertical axis behaves similarly though the positions are named differently.  
+ 
+ If a default position is specified for a particular axis and there is enough space that position will be set.  Whether there is enough space is determined by comparing the threshold size for the axis (or the height/width of the positioner if no threshold specified) to the available space for that position. If there is not enough space, or if no default position is specified, the positioner chooses the position with the most available space.
+
+The chosen position determines which side of the positioner is fixed to the anchor.  For example if the chosen position for the horizontal axis is "left" the positioner will have its "right" property set to a value corresponding to the left edge of the anchor and the "left" propertly will be unset.
+
+The *viewport positioner* adds css classes to itself based on the chosen positions (ie. "viewportPositioner__left", "viewportPositioner__right", etc...) to enable authors to style the positioner and its contents based on relative position.
+
+The "AlwaysInView" props for each axis allows authors to specify whether the positioner should remain on screen when the anchor element is scrolled off.  The *viewport positioner* accomplishes this by setting the translate transform and translate origin properties of the component.
+
+The "fixedAfterInitialPlacement" prop is set the component will not adjust positioning after the initial render.
+
+The not fully supported [ResizeObserver](https://developers.google.com/web/updates/2016/10/resizeobserver) and [IntersectionObserver](https://developers.google.com/web/updates/2016/04/intersectionobserver) are used (supported in Chrome), so it is necessary to apply a polyfill if more thorough browser support is needed (Safari and Firefox). Full adaptation is expected soon.
+
+
+
+
+
+
+
+
--- a/packages/fast-components-react-base/src/viewport-positioner/README.md
+++ b/packages/fast-components-react-base/src/viewport-positioner/README.md
@@ -8,7 +8,7 @@ The *viewport positioner* and the anchor element it is attached to must both be 
 
 If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
 
-Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "flipInward" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "inset" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
 
 FlipOutward:
 
@@ -20,7 +20,7 @@ Anchor
              Anchor
  Positioner|
 
- FlipInward:
+Inset:
 
  Anchor
  |Positioner
@@ -29,8 +29,6 @@ Anchor
 
       Anchor
  Positioner|
-
- The positioning mode chosen determines the positions that can be chosen.  The "flipOutward" mode allows the "left" and "right" positions only on the horizontal axis, while the "flipInward" mode allows the "centerLeft" and "centerRight" positions. The vertical axis behaves similarly though the positions are named differently.  
  
  If a default position is specified for a particular axis and there is enough space that position will be set.  Whether there is enough space is determined by comparing the threshold size for the axis (or the height/width of the positioner if no threshold specified) to the available space for that position. If there is not enough space, or if no default position is specified, the positioner chooses the position with the most available space.
 
--- a/packages/fast-components-react-base/src/viewport-positioner/README.md
+++ b/packages/fast-components-react-base/src/viewport-positioner/README.md
@@ -8,9 +8,9 @@ The *viewport positioner* and the anchor element it is attached to must both be 
 
 If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
 
-Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "inset" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "adjacent" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "inset" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
 
-FlipOutward:
+Adjacent:
 
 Anchor
        |Positioner
--- a/packages/fast-components-react-base/src/viewport-positioner/README.md
+++ b/packages/fast-components-react-base/src/viewport-positioner/README.md
@@ -8,7 +8,7 @@ The *viewport positioner* and the anchor element it is attached to must both be 
 
 If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
 
-Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (ie. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "flipInward" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "flipInward" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
 
 FlipOutward:
 
@@ -36,9 +36,9 @@ Anchor
 
 The chosen position determines which side of the positioner is fixed to the anchor.  For example if the chosen position for the horizontal axis is "left" the positioner will have its "right" property set to a value corresponding to the left edge of the anchor and the "left" propertly will be unset.
 
-The *viewport positioner* adds css classes to itself based on the chosen positions (ie. "viewportPositioner__left", "viewportPositioner__right", etc...) to enable authors to style the positioner and its contents based on relative position.
+The *viewport positioner* adds css classes to itself based on the chosen positions (i.e. "viewportPositioner__left", "viewportPositioner__right", etc.) to enable authors to style the positioner and its contents based on relative position.
 
-The "AlwaysInView" props for each axis allows authors to specify whether the positioner should remain on screen when the anchor element is scrolled off.  The *viewport positioner* accomplishes this by setting the translate transform and translate origin properties of the component.
+The "AlwaysInView" props (i.e. "horizontalAlwaysInView" and "verticalAlwaysInView") for each axis allows authors to specify whether the positioner should remain on screen when the anchor element is scrolled off.  The *viewport positioner* accomplishes this by setting the translate transform and translate origin properties of the component.
 
 The "fixedAfterInitialPlacement" prop is set the component will not adjust positioning after the initial render.
 
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
@@ -1,0 +1,102 @@
+import React from "react";
+import {
+    ManagedClasses,
+    ViewportPositionerClassNameContract,
+} from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface ViewportPositionerManagedClasses
+    extends ManagedClasses<ViewportPositionerClassNameContract> {}
+export interface ViewportPositionerUnhandledProps
+    extends React.HTMLAttributes<HTMLDivElement> {}
+
+export enum HorizontalPosition {
+    left = "left",
+    centerLeft = "centerLeft",
+    centerRight = "centerRight",
+    right = "right",
+    uncontrolled = "uncontrolled",
+}
+
+export enum VerticalPosition {
+    top = "top",
+    middleTop = "middleTop",
+    middleBottom = "middleBottom",
+    bottom = "bottom",
+    uncontrolled = "uncontrolled",
+}
+
+export enum AxisPositioningMode {
+    uncontrolled = "uncontrolled",
+    flipOutward = "flipOutward",
+    flipInward = "flipInward",
+}
+
+export interface ViewportPositionerHandledProps extends ViewportPositionerManagedClasses {
+    /**
+     *  Ref to the element the positioner is anchored to
+     */
+    anchor?: React.RefObject<any>;
+
+    /**
+     *  Ref to the viewport the positioner is constrained to
+     */
+    viewport?: React.RefObject<any>;
+
+    /**
+     *  The positioning mode for the horizontal axis, default is uncontrolled
+     */
+    horizontalPositioningMode?: AxisPositioningMode;
+
+    /**
+     *  The default horizontal position, layout favors the widest option if unset
+     */
+    defaultHorizontalPosition?: HorizontalPosition;
+
+    /**
+     *  The width at which the positioner switches from the default position to the widest one
+     */
+    horizontalThreshold?: number;
+
+    /**
+     *  When enabled the positioner will not move out of the viewport on the horizontal axis
+     */
+    horizontalAlwaysInView?: boolean;
+
+    /**
+     *  The positioning mode for the vertical axis, default is uncontrolled
+     */
+    verticalPositioningMode?: AxisPositioningMode;
+
+    /**
+     * The default vertical position, layout favors the tallest option if unset
+     */
+    defaultVerticalPosition?: VerticalPosition;
+
+    /**
+     *  The height at which the positioner switches from the default position to the tallest one
+     */
+    verticalThreshold?: number;
+
+    /**
+     *  When enabled the positioner will not move out of the viewport on the vertical axis
+     */
+    verticalAlwaysInView?: boolean;
+
+    /**
+     * The disabled state
+     */
+    disabled?: boolean;
+
+    /**
+     * When true the positioner remains fixed relative to it's anchor after the first render
+     */
+    fixedAfterInitialPlacement?: boolean;
+
+    /**
+     * The children of the viewport positioner
+     */
+    children?: React.ReactNode;
+}
+
+export type ViewportPositionerProps = ViewportPositionerHandledProps &
+    ViewportPositionerUnhandledProps;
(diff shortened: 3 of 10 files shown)


packages/fast-components-react-base/src/viewport-positioner/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The *viewport positioner* and the anchor element it is attached to must both be children of the specified viewport element. There should also be no other scrollable elements in the DOM hierarchy between these elements and the viewport.  The *viewport positioner* element will have its position attribute programmatically set to "relative" and the component also writes in values for the top, right, bottom, left, transform-origin and tranform attributes. 
+
+If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
+
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (ie. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "flipInward" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
```
nalogiudice: nit: i.e.

packages/fast-components-react-base/src/viewport-positioner/README.md
```diff
@@ -0,0 +1,53 @@
...
+ 
+ If a default position is specified for a particular axis and there is enough space that position will be set.  Whether there is enough space is determined by comparing the threshold size for the axis (or the height/width of the positioner if no threshold specified) to the available space for that position. If there is not enough space, or if no default position is specified, the positioner chooses the position with the most available space.
+
+The chosen position determines which side of the positioner is fixed to the anchor.  For example if the chosen position for the horizontal axis is "left" the positioner will have its "right" property set to a value corresponding to the left edge of the anchor and the "left" propertly will be unset.
+
+The *viewport positioner* adds css classes to itself based on the chosen positions (ie. "viewportPositioner__left", "viewportPositioner__right", etc...) to enable authors to style the positioner and its contents based on relative position.
```
nalogiudice: nit: `i.e.` and `etc.`

packages/fast-components-react-base/src/viewport-positioner/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The "AlwaysInView" props for each axis allows authors to specify whether the positioner should remain on screen when the anchor element is scrolled off.  The *viewport positioner* accomplishes this by setting the translate transform and translate origin properties of the component.
```
nalogiudice: I would enumerate these props to be more clear (i.e. horizontalAlwaysInView and verticalAlwaysInView

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,879 @@
...
+    private rootElement: React.RefObject<HTMLDivElement> = React.createRef<
+        HTMLDivElement
+    >();
+
+    private openRequestAnimationFrame: number = null;
+
+    /**
+     * Throttle resize request animation frame usage
+     */
+    // private throttledResize: throttle;
```
nalogiudice: commented out code?

packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * top positioner vertical state
+     */
+    viewportPositioner__top?: string;
+
+    /**
+     * middle top positioner vertical state
+     */
+    viewportPositioner__middleTop?: string;
```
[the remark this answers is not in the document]
scomea: Yeah, I did that to mirror this:
div {
  text-align: center;
  vertical-align: middle;
}

Don't feel very strongly about that choice if people prefer another approach.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
```diff
@@ -0,0 +1,102 @@
...
+    middleBottom = "middleBottom",
+    bottom = "bottom",
+    uncontrolled = "uncontrolled",
+}
+
+export enum AxisPositioningMode {
+    uncontrolled = "uncontrolled",
+    flipOutward = "flipOutward",
+    flipInward = "flipInward",
+}
```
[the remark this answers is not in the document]
scomea: ViewportPostitionerVerticalPosition?

packages/fast-components-react-base/src/viewport-positioner/README.md
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/README.md
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/README.md
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
```diff
@@ -0,0 +1,102 @@
...
+     */
+    anchor?: React.RefObject<any>;
+
+    /**
+     *  Ref to the viewport the positioner is constrained to
+     */
+    viewport?: React.RefObject<any>;
+
+    /**
+     *  The positioning mode for the horizontal axis, default is uncontrolled
```
nicholasrice: Does this mean that it won't position itself by default?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,900 @@
...
+        this.props.viewport !== undefined && this.props.viewport.current !== null
+            ? this.props.viewport.current.addEventListener("scroll", this.handleScroll)
+            : document.addEventListener("scroll", this.handleScroll);
+    };
+
+    /**
+     *  Disable the component
+     */
+    private setNoObserverMode = (): void => {
+        // observers not supported so the best we do is try to set the default position if there is one
```
nicholasrice: I'm wondering if we really want two implementations of for subscribing to resize and intersection - I feel like it adds a good amount of code and branching logic to support both. Would we be loosing any capabilities if we omitted the observers and simply implemented the code in no observer mode?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
```diff
@@ -0,0 +1,102 @@
...
+    ViewportPositionerClassNameContract,
+} from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface ViewportPositionerManagedClasses
+    extends ManagedClasses<ViewportPositionerClassNameContract> {}
+export interface ViewportPositionerUnhandledProps
+    extends React.HTMLAttributes<HTMLDivElement> {}
+
+export enum ViewportPositionerHorizontalPosition {
+    left = "left",
```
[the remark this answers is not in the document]
scomea: Nope.  If you want that you set the axis to "uncontrolled" and use regular css styling to manage placement.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
scomea: Yes, that is the current configuration as I had difficulty identifying what default should be.  Thoughts?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: The resize observer has a fairly narrow use case in that it really only matters when the anchor element resizes when the anchor is offscreen and the positioner is sticking onscreen. I'd be open to getting rid of that one altogether and seeing if it becomes an issue.  The intersection observer plays a central role though - without it we'd be dependent on the height/width/scrolltop/scrolleft properties of the elements we have refs to in order to determine positioning which would not account for transforms applied to the anchor (unless we used expensive getClientBounds() calls), also, if we wanted to reposition as a result of user action we'd need to use some timer based scenario.  No observer mode basically just tries to set the default position after checking if it is a valid choice based on the axis positioning mode, doesn't even look at the spacing.

scomea: IMO simpler more compatible version of this component would do a single positioning operation on initial mount pay the perf price for running getClientBounds at that point, and not even try to react to layout changes.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
```diff
@@ -0,0 +1,102 @@
...
+    centerLeft = "centerLeft",
```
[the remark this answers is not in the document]
scomea: In that case we'd need another flag (ie. flipInward vs flipOutward) to differentiate between "left" and "centerLeft" etc... when we calculate positioning values.  Don't think we'd gain anything by making that change other than making our switch statements more convoluted.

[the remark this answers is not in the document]
nicholasrice: Don't we already have that prop though? `horizontalPositioningMode` and `verticalPositioningMode`?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
nicholasrice: uhhh maybe bottom left? I don't know that it matters too much, but I would expect to see it do *something* by default - otherwise it just appears broken on first run without configuration.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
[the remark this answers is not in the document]
scomea: I think it's actually useful to have each position be specifically labeled.  At one point I was considering that there could exist other positioning modes that used other combinations of them.  The labels also represent the calculations for space available, for calculation translation, which positioning prop to set, etc...  These would all still be valid if we decided there were other useful combos of these named positions (I vacillate on that).

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
scomea: How about leaving the horizontal access uncontrolled and the vertical axis to flipOutwards.  That would be a vanilla setting for a vertical menu.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
```diff
@@ -0,0 +1,102 @@
...
+    uncontrolled = "uncontrolled",
+    flipOutward = "flipOutward",
+    flipInward = "flipInward",
+}
+
+export interface ViewportPositionerHandledProps extends ViewportPositionerManagedClasses {
+    /**
+     *  Ref to the element the positioner is anchored to
+     */
+    anchor?: React.RefObject<any>;
```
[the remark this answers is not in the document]
scomea: Makes sense.  Changing to
    /**
     *  Ref to the element the positioner is anchored to
     */
    anchor?: React.RefObject<any> | HTMLElement;

    /**
     *  Ref to the viewport the positioner is constrained to
     */
    viewport?: React.RefObject<any> | HTMLElement;

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,900 @@
...
+    yTransformOrigin: string;
+    xTranslate: number;
+    yTranslate: number;
+    top: number;
+    right: number;
+    bottom: number;
+    left: number;
+    currentVerticalPosition: ViewportPositionerVerticalPosition;
+    currentHorizontalPosition: ViewportPositionerHorizontalPosition;
+
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
scomea: did that.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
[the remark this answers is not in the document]
scomea: refactored to internalize labels and simplify exposed api as we discussed.

packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
```diff
@@ -0,0 +1,49 @@
...
+     */
+    viewportPositioner__horizontalFlipInward?: string;
+
+    /**
+     * vertical flip outward state
+     */
+    viewportPositioner__verticalFlipOutward?: string;
+
+    /**
+     * vertical flip inwarward state
```
janechu: nit: typo

packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * horizontal flip inward state
+     */
+    viewportPositioner__horizontalFlipInward?: string;
+
+    /**
+     * vertical flip outward state
+     */
+    viewportPositioner__verticalFlipOutward?: string;
```
janechu: I'm thinking maybe we should follow the CSS naming convention, where "flip outward" is the default and you only specify flip inward, which could be renamed to "inset"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.json
```diff
@@ -0,0 +1,85 @@
...
+        },
+        "defaultHorizontalPosition": {
+            "title": "Default horizontal position",
+            "type": "string",
+            "default": "uncontrolled",
+            "enum": [
+                "left",
+                "centerLeft",
+                "centerRight",
+                "right"
```
janechu: you should add "uncontrolled" to the enum

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1057 @@
...
+    }
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <div
+                ref={this.rootElement}
+                {...this.unhandledProps()}
```
janechu: This should occur after all other attributes or it could unintentionally override them

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1057 @@
...
+                    " ",
+                    get(
+                        this.props.managedClasses,
+                        "viewportPositioner__horizontalFlipOutward",
+                        ""
+                    )
+                );
+                break;
+
+            case ViewportPositionerHorizontalPositionLabel.centerLeft:
```
janechu: can we call this insetLeft, correct me if I'm wrong but centerLeft doesn't really tell me this is inset to the component

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1057 @@
...
+            if (
+                (this.props.disabled === undefined || this.props.disabled === false) &&
+                isNil(this.positionerRect) &&
+                this.getAnchorElement() !== null
+            ) {
+                shouldHide = true;
+            }
+        } else if (isNil(this.positionerRect) && !this.state.noOberverMode) {
+            shouldHide = true;
+        }
```
janechu: seems like this could be simplified to:

```
const shouldHide: boolean = (this.state.disabled &&
    (
        (this.props.disabled === undefined || this.props.disabled === false) &&
                isNil(this.positionerRect) &&
                this.getAnchorElement() !== null
    )
) ||
(isNil(this.positionerRect) && !this.state.noOberverMode);
```

But as that is hard to read, you might put it in another method.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1057 @@
...
+                left = this.anchorWidth;
+                break;
+        }
+
+        this.setState({
+            xTransformOrigin,
+            right,
+            left,
+            currentHorizontalPosition: desiredHorizontalPosition,
+        });
```
janechu: you are setting state too in too many locations, this is very hard to track. I would recommend using methods to return new props for the state and setting it a minimum number of times.

packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.json
scomea: yep

packages/fast-components-class-name-contracts-base/src/viewport-positioner/index.ts
scomea: Yep, "inset" is much better

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: consolidated some of that.

packages/fast-components-react-base/src/viewport-positioner/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The *viewport positioner* and the anchor element it is attached to must both be children of the specified viewport element. There should also be no other scrollable elements in the DOM hierarchy between these elements and the viewport.  The *viewport positioner* element will have its position attribute programmatically set to "relative" and the component also writes in values for the top, right, bottom, left, transform-origin and tranform attributes. 
+
+If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
+
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "flipInward" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
```
janechu: update flipOutward and flipInward to inset

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export enum ViewportPositionerVerticalPosition {
+    top = "top",
+    bottom = "bottom",
+    uncontrolled = "uncontrolled",
+}
+
+export enum AxisPositioningMode {
+    uncontrolled = "uncontrolled",
+    flipOutward = "flipOutward",
```
janechu: do you think flipOutward should now be "auto" since you are using "inset" below?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+    top: number;
+    right: number;
+    bottom: number;
+    left: number;
+    currentVerticalPosition: ViewportPositionerVerticalPositionLabel;
+    currentHorizontalPosition: ViewportPositionerHorizontalPositionLabel;
+    defaultVerticalPosition: ViewportPositionerVerticalPositionLabel;
+    defaultHorizontalPosition: ViewportPositionerHorizontalPositionLabel;
+    initialLayoutComplete: boolean;
+}
```
janechu: do any of these need comments?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+     * constructor
+     */
+    constructor(props: ViewportPositionerProps) {
+        super(props);
+
+        this.state = {
+            disabled: true,
+            noOberverMode: false,
+            xTransformOrigin: "left",
+            yTransformOrigin: "top",
```
janechu: could you use an enum on these?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                break;
+        }
+
+        return super.generateClassNames(classNames);
+    }
+
+    /**
+     *  gets the CSS classes to be programmatically applied to the component
+     */
+    private getPositioningStyles = (): {} => {
```
janechu: can you type this as a Partial?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+            right: this.state.right === null ? null : `${this.state.right}px`,
+            bottom: this.state.bottom === null ? null : `${this.state.bottom}px`,
+            left: this.state.left === null ? null : `${this.state.left}px`,
+        };
+    };
+
+    /**
+     *  Checks whether component should be disabled or not
+     */
+    private checkComponentConfig = (): void => {
```
janechu: this name is not intuitive, i would prefer something like "updateDisabledState"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                    ViewportPositionerVerticalPositionLabel.top,
+                    ViewportPositionerVerticalPositionLabel.bottom,
+                ];
+        }
+    };
+
+    /**
+     *  Get the width available for a particular horizontal position
+     */
+    private getOptionWidth = (
```
janechu: i would prefer "getHorizontalPositionAvailableWidth"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                return spaceRight + this.anchorWidth;
+            case ViewportPositionerHorizontalPositionLabel.right:
+                return spaceRight;
+        }
+    };
+
+    /**
+     *  Get the height available for a particular vertical position
+     */
+    private getOptionHeight = (
```
janechu: i would prefer getVerticalPositionAvailableHeight

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+    private getBaseHorizontalPositioningState = (
+        desiredHorizontalPosition: ViewportPositionerHorizontalPositionLabel
+    ): Partial<ViewportPositionerState> => {
+        let right: number = null;
+        let left: number = null;
+        let xTransformOrigin: string = "left";
+
+        switch (desiredHorizontalPosition) {
+            case ViewportPositionerHorizontalPositionLabel.left:
+                xTransformOrigin = "right";
```
janechu: could you use an enum for these strings?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                }
+            case AxisPositioningMode.uncontrolled:
+                return ViewportPositionerHorizontalPositionLabel.undefined;
+        }
+    };
+
+    /**
+     * Converts simple vertical position to a position label based on AxisPositioningMode
+     */
+    private ConvertVerticalPositionToLabel = (
```
janechu: the first character of this method should be lowercase, also since this is returning something I would prefer the addition of "get" so "getVerticalPositionToLabel"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+            return this.props.viewport;
+        } else {
+            return this.props.viewport.current;
+        }
+    };
+
+    /**
+     * Converts simple horizontal position to a position label based on AxisPositioningMode
+     */
+    private ConvertHorizontalPositionToLabel = (
```
janechu: the first character should be lowercase, also since this is returning something I would prefer the addition of "get" so "getHorizontalPositionToLabel"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+                translate = translate > 0 ? translate + 1 : 0;
+                break;
+        }
+        return translate;
+    };
+
+    /**
+     *  Calculate vertical tranlation to keep positioner in view
+     */
+    private calculateVerticalTranslate = (
```
janechu: prefer to add "get" here so we know there is a return value, eg. "getVerticalTranslation"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+            case ViewportPositionerHorizontalPositionLabel.right:
+                translate = this.viewportRect.left - this.anchorRight;
+                translate = translate > 0 ? translate + 1 : 0;
+                break;
+        }
+        return translate;
+    };
+
+    /**
+     *  Calculate vertical tranlation to keep positioner in view
```
janechu: nit: "translation"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+        return {
+            yTransformOrigin,
+            top,
+            bottom,
+            currentVerticalPosition: desiredVerticalPosition,
+        };
+    };
+
+    /**
+     *  Calculate horizontal tranlation to keep positioner in view
```
janechu: nit: translation

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+     */
+    private calculateHorizontalTranslate = (
```
janechu: perfer "get" here so "getHorizontalTranslation"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+            right,
+            left,
+            currentHorizontalPosition: desiredHorizontalPosition,
+        };
+    };
+
+    /**
+     * Get base vertical positioning state based on desired position
+     */
+    private getBaseVerticalPositioningState = (
```
janechu: "getDesiredVerticalPositioningState"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+        this.anchorRight = anchorEntry.boundingClientRect.right;
+        this.anchorBottom = anchorEntry.boundingClientRect.bottom;
+        this.anchorHeight = anchorEntry.boundingClientRect.height;
+        this.anchorWidth = anchorEntry.boundingClientRect.width;
+    };
+
+    /**
+     *  Update data based on positioner collisions
+     */
+    private processPositionerCollision = (
```
janechu: "setPositionerCollision"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+
+        this.scrollLeft = this.getViewportElement().scrollLeft;
+
+        this.requestFrame();
+    };
+
+    /**
+     *  Update data based on anchor collisions
+     */
+    private processAnchorCollision = (anchorEntry: IntersectionObserverEntry): void => {
```
janechu: "setAnchorCollision"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -0,0 +1,1011 @@
...
+    /**
+     *  Checks whether component should be disabled or not
+     */
+    private checkComponentConfig = (): void => {
+        if (
+            !canUseDOM() ||
+            this.props.disabled === true ||
+            this.getAnchorElement() === null
+        ) {
+            this.disableComponent();
```
janechu: I think "Component" is implied, you could remove it

packages/fast-components-react-base/src/viewport-positioner/README.md
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.props.ts
scomea: Nah, I think "flipOutward" is more descriptive.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: sure

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: A partial what?  This is a collection of arbitrary css attributes, not props or state so it would not be a part of anything that I have defined.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: sure

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: Actually I think a simple "geVerticalPositioningState" would be fine

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: Hmm...don't like "set" for this, how about "handlePositionerCollision"?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: "handle"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: "React.CSSProperties" worked

packages/fast-components-react-base/src/viewport-positioner/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The *viewport positioner* and the anchor element it is attached to must both be children of the specified viewport element. There should also be no other scrollable elements in the DOM hierarchy between these elements and the viewport.  The *viewport positioner* element will have its position attribute programmatically set to "relative" and the component also writes in values for the top, right, bottom, left, transform-origin and tranform attributes. 
+
+If no "anchor" is specified the component will be disabled and not attempt to adjust layout. If no viewport is specified the component will assume the viewport is the root document element of the page.
+
+Authors can specify a positioning mode for each axis or leave it "uncontrolled" and use other means (i.e. their own css) to manage the positioner's layout on that axis.  The "flipOutward" positioning mode ensures that the positioner does not overlap with the anchor on that axis and butts up against the outer edges of the anchor element.  With the "inset" positioning mode set the positioner will overlap with the anchor on the axis specified but will only extend beyond the anchor on the side with the most available space.
```
janechu: still have the flipOutward leftover

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.json
```diff
@@ -0,0 +1,83 @@
...
+            "type": "boolean"
+        },
+        "verticalPositioningMode": {
+            "title": "Vertical positioning mode",
+            "type": "string",
+            "default": "uncontrolled",
+            "enum": [
+                "uncontrolled",
+                "flipOutward",
+                "Inset"
```
janechu: should be lowercase "i"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.json
```diff
@@ -0,0 +1,83 @@
...
+            "type": "boolean"
+        },
+        "horizontalPositioningMode": {
+            "title": "Horizontal positioning mode",
+            "type": "string",
+            "default": "uncontrolled",
+            "enum": [
+                "uncontrolled",
+                "flipOutward",
+                "Inset"
```
janechu: should be a lowercase "i"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.json
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.json
scomea: yep

PR #1782, opened 2019-05-22
feat: update the text input components to add 'filled' appearance

bheston:
# Description

Added 'appearance' property to support outlined and filled input fields.

## Motivation & context

The update in motivation for design alignment is that controls either have an outline or a plate, but not both.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/design-system/index.ts
+++ b/packages/fast-components-styles-msft/src/design-system/index.ts
@@ -184,10 +184,10 @@ const designSystemDefaults: DesignSystem = {
     neutralFillActiveDelta: 2,
     neutralFillSelectedDelta: 8,
 
-    neutralFillInputRestDelta: 4,
-    neutralFillInputHoverDelta: 4,
-    neutralFillInputActiveDelta: 4,
-    neutralFillInputSelectedDelta: 4,
+    neutralFillInputRestDelta: 0,
+    neutralFillInputHoverDelta: 0,
+    neutralFillInputActiveDelta: 0,
+    neutralFillInputSelectedDelta: 0,
 
     neutralFillStealthRestDelta: 0,
     neutralFillStealthHoverDelta: 3,
--- a/packages/fast-components-styles-msft/src/design-system/index.ts
+++ b/packages/fast-components-styles-msft/src/design-system/index.ts
@@ -184,10 +184,10 @@ const designSystemDefaults: DesignSystem = {
     neutralFillActiveDelta: 2,
     neutralFillSelectedDelta: 8,
 
-    neutralFillInputRestDelta: 0,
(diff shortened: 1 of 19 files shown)


packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -184,10 +184,10 @@ const designSystemDefaults: DesignSystem = {
     neutralFillActiveDelta: 2,
     neutralFillSelectedDelta: 8,
 
-    neutralFillInputRestDelta: 4,
-    neutralFillInputHoverDelta: 4,
-    neutralFillInputActiveDelta: 4,
-    neutralFillInputSelectedDelta: 4,
```
nicholasrice: You want to title the PR as a "fix" and not a "chore" because "chore" signals to lerna that the change shouldn't introduce a version change (meaning this change wouldn't get published).

packages/fast-components-styles-msft/src/patterns/input-field.ts
```diff
@@ -63,7 +61,26 @@ export function inputFieldStyles(
...
+        background: neutralFillRest,
+        border: format("{0} solid transparent", toPx(outlineWidth)),
+        "&:hover:enabled": {
+            background: neutralFillHover,
+            borderColor: "transparent",
+        },
+        "&:active:enabled": {
+            background: neutralFillActive,
+            borderColor: "transparent",
+        },
```
janechu: it saves a small amount of space to do `"&:hover:enabled, &:active:enabled"` for your borderColor property.

packages/fast-components-styles-msft/src/text-action/index.ts
```diff
@@ -70,14 +74,31 @@ const styles: ComponentStyles<TextActionClassNameContract, DesignSystem> = {
...
+        background: neutralFillRest,
+        border: format("{0} solid transparent", toPx<DesignSystem>(outlineWidth)),
+        "&:hover": {
+            background: neutralFillHover,
+            borderColor: "transparent",
+        },
+        "&:active": {
+            background: neutralFillActive,
+            borderColor: "transparent",
+        },
```
janechu: same comment as above, you can consolidate the borderColor with `"&:hover, &:active"`

packages/fast-components-styles-msft/src/patterns/input-field.ts
bheston: Turns out this broke something with rule specificity and hover started taking over focus. Leaving as is.

packages/fast-components-styles-msft/src/text-action/index.ts
bheston: Same as above.

PR #1797, opened 2019-05-27
fix: add experimental color recipe calculations

bheston:
# Description

Updated color recipes such that hover is *always* lighter than rest.

## Motivation & context

We want to update the design guidance for buttons to go lighter on hover and darker on active. The current algorithm did an equal flip and didn't allow for one state to consistently be the same offset lighter or darker than another.

## Issue type checklist

This does not break interface or design. Tested by comparing the new and old algorithm with the current design system deltas.

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/color/accent.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/accent.ts
@@ -16,8 +16,17 @@ export const accentSwatch: SwatchResolver = (designSystem: DesignSystem): Swatch
         : accentPalette[Math.floor(accentPalette.length / 2)];
 };
 
+////////////////////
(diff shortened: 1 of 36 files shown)


packages/fast-components-styles-msft/src/utilities/color/accent.ts
```diff
@@ -78,7 +195,3 @@ export function findAccessibleAccentSwatchIndexes(
         ),
     };
 }
-/**
- * @deprecated - please use findAccessibleAccentSwatchIndexes
- */
-export const findAccessibleAccentSwatchIndexs: typeof findAccessibleAccentSwatchIndexes = findAccessibleAccentSwatchIndexes;
```
bheston: This wasn't exported, so just getting rid of it.

PR #1799, opened 2019-05-27
fix: track base colors in design system

bheston:
# Description

- Added base accent and neutral color to design system to enable subsequent change in palette generation.
- Updated color explorer app to use design system and to indicate base colors.

## Motivation & context

Removes the constant for accent color, which needs to update when the accent color changes or the accent palette is regenerated.

This change is necessary for another change that will adjust the palette generation such that the accent color is not necessarily in the middle.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-color-explorer/app/design-system.ts
+++ b/packages/fast-color-explorer/app/design-system.ts
@@ -7,6 +7,5 @@ import {
 export interface ColorsDesignSystem extends DesignSystem {}
 export const colorsDesignSystem: ColorsDesignSystem = Object.assign(
     {},
-    DesignSystemDefaults,
-    {}
+    DesignSystemDefaults
 );
--- a/packages/fast-color-explorer/app/design-system.ts
+++ b/packages/fast-color-explorer/app/design-system.ts
@@ -2,23 +2,11 @@ import {
     DesignSystem,
     DesignSystemDefaults,
 } from "@microsoft/fast-components-styles-msft";
-import {
-    accentPaletteConfig,
-    createColorPalette,
-    neutralPaletteConfig,
-} from "@microsoft/fast-components-styles-msft";
-import { ColorRGBA64, parseColorHexRGB } from "@microsoft/fast-colors";
-import { AccentColors } from "./colors";
 
 /* tslint:disable-next-line */
 export interface ColorsDesignSystem extends DesignSystem {}
 export const colorsDesignSystem: ColorsDesignSystem = Object.assign(
     {},
     DesignSystemDefaults,
-    {
-        neutralPalette: createColorPalette(new ColorRGBA64(0.5, 0.5, 0.5, 1)),
-        accentPalette: createColorPalette(parseColorHexRGB(
-            AccentColors.blue
-        ) as ColorRGBA64),
-    }
+    {}
 );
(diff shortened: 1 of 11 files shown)


packages/fast-color-explorer/app/design-system.ts
```diff
@@ -2,23 +2,11 @@ import {
...
 export const colorsDesignSystem: ColorsDesignSystem = Object.assign(
     {},
     DesignSystemDefaults,
-    {
-        neutralPalette: createColorPalette(new ColorRGBA64(0.5, 0.5, 0.5, 1)),
-        accentPalette: createColorPalette(parseColorHexRGB(
-            AccentColors.blue
-        ) as ColorRGBA64),
-    }
+    {}
```
nicholasrice: Probably don't need this second empty object.

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -13,6 +13,16 @@ export interface DesignSystem {
      */
     backgroundColor: string;
 
+    /**
+     * The neutral color, which the neutralPalette is based on
```
nicholasrice: Can you add a note to both of these that changes to the palette should also update these values? I want to make it super clear in documentation that changes to base colors and palettes should be synchronized.

packages/fast-components-styles-msft/src/utilities/color/color-constants.ts
```diff
@@ -17,7 +16,10 @@ export const neutralPaletteConfig: ColorPaletteConfig = {
     ...paletteConstants,
 };
 
+/**
+ * @deprecated
+ */
 export const accentPaletteConfig: ColorPaletteConfig = {
```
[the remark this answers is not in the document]
bheston: It's exported.

packages/fast-components-styles-msft/src/utilities/color/color-constants.ts
```diff
@@ -2,7 +2,6 @@ import { ColorPaletteConfig, parseColorHexRGB } from "@microsoft/fast-colors";
 
 export const white: string = "#FFFFFF";
 export const black: string = "#000000";
-export const accent: string = "#0078D4";
```
janechu: is this exported from the root package?

packages/fast-components-styles-msft/src/utilities/design-system.ts
```diff
@@ -9,6 +9,20 @@ export const backgroundColor: DesignSystemResolver<string> = getDesignSystemValu
     "backgroundColor"
 );
 
+/**
+ * Retrieve the base neutral color from the design system
```
janechu: you're putting "." at the end of your other block comments, is there a reason you're doing so some of the time but not here?

packages/fast-components-styles-msft/src/utilities/color/color-constants.ts
bheston: Nope.

packages/fast-components-styles-msft/src/utilities/design-system.ts
bheston: I was cleaning up the other file while I was in it. I'll save this file for another time. ;)

PR #1801, opened 2019-05-29
fix: updated accent getter logic

bheston:
# Description

Flips the accent swatch logic to return the configured base color when provided.

## Motivation & context

This is necessary for planned updated to place accent color off-centered based on luminance value.

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/utilities/color/accent.ts
```diff
@@ -6,14 +6,13 @@ import { contrast, Swatch, SwatchResolver } from "./common";
 import { clamp, inRange } from "lodash-es";
 
 /**
- * Returns a swatch from the middle of the accent palette
+ * Returns the design system accentBaseColor or the swatch from the middle of the accent palette if not configured.
  */
 export const accentSwatch: SwatchResolver = (designSystem: DesignSystem): Swatch => {
+    const accent: Swatch = accentBaseColor(designSystem);
```
nicholasrice: I actually don't think `accent` will ever be `null` because it defaults to the default design system value. We can probably deprecate this function and use `accentBaseColor` in its place - if we want to get fancy we can ensure the `accentBaseColor` exists in the accent palette and if not then return the middle of the accent palette, but I'm not sure that is necessary.

bheston: Good call. This felt a bit more about backwards compatibility, but as you probably noticed in the tests, I had to force it to null just to exercise that branch. Using the base color will help someone realize if they have a configuration issue as well, and we know it won't be the middle color for much longer anyway.
It was not exported, so I just removed it.

packages/fast-components-styles-msft/src/utilities/color/accent.ts
```diff
@@ -78,7 +67,3 @@ export function findAccessibleAccentSwatchIndexes(
         ),
     };
 }
-/**
- * @deprecated - please use findAccessibleAccentSwatchIndexes
- */
-export const findAccessibleAccentSwatchIndexs: typeof findAccessibleAccentSwatchIndexes = findAccessibleAccentSwatchIndexes;
```
bheston: Also not exported, so just removing.

bheston:
Closing to move experimental-related changes into another branch.

PR #1808, opened 2019-06-01
feat: add drag and drop functionality to Navigation

### LINKED ISSUE #1768: Add drag and drop functionality to the Navigation
The Navigation should be able to drag and drop child items to properties with children type.

This should:
- Use react-dnd
- Allow the data type of a child item to be dragged
- Allow the drag target to be an item with children as the data type
- Allow dragging and dropping the order inside the children

janechu:
# Description

Allows children items to be placed in other children data types and reorder.

## Motivation & context

Resolves #1768 

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/navigation/navigation-tree-item.props.ts
```diff
@@ -0,0 +1,123 @@
...
+     * The class for hovering when dragging
+     */
+    getContentDragHoverClassName: (
+        type: NavigationDataType,
+        direction?: VerticalDragDirection
+    ) => string;
+}
+
+export interface NavigationTreeItemDragSourceCollectedProps {
+    connectDragSource?: (el: JSX.Element) => JSX.Element;
```
nalogiudice: Don't we usually try to stay away from names like `el`?

janechu: We want names to be clear, this is for the actual element so I don't think there's any ambiguity

Issue #1715, opened 2019-04-25
High Contrast mode does not work for Icon Buttons


connorjsmith:
**Describe the bug; what happened?**
Buttons which use SVGs/icons seem to not be setting color correctly, making the icon nearly invisible.

**What are the steps to reproduce the issue?**
1. Enable High Contract mode in Windows10
2. https://explore.fastdna.net/components/button/examples/
3. Observe the incorrect icon color

**If applicable, provide screenshots:**
![image](https://user-images.githubusercontent.com/5892363/56759046-db8ae880-674c-11e9-98ce-0c0d0df33ce1.png)
**In what environment did you see the issue?**

* OS & Device: Windows 10 Desktop in high contrast mode
* Browser: Edge (but probably any browser which respects HC settings)
* Version: 1809

chrisdholt:
@connorjsmith This is helpful, thanks. It looks like this is related to instances where the appearance property is not set to "primary".

connorjsmith:
Also looks like none of the buttons handle high contract focus states, making it impossible to know which button has focus

PR #1812, opened 2019-06-03
fix: update button style to support ms-high-contrast

nalogiudice:
Closes [#1715](https://github.com/microsoft/fast-dna/issues/1715)

# Description

Updates button style to support ms-high-contrast.

Screen Shot 2019-06-03 at 12 54 55 PM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -61,12 +61,22 @@ const applyTransparentBackplateStyles: CSSRules<DesignSystem> = {
     // Underline
     "&:hover $button_contentRegion::before": {
         background: accentForegroundHover,
+        "@media (-ms-high-contrast:active)": {
+            background: "ButtonHighlight",
+        },
+    },
+    "&:hover$button__disabled $button_contentRegion::before": {
+        display: "none",
     },
     "&:active $button_contentRegion::before": {
         background: accentForegroundActive,
     },
     "&$button__disabled, &$button__disabled $button_contentRegion::before": {
         ...transparentBackground,
+    },
+    "@media (-ms-high-contrast:active)": {
+        border: "none",
+        fill: "ButtonHighlight",
     },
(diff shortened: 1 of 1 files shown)


packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -235,6 +248,11 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
     },
     button__disabled: {
         ...applyDisabledState(),
+        "& $button_beforeContent, & $button_afterContent": {
+            "@media (-ms-high-contrast:active)": {
+                fill: "white", // ButtonHighlight cannot be passed to -ms-high-contrast but white is safe in this context
```
chrisdholt: Can you elaborate on this? I see `ButtonHighlight` being passed above to a -ms-high-contrast query. My concern with "white" is that there are multiple high contrast modes supported by Windows and white may not meet contrast in all of them. Can we confirm these worked in all of the modes?

nalogiudice: The compound JSS statement was causing the issue. Separating `& $button_beforeContent` and `& $button_afterContent` allowed the usage of `ButtonHighlight`.

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -235,6 +248,12 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
     },
     button__disabled: {
         ...applyDisabledState(),
+        "& $button_beforeContent": {
```
nicholasrice: can you chain these? `& $button_beforeContent, & $button_afterContent`

nalogiudice: See my above comment to @chrisdholt - not sure why but chaining them doesn't allow `ButtonHighlight` to be passed.

PR #1835, opened 2019-06-07
fix: removed an unnecessary type assigned to the navigation items in Navigation

janechu:
# Description

## Motivation & context
This fixes a build break caused by a TypeScript update where unnecessary and incorrect type declarations can take precedence over implicit types.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
+++ b/packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
@@ -218,7 +218,12 @@ const NavigationTreeItem: React.RefForwardingComponent<
 
 export { NavigationTreeItem };
 
-export const DraggableNavigationTreeItem: typeof NavigationTreeItem = DropTarget(
+/*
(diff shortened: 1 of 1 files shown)


packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
```diff
@@ -218,7 +219,7 @@ const NavigationTreeItem: React.RefForwardingComponent<
 
 export { NavigationTreeItem };
 
-export const DraggableNavigationTreeItem: typeof NavigationTreeItem = DropTarget(
+export const DraggableNavigationTreeItem = DropTarget(
     NavigationTreeItemDragID,
```
nicholasrice: linter doesn't spit with no type def?

janechu: nope, it is implicitly declared with a private type called DndComponentClass

PR #1836, opened 2019-06-07
fix: add lifecycle hook to update carousel activeId when props change

### LINKED ISSUE #1828: There's no way to update activeId prop in Carousel
Carousel currently has a handled prop named activeId. This prop is mean to be used to programmatically change the current active slide. Unfortunately the logic that updates the state based on activeId prop seems to be only running in the constructor. This means that if I wanted to implement some outside mechanism to control updating the active slide, it just wouldn't work.

Take a look at where the active id is updated in state:
https://github.com/microsoft/fast-dna/blob/master/packages/fast-components-react-msft/src/carousel/carousel.tsx

Basically it's done only in the constructor of the component - so no matter how many times I try to 're-render' with new props and new activeId, it will not work.

activeId should be taken into account every time I pass in via new props.

* Browser [e.g. Edge, Chrome, Safari, FireFox]
* Version [e.g. 1.8.0]

chrisdholt:
closes #1828 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/carousel/carousel.tsx
+++ b/packages/fast-components-react-msft/src/carousel/carousel.tsx
@@ -31,6 +31,22 @@ class Carousel extends Foundation<
     public static displayName: string = `${DisplayNamePrefix}Carousel`;
 
     /**
+     * React life-cycle method
+     */
+    public static getDerivedStateFromProps(
+        nextProps: CarouselProps,
+        prevState: CarouselState
+    ): null | CarouselState {
+        if (nextProps.activeId && nextProps.activeId !== prevState.activeId) {
+            return {
+                activeId: nextProps.activeId,
+            };
+        }
+
+        return null;
+    }
+
+    /**
      * Handled props
      */
     protected handledProps: HandledProps<CarouselHandledProps> = {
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -30,6 +30,22 @@ class Carousel extends Foundation<
...
+        nextProps: CarouselProps,
+        prevState: CarouselState
+    ): null | CarouselState {
+        if (nextProps.activeId && nextProps.activeId !== prevState.activeId) {
+            return {
+                activeId: nextProps.activeId,
+            };
+        }
+
+        return null;
```
prabashn: is it ok to return empty state?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -30,6 +30,22 @@ class Carousel extends Foundation<
 > {
     public static displayName: string = `${DisplayNamePrefix}Carousel`;
 
+    /**
+     * React life-cycle method
+     */
+    public static getDerivedStateFromProps(
```
prabashn: UTs?

chrisdholt: added

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: This is based on the requirements of the lifecycle hook - either return a state object or null to not update. :)

PR #1839, opened 2019-06-08
fix: remove an issue where the design system would cause component data to be reset

### LINKED ISSUE #1837: Removing default child and adding another does not update children in properties
If you navigate to a component that has default children like (`<Card />`) remove the default child (`Image`) then add a new child the properties panel doesn't reflect the changes you have made.

Example:

1. Go `<Card/>`
2. Remove the default child `image`
3. Add another child that is not an `image`
4. The children list does not reflect the changes you made

janechu:
# Description

## Motivation & context

Solves an issue where the props being updated by an injected design system would cause the internal state of the data to change. Since this is not intended to ever be updated during the viewing of the dev site this has been removed.

Resolves #1837

## Issue type checklist

## Process & policy checklist


eljefe223:
smoked. This also resolves the crash Jason was seeing.

Issue #1632, opened 2019-04-08
Convert schema files to .ts


nicholasrice:
fast-dna maintains numerious schema files to facilitate GUI tooling. Currently, we define these JSON schemeas as `.json` files. 

I propose that we convert these to `.ts` files that export JavaScript objects representing the JSON schema. Doing so solves several problems:

1. No longer need to maintain build configs to enable importing a .json file into a .ts file, including typings and tsconfig settings. Schemas can simply be an ES import like every other piece of data.
2. Code comments - JSON does not allow comments and, while rare, somtimes commenting a schema file is usefull.
3. Code sharing - there are several areas, such as description strings and reference IDs, where we want to keep data in sync across schemas. This is significantly easier to in a programatic language.

**Proposed work:**
1. Convert all `*.schema.json` files to `*.schema.ts` 
a. each schema file should export the schema as default
2. Remove build-system configurations allowing JSON imports, typings, tsconfig, and webpack
3. Replace hard-coded schema id references with the `id` property of the imported schema
4. Update build process so that `tsc` builds the files into the distro - remove `copy-schemas.js`

awentzel:
This sounds great. What is the impact on perf for removing these configurations?  Can you do an assessment of performance building with Lerna before we remove it all to create a baseline and then again after removing.  If we can shave even 30 seconds off build/test/release times this would be outstanding. 

Either way, I approve.

awentzel:
Are there any downsides?

chrisdholt:
Currently we export all of these as `.json` files. What is the implication on distributed packages (IE, breaking change?)?

nicholasrice:
@chrisdholt we should probably treat it as breaking - we could add an interium build step to write out a `.json` version of the file to the distro, as well as a `.js` and keep it additive until the next major - or just do it for the next major.

@awentzel any build time changes will be pretty marginal.

janechu:
I like this, it would definitely make our integration easier. If we want to avoid the breaking issue we might want to add a conversion step when packaging. There are a lot of open source libraries that do this so I don't foresee it being too cumbersome and it might be more intuitive if someone is expecting to use the JSON file format.

Also for the sake of discussion, JSON schemas are supposed to be self documenting, that is what the "title" and "description" fields (among others) are for. See: http://json-schema.org/understanding-json-schema/reference/generic.html#annotations

Having said that, there are always edge cases.

janechu:
Converting this to a task since we are all agreed.

PR #1851, opened 2019-06-11
feat: convert schema files to .ts in base component package

eljefe223:
# Description
part 1 of: #1632 

Convert schemas over from `.json` to `.ts` files and export them accordingly.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/auto-suggest/examples.data.tsx
```diff
@@ -1,7 +1,6 @@
 import React from "react";
 import { ComponentFactoryExample } from "@microsoft/fast-development-site-react";
-import schema from "./auto-suggest.schema.json";
-import listboxItemSchema from "../listbox-item/listbox-item.schema.json";
+import { autoSuggestSchema, listboxItemSchema } from "../";
```
chrisdholt: This path may have issues resolving in certain environments. Can we update the pathing.

chrisdholt:
These files are currently exported as JSON files - should this be a major version change? Anyone importing or referencing the .json file would break from this with what I can see.

packages/fast-components-react-base/src/auto-suggest/examples.data.tsx
eljefe223: to "../index.ts" ?

eljefe223:
> These files are currently exported as JSON files - should this be a major version change? Anyone importing or referencing the .json file would break from this with what I can see.

We don't currently export the schema files.

packages/fast-components-react-base/src/auto-suggest/examples.data.tsx
chrisdholt: yeah, I would path it directly to be safe

janechu:
@eljefe223 I would delete these as a cleanup later, give anyone who might be using these an opportunity to fix to a version where they exist, convert to using the exported javascript and then in another version delete them.

packages/fast-components-react-base/src/auto-suggest/index.ts
```diff
@@ -1,4 +1,6 @@
 import AutoSuggest from "./auto-suggest";
+import autoSuggestSchema from "./auto-suggest.schema";
 
+export { autoSuggestSchema };
```
chrisdholt: I think we should keep naming intuitive here and elsewhere: `AutoSuggestSchema`

build/convert-schemas.js
```diff
@@ -0,0 +1,56 @@
...
+        const commonJS = "module.exports = ".concat(
+            fileContents.substring(fileContents.indexOf("{"))
+        );
+        const tempPath = path.resolve(
+            path.dirname(filePath),
+            `.tmp/${path.basename(filePath)}.js`
+        );
+
+        try {
+            fs.writeFileSync(tempPath, commonJS);
```
nicholasrice: prior to this we should ensure that the tempPath exists:

```
if (!fs.existsSync(path.dirname(tempPath)) {
  fs.mkdir(path.dirname(tempPath))
}
```

packages/fast-components-react-base/src/auto-suggest/index.ts
eljefe223: Chatted offline

PR #1864, opened 2019-06-14
chore: update to ensure custom domain name is persisted in github set…

awentzel:
# Description
Every time a PR goes into master the custom domain settings for GitHub pages is overwritten, then creating a 404 error when visiting those pages.

## Motivation & context
Fix the issue described above in Description.

## Issue type checklist

## Process & policy checklist


website/static/CNAME
```diff
@@ -0,0 +1 @@
+www.fast.design
```
janechu: it appears we have this same file at the root https://github.com/microsoft/fast-dna/blob/master/CNAME do we need this in both places?

awentzel:
I don’t know yet. I’m trying to fix the site.  Broken again. I will issue a cleanup PR when I figure it out.

website/static/CNAME
awentzel: removed

PR #1865, opened 2019-06-14
feat: add new components for each button appearance style

chrisdholt:
# Description
This PR adds a new component for each of the different `Button` appearances. There are now components for each of the following:
- Accent
- Lightweight
- Justified
- Neutral
- Outline
- Stealth

## Motivation & context
Currently, there is only an "uber" button which is a monolithic instance of all button styles. This creates a situation where any user consuming the MSFT button component is required to take a dependency on *all* styles of the button component. By breaking these instances out into individual styled components, we can save on what is delvivered to the browser asa well as what gets included into the bundle. For instance, if I *only* need a `stealth` button style, only that style is included in both the bundle and only those styles are delivered to the browser.

This PR **keeps** the current button implementation to avoid a breaking change; however, a follow-up PR for vNext would be to refactor the "uber" button to return an instance of each of the other button styles. The following PR would update the enums of the "uber" button to align with the above button names, which now map to the color system.

Guidance going forward would be to use the button style that pertains to your situation. From what I can gather from those using FAST currently, there is a limited need for an "uber" button that can be dynamically changed with a simple enum snap. Even in this situation, I think it makes the most sense for developers to manage their own requirements in this way. IF a dev needs to dynamically change the button appearance from a single prop, they could create their own instance of an "uber" button. If they need just two instances dynamically, they could create or manage that. This gives devs more control over their bundle size and their implementation based on needs. 

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/accent-button/index.ts
```diff
@@ -0,0 +1,60 @@
...
+        background: accentFillRest,
+        "&:hover:enabled": {
+            background: accentFillHover,
+        },
+        "&:active:enabled": {
+            background: accentFillActive,
+        },
+        ...applyFocusVisible<DesignSystem>({
+            borderColor: neutralFocus,
+            boxShadow: format("0 0 0 2px inset {0}", accentButtonInnerFocusRect),
```
eljefe223: do we want to use `focusOutlineWidth`?

chrisdholt: this should be a 1:1 move from what previously existed. I'd rather not change styles in this PR if we don't need to.

chrisdholt:
> > "... however, a follow-up PR for vNext would be to refactor the "uber" button to **return an instance of each of the other button styles**..."
> 
> Curious how this would play out for devs... would they be able to continue using MSFT button with all of its appearances and just get the correct new button as the render?

The MSFT button would be deprecated; guidance would be to composite these using a switch how they need. If they need *all* of them, they can create a component to manage that. If they just need primary and stealth, they can create a component that conditionally returns either of those.

> > "From what I can gather from those using FAST currently, there is a limited need for an "uber" button that can be dynamically changed with a simple enum snap..."
> 
> While I agree there is a limited use case for dynamic button switching, I have used this logic in other places to toggle between primary and other appearances depending on state.

Yep - nothing would stop you, we just wouldn't bake it into our package.

> I assume we will include helpful deprecation messaging in the vNext PR?
Correct!

PR #1874, opened 2019-06-19
fix: remove incorrect min-width values from select and button components

chrisdholt:
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -120,7 +120,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
         maxWidth: "374px",
         minWidth: ensureDesignSystemDefaults(
             (designSystem: DesignSystem): string =>
-                designSystem.density <= -2 ? "100px" : "120px"
+                designSystem.density <= -2 ? "28px" : "32px"
         ),
         padding: format("0 {0}", horizontalSpacing(focusOutlineWidth)),
         display: "inline-flex",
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -120,7 +120,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
         maxWidth: "374px",
         minWidth: ensureDesignSystemDefaults(
             (designSystem: DesignSystem): string =>
-                designSystem.density <= -2 ? "100px" : "120px"
+                designSystem.density <= -2 ? "28px" : "32px"
```
nicholasrice: should this be a function of the design unit?

packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -24,24 +24,21 @@ import { inputFieldStyles } from "../patterns/input-field";
 
 const styles: ComponentStyles<SelectClassNameContract, DesignSystem> = {
     select: {
-        minWidth: "276px",
+        display: "inline-block",
```
nicholasrice: What was this before - block? Are we worried about breaking layouts changing this?

packages/fast-components-styles-msft/src/button/index.ts
chrisdholt: It isn't anywhere else, so I'm a bit hesitant to swap the logic out for something more dynamic right now. Considering this pattern exists throughout the code, I'd prefer we make a change like that in a new PR after consulting with @bheston who drove this as part of the density changes.

packages/fast-components-styles-msft/src/select/index.ts
chrisdholt: @nicholasrice I think that's a fair point; I can remove this change if we want to hold off on this update until vNext

PR #1878, opened 2019-06-19
feat: track accentBaseColor in the design system

bheston:
# Description

Adds accentBaseColor to the design system so we know the actual color used to generate the accent palette.

## Motivation & context

This will be needed for the pending update to place the accent color relative to luminance instead of in the center.

## Issue type checklist

## Process & policy checklist


packages/fast-components-styles-msft/src/utilities/color/accent-fill.ts
```diff
@@ -21,7 +22,7 @@ import { getSwatch, isDarkMode, Palette, palette, PaletteType } from "./palette"
 function accentFillAlgorithm(contrast: number): DesignSystemResolver<FillSwatchFamily> {
     return (designSystem: DesignSystem): FillSwatchFamily => {
         const accentPalette: Palette = palette(PaletteType.accent)(designSystem);
-        const accent: Swatch = accentSwatch(designSystem);
+        const accent: Swatch = accentBaseColor(designSystem);
```
chrisdholt: Am I missing where this is a function that accepts a designSystem value? The only reference I see from `../design-system` is:

```
    /**
     * The accent color, which the accentPalette is based on
     */
    accentBaseColor: string;
```

bheston: It's a helper function all the way down in `packages/fast-components-styles-msft/src/utilities/design-system.ts`

chrisdholt: thx

nicholasrice:
I think you'll also want to update the dev site and color explorer sites to set this color when a new accent color is picked

bheston:
Strange, @nicholasrice, I can't reply to your comment, but I think that will be handled in the dev site by the design system editor panel. I think we fixed that last week so the pane either controlled everything or disabled the fields that didn't work correctly. If that's not what you were referring to, can you provide more specific pointers to the UI or code you mean?

PR #1880, opened 2019-06-19
feat: update outline field style

bheston:
# Description

Updated active state fill and fixed focus border.

## Motivation & context

Fill should not change on active state because it causes an undesirable flash.
Fixed an issue with the focus rect caused by the underlying input styling.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/patterns/input-field.ts
+++ b/packages/fast-components-styles-msft/src/patterns/input-field.ts
@@ -76,8 +76,11 @@ export function filledInputFieldStyles(): CSSRules<{}> {
             borderColor: "transparent",
         },
         "&:active:enabled": {
-            background: neutralFillActive,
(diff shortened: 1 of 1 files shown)


packages/fast-components-styles-msft/src/patterns/input-field.ts
```diff
@@ -76,9 +76,12 @@ export function filledInputFieldStyles(): CSSRules<{}> {
             borderColor: "transparent",
         },
         "&:active:enabled": {
-            background: neutralFillActive,
+            background: neutralFillHover, // Don't change on press.
```
nicholasrice: can we just remove this?

bheston: Yep, looks like we can. Updated.

PR #1887, opened 2019-06-20
feat: update lightweight button and hyperlink underline size

bheston:
# Description

Updated lightweight button underline to 1px on hover/active, 2px on focus.
Updated hyperlink underline to 1px on rest/hover/active, 2px on focus.

## Motivation & context

The hover and active underline should match with normal outline widths (1px), only focus should use the focus width (2px).

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -56,6 +56,7 @@ const applyTransparentBackplateStyles: CSSRules<DesignSystem> = {
         boxShadow: "none",
         "& $button_contentRegion::before": {
             background: neutralForegroundRest,
+            height: toPx<DesignSystem>(focusOutlineWidth),
         },
     }),
     // Underline
@@ -234,7 +235,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
         "&::before": {
             content: "''",
             display: "block",
-            height: toPx<DesignSystem>(focusOutlineWidth),
+            height: toPx<DesignSystem>(outlineWidth),
             position: "absolute",
             bottom: "-3px",
             width: "100%",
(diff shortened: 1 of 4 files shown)


chrisdholt:
Nit: Can you update the PR title to accurately reflect the component/appearance name (lightweight)?

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -56,6 +56,7 @@ const applyTransparentBackplateStyles: CSSRules<DesignSystem> = {
         boxShadow: "none",
         "& $button_contentRegion::before": {
             background: neutralForegroundRest,
+            height: toPx<DesignSystem>(focusOutlineWidth),
```
chrisdholt: With this change, what is the difference between a hyperlink and our button component? Doesn't button just address hyperlink scenarios now?

bheston: They were always visually consistent. As I understood, Hyperlink we more for use inline as an actual href and would take font styling from its parent, and Button is more for an actual semantic button that looks like a link. That is, not inline with other content.

chrisdholt: Button supports either an anchor or semantic button - these can look the same and function differently. The delta was that there was no underline on the button by default as there is no accessibility requirement for it as it is not inline with other text (as hypertext is). So, there was a visual delta between the two intentionally; the underline by default was on hypertext only as it is intended to be used inline with text.

chrisdholt: **Button:**
Screen Shot 2019-06-21 at 10 43 07 AM

**Hypertext**
Screen Shot 2019-06-21 at 10 44 50 AM

bheston: Got it. It looks like my comment is misleading. There is still the same underline setup for the two components in default state. The material change is that _when_ there is an underline it is 1px _unless_ it's focus state, at which it's 2px.

chrisdholt: Gotcha - I'll pull down just to make sure I'm seeing things correctly and sign off :thumbsup:

bheston: Title and comments updated.

PR #1889, opened 2019-06-21
feat: update neutral foreground recipes to be based on contrast

bheston:
# Description

Updated neutral foreground recipes to be based on contrast instead of offset.
Deprecated references to "light" or "dark" offsets or recipes.

## Motivation & context

This is to address the design when there are multiple app layers with increasing color depth, and the contrast increases between the layers because the foreground text was fixed white or almost black.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/design-system/design-system.schema.ts
+++ b/packages/fast-components-styles-msft/src/design-system/design-system.schema.ts
@@ -196,16 +196,6 @@ export default {
             type: "number",
             default: 2,
         },
-        neutralForegroundDarkIndex: {
-            title: "Neutral foreground dark index",
-            type: "number",
-            default: 58,
-        },
-        neutralForegroundLightIndex: {
-            title: "Neutral foreground light index",
-            type: "number",
-            default: 0,
-        },
         neutralForegroundHoverDelta: {
             title: "Neutral foreground hover delta",
             type: "number",
--- a/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
@@ -1,12 +1,11 @@
 import { isDarkMode } from "./palette";
-import { neutralForegroundDark, neutralForegroundLight } from "./neutral-foreground";
 import { ColorRecipe, colorRecipeFactory, Swatch, SwatchResolver } from "./common";
 import { DesignSystem } from "../../design-system";
 
+// These literal values are to remove the dependency on neutralForegroundLight and Dark,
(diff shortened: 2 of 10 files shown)


packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
```diff
@@ -1,12 +1,11 @@
...
 import { DesignSystem } from "../../design-system";
 
+// These literal values are to remove the dependency on neutralForegroundLight and Dark,
+// and are very temporary as focus is being updated to contrast-based as well.
 export const neutralFocus: ColorRecipe<Swatch> = colorRecipeFactory(
     (designSystem: DesignSystem): Swatch => {
-        return isDarkMode(designSystem)
-            ? neutralForegroundLight(designSystem)
-            : neutralForegroundDark(designSystem);
+        return isDarkMode(designSystem) ? "#FFFFFF" : "#101010";
```
nicholasrice: I think we still want these to pull from the neutral palette though don't we?

packages/fast-components-styles-msft/src/utilities/color/neutral-foreground.ts
```diff
@@ -1,82 +1,130 @@
...
+                : direction * stateDeltas.rest > direction * stateDeltas.hover;
+
+        const restIndex: number = indexOneIsRestState
+            ? accessibleIndex1
+            : accessibleIndex2;
+        const hoverIndex: number = indexOneIsRestState
+            ? accessibleIndex2
+            : accessibleIndex1;
+
+        const restClamped: number = clamp(restIndex, 0, maxIndex);
```
nicholasrice: This clamping work is redundant and performed by `getSwatch` - you can remove it.

packages/fast-components-styles-msft/src/design-system/design-system.schema.ts
```diff
@@ -196,16 +196,6 @@ export default {
             type: "number",
             default: 2,
         },
-        neutralForegroundDarkIndex: {
```
chrisdholt: Assuming we have to remove this based on the tooling - we can't hold off on removing this when we pull the deprecated values because the tooling is wired in to create the form from this, correct?

bheston: Yeah, it would break the tooling, but I think we’ll be fine without it. It seems safer since it’s not referenced anywhere anymore.

chrisdholt: 👍

PR #1893, opened 2019-06-23
feat:  delete rows with empty data on blur

scomea:
# Description
Refactored the component significantly.
Added row component.
Needed to manage React keys for rows to keep them consistent otherwise focus goes wonky.
Now keep an "in progress" edit state for keys so edits only get "committed" (ie. a change event is triggered) on loss of focus for key inputs

## Motivation & context
Resolves https://github.com/microsoft/fast-dna/issues/1793

Gets us closer to (still missing some interactions that add new rows on tab out and some specific "click to add a row" scenarios):
https://github.com/microsoft/fast-dna/issues/1849

## Issue type checklist

## Process & policy checklist
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/en/contributing/standards) for this project


packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
```diff
@@ -0,0 +1,62 @@
...
+    extends ManagedClasses<CSSPropertyEditorRowClassNameContract> {
+    /**
+     * The key associated with the value being edited
+     */
+    rowIndex: number;
+
+    /**
+     * The key associated with the value being edited
+     */
+    cssKey: string;
```
janechu: if the value is just called `value` can we just call this `key`?

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export interface CSSPropertyEditorRowUnhandledProps
+    extends React.HTMLAttributes<HTMLDivElement> {}
+
+export interface CSSPropertyEditorRowHandledProps
+    extends ManagedClasses<CSSPropertyEditorRowClassNameContract> {
+    /**
+     * The key associated with the value being edited
+     */
+    rowIndex: number;
```
janechu: Since this isn't necessarily going to be an index but should instead be a unique ID (no matter how this is generated), can we just call this `id`?

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Key input has lost focus
+     */
+    onCommitKeyEdit: (rowKey: string, rowIndex: number) => void;
+
+    /**
+     * Called when mouse click outside the inputs
+     */
+    onClickToInsert: (rowKey: string, rowIndex: number) => void;
```
janechu: can you rename to what it is doing, not its intention, I would call this `onClickOutside`

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Called when the key value changes
+     */
+    onKeyChange: (newkey: string, oldKey: string, rowIndex: number) => void;
+
+    /**
+     * Key input has lost focus
+     */
+    onCommitKeyEdit: (rowKey: string, rowIndex: number) => void;
```
janechu: can you update the arguments on these callbacks to `rowKey` -> `key` and `rowIndex` -> `id`

packages/fast-tooling-react/src/css-property-editor/property-editor-row.tsx
```diff
@@ -0,0 +1,230 @@
...
+        }
+    };
+
+    /**
+     * Handle click
+     */
+    private handleClick = (e: React.MouseEvent<HTMLDivElement>): void => {
+        if (
+            typeof this.props.onClickToInsert === "function" &&
+            (e.target as HTMLElement).nodeName !== "INPUT"
```
janechu: what if the click target is any other input on the page?

packages/fast-tooling-react/src/css-property-editor/property-editor.props.ts
```diff
@@ -1,59 +1,27 @@
...
-    [key: string]: CSSPropertyConfig;
-}
-
 export type CSSPropertiesOnChange = (CSS: CSSProperties) => void;
 
+export interface CSSPropertyEditorState {
+    /**
+     *  The uncommitted string for the key of the row currently being edited
+     */
+    currentEditRowUncommittedCSSPropertyKey: string;
```
janechu: Can we use the word "active" for any row that is currently being edited? So this would be `activeRowKey`

packages/fast-tooling-react/src/css-property-editor/property-editor.spec.tsx
```diff
@@ -1,41 +1,48 @@
...
 import { CSSPropertyEditor } from "./";
+import { CSSPropertyEditorClassNameContract } from "./property-editor.props";
 import { KeyCodes } from "@microsoft/fast-web-utilities";
 
 /**
  * Configure Enzyme
  */
 configure({ adapter: new Adapter() });
 
+/* tslint:disable:no-string-literal */
```
janechu: is this still needed?

packages/fast-tooling-react/src/css-property-editor/property-editor.spec.tsx
```diff
@@ -144,4 +137,43 @@ describe("CSSPropertyEditor", () => {
             shallow(<CSSPropertyEditor data={data} />);
         }).not.toThrow();
     });
+
+    test("row with empty key gets delected on input blur", () => {
```
janechu: can you change this text to be "should remove the active row with empty key on input blur"

packages/fast-tooling-react/src/css-property-editor/property-editor.spec.tsx
```diff
@@ -144,4 +137,43 @@ describe("CSSPropertyEditor", () => {
...
+        expect(inputs).toHaveLength(4);
+        inputs.at(0).simulate("focus");
+        inputs.at(0).simulate("change", { target: { value: emptyKey } });
+        inputs.at(0).simulate("blur");
+
+        inputs = rendered.find("input");
+        expect(inputs).toHaveLength(2);
+    });
+
+    test("edited key values get committed on enter key press", () => {
```
janechu: can you change this text to "should fire the onChange callback to add new key values when the enter key is pressed"

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -1,237 +1,365 @@
...
-    }
+        if (this.currentEditRowIndex === index) {
+            if (this.currentEditRowReactKey !== null) {
+                itemKey = this.currentEditRowReactKey;
+            } else if (cssKey === "") {
+                itemKey = this.generateItemKey();
+            } else {
+                this.currentEditRowReactKey = cssKey;
+            }
+        }
```
janechu: I would prefer to see all of this logic be in its own method, which results in the logic below just being:

```
<PropertyEditorRow
    key={this.getItemKey(key, index)}
    ... etc
/>
```

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -1,237 +1,365 @@
...
-        };
-    }
+        // The reason this is iterated over in this manner is to preserve
+        // the order of keys in the CSS object
+        Object.keys(this.editData).forEach((key: string, index: number) => {
+            if (index === rowIndex) {
+                newData[key] = newValue;
+            } else {
+                newData[key] = this.editData[key];
+            }
```
janechu: you could just do:
```
newData[key] = index === rowIndex
    ? newValue
    : this.editData[key];
```

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -1,237 +1,365 @@
...
+    };
+
+    /**
+     * Create a new row at the insertion index
+     */
+    private createRow = (insertionIndex: number): void => {
+        const newData: CSSProperties = {};
+        const keys: string[] = Object.keys(this.editData);
+
+        for (let i: number = 0; i <= keys.length; i = i + 1) {
```
janechu: Instead of `i = i + 1` we prefer `i++`

You should also store the value of length otherwise it will hit length each time which is a slight perf hit. See https://jsperf.com/array-length-in-loop/25

You can do this by:

`for (let i: number = 0, keysLength: number = keys.length; i <= keysLength; i++)`

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -1,237 +1,365 @@
...
+
+    /**
+     * Component got focus without any data rows, so add an empty one
+     */
+    private handleEmptyFocus = (): void => {
+        this.createRow(0);
+    };
+
+    /**
+     * Create a new row at the insertion index
```
janechu: can you add a bit more context here as it took me a while to understand it, something like:

```
/**
 * Create a new row at an insertion index
 *
 * Rows before will retain their index while rows that
 * are after the insertion index will refer to the previous index
 * to account for the newly inserted row
 */
```

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
scomea: Actually the comment is what was incorrect, rowIndex is the actual index of the row in the list.  Corrected to "The index of the row".

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
scomea: We can't.  The "key" prop on the row used as the React Key for the row (see propertty-editor.tsx "renderRow").  Don't think React even passes that value down to the component (but in this case they aren't always the same).

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
scomea: sure

packages/fast-tooling-react/src/css-property-editor/property-editor-row.tsx
scomea: This is attached to the row div so it should only respond to clicks within the row.

packages/fast-tooling-react/src/css-property-editor/property-editor-row.props.ts
scomea: rowIndex is actually an index, not an id, we could use "index" but personally I find "rowIndex" more descriptive.   For consistency we should probably use the same name for rowKey that we use for cssKey (in your previous comment) but that can't be simply key.

packages/fast-tooling-react/src/css-property-editor/property-editor.props.ts
scomea: yes, active instead of current is good

packages/fast-tooling-react/src/css-property-editor/property-editor.spec.tsx
scomea: apparently not

packages/fast-tooling-react/src/css-property-editor/property-editor.spec.tsx
scomea: yep

packages/fast-tooling-react/src/css-property-editor/property-editor.spec.tsx
scomea: yep

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
scomea: yep

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
scomea: yep

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
scomea: yep,  didn't know you could declare multiple things in that first part of the statement

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
scomea: A simple "Retains keys for existing elements" is really all that is going on here.  Added that.

packages/fast-tooling-react/src/css-property-editor/property-editor-row.tsx
```diff
@@ -0,0 +1,242 @@
...
+                this.props.rowIndex
+            );
+        }
+    };
+
+    /**
+     *  Handle focus events
+     */
+    private handleFocus = (): void => {
+        if (typeof this.props.onRowFocus === "function") {
```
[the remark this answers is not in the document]
scomea: sounds good.

PR #1894, opened 2019-06-24
fix: check to see if any child options or plugins have been passed before attempting to map their locations

janechu:
# Description

## Motivation & context

Improves a perf issue that can occur when neither child options or plugins are passed yet both location mapping functions are used against the schema.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/data-utilities/mapping.ts
+++ b/packages/fast-tooling-react/src/data-utilities/mapping.ts
@@ -32,20 +32,26 @@ export function mapDataToComponent(
     plugins: Array<Plugin<PluginProps>> = []
 ): MappedDataLocation[] {
     const mappedData: any = cloneDeep(data);
+    let reactChildrenDataLocations: string[] = [];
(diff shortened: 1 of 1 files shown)


packages/fast-tooling-react/src/data-utilities/mapping.ts
```diff
@@ -32,20 +32,26 @@ export function mapDataToComponent(
     plugins: Array<Plugin<PluginProps>> = []
 ): MappedDataLocation[] {
     const mappedData: any = cloneDeep(data);
+    let reactChildrenDataLocations: string[] = [];
+    let pluginModifiedDataLocations: PluginLocation[] = [];
+
+    if (childOptions.length > 0) {
```
chrisdholt: Do we want to check that `length` exists first? (line 47 as well)

janechu: You can see the value initializes with an empty array, it is never undefined.

PR #1896, opened 2019-06-24
feat: update neutral focus recipe to be based on contrast

bheston:
# Description

Updated neutral focus recipe to be based on contrast.

## Motivation & context

Design feedback was that the focus rect was too contrasty as previously designed and implemented.
This update makes it relative to the background, and also adds an inner ring where necessary like on an accent button.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
@@ -1,5 +1,6 @@
 import designSystemDefaults, { DesignSystem } from "../../design-system";
 import { neutralFocus } from "./neutral-focus";
+import { contrast } from "./common";
 
 describe("neutralFocus", (): void => {
     test("should return a string when invoked with an object", (): void => {
@@ -11,6 +12,8 @@ describe("neutralFocus", (): void => {
     });
 
     test("should operate on default design system if no design system is supplied", (): void => {
-        expect(neutralFocus({} as DesignSystem)).toBe("#101010");
+        expect(contrast(neutralFocus({} as DesignSystem), "#FFF")).toBeGreaterThanOrEqual(
+            3.5
+        );
     });
 });
--- a/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
@@ -1,15 +1,92 @@
-import { getSwatch, isDarkMode, Palette, palette, PaletteType } from "./palette";
-import { ColorRecipe, colorRecipeFactory, Swatch } from "./common";
-import { DesignSystem } from "../../design-system";
+import {
+    findClosestSwatchIndex,
+    isDarkMode,
+    Palette,
+    PaletteType,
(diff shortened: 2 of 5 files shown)


packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
```diff
@@ -1,15 +1,92 @@
...
+    isDarkMode,
+    Palette,
+    PaletteType,
+    swatchByContrast,
+} from "./palette";
+import { Swatch, SwatchResolver } from "./common";
+import { DesignSystem, DesignSystemResolver } from "../../design-system";
+import { accentPalette, backgroundColor, neutralPalette } from "../design-system";
+
+const targetRatio: number = 3.5;
```
janechu: should these be exported consts from web utilities or colors?

packages/fast-components-styles-msft/src/utilities/color/neutral-focus.spec.ts
```diff
@@ -11,6 +12,8 @@ describe("neutralFocus", (): void => {
     });
 
     test("should operate on default design system if no design system is supplied", (): void => {
-        expect(neutralFocus({} as DesignSystem)).toBe("#101010");
+        expect(contrast(neutralFocus({} as DesignSystem), "#FFF")).toBeGreaterThanOrEqual(
+            3.5
```
janechu: why was this test updated, contrast does not appear to have been changed

bheston: This change probably came from the future. With the focus value no longer fixed, even on a white background, this doesn’t work with the palette change that’s coming in soon. I figured this was more resilient of a test anyway.

packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
bheston: I considered that, but we’d probably add them to the design system to make them configurable. I didn’t want to add that to the interface at this point because of the potential to redesign the color recipes to make them easier to work with. If it looks like something that we want configurable it will be easy to update to later.

janechu: These numbers come from WAI ARIA and are not subject to change through a design system as that would immediately make values in-accessible.

packages/fast-components-styles-msft/src/utilities/color/neutral-focus.ts
```diff
@@ -1,15 +1,92 @@
...
+    return contrastRatio > targetRatio;
+}
+
+const neutralFocusAlgorithm: SwatchResolver = swatchByContrast(backgroundColor)(
+    neutralPalette
+)(neutralFocusIndexResolver)(neutralFocusDirectionResolver)(
+    neutralFocusContrastCondition
 );
+
+export function neutralFocus(designSystem: DesignSystem): Swatch;
```
nicholasrice: You should just be able to do:

```
export const neutralFocus: ColorRecipe<Swatch> = colorRecipeFactory(neutralFocusAlgorithm);
```

PR #1899, opened 2019-06-25
feat: color algorithm and recipe updates

bheston:
# Description

Color algorithm update to compress dark end of ramp
Ramp size update to add more room for states
Updated color parsing to remove rounding issue
Updated design system values for new state direction and style
Updated recipes to support different state directions
Updated layer recipes to be dynamic to support different ramps
Updated tests for new values

## Motivation & context

This update is to align with the greater design system aesthetic work

## Issue type checklist

## Process & policy checklist


packages/fast-colors/src/color-converters.ts
```diff
@@ -16,7 +16,12 @@ import { degreesToRadians, radiansToDegrees } from "./math-utilities";
...
 // http://brucelindbloom.com/index.html
 
-// The alpha channel of the input is ignored
+/**
+ * Get the relative luminance of a color.
+ * Adjusts the color to sRGB space, which is necessary for the WCAG contrast spec.
+ * The alpha channel of the input is ignored.
+ * @param rgb The input color
+ */
 export function rgbToLuminance(rgb: ColorRGBA64): number {
```
nicholasrice: Do you think we should deprecate this name and rename to `rgbToRelativeLuminance`?

packages/fast-colors/src/color-converters.ts
```diff
@@ -42,6 +47,19 @@ export function contrastRatio(a: ColorRGBA64, b: ColorRGBA64): number {
     return (luminanceB + 0.05) / (luminanceA + 0.05);
 }
 
+/**
+ * Get the luminance of a color in the linear RGB space.
+ * This is not the same as the relative luminance in the sRGB space for WCAG contrast calculations. Use rgbToLuminance instead.
+ * @param rgb The input color
+ */
+export function rgbToLuminanceLinear(rgb: ColorRGBA64): number {
```
nicholasrice: If we make the above change, this would probably go to `rgbToLinearLuminance` - could do both of them with the modifier after "Luminance" too

packages/fast-colors/src/color-converters.ts
```diff
@@ -42,6 +47,19 @@ export function contrastRatio(a: ColorRGBA64, b: ColorRGBA64): number {
     return (luminanceB + 0.05) / (luminanceA + 0.05);
 }
 
+/**
+ * Get the luminance of a color in the linear RGB space.
+ * This is not the same as the relative luminance in the sRGB space for WCAG contrast calculations. Use rgbToLuminance instead.
+ * @param rgb The input color
+ */
+export function rgbToLuminanceLinear(rgb: ColorRGBA64): number {
+    const r: number = rgb.r;
```
nicholasrice: You don't really need to store these as `const` variables - you're only accessing them once so it would be better to just read them in the calculation expression.

packages/fast-colors/src/color-palette.ts
```diff
@@ -29,7 +29,7 @@ export interface ColorPaletteConfig {
 
 export class ColorPalette {
     public static readonly defaultPaletteConfig: ColorPaletteConfig = {
-        baseColor: new ColorRGBA64(0.5, 0.5, 0.5, 1),
+        baseColor: parseColorHexRGB("#808080"),
```
nicholasrice: Why are you changing this? Seems like extra work to have to parse the string to middle grey.

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+import { parseColorHexRGB } from "./parse-color";
+
+export interface ComponentStateColorPaletteConfig {
+    baseColor?: ColorRGBA64;
+    steps?: number;
+}
+
+export class ComponentStateColorPalette {
+    public static readonly defaultPaletteConfig: ComponentStateColorPaletteConfig = {
+        baseColor: parseColorHexRGB("#808080"),
```
nicholasrice: same question here - can we not just use `.5` middle values and avoid the parsing step?

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+        steps: 94,
+    };
+
+    public palette: ColorRGBA64[] = [];
+    private readonly config: ComponentStateColorPaletteConfig;
+
+    constructor(config: ComponentStateColorPaletteConfig) {
```
nicholasrice: because both of these arguments are optional, I'd recommend making the argument also optional or assigning a default. As is, I have to do:

```new ComponentStateColorPalette({})```
to satisfy typescript, whereas I'd expect to be able to do
```new ComponentStateColorPalette()```
since none of the config data is actually required.

File: packages/fast-colors/src/component-state-color-palette.ts [hunk 7]
```diff
@@ -0,0 +1,165 @@
...
+        this.config = Object.assign(
+            {},
+            ComponentStateColorPalette.defaultPaletteConfig,
+            config
+        );
+        this.palette = [];
```
nicholasrice: `this.palette = []` is unnecessary - you're already defined it as an empty array with `public palette: ColorRGBA64[] = []`

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.regenPalettes();
+    }
+
+    private regenPalettes(): void {
+        let s: number = this.config.steps;
```
nicholasrice: We generally try to avoid ambiguous names - can you change this to "steps"

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+        if (isNaN(this.config.steps) || this.config.steps < 3) {
```
nicholasrice: minor nit, but at this point, you don't really care what `this.config.steps` is - you care about what `s` (`steps`) is because that is the value you use throughout this function. This condition should be validating `s` (`steps`) - not `this.config.steps`.

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+            s = 3;
+        }
+
+        // This palette is tuned to go as dark as differences between the levels can be perceived according to tests
+        // on numerous monitors in different conditions. Stay linear from white until this first cutoff.
+        const darkLum: number = 0.14;
+
+        // In the dark compression, this is the last luminance value before full black.
+        const darkestLum: number = 0.06;
+
```
nicholasrice: Thanks for adding comments about what all these values are for

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+        const r: ColorPalette = new ColorPalette({
+            ...ColorPalette.greyscalePaletteConfig,
+            baseColor: darkLumColor,
+            baseScalePosition: ((1 - darkLum) * 100) / stepsForLuminanceRamp,
+            steps: s,
+        });
+        const referencePalette: ColorRGBA64[] = r.palette;
+
+        // Find the requested base color on the adjusted luminance reference ramp.
+        // There is no _right_ way to desaturate a color, and both methods we've tested have value, so average them out.
```
nicholasrice: should this denaturation step be configurable? For instance, an option on the config?

packages/fast-colors/src/color-palette.ts
bheston: There is something happening regarding rounding that we should probably look into. Color spaces are really strange, but it has to do with the fractions in the luminance calculations or gamma I think. This will actually parse to something like 0.513 (I don’t remember the exact number), but it’s significant that we process all colors the same when it comes to other calculations and placing them relative to another.

packages/fast-colors/src/component-state-color-palette.ts
bheston: Response above.

packages/fast-colors/src/component-state-color-palette.ts
bheston: You mean in terms of which algorithm or method to use? I don’t think so. I think it can be encapsulated in the larger generation algorithm, but it’s possible we’ll tweak this so we end up with ramps that are more consistent in terms of perceived brightness between steps regardless of color. We’ve been discussing options for a v2, but it’s to make the design more robust rather than to achieve a particular color that has preference over another.

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+        );
+        const darkPercent: number = darkRefIndex / (s - 1);
+
+        // TODO: Creating a color from H, S, and a known L value is not the inverse of getting the relative
+        // luminace as above. Need to derive a relative luminance version of the color to better match on the dark end.
+
+        // Find the dark cutoff and darkest variations of the requested base color.
+        const baseColorHSL: ColorHSL = rgbToHSL(this.config.baseColor);
+
+        const darkBaseHSL: ColorHSL = ColorHSL.fromObject({
```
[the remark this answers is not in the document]
bheston: I collapsed some of them. I think this code is complicated enough that I want to keep the commented sections that each do a thing, so I left a few single-use constants in.

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,165 @@
...
+                const B: number = b.position;
+
+                if (A < B) {
+                    return -1;
+                } else if (A > B) {
+                    return 1;
+                } else {
+                    return 0;
+                }
+            }
```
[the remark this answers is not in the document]
bheston: I moved the logic into ColorScale since the implementation there is what makes the assumption on the array sequence.

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,152 @@
...
+                ColorInterpolationSpace.RGB
+            );
+            this.palette[i] = c;
+        }
+    }
+
+    private matchRelativeLuminanceIndex(input: number, reference: ColorRGBA64[]): number {
+        let bestFitValue: number = Number.MAX_VALUE;
+        let bestFitIndex: number = 0;
+        for (let i: number = 0; i < reference.length; i++) {
```
janechu: As a perf improvement, you can instead do:
`let i: number = 0, referenceLength: number = reference.length: i < referenceLength; i++`

This will store the reference length when this loop is initialized and not look it up every loop.

packages/fast-colors/src/component-state-color-palette.ts
```diff
@@ -0,0 +1,152 @@
...
+            const iLum: number = rgbToLinearLuminance(reference[i]);
+            const fitValue: number = Math.abs(iLum - input);
```
janechu: looks like `iLum` is never referred to except this one instance, you can just delete the `iLum` const and do:

`const fitValue: number = Math.abs(rgbToLinearLuminance(reference[i] - input);`

PR #1920, opened 2019-07-05
feat: add neutral divider color recipe.

bheston:
# Description

Added neutral divider color recipe.

## Motivation & context

The outline recipe is too strong for use as a divider line in many places. This can be thought of as an alternative to a drop shadow for separating content areas. Like a drop shadow, it stays dark as long as possible, even in dark mode.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/color/neutral-divider.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/neutral-divider.ts
@@ -1,5 +1,5 @@
-import { DesignSystem, DesignSystemResolver } from "../../design-system";
-import { neutralDividerDelta } from "../design-system";
+import { DesignSystem } from "../../design-system";
+import { neutralDividerRestDelta } from "../design-system";
 import {
     findClosestBackgroundIndex,
     getSwatch,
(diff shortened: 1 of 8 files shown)


chrisdholt:
@bheston can you update the project docs to reflect the addition of this new recipe?

bheston:
@chrisdholt I just looked at the docs and we only talk about the color recipe patterns, not individual recipes.

packages/fast-components-styles-msft/src/utilities/color/neutral-divider.ts
```diff
@@ -0,0 +1,22 @@
...
+    designSystem: DesignSystem
+): string => {
+    const neutralPalette: Palette = palette(PaletteType.neutral)(designSystem);
+    const backgroundIndex: number = findClosestBackgroundIndex(designSystem);
+    const delta: number = neutralDividerDelta(designSystem);
+    const swapThreshold: number = neutralPalette.length - delta;
+    const direction: 1 | -1 = backgroundIndex >= swapThreshold ? -1 : 1;
+
+    const index: number = backgroundIndex + direction * delta;
+    return getSwatch(index, neutralPalette);
```
nicholasrice: I think this should export the product of the product of the `colorRecipeFactory` so the function signature is consistent with the other color recipes.

Also, should this be namespaced to `rest` in case we add states to this?

chrisdholt: +1 on the namespace to `rest`. I can actually think of a scenario where we have a divider that also functions as a drag handle where we may want hover and focus states.

PR #1921, opened 2019-07-05
feat: add autoplay support for carousel as a controlled behavior

chrisdholt:
# Description

Adds autoplay support to carousel as a controlled behavior.
## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/carousel/carousel.props.ts
```diff
@@ -1,25 +1,84 @@
-import React from "react";
+import React, { ReactNodeArray } from "react";
 import {
     CarouselClassNameContract,
     ManagedClasses,
 } from "@microsoft/fast-components-class-name-contracts-msft";
 
+/**
+ * The carousel themes
```
nalogiudice: nit: The carousel _slide_ theme

packages/fast-components-react-msft/src/carousel/README.md
```diff
@@ -4,4 +4,8 @@
 
 ## Style guidance
 
-Class names have been provided and applied to capture the transition of the slides, and allow animation or transformation depending on incoming slide entry position.
\ No newline at end of file
+Class names have been provided and applied to capture the transition of the slides, and allow animation or transformation depending on incoming slide entry position.
+
+## Auto-rotation
+
+Auto-rotation can be enabled by setting the `autoplay` prop to "true" and including an optional `autoplayInterval` (defaults to 6000ms). If the carousel has an auto-rotation enabled, the automatic slide rotation should stop when any element in the carousel receives keyboard focus or hover. It does not resume unless the user activates a control to enable rotation. To support this scenario, the carousel treats `autoplay` as a controlled feature of this component. A callback prop of `onCarouselHoverOrFocus` is provided and executes when the carousel receives focus or mouse hover. It is the responsibility of those implementing the control to ensure that the auto-rotation stops and starts as expected.
```
nalogiudice: > _"It is the responsibility of those implementing the control to ensure that the auto-rotation stops and starts as expected."_

I can understand the desire to have this as the implementor's responsibility, but the carousel experience on the dev site feels broken and not accessible because hover/focus does not stop the carousel.

chrisdholt: You cannot reliably mix controlled and uncontrolled behavior. The most reliable way is to ensure that they manage this. I’m happy to go into more detail, but this is somewhat a limitation of react - additionally, I don’t think we should try and mix controlled and uncontrolled behavior in a given feature.

packages/fast-components-react-msft/src/carousel/carousel.props.ts
```diff
@@ -1,25 +1,84 @@
-import React from "react";
+import React, { ReactNodeArray } from "react";
```
nalogiudice: Doesn't look like you are using `ReactNodeArray` anywhere

chrisdholt: Yeah, I recall an errant tab where this got auto added. Removing.

nalogiudice:
@chrisdholt somehow the slide themes got swapped, so they are all the opposite of what they should be - can you switch those? (e.g. `CarouselSlideTheme.dark` `->` `CarouselSlideTheme.light`)

Also, the `<Flipper />` fill color for the CarouselSlideTheme.dark is incorrect.

Screen Shot 2019-07-05 at 3 31 00 PM

chrisdholt:
> @chrisdholt somehow the slide themes got swapped, so they are all the opposite of what they should be - can you switch those? (e.g. `CarouselSlideTheme.dark` `->` `CarouselSlideTheme.light`)
> 
> Also, the `<Flipper />` fill color for the CarouselSlideTheme.dark is incorrect.
> 
> Screen Shot 2019-07-05 at 3 31 00 PM

@nalogiudice is this bug part of this PR or seperate? I’d prefer to handle separate if it exists outside this PR. Otherwise, yeah - I can look into why that’s happening.

nalogiudice:
@chrisdholt it can be taken care of in another PR. I can go ahead and get that done - it'll be tiny.

packages/fast-components-react-msft/src/carousel/carousel.props.ts
```diff
@@ -4,22 +4,81 @@ import {
...
+     */
     activeId?: string;
+
+    /**
+     * Autoplay the carousel
+     */
+    autoplay?: boolean;
+
+    /**
+     * The autoplay inteval in milliseconds
```
janechu: nit: interval

packages/fast-components-react-msft/src/carousel/carousel.props.ts
```diff
@@ -4,22 +4,81 @@ import {
...
+    ) => void;
+
+    /**
+     * Callback for when the carousel hover terminates
+     */
+    onCarouselMouseLeave?: (
+        e: React.MouseEvent<HTMLDivElement>,
+        props: CarouselProps
+    ) => void;
+    /**
```
janechu: nit: extra line above

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -30,32 +22,55 @@ class Carousel extends Foundation<
...
+        onCarouselMouseLeave: void 0,
     };
 
+    /**
+     * Store a reference the autoplay interval
+     */
+    private autoplayInterval: number | null;
+
+    /**
+     * Store a reference the autoplay timer
```
janechu: nit: Store a reference *to* the autoplay timer

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -30,32 +22,55 @@ class Carousel extends Foundation<
...
         activeId: void 0,
         items: void 0,
+        onCarouselMouseEnterOrFocus: void 0,
+        onCarouselMouseLeave: void 0,
     };
 
+    /**
+     * Store a reference the autoplay interval
+     */
+    private autoplayInterval: number | null;
```
janechu: Why are we storing this as a private variable? Could we not just set the 6000 you have as default props?

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -82,20 +100,66 @@ class Carousel extends Foundation<
      */
     public render(): React.ReactElement<HTMLDivElement> {
         return (
-            <div {...this.unhandledProps()} className={this.generateClassNames()}>
+            <div
+                {...this.unhandledProps()}
```
janechu: I believe this should be below all of the declared props otherwise you risk overriding ref, onFocus etc.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -82,20 +100,66 @@ class Carousel extends Foundation<
...
+    }
+
+    /**
+     * React lifecycle hook
+     */
+    // public componentDidUnmount(): void {
+    //     if (!isNil(this.autoplayTimer)) {
+    //         this.autoplayTimer = window.clearInterval(this.autoplayTimer as number);
+    //     }
+    // }
```
janechu: still needed?

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: In terms of unhandled? Unhandled needs to be first so we ensure we don’t override those. If I pass this after the unhandled props will blow the ref and onFocus away.

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: thought I pushed this. I think it needs to be there, but shouldn’t be uncommented.

packages/fast-components-react-msft/src/carousel/carousel.tsx
janechu: I had that backwards in terms of overriding then

packages/fast-components-react-msft/src/carousel/carousel.tsx
chrisdholt: undone and passed as props. Previously I was working to try and handle stopping on focus/mouseover and decided to make the thing entirely controlled, so this can simply be prop driven.

PR #1924, opened 2019-07-08
feat: adds the component explorer

janechu:
# Description

## Motivation & context

This is part of the migration work to use storybook for the development environment and tooling to create a documentation site for the Microsoft components. This part of the work adds a starter app that 

## Issue type checklist

## Process & policy checklist


packages/fast-component-explorer/README.md
```diff
@@ -0,0 +1,25 @@
+# Introduction
+
+This is the documentation site for `@microsoft/fast-components-react-msft`.
+
+It is built using the `@microsoft/fast-tooling-react` and `@microsoft/fast-components-react-msft` packages and shows the components in a playground as well as the documentation for each component.
```
eljefe223: Sounds weird to me... maybe:

It is built using the `@microsoft/fast-tooling-react` and `@microsoft/fast-components-react-msft` packages to view, configure and document each component.

packages/fast-component-explorer/app/design-system.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface ExplorerDesignSystem {
+    backgroundColor: string;
+    foregroundColor: string;
+    accentColor: string;
+    direction: string;
+}
+
+export const creatorDesignSystem: DesignSystem<ExplorerDesignSystem> = {
+    backgroundColor: "#1A1A1A",
+    foregroundColor: "#F2F2F2",
```
eljefe223: foreground is derived from DesignSystem

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -0,0 +1,388 @@
...
+        );
+    }
+
+    private renderPivotItems(): TabsItem[] {
+        return [
+            {
+                tab: (className: string): React.ReactNode => "Code Preview",
+                content: (className: string): React.ReactNode => {
+                    return (
+                        <div className={className}>
```
eljefe223: Looks like you can't fully scroll the contents of the syntax highlighter when the row height is smaller:
image

just add `height: calc(100% - 32px)` to you `pivot_tabPanels` class

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -0,0 +1,386 @@
...
+                                            jssStyleSheet={{
+                                                pivot: {
+                                                    height: "100%",
+                                                },
+                                                pivot_tabPanels: {
+                                                    overflow: "auto",
+                                                    padding: "0 12px",
+                                                    height: "calc(100% - 32px)",
+                                                },
+                                            }}
```
nicholasrice: you shouldn't create objects inline when doing `jssStyleSheet` overrides because the JSSManager will see a new stylesheet every render. You should store this as a variable or as a member on the class

packages/fast-component-explorer/app/msft-component-helpers/design-system.schema.json
```diff
@@ -0,0 +1,258 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "title": "Design System",
+    "description": "A design system's schema definition.",
+    "id": "design-system",
+    "type": "object",
+    "properties": {
+        "designSystem": {
```
nicholasrice: Sorry if I missed it - but should we have a test that validates the default design system from `fast-components-styles-msft` against this schema, so that it will fail if changes are made to the design system but not to this schema?

packages/fast-component-explorer/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+    "compilerOptions": {
+        "moduleResolution": "node",
+        "module": "ES6",
+        "target": "ES6",
+        "baseUrl": ".",
+        "jsx": "react",
+        "strictNullChecks": true,
```
nicholasrice: yay!

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -0,0 +1,386 @@
...
+        };
+    }
+
+    private handleUpdateDirection = (): void => {
+        this.setState({
+            viewConfig: {
+                direction:
+                    this.state.viewConfig.direction === Direction.ltr
+                        ? Direction.rtl
+                        : Direction.ltr,
```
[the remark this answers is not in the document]
janechu: It does since this is the only property in `viewConfig`, I'm adding a comment to make sure if this interface is expanded, that is understood.

packages/fast-component-explorer/app/explorer.tsx
janechu: Thanks for the catch! Looks like I missed this one.

packages/fast-component-explorer/app/msft-component-helpers/design-system.schema.json
janechu: We do not yet have that, currently no testing is present for any component interface against a schema, we should do this all-up as a single set of tasks

PR #1929, opened 2019-07-10
feat: add callback for when carousel active id changes

### LINKED ISSUE #1825: Add slide navigation callback to/in Carousel
Need a callback prop on the Carousel for the parent to get notified when the user navigates back/forth between slides using click or keyboard. I need to be able to basically only need this for when the user initiates it (for example not when code triggers an active slide id change). So if you're exposing a generic callback for when next/prev is done, I need an argument to identify this is for an actual user initiated flip. I don't care about the programmatic change scenario, but if you decide to put it under the same callback, I need a way to distinguish between the two.

MSN News Infopane experience's auto rotation feature. Currently I'm using a generic 'onclick' handler to stop the auto rotation, to indicate that the user has interacted with the carousel. But I will be missing things like the keyboard navigation, etc. 

So from the infopane perspective I need a way to start/resume from where the user left off, but if the user manually interacts with the carousel in the meantime, then I don't have a way to keep track of the *current slide id* to know which slide to flip to when the next auto-rotation timer fires.

A callback on the CarouselHandledProps to indicate that the user has navigated prev/next via mouse or keyboard (or any other mechanism that I'm unaware of).

Resorting to using a blind 'onclick' handler to trigger the auto-rotation stop behavior.

**Attach screenshots or images add detail to your request:**

chrisdholt:
closes #1825 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/carousel/carousel.spec.tsx
+++ b/packages/fast-components-react-msft/src/carousel/carousel.spec.tsx
@@ -236,6 +236,62 @@ describe("carousel", (): void => {
         expect(rendered.state("activeId")).toBe("id02");
     });
 
+    test("should fire a callback when `onActiveIdUpdate` prop is passed", () => {
+        const onActiveIdUpdateMock: any = jest.fn();
+        const rendered: any = mount(
+            <MSFTCarousel {...handledProps} onActiveIdUpdate={onActiveIdUpdateMock} />
+        );
+
+        rendered
+            .find('[direction="next"]')
+            .first()
+            .simulate("click");
+
+        expect(onActiveIdUpdateMock).toHaveBeenCalledTimes(1);
+
+        rendered
+            .find('[direction="previous"]')
+            .first()
+            .simulate("click");
+
+        expect(onActiveIdUpdateMock).toHaveBeenCalledTimes(2);
+
+        rendered
+            .find('[role="tab"]')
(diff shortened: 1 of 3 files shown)


prabashn:
Does this callback get triggered whenever the carousel changes regardless of the source of the change (i.e., user clicks vs. programmatic changes?)

packages/fast-components-react-msft/src/carousel/carousel.spec.tsx
```diff
@@ -236,6 +236,62 @@ describe("carousel", (): void => {
...
+
+        rendered
+            .find('[direction="previous"]')
+            .first()
+            .simulate("click");
+
+        expect(onActiveIdUpdateFn.mock.results[1].value).toBe("id01");
+
+        rendered
+            .find('[role="tab"]')
```
prabashn: Is this the little indicators? If not, consider adding a test for those as well.

chrisdholt: Yeah - this is the indicator. Checking click for each of them.

chrisdholt:
> Does this callback get triggered whenever the carousel changes regardless of the source of the change (i.e., user clicks vs. programmatic changes?)

Yes, I believe this should trigger any time the activeID is changed regardless of input.

PR #1940, opened 2019-07-11
feat: add flyout as a new component

chrisdholt:
# Description

Adds a new styled component “flyout” to the MSFT package.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/flyout/README.md
+++ b/packages/fast-components-react-msft/src/flyout/README.md
@@ -1,0 +1,6 @@
+# Flyout
+The *flyout* component accepts children and follows the functional pattern of a non-modal dialog (non-modal dialogs allow users the ability to continue interacting with the page). The width and height of the flyout can be set by props. If no props are passed to set width or height, the flyout width will default to 280px and the height will default to 128px.
+
+## Accessibility
+
+While not required to be focused, a flyout should include a label. This can be applied as an [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute) using the `label` prop, or by referencing a child within the dialog with the [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) attribute using the `labelledby` prop. Optionally, additional context can be provided to assistive technologies by including an [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) attribute which can be applied using the `describedby` prop.
--- a/packages/fast-components-react-msft/src/flyout/flyout.schema.ts
+++ b/packages/fast-components-react-msft/src/flyout/flyout.schema.ts
@@ -1,0 +1,74 @@
+export default {
+    $schema: "http://json-schema.org/schema#",
+    title: "Flyout",
+    description: "A flyout component's schema definition.",
+    type: "object",
+    id: "@microsoft/fast-components-react-msft/flyout",
+    properties: {
+        visible: {
+            title: "Visible",
+            type: "boolean",
+        },
+        horizontalPositioningMode: {
+            title: "Horizontal positioning mode",
+            type: "string",
+            default: "uncontrolled",
+            enum: ["uncontrolled", "adjacent", "inset"],
+        },
+        defaultHorizontalPosition: {
+            title: "Default horizontal position",
+            type: "string",
+            default: "uncontrolled",
+            enum: ["left", "right", "uncontrolled"],
+        },
+        height: {
+            title: "Height",
+            type: "string",
+            default: "128px",
+        },
+        horizontalThreshold: {
+            title: "Horizontal threshold",
+            type: "number",
+        },
+        horizontalAlwaysInView: {
+            title: "Horizontal always in view",
+            type: "boolean",
+        },
+        verticalPositioningMode: {
+            title: "Vertical positioning mode",
+            type: "string",
+            default: "uncontrolled",
+            enum: ["uncontrolled", "adjacent", "inset"],
+        },
+        defaultVerticalPosition: {
+            title: "Default vertical position",
+            type: "string",
+            default: "uncontrolled",
+            enum: ["top", "bottom", "uncontrolled"],
+        },
+        verticalThreshold: {
+            title: "Vertical threshold",
+            type: "number",
+        },
+        verticalAlwaysInView: {
+            title: "Vertical always in view",
+            type: "boolean",
+        },
+        fixedAfterInitialPlacement: {
+            title: "Fixed after initial placement",
+            type: "boolean",
+        },
+        width: {
+            title: "Width",
+            type: "string",
+            default: "280px",
+        },
+    },
+    reactProperties: {
+        children: {
+            title: "Children",
+            type: "children",
+            defaults: ["text"],
+        },
+    },
+};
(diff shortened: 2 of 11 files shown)


packages/fast-components-react-msft/src/flyout/flyout.tsx
```diff
@@ -0,0 +1,201 @@
...
+        verticalThreshold: void 0,
+        viewport: void 0,
+        width: void 0,
+    };
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        // const managedClassMap: ViewportPositionerClassNameContract = this.generateManagedClassNames();
```
nicholasrice: Probably don't need this.

packages/fast-components-react-msft/src/flyout/flyout.tsx
```diff
@@ -0,0 +1,201 @@
...
+                this.props.managedClasses,
+                "flyout__verticalInset",
+                ""
+            ),
+        };
+    }
+
+    private handleWindowClick = (event: MouseEvent): void => {
+        if (typeof this.props.onDismiss === "function" && !this.props.disabled) {
+            this.props.onDismiss(event);
```
nicholasrice: Whats the story if the user clicks *inside* the flyout? Do we still want to dismiss the flyout?

packages/fast-components-react-msft/src/flyout/flyout.tsx
```diff
@@ -0,0 +1,201 @@
...
+        verticalPositioningMode: FlyoutAxisPositioningMode.adjacent,
+        defaultVerticalPosition: FlyoutVerticalPosition.bottom,
+        horizontalAlwaysInView: false,
+        verticalAlwaysInView: false,
+        fixedAfterInitialPlacement: false,
+        height: "128px",
+        width: "280px",
+    };
+
+    protected handledProps: HandledProps<FlyoutHandledProps> = {
```
nicholasrice: if you change this to `HandledProps<Required<FlyoutHandledProps>>` we won't run into the case where this doesn't cause a build error if optional props aren't in `handledProps`

packages/fast-components-react-msft/src/flyout/flyout.stories.tsx
```diff
@@ -0,0 +1,183 @@
...
+        if (!this.state.visible) {
+            return;
+        }
+
+        const { anchor, children, disabled, ...props }: FlyoutProps = this.props;
+
+        return (
+            <Flyout
+                anchor={this.anchor.current}
+                disabled={!this.state.visible}
```
nicholasrice: "disabled" isn't an intuitive name to me to convey whether the flyout is visible or not - should we deviate from the viewport positioner API here?

packages/fast-components-react-msft/src/flyout/flyout.tsx
chrisdholt: touché - good catch.

packages/fast-components-react-msft/src/flyout/flyout.stories.tsx
chrisdholt: I think we should - visible makes more sense to me. I’ll update the contract.

packages/fast-components-react-msft/src/flyout/README.md
```diff
@@ -0,0 +1,6 @@
+# Flyout
+The *flyout* component accepts children and follows the functional pattern of a non-modal dialog (non-modal dialogs allow users the ability to continue interacting with the page). The width and height of the flyout can be set by props. If no props are passed to set width or height, the dialog width will default to 280px and the height will default to 128px.
```
nalogiudice: nit: "...props are passed to set width or height, the **dialog** width will..." **dialog** should be **flyout** here to avoid confusion I think

packages/fast-components-react-msft/src/flyout/flyout.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Renders the component
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        return (
+            <ViewportPositioner
+                {...this.unhandledProps()}
+                role="dialog"
+                aria-label={this.props.label}
```
nalogiudice: These props (aria-label, aria-labelledby, aria-describedby) are optional, but you aren't checking to see if they exist before using them?

packages/fast-components-react-msft/src/flyout/flyout.tsx
nalogiudice: ^ just to be clear no, correct? The user should be able to interact with content inside the flyout - that's the point of a dialog-type flyout

chrisdholt: correct.

packages/fast-components-react-msft/src/flyout/flyout.schema.ts
```diff
@@ -0,0 +1,74 @@
+export default {
+    $schema: "http://json-schema.org/schema#",
+    title: "Flyout",
+    description: "A flyout component's schema definition.",
+    type: "object",
+    id: "@microsoft/fast-components-react-msft/flyout",
+    properties: {
```
eljefe223: does this need a `formPluginId`?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -327,6 +327,9 @@ class ViewportPositioner extends Foundation<
      *  gets the CSS classes to be programmatically applied to the component
      */
     private getPositioningStyles = (): React.CSSProperties => {
+        // Check if there is already a style object being passed as props
+        const styleProps: React.CSSProperties = get(this.props, "style");
```
janechu: It seems to me that since this change is happening in `@microsoft/fast-components-react-base` it will add to the changelog a minor bump and the item "add flyout as a new component". This doesn't seem desirable since this should be a change to msft packages only. Can we split the changes here into another PR?

packages/fast-components-react-msft/src/flyout/flyout.schema.ts
janechu: I've been adding them to schemas for the tooling, so I would say both a `formPluginId` and a `pluginId` would be great

packages/fast-components-react-msft/src/flyout/flyout.tsx
```diff
@@ -0,0 +1,219 @@
...
+                    ViewportPositionerHorizontalPosition[
+                        FlyoutHorizontalPosition[this.props.defaultHorizontalPosition]
+                    ]
+                }
+                defaultVerticalPosition={
+                    ViewportPositionerVerticalPosition[
+                        FlyoutVerticalPosition[this.props.defaultVerticalPosition]
+                    ]
+                }
+                disabled={!this.props.visible}
```
scomea: Does this make sense?  When the positioner is disabled it behaves like a regular div, the positioning logic does not run but it is visible.

packages/fast-components-react-msft/src/flyout/flyout.tsx
chrisdholt: Sure - if these are null or undefined, they won’t get passed

packages/fast-components-react-msft/src/flyout/flyout.tsx
```diff
@@ -0,0 +1,222 @@
...
+                    ViewportPositionerHorizontalPosition[
+                        FlyoutHorizontalPosition[this.props.defaultHorizontalPosition]
+                    ]
+                }
+                defaultVerticalPosition={
+                    ViewportPositionerVerticalPosition[
+                        FlyoutVerticalPosition[this.props.defaultVerticalPosition]
+                    ]
+                }
+                disabled={false}
```
scomea: This is the default, so not required

chrisdholt: thx - I’m leaving here as explicit for now

Issue #1935, opened 2019-07-11
Move the ltr/rtl selection to the right of the top bar in the Explorer


janechu:
There is a button to toggle ltr/rtl in the top left of the canvas view in `@microsoft/fast-component-explorer`, this should be moved to the top right along with other view design system controls. It should also use the MSFT toggle component.

Issue #1934, opened 2019-07-11
Add light and dark theme switching to the Explorer


janechu:
Add a way to toggle the light and dark themes in `@microsoft/fast-component-explorer`. This should use the msft toggle button and should appear with other styling controls in the top right of the canvas.

PR #1969, opened 2019-07-16
feat: add theme switching & move preview controls

eljefe223:
# Description
Update preview controls to use MSFT components, add toggle for theme switching and move controls to right side of canvas.

Closes: #1934
Closes: #1935 

Screen Shot 2019-07-16 at 9 37 51 AM
Screen Shot 2019-07-16 at 9 38 00 AM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-component-explorer/app/explorer.style.ts
+++ b/packages/fast-component-explorer/app/explorer.style.ts
@@ -3,9 +3,9 @@ import { DesignSystem } from "@microsoft/fast-components-styles-msft";
 
 export interface ExplorerClassNameContract {
     explorer?: string;
-    explorer_controls?: string;
-    explorer_viewerContrainer?: string;
-    explorer_viewerControlContrainer?: string;
+    explorer_viewerRegion?: string;
+    explorer_viewerControlRegion?: string;
+    explorer_viewerControls?: string;
 }
 
 const style: ComponentStyles<ExplorerClassNameContract, DesignSystem> = {
@@ -23,19 +23,19 @@ const style: ComponentStyles<ExplorerClassNameContract, DesignSystem> = {
         },
     },
     explorer: {},
-    explorer_controls: {
-        display: "flex",
-        padding: "7px 10px",
-        marginLeft: "auto",
-        alignItems: "center",
-    },
-    explorer_viewerContrainer: {
+    explorer_viewerRegion: {
         display: "flex",
         flexDirection: "column",
         width: "100%",
     },
-    explorer_viewerControlContrainer: {
+    explorer_viewerControlRegion: {
         display: "flex",
+    },
+    explorer_viewerControls: {
+        display: "flex",
+        padding: "7px 10px",
+        marginLeft: "auto",
+        alignItems: "center",
     },
 };
 
--- a/packages/fast-component-explorer/app/explorer.style.ts
+++ b/packages/fast-component-explorer/app/explorer.style.ts
@@ -1,6 +1,5 @@
-import { ComponentStyles, ManagedClasses } from "@microsoft/fast-jss-manager-react";
-import { DesignSystem, neutralLayerL1 } from "@microsoft/fast-components-styles-msft";
-import { reduce } from "lodash-es";
+import { ComponentStyles } from "@microsoft/fast-jss-manager-react";
+import { DesignSystem } from "@microsoft/fast-components-styles-msft";
 
 export interface ExplorerClassNameContract {
     explorer?: string;
--- a/packages/fast-component-explorer/app/explorer.tsx
+++ b/packages/fast-component-explorer/app/explorer.tsx
@@ -177,8 +177,8 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
                 this.getComponentNameSpinalCaseByPath(locationPathname)
             ),
             locationPathname,
+            theme: ThemeName.light,
             viewConfig: Object.assign({}, DesignSystemDefaults, {
-                theme: ThemeName.light,
                 direction: Direction.ltr,
             }),
         };
@@ -565,15 +565,12 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
 
     private handleUpdateTheme = (): void => {
         this.setState({
+            theme:
+                this.state.theme === ThemeName.light ? ThemeName.dark : ThemeName.light,
             viewConfig: merge({}, this.state.viewConfig, {
-                theme:
-                    this.state.viewConfig.theme === ThemeName.light
-                        ? ThemeName.dark
-                        : ThemeName.light,
-                backgroundColor:
-                    this.state.viewConfig.theme === ThemeName.light ? dark : light,
+                backgroundColor: this.state.theme === ThemeName.light ? dark : light,
                 neutralPalette: this.getNeutralPalette(
-                    this.state.viewConfig.theme === ThemeName.light ? dark : light
+                    this.state.theme === ThemeName.light ? dark : light
                 ),
             }),
         });
--- a/packages/fast-component-explorer/app/explorer.tsx
+++ b/packages/fast-component-explorer/app/explorer.tsx
@@ -167,7 +167,7 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
 
         this.resolveSchemaById = memoize(this.getSchemaById);
 
-        const paletteSource: ColorRGBA64 | null = parseColor("#808080");
+        const paletteSource: ColorRGBA64 | null = parseColor(light);
 
         this.state = {
             dataLocation: "",
@@ -177,16 +177,10 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
                 this.getComponentNameSpinalCaseByPath(locationPathname)
             ),
             locationPathname,
-            viewConfig: {
-                direction: Direction.ltr,
+            viewConfig: Object.assign({}, DesignSystemDefaults, {
                 theme: ThemeName.light,
-                backgroundColor: DesignSystemDefaults.backgroundColor,
-                designSystem: Object.assign({}, DesignSystemDefaults, {
-                    neutralPalette:
-                        paletteSource !== null ? createColorPalette(paletteSource) : [],
-                    direction: Direction.ltr,
-                }),
-            },
+                direction: Direction.ltr,
+            }),
         };
     }
 
@@ -213,13 +207,13 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
                                     <div
                                         className={get(
                                             this.props,
-                                            "managedClasses.explorer_viewerContrainer"
+                                            "managedClasses.explorer_viewerRegion"
                                         )}
                                     >
                                         <div
                                             className={get(
                                                 this.props,
-                                                "managedClasses.explorer_controls"
+                                                "managedClasses.explorer_viewerControls"
                                             )}
                                         >
                                             {this.renderScenarioSelect()}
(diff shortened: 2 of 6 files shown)


packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -100,19 +125,63 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+    ];
+
     constructor(props: ExplorerProps) {
         super(props);
 
         const locationPathname: string = get(this.props, "location.pathname", "");
 
         this.resolveSchemaById = memoize(this.getSchemaById);
+
+        const neutralColorPalette: ColorRGBA64 | null = parseColor("#808080");
```
nicholasrice: This isn't really a palette - its a palette *source*

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -123,13 +192,24 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                    direction: Direction.ltr,
+                }),
             },
         };
     }
 
     public render(): React.ReactNode {
         return (
-            <DesignSystemProvider designSystem={DesignSystemDefaults}>
+            <DesignSystemProvider
```
nicholasrice: This will create a new object every render cycle and cause components to recalculate styles every render cycle. You should store this as a variable and assign the variable to the `designSystem` prop

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -100,19 +125,63 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                return format(
+                    "0 {0} 0 {1}",
+                    toPx(designUnit)(designSystem),
+                    horizontalSpacing()(designSystem)
+                );
+            },
         },
     };
 
     private resolveSchemaById: ((id: string) => any) & MemoizedFunction;
```
nicholasrice: I don't see `this.themes` accessed anywhere - is this necessary?

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -223,9 +314,43 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                    this.props,
+                    "managedClasses.explorer_viewerControlContrainer"
+                )}
+            >
+                <Label jssStyleSheet={this.labelStyleOverrides}>RTL</Label>
+                <Toggle
+                    jssStyleSheet={this.toggleStyleOverrides}
+                    inputId={"direction"}
+                    onClick={this.handleUpdateDirection}
+                    selectedMessage={""}
```
nicholasrice: These should probably have selected / unselected messages - otherwise it seems like we're using these toggles incorrectly and they should just be checkboxes. Thoughts?

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -223,9 +314,43 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                    this.props,
+                    "managedClasses.explorer_viewerControlContrainer"
+                )}
+            >
+                <Label jssStyleSheet={this.labelStyleOverrides}>Dark mode</Label>
+                <Toggle
+                    jssStyleSheet={this.toggleStyleOverrides}
+                    inputId={"theme"}
+                    onClick={this.handleUpdateTheme}
+                    selectedMessage={""}
```
nicholasrice: same note here regarding toggles vs checkboxes.

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -437,6 +575,41 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                h: hslColor.h,
+                s: hslColor.s,
+                l: 0.5,
+            });
+
+            if (augmentedHSLColor !== null) {
+                return createColorPalette(hslToRGB(augmentedHSLColor));
+            }
+        }
+        return [];
```
nicholasrice: Maybe return the palette from the default design system if this fails?

packages/fast-component-explorer/app/preview.style.ts
```diff
@@ -0,0 +1,36 @@
+import { ComponentStyles, ManagedClasses } from "@microsoft/fast-jss-manager-react";
+import { DesignSystem, neutralLayerL1 } from "@microsoft/fast-components-styles-msft";
+import { reduce } from "lodash-es";
+
+export interface ExplorerClassNameContract {
+    explorer?: string;
+}
+
+const style: ComponentStyles<ExplorerClassNameContract, DesignSystem> = {
```
nicholasrice: should this use the "explorer" class name?

It should probably have "preview"

packages/fast-component-explorer/app/preview.tsx
```diff
@@ -21,19 +27,25 @@ class Preview extends Foundation<{}, {}, PreviewState> {
...
         };
     }
 
     public render(): React.ReactNode {
         return (
-            <div style={this.getStyle()} dir={this.state.direction}>
+            <Background
+                className={get(this.props, "managedClasses.preview")}
+                value={this.state.backgroundColor}
+                drawBackground={true}
```
nicholasrice: `drawBackground` defaults to true - you probably don't need this.

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -223,9 +314,43 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                )}
+            >
+                <Label jssStyleSheet={this.labelStyleOverrides}>RTL</Label>
+                <Toggle
+                    jssStyleSheet={this.toggleStyleOverrides}
+                    inputId={"direction"}
+                    onClick={this.handleUpdateDirection}
+                    selectedMessage={""}
+                    unselectedMessage={""}
+                    statusMessageId={"difrection"}
```
nalogiudice: Typo "difrection"

packages/fast-component-explorer/app/explorer.tsx
nalogiudice: +1 I think toggle makes more sense in this case than a checkbox, but the messages should have content.

eljefe223: @bheston thoughts on this? I know we wanted to use toggles.

eljefe223: chatted offline. Making these props optional on `<Toggle/>`

packages/fast-component-explorer/app/explorer.tsx
eljefe223: chatted offline. Making these props optional on `<Toggle/>`

packages/fast-component-explorer/app/explorer.style.ts
```diff
@@ -1,8 +1,11 @@
-import { ComponentStyles, ManagedClasses } from "@microsoft/fast-jss-manager-react";
+import { ComponentStyles } from "@microsoft/fast-jss-manager-react";
 import { DesignSystem } from "@microsoft/fast-components-styles-msft";
 
 export interface ExplorerClassNameContract {
     explorer?: string;
+    explorer_controls?: string;
+    explorer_viewerContrainer?: string;
```
janechu: contrainer? even without the spelling issue, I would like this to be `explorer_viewerRegion`

packages/fast-component-explorer/app/explorer.style.ts
```diff
@@ -1,8 +1,11 @@
...
+    explorer_viewerControlContrainer?: string;
```
janechu: would change this to `explorer_viewerControlRegion`

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -30,31 +32,54 @@ import ReactDOM from "react-dom";
...
+    ColorHSL,
+    ColorRGBA64,
+    hslToRGB,
+    parseColor,
+    rgbToHSL,
+} from "@microsoft/fast-colors";
+import { toPx } from "@microsoft/fast-jss-utilities";
+
+const dark: string = "#111111";
+const light: string = "#FFFFFF";
```
janechu: should we be taking these from the design system?

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -113,6 +166,9 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
         const locationPathname: string = get(this.props, "location.pathname", "");
 
         this.resolveSchemaById = memoize(this.getSchemaById);
+
+        const paletteSource: ColorRGBA64 | null = parseColor("#808080");
```
janechu: also here, should this be taken from the design system?

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -373,22 +491,30 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
                 direction:
                     this.state.viewConfig.direction === Direction.ltr
                         ? Direction.rtl
                         : Direction.ltr,
-            },
+                designSystem: Object.assign({}, this.state.viewConfig.designSystem, {
+                    direction:
+                        this.state.viewConfig.direction === Direction.ltr
+                            ? Direction.rtl
+                            : Direction.ltr,
```
janechu: theoretically we can now simplify this to remove direction and just use the design system, we shouldnt need to determine it twice

packages/fast-component-explorer/app/explorer.tsx
eljefe223: How? you need to input a background value into the DesignSystem in order for it to generate the color values, We could use DesignSystemDefaults.backgroundColor for `const light` but we would still need to pass a hard coded value to `const dark`

eljefe223: We do have a const `black` and a `white` in msft styles in the `color.constants.ts` file but they are not exported by the package.

janechu: Maybe we should be, but if we aren't now then no problem.

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -437,6 +545,39 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+                return createColorPalette(hslToRGB(augmentedHSLColor));
+            }
+        }
+        return DesignSystemDefaults.neutralPalette;
+    }
+
+    private handleUpdateTheme = (): void => {
+        this.setState({
+            viewConfig: merge({}, this.state.viewConfig, {
+                theme:
```
janechu: I would remove theme from the viewConfig and just have it as part of state, so `this.state.theme` since it's not part of the design system and yet if you add it to viewConfig, it will be added to the previews <DesignSystemProvider /> which is incorrect

janechu: I would remove theme from the viewConfig and just have it as part of state, so this.state.theme since it's not part of the design system and yet if you add it to viewConfig, it will be added to the previews which is incorrect

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -437,6 +545,36 @@ class Explorer extends Foundation<ExplorerHandledProps, {}, ExplorerState> {
...
+
+    private handleUpdateTheme = (): void => {
+        this.setState({
+            theme:
+                this.state.theme === ThemeName.light ? ThemeName.dark : ThemeName.light,
+            viewConfig: merge({}, this.state.viewConfig, {
+                backgroundColor: this.state.theme === ThemeName.light ? dark : light,
+                neutralPalette: this.getNeutralPalette(
+                    this.state.theme === ThemeName.light ? dark : light
+                ),
```
janechu: you seem to be doing `this.state.theme === ThemeName.light ? dark : light` over and over, can you declare a const above this that is `const updatedTheme: ThemeName = this.state.theme === ThemeName.light ? dark : light` and use updatedTheme?

PR #1977, opened 2019-07-17
feat: add support to resize pane using left/right arrow keys

### LINKED ISSUE #1913: Add keyboarding support to Pane resize
Add the ability to resize Pane using only the keyboard. The benefit would be more accessibility! 

I think the ideal solution would be to let the user tab to the resize button, then resize left and right using the arrow keys. The focus state may need to include some kind of prompt or affordance for the arrow key usage.

chrisdholt:
# Description
closes #1913 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/pane/pane.tsx
+++ b/packages/fast-layouts-react/src/pane/pane.tsx
@@ -8,10 +8,11 @@ import {
 } from "./pane.props";
 import { west } from "../row";
 import rafThrottle from "raf-throttle";
-import { toPx } from "@microsoft/fast-jss-utilities";
+import { applyFocusVisible, toPx } from "@microsoft/fast-jss-utilities";
 import { ComponentStyles } from "@microsoft/fast-jss-manager-react";
 import Foundation, { HandledProps } from "@microsoft/fast-components-foundation-react";
 import { canUseDOM } from "exenv-es6";
+import { Direction, KeyCodes } from "@microsoft/fast-web-utilities";
 import { joinClasses } from "../utilities";
 
 /**
@@ -64,6 +65,10 @@ export const paneStyleSheet: ComponentStyles<PaneClassNamesContract, undefined> 
         "&:hover": {
             cursor: "ew-resize",
         },
+        ...applyFocusVisible({
+            opacity: "1",
+            transform: "scale(1)",
+        }),
         "&:active": {
             opacity: "1",
             transform: "scale(1)",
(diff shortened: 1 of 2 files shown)


packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -233,12 +238,30 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
...
             />
         );
     }
 
+    /**
+     * Handle keyPress
+     */
+    public onKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>): void => {
+        switch (e.keyCode) {
+            case KeyCodes.arrowLeft /* tslint:disable-next-line */:
```
nicholasrice: Can we add a shift key check and increment by 10 when shift is held?

packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -233,12 +238,30 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
...
+                this.setWidth(this.rootElement.current.clientWidth - 1);
```
scomea: Maybe make the "baseResizeIncrement" a const?  I could also see it as a something that someone might want to configure in props.

packages/fast-layouts-react/src/pane/pane.tsx
chrisdholt: @nicholasrice - added and updated. Thx

packages/fast-layouts-react/src/pane/pane.tsx
chrisdholt: I’m open to adding this as a feature down the road if requested, but would prefer to hold off for now. Incrementing by 10 with shift is a common pattern and this seems relatively easy to add modifications for if requested down the road (with a default to 10).

chrisdholt:
> Do we want to do the same for Row?

Yeah, we could do that in a new PR.

packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -233,12 +238,31 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
...
 
+    /**
+     * Handle keyPress
+     */
+    public onKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>): void => {
+        const isShift: boolean = e.shiftKey;
+
+        switch (e.keyCode) {
+            case KeyCodes.arrowLeft:
+                this.setWidth(this.rootElement.current.clientWidth - (isShift ? 10 : 1));
```
bekell: Assuming this math works in RTL?

bekell: I believe it should, might want to validate though.

chrisdholt: I’ve manually tested for quite a bit - we do need to add tests to this package all up though.

chrisdholt: Since we’re using Right/Left keys to increment I think we’ll need to modify based on RTL - thanks @bekell for catching that!

chrisdholt: @bekell I’m seeing a larger issue here with RTL support; I’m going to take a separate issue and address there as it affects the click behavior as well. I’d rather align/fix both events in a separate, unified PR.

PR #1983, opened 2019-07-18
feat: add ability to avoid tag rendering for background component

Shpool:
# Description
Change to background component which allows to avoid tag rendering if tag property is set to null.

## Motivation & context
Avoid of creation of redundant div elements if <background> is used only as background context provider.
https://github.com/microsoft/fast-dna/issues/1982

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/background/background.tsx
+++ b/packages/fast-components-react-msft/src/background/background.tsx
@@ -76,9 +76,12 @@ export default class Background extends Foundation<
                 designSystem={this.getDesignSystemOverrides(color)}
                 designSystemMergingFunction={this.props.designSystemMergingFunction}
             >
-                <this.tag {...this.unhandledProps()} style={style}>
-                    {this.props.children}
-                </this.tag>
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-msft/src/background/background.tsx
```diff
@@ -76,9 +76,12 @@ export default class Background extends Foundation<
                 designSystem={this.getDesignSystemOverrides(color)}
                 designSystemMergingFunction={this.props.designSystemMergingFunction}
             >
-                <this.tag {...this.unhandledProps()} style={style}>
-                    {this.props.children}
-                </this.tag>
+                {(this.tag && (
+                    <this.tag {...this.unhandledProps()} style={style}>
```
nicholasrice: Can you change the `BackgroundHandledProps` interface to accept null as a valid value also? Otherwise I don't think projects with strict null checks will be able to use this feature. Also, can you add a note in the prop interface that when tag is `null` no element will be rendered?

PR #1986, opened 2019-07-18
feat: add slider display value converter

scomea:
# Description
Adds a prop by which developers can convert a slider value before it is shown to the user (through screen readers at this point as developers already have control of labels).

## Motivation & context
This was an issue dev encountered and the idea that a developper could want to use a slider to select values in an array seemed like a general use case we should support.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/slider/README.md
+++ b/packages/fast-components-react-base/src/slider/README.md
@@ -33,5 +33,9 @@ The "value" prop is used for controlled mode.
 
 The "thumb" prop allows authors to provide custom render functions for the thumb element(s).
 
+The "valuetextStringFormatter" prop enables authors to provide a function to format the "aria-valuetext" attribute of the thumbs.
+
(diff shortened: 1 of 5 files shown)


packages/fast-components-react-base/src/slider/README.md
```diff
@@ -33,5 +33,9 @@ The "value" prop is used for controlled mode.
 
 The "thumb" prop allows authors to provide custom render functions for the thumb element(s).
 
+The "valuetextStringFormatter" prop enables authors to provide a function to format the "aria-valuetext" attribute of the thumbs.
+
+The "displayValueConverter" prop enables authors to provide a function that converts the underlying slider value to one appropriate to display to the user if those differ.  A likely scenario for this is when the slider is actually selection an array index and the correct thing to report to the user is the value for that index and not the index itself.  This is currently applied to the "aria-valuemin", "aria-valuemax" and "aria-valuenow" attributes of the slider thumbs.
```
eljefe223: This sounds weird. "...underlying slider value to one appropriate to display to the user if those differ." maybe

"...underlying slider value to one that is appropriate to display to the user if those differ."

PR #1987, opened 2019-07-18
fix: high contrast for slider component

### LINKED ISSUE #1984: Track on slider is not visible in windows high contrast modes
Slider track bar and knob are not visible in high contrast mode, which results in an inaccessible experience. 

Example:
1. Turn on windows high contrast
2. Open edge browser
3. Go to https://explore.fastdna.net/components/slider/
4. See that the track bar is not visible

The slider track bar and knob are both visible and map accordingly to HC colors

slider_hc

khamudom:
# Description

Set high contrast media query to slider label tick mark. slider thumb (rest, hover and active state). The background and foreground of the slider track.

High contrast 1
image
High contrast 2
image
High contrast black
image
High contrast white
image

closes #1984 

## Motivation & context

When user turns on high contrast the slider does not show up

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/slider/index.ts
+++ b/packages/fast-components-styles-msft/src/slider/index.ts
@@ -70,6 +70,9 @@ const styles: ComponentStyles<SliderClassNameContract, DesignSystem> = {
         "&:active": {
             background: neutralForegroundActive,
         },
+        "@media (-ms-high-contrast:active)": {
+            background: "ButtonText",
+        },
     },
     slider_thumb__lowerValue: {},
     slider_thumb__upperValue: {},
@@ -77,11 +80,17 @@ const styles: ComponentStyles<SliderClassNameContract, DesignSystem> = {
     slider_backgroundTrack: {
         ...applyCornerRadius(),
         background: neutralOutlineRest,
+        "@media (-ms-high-contrast:active)": {
+            background: "ButtonShadow",
+        },
     },
     slider_foregroundTrack: {
         ...applyCornerRadius(),
         background: neutralForegroundHint,
         transition: "all 0.1s ease",
+        "@media (-ms-high-contrast:active)": {
+            background: "ButtonHighlight",
+        },
     },
     slider__disabled: {
         ...applyDisabledState(),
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/slider/index.ts
```diff
@@ -70,18 +70,33 @@ const styles: ComponentStyles<SliderClassNameContract, DesignSystem> = {
         "&:active": {
             background: neutralForegroundActive,
         },
+        "@media (-ms-high-contrast:active)": {
+            background: "ButtonText",
+            "&:hover": {
+                background: "Highlight",
```
janechu: Seems like this is CSS2 keywords https://www.w3.org/TR/css-color-3/#css2-system

CSS2 keywords are marked as deprecated, why are we using them here?

packages/fast-components-styles-msft/src/slider/index.ts
```diff
@@ -70,18 +70,33 @@ const styles: ComponentStyles<SliderClassNameContract, DesignSystem> = {
         "&:active": {
             background: neutralForegroundActive,
         },
+        "@media (-ms-high-contrast:active)": {
```
janechu: should this specify black-on-white or white-on-black? can you provide screenshots for both?

packages/fast-components-styles-msft/src/slider/index.ts
nicholasrice: This is - AFAIK - the only way to tap into the Windows high contrast theme. This is what we do for all other components to support high contrast.

packages/fast-components-styles-msft/src/slider/index.ts
khamudom: when you specify `-ms-high-contrast:active` it covers all high contrast styling rules, including black-on-white and white-on-black

packages/fast-components-styles-msft/src/slider/index.ts
chrisdholt: Yeah, I think this is currently the design/mapping. This article by Greg Whitworth (MSFT Edge) lays it out in more detail: http://www.gwhitworth.com/blog/2017/04/how-to-use-ms-high-contrast/

khamudom: These keywords have been added as part of a new list of system colors.
https://github.com/w3c/csswg-drafts/issues/3804

janechu: Thanks, just wanted to get this on the record

PR #1988, opened 2019-07-18
feat: refine css prop editor create row behavior

scomea:
# Description
Some minor tweaks to the css property editor's "click to create row" behavior.  Rows extend to the width of the component to more reliably insert after the click, also check for top/bottom insertion when clicking outside rows.  Hide outlines that could show up on click.

## Motivation & context
Getting closer to matching behavior of edge css prop editor

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
+++ b/packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
@@ -294,14 +294,13 @@ export default class CSSPropertyEditor extends Foundation<
         if (e.defaultPrevented) {
             return;
         }
-        if (
+
+        this.createRow(
             !isNil(this.propertyEditorRef.current) &&
             e.nativeEvent.offsetY < this.propertyEditorRef.current.clientHeight / 2
-        ) {
-            this.createRow(0);
-        } else {
-            this.createRow(Object.keys(this.editData).length);
-        }
+                ? 0
+                : Object.keys(this.editData).length
+        );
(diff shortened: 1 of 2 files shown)


packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
```diff
@@ -294,7 +294,14 @@ export default class CSSPropertyEditor extends Foundation<
...
         }
-        this.createRow(Object.keys(this.editData).length);
+        if (
+            !isNil(this.propertyEditorRef.current) &&
+            e.nativeEvent.offsetY < this.propertyEditorRef.current.clientHeight / 2
+        ) {
+            this.createRow(0);
+        } else {
+            this.createRow(Object.keys(this.editData).length);
+        }
```
janechu: nit: you could do:

```
this.createRow(
    !isNil(this.propertyEditorRef.current) &&
    e.nativeEvent.offsetY < this.propertyEditorRef.current.clientHeight / 2
        ? 0
        : Object.keys(this.editData).length
);
```

janechu:
Looks like you need to run prettier, I think @chrisdholt is looking into a solution for prettier mis-firing.

packages/fast-tooling-react/src/css-property-editor/property-editor.tsx
scomea: yep

PR #1990, opened 2019-07-19
feat: add onResize callback to Pane

### LINKED ISSUE #1912: Add onResize to Pane props
Add an onResize handler to the Pane component, so that other components can update as needed when the pane is resized.

The ideal solution would be onResize as a Pane handled prop, that passed some sort of event to the parent.

chrisdholt:
# Description

closes #1912 
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/pane/pane.tsx
+++ b/packages/fast-layouts-react/src/pane/pane.tsx
@@ -253,20 +253,23 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
     public onKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>): void => {
         const isShift: boolean = e.shiftKey;
         const offset: number = isShift ? 10 : 1;
+        let width: number;
 
         switch (e.keyCode) {
             case KeyCodes.arrowLeft:
-                this.setWidth(this.rootElement.current.clientWidth - offset);
+                width = this.rootElement.current.clientWidth - offset;
                 break;
             case KeyCodes.arrowRight:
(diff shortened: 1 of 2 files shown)


packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -262,6 +264,9 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
             default:
                 break;
         }
+
+        // Fire the resize callback
+        this.onResize(e);
```
janechu: do we also want to pass the width to this callback?

chrisdholt: Yeah, I think that’s valuable. Good call @janechu

chrisdholt: Added width as an optional param.

packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -145,6 +146,7 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
             width: this.props.initialWidth,
         };
 
+        this.onResize = throttle(this.onResize, 16);
```
scomea: Should we be throttling callbacks or letting authors do it at the other end?  My thought would be that we should send out a resize event in precise synch with our own calls to setState to update the internal size of the pane so that an author's changes can (theoretically) take effect on the same render pass.  So either not throttle at all, or if we throttle we do so such that our internal calls that set the width in state are in the same throttled code that sends out the event?

PR #2000, opened 2019-07-22
fix: resolve plugins against objects without props

nicholasrice:
# Description
Plugins were only being resolve against data structures that had a "props" key. This meant that a react child would not have a plugin resolved if `children` was of a type other than object, meaning strings, booleans, and numbers could would not be correctly resolved. This fixes that issue.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/data-utilities/location.spec.ts
+++ b/packages/fast-tooling-react/src/data-utilities/location.spec.ts
@@ -380,14 +380,18 @@ describe("getDataLocationsOfPlugins", () => {
             };
         }
 
-        ["", 1, true].map((value: unknown) => getDataLocationsOfPlugins(
-            childrenSchema,
-            factory(value),
-            childOptions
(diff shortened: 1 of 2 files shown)


packages/fast-tooling-react/src/data-utilities/location.spec.ts
```diff
@@ -358,6 +358,46 @@ describe("getDataLocationsOfPlugins", () => {
...
+        });
+
+        //        const dataLocationsOfPluginsWithStringChild: PluginLocation[] = getDataLocationsOfPlugins(
+        //            childrenSchema,
+        //            factory("string"),
+        //            childOptions
+        //        );
+        //
+        //        expect(dataLocationsOfPluginsWithStringChild.length).toBe(1);
+        //        expect(dataLocationsOfPluginsWithStringChild[0].dataLocation).toBe("children.props.children.props.children.props.render");
```
janechu: still needed?

PR #2026, opened 2019-07-28
feat: add css border component

eljefe223:
# Description
Added css border component.

image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/css-editor/border/border.props.ts
+++ b/packages/fast-tooling-react/src/css-editor/border/border.props.ts
@@ -1,0 +1,48 @@
+import { ManagedClasses } from "@microsoft/fast-jss-manager-react";
+import { CSSBorderClassNameContract } from "./border.style";
+
+export enum BorderStyleValue {
+    dashed = "dashed",
+    dotted = "dotted",
+    double = "double",
+    groove = "groove",
+    hidden = "hidden",
+    inherit = "inherit",
+    initial = "initial",
+    inset = "inset",
+    none = "none",
+    outset = "outset",
+    ridge = "ridge",
+    solid = "solid",
+    unset = "unset",
+}
+
+export interface CSSBorderState {
+    borderColor?: string;
+    borderStyle?: BorderStyleValue;
+    borderWidth?: string
+}
+
+export interface CSSBorderValues {
+    border?: string;
+    borderColor?: string;
+    borderStyle?: BorderStyleValue;
+    borderWidth?: string;
+}
+
+export interface CSSBorderUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+export interface CSSBorderHandledProps
+    extends ManagedClasses<CSSBorderClassNameContract> {
+    /**
+     * The data
+     */
+    data?: CSSBorderValues;
+
+    /**
+     * The onChange callback
+     */
+    onChange?: (border: CSSBorderValues) => void;
+}
+
+export type CSSBorderProps = CSSBorderHandledProps & CSSBorderUnhandledProps;
--- a/packages/fast-tooling-react/src/css-editor/border/border.props.ts
+++ b/packages/fast-tooling-react/src/css-editor/border/border.props.ts
@@ -20,7 +20,7 @@ export enum BorderStyleValue {
 export interface CSSBorderState {
     borderColor?: string;
     borderStyle?: BorderStyleValue;
-    borderWidth?: string
+    borderWidth?: string;
 }
 
 export interface CSSBorderValues {
--- a/packages/fast-tooling-react/src/css-editor/border/border.props.ts
+++ b/packages/fast-tooling-react/src/css-editor/border/border.props.ts
@@ -24,7 +24,6 @@ export interface CSSBorderState {
 }
 
 export interface CSSBorderValues {
-    border?: string;
     borderColor?: string;
     borderStyle?: BorderStyleValue;
     borderWidth?: string;
--- a/packages/fast-tooling-react/src/css-editor/border/border.schema.json
+++ b/packages/fast-tooling-react/src/css-editor/border/border.schema.json
@@ -9,8 +9,31 @@
             "title": "CSS Values",
             "type": "object",
             "properties": {
-                "border": {
-                    "title": "Border",
+                "borderColor": {
+                    "title": "Color",
+                    "type": "string"
+                },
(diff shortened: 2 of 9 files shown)


packages/fast-tooling-react/src/css-editor/border/border.schema.json
```diff
@@ -0,0 +1,19 @@
...
+    "type": "object",
+    "properties": {
+        "data": {
+            "title": "CSS Values",
+            "type": "object",
+            "properties": {
+                "color": {
+                    "title": "Border",
+                    "type": "string"
+                }
```
janechu: I think this should be border not color?

packages/fast-tooling-react/src/css-editor/border/border.tsx
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public render(): React.ReactNode {
+        return (
+            <div className={get(this.props, "managedClasses.cssBorder")}>
+                <label className={get(this.props, "managedClasses.cssBorder_label")}>
+                    Border
+                </label>
+                <div className={get(this.props, "managedClasses.cssBorder_control")}>
+                    <div className={get(this.props, "managedClasses.cssBorder_borderInputContainer")} style={{background: this.state.borderColor}}>
```
janechu: should this be `cssBorder_colorInputRegion` like in the color component? It seems like it contains the input with type "color" so that naming makes sense to be repeated

janechu:
Whats the deal with default border values in browsers? Is there for example a consensus on browser for a default border width?

nicholasrice:
The chevron looks miss-aligned / too high and too close to the word  - is that intentional

packages/fast-tooling-react/src/css-editor/border/border.tsx
```diff
@@ -0,0 +1,129 @@
...
+        managedClasses: void 0,
+    };
+
+    constructor(props: CSSBorderProps) {
+        super(props);
+
+        this.state = {
+            borderColor: get(this.props.data, "borderColor", ""),
+            borderStyle: get(this.props.data, "borderStyle", BorderStyleValue.solid),
+            borderWidth: get(this.props.data, "borderWidth", ""),
```
janechu: whats the reasoning for keeping this in state?

packages/fast-tooling-react/src/css-editor/border/border.props.ts
```diff
@@ -0,0 +1,48 @@
...
+    borderColor?: string;
+    borderStyle?: BorderStyleValue;
+    borderWidth?: string;
+}
+
+export interface CSSBorderValues {
+    border?: string;
+    borderColor?: string;
+    borderStyle?: BorderStyleValue;
+    borderWidth?: string;
```
janechu: This suggests that we might return any subset of these values, but the reality is that we only ever return border, I think if that's what we expect you can remove `borderColor`, `borderStyle`, `borderWidth`

packages/fast-tooling-react/src/css-editor/border/border.tsx
```diff
@@ -0,0 +1,129 @@
...
+        this.setState({
+            borderStyle: e.target.value as BorderStyleValue,
+        });
+    };
+
+    private handleBorderColorOnChange = (
+        e: React.ChangeEvent<HTMLInputElement>
+    ): void => {
+        this.setState({
+            borderColor: e.target.value.toUpperCase(),
```
janechu: This could conceivably fall out of sync if these values are changed by an outside source.

I would actually suggest ditching `borderColor` etc. for now, this can be added as a feature afterwards because you will need to identify the sequence of these values in the object.

For example:
```ts
{
    borderWidth: "5px",
    border: "1px solid black",
    borderColor: "red"
}
```

Should identify as `borderColor` as `red`, `borderWidth` as `1px` and `borderStyle` as `solid`.

packages/fast-tooling-react/src/css-editor/border/border.schema.json
```diff
@@ -0,0 +1,19 @@
...
+    "title": "CSS Border",
+    "description": "A CSS border component's schema definition.",
+    "id": "@microsoft/fast-tooling-react/css-border",
+    "type": "object",
+    "properties": {
+        "data": {
+            "title": "CSS Values",
+            "type": "object",
+            "properties": {
+                "border": {
```
janechu: you want to change this to not have `border` but instead have `borderWidth`, `borderStyle`, `borderColor`

packages/fast-tooling-react/src/css-editor/border/border.tsx
```diff
@@ -0,0 +1,127 @@
...
+        managedClasses: void 0,
+    };
+
+    constructor(props: CSSBorderProps) {
+        super(props);
+
+        this.state = {
+            borderColor: get(this.props.data, "borderColor", "#000"),
+            borderStyle: get(this.props.data, "borderStyle", BorderStyleValue.solid),
+            borderWidth: get(this.props.data, "borderWidth", "1px"),
```
janechu: I would suggest that you save these as private variables, then where you set state for componentDidUpdate, use those to make sure if they are ever undefined they revert back to the defaults

packages/fast-tooling-react/src/css-editor/border/border.tsx
```diff
@@ -0,0 +1,127 @@
...
+        this.setState({
+            borderStyle: e.target.value as BorderStyleValue,
+        });
+    };
+
+    private handleBorderColorOnChange = (
+        e: React.ChangeEvent<HTMLInputElement>
+    ): void => {
+        this.setState({
+            borderColor: e.target.value,
```
janechu: per my comment above, you will want to change this to `e.target.value || this.defaultBorderColor` same with all the other handlers setting state for these values.

packages/fast-tooling-react/src/css-editor/editor.tsx
```diff
@@ -99,6 +101,13 @@ export default class CSSEditor extends Foundation<
         return pick(this.props.data, ["color"]);
     }
 
+    /**
+     * Gets the border value
+     */
+    private getBorderData(): CSSBorderValues {
+        return pick(this.props.data, ["border"]);
```
janechu: this should not be "border", should be "borderWidth", "borderColor", "borderStyle"

PR #2064, opened 2019-07-30
fix: allow the deletion via keyboard of dashes in property names

janechu:
# Description

## Motivation & context

Only transform the name of a property once it is being committed so that there is no need to store the trailing dash in state, this also allows the deletion of dashes.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/css-property-editor/property-editor-row.tsx
+++ b/packages/fast-tooling-react/src/css-property-editor/property-editor-row.tsx
@@ -1,12 +1,11 @@
 import React from "react";
-import { camelCase, get, isNil } from "lodash-es";
+import { get, isNil } from "lodash-es";
 import Foundation, {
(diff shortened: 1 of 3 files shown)


packages/fast-tooling-react/src/css-property-editor/property-editor-row.tsx
```diff
@@ -129,9 +128,7 @@ export default class CSSPropertyEditorRow extends Foundation<
      * a user to continue typing using spinal case
      */
```
nicholasrice: This comment isn't really accurate - do we even need this function anymore? Can we just parse the string with `spinalCase` prior to passing it into the input value?

janechu: Good point, I'll remove

PR #2076, opened 2019-07-31
feat: change Navigation selection UX and update UI to new design guidance

### LINKED ISSUE #2046: Add a selection state to the Navigation
The Navigation should have a selection state similar to that of the Navigation Menu, see #2035

janechu:
# Description

## Motivation & context

Updates to make the expand different from the selection of an item in Navigation and updates it to the new design.

Resolves #2046

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/navigation/navigation.style.ts
+++ b/packages/fast-tooling-react/src/navigation/navigation.style.ts
@@ -3,14 +3,15 @@ import { applyTriggerStyle, insetStrongBoxShadow } from "../style";
 import {
     accent,
     foreground300,
-    neutralLayerL3,
     neutralLayerL4,
     neutralOutlineActive,
+    neutralFillStealthSelected,
 } from "../style/constants";
 
 export interface NavigationClassNameContract {
     navigation?: string;
     navigation_item?: string;
+    navigation_itemExpandTrigger?: string;
     /**
      * @deprecated
      */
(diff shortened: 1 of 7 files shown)


packages/fast-tooling-react/src/navigation/navigation.style.ts
```diff
@@ -101,12 +119,21 @@ const styles: ComponentStyles<NavigationClassNameContract, {}> = {
         cursor: "inherit",
         textDecoration: "none",
         fontStyle: "italic",
-        "&:focus": {
```
eljefe223: I think we still want focus? Also should the expand collapse arrows have focus?

eljefe223:
image

looks like we can get in a state where 2 elements appear selected. Also selecting the child element in this case collapses the node.

packages/fast-tooling-react/src/navigation/navigation.style.ts
eljefe223: I think we still want focus. Also should the expand collapse arrows have focus?

janechu:
This is an existing issue and happens on non default react children, tracking now as #2082

packages/fast-tooling-react/src/navigation/navigation.style.ts
janechu: Added back in focus visible, the expand buttons shouldn't get focus (i believe)

PR #2085, opened 2019-08-01
feat: addresses an issue where duplication of children would not work when there was only a single child, also allows the duplication of string children

### LINKED ISSUE #2080: extend duplicate functionality in Navigation component
- add ability to duplicate non-component children in the react "children" property
- add ability to duplicate component and non-component children in any react children property

Updates following the feature added in #2069
https://github.com/microsoft/fast-dna/pull/2069

janechu:
# Description

## Motivation & context

This addresses issues where:
- Primitive children (strings, etc) could not be copied
- items in arrays could not be copied
- children in non-react default properties could not be copied

Closes #2080 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/navigation/navigation.utilities.ts
+++ b/packages/fast-tooling-react/src/navigation/navigation.utilities.ts
@@ -758,38 +758,32 @@ export function getDataWithDuplicate<T>(sourceDataLocation: string, data: T): T 
     if (isInArray(clonedData, normalizedSourceDataLocation)) {
         duplicateDataInArray(clonedData, normalizedSourceDataLocation);
     } else {
-        // check for case of single child as object
-        const sourceDataLocationSegments: string[] = sourceDataLocation.split(".");
-        if (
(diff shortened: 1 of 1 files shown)


packages/fast-tooling-react/src/navigation/navigation.utilities.ts
```diff
@@ -758,38 +758,32 @@ export function getDataWithDuplicate<T>(sourceDataLocation: string, data: T): T
...
-        }
+        duplicateData(clonedData, normalizedSourceDataLocation);
     }
 
     return clonedData;
 }
 
+function duplicateData(data: unknown, sourceDataLocation: string): void {
+    const sourceData: unknown = get(data as object, sourceDataLocation);
+    const targetData: unknown = [sourceData, sourceData];
```
nicholasrice: this isn't really unknown - its an array of unknowns

janechu: That’s true

PR #2105, opened 2019-08-04
fix: set opacity to 1 on high contrast

### LINKED ISSUE #2104: High contrast disabled state are all hard to see
When our components are disabled they could be using the correct disabled color, but the low opacity makes it hard what it is.
 
1. Turn on Windows high contrast
2. Open Edge browser (I use Spartan Edge)
3. Go to: https://explore.fast.design/components
4. On any component, select disabled, on any of the components

We should see the components at 100% opacity

some examples:
image
image
image
image
image

khamudom:
# Description
All of our components, when disabled, in high contrast, are hard to see.  
Setting opacity to 1 on `-ms-high-contrast:active` media query.

**Note:** 
This only resolves the issue of making disabled easy to read. I have issues, assigned to me, to make sure the components are using the correct "disabled" keyword.

closes #2104 

image
image
image

## Motivation & context
When we lower opacity in high contrast we are making it worst for users to see.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/disabled.ts
+++ b/packages/fast-components-styles-msft/src/utilities/disabled.ts
@@ -10,5 +10,8 @@ export function applyDisabledState(
     return {
         opacity: toString(disabledOpacity),
         ...applyCursorDisabled(),
+        "@media (-ms-high-contrast:active)": {
+            opacity: "1",
+        },
     };
 }


packages/fast-components-styles-msft/src/utilities/disabled.ts
```diff
@@ -10,5 +10,8 @@ export function applyDisabledState(
     return {
         opacity: toString(disabledOpacity),
         ...applyCursorDisabled(),
+        "@media (-ms-high-contrast:active)": {
```
chrisdholt: This solution seems to create a new problem for users - how do they tell between a disabled control and a non-disabled control? While opacity certainly does make it harder to see, I don’t think this *the* best solution to solve this problem. Instead of the disabled control being difficult to see, it’s now difficult to tell the two apart unless your cursor is over it.

chrisdholt:
This would be a fix and not a chore; please review the PR template if need be: https://github.com/Microsoft/fast-dna/blob/master/.github/pull_request_template.md

packages/fast-components-styles-msft/src/utilities/disabled.ts
khamudom: I did make a note saying, this is one part of the fix. the components need the correct keyword, letting users know it is disabled. There are a couple things we can do here. Make sure all components, have the correct keyword in the disabled classname, then add this change. 
or
we leave this alone and put the keyword in each disabled classname. I figured fixing this in the utility file would makes sense, because it fixes all components.

chrisdholt: Can you elaborate on what you mean by the “disabled” keyword? Do you mean the `disabled` HTML attribute? I’m not sure that will scale across 100% of needed cases. In order to style across browsers, we need to enhance traditionally native controls which means instead of always being able to use `disabled` we may need to use something alternative such as `aria-disabled`. 

It may help to spell out the details of the proposed solution so we can better understand how this problem can be systematically solved throughout the repo.

khamudom: Within the high contrast media query, you can sets the background, color, borders etc. to a high contrast keyword, like "ButtonText", "GrayText", "Background" etc.
https://github.com/w3c/csswg-drafts/issues/3804

The keyword are mapped to MSFT high contrast system colors, so this is a PC-only feature.
image

example syntax:
"@media (-ms-high-contrast:active)": {
            borderColor: "ButtonText",
        },

chrisdholt: I understand how HC works. I’m curious what the overall proposed solution is to the problem here. This partial fix actually exacerbates the problem as it makes disabled controls the same opacity as non-disabled controls. I’m not understanding the full solution being proposed here and I’m confused by what you’re referring to when you say, “correct keyword in the disabled classname”. 

It will be helpful to understand how making the inputs the same opacity work into your overall solution to the problem linked above.

khamudom: In Windows high contrast theme, an element or state are mapped to a color.
In the case of a disabled element, anywhere you see a maroon(in screenshot) color element, users will know that is disabled.
image

To make sure users know that our components are in a disabled state, the style has to be set to `"GrayText"`. This is the keyword that is mapped to Windows high contrast "Disable Text"
We have a couple of issues in our components, 
1. they are not rendering with the correct disabled color
2. the opacity is too low, and does not meet high contrast ratio.

In high contrast, using opacity is incorrect when trying to convey disabled state, because we are taking the ability away for users to see the component. What MSFT does is use a theme color to let users know it is in a disabled state
You bring up the point that if we turn up opacity, then users cannot distinguish between disabled and non disabled state. The users will be able to tell by the "Disable Text" color theme that is set in high contrast settings, given it be in Black, White or custom.

We have a couple of solutions to meet MSFT high contrast standard regarding disabled state.
- In disabled utilities file, set opacity to 1.  This covers all places that uses disabled function
**BUT** with this approach, we need to make sure all disabled state styles, on the components, are set to using "DisableText" keyword in the style.
- We leave disabled utilities alone and just set the opacity fix in the individual component's style file.

chrisdholt: Thanks for the update with additional context and proposal. I’m fine with either 1 or 2, but the changes need to be made together. If we push through *just* this portion of the fix, we either have to wait for all the fixes related to this to come in or we release a partial fix with a regression. 

I’m fine with multiple commits, but if that’s the direction we want to go we should do so to a feature branch and not master. That’s likely a better path forward than a monolithic PR, which I think is what you intended with this partial change, but in this case it doesn’t help. Should we target a feature branch here instead of master so you can get the entire thing done and then commit it as a singular PR to master?

khamudom: Yes, a feature branch would be a good idea. Getting them all done together would be great. I will create a feature branch off master called `high-contrast-featured-branch`

khamudom:
This work is being done in feature branch along with other high contrast work.

PR #2112, opened 2019-08-06
feat: remove ColorRGBA64 instance exports

nicholasrice:
# Description
BREAKING CHANGE: Removes named `ColorRGBA64` exports

There is a significant terser bug preventing tree-shaking of class instances. This means that all named colors are being bundled regardless of if they are used or not. Re-factoring to a function to retrieve a color by name allows us to properly tree-shake named colors when not explicitly imported, saving roughy 4kb of parsed size and 1kb gzipped.

## Motivation & context
Reducing bundle size

## Issue type checklist

## Process & policy checklist


packages/fast-colors/src/color-converters.ts
```diff
@@ -16,18 +16,6 @@ import { degreesToRadians, radiansToDegrees } from "./math-utilities";
...
- * Get the relative luminance of a color.
- * Adjusts the color to sRGB space, which is necessary for the WCAG contrast spec.
- * The alpha channel of the input is ignored.
- * @param rgb The input color
- *
- * @deprecated Use rgbToRelativeLuminance instead.
- */
-export function rgbToLuminance(rgb: ColorRGBA64): number {
-    return rgbToRelativeLuminance(rgb);
-}
```
janechu: should we wait for a breaking change before removing this?

packages/fast-colors/src/named-colors.ts
```diff
@@ -0,0 +1,894 @@
+import { ColorRGBA64Config } from "./color-rgba-64";
+export type NamedColors =
```
janechu: can we get a documentation link to where these colors are defined, i assume mdn has these listed

packages/fast-colors/src/color-converters.ts
nicholasrice: this *is* a breaking change

packages/fast-colors/src/named-colors.ts
nicholasrice: added

PR #2117, opened 2019-08-08
fix: export observer support code from base package

scomea:
<!--- Provide a summary of your changes in the title field above. For guidance on formatting, see the # Description
Added support code for intersection observer and resize observer to the package exports.

## Motivation & context
Need to use observers in other packages.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/utilities/index.ts
+++ b/packages/fast-components-react-base/src/utilities/index.ts
@@ -13,7 +13,8 @@ declare global {
     }
 }
 
-export { DisplayNamePrefix };
+export { DisplayNamePrefix, WindowWithIntersectionObserver, WindowWithResizeObserver };
+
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/utilities/index.ts
```diff
@@ -13,7 +13,8 @@ declare global {
     }
 }
 
-export { DisplayNamePrefix };
+export { DisplayNamePrefix, WindowWithIntersectionObserver, WindowWithResizeObserver };
```
janechu: since you are wholesale exporting this it will include the export of `DisplayNamePrefix` which I don't think we want, that's more of an private variable just for the package.

scomea: Yep, good catch.

PR #2119, opened 2019-08-08
fix: slider moves to nearest increment on click

scomea:
# Description
Slider was previously always rounding down on clicks between increments, now it moves to the nearest one

## Motivation & context
Previous behavior was incorrect and made it nearly impossible to click on the track to get to the highest value increment on the track

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/slider/slider.spec.tsx
+++ b/packages/fast-components-react-base/src/slider/slider.spec.tsx
@@ -684,6 +684,8 @@ describe("Slider", (): void => {
             />
         );
 
+        // constrainToStep = (value: number, step: number): number 
+        // where value is the number to be constrained and step the step increment
         expect(rendered.instance()["constrainToStep"](14, 10)).toBe(10);
         expect(rendered.instance()["constrainToStep"](16, 10)).toBe(20);
     });
--- a/packages/fast-components-react-base/src/slider/slider.spec.tsx
+++ b/packages/fast-components-react-base/src/slider/slider.spec.tsx
@@ -673,6 +673,22 @@ describe("Slider", (): void => {
         document.body.removeChild(container);
     });
 
+    test("constrainToStep rounds fractions of a step properly", (): void => {
+        const rendered: any = mount(
+            <Slider
+                range={{
+                    minValue: 0,
+                    maxValue: 100,
+                }}
+                managedClasses={managedClasses}
+            />
+        );
+
+        expect(rendered.instance()["constrainToStep"](14, 10)).toBe(10);
+        expect(rendered.instance()["constrainToStep"](16, 10)).toBe(20);
+
+    });
+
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -1135,10 +1135,14 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
-            this.props.range.minValue
-        );
+        let constrainedValue: number =
+            Math.floor((value - this.props.range.minValue) / step) * step;
+        constrainedValue =
+            (value - constrainedValue) / step >= 0.5
+                ? constrainedValue + step
+                : constrainedValue;
+
+        return constrainedValue + this.props.range.minValue;
```
janechu: Can you not use `%`:
```ts
const remainder = value % step;
const updatedValue = remainder >= (step / 2) // check to see if this is over half a single step
    ? value - remainder + step // if so add a step
    : value - remainder;

return updatedValue <= this.props.range.minValue
    ? this.props.range.minValue
    : updatedValue > this.props.range.maxValue
        ? this.props.range.maxValue
        : updatedValue;
```

In this way you can skip your if (step === 0) as well.

packages/fast-components-react-base/src/slider/slider.spec.tsx
```diff
@@ -673,6 +673,21 @@ describe("Slider", (): void => {
         document.body.removeChild(container);
     });
 
+    test("constrainToStep rounds fractions of a step properly", (): void => {
+        const rendered: any = mount(
+            <Slider
+                range={{
+                    minValue: 0,
+                    maxValue: 100,
```
janechu: you didn't supply a step here?

scomea: The function we are testing is called directly and takes step as a second parameter. 
"expect(rendered.instance()["constrainToStep"](16, 10)).toBe(20);"

janechu: I see, I expected it to be part of the range property like it is in JSON schema - https://json-schema.org/understanding-json-schema/reference/numeric.html#multiples and since step is an attribute added to the input along with min/max this test seemed unintuitive from a prop perspective

scomea: I will add a note identifying the params on the function test

packages/fast-components-react-base/src/slider/slider.tsx
scomea: Range checking is a different fn, so:
       const remainder = value % step;
        const constrainedValue = remainder >= (step / 2) // check to see if this is over half a single step
            ? value - remainder + step // if so add a step
            : value - remainder;

        return constrainedValue + this.props.range.minValue;

PR #2122, opened 2019-08-09
feat: adds abstraction for conditionally applying class-names

nicholasrice:
# Description
Adds and implements a `classNames` function to conditionally add classes.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-web-utilities/src/class-names.ts
+++ b/packages/fast-web-utilities/src/class-names.ts
@@ -2,8 +2,6 @@ import { isFunction, isString } from "lodash-es";
 
 type ClassNamesArg = string | (() => string) | [string | (() => string), boolean];
 
-// const isString: (val: unknown) => val is string = (val: unknown): val is string => typeof val === "string"
-
 export function classNames(...args: ClassNamesArg[]): string {
     let classes: string = "";
     let leadingChar: string = "";
--- a/packages/fast-web-utilities/src/class-names.ts
+++ b/packages/fast-web-utilities/src/class-names.ts
@@ -1,0 +1,22 @@
+interface Foo {
+    [key: string]: boolean
+}
+type ClassNamesArg = (string | Foo);
+
+export function classNames(...args: Array<(string|{[key: string]: boolean})>): string {
+    let classes: string = "";
+
+    for (let i: number = 0, length: number = args.length; i < length; i++) {
+        const arg: ClassNamesArg = args[i];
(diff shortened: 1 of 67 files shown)


packages/fast-web-utilities/src/class-names.ts
```diff
@@ -0,0 +1,33 @@
+import { isFunction, isString } from "lodash-es";
+
+type ClassNamesArg = string | (() => string) | [string | (() => string), boolean];
+
+// const isString: (val: unknown) => val is string = (val: unknown): val is string => typeof val === "string"
```
chrisdholt: Looks like this commented out code can go bye-bye.

packages/fast-web-utilities/src/class-names.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function classNames(...args: ClassNamesArg[]): string {
+    let classes: string = "";
+    let leadingChar: string = "";
+    const addLeadingChar: (value: string) => string = (value: string): string =>
+        (leadingChar += value);
+
+    for (let i: number = 0, length: number = args.length; i < length; i++) {
```
chrisdholt: could `args.length` ever end up as undefined or null?

nicholasrice: no, it uses spread syntax (or the arguments keyword) so this will always have a `length` property. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

I'm testing for no arguments also

packages/fast-web-utilities/src/class-names.ts
```diff
@@ -0,0 +1,31 @@
...
+        } else if (Array.isArray(arg) && arg[1]) {
+            const value: string | (() => string) = arg[0];
+
+            if (isString(value)) {
+                classes += addLeadingChar(value);
+            } else if (isFunction(value)) {
+                classes += addLeadingChar(value());
+            }
+        }
+    }
```
janechu: Would this work?

```ts
return args.reduce((classNames: string, newClassName: ClassNamesArg): string {
    const className: [string | (() => string), boolean] = Array.isArray(newClassName)
        ? newClassName
        : [newClassName, true];

    return !className[1]
        ? classNames
        : isFunction(className[0])
            ? `${classNames} ${className[0]()}`
            : `${classNames} ${className[0]}`;
}).trim();
```

PR #2127, opened 2019-08-13
feat: increase perf on determining the navigation from the Form

### LINKED ISSUE #1999: Increase performance in Form when getting the navigation
The navigation is retrieved every time navigation is performed, causing a performance problem on very large JSON schemas.

To address this issue it should:
- Be abstracted to a class and only called when necessary
- Save the results of navigation
- Be able to find navigation relative to it's position more easily

janechu:
# Description

## Motivation & context

This resolves an issue with not caching the navigation which results in slow perf when navigating using the breadcrumbs or section links.

Resolves #1999

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/utilities/navigation.ts
```diff
@@ -0,0 +1,381 @@
+import { cloneDeep, get, memoize } from "lodash-es";
+import { MemoizedFunction } from "lodash";
```
nicholasrice: lodash?

packages/fast-tooling-react/src/form/utilities/navigation.ts
```diff
@@ -0,0 +1,381 @@
...
+    private config: NavigationConfig;
+
+    private dataLocation: string;
+
+    private reactChildrenDataLocations: string[];
+
+    private resolveNormalizeDataLocation: ((dataLocation: string) => string) &
+        MemoizedFunction;
+
+    private resolveDataLocations: ((dataLocation: string) => string[]) & MemoizedFunction;
```
nicholasrice: Do you need to type this as a `MemoizedFunction`? A memorzed function has the same signature as the callback - not sure why this needs extra type information.

packages/fast-tooling-react/src/form/utilities/navigation.ts
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    /**
+     * Gets an array of navigation items
+     */
+    public get(): NavigationItem[] {
+        const navigationItems: NavigationItem[] = this.resolveDataLocations(
+            this.dataLocation
+        ).map(
+            (dataLocation: string): NavigationItem => {
```
nicholasrice: You're looking this array twice with different map functions - you should be able to combine this into one loop I think

packages/fast-tooling-react/src/form/utilities/navigation.ts
```diff
@@ -0,0 +1,381 @@
...
+        );
+    }
+
+    private getDataLocationArray(dataLocation: string): string[] {
+        const dataLocations: Set<string> = this.getDataLocations(dataLocation);
+        const dataLocationIterator: IterableIterator<string> = dataLocations.values();
+        const dataLocationsArray: string[] = [];
+
+        // convert the set into of data locations into an array
+        for (let i: number = 0, size: number = dataLocations.size; i !== size; i++) {
```
nicholasrice: same here - you're iterating through twice - can you combine these?

packages/fast-tooling-react/src/form/utilities/navigation.ts
janechu: That was erroring when I tried to import from lodash-es, turns out I don't need it though not sure what that was about.

packages/fast-tooling-react/src/form/utilities/navigation.ts
janechu: Yep, turns out I didn't need it, not sure why it was erroring elsewhere when I tried to type it without using that.

packages/fast-tooling-react/src/form/utilities/navigation.ts
janechu: I actually need to do it twice because until its converted to an array I don't know how to resolve inherited default values, I've simplified it though!

packages/fast-tooling-react/src/form/utilities/navigation.ts
janechu: I'm using a Set here, so I need to convert it to an array

PR #2144, opened 2019-08-16
chore: add app insights sdk to component explorer website

awentzel:
# Description
This additional enables application telemetry integration using [Azure Insights](https://github.com/Microsoft/ApplicationInsights-js) and monitors your backend services and components after you deploy them to help you discover and rapidly diagnose performance and other issues.

## Motivation & context
All of our websites require telemetry and integration to Azure so we can properly monitor the health of our applications and proactively act when necessary.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-component-explorer/app/index.tsx
+++ b/packages/fast-component-explorer/app/index.tsx
@@ -3,6 +3,9 @@ import ReactDOM from "react-dom";
 import App from "./app";
 import HTML5Backend from "react-dnd-html5-backend";
 import { DragDropContext } from "react-dnd";
+import { ApplicationInsights } from "@microsoft/applicationinsights-web";
+import { ReactPlugin, withAITracking } from "@microsoft/applicationinsights-react-js";
+import { createBrowserHistory } from "history";
 
 /**
  * Create the root node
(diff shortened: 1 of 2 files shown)


janechu:
How have these changes been tested?

awentzel:
@janechu - these have been tested through Azure which when the app is running collects application data and enables a mechanism to view and analyze.

janechu:
Should we have separate instrumentation keys for development and production so that we are only gathering data from the live site? See: https://docs.microsoft.com/en-us/azure/azure-monitor/app/separate-resources

awentzel:
@janechu sure that can be added in a future PR.  Once this goes in there will be further testing and configuration necessary.  That would be a good point to add it.  Additionally, we'll want to break this out in a class of its own so it's easily replicated across websites source code.

janechu:
If we already know we want to break this out now, I would suggest that we include that work in this PR. Is there a reason we can't do that yet?

awentzel:
@janechu because I don't have the bandwidth to do the other stuff at this time which will also take further research and investigation.

nicholasrice:
I really don't think we should be checking this token into source control.

packages/fast-component-explorer/app/index.tsx
```diff
@@ -11,10 +14,30 @@ const root: HTMLElement = document.createElement("div");
...
-const DragDropApp: any = DragDropContext(HTML5Backend)(App);
+/**
+ * Create Application Insights SDK for Telemetry
+ * https://docs.microsoft.com/en-us/azure/azure-monitor/learn/nodejs-quick-start
+ */
+const browserHistory = createBrowserHistory({ basename: "" });
+var reactPlugin = new ReactPlugin();
+var appInsights = new ApplicationInsights({
+    config: {
+        instrumentationKey: "ecda4203-b38b-42ad-b181-f0f7f2b1d8fa",
```
nicholasrice: I understand this isn't a security threat, leaving this in source for the world to see opens the opportunity for this to get stolen and misused.

We need to tokenize this as part of a build step so private information stays private.

janechu:
@nicholasrice I agree, it leaves us open to data being passed that does not originate from the relevant sites.

I would like to see:
- Tokens used in place of generated keys
- Tokens for production and staging
- Factoring out the logic for use in both the explore site and the doc site if this is expected to happen in the future

packages/fast-component-explorer/app/index.tsx
awentzel: @nicholasrice Please read https://devblogs.microsoft.com/premier-developer/alternative-way-to-protect-your-application-insights-instrumentation-key-in-javascript/ and then let me know how you'd like to proceed.

awentzel:
@janechu - Could you comment as part of the review so they stay threaded.  For feature requests, your ideas are great but you should file an issue to be prioritized and allocated as availability allows.

packages/fast-component-explorer/app/index.tsx
janechu: *Added this to the original comment thread*

@nicholasrice I agree, it leaves us open to data being passed that does not originate from the relevant sites.

I would like to see:

Tokens used in place of generated keys
Tokens for production and staging
Factoring out the logic for use in both the explore site and the doc site if this is expected to happen in the future

nicholasrice: My understanding is that, with this implementation, anyone that comes across this key could pollute our data - is that correct?

nicholasrice: My understanding is that, with this implementation, anyone that comes across this key could pollute our data - is that correct?

PR #2150, opened 2019-08-17
feat: allow additional properties to be added as a dictionary of keys

### LINKED ISSUE #1652: Additional properties should be added to the `Form`
Any additional properties not specified by the JSON schema should still be displayed by the `Form` if they have been specified using the `additionalProperties` keyword. See: https://json-schema.org/understanding-json-schema/reference/object.html#properties

This does not include the `additionalProperties` being a boolean.

This requires the UI to be able to add arbitrarily named property keys, similar to the handling of "children". Equally similarly if the items are in data (and not defaults) then they should be removable. See also the "children" UI.

janechu:
# Description

Deciphers the `additionalProperties` keyword in JSON schemas to create a dictionary of properties based on the definition included in the keyword.

This allows:
- Adding a new key
- Modifying a key
- Removing a key

~The "soft remove" functionality inherent in all form controls allows for the removal of dictionary items.~ Soft remove is disabled for dictionary keys.

## Motivation & context

Closes #1652 

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/form/form-item.dictionary.tsx
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    private renderFormControls(): React.ReactNode {
+        return (typeof this.props.data !== "undefined"
+            ? Object.keys(this.props.data)
+            : []
+        )
+            .filter((dataKey: string) => {
+                return !this.props.enumeratedProperties.includes(dataKey);
+            })
```
nicholasrice: You could just use reduce and avoid multiple loops on the array.

just saying...

packages/fast-tooling-react/src/form/form/form-section.tsx
```diff
@@ -359,11 +400,23 @@ class FormSection extends React.Component<
...
                 </div>
             </div>
         );
     }
 
+    /**
+     * Get all enumerated properties for the object
+     */
+    private getEnumeratedProperties(schema: any): string[] {
+        if (typeof schema.properties === "undefined") {
```
nicholasrice: you could probably just check by value here - `schema.properties === undefined`

packages/fast-tooling-react/src/style/utilities.ts
```diff
@@ -292,6 +292,45 @@ export function applyRemoveItemStyle(): CSSRules<{}> {
...
+            background: foreground300,
+        },
+        "&::after": {
+            position: "absolute",
+            content: "''",
+            pointerEvents: "none",
+            width: "1px",
+            height: "9px",
+            left: "9.5px",
+            top: "5.5px",
```
nicholasrice: `"&::before, &::after"` will save some duplication here

packages/fast-tooling-react/src/form/form/form-item.dictionary.tsx
janechu: Darn it

PR #2155, opened 2019-08-19
feat: update modal dialog background to simple opacity dark mask

bheston:
# Description

Updated modal dialog background to simple opacity dark mask.

## Motivation & context

Aligning with the visual direction of related teams and responding to feedback regarding contrast in dark mode.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/dialog/index.ts
+++ b/packages/fast-components-styles-msft/src/dialog/index.ts
@@ -1,53 +1,45 @@
-import { DesignSystem, withDesignSystemDefaults } from "../design-system";
-import { ComponentStyles, ComponentStyleSheet } from "@microsoft/fast-jss-manager";
+import { DesignSystem } from "../design-system";
+import { ComponentStyles } from "@microsoft/fast-jss-manager";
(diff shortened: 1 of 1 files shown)


packages/fast-components-styles-msft/src/dialog/index.ts
```diff
@@ -1,51 +1,43 @@
...
+        right: "0",
+        overflow: "auto",
+    },
+    dialog_modalOverlay: {
+        position: "fixed",
+        top: "0",
+        left: "0",
+        right: "0",
+        bottom: "0",
+        background: "rgba(0, 0, 0, 0.3)",
```
chrisdholt: @bheston this should *never* theme whatsoever?

bheston: Right, always darker, which the transparency gives us. So if the neutral palette was red, it would still look and feel correct.

bheston:
@chrisdholt I don't see it as two changes. The only reason the design system was accessed as it was is because the acrylic still requires the old function call format, so now that it's gone none of that is needed anymore. That is, we wouldn't have been able to pull out the design defaults without the acrylic change, so they are coupled into one update.

PR #2160, opened 2019-08-20
Bundle size reductions

nicholasrice:
# Description
various bundle size reductions

## Issue type checklist

## Process & policy checklist


janechu:
If the intent is to have this be a breaking change and to add each of these commits to the git history, can we make sure to add the keywords "BREAKING CHANGE" into each of this commit messages? If it's the case that this will be squash merged then we can add it at that time, just let us know which one we're doing so we can verify.

nicholasrice:
@janechu we don't want to add BREAKING CHANGE to all commits because only one commit was breaking - I'll ensure it's added though because it appears to have gotten lost

PR #2162, opened 2019-08-20
feat: export individual keycodes as named exports and unreference KeyCodes

nicholasrice:
# Description
Exports keycodes as named exports so resulting bundles don't take dependencies on keycodes they don't need.

## Issue type checklist

## Process & policy checklist


scomea:
Isn't this a breaking change for anyone who is using utilities?

scomea:
Have we confirmed this actually improves perf/load time?

nicholasrice:
@scomea It is not breaking - we do not remove the export of the `KeyCodes` enum.

I've confirmed it reduces the bundle size of importing fast components - there is not a single component that imports even a majority of these values. The goal is to reduce bundle size. Reducing parsed bundle size has a positive impact on performance. Being an object, the keys of this export cannot be minified - naming these as individual exports circumvents that.

PR #2187, opened 2019-08-27
fix: improve slider thumb animations

scomea:
# Description
Added slider css modifier classes for dragging and incrementing which allowed the addition of transitions specifically intended for those modes vs. the single transition for all cases we had before.  Now when incrementing (i.e. holding the arrow key down) we have no easing and when dragging there is no transition at all.  Additionally the thumb now tracks dragging on stepped sliders and only snaps to the step on release

## Motivation & context
Slider animations were not as smooth as they should be

## Issue type checklist

## Process & policy checklist


eljefe223:
Smoked.

packages/fast-components-class-name-contracts-base/src/slider/index.ts
```diff
@@ -68,6 +68,16 @@ export interface SliderClassNameContract {
      */
     slider__disabled?: string;
 
+    /**
+     * The dragging modifier
+     */
+    slider__dragging?: string;
+
+    /**
+     * The incrmenting modifier
```
janechu: nit: spelling "incrementing"

PR #2190, opened 2019-08-28
fix: Updating README for managed-jss-react to remove typo

dzearing:
# Description

Minor typo fix!


dzearing:
Note that prettier kicked in here! I only was trying to change "improt" ;)

If you're good with the other changes, then I'll leave as is.

PR #2200, opened 2019-08-29
fix: :bangbang: single mode select menu changes selection on invoke, not focus

scomea:
# Description
Previously the select menu changed the selected state of the select as options got focus, now this only happens when an item is invoked by a click or keyboard enter or space.  A new callback has been added for those that may want to act on menu focus changes.

!!! NOTE: It is possible that some implementations based on the old behavior may not behave as expected with this change.  Please test your Select implementations  !!!

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2181

## Issue type checklist

## Process & policy checklist


janechu:
I'm not sure this should be marked as a breaking change since it is fixing unintended behavior.

scomea:
> I'm not sure this should be marked as a breaking change since it is fixing unintended behavior
The change could cause something that works with the old functionality to no longer work as expected.  I think we should flag it somehow.

khamudom:
We should have an example of both scenarios. Where the change happens when invoked and while menu item is focused.

PR #2203, opened 2019-08-30
fix: high-contrast style in Button, AccentButton, ActionToggle, ActionTrigger, Call to Action, NeutralButton, LightweightButton, OutlineButton, and StealthButton to support Edge icon components and add more common constants to high-contrast utility file.

khamudom:
# Description
While working on high contrast in Edge Web UI, I was noticing the icon components, from our icon packages, were not changing colors on hover. The fix was to set its fill property to the correct color on hover. We should not have to do this. The components should take care of setting the correct colors already.
After investigating, the `fill` property, in our button styles, was not set on hover. During the initial high contrast pass, we were testing to make sure text and icons were working properly, which they did, but did not test to see if icons from edge-icons or news-icons packages worked. If I had imported an edge-icon into a storybook example, I would have noticed that the fill property should have also been set on hover.
The fix is to set, as an example, `fill: "HighlightText"` (or whatever color it should be) to be included in the hover state. Now when any icon is being used in our Buttons, whether it is a beforeContent, afterContent, or an imported icon from one of our icon packages, the high contrast colors will be correct in all states.

There were also more common patterns that were be added to the `high-contrast` utility file.
Added:
- highContrastAccent
- highContrastSelectionOutline
- highContrastSelectionForeground
- highContrastSelectionBackground

The Button storybook example was also updated. The Stealth example was incorrect.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -134,7 +134,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
             ...highContrastSelectionOutline,
             "& $button_beforeContent, & $button_afterContent": {
                 fill: accentForegroundCut,
-                ...highContrastSelectionForeground
+                ...highContrastSelectionForeground,
             },
         },
         "&:active:enabled": {
@@ -145,7 +145,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
             "border-color": neutralFocus,
         }),
         "&:disabled": {
-            ...highContrastDisabledBorder
+            ...highContrastDisabledBorder,
         },
         "&::-moz-focus-inner": {
             border: "0",
@@ -204,7 +204,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
                 toPx<DesignSystem>(outlineWidth),
                 neutralOutlineHover
             ),
-            ...highContrastSelectionOutline
+            ...highContrastSelectionOutline,
         },
         "&:active:enabled": {
             background: "transparent",
@@ -239,7 +239,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
         background: neutralFillStealthRest,
         "&:hover:enabled": {
             "background-color": neutralFillStealthHover,
-            ...highContrastSelection
+            ...highContrastSelection,
         },
         "&:active:enabled": {
             "background-color": neutralFillStealthActive,
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -39,11 +39,16 @@ import {
 } from "../utilities/design-system";
 import { applyDisabledState } from "../utilities/disabled";
 import {
+    highContrastAccent,
     highContrastBackground,
     highContrastDisabledBorder,
     highContrastDisabledForeground,
     highContrastDoubleFocus,
+    highContrastOutline,
     highContrastOutlineFocus,
+    highContrastSelection,
+    highContrastSelectionForeground,
+    highContrastSelectionOutline,
     highContrastSelector,
     highContrastStealth,
 } from "../utilities/high-contrast";
@@ -89,6 +94,7 @@ const applyTransparentBackplateStyles: CSSRules<DesignSystem> = {
         [highContrastSelector]: {
             background: "ButtonFace !important",
             color: "ButtonText !important",
+            fill: "Highlight",
         },
     },
     "&:active:enabled": {
@@ -125,16 +131,10 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
         background: neutralFillRest,
         "&:hover:enabled": {
             background: neutralFillHover,
-            [highContrastSelector]: {
-                background: "Highlight !important",
-                "border-color": "ButtonText !important",
-                color: "HighlightText !important",
-            },
+            ...highContrastSelectionOutline,
             "& $button_beforeContent, & $button_afterContent": {
                 fill: accentForegroundCut,
-                [highContrastSelector]: {
-                    fill: "HighlightText",
-                },
+                ...highContrastSelectionForeground
             },
         },
         "&:active:enabled": {
@@ -145,22 +145,12 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
             "border-color": neutralFocus,
         }),
         "&:disabled": {
-            [highContrastSelector]: {
-                background: "ButtonFace",
-                "border-color": "GrayText",
-                color: "GrayText",
-            },
+            ...highContrastDisabledBorder
         },
         "&::-moz-focus-inner": {
             border: "0",
         },
-        [highContrastSelector]: {
-            background: "ButtonFace",
-            "border-color": "ButtonText",
-            color: "ButtonText",
-            fill: "ButtonText",
-            "-ms-high-contrast-adjust": "none",
-        },
+        ...highContrastOutline,
         "a&": {
             "&$button__disabled": {
                 "&:hover": {
@@ -179,6 +169,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
                 background: "HighlightText !important",
                 "border-color": "Highlight !important",
                 color: "Highlight !important",
+                fill: "Highlight",
             },
         },
         "&:active:enabled": {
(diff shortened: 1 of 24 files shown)


packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -94,6 +105,15 @@ export const highContrastDoubleFocus: CSSRules<DesignSystem> = {
...
 export const highContrastSelection: CSSRules<DesignSystem> = {
+    [highContrastSelector]: {
+        background: HighContrastColor.selectedBackground,
+        color: HighContrastColor.selectedText,
+        fill: HighContrastColor.selectedText,
+    },
+};
+
+// Used to set 'selected-text' color with an outline
+export const highContrastSelectionOutline: CSSRules<DesignSystem> = {
```
bheston: Looks like I missed one last time, but "selection" should be "selected". Can we rename the new ones and deprecate/alias the other since it's already been published. "Selection" sounds more like highlighted text or a single item in a list. "Selected" is a state like we're using it here. That will coordinate better with a new feature I'm about to do for non-HC selected colors.

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -89,6 +94,7 @@ const applyTransparentBackplateStyles: CSSRules<DesignSystem> = {
         [highContrastSelector]: {
             background: "ButtonFace !important",
             color: "ButtonText !important",
+            fill: "Highlight",
```
bheston: Shouldn't the icon fill be the same color as the foreground text?

packages/fast-components-styles-msft/src/button/index.ts
```diff
@@ -179,6 +169,7 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
                 background: "HighlightText !important",
```
bheston: Should this whole thing be `highContrastSelection` maybe with `border-color` added on? Or should the border always be a part of this combo?

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -110,6 +130,14 @@ export const highContrastForeground: CSSRules<DesignSystem> = {
     },
 };
 
+// Used to set foreground and glyph to be 'select-text' color
+export const highContrastSelectionForeground: CSSRules<DesignSystem> = {
+    [highContrastSelector]: {
+        color: "HighlightText !important",
```
bheston: Can all values in this file use the constants? Maybe add `important` to each color anyway.

packages/fast-components-styles-msft/src/button/index.ts
khamudom: This is a one off situation. in our Lightweight and Justified buttons, the beforeContent and afterContent do not get an underline indicator when a user hovers over it. It does slightly change color. I had to separate the color from the icon and the text, because on hover the text does not change, we just add an underline. If both the icon and text are the same color, and, they are both together, its fine when the user hovers, because you at least get an underline on the text. But if it is only the icon, and you hover over it, there is indication of hover., because the icon is is the same color in rest and hover state.
As I type this, I am thinking I can just change `color` to "Highlight", that should work.

bheston: Sounds good. Maybe add a comment. That file is a little hard to parse because of those types of situations.

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: create a function called `ImportantColor` to use here

packages/fast-components-styles-msft/src/button/index.ts
khamudom: put them in a const called `highContrastSelectedOutline`

packages/fast-components-styles-msft/src/button/index.ts
khamudom: updated

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: updated to be `selected`

janechu:
Can you change the title to "fix: add missing style properties to Button, AccentButton (etc.)" and fill in the etc. with all relevant components?

janechu:
The language is still off, we want the language to read more like what this is accomplishing, so "adds missing style properties to [list of affected components]".

When this is added to the changelog it will go under fixes with this line so it should be as descriptive as possible in what it is doing. I would also leave out "our", this doesn't make sense given the context (as a means of communication to developers taking a dependency on our package).

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -16,18 +16,22 @@ export enum HighContrastColor {
     background = "Background",
 }
 
+function ImportantColor(color: HighContrastColor): string {
+    return color + "!important";
```
janechu: no space between the color and !important, also prefer using back ticks here.

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -134,3 +174,17 @@ export const highContrastBackground: CSSRules<DesignSystem> = {
         background: HighContrastColor.buttonText,
     },
 };
+
+// Used to set background to be 'selected-text' color
+export const highContrastSelectionBackground: CSSRules<DesignSystem> = {
```
janechu: where is this getting used?

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -93,20 +108,45 @@ export const highContrastDoubleFocus: CSSRules<DesignSystem> = {
 };
 
 // Used to set 'selected-text' color
-export const highContrastSelection: CSSRules<DesignSystem> = {
+export const highContrastSelected: CSSRules<DesignSystem> = {
```
janechu: this looks like a breaking change :bangbang:

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: updated with back ticks

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: this is now being used

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: brought back the const and add deprecated comment

janechu: if this is just a name change you can make the deprecated version simply point to the new version, like so:
`export const highContrastSelection: CSSRules<DesignSystem> = highContrastSelected;`

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -16,18 +16,22 @@ export enum HighContrastColor {
     background = "Background",
 }
```
janechu: I can't comment on it (because github UI doesn't allow me) but the disabledText is marked with an !important, given the new Important utility, should you remove it here and add it in like you are doing with the rest of these?

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -16,18 +16,22 @@ export enum HighContrastColor {
...
 
+function ImportantColor(color: HighContrastColor): string {
```
nicholasrice: typically capitol letters are reserved for class constructors, enums, interfaces, and type aliases - can you rename to `importantColor`?

You should also consider abstracting - the only assumption this code really makes is that the argument type is a string. You could move this to a generic utility file and change the type signature to simply expect a string value. Then this could be used to add important to any css value

khamudom: abstract the function to its own utility

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: updated

packages/fast-components-styles-msft/src/utilities/important.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Sets a css value to be '!important'.
+ */
+export function importantValue(value: string): string {
```
bheston: Since this is intended for style sheets, and most style sheet functions are DesignSystemResolvers, can you change this to `value: string | DesignSystemResolver<string>` and then call `checkDesignSystemResolver(value)`?

khamudom: updated. thanks Brian

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: updated

Issue #2202, opened 2019-08-29
Add custom spin buttons to number field


chrisdholt:
Currently the browser adds inner spin buttons to the number field input type. These spin buttons cannot be styled easily as they are part of the shadow DOM. This creates problems as the spin buttons take the browser styling and cannot be styled to align with the design system. Considering this, the spin buttons should be hidden for the time being.

I think it's worth a discussion as to whether or not we should add custom spin buttons to the number field component.

Outstanding questions:
- What are the perf tradeoffs for including this behavior?
- Should this be done at the base or MSFT level? 
- Is work currently being done on the browser side to expose similar functionality?
- Should this be a render prop approach? This would allow users to opt into the behavior by passing their own spin buttons. We could build out the spin button as components which would make for an easier implementation.

mdtauk:
WinUI is in the process of adding a NumberBox control to its Xaml library.  There is a focus on it being touch friendly, but the same design could work for the Fast DNA Number Input Field.

![image](https://user-images.githubusercontent.com/7389110/63986740-35ba0280-cacd-11e9-9b21-400e996fcb15.png)

chrisdholt:
Thanks for the insight @mdtauk! I think we'd want this to support touch for sure. I'm leaning towards implementing a spin button specific to the increment and decrement functionality which the number field could optionally be composed with: https://w3c.github.io/aria-practices/#spinbutton

mdtauk:
Yes, it should conform to the W3C expected behaviour, but I don't think that prohibits aligning the Up and Down buttons horizontally, instead of stacking them vertically, where there is not a lot of height available in the TextField.

bheston:
Here is the design we proposed for the native platform controls update. We had to hold on it for now as the overlay proved to be difficult.
My goal was to maintain the up down visual relationship while supporting touch or generally larger hit points.
On the platform control we needed to maintain the narrow width due to an incredibly narrow 45px default field width.
![image](https://user-images.githubusercontent.com/47367562/64041733-18108b80-cb15-11e9-84e5-10ec39749a1c.png)

bheston:
@mdtauk for the WinUI control did you consider arrows instead of chevrons? Obviously there are lines and another "up" icon in the design you shared, but it might not be as discernible from a select box. Perhaps + and - would be good universal options too.

mdtauk:
[...]

PR #2204, opened 2019-08-30
fix: hide default spin button on number field

chrisdholt:
# Description
Currently the browser adds spin buttons to inputs with type of number. These buttons only support minimal styling and when coupled with certain MSFT design system changes, they look awful. This change serves to fix the visual problem. The longer term solution is proposed with #2202 

## Issue type checklist

## Process & policy checklist


bheston:
I think the spinners are desirable to have by default on this control. There have been some rendering issues in the browser, but it seems normal now and looks alright by default in Explorer. Can you post an example of what's not working?

chrisdholt:
> I think the spinners are desirable to have by default on this control. There have been some rendering issues in the browser, but it seems normal now and looks alright by default in Explorer. Can you post an example of what's not working?

These don't theme according to our system and they are squished in density driven scenarios. @Falkicon can provide more insight as he brought the issue to my attention.

Screen Shot 2019-08-30 at 9 52 08 AM

mdtauk:
> I think the spinners are desirable to have by default on this control. There have been some rendering issues in the browser, but it seems normal now and looks alright by default in Explorer. Can you post an example of what's not working?

I don't think the idea is to not have spinners when you add the control to the page, but to suppress the browser's buttons, and use custom ones which adapt better to density, and touch usage.

packages/fast-components-styles-msft/src/number-field/index.ts
```diff
@@ -13,6 +13,11 @@ const styles: ComponentStyles<NumberFieldClassNameContract, DesignSystem> = {
             format("0 6px 0 {0}", horizontalSpacing()),
             format("0 {0} 0 6px", horizontalSpacing())
         ),
+        "&::-webkit-inner-spin-button": {
+            appearance: "none",
+            "-webkit-appearance": "none",
+            "-moz-appearance": "none",
```
janechu: Firefox does not support `-webkit-inner-spin-button`.
Theoretically assigning it an appearance within this selector will not affect how the control looks on firefox.

chrisdholt: Yeah, I'll get this updated. I saw that it was changing the appearance in FF. Still need to land this with others.

janechu:
smoked, nice solution

PR #2210, opened 2019-09-03
feat: added input toggle fill recipe.

bheston:
# Description

Adds a new fill recipe to be used for toggled or selected state of input controls like checkboxes, radio buttons, and toggle switches.

## Motivation & context

The current implementation of these controls uses more of an outlined style in selected and unselected states. There has been some feedback and discussion about making the states more distinct. This new fill recipe opens the path for a soon-to-be-updated set of input controls that uses a more bold fill color on selected state.

The updates in the color explorer are temporary to illustrate and test the new colors, and will be rolled into the main components in another PR.

This also cleans up some code in other recipes that use contrast and deltas to determine swatch colors. I would like to apply this to accent color based recipes in the future, so looking for any points that may not work in that model where the desired color is not the *first* compliant contrast color, but the *closest* to the base accent color.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-color-explorer/app/color-blocks.tsx
+++ b/packages/fast-color-explorer/app/color-blocks.tsx
@@ -1,8 +1,17 @@
-import React from "react";
 import {
-    ButtonClassNameContract,
-    DividerClassNameContract,
-} from "@microsoft/fast-components-react-base";
+    Background,
+    Button,
+    ButtonAppearance,
+    Caption,
+    CaptionClassNameContract,
+    Checkbox,
+    Divider,
+    Hypertext,
+    Paragraph,
+    ParagraphClassNameContract,
+    TextField,
+} from "@microsoft/fast-components-react-msft";
+import React from "react";
 import {
     accentFillActive,
     accentFillHover,
@@ -24,6 +33,9 @@ import {
     neutralFillStealthHover,
     neutralFillStealthRest,
     neutralFillStealthSelected,
+    neutralFillToggleActive,
+    neutralFillToggleHover,
+    neutralFillToggleRest,
     neutralFocus,
     neutralFocusInnerAccent,
     neutralForegroundActive,
@@ -38,18 +50,9 @@ import manageJss, { ComponentStyleSheet } from "@microsoft/fast-jss-manager-reac
 import classnames from "classnames";
 import { get, isEqual, uniqueId } from "lodash-es";
 import {
-    Background,
-    Button,
-    ButtonAppearance,
-    Caption,
-    CaptionClassNameContract,
-    Checkbox,
-    Divider,
-    Hypertext,
-    Paragraph,
-    ParagraphClassNameContract,
-    TextField,
-} from "@microsoft/fast-components-react-msft";
+    CheckboxClassNameContract,
+    DividerClassNameContract,
+} from "@microsoft/fast-components-react-base";
 import { connect } from "react-redux";
 import { Omit } from "utility-types";
 import { ColorsDesignSystem } from "./design-system";
@@ -204,6 +207,31 @@ class ColorBlocksBase extends React.Component<ColorBlocksProps, ColorBlocksState
     > = {
         caption: {
             color: neutralForegroundHint,
+        },
+    };
+
+    private checkboxStyleOverrides: ComponentStyleSheet<
+        CheckboxClassNameContract,
+        ColorsDesignSystem
+    > = {
+        checkbox_input: {},
+        checkbox_stateIndicator: {},
+        checkbox__checked: {
+            "& $checkbox_input": {
+                background: neutralFillToggleRest,
+                "border-color": "transparent",
+                "&:hover": {
+                    background: neutralFillToggleHover,
+                },
+                "&:active": {
+                    background: neutralFillToggleActive,
+                },
+            },
+            "& $checkbox_stateIndicator": {
+                "&::before": {
+                    background: `url('data:image/svg+xml;utf8,<svg viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path fill="white" fill-rule="evenodd" clip-rule="evenodd" d="M8.143 12.6697L15.235 4.5L16.8 5.90363L8.23812 15.7667L3.80005 11.2556L5.27591 9.7555L8.143 12.6697Z"/></svg>')`,
+                }
+            },
         },
     };
 
@@ -535,7 +563,7 @@ class ColorBlocksBase extends React.Component<ColorBlocksProps, ColorBlocksState
                 />
                 <FocusSwatch />
                 {this.renderExample(
-                    <Checkbox inputId={uniqueId()}>
+                    <Checkbox inputId={uniqueId()} jssStyleSheet={this.checkboxStyleOverrides}>
                         <label slot="label">Checkbox</label>
                     </Checkbox>
                 )}
@@ -543,6 +571,24 @@ class ColorBlocksBase extends React.Component<ColorBlocksProps, ColorBlocksState
                     type={SwatchTypes.fill}
                     fillRecipe={neutralFillInputRest}
                     recipeName="neutralFillInputRest"
+                    foregroundRecipe={neutralForegroundRest}
+                />
+                <Swatch
+                    type={SwatchTypes.fill}
+                    fillRecipe={neutralFillToggleRest}
+                    recipeName="neutralFillToggleRest"
+                    foregroundRecipe={neutralForegroundRest}
+                />
+                <Swatch
+                    type={SwatchTypes.fill}
+                    fillRecipe={neutralFillToggleHover}
+                    recipeName="neutralFillToggleHover"
+                    foregroundRecipe={neutralForegroundRest}
+                />
+                <Swatch
+                    type={SwatchTypes.fill}
+                    fillRecipe={neutralFillToggleActive}
+                    recipeName="neutralFillToggleActive"
                     foregroundRecipe={neutralForegroundRest}
                 />
                 <Swatch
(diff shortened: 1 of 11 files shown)


packages/fast-components-styles-msft/src/utilities/color/recipe.ts
```diff
@@ -0,0 +1,113 @@
...
+    };
+}
+
+/**
+ * Function to derive accessible colors from contrast and delta configuration.
+ * Performs a simple contrast check against the colors and returns
+ * the color that has the most contrast against the background. If contrast
+ * cannot be retrieved correctly, function returns black.
+ */
+export function accessibleAlgorithm(
```
nicholasrice: Nice abstraction

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
```diff
@@ -0,0 +1,113 @@
...
+    palette: Palette | DesignSystemResolver<Palette>,
+    minContrast: number | DesignSystemResolver<number>,
+    restDelta: number | DesignSystemResolver<number>,
```
nicholasrice: Same question for deltas - at some level it enables circumventing the DesignSystem which could be a bad thing.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
```diff
@@ -0,0 +1,113 @@
...
+    hoverDelta: number | DesignSystemResolver<number>,
+    activeDelta: number | DesignSystemResolver<number>
```
nicholasrice: Do we need to support selected here?

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
```diff
@@ -0,0 +1,113 @@
+import {
```
nicholasrice: Note on the file name - everything about these color utilities relates to recipes - can we  just put this stuff in common? Or maybe accessability.ts?

packages/fast-color-explorer/app/color-blocks.tsx
```diff
@@ -207,6 +210,31 @@ class ColorBlocksBase extends React.Component<ColorBlocksProps, ColorBlocksState
...
+            "& $checkbox_input": {
+                background: neutralFillToggleRest,
+                "border-color": "transparent",
+                "&:hover": {
+                    background: neutralFillToggleHover,
+                },
+                "&:active": {
+                    background: neutralFillToggleActive,
+                },
+            },
```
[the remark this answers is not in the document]
bheston: It will. I was going to do a separate PR to implement the styles. This is for testing only.

packages/fast-components-styles-msft/src/design-system/index.ts
```diff
@@ -139,6 +139,13 @@ export interface DesignSystem {
     neutralFillStealthActiveDelta: number;
     neutralFillStealthSelectedDelta: number;
 
+    /**
+     * Configuration for the neutral-fill-toggle recipe.
+     */
+    neutralFillToggleRestContrast: number;
```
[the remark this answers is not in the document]
bheston: I think we will expose those soon. Perhaps there will be another way we want to support increased contrast, but for instance, to coordinate with Mac increased contrast mode, we will want to support increasing the default 4.5 up to 7.

In this case I wasn't thinking of it being tied as directly to W3C, but perhaps the base recommendation is. In testing the design for these components we evaluated lighter or darker values. I was considering this the equivalent of a "rest" delta like the other recipes.

Long term I would like to see a better way to define and configure recipes, which this falls into. For instance, maybe I want my outline recipe to be contrast based instead of offset based. Perhaps we accept either a contrast or an offset and let the recipe resolve from there.

I'm fine hard-coding it like the other recipes until we have a chance to think about a wider strategy.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
bheston: I'm actually trying to get rid of selected, so I didn't add it to the new recipe. We currently only use the base neutral fill for selected, and it's complicating a lot of the rules and usage patterns. One of the things removing it would address is the ability to then have hover and active state for a selected element.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
bheston: Yeah, funny name. I thought I had something better. I was trying to encapsulate it and avoid cluttering common. I'll think of something more descriptive.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
```diff
@@ -0,0 +1,113 @@
...
+    findClosestSwatchIndex,
+    findSwatchIndex,
+    getSwatch,
+    isDarkMode,
+    Palette,
+    swatchByContrast,
+} from "./palette";
+
+/**
```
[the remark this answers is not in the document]
bheston: It's just here for the swatchByContrast call. I could inline it, but it might be harder to read with all the parameters inline. I was considering default implementations of these swatchByContrast methods though, since most of the time we are doing the same thing.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * Function to derive accessible colors from contrast and delta configuration.
+ * Performs a simple contrast check against the colors and returns
+ * the color that has the most contrast against the background. If contrast
+ * cannot be retrieved correctly, function returns black.
+ */
+export function accessibleAlgorithm(
+    palette: Palette | DesignSystemResolver<Palette>,
```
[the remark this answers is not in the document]
bheston: I added this throughout to better support the swatchByContrast initial index resolver function. It provides a palette, but we never use it because findClosestSwatchIndex took only a palette resolver.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
bheston: With this being an underlying abstraction I'm less concerned about circumventing the design system, but the reason it is here specifically is for cases where we needed a constant value or the value was already resolved, so avoiding doing it twice. For the constant, sometimes we want "0". I suppose we could write `() => 0`.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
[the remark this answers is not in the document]
nicholasrice: In the interest of bundle size I think you should remove - it doesn't provide new utility but adds code.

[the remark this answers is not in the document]
nicholasrice: In the interest of bundle size I think you should remove - it doesn't provide new utility but adds code.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
nicholasrice: Are there cases where this information *isn't* quantified in the DesignSystem?

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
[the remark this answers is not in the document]
nicholasrice: Gotcha - makes sense. Another option would be to change *that* API to pass the resolver around - not really sure which is better. I'm good with this.

packages/fast-components-styles-msft/src/design-system/index.ts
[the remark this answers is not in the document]
nicholasrice: Yea let's hard-code for now. When we figure the story out we can update holistically.

packages/fast-components-styles-msft/src/utilities/color/recipe.ts
bheston: At some point more of the attributes might be in the design system, like the min-contrast, but for now they were hard-coded. For instance, foreground text is 14:1 and the rest offset is 0. We only have attributes for hover and active deltas.

packages/fast-components-styles-msft/src/utilities/color/accessible-recipe.ts
```diff
@@ -0,0 +1,95 @@
...
+    minContrast: number | DesignSystemResolver<number>,
+    restDelta: number | DesignSystemResolver<number>,
+    hoverDelta: number | DesignSystemResolver<number>,
+    activeDelta: number | DesignSystemResolver<number>
+): DesignSystemResolver<SwatchFamily> {
+    return (designSystem: DesignSystem): SwatchFamily => {
+        const resolvedPalette: Palette = checkDesignSystemResolver(palette, designSystem);
+        const direction: 1 | -1 = isDarkMode(designSystem) ? -1 : 1;
+
+        const accessibleSwatch: Swatch = swatchByContrast(
```
nicholasrice: This should be defined outside of the returned callback to prevent a new function from being created with the same arguments every invocation.

packages/fast-components-styles-msft/src/utilities/color/palette.ts
```diff
@@ -235,12 +243,14 @@ export function swatchByContrast(referenceColor: string | SwatchResolver) {
                      * we return the color
                      */
                     return (designSystem: DesignSystem): Swatch => {
-                        const color: Swatch =
-                            typeof referenceColor === "function"
-                                ? referenceColor(designSystem)
-                                : referenceColor;
-                        const sourcePalette: Palette = paletteResolver(designSystem);
```
nicholasrice: You're deleting length but still referencing it. I don't think you want to do that.

bheston: Doh! I thought I caught that merge.

packages/fast-components-styles-msft/src/utilities/color/accessible-recipe.ts
bheston: Per our chat, this should only be called once due to memoization.

Issue #2208, opened 2019-08-30
Contexts used in components seem to be an anti-pattern


janechu:
Some components use context as a means of passing information down to related components that are expected to be there.

According to React this is an anti-pattern: https://reactjs.org/docs/context.html

This would require a breaking change to be made in the following components:
- AutoSuggest
- Listbox
- ListboxItem
- Slider
- SliderTrackItem

There are alternatives presented in the React context documentation that would follow more closely with the principles React has established for passing information down to components.

Once it is determined which (if any) of these components should be updated to reflect the accepted React patterns, new tasks should be created for each component with a referral to this issue.

chrisdholt:
@scomea can you investigate based on Jane's recommendation and consider what changes need to take place in the above components to better align with react patterns?

scomea:
Where is the anti-pattern argument?  We are using context as intended as far as I can tell.

janechu:
From the linked documentation provided by React:

> Context is designed to share data that can be considered “global” for a tree of React components, such as the current authenticated user, theme, or preferred language.

> Context is primarily used when some data needs to be accessible by many components at different nesting levels. Apply it sparingly because it makes component reuse more difficult.

> [M]oving more complexity higher in the tree makes those higher-level components more complicated and forces the lower-level components to be more flexible than you may want.

As far as I can tell, some of the components I listed pass logic or properties down that are specifically supposed to be used by a child related to that component. There are some different patterns listed in the document such as using readerProps and the React version of slots which are named children.

scomea:
Context is used to expose data about the parent to all of it's children (ie. global for that tree) or to expose global functions.  I want to double check to ensure how we're using it does not trigger extra renders (something I read made me wonder) but I don't see the problem otherwise.

janechu:
So, I would say what we're really talking about is a scoping issue.

It's the same reason I suggested that the name in context changed to "@microsoft/fast-components-react-msft/*", because of potential conflicts.

If there are better patterns available, such as renderProps, we should use those instead of injecting component specific logic into the context.

The examples given by React illustrate this point (as to what should be considered global):
> Common examples where using context might be simpler than the alternatives include managing the current locale, theme, or a data cache

I don't know that this is super pressing, but I do believe we should tackle it for our next upcoming major version.

chrisdholt:
closed by #3517

PR #2213, opened 2019-09-03
fix: Enable slider touch dragging

scomea:
# Description
Added support for touch dragging slider thumbs.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2191

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -894,6 +908,99 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
+    };
+
+    /**
+     *  Returns first valid touch found in a touch event
+     */
+    private getValidTouch = (event: TouchEvent): Touch => {
+        if (isNil(this.rootElement.current)) {
+            return null;
+        }
+        for (let i: number = 0; i < event.touches.length; i++) {
```
janechu: It's best practice to always store the length or it will use the method to get it every loop iteration which is a slight perf problem, you can do it like this `for (let i: number = 0, touchLength: number = event.touches.length; i < touchLength; i++) `.

Additionally, if you aren't using the index for anything you could use [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) instead, it reads a little better.

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -894,6 +908,99 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
+        this.updateSliderDimensions();
+        const pixelCoordinate: number =
+            this.props.orientation === SliderOrientation.vertical
+                ? touch.clientY
+                : touch.clientX;
+        const dragValue: number =
+            (this.props.range.maxValue - this.props.range.minValue) *
+                this.convertPixelToPercent(pixelCoordinate) +
+            this.props.range.minValue;
+        return dragValue;
```
janechu: Seems like this method copies a lot of logic from `handleMouseMove`, can we consolidate it a bit and get the number of added lines down?

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -894,6 +908,99 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
+    private handleThumbTouchStart = (e: React.TouchEvent, thumb: SliderThumb): void => {
+        if (e.defaultPrevented || this.isBusyOrDisabled()) {
+            return;
+        }
+        
+        e.preventDefault();
+        (e.target as HTMLElement).focus();
+        window.addEventListener("touchend", this.handleTouchEnd);
+        window.addEventListener("touchcancel", this.handleTouchEnd);
+        window.addEventListener("touchmove", this.handleTouchMove);
```
janechu: Just a note: I think we should consider re-evaluating these window events along with #2208 as a breaking change, it seems like we could use forwardRef and localize this event listening to the component.

Not for this PR but as a potential future update.

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -894,6 +908,99 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
     };
 
+    /**
+     * Handles touch dragging
+     */
+    private handleUpperThumbTouchStart = (e: React.TouchEvent): void => {
+        this.handleThumbTouchStart(e, SliderThumb.upperThumb);
+    };
+
+    private handleLowerThumbMouseTouchStart = (e: React.TouchEvent): void => {
```
janechu: Should this have "Mouse" in the touch handler name?

packages/fast-components-react-base/src/slider/slider.tsx
scomea: optimizations could be had

packages/fast-components-react-base/src/slider/slider.tsx
scomea: yep.  Event.touches didn't work right off with the for...of syntax so didn't spend a lot of time on it as the iterator being a function call seemed odd (event.touches.item(index)).

packages/fast-components-react-base/src/slider/slider.tsx
scomea: yep

scomea: (actually no it shouldn't, but "yep" I handled it...)

packages/fast-components-react-base/src/slider/slider.tsx
scomea: Yes, consolidated into one fn that takes either a touch or mouseevent as a param

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -295,6 +299,7 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
                 slider,
                 [slider__disabled, this.props.disabled],
                 [slider__dragging, this.state.isDragging],
+                [slider__touchDragging, this.state.isTouchDragging],
```
janechu: Actually, should there be another class and callback for this? Because I would actually assume touch dragging is the same as mouse dragging, both visually and for the information passed as a callback. This way you can just check to see if there is dragging.

scomea: Developers may want to have different visuals or transitions when in touch dragging vs mouse dragging, so I think there are valid use cases for keeping track of both separately.

nicholasrice: are both `slider__dragging` and `slider__touchDragging` applied when a user is touch dragging, or are dragging and touchDragging mutually exclusive?

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -894,6 +908,86 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
+    };
+
+    private handleThumbTouchStart = (e: React.TouchEvent, thumb: SliderThumb): void => {
+        if (e.defaultPrevented || this.isBusyOrDisabled()) {
+            return;
+        }
+
+        e.preventDefault();
+        (e.target as HTMLElement).focus();
+        window.addEventListener("touchend", this.handleTouchEnd);
```
nicholasrice: You should be sure to remove these listeners on unmount to ensure they aren't called and causing an unmounted component to setState.

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -894,6 +908,86 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
+            activeThumb: thumb,
+        });
+        const thisTouch: Touch = e.nativeEvent.touches.item(0);
+        this.updateDragValue(this.getDragValue(thisTouch, thumb), thumb);
+    };
+
+    /**
+     *  Returns first valid touch found in a touch event
+     */
+    private getValidTouch = (event: TouchEvent): Touch => {
```
nicholasrice: this method *can* return `null` - the type signature should reflect that.

scomea: Really?  I guess my c# single return type background has me expecting that any function that returns an object could return null.  But sure.  I've probably done this a bunch...

packages/fast-components-react-base/src/slider/slider.tsx
scomea: We call "suspendActiveOperations" on unmount which shuts down all dragging/incrementing listeners.

packages/fast-components-react-base/src/slider/slider.tsx
scomea: made changes so we always apply the dragging class while dragging for mouse or touch, then apply a touch dragging class if we are doing that.  renamed internal state props to be more clear as they are mutually exclusive.

packages/fast-components-react-base/src/slider/slider.tsx
nicholasrice: awesome - must have missed that. Thanks

packages/fast-components-react-base/src/slider/slider.tsx
nicholasrice: Looks great - thanks

packages/fast-components-react-base/src/slider/slider.tsx
nicholasrice: Yea it's not illegal in TypeScript either (which is why this doesn't cause an error) but TypeScript does expose a `strictNullChecks` flag on the compiler to avoid that behavior. I'd like to get us there so our type system is more sound but there are quite a few issues to fix

PR #2228, opened 2019-09-07
fix: consolidate drag and drop dependencies

janechu:
# Description

Consolidates drag and drop dependencies to only use react-dnd.

## Motivation & context

This reduces package overhead when using multiple drag and drop dependencies across the tooling. It also resolves an issue with the other drag and drop dependency that was creating warnings during development.

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/form/form-item.array.props.ts
```diff
@@ -58,3 +59,6 @@ export interface FormItemArrayProps extends FormItemCommon {
      */
     onUpdateActiveSection: updateActiveSection;
 }
+
+/* tslint:disable */
```
chrisdholt: Do you want to disable all-up or just the next line?

PR #2241, opened 2019-09-13
fix: improve render performance of page, grid, and column components

nicholasrice:
# Description
Using JSS to create a few simple style rules adds a significant amout of overhead when rules can simply be added to existing inline styles. This change removes the JSS stylesheet ad applies all CSS inline.

This change also introduces an optimization for browsers that support CSS grid, reducing the render time for the Grid component by roughly half in those browsers.

## Motivation & context
Performance

## Issue type checklist

## Process & policy checklist


mdtauk:
CSS Grid with a JSS polyfill fallback maybe?

nicholasrice:
@mdtauk I don't follow that - can you elaborate?

mdtauk:
> @mdtauk I don't follow that - can you elaborate?

I just meant that anything layout wise being done with JS now, that can be replaced with HTML and CSS, like Grids, Rows, Columns - should be optimised by the browser developers.

But older browser support may require that JS code to make it work.

PR #2242, opened 2019-09-14
fix: use binary search as a lookup in swatchByContrast utility

janechu:
# Description

This changes a linear search to use a binary search.

## Motivation & context

This should increase performance when doing contrast checking.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/color/palette.ts
+++ b/packages/fast-components-styles-msft/src/utilities/color/palette.ts
@@ -1,4 +1,4 @@
-import { clamp, inRange } from "lodash-es";
+import { clamp } from "lodash-es";
 import { DesignSystem, DesignSystemResolver } from "../../design-system";
 import { backgroundColor } from "../../utilities/design-system";
 import { accentPalette, neutralPalette } from "../design-system";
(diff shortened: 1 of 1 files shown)


packages/fast-components-styles-msft/src/utilities/color/palette.ts
```diff
@@ -251,24 +251,74 @@ export function swatchByContrast(referenceColor: string | SwatchResolver) {
...
+                            constrainedSourcePalette = sourcePalette.slice(
+                                initialSearchIndex,
+                                length
+                            );
+                        } else {
+                            constrainedSourcePalette = sourcePalette.slice(
+                                0,
+                                initialSearchIndex + 1
+                            );
+                            // reverse the palette array when the direction that
```
nicholasrice: Do you need to slice the palette now that you're simply tracking an index?

janechu: I'll try it without, I was attempting to do the binary search on the assumption that 0 is always the searching index and it always moves in a positive direction (aka darker).

PR #2263, opened 2019-09-23
feat: add viewport positioner to select menu

scomea:
# Description
This change gives developers the ability to have finer control over where the select component's flyout menu is positioned.  The select component implements the viewport positioner as a wrapper around the menu and the component's props include a new flyout menu config prop that closely mirrors viewport positioner props.  

## Motivation & context
Enables select to be more easily adaptable to various layout scenarios.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/select/select.props.ts
```diff
@@ -95,6 +173,11 @@ export interface SelectHandledProps extends SelectManagedClasses {
      * element that provides it an accessible name
      */
     labelledBy?: string;
+
+    /**
+     * The onValueChange event handler
```
bheston: Wrong comment.

packages/fast-components-react-base/src/select/select.spec.tsx
```diff
@@ -22,6 +26,22 @@ const itemADisabled: JSX.Element = (
 const itemB: JSX.Element = <ListboxItem id="b" value="b" displayString="ab" />;
 const itemC: JSX.Element = <ListboxItem id="c" value="c" displayString="abc" />;
 
+const managedClasses: SelectClassNameContract = {
+    select: "select",
+    select__disabled: "select__disabled",
+    select_menu: "elect_menu",
```
bheston: Missing "s"

packages/fast-components-react-msft/src/select/select.schema.ts
```diff
@@ -3,8 +3,8 @@ export default {
     title: "Select",
     description: "A select component's schema definition.",
     type: "object",
-    id: "@microsoft/fast-components-react-msft/select",
-    formPluginId: "@microsoft/fast-components-react-msft/select",
+    id: "@microsoft/fast-components-react-base/select",
```
bheston: Making sure this is correct. I'm not quite sure how these work.

janechu: we are in the @microsoft/fast-components-react-msft package so we shouldn't be changing these to base.

packages/fast-components-react-msft/src/select/select.schema.ts
```diff
@@ -50,13 +50,68 @@ export default {
...
+            },
+        },
     },
     reactProperties: {
         children: {
             title: "Building blocks",
             type: "children",
-            formPluginId: "@microsoft/fast-components-react-msft/select/children",
-            ids: ["@microsoft/fast-components-react-msft/select-option"],
+            formPluginId: "@microsoft/fast-components-react-base/select/children",
```
janechu: also here, we should not be changing these to base, or removing the ids for available children

packages/fast-components-react-base/src/select/select.schema.ts
```diff
@@ -50,6 +50,61 @@ export default {
             title: "Labelled by",
             type: "string",
         },
+        menuFlyoutConfig: {
```
janechu: actually what you could do here is add this object as a separate export and just merge it in here, then you wont be in danger of this falling out of sync, since this looks like a copy from the viewport positioner schema

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -241,30 +251,74 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
...
-                    listbox: get(this.props.managedClasses, "select_menu", ""),
-                    listbox__disabled: get(
-                        this.props.managedClasses,
-                        "select_menuDisabled",
-                        ""
-                    ),
-                }}
+            <ViewportPositioner
+                anchor={this.rootElement}
+                {...this.getFlyoutMenuConfig()}
```
janechu: i'd suggest re-arranging this so that anchor is below the ...this.getFlyoutMenuConfig, theres a slight chance someone could pass in "anchor" as part of the config if they arent using typescript

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -241,30 +251,74 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
...
+            viewportPositioner__right: select_viewportPositioner__right,
+            viewportPositioner__top: select_viewportPositioner__top,
+            viewportPositioner__bottom: select_viewportPositioner__bottom,
+            viewportPositioner__horizontalInset: select_viewportPositioner__horizontalInset,
+            viewportPositioner__verticalInset: select_viewportPositioner__verticalInset,
+        };
+    }
+
+    /**
+     *
```
janechu: empty comment block?

packages/fast-components-react-base/src/select/select.schema.ts
scomea: Flyout props now extends viewport positioner props (omit "anchor")

packages/fast-components-react-base/src/select/select.tsx
scomea: yep

packages/fast-components-react-base/src/select/select.tsx
scomea: yep, fixed

packages/fast-components-react-base/src/select/select.schema.ts
janechu: Sure, I meant though that you can do that here in the schema as well. These are all `.ts` files, therefore you can import them just as you would any other javascript object. Then your schema could look like:

```
import FlyoutSchema from "./path/to/schema.ts";

{
    // select schema props
    menuFlyoutConfig: FlyoutSchema
}
```

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -421,7 +467,7 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         }
         const isItemSelected: boolean = state.selectedItemIndex !== 0;
         return (
-            <button
+            <Button
```
chrisdholt: Is this related to your change or a different bug? If it's not directly related can we pull it out and tackle it elsewhere?

packages/fast-components-react-base/src/select/select.schema.ts
scomea: ok

packages/fast-components-class-name-contracts-base/src/select/index.ts
```diff
@@ -26,4 +26,39 @@ export interface SelectClassNameContract {
...
+
+    /**
+     * horizontal flip inward state
+     */
+    select_viewportPositioner__horizontalInset?: string;
+
+    /**
+     * vertical flip inward state
+     */
+    select_viewportPositioner__verticalInset?: string;
```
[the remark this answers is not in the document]
scomea: These denote the position of the menu wrapper relative to the anchor.  You'd use these if you wanted to change the look/layout of the flyout based on that.  Names like "select__menuPositionTop" would probably work. But what to call the positioner root class ?  "select_menuWrapper"?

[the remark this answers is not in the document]
scomea: Actually...we could leave the wrapper itself out and make it not styleable from here.

packages/fast-components-react-base/src/select/select.schema.ts
scomea: Doesn't like it when I bootstrap that.
ReferenceError: FlyoutSchema is not defined
    at Object.<anonymous> (D:\repos\fast-dna\packages\fast-components-react-base\src\select\.tmp\select.schema.ts.js:53:27)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Module.require (internal/modules/cjs/loader.js:681:19)
    at require (internal/modules/cjs/helpers.js:16:16)
    at D:\repos\fast-dna\build\convert-schemas.js:40:32
    at Array.forEach (<anonymous>)
    at copySchemaFiles (D:\repos\fast-dna\build\convert-schemas.js:21:47)

packages/fast-components-react-base/src/select/select.tsx

packages/fast-components-class-name-contracts-base/src/select/index.ts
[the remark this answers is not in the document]
scomea: pushed up changes

packages/fast-components-react-base/src/select/select.schema.ts
janechu: Oh, you want to leave off the ".ts" from the path probably, looks like its trying to add ".js"

scomea: We discovered an issue that needs to be fixed before we can do this:
https://github.com/microsoft/fast-dna/issues/2302

packages/fast-components-class-name-contracts-base/src/select/index.ts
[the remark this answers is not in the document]
chrisdholt: @scomea does this reflect your changes? This still appears to be the same from what I can tell...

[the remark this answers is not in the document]
scomea: This commit - https://github.com/microsoft/fast-dna/pull/2263/commits/83ca4d82e1fb7d0a6aaf49551d531c3ab5f9faeb

PR #2272, opened 2019-09-25
fix: correct math error in viewport positioner

scomea:
# Description
Found a math error in viewport positioner code while looking at adding functionality. 

## Motivation & context
More accurate internal math. 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
@@ -501,9 +501,8 @@ class ViewportPositioner extends Foundation<
     private getHorizontalPositionAvailableWidth = (
         positionOption: ViewportPositionerHorizontalPositionLabel
     ): number => {
-        const spaceLeft: number = this.anchorLeft - this.viewportRect.left;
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -523,9 +522,9 @@ class ViewportPositioner extends Foundation<
     private getVerticalPositionAvailableHeight = (
         positionOption: ViewportPositionerVerticalPositionLabel
     ): number => {
-        const spaceAbove: number = this.anchorTop - this.viewportRect.top;
+        const spaceAbove: number = this.anchorTop;
         const spaceBelow: number =
-            this.viewportRect.bottom - (this.anchorTop + this.anchorHeight);
+            this.viewportRect.height - spaceAbove - this.anchorHeight;
```
janechu: I'd like it if we could extrapolate this logic in a way that we can test it

scomea: I'll add some function tests

Issue #2266, opened 2019-09-24
Add ability to wrap menu in base select component


bheston:
**What feature are you requesting?**
Ability to wrap or modify the menu portion of the base select component.

**What would be the benefit of adding this feature?**
This is currently possible for the trigger portion. In the styled component, the design system context cannot be set appropriately, because there is no way to define that the menu portion should start a different context based on a different background color recipe.

**What solution would be ideal for you?**
Please see #2211 and the comments thread including a proposed solution that seems like it would work.

**What alternatives have you considered?**
Alternatively we could redefine the entire menu in the styled component. That runs the risk of getting out of sync and having duplication.

connorjsmith:
Hm unable to assign myself (I guess I'm missing some permissions), but I'm actively working on this

scomea:
fixed by connonjsmith's merge

Issue #2250, opened 2019-09-18
ContextMenu and Select menus should have a different background level


connorjsmith:
**Describe the bug; what happened?**
ContextMenus and Select menus appear "on top" of the background, so they should select a lighter background color (especially in dark themes). This makes it hard to distinguish a menu from the background when shadows are less apparent

**If applicable, provide screenshots:**

![image](https://user-images.githubusercontent.com/5892363/65185402-56b2ab00-da1c-11e9-8669-db1c7894592f.png)

We should probably be using a lighter grey, as seen in this native context menu
![image](https://user-images.githubusercontent.com/5892363/65185423-5f0ae600-da1c-11e9-8851-6a6abb0e8966.png)

chrisdholt:
@bheston I think you are tracking this.

connorjsmith:
Note that I could wrap the entire component in a BackgroundProvider, but this would affect the trigger of a Select and could lead to contrast issues

bheston:
Yeah, found it. #2211.

bheston:
Bumping this back up. Select has been fixed, but we need the same support in Context Menu. The background for the menu should also be neutralLayerFloating, which from what I can tell should produce the correct color for any of the supported design system configurations, using background color or luminance.

chrisdholt:
@bheston are there any negative implications with updating/adding this *today*?

bheston:
I don't believe so. Currently the background color is based on light / dark mode, which flips near the middle of the ramp. I looked at these possible configurations:
- current defaults (luminance: -1 / off, background: white, no layer recipes): currently it will render with a white background, and updating to the floating recipe will also render white.
- menu within background: if someone was already styling it according to guidance, say an L2 background for a content area with a Floating background containing the menu, L2 would render as expected, floating would render as white, and floating again applied within the menu would also render as white.
- The above should work the same in dark mode.

The only potential color swap I can think of is if the app is in the middle area of the ramp, it's possible the content background could be middle grey, but the backwards compatible portion of this style (no luminance setting) could make the floating layer white. This seems highly unlikely though.

PR #2274, opened 2019-09-26
feat: support custom menu render functions in Select

connorjsmith:
# Description
This change allows for custom menu render functions in the `Select` component. This mirrors the existing custom `trigger` render function allows for a user to override the default trigger.

## Motivation & context
Allow for more flexibility in the menu rendered by a `Select`, including wrapping the default menu in a background provider as described in #2250  and #2266

## Issue type checklist

## Process & policy checklist


bheston:
Looks good to me. Feel like fixing the msft style control, or in another PR?

connorjsmith:
Forgot the msft component isn't a straight manageJss call. Will update that in a sec

packages/fast-components-react-msft/src/select/select.tsx
```diff
@@ -26,6 +27,7 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, {}> {
                 {...this.unhandledProps()}
                 managedClasses={this.props.managedClasses}
                 disabled={this.props.disabled}
+                menu={this.props.menu}
```
bheston: This doesn't quite resolve 2250, but it lets you fix it in your code. We should address it as the default stying though. Can you wrap the default menu in a Background and add neutralLayerFloating for its color?

janechu:
@connorjsmith looks like you need to run `lerna run prettier` in the  to do our auto-formatting, for some reason our npm hooks which should automatically add that on commit aren't quite working

PR #2290, opened 2019-10-02
feat: add dynamic scaling capability to viewport positioner

scomea:
# Description
This change adds a new "scaleToAvailableSpace" prop to viewport positioner.  Default is false which is the current behavior of a fixed size positioner. When it is set to true the positioner will dynamically scale to fill all the available space in the assigned region.  This is useful for scaling menus and the such.

## Motivation & context
Provide a means of scaling positioned content to available space.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/select/select.schema.ts
+++ b/packages/fast-components-react-base/src/select/select.schema.ts
@@ -103,6 +103,10 @@ export default {
                     title: "Fixed after initial placement",
                     type: "boolean",
                 },
+                scaleToFit:{
+                    title: "Scale to fit",
+                    type: "boolean",
+                }
             },
         },
     },
--- a/packages/fast-components-react-base/src/select/select.schema.ts
+++ b/packages/fast-components-react-base/src/select/select.schema.ts
@@ -103,10 +103,11 @@ export default {
                     title: "Fixed after initial placement",
                     type: "boolean",
                 },
-                scaleToFit:{
+                scaleToFit: {
                     title: "Scale to fit",
                     type: "boolean",
-                }
+                    default: "false",
+                },
             },
         },
     },
--- a/packages/fast-components-react-base/src/select/select.schema.ts
+++ b/packages/fast-components-react-base/src/select/select.schema.ts
@@ -106,7 +106,7 @@ export default {
                 scaleToFit: {
                     title: "Scale to fit",
                     type: "boolean",
-                    default: "false",
+                    default: false,
                 },
             },
         },
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
@@ -56,6 +56,7 @@ export default {
         scaleToFit: {
             title: "Scale to fit",
             type: "boolean",
+            default: "false",
         },
     },
     reactProperties: {
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
@@ -53,6 +53,10 @@ export default {
             title: "Fixed after initial placement",
             type: "boolean",
         },
+        scaleToFit: {
+            title: "Scale to fit",
+            type: "boolean",
+        },
     },
     reactProperties: {
         children: {
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
@@ -56,7 +56,7 @@ export default {
         scaleToFit: {
             title: "Scale to fit",
             type: "boolean",
-            default: "false",
+            default: false,
         },
     },
     reactProperties: {
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
@@ -1345,4 +1345,191 @@ describe("viewport positioner", (): void => {
         expect(positioner.instance().baseHorizontalOffset).toBe(-10);
         expect(positioner.instance().baseVerticalOffset).toBe(-10);
     });
+
+    test("Positioner scaled sizes calculated correctly", (): void => {
+        const anchorElement: React.RefObject<HTMLDivElement> = React.createRef<
+            HTMLDivElement
+        >();
+
+        const rendered: any = mount(
+            <div
+                style={{
+                    height: "100px",
+                    width: "100px",
+                }}
+            >
+                <div
+                    style={{
+                        height: "10px",
+                        width: "10px",
+                    }}
+                    ref={anchorElement}
+                />
+                <ViewportPositioner
+                    horizontalPositioningMode={AxisPositioningMode.inset}
+                    verticalPositioningMode={AxisPositioningMode.inset}
+                    anchor={anchorElement}
+                    viewport={document.firstElementChild as HTMLElement}
+                    managedClasses={managedClasses}
+                    scaleToAvailableSpace={true}
+                />
+            </div>
+        );
+
+        const positioner: any = rendered.find("BaseViewportPositioner");
+
+        positioner.instance().viewportRect = viewportRect;
+        positioner.instance().anchorTop = 100;
+        positioner.instance().anchorRight = 160;
+        positioner.instance().anchorBottom = 110;
+        positioner.instance().anchorLeft = 150;
+        positioner.instance().anchorWidth = 10;
+        positioner.instance().anchorHeight = 10;
+
+        expect(
+            positioner
+                .instance()
+                ["getNextPositionerWidth"](
+                    ViewportPositionerHorizontalPositionLabel.insetRight
+                )
+        ).toBe(100);
+        expect(
+            positioner
+                .instance()
+                ["getNextPositionerWidth"](ViewportPositionerHorizontalPosition.right)
+        ).toBe(90);
+        expect(
+            positioner
+                .instance()
+                ["getNextPositionerHeight"](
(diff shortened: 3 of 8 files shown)


packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -134,6 +142,7 @@ class ViewportPositioner extends Foundation<
         verticalAlwaysInView: void 0,
         verticalLockToDefault: void 0,
         fixedAfterInitialPlacement: void 0,
+        scaleToAvailableSpace: void 0,
```
janechu: I'd recommend `scaleToFit`, or `autoResize` as more common naming for this action.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
```diff
@@ -1271,4 +1309,191 @@ describe("viewport positioner", (): void => {
         expect(positioner.instance().baseHorizontalOffset).toBe(-10);
         expect(positioner.instance().baseVerticalOffset).toBe(-10);
     });
+
+    test("Positioner scaled sizes calculated correctly", (): void => {
```
janechu: I think it would be preferable to say *what* was calculated correctly.
Example "should scale the height to match the viewports height when scaleToAvailableSpace is set to true"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: "ScaleToFit" I like

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -303,6 +318,12 @@ class ViewportPositioner extends Foundation<
             (isNil(this.positionerRect) && !this.state.noObserverMode);
 
         return {
+            height: this.props.scaleToFit
+                ? `${this.state.verticalSelectedPositionHeight}px`
+                : null,
+            width: this.props.scaleToFit
+                ? `${this.state.horizontalSelectedPositionWidth}px`
+                : null,
```
janechu: nit: in order to not do this same query twice you could change this to:
```typescript
return {
    ...(
        this.props.scaleToFit
            ? {
                height: `${this.state.verticalSelectedPositionHeight}px`,
                width: `${this.state.verticalSelectedPositionHeight}px`
            }
            : {}
    ),
    // other properties
}
```

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -594,8 +615,12 @@ class ViewportPositioner extends Foundation<
             }
         });
 
-        this.scrollTop = this.getViewportElement().scrollTop;
-        this.scrollLeft = this.getViewportElement().scrollLeft;
+        const viewPortElement: HTMLElement | null = this.getViewportElement();
+
+        if (!isNil(viewPortElement)) {
+            this.scrollTop = this.getViewportElement().scrollTop;
+            this.scrollLeft = this.getViewportElement().scrollLeft;
```
janechu: should these just be `viewPortElement.scrollTop` etc since you're alerady storing the result of viewPortElement?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -879,6 +903,44 @@ class ViewportPositioner extends Foundation<
...
+            const availableHeight: number = this.getAvailableHeight(
+                desiredVerticalPosition
+            );
+            newPositionerRectHeight =
+                availableHeight > this.viewportRect.height
+                    ? this.viewportRect.height
+                    : availableHeight;
+        }
+        return newPositionerRectHeight;
+    };
```
janechu: These functions do the same thing, we can create some abstraction here so that you could combine and do something like this:

```typescript
export enum Dimension {
    height,
    width
}

// inside class

private getNextPositionDimension = (
    dimension: Dimension,
    desiredDimensionPosition: ViewportPositionerPositionLabel // would need to make this generic
): number => {
    if (this.props.scaleToFit) {
        const availableSize: number = this.getAvailableDimension( // would need to make this generic
            desiredDimensionPosition,
            dimension
        );
        return availableSize > this.viewportRect[dimension]
            ? this.viewportRect[dimension]
            : availableSize;
    }

    return this.positionRect[dimension];
}

File: packages/fast-components-react-msft/src/select/select.schema.ts [hunk 6]
```diff
@@ -103,6 +103,10 @@ export default {
                     title: "Fixed after initial placement",
                     type: "boolean",
                 },
+                scaleToFit: {
+                    title: "Scale to fit",
+                    type: "boolean",
```
janechu: `default: false`

packages/fast-components-react-base/src/select/select.schema.ts
```diff
@@ -103,6 +103,10 @@ export default {
                     title: "Fixed after initial placement",
                     type: "boolean",
                 },
+                scaleToFit:{
+                    title: "Scale to fit",
+                    type: "boolean",
```
janechu: `default: false`

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yes, better

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: yep

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: I would like to do a refinement pass with the specific intention of collapsing the axis calculations into more generic functions on this whole class.  I think that should be it's own chore and PR as it could involve a lot of refactoring.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.stories.tsx
```diff
@@ -1,53 +1,193 @@
...
-            style={{
-                height: "600px",
-                width: "600px",
-                padding: "250px",
-                background: "blue",
+class TestViewport extends React.Component<TestViewportProps, {}> {
+    private rootElement: React.RefObject<HTMLDivElement> = React.createRef<
+        HTMLDivElement
+    >();
+
```
nicholasrice: You shouldn't need this constructor

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -879,6 +904,44 @@ class ViewportPositioner extends Foundation<
         ) as ViewportPositionerState);
     };
 
+    /**
+     * Get width of positioner for next render
+     */
+    private getNextPositionerWidth = (
+        desiredHorizontalPosition: ViewportPositionerHorizontalPositionLabel
+    ): number => {
+        let newPositionerRectWidth: number = this.positionerRect.width;
```
nicholasrice: You could make this less verbose by changing to the following:
```
return this.props.scaleToFit
? Math.min(this.viewportRect.width, this.getAvailableWidth(
    desiredHorizontalPosition
))
: this.positionerRect.width;
```

packages/fast-components-react-base/src/select/select.schema.ts
```diff
@@ -103,6 +103,11 @@ export default {
                     title: "Fixed after initial placement",
                     type: "boolean",
                 },
+                scaleToFit: {
+                    title: "Scale to fit",
+                    type: "boolean",
+                    default: "false",
```
janechu: This should not be in quotes, otherwise its a string value and not a boolean.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.schema.ts
```diff
@@ -53,6 +53,11 @@ export default {
             title: "Fixed after initial placement",
             type: "boolean",
         },
+        scaleToFit: {
+            title: "Scale to fit",
+            type: "boolean",
+            default: "false",
```
janechu: also here, you need to remove the quotes or it will be considered a string.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
janechu: We shouldn't really need to plan abstraction in the future if we can just do it now. I think it's worthwhile to make sure the cleanest commits and code goes in the first time so that we don't have to come back to it.

packages/fast-components-react-msft/src/select/select.schema.ts
```diff
@@ -103,6 +103,11 @@ export default {
...
+                    default: "false",
```
janechu: also here, this should be without quotes or it will not be considered a boolean, it will be a string.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.stories.tsx
scomea: Actually I do.  There is an instanciation order issue where the ref gets passed into one component before the other has initialized it.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: refactored those functions, but used min in the updated version

PR #2297, opened 2019-10-04
feat: add looping to carousel as optional prop

### LINKED ISSUE #2295: Carousel component should be able to disable looping
The carousel component should loop by default (hitting next on the last slide goes to the first slide), but should have an optional prop to disable looping.

Allows for better configuration to fit more use cases.

An optional boolean "loop" prop that defaults to true. When loop is set to false, on the first slide the previous button doesn't exist, and on the last slide the next button doesn't exist. This prevents the user from looping through the carousel.

This may be possible by providing values to nextFlipper/previousFlipper, but seems a bit cumbersome.

SethDonohue:
# Description
- Add looping functionality and optional loop prop to Carousel component

## Motivation & context
The carousel currently does not have a looping control so looping can not be turned off.

This will allow for better configuration in more use cases.

closes #2295

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-msft/src/carousel/carousel.spec.tsx
```diff
@@ -488,4 +486,44 @@ describe("carousel", (): void => {
...
+            expect(rendered.props().loop).toBe(false);
+        });
+
+        test("should not generate a previous flipper for first slide when `loop` is passed as false", () => {
+            const rendered: any = mount(<MSFTCarousel {...props} loop={false} />);
+
+            expect(rendered.state("activeId")).toBe("id01");
+            expect(rendered.exists(".carousel_next")).toBe(true);
+            expect(rendered.exists(".carousel_previous")).toBe(false);
+        });
```
nicholasrice: If the test doesn't already exist, you might want to also test that the flippers *do* exist when loop is true.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -223,6 +225,33 @@ class Carousel extends Foundation<
         return this.slides.length !== 1;
     }
 
+    /**
+     * Check if it is the FIRST slide for looping
+     */
+    private get isFirstSlide(): boolean {
+        if (this.props.items.length) {
+            const firstSlideId: string = this.props.items[0].id;
+            if (firstSlideId) {
```
nicholasrice: Do you need this `if (firstSlideId)` check? Presumably if this was falsey, `firstSlideId === this.state.activeId` would also be falsey

packages/fast-components-react-msft/src/carousel/carousel.spec.tsx
SethDonohue: loop defaults to true, and there is a flipper test that checks for default flippers showing in the, so that would you agree that those tests cover it?

chrisdholt: If the test isn't already written, I'd add it just to be safe IMO. If someone follows with a change we would want that case to be accounted for to ensure the test breaks

packages/fast-components-react-msft/src/carousel/carousel.tsx
SethDonohue: true, i am already checking to make sure the length exists. making this change passes the tests too so I'll add it.

packages/fast-components-react-msft/src/carousel/carousel.spec.tsx
SethDonohue: tests added for clarity and to guarantee coverage

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -223,6 +225,31 @@ class Carousel extends Foundation<
         return this.slides.length !== 1;
     }
 
+    /**
+     * Check if it is the FIRST slide for looping
+     */
+    private get isFirstSlide(): boolean {
+        if (this.props.items.length) {
```
janechu: It seems like slides is always an array, unlike items which can be undefined, so I think you should change this to check the length on slides.

bekell:
LGTM

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -223,6 +225,31 @@ class Carousel extends Foundation<
...
+        } else {
+            return false;
+        }
+    }
+
+    /**
+     * Check if it is the LAST slide for looping
+     */
+    private get isLastSlide(): boolean {
+        const lastItemKey: number = this.props.items.length - 1;
```
janechu: same issue here, it is possible for items to be undefined.

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -223,6 +225,31 @@ class Carousel extends Foundation<
...
+        if (lastItemKey >= 0) {
+            const lastSlideId: string = this.props.items[lastItemKey].id;
```
chrisdholt: should likely check this.slides here as well instead of `props.items`

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -295,6 +322,9 @@ class Carousel extends Foundation<
         if (!this.isMultipleSlides) {
             return;
         }
+        if (this.isFirstSlide && !this.props.loop) {
+            return;
+        }
```
chrisdholt: Can these be combined?
```
if (!this.isMultipleSlides || this.isFirstSlide && !this.props.loop) {
    return;
}
```

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -322,6 +352,9 @@ class Carousel extends Foundation<
         if (!this.isMultipleSlides) {
             return;
         }
+        if (this.isLastSlide && !this.props.loop) {
+            return;
+        }
```
chrisdholt: Can these be combined?
```
if (!this.isMultipleSlides || this.isLastSlide && !this.props.loop) {
    return;
}
```

packages/fast-components-react-msft/src/carousel/carousel.tsx
```diff
@@ -292,7 +319,7 @@ class Carousel extends Foundation<
      * Generates previous flipper if more than one slide
      */
     private generatePreviousFlipper(): any {
-        if (!this.isMultipleSlides) {
+        if (!this.isMultipleSlides || (this.isFirstSlide && !this.props.loop)) {
```
janechu: I think you should do:
`(!this.isMultipleSlides || (this.isMultipleSlides && this.isFirstSlide && !this.props.loop))`

Because you probably only want to render this if its the first slide, not looping AND there are multiple slides.

janechu: Actually, I've just realized that's redundant, nevermind.

PR #2298, opened 2019-10-07
feat: allow control plugins to be passed to the form

### LINKED ISSUE #551: Add ability to pass replacement components to override the default inputs
A developer implementing the form generator should be able to replace the default simple form inputs for styling and as an additional hook.

Sufficient documentation to show how this will work should be provided in the README.

These inputs include:
- checkbox
- file upload
- number field
- select
- text field
- textarea

To add to this, the custom mapping for components will be removed and these components will be provided as additional exports.

Related to #1423 

To use custom form controls should be a similar implementation as the other plugin systems.

```jsx
<Form
    data={this.state.currentComponentData}
    schema={this.state.currentComponentSchema}
    onChange={handleChange}
    formControls={[
        new SliderFormControl({
            id: ["myJSONSchemaSliderFormControlId"]
        })
    ]}
/>
```
JSON Schema:
```json
{
    "type": "object",
    "properties": {
        "foo": {
            "title": "Foo",
            "description": "Foo slider",
            "type": "number",
            "minimum": 0,
            "maximum": 100,
            "formControlId": "myJSONSchemaSliderFormControlId"
        }
    }
}
```

janechu:
# Description

Adds a plugin for use in creating custom controls for the form, allows the form to accept custom controls.

## Motivation & context

This allows the passing of custom controls to the Form to use in place of the default form controls.

Closes #551 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/form/custom-controls/index.ts
+++ b/packages/fast-tooling-react/src/form/custom-controls/index.ts
@@ -10,10 +10,10 @@ import { ThemeControlProps } from "./control.theme.props";
 export {
     AlignControl,
     AlignControlProps,
-    FileUploadControl,
-    FileUploadControlProps,
+    // FileUploadControl, // TODO: this needs a visual pass
+    // FileUploadControlProps,
     TextAlignControl,
     TextAlignControlProps,
-    ThemeControl,
-    ThemeControlProps,
+    // ThemeControl, // TODO: this needs a visual pass
+    // ThemeControlProps,
 };
--- a/packages/fast-tooling-react/src/form/form-control-switch.tsx
+++ b/packages/fast-tooling-react/src/form/form-control-switch.tsx
@@ -93,9 +93,7 @@ class FormControlSwitch extends React.Component<FormControlSwitchProps, {}> {
     }
 
     private renderChildren(control: StandardControlPlugin): React.ReactNode {
-        const standardControl: StandardControlPlugin = control;
-
-        standardControl.updateProps(
+        control.updateProps(
             Object.assign(
                 {
                     childOptions: this.getChildOptions(),
@@ -105,16 +103,14 @@ class FormControlSwitch extends React.Component<FormControlSwitchProps, {}> {
             )
         );
 
-        return standardControl.render();
+        return control.render();
     }
 
     /**
      * Renders the array form item
      */
     private renderArray(control: StandardControlPlugin): React.ReactNode {
-        const standardControl: StandardControlPlugin = control;
-
-        standardControl.updateProps(
+        control.updateProps(
             Object.assign(
                 {
                     minItems: get(this.props.schema, ItemConstraints.minItems, 0),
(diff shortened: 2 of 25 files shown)


packages/fast-tooling-react/src/form/custom-controls/index.ts
```diff
@@ -10,10 +10,10 @@ import { ThemeControlProps } from "./control.theme.props";
 export {
     AlignControl,
     AlignControlProps,
```
nicholasrice: Still TODO?

packages/fast-tooling-react/src/form/form-control-switch.tsx
```diff
@@ -76,8 +111,8 @@ class FormControlSwitch extends React.Component<FormControlSwitchProps, {}> {
     /**
      * Renders the array form item
      */
-    private renderArray(): React.ReactNode {
-        const standardControl: StandardControlPlugin = this.props.controls.array;
+    private renderArray(control: StandardControlPlugin): React.ReactNode {
+        const standardControl: StandardControlPlugin = control;
```
nicholasrice: Whats the purpose of re-assigning this argument?

packages/fast-tooling-react/src/form/form.tsx
```diff
@@ -172,54 +175,134 @@ class Form extends React.Component<
...
+                    },
+                });
+        }
+    }
+
+    private updateControls(): void {
+        const hasCustomControlPlugins: boolean = Array.isArray(this.props.controlPlugins);
+
+        this.selectControl = this.findControlPlugin(
+            hasCustomControlPlugins,
```
nicholasrice: If the logic for determining if there are custom control plugins is always the same, is should `this.findControlPlugin` just determine this itself? Then you don't have to pass it as an argument for every invocation.

packages/fast-tooling-react/src/form/templates/plugin.control.utilities.tsx
```diff
@@ -27,8 +54,25 @@ export default abstract class ControlPluginUtilities<
         this.updateConfig(config);
     }
 
+    /**
+     * Determines if there is a match for the IDs set for the plugin
+     * and a provided ID
+     */
+    public matchesId(id: string): boolean {
+        return this.config.id.indexOf(id) !== -1;
```
nicholasrice: why `indexOf` and not direct comparison? couldn't you get unexpected id matches?

packages/fast-tooling-react/src/form/custom-controls/index.ts
janechu: Yes, I didn't vet those custom controls in this PR.

packages/fast-tooling-react/src/form/form-control-switch.tsx
janechu: I forgot to remove it :)

packages/fast-tooling-react/src/form/form.tsx
janechu: Passing it as an argument means I only have to do the check once, otherwise I would do the same conditional 9 times.

packages/fast-tooling-react/src/form/templates/plugin.control.utilities.tsx
janechu: No, this is a array indexOf not string indexOf

PR #2306, opened 2019-10-09
fix: ensure minWidth value is passed as pixels in pane

chrisdholt:
# Description
In certain scenarios it's possible for pane to be passed as a number and not a pixel value when a pixel value is expected.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/pane/pane.tsx
+++ b/packages/fast-layouts-react/src/pane/pane.tsx
@@ -217,7 +217,7 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
         const styles: React.CSSProperties = {};
 
         styles.minWidth = this.props.collapsed
-            ? this.props.collapsedWidth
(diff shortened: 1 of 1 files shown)


packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -217,7 +217,7 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
         const styles: React.CSSProperties = {};
 
         styles.minWidth = this.props.collapsed
-            ? this.props.collapsedWidth
+            ? toPx(this.props.collapsedWidth)
```
nicholasrice: Doesn't react convert this for us?

chrisdholt: If it does we should remove toPx from the minWidth value below. I found this writing tests where I'm getting a number back as opposed to a string. I'm fine either way; just think we should be consistent with our application here.

PR #2307, opened 2019-10-09
feat: add validation errors to a section

### LINKED ISSUE #1761: Surface validation in the UI when an object is invalid inside a section
The form currently has validation that chains up from nested locations however if there is a validation error happening at on an object there is no place to see it.

An example of this is a schema with type object and a required list with a property that is missing. This should cause an error on a section, and if the errors enabled inline it should be displayed as an alert type UI.

janechu:
# Description

This adds validation errors at the top of the form sections.

## Motivation & context

This will allow each object to have validation errors pertaining to the object, instead of only showing validation on simpler form controls.

Closes #1761 

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/form-section.tsx
```diff
@@ -393,7 +407,17 @@ class FormSection extends React.Component<
         return null;
     }
 
-    private renderFormSection(): React.ReactNode {
+    private renderFormValidation(invalidMessage: string): React.ReactNode {
+        if (invalidMessage !== "") {
+            return (
+                <div className={this.props.managedClasses.formSection_invalidMessage}>
```
eljefe223: should we be using the `classNames` utility?

janechu: there's no reason when it's just a single class name and there is no logic for determining whether to add/remove it.

PR #2310, opened 2019-10-09
fix: allow tabbing away from controlled auto-suggest without opening menu

### LINKED ISSUE #2309: Controlled mode keyboard tab navigation always triggers menu in AutoSuggest
When using the auto-suggest component in controlled mode (passing `value` prop, and updating it from `onValueChange`) with the keyboard, the suggestion menu is always triggered when tabbing out of the input text box.

Create an auto suggest in controlled mode - `value={this.state.input}` and `onValueChange={value => this.setState ...}`. Create some mock function or something to population `AutoSuggestOption` children on input.

Tab to the input, type into it, select an option using the keyboard. When you press tab next, to move onto something else, the menu will trigger.

The root cause is because [`handleInputRegionKeydown`](https://github.com/microsoft/fast-dna/blob/af5f04a7c54c208f5475ec8ee0372bb1038f4411/packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx#L309) has a `default` case which calls `this.updateValue()` and `this.focusOnInput()`. `updateValue` always calls [`this.toggleMenu(true)`](https://github.com/microsoft/fast-dna/blob/af5f04a7c54c208f5475ec8ee0372bb1038f4411/packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx#L299) when `props.value !== undefined` which triggers the menu.

Additionally, the `focusOnInput` call makes this a key accessibility bug - if elements are covered by the auto-suggest menu, these elements are always hidden under the menu since you can't tab away from the input without opening the menu.

Pressing `tab` would not trigger the default case that opens the menu and refocuses the input.

Observed in Edge Chromium on Windows, would be reproducible on all supported browsers/OS.

* Version [e.g. 1.8.0]
`@microsoft/fast-components-react-base` - 4.14.3
Code is there in latest master, as of writing.

The fix may be as simple as an empty `case (keyCode): break` for tab, depending on what the intended behaviour for tab with inactive menu is.

hchokshi:
# Description
Removes default (update input value and re-focus to input) handling for tab keypress to move out of AutoSuggest component. Allows the user to tab away from a controller auto suggest component without opening the suggestions.

## Motivation & context

Resolves #2309. Makes AutoSuggest more accessible.

## Issue type checklist

## Process & policy checklist


janechu:
Could you add a test for this in the `auto-suggest.spec.tsx`? I would suggest that the test makes sure that default is not prevented, or that focus is now on the next tab-able element and that invoke has not been triggered.

hchokshi:
I tried testing focus being passed to an adjacent element (with `onFocus` event handler, or `document.activeElement` + ref), but it seems this is not testable in Jest as simulating keypresses won't trigger default behaviour that would be expected for explicit user input for security reasons (to prevent JavaScript from pretending to be the user).

https://stackoverflow.com/a/1601732

janechu:
No worries, this looks good to me, thanks!

PR #2313, opened 2019-10-10
chore: add tests for fast-layouts components

chrisdholt:
# Description
Adds tests for components in `fast-layouts-react`

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/grid/grid.no-css-grid.spec.tsx
+++ b/packages/fast-layouts-react/src/grid/grid.no-css-grid.spec.tsx
@@ -1,0 +1,80 @@
+import React from "react";
+import Adapter from "enzyme-adapter-react-16";
+import { configure, mount, shallow } from "enzyme";
+import {
+    Grid,
+    GridAlignment,
+    GridClassNamesContract,
+    GridHandledProps,
+    GridUnhandledProps,
+} from "./grid";
+import { GridTag } from "./grid.props";
+import { Column, ColumnClassNamesContract } from "../column";
+import { GridGutter } from ".";
+import { uniqueId } from "lodash-es";
+
+/**
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("Grid - without CSS grid support", (): void => {
+    const managedClasses: GridClassNamesContract = {
+        grid: "grid",
+    };
+
+    test("should set an inline style of `display: -ms-grid` when CSS grid is NOT supported", () => {
+        const rendered: any = shallow(<Grid />);
+
+        expect(rendered.props().style.display).toEqual("-ms-grid");
+    });
+
+    test("should pass `gutter` props to children of type `column` when CSS grid is not supported and `column` does not have `gutter` prop", () => {
+        const columnId: string = "foo";
+        const gridGutter: number = 4;
+
+        const rendered: any = shallow(
+            <Grid gutter={gridGutter} managedClasses={managedClasses}>
+                <Column id={columnId} />
+            </Grid>
+        );
+
+        expect(rendered.find(`#${columnId}`).props().gutter).toBe(gridGutter);
+    });
+
+    test("should NOT pass `gutter` props to children of type `column` when CSS grid is not supported and `column` has `gutter` prop", () => {
+        const columnId: string = "bar";
+        const gridGutter: number = 4;
+        const columnGutter: GridGutter = 6;
+
+        const rendered: any = shallow(
+            <Grid gutter={gridGutter} managedClasses={managedClasses}>
+                <Column id={columnId} gutter={columnGutter} />
+            </Grid>
+        );
+
+        expect(rendered.find(`#${columnId}`).props().gutter).toBe(columnGutter);
+    });
+
+    test("should NOT pass `gutter` props to children which are not of type `column` when CSS grid is not supported", () => {
+        const gridGutter: number = 4;
+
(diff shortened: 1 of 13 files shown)


packages/fast-layouts-react/src/grid/grid.spec.tsx
```diff
@@ -0,0 +1,234 @@
...
+        expect(rendered.props().tag).not.toBe(undefined);
+        expect(rendered.props().tag).toBe(GridTag.div);
+    });
+
+    test("should set a tag value equal to the `tag` prop when passed", () => {
+        const rendered: any = mount(
+            <Grid tag={GridTag.main} managedClasses={managedClasses} />
+        );
+
+        expect(rendered.props().tag).toBe(GridTag.main);
```
janechu: it would be nice if we tested that the tag was actually used to create a `<main>` element.

packages/fast-layouts-react/src/page/page.spec.tsx
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    test("should set a margin value equal to the `margin` prop when passed", () => {
+        const rendered: any = mount(
+            <Page margin={"10%"} managedClasses={managedClasses} />
+        );
+
+        expect(rendered.props().margin).toBe("10%");
+    });
+});
```
janechu: can we check the generated styles on this, namely `display`, `gridTemplateColumns`, `msGridColumns`?

packages/fast-layouts-react/src/pane/pane.spec.tsx
```diff
@@ -0,0 +1,632 @@
...
+                    managedClasses={managedClasses}
+                />
+            );
+            const initialWidth: number = rendered.state().width;
+
+            rendered
+                .find(`.${managedClasses.pane_resizeHandle}`)
+                .simulate("keyDown", { keyCode: KeyCodes.arrowRight });
+
+            expect(rendered.state().width).toBe(initialWidth + 1);
```
janechu: should we check the style width instead?

packages/fast-layouts-react/src/pane/pane.spec.tsx
```diff
@@ -0,0 +1,632 @@
...
+        expect(rendered.props().overlay).toBe(false);
+    });
+
+    test("should set `overlay` to value of `overlay` prop when passed", () => {
+        const rendered: any = mount(
+            <Pane overlay={true} managedClasses={managedClasses} />
+        );
+
+        expect(rendered.props().overlay).toBe(true);
+    });
```
janechu: can we add another overlay test and check the styles.flexBasis has been set?

packages/fast-layouts-react/src/row/row.spec.tsx
```diff
@@ -0,0 +1,568 @@
...
+import { Row, RowClassNamesContract, RowHandledProps, RowUnhandledProps } from "./row";
+import { KeyCodes } from "@microsoft/fast-web-utilities";
+import { RowResizeDirection } from "./row.props";
+
+/**
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("Row", (): void => {
```
janechu: seems to me like there is a lot of duplication in row and pane, separate from this PR, should we file a task to abstract the logic into a class these can both extend and consolidate tests?

packages/fast-layouts-react/src/grid/grid.no-css-grid.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Column, ColumnClassNamesContract } from "../column";
+import { GridGutter } from ".";
+import { uniqueId } from "lodash-es";
+
+/**
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("Grid - without CSS grid support", (): void => {
```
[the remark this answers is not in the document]
chrisdholt: I technically can't due to issues I'm having with mocks and needing to scope them to a specific block. With babel-jest mocks get hoisted and while they state that variables which start with `mock` are also hoisted, that seems to not be the case. I'm open to changing the convention, but we should have these tests.

packages/fast-layouts-react/src/grid/grid.spec.tsx
chrisdholt: Sure. Easy enough

packages/fast-layouts-react/src/row/row.spec.tsx
chrisdholt: I think that task exists: https://github.com/microsoft/fast-dna/issues/564

packages/fast-layouts-react/src/page/page.spec.tsx
chrisdholt: Sure - display is on line 58 already; I'll add the others. `msGrid` will require a seperate file at this time similar to grid.

packages/fast-layouts-react/src/pane/pane.spec.tsx
chrisdholt: Yeah I had this and then pulled it...I'll add it back in.

packages/fast-layouts-react/src/pane/pane.spec.tsx
chrisdholt: Technically can't do the lifecycle here to get the style update to propagate due to internal refs being undefined.

PR #2314, opened 2019-10-10
fix: add the prefix forced-colors to fix high contrast in Edge chromium

khamudom:
Edge chromium canary uses the new standard prefix, which is 'forced-color' adjust property. We currently are only using `-ms-high-contrast` adjust property. We still want to keep the `-ms-high-contrast` standard to support Edge and IE. 
The color values have also been updated, for example, for background, `forced-colors` uses "Canvas". For an anchor in `-ms-high-contrast` it uses "-ms-hotlight", where `forced-colors` use "LinkText".

# Description
This change is to make sure, when in high contrast, our components will work in Edge chromium, using `forced-colors` and that `-ms-high-contrast` will also work when the browser does not have the `forced-colors` flag.
- I created a function,`applyHighContrastSelector `,  it will check if `forced-colors` flag is enabled on the browser, if so, then prefix with `forced-colors` query, if not that `-ms-high-contrast` query will be used.
- For the scenario where we have two anchor color value. I created a similar function as the media query, called, `applyHighContrastlinkValue`. This detects that, in `forced-colors` we use "LinkText", and in `-ms-high-contrast` we use "-ms-hotlight".
- Now that "LinkText" is available, I created extra const objects to use on our anchor buttons.
`highContrastLinkOutline `, `highContrastLinkBorder ` and `highContrastLinkForeground `.
- Components like `Radio`, `Checkbox`, `Toggle`, `Pivot`, `Progress`, `Carousel`, `Slider label`, `Slider` had to be re-addressed, when using `forced-colors`.  The fix is to use the opt out property, `forced-colors-adjust` and `-ms-high-contrast-adjust`. These property allowed me to opt particular elements out of `forced-colors` and `-ms-high-contrast` mode. Then I added the correct color values to elements. I created a function, `applyhighContrastOptOutProperty`, that uses both property.
- Making sure there no hardcode color values. Instead we are using the enum or the color pattern objects.
- Create extra stories that shows Anchor and Disabled on some of our button components.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/accent-button/accent-button.stories.tsx
+++ b/packages/fast-components-react-msft/src/accent-button/accent-button.stories.tsx
@@ -5,7 +5,14 @@ import { glyphFactory, SVGGlyph } from "../../assets/svg-element";
 
 storiesOf("Accent button", module)
     .add("Default", () => <AccentButton>Accent button</AccentButton>)
-    .add("Anchor", () => <AccentButton href="#">Anchor button</AccentButton>)
+    .add("Anchor", () => 
+    <AccentButton 
+        href="#"
+        beforeContent={glyphFactory(SVGGlyph.download)}
+        afterContent={glyphFactory(SVGGlyph.user)}
+    >
+        Anchor button
+    </AccentButton>)
     .add("Before content", () => (
         <AccentButton beforeContent={glyphFactory(SVGGlyph.download)}>
             With before content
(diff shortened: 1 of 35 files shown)


eljefe223:
smoked non-high contrast.

janechu:
Please remove any mention of internally used names or any mention of commentary not useful for the open source community.

khamudom:
Holding off on using `forced-colors`. Once all the color values are supported I will address adding the `forced-colors` prefix. Using `-ms-high-contrast` prefix will be fine.

PR #2316, opened 2019-10-10
fix: progress `size` class names conflicting

eljefe223:
# Description
Looks when we refactored to use `classNames` utility there was a regression in how the `size` `className` was being applied and the `control` size `className` was always being overwritten by the `container` size `className`. We are defaulting the prop `size` so just  applying a single class fixes the issue.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/progress/progress.tsx
+++ b/packages/fast-components-react-msft/src/progress/progress.tsx
@@ -92,10 +92,7 @@ class Progress extends Foundation<ProgressHandledProps, ProgressUnhandledProps, 
         const managedClasses: Partial<ProgressClassNameContract> = this.props
             .managedClasses;
 
-        return classNames(
-            managedClasses.progress_circularSVG__container,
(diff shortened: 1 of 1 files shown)


packages/fast-components-react-msft/src/progress/progress.tsx
```diff
@@ -92,10 +92,7 @@ class Progress extends Foundation<ProgressHandledProps, ProgressUnhandledProps,
         const managedClasses: Partial<ProgressClassNameContract> = this.props
             .managedClasses;
 
-        return classNames(
-            managedClasses.progress_circularSVG__container,
-            managedClasses[`progress_circularSVG__${this.props.size}`]
-        );
+        return classNames(managedClasses[`progress_circularSVG__${this.props.size}`]);
```
janechu: nit: there doesn't seem to be a need to reassign the class name contract to a `const`.

PR #2319, opened 2019-10-10
fix: modal dialogs actively take focus

scomea:
# Description
Intercept document "focusin" events and focus the dialog if the element receiving focus is outside of it.

## Motivation & context
This is a proposed fix for the problem of users being able to focus elements on the page outside of the dialog.  The common scenario being that the user clicks on the browser title bar, then tabs back to the document and focus is passed to an element outside the dialog.

It's a fairly heavy handed approach, but I'm thinking it might be warranted.  Discuss. 

Accessibility reference:
"modal dialogs contain their tab sequence. That is, Tab and Shift + Tab do not move focus outside the dialog. However, unlike most non-modal dialogs, modal dialogs do not provide means for moving keyboard focus outside the dialog window without closing the dialog. "
https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -140,6 +153,20 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
+        if (isNil(this.rootElement.current)) {
+            return;
+        }
+        if (
+            this.props.visible &&
+            !(this.rootElement.current as HTMLElement).contains(
+                event.relatedTarget as HTMLElement
+            )
+        ) {
+            this.rootElement.current.focus();
```
janechu: This would seem to suggest that if focus leaves the modal it will refocus on the modal, however that doesn't seem to support what aria specs have which is:

- Tab from last focusable element in the modal should focus on first focusable element,
- Shift + Tab on first focusable element should focus on last focusable element.

It seems like what we should do here is:
- Store first focusable item
- Store last focusable item
- Check during appropriate lifecycles in case a new focusable item has been rendered to see if there is a new first or last focusable item
- When the keydown event is Shift + Tab on the first item, focus on the last item
- When the keydown event is Tab on the last item, focus on the first item

You could probably use the current `handleWindowKeyDown` method.

scomea: I thought about that and came to the conclusion that developers can manage their own focus loops in dialog ui.  When we refocus the dialog focus goes back to the last thing focused.  Why get involved in managing that if we don't have to?

scomea: (there are existing implementations of these dialogs that already have focus loops)

chrisdholt: It seems like an issue may be useful here to actually track feedback and the problem a bit more. If the goal is to trap focus within modal dialogs, we should investigate ways it can be accomplished now (if at all) and then move towards a proposed fix/feature.

I'm also fine if we want to have the discussion here, but it may be difficult to track in this form.

Either way, I'm waiting on this PR as it's "live" and not a "draft" until we can land the overall problem, the current implementation, and then what's missing.

scomea: Happy to move discussion wherever, the intent of the all caps "PROPOSED" tag was to indicate that I consider this a decision of consequence that needs discussion (and I tend towards having an implementation to discuss over).  Created an issue here - https://github.com/microsoft/fast-dna/issues/2329

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -130,14 +149,98 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
+     * forces focus to first tabbable element of modal dialog if document gains focus while dialog is open
+     */
+    private handleDocumentFocus = (event: FocusEvent): void => {
         if (
-            this.props.onDismiss &&
-            typeof this.props.onDismiss === "function" &&
             this.props.visible &&
-            event.keyCode === keyCodeEscape
+            !event.defaultPrevented &&
+            !isNil(this.rootElement.current) &&
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -130,14 +149,98 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
-    private handleWindowKeyDown = (event: KeyboardEvent): void => {
+    /**
+     * handles document key down events
+     */
+    private handleDocumentKeyDown = (event: KeyboardEvent): void => {
+        if (!event.defaultPrevented && this.props.visible) {
+            switch (event.keyCode) {
+                case keyCodeEscape:
+                    if (
+                        this.props.onDismiss &&
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -130,14 +149,98 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
+            }
+        }
+    };
+
+    /**
+     * if no tabbable elements try to focus root element 
+     * generally a modal dialog should be expected to have a focusable element
+     */
+    private tryFocusOnRootElement = (): void => {
+        if (!isNil(this.rootElement.current)) {
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -68,8 +75,12 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
      * React life-cycle method
      */
     public componentDidMount(): void {
-        if (canUseDOM() && this.props.onDismiss) {
-            window.addEventListener("keydown", this.handleWindowKeyDown);
+        if (canUseDOM()) {
+            document.addEventListener("keydown", this.handleDocumentKeyDown);
```
chrisdholt: Do we want to add the keydown event even if we don't need to listen to it (IE, there is no onDismiss handler, so escape does not close the dialog)

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -78,10 +89,11 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
      */
     public componentDidUpdate(prevProps: Partial<DialogHandledProps>): void {
         if (canUseDOM()) {
-            if (!prevProps.onDismiss && this.props.onDismiss) {
-                window.addEventListener("keydown", this.handleWindowKeyDown);
-            } else if (prevProps.onDismiss && !this.props.onDismiss) {
-                window.removeEventListener("keydown", this.handleWindowKeyDown);
```
chrisdholt: Are we pulling this logic for a reason?

packages/fast-components-react-base/src/dialog/dialog.tsx
chrisdholt: I see - you're using this for the modal now as well.

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -1,10 +1,12 @@
 import { DialogClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
 import Foundation, { HandledProps } from "@microsoft/fast-components-foundation-react";
-import { classNames, keyCodeEscape } from "@microsoft/fast-web-utilities";
+import { classNames, keyCodeEscape, keyCodeTab } from "@microsoft/fast-web-utilities";
 import { canUseDOM } from "exenv-es6";
 import React from "react";
 import { DisplayNamePrefix } from "../utilities";
 import { DialogHandledProps, DialogProps, DialogUnhandledProps } from "./dialog.props";
+import { isNil } from "lodash-es";
+import Tabbable from "tabbable";
```
chrisdholt: Do we have any concerns about this *not* being ES6 and the size delta it could create downstream? @nicholasrice

packages/fast-components-react-base/src/dialog/dialog.tsx
scomea: Restored similar logic to not listen when we don't need to.

packages/fast-components-react-base/src/dialog/dialog.tsx
scomea: Restored similar logic to not listen when we don't need to

janechu:
If this PR is meant to be the final approach, can you remove the [PROPOSED] from the title?

packages/fast-components-react-base/src/dialog/dialog.tsx
nicholasrice: I don't have much concern about this not using ES6 modules - it is a purpose-specific dep that doesn't (at least at this point) appear to have any code that can be tree-shaken.

Bundle size is aways a concern - if we think we can do this with less code then we should.

chrisdholt: Thoughts on the above @nicholasrice? I'm open to either here, but want to be conscious of bundle size increases on folks.

Issue #2318, opened 2019-10-10
MSFT radio gets offset at different pixel densities


chrisdholt:
**Describe the bug; what happened?**
The radio component state indicator appears offset at different pixel densities and zoom levels - this means that different devices may render the issue and others may not.

**What are the steps to reproduce the issue?**
Go to the radio component and ensure it is in a "checked" state. Zoom in to 110%. Alternatively, open chromium based Edge browser on Surface Pro. On the default home page, click the settings cog and notice the radios are offset. This also repros in Edge settings instances which have radios.

**What behavior did you expect?**
The state indicator should center no matter the device density or zoom level.

**If applicable, provide screenshots:**
<img width="146" alt="GetImage" src="https://user-images.githubusercontent.com/13071055/66600687-e69acf00-eb5a-11e9-9103-4f8b2a523431.png">

* Browser [e.g. Edge, Chrome, Safari, FireFox]
* Version [e.g. 1.8.0]

**Is there any additional context?**

chrisdholt:
closed with #3517

PR #2324, opened 2019-10-11
fix: subpixel rendering issues in radio

eljefe223:
# Description
Closes: #2318 

The current layout of radio centers elements within each other this causes rendering issues at some zoom levels.

Example of 90% zoom in Chromium:
image

Solution is to render the selected indicator with a single elements style.

67%
image

75%
image

80%
image

90%
image

110%
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/radio/index.ts
+++ b/packages/fast-components-styles-msft/src/radio/index.ts
@@ -12,6 +12,8 @@ import {
     divide,
     format,
     toPx,
+    subtract,
+    multiply,
 } from "@microsoft/fast-jss-utilities";
 import {
     neutralFillInputActive,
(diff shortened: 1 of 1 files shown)


packages/fast-components-styles-msft/src/radio/index.ts
```diff
@@ -122,13 +111,40 @@ const styles: ComponentStyles<RadioClassNameContract, DesignSystem> = {
...
-        "&:hover $radio_stateIndicator::before": {
+        "&:hover $radio_stateIndicator": {
             ...highContrastSelectedBackground,
+            "box-shadow": format(
+                "inset 0 0 0 {0} {1}",
+                indicatorMargin,
+                neutralFillInputHover
+            ),
+        },
+        "&:hover:enabled $radio_stateIndicator": {
```
nicholasrice: why is this selector necessary? It applies the same box-shadow as above to the same element

bheston: Ideally we keep the ones with ":enabled" for elements that support this selector. This should be on "active" as well below.

PR #2327, opened 2019-10-14
fix: components are not showing correct high contrast colors in Edge chromium, removed hard code value, and add more examples

khamudom:
Edge chromium canary now supports the high contrast flags,  `ForcedColors`, `UsedWindowsSystemColors`, `LinkSystemColors`. 

Steps to smoking (only for Windows PC):
-  Enable flag on Edge chromium canary by using `--enable-blink-features=ForcedColors,UsedWindowsSystemColors,LinkSystemColors`
- `npm start` at `packages/fast-components-react-msft`
- Turn on high contrast in Settings, choosing high contrast black or white.
When smoking in default mode, please make sure there are no regressions in all states, rest, hover active and keyboard focus.
For high contrast, most importantly, make sure the all parts of the components are visible. This is the issue we are having in chromium. HC works on Edge, but not Edge chromium, so that was addressed in this PR. Check all states, rest, hover, active and keyboard focus.

# Description
This change is to make sure, when high contrast is enabled, our components will work in Edge chromium.
- There are two different link values, `-ms-hotlight` for Edge and `LinkText` for Edge chromium . Created, applyHighContrastlinkValue , that will detects if forced-colors query is set, we use "LinkText", and for -ms-high-contrast, we use "-ms-hotlight".
- I created extra const objects to use on our anchor buttons.
highContrastLinkOutline , highContrastLinkBorder and highContrastLinkForeground .
- Components like Radio, Checkbox, Toggle, Pivot, Progress, Carousel, Slider label, Slider had to be re-addressed. The fix is to use the opt out property, `-ms-high-contrast-adjust`. This property allows us to opt, particular elements, out of -ms-high-contrast color scheme. Then added the correct color values to elements.
- Making sure there no hardcode color values. Using the enum or the high contrast color pattern objects in its place.
- Create extra stories that shows Anchor and Disabled examples on our button components.
- When using `window.matchMedia()`, Jest will return `TypeError: window.matchMedia is not a function` because JSDOM has not implemented yet.
So I had to create a file that mocks `matchMedia`, and import the file, into the test files.
https://jestjs.io/docs/en/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

**Update:** Took a different approach to testing `matchMedia`. Instead of having to import the match-media file to every spec file, I created a mock.js file in the package, and inside package.json use `setupFilesAfterEnv` pointing to the mock file. This way it will test for match media in any spec file.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -24,6 +24,8 @@ import {
     highContrastDisabledForeground,
     highContrastForeground,
     highContrastHighlightForeground,
+    highContrastLinkForeground,
+    highContrastOptOutProperty,
     highContrastSelectedForeground,
     highContrastSelector,
 } from "../utilities/high-contrast";
@@ -97,6 +99,9 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = {
         ...applyFocusVisible("& $callToAction_glyph", {
             ...applyGlyphTransform(),
         }),
+        [highContrastSelector]: {
+            ...highContrastOptOutProperty,
+        },
     },
     callToAction_glyph: {
         fill: neutralForegroundRest,
@@ -111,12 +116,27 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = {
         ...highContrastForeground,
     },
     callToAction__primary: {
-        "& $callToAction_glyph": {
+        "& $button_contentRegion, $callToAction_glyph": {
             fill: accentForegroundCut,
             ...highContrastSelectedForeground,
         },
-        "&:hover $callToAction_glyph": {
-            ...highContrastHighlightForeground,
+        "&:hover": {
+            "& $callToAction_glyph": {
+                ...highContrastHighlightForeground,
+            },
+        },
+        "a&": {
+            "& $button_contentRegion, & $callToAction_glyph": {
+                ...highContrastLinkForeground,
+            },
+            "&:hover": {
+                [highContrastSelector]: {
+                    "box-shadow": "none !important",
+                },
+                "a& & $button_contentRegion, & $callToAction_glyph": {
+                    ...highContrastLinkForeground,
+                },
+            },
         },
     },
     callToAction__lightweight: {
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -124,18 +124,18 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = {
             "& $callToAction_glyph": {
                 ...highContrastHighlightForeground,
             },
-        },
-        "a&": {
-            "& $button_contentRegion, & $callToAction_glyph": {
-                ...highContrastLinkForeground,
-            },
-            "&:hover": {
+            "a&": {
                 [highContrastSelector]: {
                     "box-shadow": "none !important",
                 },
                 "a& & $button_contentRegion, & $callToAction_glyph": {
                     ...highContrastLinkForeground,
                 },
+            },
+        },
+        "a&": {
+            "& $button_contentRegion, & $callToAction_glyph": {
+                ...highContrastLinkForeground,
             },
         },
     },
--- a/packages/fast-components-styles-msft/src/call-to-action/index.ts
+++ b/packages/fast-components-styles-msft/src/call-to-action/index.ts
@@ -128,15 +128,10 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = {
                 [highContrastSelector]: {
                     "box-shadow": "none !important",
                 },
-                "a& & $button_contentRegion, & $callToAction_glyph": {
-                    ...highContrastLinkForeground,
-                },
             },
         },
-        "a&": {
-            "& $button_contentRegion, & $callToAction_glyph": {
-                ...highContrastLinkForeground,
-            },
+        "a& $button_contentRegion, a& $callToAction_glyph, a&:hover $button_contentRegion, a&:hover $callToAction_glyph": {
+            ...highContrastLinkForeground,
         },
     },
     callToAction__lightweight: {
--- a/packages/fast-components-styles-msft/src/carousel/index.ts
+++ b/packages/fast-components-styles-msft/src/carousel/index.ts
@@ -96,7 +96,7 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
             "&:hover": {
                 "&::before": {
                     opacity: "0.9",
-                    ...highContrastHighlightColorIndicator
+                    ...highContrastHighlightColorIndicator,
                 },
             },
         },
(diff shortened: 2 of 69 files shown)


chrisdholt:
Can you update the PR title to better reflect exactly what is going in? The current title makes it seem like it's a feature `high contrast in Edge chromium`. What **exactly** is being fixed here?

eljefe223:
Auto Suggest:
image

Text Action:
image

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -111,12 +116,27 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = {
...
+        },
+        "a&": {
+            "& $button_contentRegion, & $callToAction_glyph": {
+                ...highContrastLinkForeground,
+            },
+            "&:hover": {
+                [highContrastSelector]: {
+                    "box-shadow": "none !important",
+                },
+                "a& & $button_contentRegion, & $callToAction_glyph": {
```
nicholasrice: I think this selector can be combined with the above.

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -8,45 +8,75 @@ export const highContrastSelector: string = "@media (-ms-high-contrast:active)";
...
+    return {
+        "-ms-high-contrast-adjust": "none",
+    };
+}
+// Used to to opt-out of high contrast color scheme
+export const highContrastOptOutProperty: CSSRules<{}> = applyhighContrastOptOutProperty();
+
+// Function used to to set link color base on 'forced-color' query
+export function applyHighContrastlinkValue(): string {
+    return window.matchMedia("(forced-colors: none)").matches ||
```
nicholasrice: I think you should break this out further - can you create a `supportsForcedColors` in `fast-web-utilities` and use that function here?

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -8,45 +8,75 @@ export const highContrastSelector: string = "@media (-ms-high-contrast:active)";
...
+export function applyhighContrastOptOutProperty(): CSSRules<{}> {
+    return {
+        "-ms-high-contrast-adjust": "none",
+    };
+}
+// Used to to opt-out of high contrast color scheme
+export const highContrastOptOutProperty: CSSRules<{}> = applyhighContrastOptOutProperty();
+
+// Function used to to set link color base on 'forced-color' query
+export function applyHighContrastlinkValue(): string {
```
nicholasrice: naming nit: applyHighContrast**L**inkValue

packages/fast-components-styles-msft/src/toggle/index.ts
```diff
@@ -153,39 +164,47 @@ const styles: ComponentStyles<ToggleClassNameContract, DesignSystem> = {
...
+                    "border-color": importantValue(HighContrastColor.selectedBackground),
+                    "& + span": {
+                        background: importantValue(HighContrastColor.selectedText),
                     },
                 },
             },
+            [highContrastSelector]: {
+                background: HighContrastColor.selectedBackground,
+                "border-color": HighContrastColor.selectedBackground,
+                "& + span": {
```
janechu: I realize we are using "+" in a few places, but we should add more selectors so we won't need to. This is costly during rendering as the browser has to do more work to determine styles on a give element.

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
```diff
@@ -8,45 +8,75 @@ export const highContrastSelector: string = "@media (-ms-high-contrast:active)";
...
     selectedBackground = "Highlight",
     buttonText = "ButtonText",
     buttonBackground = "ButtonFace",
-    background = "Background",
+    background = "Window",
 }
 
-// Used to remove text backplate and borders in 'button-text' colors
+// Function used to to opt-out of high contrast color scheme for '-ms' prefixes
+export function applyhighContrastOptOutProperty(): CSSRules<{}> {
```
janechu: if this function is only being used by the `highContrastOptOutProperty` you can remove the export.

khamudom: export removed

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: updated

packages/fast-components-styles-msft/src/utilities/high-contrast.ts
khamudom: add function to the dom file in fast-web-utilities

packages/fast-components-styles-msft/src/call-to-action/index.ts
khamudom: updated

packages/fast-components-styles-msft/src/call-to-action/index.ts
```diff
@@ -111,12 +116,27 @@ const styles: ComponentStyles<CallToActionClassNameContract, DesignSystem> = {
...
+            "a&": {
+                [highContrastSelector]: {
+                    "box-shadow": "none !important",
+                },
+                "a& & $button_contentRegion, & $callToAction_glyph": {
+                    ...highContrastLinkForeground,
+                },
+            },
+        },
+        "a&": {
```
janechu: this could be:
```typescript
"a& $button_contentRegion, a& $callToAction_glyph, a&:hover $button_contentRegion, a&:hover $callToAction_glyph": {
    ...highContrastLinkForeground,
}
```

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -210,20 +212,22 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
...
                 "border-color": lightModeNeutralOutlineRest,
+                ...highContrastButtonColorIndicator,
             },
             "&$carousel_sequenceIndicator__active": {
                 "&::before": {
                     background: lightModeNeutralFillStealthRest,
-                    ...highContrastHighlightBackground,
+                    ...highContrastHighlightColorIndicator,
                 },
             },
```
janechu: I think this might be better as:

```typescript
"&::before: {
    "border-color": lightModeNeutralOutlineRest,
},
"&::before, &$carousel_sequenceIndicator__active": {
    background: lightModeNeutralFillStealthRest,
    ...highContrastButtonColorIndicator,
}
```

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -160,20 +158,22 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
...
                 "border-color": darkModeNeutralOutlineRest,
+                ...highContrastButtonColorIndicator,
             },
             "&$carousel_sequenceIndicator__active": {
                 "&::before": {
                     background: darkModeNeutralFillStealthRest,
-                    ...highContrastHighlightBackground,
+                    ...highContrastHighlightColorIndicator,
                 },
             },
```
janechu: This could also be consolidated a little,

```typescript
"&::before": {
    "border-color": darkModeNeutralOutlineRest,
},
"&::before, &$carousel_sequenceIndicator__active": {
    background: darkModeNeutralFillStealthRest,
    ...highContrastHighlightColorIndicator,
}
```

PR #2359, opened 2019-10-18
feat: add viewport context

scomea:
# Description
Viewport context is intended to provide developers with an easy way to set the viewport for a region of the dom tree without having to pass down the reference through multiple component layers.  Any viewport positioners that are children of the context will use that as the viewport if none are specified in it's props.

## Motivation & context
Make developers life easier

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
@@ -18,6 +18,7 @@ import {
     ViewportPositionerUnhandledProps,
     ViewportPositionerVerticalPosition,
 } from "./viewport-positioner.props";
+import { ThemeProvider } from "emotion-theming";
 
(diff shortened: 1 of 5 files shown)


packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -18,6 +19,8 @@ import {
     ViewportPositionerUnhandledProps,
     ViewportPositionerVerticalPosition,
 } from "./viewport-positioner.props";
+import { ThemeProvider } from "emotion-theming";
```
janechu: this doesn't appear to get used, it's also not part of our package dependencies.

scomea: Thanks for catching.  VS likely inserted that trying to be helpful.

PR #2360, opened 2019-10-19
fix: use only context and jss styles to apply select menu background

scomea:
# Description
The approach of using actual wrapping dom elements was interfering with advanced layout options for select menu.  A slight change of approach allows passing of color to base component using only context and not actually wrapping the menu in a dom element. 

## Motivation & context
Removing wrapping dom element and using only context + css makes layout easier (and removes an element from the dom)

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/select/index.ts
+++ b/packages/fast-components-styles-msft/src/select/index.ts
@@ -67,7 +67,9 @@ const styles: ComponentStyles<SelectClassNameContract, DesignSystem> = {
         ...highContrastForeground,
     },
     select_menu: {
-        background: neutralFillStealthRest,
+        background: (designSystem: DesignSystem): string => {
+            return designSystem.backgroundColor;
+        },
         ...applyElevatedCornerRadius(),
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/select/index.ts
```diff
@@ -67,7 +67,9 @@ const styles: ComponentStyles<SelectClassNameContract, DesignSystem> = {
         ...highContrastForeground,
     },
     select_menu: {
-        background: neutralFillStealthRest,
+        background: (designSystem: DesignSystem): string => {
+            return designSystem.backgroundColor;
+        },
```
bheston: You should be able to just do `background: backgroundColor,`.

scomea: Actually that doesn't work.  Looking at this with Nick it became apparent that the new background color got inherited by the select options but not the select menu BG.  The fix is to actually set the menu background in the style as well as the context to cover both cases.

PR #2367, opened 2019-10-22
fix: refactoring elevation utilities

nicholasrice:
# Description
Reduce size of elevation utility code.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/elevation.spec.ts
+++ b/packages/fast-components-styles-msft/src/utilities/elevation.spec.ts
@@ -48,16 +48,16 @@ describe("applyElevation", () => {
     });
     test("should calculate elevation value correctly", () => {
         expect(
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/utilities/elevation.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        expect(
+            elevationShadow(ElevationMultiplier.e17, "#000", ambientShadowConfig)(
+                defaultDesignSystem
+            )
+        ).toBe("0px 14.4px 43.2px rgba(0,0,0,0.18)");
+    });
+});
+
+describe("applyElevation", () => {
+    test("should calculate elevation value correctly", () => {
```
janechu: These tests all say the same thing, if one fails we need to know which so I would recommend you add in something to distinguish each one.

PR #2375, opened 2019-10-23
feat: add ability to pass custom resize control to pane and row

chrisdholt:
# Description
Currently, the `row` and `pane` components in `fast-layouts-react` create their own button elements to support resizing via click and keyboard. For a completely accessible experience, there likely needs to be additional context passed such as an aria-label, name, or title attribute to provide non-sighted users details about the control. Because these controls are created within the component, any attempt to pass additional attributes, aria-labels, etc would require either a complex object or props enumerated for each attribute we would want to add.

Rather than trying to enumerate each or create a complex object, this change gives users the ability to pass a custom control with attributes specific to their requirements. We will pass them the className, the role, and the handlers for click and keydown.

This change allows us to support this in a non-breaking manner.

## Issue type checklist

## Process & policy checklist


packages/fast-layouts-react/src/pane/pane.tsx
```diff
@@ -234,19 +235,25 @@ export class Pane extends Foundation<PaneHandledProps, PaneUnhandledProps, PaneS
...
-                role={"separator"}
-            />
-        );
+        const resizeProps: PaneResizeControlProps = {
+            className: this.props.managedClasses.pane_resizeHandle,
+            onMouseDown: this.onMouseDown,
+            onKeyDown: this.onKeyDown,
+            role: "separator",
+        };
+
```
nicholasrice: can you set a default prop and then just always use the prop?

packages/fast-layouts-react/src/row/row.tsx
```diff
@@ -222,19 +227,25 @@ export class Row extends Foundation<
...
 
-        return (
-            <button
-                className={this.props.managedClasses.row_resizeHandle}
-                onMouseDown={this.onMouseDown}
-                onKeyDown={this.onKeyDown}
-                role={"separator"}
-            />
-        );
+        const resizeProps: RowResizeControlProps = {
```
nicholasrice: same

PR #2382, opened 2019-10-24
chore: update repository to yarn

chrisdholt:
# Description

## Issue type checklist

## Process & policy checklist


nicholasrice:
Looks great if we can get it to build

CONTRIBUTING.md
```diff
@@ -7,19 +7,13 @@
...
+Install all package dependencies:
 
 ```bash
-npm install
-```
-
-Install dependencies for packages within the project:
-
-```bash
-lerna bootstrap
```
nicholasrice: Can you add a new step here - `lerna run prepare`.

That command will largely take the place of `lerna bootstrap` with yarn

chrisdholt: Updated - Do we perhaps want to see if we can setup a githook for post-install to just run this?

chrisdholt: ...in a different PR

packages/fast-breakpoint-tracker-react/README.md
```diff
@@ -4,7 +4,7 @@ A library for managing viewport breakpoints. Given a set of breakpoints, this co
 
 ## Installation
 
-`npm i --save @microsoft/fast-breakpoint-tracker-react`
+`yarn i --save @microsoft/fast-breakpoint-tracker-react`
```
janechu: I think this should still be npm

chrisdholt: updated here and everywhere.

packages/fast-tooling-react/src/form/templates/template.control.utilities.tsx
```diff
@@ -32,7 +32,7 @@ abstract class ControlTemplateUtilities<P, S> extends React.Component<
         }
     }
 
-    public renderSoftRemove(className: string): React.ReactNode {
+    public renderSoftRemove(className: string): JSX.Element {
```
janechu: can we change this file back? we may be exporting it at some point and we want to be as permissive as possible, it shouldn't error.

CONTRIBUTING.md
nicholasrice: Yea let's do any of that as a follow up. I'm not convinced yet that we really need to add additional complications, but if we do I think it will become readily apparent.

packages/fast-tooling-react/src/form/templates/template.control.utilities.tsx
nicholasrice: Tried this out - reverting doesn't cause issues.

PR #2393, opened 2019-10-29
fix: improve select aria

scomea:
# Description
On focus the Select component was not providing desired info to assistive technologies.  When a select menu is not open screen readers should recite: the label of the select control, the current value, that it is a combo box, that it is collapsed

## Motivation & context
Address accessibility requirements
https://github.com/microsoft/fast-dna/issues/2214

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/select/select.props.ts
```diff
@@ -22,11 +17,13 @@ export interface SelectHandledProps extends SelectManagedClasses {
...
      * Custom function to render the trigger of the control
      * (ie. the ui that typically displays the current value and opens the menu in single select mode)
+     * Note: the optional triggerId param has been depracated and will always be null until it is removed
+     * in a future major version update
      */
     trigger?: (
         props: SelectProps,
         state: SelectState,
-        triggerId?: string
+        triggerId?: string //depracated
```
janechu: nit: I think this is the first time we're deprecating a param, I found this from JSDoc as a solution for how to mark it as deprecated: https://github.com/jsdoc/jsdoc.github.io/issues/39

scomea: Thanks, shows up in VS tooltips.

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -464,20 +464,20 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
...
-                aria-selected={isItemSelected}
-                aria-posinset={isItemSelected ? state.selectedItemIndex : null}
-                aria-setsize={isItemSelected ? state.selectableItemCount : null}
+                disabled={props.disabled}
+                aria-disabled={this.props.disabled}
+                aria-expanded={this.state.isMenuOpen}
+                aria-labelledby={this.props.labelledBy || null}
+                aria-multiline={false}
+                value={state.displayString}
+                role="textbox"
```
janechu: Why is this a button element if it's a textbox? Could it be an input?

scomea: Basically Aria does not have good way to describe this control (silly but true).  We wanted the act of focusing on a collapsed select component to cause screen readers to enumerate "control label", "current value", "control type", "iscollapsed".  The only component that does this in the widget list (https://www.w3.org/TR/wai-aria-1.1/#widget_roles) turned out to be combobox.  We tried menu, but that doesn't enumerate value.  Aria combo boxes expect to own an element with role of "textbox" and need that in order to have something to extract the current value from.  So no, we should not make it actually be an input as we don't actually want that functionality, but we should have our accessibility folks raise this with the spec making folks at W3.

janechu: Without context in the code there's no way for anyone to know that, so could you drop a comment in here explaining why we are using a combobox when this is not actually one.

scomea: Added some comments.

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -157,6 +157,8 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
      * Renders the component
      */
     public render(): React.ReactElement<HTMLDivElement> {
+        // we set role="combobox" in order to get assistive technologies to enumerurate
```
janechu: nit: spelling "enumerate", "focused", also I think a little copy editing might make this more understandable, I am still not quite sure what setting the role to combobox is achieving from this comment. What does it prevent from happening, what does it ensure happens? I just see a list of properties. Does it read them in that order? If so, I would also reword that to be more like "to enumerate a focused trigger elements label, value, role and whether the listbox is expanded".

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -464,20 +466,25 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (props.multiselectable) {
             return null;
         }
+
         const isItemSelected: boolean = state.selectedItemIndex !== 0;
+
+        // we give the trigger button the role of textbox in order to get
+        // because Aria expects elements with role of "combobox", as is the case
+        // the select parent, to own a textbox from which assistive technologies get
+        // the current value.
```
janechu: This is a confusing read, can you take a stab at re-wording it a little?

packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -166,11 +168,10 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
                 onClick={this.handleClick}
                 onFocus={this.handleFocus}
                 tabIndex={-1}
-                role="listbox"
+                role="combobox"
```
nicholasrice: What is the justification for making this a `combobox`? https://www.w3.org/TR/wai-aria-1.1/#combobox establishes a combobox as a textbox with autosuggest functionality - isn't select an implementation of the native `select` element?

scomea: Native selects are announce as a combobox.  There are some definite inconsistencies between native implementations and aria spec.

packages/fast-components-react-base/src/select/select.tsx
scomea: Change to a more concise  
// we set role="combobox" in order to match how native select elements
// are described by assistive technologies

packages/fast-components-react-base/src/select/select.tsx
scomea: More concise -
        // we give the trigger button the role of "textbox"
        // because Aria expects elements with role of "combobox" to contain a textbox

packages/fast-components-react-base/src/select/select.tsx
nicholasrice: It appears that ARIA is moving their implementation guidance towards the native select: http://w3c.github.io/aria/#combobox, so I'm supportive of this change.

chrisdholt: Can we audit this control to ensure we're meeting the latest standards linked to above? If we're going to change the role here it's the best time to make sure the control accurately reflects the above draft.

scomea: I've messed with it some and I have not been able to come up with a better behavior than what we have.  Biggest issue I see now is that narrator calls it a "ComboEdit" instead of a "ComboBox".  My recommendation is that we go with this and create a task to revisit when aria 1.2 is implemented and we can tweak over to something closer to this (I don't think we'll be too far off) - https://github.com/w3c/aria/wiki/Resolving-ARIA-1.1-Combobox-Issues

chrisdholt: Sounds good - thanks for looking deeper into this. Can you link to the issue when you get it written up? Thx!

scomea: Here you go - https://github.com/microsoft/fast-dna/issues/2454

rfairweather: **Current behavior**
-	Narrator+UIA: “Zoom, 100%, 8 of 17, selected”
-	NVDA+IA2: “Zoom  list  collapsed  100%, 100%  8 of 17”

I think this is a solution we should seriously: [https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html
](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html)

_“The following example implementation of the design pattern for listbox demonstrates a collapsible single-select listbox widget that is functionally similar to an HTML select input with the attribute size="1"”_ 

**Proposed current behavior**
-	Narrator+UIA: “Choose an element: Neptunium, menu, has popup”
-	NVDA+IA2: “Choose an element: Neptunium, menu button  submenu”

PR #2399, opened 2019-10-31
fix: slider can miss dir setting on initial render.

scomea:
## Description
Slider's mechanism for detecting direction (ltr/rtl) could fail to detect the correct setting on initial render.  We now check after mounting and after any renders.

## Motivation & context
Sliders would be incorrect on initial render in rtl scenarios.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/slider/slider.tsx
+++ b/packages/fast-components-react-base/src/slider/slider.tsx
@@ -21,11 +21,7 @@ import {
 } from "@microsoft/fast-web-utilities";
 import ReactDOM from "react-dom";
 import { SliderClassNameContract } from "@microsoft/fast-components-class-name-contracts-base";
-import {
-    classNames,
-    Direction,
-    DirectionAttributeName,
-} from "@microsoft/fast-web-utilities";
+import { classNames, Direction } from "@microsoft/fast-web-utilities";
 import { DisplayNamePrefix } from "../utilities";
 import { SliderContext, SliderContextType } from "./slider-context";
 import SliderTrackItem, {
@@ -70,6 +66,7 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
     private static minIncrementDelay: number = 100;
     private static incrementAcceleration: number = 50;
     private static rolePropName: string = "role";
+    private static DirectionAttributeName: string = "dir";
 
     protected handledProps: HandledProps<SliderHandledProps> = {
         disabled: void 0,
@@ -657,11 +654,11 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
         }
 
         const closest: Element = this.rootElement.current.closest(
-            `[${DirectionAttributeName}]`
+            `[${Slider.DirectionAttributeName}]`
         );
 
         return closest === null ||
-            closest.getAttribute(DirectionAttributeName) === Direction.ltr
+            closest.getAttribute(Slider.DirectionAttributeName) === Direction.ltr
             ? Direction.ltr
             : Direction.rtl;
     };
--- a/packages/fast-components-react-base/src/slider/slider.tsx
+++ b/packages/fast-components-react-base/src/slider/slider.tsx
@@ -45,6 +45,7 @@ export interface SliderState {
     isIncrementing: boolean;
     incrementDirection: number;
     usePageStep: boolean;
+    direction: Direction;
 }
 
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -628,22 +634,35 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
+    /**
+     *  gets the current direction
+     */
+    private getDirection = (): Direction | null => {
         if (this.rootElement.current === null) {
-            return;
+            return Direction.ltr;
         }
 
         const closest: Element = this.rootElement.current.closest(`[dir]`);
```
janechu: nit: don't need backticks here

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -628,22 +634,35 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
...
 
         if (closest === null) {
-            this.direction = Direction.ltr;
-            return;
+            return Direction.ltr;
         }
 
-        this.direction =
-            closest.getAttribute("dir") === "rtl" ? Direction.rtl : Direction.ltr;
+        return closest.getAttribute(Slider.dirPropName) === "rtl"
```
janechu: I'd assume this can just be `this.dirPropName` to refer to the local private instance. Also "rtl" should be `Direction.ltr`.

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -786,7 +805,7 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
         }
 
         if (
-            this.direction === "rtl" &&
+            this.state.direction === "rtl" &&
```
janechu: same here, I would use the `Direction` enum.

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -308,7 +315,7 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
                 [slider__incrementing, this.state.isIncrementing],
                 [slider__vertical, isVertical],
                 [slider__horizontal, !isVertical],
-                [slider__rtl, this.direction === "rtl"],
+                [slider__rtl, this.state.direction === "rtl"],
```
janechu: and here, would use the `Direction` enum.

packages/fast-components-react-base/src/slider/slider.tsx
janechu: nit: You could combine this with the previous if statement:

```typescript
return closest === null || closest.getAttribute(this.dirPropName) === Direction.rtl
    ? Direction.ltr
    : Direction.rtl;
```

scomea: dirPropName is static so it doesn't exist on "this".

packages/fast-components-react-base/src/slider/slider.tsx
janechu: you could use your dirPropName here.

packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -66,6 +67,7 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
     private static minIncrementDelay: number = 100;
     private static incrementAcceleration: number = 50;
     private static rolePropName: string = "role";
+    private static dirPropName: string = "dir";
```
janechu: I'm realizing that "dir" and "role" should exist in the `fast-web-utilities` package, not as part of a single class.

@chrisdholt @nicholasrice do we want to provide an exported set of attribute strings for these purposes?

nicholasrice: we do export a `Direction` object.

packages/fast-components-react-base/src/slider/slider.tsx
scomea: yep

packages/fast-components-react-base/src/slider/slider.tsx
scomea: I added a string for 'dir' called "DirectionAttributeName" to utilities.  Does that fit with your thoughts?

PR #2401, opened 2019-11-01
fix: modal dialog now blocks touch events

scomea:
# Description
The modal overlay was blocking click events but not touch events.

## Motivation & context
Touch users should not be able to interact with elements behind modal overlay 

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -114,28 +114,49 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
+     */
     private handleOverlayClick = (event: React.MouseEvent): void => {
-        if (
-            this.props.onDismiss &&
-            typeof this.props.onDismiss === "function" &&
-            this.props.visible
-        ) {
-            this.props.onDismiss(event);
-        }
+        this.checkForSoftDismiss(event);
```
janechu: To reduce falling out of sync by having to maintain two methods that do the same thing, can we make touch and click the same method? Perhaps `handleOverlaySelection`.

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -163,16 +167,12 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
-    private handleOverlayClick = (event: React.MouseEvent): void => {
-        if (
-            this.props.onDismiss &&
-            typeof this.props.onDismiss === "function" &&
-            this.props.visible
-        ) {
-            this.props.onDismiss(event);
-        }
+    private handleOverlaySelection = (
+        event: React.TouchEvent | React.MouseEvent
```
nicholasrice: why do you need this method? Seems since you're simply invoking `checkForSoftDismiss` with the event object that you can just provide `this.checkForSoftDismiss` as the listener

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -145,9 +145,13 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
         return (
             <div
                 className={classNames(this.props.managedClasses.dialog_modalOverlay)}
-                onClick={this.handleOverlayClick}
+                onClick={this.handleOverlaySelection}
+                onTouchStart={this.handleOverlaySelection}
                 role={"presentation"}
                 tabIndex={-1}
+                style={{
```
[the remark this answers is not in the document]
scomea: It's intrinsic so I think it should be applied to the base level component.

packages/fast-components-react-base/src/dialog/dialog.tsx
scomea: yep

awentzel:
The PR title should describe what the change or addition is rather than what the problem is.

PR #2411, opened 2019-11-06
fix: update divider recipe to invert in dark mode and adjusted delta

bheston:
# Description

Update divider recipe to invert in dark mode and adjusted delta.

## Motivation & context

The earlier dark on dark was an exploration with partner teams, but is not a general convention, so adjusting for a more traditional style.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/design-system/index.ts
+++ b/packages/fast-components-styles-msft/src/design-system/index.ts
@@ -268,7 +268,7 @@ const designSystemDefaults: DesignSystem = {
     neutralForegroundActiveDelta: 0,
     neutralForegroundFocusDelta: 0,
 
-    neutralDividerRestDelta: 6,
+    neutralDividerRestDelta: 8,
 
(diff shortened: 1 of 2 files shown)


bheston:
@chrisdholt It seems likely we will need a way to achieve guaranteed 3:1 contrast, but I think that fits in to the existing outline recipe. We may need to add some rails to it to ensure it meets contrast as well as / instead of using deltas.

chrisdholt:
> @chrisdholt It seems likely we will need a way to achieve guaranteed 3:1 contrast, but I think that fits in to the existing outline recipe. We may need to add some rails to it to ensure it meets contrast as well as / instead of using deltas.

Sounds good - likely a follow-up task, but would be great if we mapped the outline recipe to the root element here and then the inaccessible recipe to a selector of `&:[role="presentation"]: {}`. That would ensure that dividers with semantic roles meet contrast and those with a role of presentation (the non-implicit role) would not have to.

PR #2424, opened 2019-11-14
feat: adds React utilities package

nicholasrice:
# Description
Adds a React utilities library for general purpose utilities that depend on React APIs.

## Motivation & context
As we move toward using more functional components and hooks, having a general purpose React utility library that can be leveraged by all packages will be useful and promote general-purpose factoring of UI logic.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-react-utilities/README.md
+++ b/packages/fast-react-utilities/README.md
@@ -1,0 +1,36 @@
+# FAST Components React MSFT
+
+A set of React components which implements the Microsoft styling.
+
+## Installation
+
+`npm i --save @microsoft/fast-components-react-msft`
+
+## Usage
+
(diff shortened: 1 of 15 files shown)


packages/fast-react-utilities/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+When two duration values are provided, the first will be used when `active` goes from `false`->`true` and the second will be used when `active` goes from `true`->`false`.
+
+```js
+function FancyDialog(props) {
+    const state = useTransitionState(props.visible); 
+
+    // eg, apply classes dependent on if the dialog is
+    // resting in an active / inactive state or if it is in the
+    // middle of trasitioning to active / inactive
```
chrisdholt: Could we provide one example of this either in place of or in addition to the comment here?

PR #2425, opened 2019-11-14
fix: update state for sections and children control when schemas or data have been updated

janechu:
# Description

This fixes an issue where state was falling out of sync when props were updated in the form section and children control.

## Motivation & context

This caused an error when switching between certain schemas which had similar properties after render had been invoked. This logic has now been moved to the getDerivedStateFromProps lifecycle hook so that state may be updated before rendering.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/form/controls/control.children.tsx
+++ b/packages/fast-tooling-react/src/form/controls/control.children.tsx
@@ -800,10 +800,12 @@ class ChildrenControl extends React.Component<
         return `${this.props.dataLocation}${propLocation}`;
     }
 
-    private getChildOptionTextString(item: any): string {
-        const textString: string = typeof item.props === "string" ? item.props : item;
-
-        return textString ? textString : "Untitled";
+    private getChildOptionTextString(item: string | ChildComponentConfig): string {
+        return typeof item === "string"
+            ? item
+            : typeof item.props === "string"
+                ? item.props
+                : "Untitled";
     }
 }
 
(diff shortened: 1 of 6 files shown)


packages/fast-tooling-react/src/form/controls/control.children.tsx
```diff
@@ -39,6 +39,19 @@ class ChildrenControl extends React.Component<
         managedClasses: {},
     };
 
+    public static getDerivedStateFromProps(
+        props: ChildrenControlProps,
+        state: ChildrenControlState
```
nicholasrice: You're doing somthing very similar in `compnentDidUpdate` - is that still needed?

packages/fast-tooling-react/src/form/controls/control.children.tsx
```diff
@@ -39,6 +39,19 @@ class ChildrenControl extends React.Component<
...
+    ): Partial<ChildrenControlState> {
+        if (props.value !== state.data) {
+            return {
+                data: [].concat(props.value || []),
```
nicholasrice: if `this.state.data` is really just a shallow copy of `this.props.value`, do you really need to do this at all? props and state should never be mutated, so it seems like you could simply substitute `this.props.value` where you're using `this.state.data`

janechu: We need it for dragging changes, since when you drag you need to show the updated state but the props arent updated until the drop is finished.

packages/fast-tooling-react/src/form/controls/control.children.tsx
janechu: Turns out it was unnecessary, removed.

PR #2441, opened 2019-11-15
fix: delay rendering slider internals until after direction check

scomea:
# Description
The slider component can only query the dom for direction until after it has mounted.  Previously we rendered in default ltr and if we detect rtl after the component mounted we reset the state and triggered a render to correct the layout to rtl.  This meant we ran layout logic twice in rtl and that css transitions could be triggered by the ltr to rtl layout changes.  With this change we only render the component root element on the first pass, check for direction on componentDidMount() and only render the component internals at that point.

## Motivation & context
Users reported seeing transition animations on rtl ui.  Slight perf gain on rtl by avoiding running layout twice.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/slider/slider.spec.tsx
+++ b/packages/fast-components-react-base/src/slider/slider.spec.tsx
@@ -7,6 +7,7 @@ import Slider, {
     SliderUnhandledProps,
 } from "./slider";
 import {
+    Direction,
     keyCodeArrowDown,
     keyCodeArrowLeft,
     keyCodeArrowRight,
@@ -767,7 +768,7 @@ describe("Slider", (): void => {
         document.body.removeChild(container);
     });
 
-    test("first render state flag is false after component has mounted", (): void => {
+    test("direction is set after component has mounted", (): void => {
         const container: HTMLDivElement = document.createElement("div");
         document.body.appendChild(container);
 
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/slider/slider.tsx
```diff
@@ -174,6 +175,7 @@ class Slider extends Foundation<SliderHandledProps, SliderUnhandledProps, Slider
             incrementDirection: 1,
             usePageStep: false,
             direction: Direction.ltr,
```
nicholasrice: I think you can avoid adding an additional state property by defaulting the `direction` state property to `null`. Then the check in render becomes "do I know my direction? If not, don't render anything", which feels more in line with the purpose of this change.

chrisdholt: 🔥 🔥 🔥 - I like this.

scomea: yup, good call

packages/fast-components-react-base/src/slider/slider.spec.tsx
```diff
@@ -767,6 +768,21 @@ describe("Slider", (): void => {
...
+        const rendered: any = mount(
+            <Slider managedClasses={managedClasses} initialValue={50} />,
+            {
+                attachTo: container,
+            }
+        );
+
+        expect(rendered.state("direction")).toBe(Direction.ltr);
+        document.body.removeChild(container);
+    });
```
chrisdholt: @scomea should we add a test to ensure that if the direction is null the slider does not render?

scomea: figured one out

PR #2405, opened 2019-11-05
fix: update carousel sequence indicator to be centered without using full width of parent

SethDonohue:
# Description

This changes the sequence indicators container to be centered without using the full width of its parent and therefor not covering up the actions on content in the carousel.
## Motivation & context

The carousel sequence indicators container is currently the full width of the carousel, which can cover content actions, such as when a card is used that has actions in the lower right or left corner of the carousel view.

BEFORE:
Screen Shot 2019-11-05 at 1 27 18 PM

AFTER:
Screen Shot 2019-11-05 at 1 34 22 PM

## Issue type checklist

- [ ] **Chore**: A change that does not impact distributed packages.
- [x] **Bug fix**: A change that fixes an issue, link to the issue above.
- [ ] **New feature**: A change that adds functionality.

**Is this a breaking change?**
- [ ] This change causes current functionality to break.

## Process & policy checklist

- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/en/contributing/standards) for this project.

PR #2447, opened 2019-11-15
chore: add story with buttons to track sequence indicator container width

SethDonohue:
# Description

## Motivation & context

Add story to Carousel to showcase that the sequence indicator container should not be overlapping any actions in the content corners of the carousel. This is partially dependent on the implementation of the carousel content (as the width of the indicators must be reduced), but is described in the story so it is clearly known what to look for when viewing the story.
This story is associated with PR #2405 which fixed the sequence indicator container going full width across the carousel and covering up content.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/assets/carousel-hero-content.tsx
+++ b/packages/fast-components-react-msft/assets/carousel-hero-content.tsx
@@ -1,18 +1,16 @@
-import React from "react";
+import React, { Fragment } from "react";
 import {
+    Button,
     CallToAction,
     CallToActionAppearance,
     CallToActionProps,
     CarouselSlideTheme,
     Heading,
     HeadingProps,
-    HeadingSize,
-    HeadingTag,
     Image,
     ImageProps,
     Paragraph,
     ParagraphProps,
-    ParagraphSize,
 } from "../src";
 
 export interface CarouselHeroContentProps {
@@ -22,6 +20,7 @@ export interface CarouselHeroContentProps {
     image?: ImageProps;
     className?: string;
     theme?: CarouselSlideTheme;
+    extraButtons?: boolean;
 }
 
 /**
@@ -46,6 +45,7 @@ export default class CarouselHeroContent extends React.Component<
             href: "#",
             appearance: CallToActionAppearance.primary,
         },
+        extraButtons: false,
     };
 
     private imageSrc: string =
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-msft/assets/carousel-hero-content.tsx
```diff
@@ -46,16 +45,31 @@ export default class CarouselHeroContent extends React.Component<
...
         },
+        extraButtons: false,
     };
 
     private imageSrc: string =
         this.props.theme === CarouselSlideTheme.light
             ? "http://placehold.it/1399x600/2F2F2F/171717"
             : "http://placehold.it/1399x600/";
 
+    private extraButtonRightStyle: any = {
```
chrisdholt: Are these actually *any*? (and all the way down)

packages/fast-components-react-msft/src/carousel/carousel.stories.tsx
```diff
@@ -3,163 +3,87 @@ import React from "react";
...
+                    }
+                    theme={theme}
+                    extraButtons={extraButtons}
+                />
+            ),
+        });
+    }
+    return fillerArray;
+}
+
```
nicholasrice: Can you ensure the types are accurate here? A number of `any` types.

packages/fast-components-react-msft/assets/carousel-hero-content.tsx
```diff
@@ -1,18 +1,16 @@
-import React from "react";
+import React, { Fragment } from "react";
 import {
+    Button,
```
chrisdholt: nit: Can we use something other than button here? If it's "accent", then lets just use `<AccentButton />`

packages/fast-components-react-msft/assets/carousel-hero-content.tsx
```diff
@@ -1,18 +1,16 @@
-import React from "react";
+import React, { Fragment } from "react";
```
chrisdholt: FYI - You're already bringing in React, you *could* reference here as `React.Fragment` without the import.

packages/fast-components-react-msft/src/carousel/carousel.stories.tsx
```diff
@@ -3,163 +3,87 @@ import React from "react";
...
+const carouselSequenceStylesOverrides: any = {
```
chrisdholt: Can we type this?

packages/fast-components-react-msft/src/carousel/carousel.stories.tsx
SethDonohue: Yeah, sorry spaced on that.

packages/fast-components-react-msft/assets/carousel-hero-content.tsx
```diff
@@ -81,13 +101,13 @@ export default class CarouselHeroContent extends React.Component<
                         }}
                     >
                         <Heading
-                            tag={HeadingTag.h2}
-                            size={HeadingSize._3}
+                            tag={this.props.heading.tag}
+                            size={this.props.heading.size}
```
janechu: heading is optional, this will throw an error if it is not passed.

PR #2449, opened 2019-11-16
fix: allow default breakpoint for fast-layouts-react column component

### LINKED ISSUE #2398: Add new prop for width to fast-layouts-react Column component to support SSR
Like to be able to provide the Column component inside fast-layouts-react with a specific width value to be used in place of window.innerWidth on the server side, would be great if we can have a prop added so we can pass that down.

The intended usage is inside getValueByBreakpoint(). If canUseDOM is false (we are on the server) we would want to pass identifyBreakpoint() the prop to use in place of window.innerWidth.

Thanks :)

Reza-Rajabi:
Fixes #2398

# Description

updates fast-layouts-react column component to allow default breakpoint

## Motivation & context
fixing the issue addressed in #2398 to customize column width in layout component

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/column/column.props.ts
+++ b/packages/fast-layouts-react/src/column/column.props.ts
@@ -38,6 +38,12 @@ export interface ColumnHandledProps extends ColumnManagedClasses {
      * Provide this prop when doing server side rendering
      */
     cssGridPropertyName?: GridDisplay;
+
+    /**
+     * The default breakpoint (helpful when rendering server-side)
+     * Default to 0
+     */
+    defaultBreakpoint?: number;
 }
 
 export type ColumnProps = ColumnHandledProps & ColumnUnhandledProps;
--- a/packages/fast-layouts-react/src/column/column.tsx
+++ b/packages/fast-layouts-react/src/column/column.tsx
@@ -18,6 +18,7 @@ export class Column extends Foundation<ColumnHandledProps, ColumnUnhandledProps,
     public static defaultProps: Partial<ColumnProps> = {
         managedClasses: {},
         span: 12,
+        defaultBreakpoint: 0,
     };
 
     protected handledProps: HandledProps<ColumnHandledProps> = {
@@ -91,11 +92,9 @@ export class Column extends Foundation<ColumnHandledProps, ColumnUnhandledProps,
      * or the nearest preceding break-point if no entry exists for the current break-point
      */
     private getValueByBreakpoint<T>(breakpointSet: T[]): T {
-        if (!canUseDOM()) {
-            return breakpointSet[0];
-        }
-
-        const breakpoint: Breakpoint = identifyBreakpoint(window.innerWidth);
+        const breakpoint: Breakpoint = canUseDOM()
+            ? identifyBreakpoint(window.innerWidth)
+            : this.props.defaultBreakpoint;
 
         return breakpointSet.slice(0, breakpoint + 1).pop();
     }


packages/fast-layouts-react/src/column/column.props.ts
```diff
@@ -31,6 +31,11 @@ export interface ColumnHandledProps extends ColumnManagedClasses {
      * The gutter size of the parent Grid component
      */
     gutter?: GridGutter;
+
+    /**
+     * The default breakpoint of the parent Grid component
```
chrisdholt: This *technically* isn't the breakpoint of the grid; I think we want something like:
```suggestion
     * The default breakpoint (helpful when rendering server-side).
     * Defaults to 0

File: packages/fast-layouts-react/src/column/column.tsx [hunk 2]
```diff
@@ -91,7 +92,7 @@ export class Column extends Foundation<ColumnHandledProps, ColumnUnhandledProps,
      */
     private getValueByBreakpoint<T>(breakpointSet: T[]): T {
         if (!canUseDOM()) {
-            return breakpointSet[0];
+            return breakpointSet[this.props.defaultBreakpoint];
```
chrisdholt: Can you add a test for this configuration? If there isn't one already, would be great to have a test for the default as well.

Reza-Rajabi: Sure. I will check if there is any test already or not, and I will either update the existing or make a new one.

chrisdholt:
@Reza-Rajabi just checking in to see if I can help with anything on this PR - let me know!

Reza-Rajabi:
@chrisdholt Hi. Thank you for your follow up, and sorry for the delay. I am a bit busy with a school project, but I would like to get this task done as soon as possible. I wish I could have a bit more time to do the test part if it has no harm to the project.

packages/fast-layouts-react/src/column/column.tsx
JamesBurnside: This doesn't guard that the default breakpoint won't be outside the array. i.e. breakpointSet could be of size 5, and this.props.defaultBreakpoint could be 20.
It would be better to remove the canUseDOM check there and instead do

    const breakpoint: Breakpoint = canUseDOM()
            ? identifyBreakpoint(window.innerWidth)
            : this.props.defaultBreakpoint;

and let the slice-pop do its thing

Reza-Rajabi:
Will work on this in this weekend.

Reza-Rajabi:
@JamesBurnside for the `canUseViewport()`, I couldn't see if it has anything to do with this issue. Are we worry about `initial-scale`?

Reza-Rajabi:
I have applied the review changes.
Working on test now.

packages/fast-layouts-react/src/column/column.tsx
```diff
@@ -91,11 +92,9 @@ export class Column extends Foundation<ColumnHandledProps, ColumnUnhandledProps,
...
      */
     private getValueByBreakpoint<T>(breakpointSet: T[]): T {
-        if (!canUseDOM()) {
-            return breakpointSet[0];
-        }
-
-        const breakpoint: Breakpoint = identifyBreakpoint(window.innerWidth);
+        const breakpoint: Breakpoint = canUseDOM()
+            ? identifyBreakpoint(window.innerWidth)
+            : this.props.defaultBreakpoint;
```
chrisdholt: thanks for updating @Reza-Rajabi! Could you write a test for this scenario? If it helps, you can mimic @JamesBurnside's recent change to the Breakpoint Tracker in this package and mock out the `canUseDOM` call to be false to ensure that the column gets the expected defaultBreakpoint.

Reza-Rajabi: Sure

JamesBurnside:
> @JamesBurnside for the `canUseViewport()`, I couldn't see if it has anything to do with this issue. Are we worry about `initial-scale`?

Worried about the column span being wrong.

The issue is that canUseViewport under the hood does the following: `canUseDOM() && !!window.screen;`

The check currently is:

    if (!canUseViewport() || !Array.isArray(this.props.span)) {
            return Column.defaultProps.span as number;
    }

This means that if canUseDom() is false (as it is on the server side) and a default breakpoint is set, it won't retrieve the column span based on the defaultBreakpoint but instead just return `Column.defaultProps.span`

I'm not 100% sure but I would've thought this function could now be:

    private generateColumnSpan(): number {
        if (typeof this.props.span === "number") {
            return this.props.span;
        }

        if (!Array.isArray(this.props.span)) {
            return Column.defaultProps.span as number;
        }

        return this.getValueByBreakpoint(this.props.span);
    }

(i.e. remove the canUseViewport() check)

JamesBurnside:
@Reza-Rajabi are you able to work on this this week? this is an issue currently for us, let me know if not and I can finish it up tomorrow

Reza-Rajabi:
@JamesBurnside Thank you for asking. I can barely work this week. Feel free to work on it.

JamesBurnside:
Created here: #2531

chrisdholt:
> Created here: #2531

Thanks @JamesBurnside and @Reza-Rajabi for working together to get this landed. We'll move forward with the latest PR on this. 

@Reza-Rajabi feel free to grab other issues when you come back up for air!

Issue #2269, opened 2019-09-24
Export data-props from @microsoft/fast-components-explorer package


khamudom:
Move props in fast-dna\packages\fast-component-explorer\app\utilities\configs\data.props.ts
into @microsoft/fast-components-react-msft package

Manan311:
Hello,
I can move the data.props.ts. But before I do I just want to confirm I am moving from packages/fast-component-explorer/app/utilities/configs/data.props.ts     to     fast-dna/packages/fast-components-react-msft/. Is that right?

chrisdholt:
Adding @janechu - @khamudom can you elaborate on the goal here? Removing as good first issue for now until we get more information.

khamudom:
**Description**
The task is to export `data.props.ts` in fast-component-explorer\app\utilities\configs

**Motivation**
In the case, an author would want to create a page similar to the component explorer and use the same data interfaces, we should export the data.props.ts file from a common src file. 

**Proposal**
A suggested place, to export, could be in the fast-toolings-react utilities directory.

janechu:
Closing as this file no longer exists due to the v2 refactor in #3238

PR #2451, opened 2019-11-17
chore: moved data.props.ts to fast-components-react-msft

Manan311:
# Description

## Motivation & context
Issue [#2269](https://github.com/microsoft/fast-dna/issues/2269)
I have moved file data.props.ts from fast-component-explorer directory to fast-components-react-msft directory.

## Issue type checklist

## Process & policy checklist


chrisdholt:
Thanks @Manan311 for the contribution!  I think there's a couple things we'll need to finalize to get this to successfully pass the build. Because the interface is used in the `fast-component-explorer` package, we'll need it exported - that's why the unit tests are failing. Looks like this issue could've used a bit more detail; I likely should have caught that when adding the label.

Curious if @janechu thinks the `/utilities` folder is a good place to export simply the interface here?

janechu:
@chrisdholt I agree
@Manan311 I think this should be moved to `packages/fast-component-explorer/app/utilities/`. Per Chris' suggestion, to pass the build you will need to update the imports since the location of the file will be changing, you should be able to find them mostly in the configs folder but they may also exist elsewhere in this package.

Manan311:
I totally forgot about changing the paths before I made the pr, so I been working on that and I'll update this and add the updated file again

PR #2455, opened 2019-11-19
revert: "fix: improve select aria"

scomea:
Reverts microsoft/fast-dna#2393


scomea:
Reverting because another possibly better solution was proposed and we found an issue with this change in NVDA.

PR #2457, opened 2019-11-19
fix: adjust select Aria

scomea:
# Description
On focus the Select component was not providing desired info to assistive technologies. When a select menu is not open screen readers should recite: the label of the select control, the current value, the component name ("menu button" on Anaheim & Chrome, but "button" on spartan), that it is collapsed.

This is a second approach at this (https://github.com/microsoft/fast-dna/pull/2393);

## Motivation & context
Improve Select component accesibility

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/select/select.tsx
+++ b/packages/fast-components-react-base/src/select/select.tsx
@@ -460,18 +460,15 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (props.multiselectable) {
             return null;
         }
-        const isItemSelected: boolean = state.selectedItemIndex !== 0;
+        const labelledBy: string = `${this.props.labelledBy} ${triggerId}`;
         return (
             <button
                 disabled={props.disabled}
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -466,18 +460,15 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (props.multiselectable) {
             return null;
         }
-        const isItemSelected: boolean = state.selectedItemIndex !== 0;
+        const labelledBy: string = `${this.props.labelledBy} ${triggerId}`;
```
janechu: Can you include a test that ensures that both of these items are part of the labeled by prop? You could do something like:
```
expect(propStringForLabelledBy.split(" ")).toEqual([labelledById, triggerId]);
```

scomea: Added a test along those lines to the trigger aria test cases.

PR #2459, opened 2019-11-20
feat: add display prop to Grid component for SSR support

kim-msft:
# Description

## Motivation & context

Related to: [https://github.com/microsoft/fast-dna/issues/2408](https://github.com/microsoft/fast-dna/issues/2408)

For server side rendering we want to ensure we use the same grid styles that the client would produce. To do this we want to be able to set the grid styles ourselves instead of calling canUseCssGrid() which defaults to false on the server.

This change adds a new prop "display" to the grid component which can be one of "grid" or "-ms-grid". When provided this prop will be used to determine the styles that are used.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-layouts-react/src/grid/grid.no-css-grid.spec.tsx
+++ b/packages/fast-layouts-react/src/grid/grid.no-css-grid.spec.tsx
@@ -84,24 +84,24 @@ describe("Grid - without CSS grid support but `display` prop is `grid`", (): voi
         grid: "grid",
     };
 
-    test("should set an inline style of `display: grid` when CSS grid is NOT supported but the `display` prop passed is `grid`", () => {
-        const rendered: any = shallow(<Grid display={"grid"} />);
+    test("should set an inline style of `display: grid` when CSS grid is NOT supported but the `displayType` prop passed is `grid`", () => {
+        const rendered: any = shallow(<Grid displayType={"grid"} />);
 
         expect(rendered.props().style.display).toEqual("grid");
     });
 
-    test("should set an inline style for `gridColumn` with a value equal to the `gridColumn` prop when  CSS grid is NOT supported but the `display` prop passed is equal to `grid`", () => {
+    test("should set an inline style for `gridColumn` with a value equal to the `gridColumn` prop when  CSS grid is NOT supported but the `displayType` prop passed is equal to `grid`", () => {
         const rendered: any = shallow(
-            <Grid display={"grid"} gridColumn={2} managedClasses={managedClasses} />
+            <Grid displayType={"grid"} gridColumn={2} managedClasses={managedClasses} />
         );
 
         expect(rendered.props().style.gridColumn).toBe(2);
         expect(rendered.props().style.msGridColumn).toBe(undefined);
     });
 
-    test("should set an inline style for `gridRow` with a value equal to the `gridColumn` prop when CSS grid is NOT supported but the `display` prop passed is equal to `grid`", () => {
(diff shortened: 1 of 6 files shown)


packages/fast-layouts-react/src/grid/grid.props.ts
```diff
@@ -69,6 +74,12 @@ export interface GridHandledProps extends GridManagedClasses {
      * Use this property to support -ms-grid when Grid is a child of an element who's display property is set to -ms-grid
      */
     row?: number;
+
+    /**
+     * The CSS grid display, "grid" when true and "-ms-grid" when false
+     * Provide this prop when doing server side rendering
+     */
+    display?: GridDisplay;
```
chrisdholt: I'm torn on the naming convention here a bit. `display` initially causes me to think this relates to whether the grid should be displayed or not - though I do understand we have the type system providing context. I do wonder if a modifier may possibly help here? Perhaps something like `displayType` or something which eludes a bit more to it being related to the display property?

kim-msft: I can see how display is a bit vague, displayType sounds good! I'll make the change

packages/fast-layouts-react/src/grid/grid.props.ts
```diff
@@ -30,6 +30,11 @@ export enum GridAlignment {
  */
 export type GridGutter = 0 | 2 | 3 | 6 | 12;
 
+/**
+ * The CSS grid display style
+ */
+export type GridDisplay = "grid" | "-ms-grid";
```
chrisdholt: @kim-msft can you make sure this is being exported from the `index.ts` file?

kim-msft: Updated - thank you!

packages/fast-layouts-react/src/grid/grid.props.ts
```diff
@@ -69,6 +74,12 @@ export interface GridHandledProps extends GridManagedClasses {
      * Use this property to support -ms-grid when Grid is a child of an element who's display property is set to -ms-grid
      */
     row?: number;
+
+    /**
+     * The CSS grid display, "grid" when true and "-ms-grid" when false
+     * Provide this prop when doing server side rendering
+     */
+    displayType?: GridDisplay;
```
janechu: This seems a little misleading, as what it is is a CSS property name for the grid property. My preference would be:
`cssGridPropertyName`.

chrisdholt: I'm good with this.

kim-msft: sounds good

packages/fast-layouts-react/src/grid/grid.no-css-grid.spec.tsx
```diff
@@ -78,3 +78,85 @@ describe("Grid - without CSS grid support", (): void => {
         expect(rendered.props().style.msGridColumn).toBe(2);
     });
 });
+
+describe("Grid - without CSS grid support but `display` prop is `grid`", (): void => {
```
janechu: This should be updated to `cssGridPropertyName` (same below)

kim-msft: Updated, thanks!

chrisdholt:
Merged - thanks @kim-msft!

PR #2463, opened 2019-11-21
feat: update elevation shadow

bheston:
# Description

Updated elevation shadow calculation to increase top definition on lower levels and darker backgrounds.

Before (top) and after (bottom):
Screen Shot 2019-11-21 at 1 46 58 PM

## Motivation & context

The previous shadows didn't leave much definition on the top edge, making it hard to determine elevation particularly for something like a pane that is docked to the bottom.

Also, shadows against darker backgrounds wash out, so increased the opacity based on the background.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/utilities/elevation.ts
+++ b/packages/fast-components-styles-msft/src/utilities/elevation.ts
@@ -3,13 +3,15 @@ import { CSSRules } from "@microsoft/fast-jss-manager";
 import { DesignSystem, DesignSystemResolver } from "../design-system";
 import { black } from "../utilities/color/color-constants";
 import { parseColorString } from "../utilities/color/common";
-import { ColorRGBA64 } from "@microsoft/fast-colors";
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/utilities/elevation.ts
```diff
@@ -121,20 +125,18 @@ export function elevationShadow(
             yOffsetMultiplier,
             opacity,
             blurMultiplier,
+            blurBase,
```
janechu: is the elevationShadow exposed from this package?

bheston: Yes, looks like all exported functions in this file are exported from the package.

janechu: In that case if anyone is using it, `blurBase` will be `undefined` and `elevationShadow()` will not return the right thing, can you add a default 0 for `blurBase`?

SethDonohue:
Smoked and applying to the components as expected visually.

Issue #2458, opened 2019-11-20
Change menu open functionality in AutoSuggest


eljefe223:
**What feature are you requesting?**
Currently the menu opens on focus, while acceptable according to [ARIA spec](https://w3c.github.io/aria-practices/#combobox) this does not match what design wants or what we see in similar experiences like bing and google. 

> The popup is hidden by default, i.e., the default state is collapsed. The conditions that trigger expansion -- display of the popup --are specific to each implementation. Some possible conditions that trigger expansion include:
> 
> 1. It is displayed when the Down Arrow key is pressed or the Open button is activated. Optionally, if the combobox is editable and contains a string that does not match an allowed value, expansion does not occur.
> 2. It is displayed when the combobox receives focus even if the combobox is editable and empty.
> 3. If the combobox is editable, the popup is displayed only if a certain number of characters are typed in the combobox and those characters match some portion of one of the suggested values.

Suggesting to change the functionality to match the the 3rd option above. Or have some configurable prop to choose the current implementation or the one suggested here

**What would be the benefit of adding this feature?**
This would match other experiences we see elsewhere on the web.

![bing](https://user-images.githubusercontent.com/37851214/69265461-27383000-0b7e-11ea-927c-c779d0af89e5.gif)

janechu:
I would like to see the second suggestion implemented:
> Or have some configurable prop to choose the current implementation or the one suggested here

My opinion is that we include a prop which accepts an enumeration of all of the expansion triggers and provide them as an array, so for example (if all three suggested behaviors are possible):

```js
["keyDownPress", "focus", "characterMatch"]
```

With some implementations adding extra props, perhaps `"characterMatch"` could default to `1`, but AutoSuggest could have another prop such as `characterMatchCount`.

SethDonohue:
I don't see an issue with this change as it will still match ARIA spec.

eljefe223:
I think we should also support filtering as that is related ot the menu being open or not.

> Alternatively, if a combobox supports text input, it is referred to as editable. An editable combobox may either allow users to input any arbitrary value, or it may restrict its value to a discrete set of allowed values, in which case typing input serves to filter suggestions presented in the popup.

That being said I think there are 2 possible scenerios both of which could support filtering. `filtering` could default to false and `menuTrigger` could default to the current implementation `menuTriggers.onFocus`. 

Mark-up suggestion:

**1. On focus.**

[...]

PR #2466, opened 2019-11-24
feat: add filtering and onFocus to autoSuggest

eljefe223:
# Description
Added optional `filterSuggestions` and `menuTrigger` prop to base `autoSuggest`.

Closes: #2458 

ScreenRecording2019-11-23at45951

## Motivation & context
Added functionality to better align with aria spec and user expectations

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
+++ b/packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
@@ -6,6 +6,11 @@ import {
 import { ListboxItemProps } from "../listbox-item";
 import { AutoSuggestState } from "./auto-suggest";
 
+export enum MenuTriggers {
+    onFocus,
+    onKeyDown,
+}
+
 export interface AutoSuggestManagedClasses
     extends ManagedClasses<AutoSuggestClassNameContract> {}
 export interface AutoSuggestUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
@@ -19,7 +24,8 @@ export interface AutoSuggestHandledProps extends AutoSuggestManagedClasses {
         state: AutoSuggestState,
         changeCallback: (event: React.ChangeEvent) => void,
         buttonClickCallback: (event: React.MouseEvent) => void,
-        keyDownCallback: (event: React.KeyboardEvent) => void
+        keyDownCallback: (event: React.KeyboardEvent) => void,
+        focusCallback?: (event: React.FocusEvent) => void
     ) => React.ReactNode;
 
     /**
@@ -69,6 +75,16 @@ export interface AutoSuggestHandledProps extends AutoSuggestManagedClasses {
      * The unique id that will be assigned to the listbox created by this component
      */
     listboxId: string;
+
+    /**
+     * The action that triggers the visibility of the menu
+     */
+    menuTrigger?: MenuTriggers;
+
+    /**
+     * Specifies whether the suggestions dhould filter
+     */
+    filterSuggestions?: boolean;
 }
 
 export type AutoSuggestProps = AutoSuggestHandledProps & AutoSuggestUnhandledProps;
--- a/packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
+++ b/packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
@@ -6,11 +6,6 @@ import {
 import { ListboxItemProps } from "../listbox-item";
 import { AutoSuggestState } from "./auto-suggest";
 
-export enum MenuTriggers {
-    onFocus,
-    onKeyDown,
-}
-
 export interface AutoSuggestManagedClasses
     extends ManagedClasses<AutoSuggestClassNameContract> {}
 export interface AutoSuggestUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
@@ -77,12 +72,7 @@ export interface AutoSuggestHandledProps extends AutoSuggestManagedClasses {
     listboxId: string;
 
     /**
-     * The action that triggers the visibility of the menu
-     */
-    menuTrigger?: MenuTriggers;
-
-    /**
-     * Specifies whether the suggestions dhould filter
+     * Specifies whether the suggestions should filter
      */
     filterSuggestions?: boolean;
 }
--- a/packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
+++ b/packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
@@ -14,6 +14,8 @@ import {
     keyCodeTab,
 } from "@microsoft/fast-web-utilities";
 import { DisplayNamePrefix } from "../utilities";
+import { MenuTriggers } from "./auto-suggest.props";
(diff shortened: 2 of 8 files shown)


packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -442,4 +462,22 @@ describe("auto suggest", (): void => {
 
         document.body.removeChild(container);
     });
+
+    test("intialValue prop applied correctly", (): void => {
```
janechu: nit: initialValue

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -63,6 +68,8 @@ class AutoSuggest extends Foundation<
 
     private shouldFocusMenuOnNextRender: boolean = false;
 
+    private storedSearchString: string;
```
janechu: I think this should be `storedValueString` since AutoSuggest may not be used for searching

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -201,11 +212,41 @@ class AutoSuggest extends Foundation<
...
+        if (this.props.filterSuggestions) {
+            const children: React.ReactNode = this.props.children;
+            return React.Children.map(
+                children,
+                (node: React.ReactElement<any>): React.ReactNode | null => {
+                    if (node.props !== undefined) {
+                        if (node.props[AutoSuggest.valuePropertyKey] === undefined) {
+                            return node;
+                        }
+                        return this.isMatch(node.props) ? node : null;
```
janechu: This does imply children are directly nested in order to be filtered, @scomea can you weigh in on this.

eljefe223: AutoSuggest uses ListBox and ListBoxItem, structurally you need to directly nest ListBoxItem inside Listbox

scomea: The data input for autosuggest is "children", so any operation done in that component will consist of evaluating this.props.children somehow.  I think this is fine for "local typeahead" type implementations like a relatively small list of options like "enter your state".  When the component is used for larger datasets I would not expect filtering to happen in the component but rather upstream before children are created (a controlled scenario).

packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
```diff
@@ -6,6 +6,11 @@ import {
 import { ListboxItemProps } from "../listbox-item";
 import { AutoSuggestState } from "./auto-suggest";
 
+export enum MenuTriggers {
```
scomea: Shouldn't the choice to open the menu on focus be in addition to click and keystrokes?  Maybe this should just be a Boolean?

packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
```diff
@@ -69,6 +75,16 @@ export interface AutoSuggestHandledProps extends AutoSuggestManagedClasses {
...
      */
     listboxId: string;
+
+    /**
+     * The action that triggers the visibility of the menu
+     */
+    menuTrigger?: MenuTriggers;
+
+    /**
+     * Specifies whether the suggestions dhould filter
```
scomea: "should filter"

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -77,6 +84,8 @@ class AutoSuggest extends Foundation<
             focusedItem: null,
             isMenuOpen: this.validateMenuState(false),
         };
+
+        this.storedValueString = this.state.value;
```
scomea: nit: Could just be "this.storedValueString = value"

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -201,11 +212,41 @@ class AutoSuggest extends Foundation<
...
+        } else {
+            return this.props.children;
+        }
+    }
+
+    /**
+     * Determine if a single node is a match
+     */
+    private isMatch(node: ListboxItemProps): boolean {
+        if (this.storedValueString !== undefined) {
```
scomea: nit: I tend to use isNil to cover null + undefined in checks like this.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -214,17 +255,33 @@ class AutoSuggest extends Foundation<
...
+                menuTriggerProps = {
+                    onFocus,
+                    onKeyDown,
+                };
+                break;
+
+            default:
+                menuTriggerProps = {
+                    onClick,
+                    onKeyDown,
```
scomea: You could keep the logic in the render function to a minimum by hooking up all events here and deciding whether or not to act on them when you catch them.  The less logic we have in these customizable functions the better as any change we make needs to be picked up by folks who have rolled their own.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
eljefe223: What about other scenarios? Like character match? This gives us more flexibility to add on functionality in the future.

scomea: What's character match?

scomea: I guess "should the menu open on focus" seems like yes/no option to me.

eljefe223: Character match is another potential "MenuTrigger" where the menu will only open when a certain number of inputed characters match. this is all from the aria spec linked in the issue this PR is resolving.
https://github.com/microsoft/fast-dna/issues/2458

SethDonohue:
This is probably not related to this PR, but should a filtered search of just leading space(s) search  for just spaces as it does with this component, or should they be ignored?

ScreenRecording2019-11-26at21554

packages/fast-components-react-base/src/auto-suggest/auto-suggest.spec.tsx
```diff
@@ -442,4 +462,22 @@ describe("auto suggest", (): void => {
 
         document.body.removeChild(container);
     });
+
+    test("Should render none listbox item when filterSuggestions id set to true", (): void => {
```
SethDonohue: typo on "none" maybe?

packages/fast-components-react-base/src/auto-suggest/auto-suggest.props.ts
eljefe223: Chatted with @scomea offline and decided its best for now to remove `MenuTriggers` from this PR until we discuss in more detail. The filtering / storing the search string was the most important bits of this PR anyway.

PR #2467, opened 2019-11-25
fix: simplify viewport-positioner flicker avoidance logic and remove dom ref

scomea:
# Description
viewport positioner sets it's opacity to 0 until after initial placement is complete.  This change simplifies that logic and removes code that checked that dom during render which was generating a React warning.

## Motivation & context
Looking at the dom during render bad, simpler logic good

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
@@ -160,8 +160,8 @@ class ViewportPositioner extends Foundation<
 
     private openRequestAnimationFrame: number = null;
 
-    private collisionDetector: IntersectionObserver = null;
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -364,6 +357,8 @@ class ViewportPositioner extends Foundation<
             if (this.state.validRefChecksRemaining > 0) {
                 this.setState({
                     validRefChecksRemaining: this.state.validRefChecksRemaining - 1,
+                    initialLayoutComplete:
+                        this.state.validRefChecksRemaining > 1 ? false : true,
```
janechu: for true/false you can just do `initialLayoutComplete: this.state.validRefChecksRemaining <= 1` no need for ternary.

PR #2471, opened 2019-11-27
fix:  horizontal overflow position data error, add test

scomea:
# Description
Was a bug in my last commit to the previous horizontal overflow fix that evaded my tests, so fixed it and added a test.

## Motivation & context
function was returning incorrect result

## Issue type checklist

## Process & policy checklist:

### DIFF
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -305,7 +305,7 @@ describe("horizontal overflow", (): void => {
                     [10, 20, 20, 50, 20],
                     10
                 )
-        ).toBe(-0);
+        ).toBe(0);
 
         expect(
             renderedWithImages
@@ -316,7 +316,7 @@ describe("horizontal overflow", (): void => {
                     [50, 50, 50, 50],
                     100
                 )
-        ).toBe(-50);
+        ).toBe(50);
     });
     test("should an state property `itemsHeight`", () => {
         const renderedWithImages: any = mount(
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -617,8 +617,8 @@ describe("horizontal overflow", (): void => {
             scrollLeft: 0,
         };
         expect(rendered.instance()["getPositionData"]()).toEqual({
-            start: false,
-            end: true,
+            start: true,
+            end: false,
         });
     });
     test("getPositionData returns correct value when scrolled to middle", (): void => {
@@ -634,8 +634,8 @@ describe("horizontal overflow", (): void => {
             scrollLeft: 50,
         };
         expect(rendered.instance()["getPositionData"]()).toEqual({
-            start: true,
-            end: true,
+            start: false,
+            end: false,
         });
     });
     test("getPositionData returns correct value when scrolled to the end", (): void => {
@@ -651,8 +651,25 @@ describe("horizontal overflow", (): void => {
             scrollLeft: 100,
         };
         expect(rendered.instance()["getPositionData"]()).toEqual({
+            start: false,
+            end: true,
+        });
+    });
+    test("getPositionData returns correct value when container does not scroll", (): void => {
+        const rendered: any = mount(
+            <HorizontalOverflow managedClasses={managedClasses}>
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+
+        rendered.instance().horizontalOverflowItemsRef.current = {
+            scrollWidth: 100,
+            clientWidth: 100,
+            scrollLeft: 0,
+        };
+        expect(rendered.instance()["getPositionData"]()).toEqual({
             start: true,
-            end: false,
+            end: true,
         });
     });
 });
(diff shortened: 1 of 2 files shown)


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -280,19 +280,19 @@ class HorizontalOverflow extends Foundation<
...
-        const scrollPositionIsInMiddle: boolean =
-            !scrollPositionIsAtEnd && !scrollPositionIsAtBeginning;
+        const isAtBeginning: boolean = scrollPosition === 0;
+        const isAtEnd: boolean =
+            this.horizontalOverflowItemsRef.current.scrollWidth - scrollPosition ===
+            this.horizontalOverflowItemsRef.current.clientWidth;
 
-        return {
-            start: scrollPositionIsInMiddle ? false : scrollPositionIsAtBeginning,
-            end: scrollPositionIsInMiddle ? false : scrollPositionIsAtEnd,
```
janechu: looks like you just need to change the false here to `true`

scomea: I kinda prefer the changed version, a bit more compact

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -280,19 +280,19 @@ class HorizontalOverflow extends Foundation<
...
-        };
+        if (isAtBeginning) {
+            return { start: false, end: true };
+        } else if (isAtEnd) {
+            return { start: true, end: false };
```
janechu: seems backwards, shouldnt start be `false` and end be `true`? and vice versa above?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -604,5 +604,34 @@ describe("horizontal overflow", (): void => {
 
         (rafThrottle as any) = actualRafThrottle;
     });
+    test("getPositionData returns correct values as scrollLeft value changes", (): void => {
```
janechu: i would break this test out to have:
1. a test to check you are at the beginning
2. a test to check you are at the end
3. a test to check you are in the middle

janechu:
This will end up in the changelog for components, please specify what component you are fixing in your title and merge message when this PR completes.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
scomea: If you are ate the beginning there is nothing to scroll to going left, so it returns false, and there is content to scroll to on the end side, so true.  I agree this terminology is confusing.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
scomea: Actually true, because when you are in the middle and there is overflow you can scroll in either direction.

scomea: Most common use of this is to control visibility of scroll buttons I'd expect, and when you are in the middle you want both because there is content on either side.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
scomea: Sure

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: The documentation for start and end interface is currently as follows:
```
/**
 * Scroll interface for consumers
 * 'start' is when the horizontal overflow scroll is all the way left in LTR (all the way right in RTL)
 * 'end' is when the horizontal overflow scroll is all the right in LTR (all the way left in RTL)
 * TODO #1177: Change interface property names to be more meaningful
 */
```

I would expect then if you are at the beginning start would be `true`, and vice versa for end.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: it seems to me all of the expected values are flipped, see my previous comment on the documentation that dictates what start and end represent.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
janechu: In fact, I'm not sure why this couldn't simple be:
```
return {
    start: isAtStart,
    end: isAtEnd
}
```

scomea: But if you read down this is what we say about overflow start/end.  
 * 'overflowStart' is true when there are items to the left in LTR (right in RTL)
 * 'overflowEnd' is true when there are items to the right in LTR (left in RTL)

So we have two interfaces that report essentially the same information but the meanings of true/false flipped?

scomea: I updated the PR to reflect but I really think we need to revisit this interface on the next version bump, I don't see the value of having both callbacks.

scomea: There is already an issue for deprecating scroll position - https://github.com/microsoft/fast-dna/issues/1178

Issue #2217, opened 2019-09-04
Add server side support to Horizontal Overflow styling


minnieliu:
fast-dna/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx

Currently, horizontal overflow does not throw exception with server side rendering but the styling is off. Would be beneficial to add some workaround (e.g. default props) so that components rendered on the server side can still be styled.

chrisdholt:
Updated the name here - the component technically supports SSR, but I believe the ask is for something that prevents the control from not visually showing up as height is 0. Perhaps a prop for default height could help here?

erhuan-msft:
Default prop should satisfy the current requirement. Thanks!

chrisdholt:
closed with #2472

PR #2472, opened 2019-11-28
fix: update horizontal-overflow default height to null instead of 0px

JamesBurnside:
# Description

Set Horizontal-Overflow default height to unset instead of 0px.

## Motivation & context

#2217
Having 0px can unintentionally hide children.
This is more apparent when server side rendering. The issue is that this.state.itemsHeight is never set to this.getItemMaxHeight();
This is because componentDidMount and itemsLoaded are never triggered when server side rendering and hence itemHeight stays at a default of 0px.

## Issue type checklist

Components depending on this being set to 0px may experience different functionality.

## Process & policy checklist


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -118,7 +118,10 @@ class HorizontalOverflow extends Foundation<
             >
                 <div
                     style={{
-                        height: `${this.state.itemsHeight}px`,
+                        height:
+                            this.state.itemsHeight !== null
+                                ? `${this.state.itemsHeight}px`
+                                : "unset",
```
chrisdholt: I don't believe `unset` is supported whatsoever in IE. https://developer.mozilla.org/en-US/docs/Web/CSS/unset

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -28,7 +28,7 @@ export enum ButtonDirection {
 }
 
 export interface HorizontalOverflowState {
-    itemsHeight: number;
+    itemsHeight?: number;
```
nicholasrice: What is the reason for making `itemsHeight` an optional property? It always exists and is set. Is it more accurate to change to `itemsHeight: number | null` to make clear that `null` is a valid value of this property?

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
nicholasrice: We should validate it works correctly in IE, but I expect it does - the property should be ignored if the value is unrecognized, which *should* have the same effect as "unset". Another option would be to assign `height` to `undefined` and let React filter the property out.

JamesBurnside:
> You mention that having 0px can unintentionally hide children - this concern is specific to rendering server side, correct?

I assume it is visibly an issue on client because itemsHeight is set to a new value on ComponentDidMount (which isn't fired when server side rendering) so it wouldn't really be seen on client. Let me check

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
JamesBurnside: If React filters out undefined then that is likely the better alternative (that have the possibility of IE give a warning/undefined behavior on a property)

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
JamesBurnside: The issue currently is that itemsHeight defaults to 0px as it is expected to be updated by componentDidMount() or by itemsLoaded() callbacks.
However these are not triggered server side so itemsHeight has an incorrect value of 0px (and technically incorrect at points on client side but a non-issue as componentDidMount/itemsLoaded will update the state)
So I was making it optional to indicate that it hadn't been set. An alternative would be to use NaN as the default instead

nicholasrice: It *is* always set though right? Sorry if I'm missing something, but it appears it is always set to a pixel value or `null`, meaning there is never a case where the property *doesn't exist*

JamesBurnside: Ah new to typescript, my bad thought the `?` meant nullable as opposed to optional. Will remove the `?`.

JamesBurnside:
@janechu for review

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -118,7 +118,10 @@ class HorizontalOverflow extends Foundation<
             >
                 <div
                     style={{
-                        height: `${this.state.itemsHeight}px`,
+                        height:
+                            this.state.itemsHeight !== null
+                                ? `${this.state.itemsHeight}px`
+                                : "undefined",
```
janechu: There is no such state as "undefined" for height, I think it should default to "auto" per https://developer.mozilla.org/en-US/docs/Web/CSS/height

JamesBurnside: Sounds good, I'll update to auto now 👍

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
JamesBurnside: Updated to auto per @janechu's comment, resolving this conversation to only have one thread open on the issue

chrisdholt:
Thanks @JamesBurnside for the PR!

Issue #2409, opened 2019-11-06
Add SSR support to Page and Grid components


k1mby:
Both Page and Grid have static "display" which is being calculated using canUseCssGrid(). 

This is always evaluating to use -ms-grid but we want it to reflect whatever the client's browser supports, we can pass the client's browser version to the server to do this check

chrisdholt:
closed with #3517 but I also believe a PR was merged pertaining to this

PR #2473, opened 2019-11-28
feat: Add cssGridPropertyName to Column and Page for Server Sider Rendering

JamesBurnside:
# Description

Add cssGridPropertyName to Column and Page for Server Sider Rendering

## Motivation & context

#2409
To avoid html mismatches we need the server to be able to specify grid vs ms-grid css property.
Similar to https://github.com/microsoft/fast-dna/pull/2459

## Issue type checklist

## Process & policy checklist


packages/fast-layouts-react/src/page/page.props.ts
```diff
@@ -1,4 +1,5 @@
 import React from "react";
+import { GridDisplay } from "../grid/grid.props";
```
chrisdholt: @JamesBurnside @nicholasrice I wonder if we want to move this definition *out* from Grid? This change makes the interface a bit more of a utility specific to the CSS grid property name. It no longer is specific to the grid component implementation, and I think we should likely move it as part of this change.

packages/fast-layouts-react/src/page/page.props.ts
```diff
@@ -14,6 +15,12 @@ export interface PageHandledProps extends PageManagedClasses {
      * The max width of the page
      */
     maxWidth?: string;
+
+    /**
+     * The CSS grid display, "grid" when true and "-ms-grid" when false
```
nicholasrice: I'm not following this comment - what is "true" or "false" here?

Maybe change to "The value assigned to the CSS display property"

packages/fast-layouts-react/src/column/column.tsx
```diff
@@ -170,7 +171,9 @@ export class Column extends Foundation<ColumnHandledProps, ColumnUnhandledProps,
                 : this.props.order;
         }
 
-        const gridStyles: React.CSSProperties = canUseCssGrid()
+        const canUseCssGridStyle: boolean =
+            this.props.cssGridPropertyName === "grid" ? true : canUseCssGrid();
```
nicholasrice: Do you think it's unintuitive that this still has the potential to generate CSS grid properties even if I'm explicitly setting the `cssGridPropertyName` to `"-ms-grid"`. Probably not a common scenario, but I think this should mimic the handled/unhandled pattern of form controls where, if set, `cssGridPropertyName` would always be respected. Thoughts?

packages/fast-layouts-react/src/page/page.props.ts
nicholasrice: I'd be open to moving within this package, but I think it's a bit fragmented of an implementation to put into web-utilities or other packages.

packages/fast-layouts-react/src/page/page.props.ts
JamesBurnside: I'll update to that 👍 (The value assigned to the CSS display property. The property is assigned "grid" when true and "-ms-grid" when false.

packages/fast-layouts-react/src/page/page.props.ts
JamesBurnside: I could move into fast-layouts-react/utilities into its own file of something like "grid-display.ts" or "layout-styles.ts"

packages/fast-layouts-react/src/column/column.tsx
JamesBurnside: Yeah that's a fair point, I'll update to

    const canUseCssGridStyle: boolean =
          this.props.cssGridPropertyName === "grid" ? true :
          this.props.cssGridPropertyName === "-ms-grid" ? false : 
          canUseCssGrid();

packages/fast-layouts-react/src/page/page.props.ts
chrisdholt: Yes, this is what I'm referring to. I think we want it within this package, just not in the grid.props.ts file.

I think for now it's likely fine to even just have an index file in a utils folder until we need to add more:
```
fast-layouts-react/utilities/index.ts
```

nicholasrice:
Thanks for the PR @JamesBurnside

chrisdholt:
Merged - thx @JamesBurnside!

PR #2474, opened 2019-11-28
feat: Add default breakpoint to BreakpointManager

JamesBurnside:
# Description

Add optional default breakpoints that can be used when access to the DOM is not possible (e.g. with server side rendering).

## Motivation & context

Server side rendering cannot control the breakpoint used.
This means it can not emulate the client side window width to produce the same render a client side render would.

## Issue type checklist

Removed default breakpoint manager.

## Process & policy checklist


packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -117,8 +128,9 @@ class BreakpointTracker {
...
         this.openRequestAnimationFrame = true;
-        window.requestAnimationFrame(this.update);
+
+        if (canUseDOM()) {
+            window.requestAnimationFrame(this.update);
+        }
     };
 }
-
-export default new BreakpointTracker();
```
JamesBurnside: This currently removes BreakpointManager default singleton. This was only removed to add defaultBreakpoint as a constructor Param - is there a way to keep this in?

chrisdholt: This change would be breaking.

JamesBurnside: Is there a way to keep in the default constructor (so it doesn't have to be breaking?)

chrisdholt: Is there a reason not to use a public getter and setter as we currently do to enable configuration of breakpoints? That would allow you to customize it for the instance without removing the singleton.

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -56,7 +62,27 @@ class BreakpointTracker {
...
+    }
+
+    /**
+     * Sets the default breakpoint value
+     */
+    public set defaultBreakpoint(breakpoint: Breakpoint) {
+        this._defaultBreakpoint = breakpoint;
+
+        if (!canUseDOM()) {
+            this.breakpoint = this.defaultBreakpoint;
```
JamesBurnside: should this call update() instead here?

JamesBurnside: likewise with `set breakpoints()` function. these both update `this.breakpoint` but do not trigger any `notifySubscribers`. Meaning even if `update()` is called notifySubscribers _**still**_ won't be called which feel wrong.

nicholasrice: Good question - my expectation would be if current breakpoint *changes* then update would get called. Theoretically you could change the default breakpoint and breakpoints array without the active breakpoint changing.

JamesBurnside: `update()` already contains the check `if (this.breakpoint !== breakpoint) { //do work }`
To reduce the redundancy of calculating the new breakpoint I'll update to calling `update()` here and in `set breakpoints`

nicholasrice: Yea I think adding to both is a good call. Thanks @JamesBurnside

JamesBurnside:
@chrisdholt @nicholasrice friendly ping for review

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -32,15 +37,16 @@ class BreakpointTracker {
...
-
-        this.breakpoint = identifyBreakpoint(window.innerWidth, this._breakpoints);
+        this.breakpoint = canUseDOM()
+            ? identifyBreakpoint(window.innerWidth, this._breakpoints)
+            : this.defaultBreakpoint;
 
-        window.addEventListener("resize", this.requestFrame);
+        if (canUseDOM()) {
+            window.addEventListener("resize", this.requestFrame);
+        }
```
janechu: We are using `canUseDOM()` twice, can we store this as a variable?
Either that or do something like:

```typescript
if (canUseDOM()) {
    this.breakpoint = identifyBreakpoint(window.innerWidth, this._breakpoints);
    window.addEventListener("resize", this.requestFrame);
} else {
    this.breakpoint = this.defaultBreakpoint;
}
```

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
```diff
@@ -117,7 +138,10 @@ class BreakpointTracker {
         }
 
         this.openRequestAnimationFrame = true;
-        window.requestAnimationFrame(this.update);
+
+        if (canUseDOM()) {
```
janechu: I don't think this is necessary, we only requestFrame when canUseDOM() is true as it is only assigned in the constructor after that check has already passed.

JamesBurnside: great point, will remove check

packages/fast-layouts-react/src/utilities/breakpoint-tracker.ts
JamesBurnside: Yep I'll update to your if-else, that looks better than storing in a const

Issue #2479, opened 2019-12-04
Add Tree View and Tree View Item as new base components


chrisdholt:
There should be a new components in fast-components-react-base for Tree View (and thus Tree View Items).

<img width="1040" alt="Screen Shot 2019-12-04 at 11 28 15 AM" src="https://user-images.githubusercontent.com/13071055/70174134-34c6dd00-1689-11ea-8e38-94c43ffced7b.png">

As defined by the W3C:
> A tree view widget presents a hierarchical list. Any item in the hierarchy may have child items, and items that have children may be expanded or collapsed to show or hide the children. For example, in a file system navigator that uses a tree view to display folders and files, an item representing a folder can be expanded to reveal the contents of the folder, which may be files, folders, or both.

**Functional requirements:**
- Nodes may or may not have child nodes
- Nodes with child nodes may expand and collapse
- Tree view should support multi-select
- Tree view should enable items to be composed as draggable nodes
- Tree view items should support infinite nesting
- The Tree View should manage focus within its context
- The Tree View should support both controlled and uncontrolled behavior by allowing a user to declare a selected node via props.
- Tree view items should support a slot for glyph content
- Tree view items should support a slot for the expand/collapse glyph

**Interaction model:**
The tree view should align to the interaction model provided by the W3C: https://w3c.github.io/aria-practices/#TreeView

**Implementation examples:**
Flat:
```jsx
<TreeView>
    <TreeViewItem title="Favorites" glyph={favIcon} />
    <TreeViewItem title="Read later" glyph={folderIcon} />
    <TreeViewItem title="Other" glyph={folderIcon} />
</TreeView>
```

With selected defined:
```jsx
<TreeView>
    <TreeViewItem title="Favorites" glyph={favIcon} />
    <TreeViewItem title="Read later" glyph={folderIcon} selected={true} />
    <TreeViewItem title="Other" glyph={folderIcon} />
</TreeView>
```

Nested:
```jsx
<TreeView>
    <TreeViewItem title="Root node & parent node" glyph={favIcon}>
        <TreeViewItem title="Child node" glyph={folderIcon} />
        <TreeViewItem title="Child node" glyph={folderIcon} />
        <TreeViewItem title="Child node & Parent node" glyph={folderIcon}>
            <TreeViewItem title="Child node" glyph={folderIcon} />
            <TreeViewItem title="Child node" glyph={folderIcon} />
        </TreeViewItem>
    </TreeViewItem>
[...]

PR #2481, opened 2019-12-04
feat: add new utilities for checking HTML elements

chrisdholt:
# Description
This PR adds new utilities for checking valid HTMLElements and checking when those elements are in view.

Adding these utilities outside the context of #2479 though they will be useful for that task.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-web-utilities/src/dom.ts
+++ b/packages/fast-web-utilities/src/dom.ts
@@ -2,6 +2,28 @@ import { canUseDOM } from "exenv-es6";
 import { isBoolean } from "lodash-es";
 
 /**
+ * A test that ensures that all arguments are HTML Elements
+ */
+export function isHTMLElement(...args: unknown[]): boolean {
+    return args.every((arg: unknown) => arg instanceof HTMLElement);
+}
+
+/**
+ * Returns all displayed elements inside of a root node that match a provided selector
+ */
+export function getDisplayedNodes(rootNode: Element, selector: string): HTMLElement[] {
+    if (!isHTMLElement(rootNode)) {
+        return;
+    }
+
+    const nodes: HTMLElement[] = Array.from(rootNode.querySelectorAll(selector));
+
+    // offsetParent will be null if the element isn't currently displayed,
+    // so this will allow us to operate only on visible nodes
+    return nodes.filter((node: HTMLElement) => node.offsetParent !== null);
+}
+
+/**
  * Gets the numeric key code associated with a keyboard event. This method is for use with DOM level 3 events
  * that still use the deprecated keyCode property.
  */
--- a/packages/fast-web-utilities/src/dom.ts
+++ b/packages/fast-web-utilities/src/dom.ts
@@ -4,14 +4,17 @@ import { isBoolean } from "lodash-es";
 /**
  * A test that ensures that all arguments are HTML Elements
  */
-export function isHTMLElement(...args: unknown[]): boolean {
-    return args.every((arg: unknown) => arg instanceof HTMLElement);
+export function isHTMLElement(...args: HTMLElement[]): boolean {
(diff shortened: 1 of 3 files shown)


packages/fast-web-utilities/src/dom.ts
```diff
@@ -1,6 +1,28 @@
 import { canUseDOM } from "exenv-es6";
 import { isBoolean } from "lodash-es";
 
+/**
+ * A test that ensures that all arguments are HTML Elements
+ */
+export function isHTMLElement(...args: unknown[]): boolean {
```
nicholasrice: Is it possible to make this a type guard so the type system knows these are HTMLElements?

chrisdholt: I can update to `HTMLElement[]`

packages/fast-web-utilities/src/dom.ts
```diff
@@ -1,6 +1,31 @@
...
+    return args.every((arg: HTMLElement) => arg instanceof HTMLElement);
+}
+
+/**
+ * Returns all displayed elements inside of a root node that match a provided selector
+ */
+export function getDisplayedNodes(
+    rootNode: HTMLElement,
+    selector: string
+): HTMLElement[] {
```
janechu: return value should be `HTMLElement[] | void`

packages/fast-web-utilities/src/dom.ts
```diff
@@ -1,6 +1,31 @@
...
+    if (!isHTMLElement(rootNode)) {
+        return;
+    }
+
+    const nodes: HTMLElement[] = Array.from(rootNode.querySelectorAll(selector));
+
+    // offsetParent will be null if the element isn't currently displayed,
+    // so this will allow us to operate only on visible nodes
+    return nodes.filter((node: HTMLElement) => node.offsetParent !== null);
```
janechu: is there a potential issue if the parent of a parent is causing this to be not visible?

packages/fast-web-utilities/src/dom.ts
```diff
@@ -1,6 +1,31 @@
...
+ * A test that ensures that all arguments are HTML Elements
+ */
+export function isHTMLElement(...args: HTMLElement[]): boolean {
+    return args.every((arg: HTMLElement) => arg instanceof HTMLElement);
+}
+
+/**
+ * Returns all displayed elements inside of a root node that match a provided selector
+ */
+export function getDisplayedNodes(
```
janechu: Do you mean rendered instead of displayed?

nicholasrice: No - this is specifically asserting the display property of HTML nodes. This approach is taken to determine if nodes are displayed because the alternative is to get the computed styles of the element, which forces a layout pass (bad for perf)

https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom

chrisdholt: Oops, was going to respond with a link here as well: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent

packages/fast-web-utilities/src/dom.ts
```diff
@@ -1,6 +1,31 @@
 import { canUseDOM } from "exenv-es6";
 import { isBoolean } from "lodash-es";
 
+/**
+ * A test that ensures that all arguments are HTML Elements
+ */
+export function isHTMLElement(...args: HTMLElement[]): boolean {
```
nicholasrice: sorry - we don't want to assert these are `HTMLElement` because the point of the fn is to *determine* if they're HTML element types.

Just tried it locally, doesn't work. This should probably just be `any[]` instead of `unknown[]` since we actually don't care what the types are.

PR #2491, opened 2019-12-06
feat: performance enhancement when the JSON schema has deeply nested oneOfs

### LINKED ISSUE #2304: [Form] Performance issues seem to occur on deeply nested oneOfs
There seems to be a performance hit on deeply nested oneOfs in a JSON schema.

This should be stress tested to see what improvements can be made so that no matter how deep the nesting goes, the location is stored and does not need to be re-fetched.

This will need to be judged against whether the form is controlled or uncontrolled and whether the data is changing from other sources.

fmamic:
[Form] Performance issues seem to occur on deeply nested oneOfs

# Description

There seems to be a performance hit on deeply nested oneOfs in a JSON schema.

This should be stress tested to see what improvements can be made so that no matter how deep the nesting goes, the location is stored and does not need to be re-fetched.

This will need to be judged against whether the form is controlled or uncontrolled and whether the data is changing from other sources.

## Motivation & context

User experience improvement.

Closes #2304

## Issue type checklist

## Process & policy checklist

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/en/contributing/standards) for this project.


packages/fast-tooling-react/package.json
```diff
@@ -66,6 +66,7 @@
     "@babel/preset-env": "^7.1.0",
     "@babel/preset-react": "^7.0.0",
     "@microsoft/fast-tslint-rules": "^3.0.4",
+    "@types/fast-json-stable-stringify": "^2.0.0",
```
janechu: I think we can remove this, we will want our package to break if `ajv` removes this as a dependency, but it will pass if we leave this here.

packages/fast-tooling-react/src/utilities/ajv-validation.ts
```diff
@@ -0,0 +1,83 @@
...
+    } else {
+        ajvDictionary[schemaRefKey] = new ajv({ schemaId: "auto", allErrors: true });
+        validateFunctionDictionary[schemaRefKey] = ajvDictionary[schemaRefKey].compile(
+            schema
+        );
+        dataValidationResult = validateFunctionDictionary[schemaRefKey](data) || false;
+    }
+
+    return dataValidationResult;
+}
```
janechu: nit: no need to store the result, you could just do:

```ts
const schemaRefKey: string = stringify(schema);

if (!validateFunctionDictionary[schemaRefKey]) {
    ajvDictionary[schemaRefKey] = new ajv({ schemaId: "auto", allErrors: true });
    validateFunctionDictionary[schemaRefKey] = ajvDictionary[schemaRefKey].compile(
        schema
    );
}

return validateFunctionDictionary[schemaRefKey](data) || false;
```

janechu:
Just a couple notes, otherwise this looks really good 👍

packages/fast-tooling-react/package.json
fmamic: fast-json-stable-stringify dependency doesn't come with the typescript types so this one is just to enable safe typing while using this dependency. It is in the dev dependency list so it won't affect production bundle. I didn't include fast-json-stable-stringify itself and code will break if ajv removes it.

packages/fast-tooling-react/src/utilities/ajv-validation.ts
fmamic: Nice thanks :)

PR #2498, opened 2019-12-08
fix: modal dialogs do not move focus if already within

scomea:
# Description
Modal dialogs were always focusing to the top of the tab queue on mount which was interfering with the autofocus option of other components.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2495

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/dialog/dialog.spec.tsx
+++ b/packages/fast-components-react-base/src/dialog/dialog.spec.tsx
@@ -18,9 +18,7 @@ import { DisplayNamePrefix } from "../utilities";
  */
 configure({ adapter: new Adapter() });
 
-const container: HTMLDivElement = document.createElement("div");
-document.body.appendChild(container);
-
+/* tslint:disable:no-string-literal */
 describe("dialog", (): void => {
     const managedClasses: DialogClassNameContract = {
         dialog: "dialog-class",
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -231,18 +233,28 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
...
             event.preventDefault();
         }
     };
 
+    /**
+     * test to avoid forcing focus when focus is already within
+     */
+    private shouldForceFocus = (currentFocusElement: Element): boolean => {
+        if (
+            this.props.visible &&
```
nicholasrice: This conditional evaluates to true/false, so you can reduce this to just return the product of the conditional:

```
return this.props.visible && this.rootElement.current instanceof HTMLElement && !(this.rootElement.current as HTMLElement).contains(currentFocusElement)

File: packages/fast-components-react-base/src/dialog/dialog.spec.tsx [hunk 2]
```diff
@@ -243,4 +241,26 @@ describe("dialog", (): void => {
         expect(mockRemoveListenerFn).toHaveBeenCalledTimes(2);
         expect(mockRemoveListenerFn.mock.calls[1][0]).toBe("focusin");
     });
+    test("shouldForceFocus correctly detects whether the prodided element is in or out of dialog", () => {
```
janechu: nit: provided* i would update the wording here to something like "should detect whether the provided element is in or out of dialog"

packages/fast-components-react-base/src/dialog/dialog.spec.tsx
```diff
@@ -243,4 +241,26 @@ describe("dialog", (): void => {
...
+                </Dialog>
+                <button id="outerButton" />
+            </div>,
+            { attachTo: container }
+        );
+
+        const dialog: any = rendered.find("BaseDialog");
+        const innerButton: Element = document.getElementById("innerButton");
+        const outerButton: Element = document.getElementById("outerButton");
+        expect(dialog.instance()["shouldForceFocus"](innerButton)).toEqual(false);
```
connorjsmith: nit: I think we can avoid the tslint disable if we do something like
```typescript
const instance: Dialog = dialog.instance();
expect(instance.shouldForceFocus(innerButton)).toEqual(false);
expect(instance.shouldForceFocus(outerButton)).toEqual(true);
```

packages/fast-components-react-base/src/dialog/dialog.spec.tsx
```diff
@@ -18,9 +18,7 @@ import { DisplayNamePrefix } from "../utilities";
  */
 configure({ adapter: new Adapter() });
 
-const container: HTMLDivElement = document.createElement("div");
-document.body.appendChild(container);
-
+/* tslint:disable:no-string-literal */
```
connorjsmith: Can we disable this only for the method that needs it?

PR #2503, opened 2019-12-10
fix: Change dismissals on pointer down instead of click

prateekchandan:
# Description

In this PR we change the following behavior:
1. We changed the click handling to pointer down. The flyouts will get dismissed on pointer down and not pointer up. This will ensure, that while flyout is open, and the user starts interacting outside the flyout where the events are (pointer down -> move -> up) which will not generate a click event, will also dismiss the flyout
2. I changed mouse to pointer events, so that we are dismissing the flyout from mouse interactions as well as pen interactions.

## Motivation & context

Motivation: For inking scenarios in PDF, we are using this component. Now there is a user scenario that user selects some inking settings and then if he starts inking without first dismissing the flyout, he will be ale to ink, but flyout will be taking some of the area. And now once he finishes drawing first stroke, then the flyout goes away.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/flyout/flyout.spec.tsx
+++ b/packages/fast-components-react-msft/src/flyout/flyout.spec.tsx
@@ -113,7 +113,7 @@ describe("flyout", (): void => {
         ).toEqual(describedby);
     });
 
-    test("should call the `onDismiss` callback after a click event on the window when `visible` prop is true", () => {
+    test("should call the `onDismiss` callback after a pointer down event on the window when `visible` prop is true", () => {
         const onDismiss: any = jest.fn();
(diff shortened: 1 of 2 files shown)


chrisdholt:
It's been 17 days without feedback on this, so I'm going to close this PR. @prateekchandan if you still feel strongly about this change needing to be made, feel free to create a larger [feature request](https://github.com/microsoft/fast-dna/issues/new?assignees=&labels=&template=request-a-feature.md&title=Add+%5Bwhat%5D+to%2Fin+%5Bwhere%5D) detailing how this might be done across all components as well as the benefit to components with this change.

Issue #2482, opened 2019-12-05
Create a messaging system


janechu:
# Abstract

Create a messaging system outlined in [this document](https://github.com/microsoft/fast-dna/wiki/Tooling:-Messaging-system).

The messaging system will be used by various components in the tooling, which will necessitate breaking changes, see:

- #2483 
- #2484

## Requirements

- Components should be able to register
- Components should be able to deregister
- Components should be able to subscribed to different types
- Components should be able to subscribe to different actions

Refer to the above document for the types and actions that should be available.

janechu:
Closed when #2919 merged the changes to master.

PR #2504, opened 2019-12-10
feat: add message system

janechu:
# Description

This is the initial message system which registers and de-registers components.

## Motivation & context

Begins addressing #2482

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/message-system/index.ts
+++ b/packages/fast-tooling-react/src/message-system/index.ts
@@ -2,6 +2,7 @@
 
 import {
     MessageSystemComponentTypeAction,
+    MessageSystemData,
     MessageSystemType,
 } from "./message-system.props";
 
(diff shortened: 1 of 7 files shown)


packages/fast-tooling-react/src/message-system/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * The main purpose of this is to tie together
+ * process heavy actions onto a separate thread,
+ * as well as to allow components to opt into a
+ * single source for data updates.
+ */
+
+const registeredComponents: object = {};
+
+onmessage = function(e: MessageEvent): void {
```
nicholasrice: no `const` / `let` / `var`? Is this some webworker syntax?

janechu: it's a bit odd, but it's declared in the global scope see: https://github.com/microsoft/TypeScript/blob/master/lib/lib.webworker.d.ts#L5768

this is a plain javascript example but here it is implemented simply: https://github.com/mdn/simple-web-worker/blob/gh-pages/worker.js

PR #2505, opened 2019-12-10
feat: add initialization with data to message system

janechu:
# Description

Adds initialization message to be passed to the message system which initializes data.

## Motivation & context

The data will serve as a single source to be used among the various tooling components. As such, the message system will handle it.

Adds to #2482

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/message-system/message-system.props.ts
+++ b/packages/fast-tooling-react/src/message-system/message-system.props.ts
@@ -1,21 +1,41 @@
 export enum MessageSystemComponentTypeAction {
     register = "register",
     deregister = "deregister",
+    registered = "registered",
+    deregistered = "deregistered",
 }
 
 export enum MessageSystemDataTypeAction {
     update = "update",
     move = "move",
     duplicate = "duplicate",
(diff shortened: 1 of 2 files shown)


packages/fast-tooling-react/src/message-system/message-system.props.ts
```diff
@@ -1,23 +1,43 @@
 export enum MessageSystemComponentTypeAction {
     register = "register",
     deregister = "deregister",
+    registered = "registered",
```
nicholasrice: I'm not quite following the naming here - "registered", "deregistered", and "duplicated" seem more like states or event types than actions. Maybe there is a way to organize these differently? You call - just caught me as inconsistent.

janechu: I'll take a look on my next pass to see if I can clarify, but the outgoing messages are all notifications that the action has been completed with any extra parameters that should be passed. So you might have an incoming message to "register" the component, and an outgoing message confirming that the component has been "registered".

PR #2510, opened 2019-12-11
feat: add tree view as a new component

### LINKED ISSUE #2479: Add Tree View and Tree View Item as new base components

chrisdholt:
# Description
closes #2479 

## Issue type checklist

## Process & policy checklist

- [] I have updated the project documentation to reflect my changes.


packages/fast-component-explorer/app/utilities/plugins/index.ts
```diff
@@ -14,6 +14,8 @@ const initializedPlugins: Array<Plugin<PluginProps>> = [
             "@microsoft/fast-components-react-msft/select-option/glyph",
             "@microsoft/fast-components-react-msft/text-action/button",
             "@microsoft/fast-components-react-msft/text-action/beforeGlyph",
+            "@microsoft/fast-components-react-msft/tree-view-item/beforeContent",
+            "@microsoft/fast-components-react-msft/tree-view-item/beforeContent",
```
janechu: same key being used here

packages/fast-components-class-name-contracts-msft/src/index.ts
```diff
@@ -24,3 +24,4 @@ export * from "./slider-label";
 export * from "./subheading";
 export * from "./text-action";
 export * from "./text-field";
+export * from "./tree-view-item";
```
janechu: no tree view class name contract?

packages/fast-components-react-base/src/tree-view-item/README.md
```diff
@@ -0,0 +1,5 @@
+# Tree View Item
+A TreeViewItem is an item in a TreeView. It can contain nested child items and be selected.
```
janechu: nit: and *can* be selected

packages/fast-components-react-base/src/tree-view/README.md
```diff
@@ -0,0 +1,5 @@
+# Tree View
+A TreeView displays a hierarchical set of information like an outline or folder structure. It is populated with one or more `TreeViewItems`, which may further contain child items. An item that has children can be expanded or collapsed. An item can also be selected, typically supported to present related information in an adjacent panel.
```
janechu: nit: prefer "...~like~ such as an outline or folder structure."

packages/fast-components-react-base/src/tree-view-item/tree-view-item.props.ts
```diff
@@ -0,0 +1,60 @@
...
+     */
+    expandCollapseGlyph?: (className?: string) => React.ReactNode;
+
+    /**
+     * The expanded state of the tree view item on mount.
+     * After mounting, the tree view item will track the expanded
+     * state internally to align to accessibility guidelines.
+     * This property will have no effect if the item has no child nodes
+     */
+    defaultExpanded?: boolean;
```
scomea: Do we need an expanded prop for controlled mode?

packages/fast-components-react-base/src/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,276 @@
...
+            }
+        }
+    }
+
+    private handleKeyDown = (e: React.KeyboardEvent<HTMLDivElement>): void => {
+        let nodes: HTMLElement[] | void;
+
+        if (canUseDOM()) {
+            switch (e.keyCode) {
+                case keyCodeHome:
```
scomea: Prevent default on key events I think.

packages/fast-component-explorer/app/utilities/plugins/index.ts
chrisdholt: oops!

packages/fast-components-class-name-contracts-msft/src/index.ts
chrisdholt: We typically don't dupe all contracts here - there's not a MSFT specific tree view so it simply lives in base.

packages/fast-components-react-base/src/tree-view-item/tree-view-item.props.ts
chrisdholt: This is technically the `expanded` property, so we could likely just call it `expanded` instead of default expanded.

packages/fast-components-react-base/src/tree-view/tree-view.tsx
chrisdholt: Do we need to prevent default on *all* potential key events? I'm not sure we do

connorjsmith:
When we were designing the treeview, we had a few requirements:
1. When an item becomes selected, it should be revealed in the tree if it isn't already (ie. if it has a collapsed ancestor(s), they should become expanded if a descendant becomes selected). This is the primary use case of the `onSelected` prop.
2. When an item is invoked (either with keyboard or mouse), the owner of the component should be aware
3. Easy to generate complex trees in a loop


## Issues with current implementation

The current implementation combines 1 and 2, which recently was discovered to be an issue within one of our pages. Effectively we were doing something along the lines of
```typescript
const selected = /* true iff we are on a certain route exactly */

function onSelected() {
  if (selected) { return; } // we don't want to navigate if the pivot is already selected
  this.props.history.push(/* new route */); // Uses react-router
}

<TreeViewItem
  onSelected={onSelected}
  selected={selected}
/>
```

The issue here is that for sub-routes (ie. /foo/bar is a sub-route of /foo), navigating from a route to an unrelated sub-route (ie. /baz -> /foo/bar) triggers the onSelected function and navigates to /foo instead.

## Proposed fix

We separate the onSelected listener into an `onInvoke` and `onSelected`. We need to keep `onSelected` as a way of expanding the tree (goal 1 above), but most users should instead use `onInvoke` to detect activation of the item. This `onInvoke` does not yet exist in the current implementation, but should should be relatively straight forward to achieve. To preserve backwards compatibility, we can still call the `onSelected` handler if no `onInvoke` is provided.

In short:
* `onInvoke` provided => `onInvoke` called on mouse/keyboard interactions. `onSelected` called for `componentDidUpdate` changes.
	* For the scenario above, we would *not* provided an `onSelected` callback when instantiating a `TreeViewItem`.
* `onInvoke` not provided => `onSelected` called for mouse/keyboard interactions and `componentDidUpdate` for `props.selected` changing. This is the behavior we have today.

chrisdholt:
Thanks @connorjsmith - I just did some work to manage the expanded state if instantiated with a selected child as part of the component. I think this works with the intent of the `defaultExpanded` prop - if you tell a certain portion of the tree to be collapsed, but a child is selected we're going to render it in an expanded state to ensure an accessible experience. Since I believe this is called out as part of the spec, I think it makes sense for the component to handle this. 

Is there another scenario you can image where we would want to keep onSelected? In the example above, it seems like onSelected isn't completely accurate of the new behavior with us handling the expanded state - selected has not changed, but the expanded state of a given item has. If that's the primary scenario you can think of for the onSelected prop, I wonder if we should look at a more semantic name related to the expanded property changing?

packages/fast-components-react-base/src/tree-view-item/tree-view-item.props.ts
chrisdholt: @scomea, after sleeping on this - I'm not sure that `expanded` would be *as* accurate for this control as `defaultExpanded`. Essentially, the expanded state is not *explicitly* controlled here. 

I could set expanded on a `treeitem`, but if it does not have childNodes, it would not (nor do I think it should) actually behave as expanded. If children were dynamically added after the initial render, it would default to the `defaultExpanded` prop.

Similarly, if I wanted to default a given parent node to NOT be expanded, but a child node within the tree were `selected` on mount, the parent node(s) above that child should expand for accessibility concerns. 

With the above scenarios, I feel like `defaultExpanded` provides a better API as it is more informative of the actual behavior of the tree. We'll honor that so far as we can given the accessible requirements. As a user, I would expect `expanded` to be more definitive, whereas `defaultExpanded` seems to convey the potential for it to *not be the case*.

connorjsmith:
Yeah, we can probably simulate `onSelected` with a combination of `onInvoke` and `onExpandedChange`.

Just to confirm, the expected use case is:

* On initial mount, whatever items should be expanded by default will see `defaultExpanded=true`. These will be expanded even if none of their descendants are selected.
* If any items in the tree are `selected=true`, all of their ancestors will be expanded regardless of their `defaultExpanded` prop on initial mount.
* If a user chooses to collapse one of the ancestors of a selected item, then the selected item will "disappear", but remain selected.
* If a descendant of that manually collapsed item is selected again, all ancestors will become expanded again, overriding the manual collapsing.

chrisdholt:
@connorjsmith 
> Yeah, we can probably simulate `onSelected` with a combination of `onInvoke` and `onExpandedChange`.
> 
> Just to confirm, the expected use case is:
> 
> * On initial mount, whatever items should be expanded by default will see `defaultExpanded=true`. These will be expanded even if none of their descendants are selected.
Correct, currently supported behavior; one caveat is that if there are *no* descendants, it will not be expanded.
> * If any items in the tree are `selected=true`, all of their ancestors will be expanded regardless of their `defaultExpanded` prop on initial mount.
Correct, currently supported in this PR. I think this gives weight to why we are currently moving ahead with `defaultExpanded` rather than expanded.
> * If a user chooses to collapse one of the ancestors of a selected item, then the selected item will "disappear", but remain selected.
I think this is currently what will happen - do you have any thoughts on this?
> * If a descendant of that manually collapsed item is selected again, all ancestors will become expanded again, overriding the manual collapsing.
I think with the above behaviors this is what I would *expect* to happen. The functionality needs to be added to this PR though to check and set expanded as part of an update cycle if the selected prop has changed.

packages/fast-components-styles-msft/src/tree-view-item/index.ts
```diff
@@ -0,0 +1,228 @@
...
+    horizontalSpacing,
+} from "../utilities/density";
+import {
+    applyScaledLineHeight,
+    applyScaledTypeRamp,
+    getScaledLineHeight,
+} from "../utilities/typography";
+import { applyFontWeightNormal } from "../utilities/fonts";
+
+const nestedNodeOffest: (designSystem?: DesignSystem) => string = toPx(
```
bheston: Seems like this needs to be typed as DesignSystemResolver<string>. If the designSystem is optional, the default values will be used, so any design system changes won't apply.

packages/fast-components-styles-msft/src/tree-view-item/index.ts
```diff
@@ -0,0 +1,228 @@
...
+        display: "flex",
+        "align-items": "center",
+        "white-space": "nowrap",
+        ...applyScaledTypeRamp("t7"),
+        ...applyFontWeightNormal(),
+    },
+    // Container for all child nodes of an item
+    treeViewItem_childNodeRegion: {
+        display: "none",
+        "font-size": `calc(1em + ${nestedNodeOffest()})`,
```
bheston: Per previous comment, this needs to be `format()` so the design system is passed.

packages/fast-components-react-base/src/tree-view-item/tree-view-item.tsx
```diff
@@ -0,0 +1,491 @@
...
+            return null;
+        }
+
+        return !!currentNode.querySelector("[aria-selected='true'");
+    }
+
+    /**
+     * Handles when the left arrow is pressed
+     */
+    private handleArrowLeft(): void {
```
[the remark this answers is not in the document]
nicholasrice: Good question - we don't have a loc bug on it currently but I would expect it would

packages/fast-components-react-base/src/tree-view/tree-view.tsx
nicholasrice: If this is currently working as expected I would hesitate to change it.

packages/fast-components-react-base/src/tree-view-item/tree-view-item.tsx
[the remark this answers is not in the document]
chrisdholt: I'm not sure - The spec details the following:

> Right arrow:
- When focus is on a closed node, opens the node; focus does not move.
- When focus is on a open node, moves focus to the first child node.
- When focus is on an end node, does nothing.

> Left arrow:
- When focus is on an open node, closes the node.
- When focus is on a child node that is also either an end node or a closed node, moves focus to its parent node.
- When focus is on a root node that is also either an end node or a closed node, does nothing.

I'd prefer to ask this question of the standards body and follow-up rather than assume here.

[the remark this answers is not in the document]
nicholasrice: Yea, I think we can take it as a follow-up

packages/fast-components-react-base/src/tree-view/tree-view.tsx
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * Simple state machine to track how many nodes have nested nodes
+ */
+class ChildNodeCountTracker {
+    private _count: number = 0;
+    public get count(): number {
+        return this._count;
+    }
+
+    public adjust(delta: number): void {
```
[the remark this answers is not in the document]
chrisdholt: We're adding the delta now. I think the primary reason we didn't see any issues was that we weren't un-mounting nodes which would change the nested depth - I think I was only checking for adding nodes, so += 1 is fine for that. Things seem to be working as expected with this change.

PR #2531, opened 2019-12-16
feat: add default breakpoint for fast-layouts-react column

JamesBurnside:
# Description

Add default breakpoint to the column component fast-layouts-react

## Motivation & context

When performing server side rendering there is no way to set the column width,span&order.
#2398 

## Issue type checklist

## Process & policy checklist


packages/fast-layouts-react/src/column/column.spec.tsx
```diff
@@ -241,6 +242,53 @@ describe("Column", (): void => {
...
+
+    test("column should use a default breakpoint of zero when the DOM is unavailable and default breakpoint is unset", () => {
+        // make DOM unavailable for test
+        const windowSpy: jest.SpyInstance<any> = jest.spyOn(
+            global as any,
+            "window",
+            "get"
+        );
+        windowSpy.mockImplementation(() => ({ undefined }));
+        expect(canUseDOM()).toEqual(false);
```
chrisdholt: We had some issues with tests failing with this implementation. Can you check out the breakpoint tracker test updates in this commit and see if we can leverage that pattern here? https://github.com/microsoft/fast-dna/commit/25633f6e63c69a565fb65e654607b6a87b7fe9d2

chrisdholt:
@JamesBurnside Just an FYI that I updated your PR message - thanks!

packages/fast-layouts-react/src/column/column.spec.tsx
JamesBurnside: That's much cleaner as well, yep I'll update

packages/fast-layouts-react/src/column/column.spec.tsx
```diff
@@ -241,9 +252,50 @@ describe("Column", (): void => {
...
+        const rendered: any = shallow(
+            <Column
+                defaultBreakpoint={2}
+                span={[7, 8, 9]}
+                row={[10, 11, 12]}
+                order={[13, 14, 15]}
+            />
+        );
+
+        // when breakpoint is zero, GridColumn, GridRowStart and order styles should use the default breakpoint value
```
chrisdholt:
```suggestion
        // when the DOM is unavailable, GridColumn, GridRowStart and order styles should use the default breakpoint value
```

Issue #2494, opened 2019-12-06
Progress is width 100% even when circular=true


adkaufm:
Progress still has width 100% even when it's the circular version. Can we not

chrisdholt:
@eljefe223 can you take a look at this? I don't think this should be required to be wrapped in a containing div just to properly size it. (@bheston as FYI)

eljefe223:
@adkaufm How is width 100% negatively impacting your use case? Just trying to understand what the expectation is. Even if we remove `width: 100%` it will default to that same value.

adkaufm:
The issue is that the containing div is width 100%. So if I place the progress in a larger div, it takes up the full width but the progress is left aligned. So annoying to have to add a jss stylesheet just to center it, when it shouldn't be 100% in the first place.

adkaufm:
![image](https://user-images.githubusercontent.com/44783604/71026957-20d69e80-20bf-11ea-9a27-7bb958880f46.png)
why

eljefe223:
Just implemented a progress in a prototype real quick. I see the `width` issue, but as an fyi there is no need to use `jssStyleSheet` in the current implementation you can apply the style directly to the `<Progress/>` itself.

https://stackblitz.com/edit/progress-circular

PR #2537, opened 2019-12-19
fix: circular progress is 100% width

eljefe223:
# Description
This fix changes the outer most div display to be `inline` instead of `block`.

Closes: #2494 

**After:**
Screen Shot 2019-12-19 at 8 28 30 AM

## Motivation & context
This allows the circular `progress` to be easily positioned within UI without any style overrides

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/progress/index.ts
+++ b/packages/fast-components-styles-msft/src/progress/index.ts
@@ -29,6 +29,8 @@ const styles: ComponentStyles<ProgressClassNameContract, DesignSystem> = {
     },
     progress__circular: {
         height: "unset",
+        width: "unset",
+        display: "inline-block",
         "& $progress_valueIndicator, & $progress_indicator": {
             fill: "none",
             "stroke-width": "2px",


chrisdholt:
@eljefe223 Question: With `display: inline` we can run into an issue where user-defined width, height, and padding are not respected without changing the display value. I don't think this would be an issue with `display: inline-block` (or `display: inline-flex` FWIW) - is there a reason to opt for inline over inline-block?

eljefe223:
> @eljefe223 Question: With `display: inline` we can run into an issue where user-defined width, height, and padding are not respected without changing the display value. I don't think this would be an issue with `display: inline-block` (or `display: inline-flex` FWIW) - is there a reason to opt for inline over inline-block?

Sorry, I may not be following, but I did use `display: inline-block`.

chrisdholt:
> > @eljefe223 Question: With `display: inline` we can run into an issue where user-defined width, height, and padding are not respected without changing the display value. I don't think this would be an issue with `display: inline-block` (or `display: inline-flex` FWIW) - is there a reason to opt for inline over inline-block?
> 
> Sorry, I may not be following, but I did use `display: inline-block`.

I must have misread it. Thanks for the quick follow up!

PR #2538, opened 2019-12-24
fix: viewport positioner resets on anchor change

scomea:
# Description
Viewport positioner was not resetting on anchor change.  This fix causes it to check for anchor changes and reposition if required.

## Motivation & context
Users had encountered issue with positioner not reacting to anchor change.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
@@ -189,33 +189,7 @@ class ViewportPositioner extends Foundation<
     constructor(props: ViewportPositionerProps) {
         super(props);
 
-        this.state = {
-            disabled: true,
-            noObserverMode: false,
-            xTransformOrigin: Location.left,
-            yTransformOrigin: Location.top,
-            xTranslate: 0,
-            yTranslate: 0,
-            top: null,
-            right: null,
-            bottom: null,
-            left: null,
-            currentHorizontalPosition:
-                ViewportPositionerHorizontalPositionLabel.undefined,
-            currentVerticalPosition: ViewportPositionerVerticalPositionLabel.undefined,
-            defaultHorizontalPosition: this.getHorizontalPositionToLabel(
-                this.props.horizontalPositioningMode,
-                this.props.defaultHorizontalPosition
-            ),
-            defaultVerticalPosition: this.getVerticalPositionToLabel(
-                this.props.verticalPositioningMode,
-                this.props.defaultVerticalPosition
-            ),
-            horizontalSelectedPositionWidth: null,
-            verticalSelectedPositionHeight: null,
-            initialLayoutComplete: false,
-            validRefChecksRemaining: 2,
-        };
+        this.state = this.getInitialState();
     }
 
     public componentDidMount(): void {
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -228,12 +202,27 @@ class ViewportPositioner extends Foundation<
     }
 
     public componentDidUpdate(prevProps: ViewportPositionerProps): void {
+        // if anchor or viewport changes the component needs to reset
+        if (
+            prevProps.anchor !== this.props.anchor ||
+            prevProps.viewport !== this.props.viewport
+        ) {
+            this.disconnectObservers();
```
janechu: could you add a method for resetting the component and just run both of these together? seems like there's no case in which you aren't doing both `disconnectObservers` and `disconnectViewport`.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -1245,6 +1247,39 @@ class ViewportPositioner extends Foundation<
                 return ViewportPositionerVerticalPositionLabel.undefined;
         }
     };
+
+    /**
+     * Gets the uninitialized state
+     */
+    private getInitialState = (): ViewportPositionerState => {
+        return {
+            disabled: true,
```
janechu: should this check for props first since disabled is settable?

scomea: State does not necessarily match props.  Before initialization the component state is disabled and it only gets enabled if not disabled by props *and* gets valid viewport/anchor values.  Setting disabled = false on the prop does not mean the disabled state will be false if the other essential resources needed by the component fail to validate.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: Merged into "detachListeners"

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
janechu: What if I want to initialize it as open?

janechu: I'm going to be away next week, pinging @chrisdholt for an opinion.

scomea: If the prop is set to true then that should override the state appropriately, the state should end up as true and the menu should be visible.

chrisdholt: @scomea is this an issue where "disabled" means something different than it typically does? There isn't a prop to set disabled as `true`, correct? This is more a function of the internals of the component?

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
```diff
@@ -1245,6 +1237,39 @@ class ViewportPositioner extends Foundation<
                 return ViewportPositionerVerticalPositionLabel.undefined;
         }
     };
+
+    /**
+     * Gets the uninitialized state
+     */
+    private getInitialState = (): ViewportPositionerState => {
```
chrisdholt: This is somewhat of a nit but I think important. While this function is descriptive of what is happening, there previously was a pattern of getInitialState in react and I'd like to avoid confusion with it. Can we modify this name slightly to do so? Perhaps `generateInitialState` or something close but not the same as the above? :)

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: Sort of.  There is a disabled prop which disables positioning logic and causes the positioner to just show up where it would in normal document flow.  The internal disabled "state" starts as true and unless it is disabled in props only gets set to false once the component's initialization completes successfully (ie. we get valid anchor & viewport).  If initialization fails to get resources it needs it ultimately ends up in a disabled state which renders the component in normal document flow.  The component remains hidden from view during initialization so we don't get flicker as we determine resources and calculate the position.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.tsx
scomea: generateInitialState it is.

PR #2539, opened 2019-12-27
chore: add fast MSFT component version to explore site

### LINKED ISSUE #2257: Add build info to all FAST apps
Let's have some way to see basic build information in the tools; at least the build date, but possibly a build number if that's useful, or version numbers of component libraries.

As the number of apps and deployment location grows, it's increasingly harder to know what's current and how current it is. We need to know that if we direct a partner somewhere, that it's up to date.

Any of these options would be a good start:
- A url that's not even linked or indexed, like /ver.txt
- A tooltip over the product header
image
- A question mark icon in that same space, with tooltip or more detailed flyout

chrisdholt:
# Description
This PR enables the ability to dynamically add version information for the MSFT components to the explore site. 
**Edit: I've update this to use the title attribute rather than passing the version as part of the heading itself.**

## Motivation & context
closes #2257 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-component-explorer/app/explorer.tsx
+++ b/packages/fast-component-explorer/app/explorer.tsx
@@ -100,22 +100,20 @@ function setViewConfigsWithCustomConfig(
 ): ObjectOfComponentViewConfigs {
     const componentViewConfigs: ObjectOfComponentViewConfigs = {};
 
-    Object.keys(viewConfigs).forEach(
-        (viewConfigKey: string): void => {
-            componentViewConfigs[viewConfigKey] = Object.assign(
-                {},
-                viewConfigs[viewConfigKey],
-                {
-                    scenarios: [
-                        {
-                            displayName: "Custom",
-                            data: viewConfigs[viewConfigKey].scenarios[0].data,
-                        },
-                    ].concat(viewConfigs[viewConfigKey].scenarios),
-                }
-            );
-        }
-    );
+    Object.keys(viewConfigs).forEach((viewConfigKey: string): void => {
+        componentViewConfigs[viewConfigKey] = Object.assign(
+            {},
+            viewConfigs[viewConfigKey],
+            {
+                scenarios: [
+                    {
+                        displayName: "Custom",
+                        data: viewConfigs[viewConfigKey].scenarios[0].data,
+                    },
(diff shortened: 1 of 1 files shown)


packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -126,6 +128,8 @@ class Explorer extends Foundation<
 > {
     public static displayName: string = "Explorer";
 
+    public static version: string = version;
```
janechu: does this need to be added? seems like we're just using the imported version

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -321,7 +325,12 @@ class Explorer extends Foundation<
                         >
                             <Background value={neutralLayerL3} drawBackground={false}>
                                 <div className={classNames(explorer_paneTitleContainer)}>
-                                    <Heading size={HeadingSize._6}>FAST Explorer</Heading>
+                                    <Heading
+                                        size={HeadingSize._6}
+                                        title={`Version ${version}`}
```
janechu: should we consider using `Badge`?

chrisdholt: We could add Badge - I meant to open as a draft initially, but here we are :)

bheston: That would be consistent with what we did for Creator. I think that works well considering the related change to publish the latest published version.
image

packages/fast-component-explorer/app/explorer.tsx
```diff
@@ -321,7 +325,12 @@ class Explorer extends Foundation<
                         >
                             <Background value={neutralLayerL3} drawBackground={false}>
                                 <div className={classNames(explorer_paneTitleContainer)}>
-                                    <Heading size={HeadingSize._6}>FAST Explorer</Heading>
+                                    <Heading
```
nicholasrice: We discussed somehow denoting that these changes could be pre-release and not available via npm in the staging case - is that work going into this PR?

chrisdholt: I'm working on that as part of this PR.

PR #2571, opened 2020-01-09
fix: lightweight button looks like plain text in high contrast mode

khamudom:
# Description
Bug: 
Accessibility: MAS4.3.1: HighContrast:  "Back to extensions home" button looks like static text in High contrast themes.
https://microsoft.visualstudio.com/Edge/_workitems/edit/24461248/

In high contrast mode, the lightweight button looks like plain text, so I added the underline.

High Contrast Black
rest
image
hover
image

High Contrast White
rest
image
hover
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/button/index.ts
+++ b/packages/fast-components-styles-msft/src/button/index.ts
@@ -40,7 +40,7 @@ import {
 import { applyDisabledState } from "../utilities/disabled";
 import {
     highContrastAccent,
-    highContrastBackground,
+    HighContrastColor,
     highContrastDisabledBorder,
     highContrastDisabledForeground,
     highContrastDoubleFocus,
@@ -73,13 +73,19 @@ const applyTransparentBackplateStyles: CSSRules<DesignSystem> = {
     ...applyFocusVisible({
         "border-color": "transparent",
         "box-shadow": "none",
+        ...highContrastHighlightForeground,
         "& $button_contentRegion::before": {
             background: neutralForegroundRest,
             height: toPx<DesignSystem>(focusOutlineWidth),
-            ...highContrastBackground,
+            ...highContrastHighlightBackground,
         },
     }),
     // Underline
+    "& $button_contentRegion::before": {
+        [highContrastSelector]: {
+            background: HighContrastColor.buttonText,
+        },
+    },
     "&:hover $button_contentRegion::before": {
         background: accentForegroundHover,
         ...highContrastHighlightBackground,
@@ -249,6 +255,11 @@ const styles: ComponentStyles<ButtonClassNameContract, DesignSystem> = {
             },
             "&$button__disabled": {
                 ...highContrastDisabledBorder,
+            },
+            "& $button_contentRegion::before": {
+                [highContrastSelector]: {
+                    background: "transparent",
+                },
             },
         },
     },
--- a/packages/fast-components-styles-msft/src/lightweight-button/index.ts
+++ b/packages/fast-components-styles-msft/src/lightweight-button/index.ts
@@ -11,7 +11,6 @@ import {
 import { applyFocusVisible, toPx } from "@microsoft/fast-jss-utilities";
 import { focusOutlineWidth } from "../utilities/design-system";
 import {
-    highContrastBackground,
     HighContrastColor,
     highContrastDisabledForeground,
     highContrastHighlightBackground,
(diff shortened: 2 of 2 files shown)


bheston:
I missed this when we talked, but the lightweight button style is supposed to match that of a link. In high contrast mode I believe our default link treatment sticks to what Windows does using yellow at rest, with no underline. Hover goes white in Windows, and we add the underline consistently in FAST. Is that how the other link-styled components behave today?

chrisdholt:
> I missed this when we talked, but the lightweight button style is supposed to match that of a link. In high contrast mode I believe our default link treatment sticks to what Windows does using yellow at rest, with no underline. Hover goes white in Windows, and we add the underline consistently in FAST. Is that how the other link-styled components behave today?

I think this assumption is incorrect. Our guidance can be that this should primarily be used as buttons, but if this is a functional button it should look like a button. If it's a link it should look like a link. The issue raised from an accessibility standpoint was that when the control *is* a button it looks like text. In this case, I don't think we should force it to look like a link; it should look like a button but have an affordance that differentiates it from text.

bheston:
Previously we've talked about the styles being a convenience for how someone wants to develop an app vs. what they want it to look like. In this case, someone may want to handle the click event provided by a button, but have it look like a link (which would normally navigate to another page).

If someone chooses for their button to look like a link in normal mode, I think it should also look like a link in HC mode. The counter to that seems like it's giving more precedence to someone using HC to understand an implementation detail of whether it's a normal button or a link that will go to another page. I'm not sure why this would be more important for them to know though.

I'd like to formalize the HC designs and check for consistency across the board, but it still seems that this should follow the colors someone expects for being able to click, as opposed to something that might otherwise just be underlined text.

Underlined text doesn't "look like a button" either. That would need some sort of outline, which would cause margin issues because a lightweight button wouldn't necessarily be laid out the same, for instance when stacking a few links vs stacking a few buttons.

chrisdholt:
> Underlined text doesn't "look like a button" either. That would need some sort of outline, which would cause margin issues because a lightweight button wouldn't necessarily be laid out the same, for instance when stacking a few links vs stacking a few buttons.

I think there’s a fair point on this looking specifically like underlined text. I understand the concern about implementation details, but high contrast is useful to help those who are visually impaired or even those who have cognitive impairments navigate the UI and understand what kind of reasonable expectation they may have from interacting with it. For that reason, I agree that potentially having button look like underlined text is fair. I don’t know that the solution though is to have it appear as a link.
 
A link and a button fundamentally do different things and I believe there is a reason that there are different styles for each in HC mode. It gives low vision users an affordance of how they can reasonably expect something to behave. A link *should* navigate a user somewhere. A button *should* submit a form or perform an on page action. I don’t believe the intention of HC mode is for authors to persist their visual design language in the way described below, I think it exists to better aid and inform the user. When a user is setting their high contrast settings, they are shown (and able to choose) colors for hyperlinks, button text, disabled text, selected text, etc. I think these map to elements and attributes purposefully. 

As an example of how this could cause issues:
A disabled hyperlink is really just text – it’s an anchor with no href and in HC mode it will look like text because that is how it functionally behaves.  A disabled button however, has a different affordance than that and functionally behaves differently. Consider a user interacting with a page which has quite a few links on it. While interacting with a form a link at the bottom suddenly turns into a disabled button – why? Those are fundamentally different elements, what caused the sudden change?  I see a lot of opportunity for confusion this way.
 
Would we similarly style all back-plated buttons which functionally are links to take on the appearance of buttons in HC mode? I would hope not, and I don’t think we do. For similar reasons to the above scenario I think this would create a ton of confusion. 

**Rather than going the route of convoluting button and link in the UI, perhaps the actual design problem is:**
-> How do we style lightweight buttons which are actually buttons in High Contrast mode without taking on the high contrast affordance of a functional link?

bheston:
This change is better than the current style, and it seems like coming up with a more comprehensive design is more involved and potentially undesirable. I've been planning to incorporate HC design into the current modular styling investigation, so perhaps that will reveal something useful here.

khamudom:
There is a reason why High Contrast has a separate option to customize hyperlink buttons and non-anchor buttons. The reason we have a Link color value is to let users know that, I am a hyperlink and I am different from all the other buttons, I am going to take you somewhere. To me, in high contrast, the Link color value is not just to look like a clickable button or a link button, but what it actually does.
I get the concern that, in a sea of text, a non-anchor lightweight button, may look like and underline text. I apologize, I don't mean to assume, but I would hope base on the context of where and how the button is used, the end-user would know it is a clickable button.
This work resolves the current issue. But that does not mean we have to stop here, we can continue to research and investigate.

PR #2573, opened 2020-01-10
fix: ensure that setting a value to null does not trigger any default button click behavior

janechu:
# Description

## Motivation & context

The value setting to null was triggering form submittal behavior such as reporting validity on invalid form elements. This fix stops any default button clicking behavior.

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/controls/control.button.tsx
```diff
@@ -52,6 +52,8 @@ class ButtonControl extends React.Component<
         e: React.MouseEvent<HTMLButtonElement>
     ) => void) => {
         return (e: React.MouseEvent<HTMLButtonElement>): void => {
```
nicholasrice: Not really this change - but does this really need to *return* a function?

janechu: Good point, i'll take a look at it on my next pass

PR #2577, opened 2020-01-10
fix: autosuggest opens menu when suggestions added 

scomea:
# Description
Auto suggest does not open menu when suggestions come in after some latency as may happen with when a server query is involved.  This change opens the menu when new suggestions come in to a previously empty component and the component has focus.

## Motivation & context
Developer feedback

## Issue type checklist

## Process & policy checklist

- [] I have updated the project documentation to reflect my changes.

### DIFF
--- a/packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
+++ b/packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
@@ -101,8 +101,6 @@ class AutoSuggest extends Foundation<
             updatedMenuVisibility = true;
         }
 
-        updatedMenuVisibility = this.validateMenuState(updatedMenuVisibility);
-
         if (updatedMenuVisibility !== this.state.isMenuOpen) {
             this.toggleMenu(updatedMenuVisibility);
         }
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -87,9 +87,21 @@ class AutoSuggest extends Foundation<
     }
 
     public componentDidUpdate(prevProps: AutoSuggestProps): void {
-        const updatedMenuVisibility: boolean = this.validateMenuState(
-            this.state.isMenuOpen
-        );
+        let updatedMenuVisibility: boolean = this.state.isMenuOpen;
+
+        if (
```
nicholasrice: It seems to me this logic is really *part* of menu state validation - do you agree? should this logic live in `validateMenuState`?

scomea: I was able to remove a redundant line there, but no. "validateMenuState" is a test we run every time we try to toggle to menu, it is very generic.  The logic here is specifically deciding whether we need to toggle the menu because suggestions changed in an update and does not run at any other time

janechu:
In what way are we testing the suggestions being added after rendering and the menu opening?

Also, are we sure this is really behavior we want? I'm imagining a scenario where the menu is closed, and the suggestions are added later, then the autosuggest is opened just because the autosuggest has focus.

PR #2578, opened 2020-01-13
feat: add stack panel to base component set

scomea:
# Description
The stack panel component enables authors to create a horizontally or vertically scrolling list of child items that can be virtualized (ie. not rendered to the dom) when not within the viewport.  Authors provide the span (height or width depending on orientation) of each item, either a single value when all child items have identical spans, or as an array of values when children have variable spans. Because the stackpanel knows the space to allocate to each item it can create layouts without actually having to render each child which can be a significant performance gain.

The stack panel accepts children in the usual way for html markup.
`<StackPanel>`
`               <button/>`
`                <button/>`
`               <button/> `
`                etc...`
</StackPanel>`
The expectation is that the stackpanel's role is to manage virtualization at React node level and that other parent components, like a data grid or a feed for example, would handle more data driven inputs and handle converting those to nodes to be rendered  in an efficient manner.  

The prototype [data grid component](https://github.com/microsoft/fast-dna/tree/users/scomea/data-grid) takes in data and interacts with the stackpanel to only create nodes only for the rows that are in the region that is being rendered.  Browsing a large dataset in this scenario could end up with a stackpanel with children that look something like this:

`<StackPanel itemSpan={[500000, 100, ...more spans...100, 300000]}>`
`                <div>Filler child that represents first 5000 rows</div>`
`                <dataRow>Row index 5001</dataRow>`
`               ...more data rows...`
 `             <dataRow>Row index 6000</dataRow>`
 `           <div>Filler child that represents rows after 6000</div>`
`</StackPanel>`

As the user scrolls the data grid tracks the scroll position and updates the children as required.  The first and last child merely exist to take up virtual space and their spans change as the component is scrolled and which data rows are actually passed to the data grid changes accordingly.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2501

Consider a stackpanel with 200 child items rendered with virtualization turned off which takes about 5 seconds of churn on the main thread when first rendered:

200NoVirt

Turning on virtualization reduces that time to less than a second:

200wVirt

Further, upping the item count from 200 to 10000 does not have a significant impact:

10kwVirt

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
@@ -16,14 +16,4 @@ export interface StackPanelClassNameContract {
      *
      */
     stackPanel__isScrolling?: string;
-
-    /**
-     *
-     */
-    stackPanel_item?: string;
-
-    /**
-     *
-     */
-    stackPanel_item__focusWithin?: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
@@ -13,7 +13,8 @@ export interface StackPanelClassNameContract {
     stackPanel_items?: string;
 
     /**
-     *
+     * indicates that the panel is scrollable
+     * (ie. contents exceed height of container)
      */
-    stackPanel__isScrolling?: string;
+    stackPanel__scrolling?: string;
 }
--- a/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
@@ -10,7 +10,7 @@ export interface StackPanelClassNameContract {
     /**
      * the container that contains the stack panel items
      */
-    stackPanel_itemContainer?: string;
+    stackPanel_items?: string;
 
     /**
      *
--- a/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
@@ -1,0 +1,29 @@
+/**
+ * The class name contract for the stackpanel component
+ */
+export interface StackPanelClassNameContract {
+    /**
+     * The root of the stackpanel component
+     */
+    stackPanel?: string;
+
+    /**
+     * the container that contains the stack panel items
+     */
+    stackPanel_itemContainer?: string;
+
+    /**
+     *
+     */
+    stackPanel__isScrolling?: string;
+
+    /**
+     *
+     */
+    stackPanel_item?: string;
+
+    /**
+     *
+     */
+    stackPanel_item__focusWithin?: string;
+}
--- a/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
+++ b/packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
@@ -13,8 +13,7 @@ export interface StackPanelClassNameContract {
     stackPanel_items?: string;
 
     /**
-     * indicates that the panel is scrollable
-     * (ie. contents exceed height of container)
+     * Applied if the stack panel is scrollable
      */
-    stackPanel__scrolling?: string;
+    stackPanel__scrollable?: string;
 }
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -347,6 +347,24 @@ describe("horizontal overflow", (): void => {
             0.0004
         );
     });
+    test("should ease the animation correctly when at animation start time", () => {
+        const renderedWithImages: any = mount(
+            <HorizontalOverflow managedClasses={managedClasses}>
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+
+        expect(renderedWithImages.instance()["easeInOutQuad"](0, 0, 0.5, 50)).toBe(0);
+    });
+    test("should ease the animation correctly when at animation end time", () => {
+        const renderedWithImages: any = mount(
+            <HorizontalOverflow managedClasses={managedClasses}>
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+
+        expect(renderedWithImages.instance()["easeInOutQuad"](50, 0, 0.5, 50)).toBe(0.5);
+    });
     test("getScrollAnimationPosition returns correct start and end values", () => {
         const renderedWithImages: any = mount(
             <HorizontalOverflow managedClasses={managedClasses}>
@@ -534,19 +552,6 @@ describe("horizontal overflow", (): void => {
         renderedWithImages.find("div.foo").simulate("load", {});
 
         expect(renderedWithImages.instance().state.itemsHeight).toBe(0);
-    });
-    test("should execute a scroll animation on the element", () => {
-        const renderedWithImages: any = mount(
-            <HorizontalOverflow managedClasses={managedClasses}>
-                {imageSet1}
-            </HorizontalOverflow>
-        );
-
-        const itemsElement: HTMLDivElement = renderedWithImages.find(
-            ".horizontal-overflow-items-class"
-        );
-
-        expect(renderedWithImages.instance()["scrollContent"](50, 0)).toBe(undefined);
     });
     test("should allow clicks on previous and next buttons", () => {
         const renderedWithImagesAndNextAndPrevious: any = mount(
@@ -791,6 +796,16 @@ describe("horizontal overflow", (): void => {
         rendered.instance()["setScrollPosition"](100);
         expect(rendered.instance()["getScrollPosition"]()).toEqual(100);
     });
+    test("settting and getting scroll position works correctly in rtl", (): void => {
+        const rendered: any = mount(
+            <HorizontalOverflow managedClasses={managedClasses} dir="rtl">
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+        expect(rendered.instance()["getScrollPosition"]()).toEqual(-0);
+        rendered.instance()["setScrollPosition"](100);
+        expect(rendered.instance()["getScrollPosition"]()).toEqual(100);
+    });
     test("getting and setting scroll position returns 0 when horizontalOverflowItemsRef isn't populated ", (): void => {
         const rendered: any = mount(
             <HorizontalOverflow managedClasses={managedClasses}>
@@ -843,6 +858,23 @@ describe("horizontal overflow", (): void => {
         }
         rendered.instance()["updateScrollAnimation"]();
         expect(rendered.instance().isScrollAnimating).toEqual(false);
+    });
+    test("scrollContent properly configures a smooth scrolling animation", () => {
+        const maxScrollFn: any = jest.fn();
+        maxScrollFn.mockReturnValue(1000);
+
+        const rendered: any = mount(
+            <HorizontalOverflow managedClasses={managedClasses}>
+                {imageSet1}
+            </HorizontalOverflow>
+        );
+        expect(rendered.instance().isScrollAnimating).toEqual(false);
+        rendered.instance().getMaxScrollDistance = maxScrollFn;
+        rendered.instance().requestFrame = jest.fn();
+        rendered.instance()["scrollContent"](0, 100);
+        expect(rendered.instance().isScrollAnimating).toEqual(true);
+        expect(rendered.instance().currentScrollAnimStartPosition).toEqual(0);
+        expect(rendered.instance().currentScrollAnimEndPosition).toEqual(100);
     });
 });
 /* tslint:enable:no-string-literal */
--- a/packages/fast-components-react-base/src/index.ts
+++ b/packages/fast-components-react-base/src/index.ts
@@ -86,6 +86,10 @@ import SliderTrackItem from "./slider-track-item";
 export { SliderTrackItem };
 export * from "./slider-track-item";
 
+import StackPanel from "./stack-panel";
+export { StackPanel };
+export * from "./stack-panel";
+
 import Tabs from "./tabs";
 export { Tabs };
 export * from "./tabs";
--- a/packages/fast-components-react-base/src/index.ts
+++ b/packages/fast-components-react-base/src/index.ts
@@ -86,9 +86,9 @@ import SliderTrackItem from "./slider-track-item";
 export { SliderTrackItem };
 export * from "./slider-track-item";
 
-import StackPanel from "./stack-panel";
-export { StackPanel };
-export * from "./stack-panel";
+// import StackPanel from "./stack-panel";
+// export { StackPanel };
+// export * from "./stack-panel";
 
 import Tabs from "./tabs";
 export { Tabs };
--- a/packages/fast-components-react-base/src/stack-panel/README.md
+++ b/packages/fast-components-react-base/src/stack-panel/README.md
@@ -12,6 +12,8 @@ Smooth scrolling: When smooth scrolling is enabled the component monitors focus 
 
 Authors should be aware that this component writes to the width/left or top/height css style attributes of children on render depending on the orientation of the stack panel.
 
+Keys: authors should strongly consider providing keys (https://reactjs.org/docs/lists-and-keys.html) to child elements to maintain stable identies for child items. 
+
 ### Props
 
  enableVirtualization:  Whether child elements are virtualized or not. Defaults to true.
--- a/packages/fast-components-react-base/src/stack-panel/README.md
+++ b/packages/fast-components-react-base/src/stack-panel/README.md
@@ -1,0 +1,1 @@
+## Stack Panel
--- a/packages/fast-components-react-base/src/stack-panel/README.md
+++ b/packages/fast-components-react-base/src/stack-panel/README.md
@@ -1,5 +1,7 @@
 ## Stack Panel
 
+NOTE: this component is in initial development and should not yet be considered stable.
+
 The *stack panel* component is a container that enables smooth scrolling and virtualization of child elements arranged horizontally or vertically.
 
 ### Usage
--- a/packages/fast-components-react-base/src/stack-panel/README.md
+++ b/packages/fast-components-react-base/src/stack-panel/README.md
@@ -7,3 +7,32 @@ The *stack panel* component is a container that enables smooth scrolling and vir
 Virtualization:  When virtualization is enabled the component does not render child items that are not in the viewport or within the defined buffer zone (ie one or more items outside the viewport).
 
 Smooth scrolling: When smooth scrolling is enabled the component monitors focus changes in child elements and will scroll the element that gains focus into view.  Note that virtualized elements can't get focus so typically authors will want at least one out of view item instanciated otherwise it will not be possible to tab outside of the viewport.
+
+Authors should be aware that this component writes to the width/left or top/height css style attributes of children on render depending on the orientation of the stack panel.
+
+### Props
+
+ enableVirtualization:  Whether child elements are virtualized or not. Defaults to true.
+
+ neverVirtualizeIndexes: Authors can provide an array of indexes which flag child indexes that should not be virtualized.  For example providing [0, 10] to this prop would mean that child items at indexes 0 and 10 would be rendered.
+
+ defaultItemSpan:  The default span of the child items in pixels, this can be overriden on an item by item basis using the "itemSpanOverrides" prop. Default is 100. 
+
+ itemSpanOverrides: An object where the keys correspond to child indexes and the values to the span to be used to calculate layout for the child at that index. For example {1:200, 5:500} would specify that the child item at index 1 has a span of 200 pixels and the one at index 5 has a span of 500 pixels.
+
+preloadBufferLength: How many out of view child items to render on either side of the viewport. Default is 1.
+
+orientation: Whether the child elements stack vertically or horizontally.  Default is vertical.
+
+onScrollChange:  Callback for when the component scrolls.
+
+nextItemPeek:  When the component scrolls focused content into view how many pixels should the next item "peek" into view.  Note that the component will clip the peek value applied based on available viewport size.  Default is 50.
+
+enableSmoothScrolling:  Whether smooth scrolling is enabled or not.  When disabled focused items will be brought into view by the browser (ie. not animated).  Default is true.
+
+scrollDuration: The lenght of the scroll animation in milliseconds.  Default is 500.
+
+initiallyVisibleItemIndex:  When provided the component will initially render with the child at that index scrolled into view.  After the component is initialzed if this value is changed 
+
+
+
--- a/packages/fast-components-react-base/src/stack-panel/README.md
+++ b/packages/fast-components-react-base/src/stack-panel/README.md
@@ -1,1 +1,9 @@
 ## Stack Panel
+
+The *stack panel* component is a container that enables smooth scrolling and virtualization of child elements arranged horizontally or vertically.
+
+### Usage
+
+Virtualization:  When virtualization is enabled the component does not render child items that are not in the viewport or within the defined buffer zone (ie one or more items outside the viewport).
+
+Smooth scrolling: When smooth scrolling is enabled the component monitors focus changes in child elements and will scroll the element that gains focus into view.  Note that virtualized elements can't get focus so typically authors will want at least one out of view item instanciated otherwise it will not be possible to tab outside of the viewport.
--- a/packages/fast-components-react-base/src/stack-panel/README.md
+++ b/packages/fast-components-react-base/src/stack-panel/README.md
@@ -18,11 +18,11 @@ Keys: authors should strongly consider providing keys (https://reactjs.org/docs/
 
  enableVirtualization:  Whether child elements are virtualized or not. Defaults to true.
 
- neverVirtualizeIndexes: Authors can provide an array of indexes which flag child indexes that should not be virtualized.  For example providing [0, 10] to this prop would mean that child items at indexes 0 and 10 would be rendered.
+ neverVirtualizeIndexes: Authors can provide an array of indexes which flag child indexes that should not be virtualized.  For example providing "[0, 10]" to this prop would mean that child items at indexes 0 and 10 would be rendered.
 
  defaultItemSpan:  The default span of the child items in pixels, this can be overriden on an item by item basis using the "itemSpanOverrides" prop. Default is 100. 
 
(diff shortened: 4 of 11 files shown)


packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+import {
+    StackPanelHandledProps,
+    stackPanelOrientation,
+    StackPanelProps,
+    StackPanelUnhandledProps,
+} from "./stack-panel.props";
+import { isFunction } from "util";
+
+export interface StackPanelState {
+    disabled: boolean;
```
nicholasrice: I don't see disabled used anywhere - is it needed?

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+    isScrolling: boolean;
+    renderedRangeStartIndex: number;
+    renderedRangeEndIndex: number;
+    direction: Direction;
+}
+
+interface ItemPosition {
```
nicholasrice: What do the values in `ItemPosition` represent? Are these pixel offset values? Is some of this information derived?

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+     */
+    public render(): React.ReactElement<HTMLDivElement> {
+        const {
+            stackPanel_itemContainer,
+        }: StackPanelClassNameContract = this.props.managedClasses;
+
+        return (
+            <div
+                {...this.unhandledProps()}
+                tabIndex={-1}
```
nicholasrice: Why do both of these elements need tabIndex?

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+                      left: toPx(this.itemPositions[index].start),
+                  }
+                : {
+                      height: toPx(this.itemPositions[index].span),
+                      top: toPx(this.itemPositions[index].start),
+                  }),
+        };
+
+        if (this.isClonableElement(item)) {
+            const styleProps: React.CSSProperties = get(item.props, "style");
```
nicholasrice: are we cloning *every* child?

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: Let's pull it until we figure out we need it.

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: Added comment "Used to store the pixel coordinates and span of items along the stackpanel's axis".  Basically it's used to create a lookup table of the coordinates and spans of the children on mount and update (see "updateItemPositions()") so we don't have to recalculate on scroll events.

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: removed, may not be necessary

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: Only the ones we actually render - those in the viewport and configurable buffer region, and those in the "never virtualize" list.  Of course if we turn off virtualization we clone them all

scomea: Might be an optimization possible *if* we can safely cache the clones between updates.

packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface StackPanelClassNameContract {
+    /**
+     * The root of the stackpanel component
+     */
+    stackPanel?: string;
+
+    /**
+     * the container that contains the stack panel items
+     */
+    stackPanel_itemContainer?: string;
```
janechu: I would prefer to just say items, so `stackPanel_items`.

packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    /**
+     *
+     */
+    stackPanel__isScrolling?: string;
```
janechu: just as we would say `stackPanel__active`, and not `stackPanel__isActive` I prefer `stackPanel__scrolling`.

Also, think you need a comment above this.

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -347,6 +347,24 @@ describe("horizontal overflow", (): void => {
             0.0004
         );
     });
+    test("should ease the animation correctly when at animation start time", () => {
```
janechu: I know it's rife through this file, but I prefer to leave out the words "correct" and "correctly" rather "should do x" would suffice since all actions should be performed correctly. In this case I think it might be better to re-title the test "should use an ease animation at the animation start time".

packages/fast-components-react-base/src/index.ts
```diff
@@ -86,6 +86,10 @@ import SliderTrackItem from "./slider-track-item";
 export { SliderTrackItem };
 export * from "./slider-track-item";
 
+import StackPanel from "./stack-panel";
+export { StackPanel };
+export * from "./stack-panel";
```
janechu: I'm not sure this should be available from the index location due to the note on this still being in initial development.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+export interface ItemSpanOverride {
+    [key: number]: number;
+}
+
+export interface StackPanelHandledProps extends StackPanelManagedClasses {
+    /**
+     * Whether items outside the viewport should be rendered or not
+     * default is true
+     */
+    enableVirtualization?: boolean;
```
janechu: I think this could just be `virtualize`.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     *  the indexes of children that should always be rendered
```
janechu: Can you provide a use case?

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+     */
+    neverVirtualizeIndexes?: number[];
+
+    /**
+     *  The default span of each item (width in horizontal orientation, height in vertical)
```
janechu: Not sure that "span" gets across the dimensionality of the items, also not sure if you shouldn't just pass the width and height separately and let the user input them, perhaps `itemWidth` and `itemHeight`.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+     */
+    defaultItemSpan?: number;
+
+    /**
+     * An object whose entries specifies the height of items that differ from the default height.
+     * Format is {itemIndex: heightValue}
+     */
+    itemSpanOverrides?: ItemSpanOverride;
```
janechu: Is this necessary in the first iteration of this component? I think you can specify an override strategy and have it be a feature.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * How many items to preload before and after the range defined by viewportMaxHeight
+     */
+    preloadBufferLength?: number;
```
janechu: Should this be settable for now? I would think that either there would be some kind of calculation we are doing to determine how much of a buffer we should have or we allow a function to be passed that would provide a calculation based on some parameters like item size and viewport size.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * The orientation of the stackpanel
+     */
+    orientation?: stackPanelOrientation;
```
janechu: I think we have orientation in the `fast-web-utilities/src/aria.ts` you could use

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * The children of the panel
+     */
+    children?: React.ReactNode;
```
janechu: I think children doesn't need to be specified

packages/fast-components-react-base/src/stack-panel/stack-panel.schema.ts
```diff
@@ -0,0 +1,58 @@
+export default {
+    $schema: "http://json-schema.org/schema#",
+    title: " stack panel",
```
janechu: nit: space in from of title, would capitalize the "S".

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+        }
+
+        let renderEndIndex: number = visibleRangeEndIndex + bufferLength;
+        if (renderEndIndex > lastIndex) {
+            renderEndIndex = lastIndex;
+        }
+        this.setState({
+            renderedRangeStartIndex: renderStartIndex,
+            renderedRangeEndIndex: renderEndIndex,
+            isScrolling: this.itemContainerSpan > this.viewPortSpan ? true : false,
```
janechu: can just be `this.itemContainerSpan > this.viewPortSpan`, don't need a ternary for true/false.

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+     * Builds a list of positioning data for all items
+     */
+    private updateItemPositions = (): void => {
+        const itemPositions: ItemPosition[] = [];
+        let currentItemTop: number = 0;
+        const childrenAsArray: React.ReactNode[] = React.Children.toArray(
+            this.props.children
+        );
+        const itemCount: number = childrenAsArray.length;
+        const keys: string[] = Object.keys(this.props.itemSpanOverrides);
```
janechu: What's the benefit of the itemSpanOverrides not being just an array?

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
janechu: Seems to me if the itemContainerSpan is larger than the viewPortSpan, then that implies it is scrollable, not the isScrolling is happening.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * When the component scrolls focused content into view how many pixels should the next item "peek" into view
+     */
+    nextItemPeek?: number;
+
+    /**
+     * Whether to enable smooth scrolling
+     */
+    enableSmoothScrolling?: boolean;
```
janechu: I wonder if we can't just provide a prop that is an animation function that can be passed, so instead of having a smooth scrolling function, we have a secondary smooth scrolling function they can pass in to a `scrollAnimation` prop

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Callback for when scroll values change
+     */
+    onScrollChange?: (newScrollValue: number, scrollMaxValue: number) => void;
+
+    /**
+     * When the component scrolls focused content into view how many pixels should the next item "peek" into view
+     */
+    nextItemPeek?: number;
```
janechu: should we instead of offering a peek, just make this controlled with regards to the scrolled distance?

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
janechu: Should we even have orientation? Can virtualization happen in both horizontal and vertical axis? Take excel for example.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: The scenario you can see working in the stories is the case where we always render the first and last item in the list so that tabbing to the component brings one of those into focus.  The other use case is if someone wanted to implement something that resembled lazy loading where some items don't unload after they have been loaded.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: I want to bring over a datagrid component to use this and it would require variable spans.

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: It needs to be keyed as it only specifies the items that vary from the default span (ie. it does not need to include an entry for each item). I thought about using a Map but didn't think that seemed better.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: I've thought about having a function to calculate the animation curve but I'm not sure we should do that instead of a Boolean.  I'm thinking we want the scrolling on by default but if all we have is a function prop then how do you turn it off?

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: Yes, IMO we should have orientation.  A horizontal carousel could totally benefit from virtualization, for example,

packages/fast-components-react-base/src/stack-panel/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Authors should be aware that this component writes to the width/left or top/height css style attributes of children on render depending on the orientation of the stack panel.
+
+### Props
+
+ enableVirtualization:  Whether child elements are virtualized or not. Defaults to true.
+
+ neverVirtualizeIndexes: Authors can provide an array of indexes which flag child indexes that should not be virtualized.  For example providing [0, 10] to this prop would mean that child items at indexes 0 and 10 would be rendered.
+
+ defaultItemSpan:  The default span of the child items in pixels, this can be overriden on an item by item basis using the "itemSpanOverrides" prop. Default is 100.
```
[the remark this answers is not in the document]
scomea: Yes, "span" is meant to refer to height or width based on orientation.  Happy to change the term if someone has a better idea,  "size" works for me.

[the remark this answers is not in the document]
nicholasrice: I had this thought also, but then looked up at it seemed accurate to me - in particular the example:

> the full extent of something from end to end; the amount of space that something covers.
"a warehouse with a clear span of 28 feet"

https://www.google.com/search?sxsrf=ACYBGNR6PRYcr1tTn75z_MHjdyiuCCSBOA%3A1579799345198&ei=MdMpXunaC4HN-gTT14_IDQ&q=span+definition&oq=span+definition&gs_l=psy-ab.3..35i39l2j0l8.2918.7009..7253...1.2..0.62.578.12......0....1..gws-wiz.......0i71j0i67j0i131j0i10j0i22i30j0i22i10i30.5vqFrfo9VXU&ved=0ahUKEwipt6WWm5rnAhWBpp4KHdPrA9kQ4dUDCAs&uact=5

"size" seems good to me also, and avoids that conflation

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
nicholasrice: should we expect a render function instead of children, so that we can pass the required data as an arg to that cb, instead of cloning children?

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
janechu: I'm in agreement, I meant what if it was virtualizing both horizontally and vertically.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
janechu: Well, this is in base correct? Animation is something I would consider as part of styling, so if anything I would offer a prop to inject the animation function, any implementation in a styled lib such as `fast-component-styles-msft` would then have the exported animation. Maybe @bheston or @nicholasrice has an opinion.

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
janechu: What I mean is, why do:
```
{
    "0": "foo",
    "1": "bar"
}
```

and not just:
```
[
    "foo",
    "bar"
]
```

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: The only span/size we are concerned with is the along the scrolling axis of the panel, so it doesn't seem ideal to take both height and width.  I think that would be confusing as we'd just be throwing away the value that isn't on the axis we scroll on.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: I think it's useful and straightforward.  It does not mean that we can't have some calculation that figures out the buffer based on an itemheight/containerHeight calculation.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: I was not intending to support virtualization on both axis simultaneously in this component.  The scenarios this is intended for initially is a feed and datagrid scenario which does not require more than one axis.

packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
scomea: sure

packages/fast-components-class-name-contracts-base/src/stack-panel/index.ts
scomea: done

packages/fast-components-react-base/src/index.ts
scomea: good idea

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: done

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: I've thought about that a fair bit and have the idea that the base stackpanel component should accept children directly so that an author could populate it directly with markup, I think that's valuable.  Components derived from this one, like say a datagrid, would be the place where data inputs are converted to children with render functions and could use callback events that pass the rendered range to update children as the component is interacted with.  This could also support a pagination strategy where the derived component could pass a single child that represents a range of out of view children as a "page" but maintains the height.  Working on a proof of concept for this.

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
bheston: As we chatted, I like the idea of not baking the animation into the component, though I understand the particular list / tabbing issue you're solving. If there's a good way to encapsulate this and allow it to be easily updated I think that makes sense.

packages/fast-components-react-base/src/stack-panel/stack-panel.schema.ts
```diff
@@ -0,0 +1,58 @@
...
+    description: "A  stack panel component's schema definition.",
```
eljefe223: extra space

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: done

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+            <div onFocusCapture={this.onItemFocus} style={newStyleProps}>
+                {item}
+            </div>
+        );
+    };
+
+    /**
+     * Builds a list of positioning data for all items
+     */
+    private updateItemPositions = (): void => {
```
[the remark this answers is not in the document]
scomea: The developer using the stack panel must be able to provide correct dimensions for the children that take into account whatever css is applied to them.  We don't measure dom elements in any way.

packages/fast-components-react-base/src/stack-panel/stack-panel.schema.ts
scomea: done

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: Because if we do a straight array we need to populate a value for every item, with keys we can limit to just those items that deviate from the default height.  Think a data grid that has a large number of rows but only the one the user is operating on is expanded.  With keys we can do:
 {
    "375": 200,
}

vs. 
[
    …, to the beginning
    100,
    100,
    200,
    100,
    100,
    … to the end
]

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: yep

packages/fast-components-react-base/src/stack-panel/README.md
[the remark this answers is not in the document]
scomea: Folks seem to be ok with "span" after some discussion.  Please reactivate if further discussion needed.

packages/fast-components-react-base/src/stack-panel/stack-panel.schema.ts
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,738 @@
...
+        if (!isNil(this.rootElement.current)) {
+            if (((window as unknown) as WindowWithResizeObserver).ResizeObserver) {
+                this.resizeDetector = new ((window as unknown) as WindowWithResizeObserver).ResizeObserver(
+                    this.handleResize
+                );
+                this.resizeDetector.observe(this.rootElement.current);
+            }
+        }
+
+        this.viewPortSpan = this.getViewportSpan();
```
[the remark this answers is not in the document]
scomea: yep

packages/fast-components-react-base/src/stack-panel/stack-panel.schema.ts
scomea: zapped

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,84 @@
...
+import {
+    ManagedClasses,
+    StackPanelClassNameContract,
+} from "@microsoft/fast-components-class-name-contracts-base";
+
+export interface StackPanelManagedClasses
+    extends ManagedClasses<StackPanelClassNameContract> {}
+export interface StackPanelUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+export enum stackPanelOrientation {
```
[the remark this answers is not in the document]
scomea: enum no longer exists

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: See the concept working in the data-grid working branch 
https://github.com/microsoft/fast-dna/tree/users/scomea/data-grid

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
janechu: You do not need to populate a real value for every item, if you did `var test = new Array(100)` you would get an array of 100 empty items.

janechu:
I'd like to backtrack on this PR a little, I think this is too complex a task to not have an associated issue filed before this feature can be put in.

I think we need to update our contributor documentation, but what I would suggest is closing this issue and filing a feature request that includes:

- abstract: high level description of the component and what it should do
- background: any background information needed to understand the ask
- proposed implementation: how will the component be implemented
- additional information: any extra information that should be taken into consideration (optional)

packages/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: Replaced "defaultItemSpan" prop with simply "itemSpan" that takes either a single number for uniform sized items or a flat array (number[]) for variable sized items.  Seems more intuitive.

scomea: Switched to a number[] array for simplicity.

packages/fast-components-react-base/src/stack-panel/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+preloadBufferLength: How many out of view child items to render on either side of the viewport. Default is 1.
+
+orientation: Whether the child elements stack vertically or horizontally.  Default is vertical.
+
+onScrollChange:  Callback for when the component scrolls.
+
+nextItemPeek:  When the component scrolls focused content into view how many pixels should the next item "peek" into view.  Note that the component will clip the peek value applied based on available viewport size.  Default is 50.
+
+enableSmoothScrolling:  Whether smooth scrolling is enabled or not.  When disabled focused items will be brought into view by the browser (ie. not animated).  Default is true.
```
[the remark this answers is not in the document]
scomea: removed animation code as I found a css based solution

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: removed animation code as I found a css based solution

packages/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: removed animation code as I found a css based solution

PR #2582, opened 2020-01-14
feat: update ContextMenu to neutralLayerFloating

eljefe223:
# Description
Closes: #2250 

Updated `<ContextMenu/>` to use `neutralLayerFloating`

Before:
Screen Shot 2020-01-14 at 2 43 24 PM

After:
Screen Shot 2020-01-14 at 2 54 55 PM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/context-menu-item/index.ts
+++ b/packages/fast-components-styles-msft/src/context-menu-item/index.ts
@@ -8,7 +8,6 @@ import {
     neutralFillStealthHover,
     neutralFocus,
     neutralForegroundRest,
-    neutralLayerFloating,
 } from "../utilities/color";
 import { ContextMenuItemClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 import { height, horizontalSpacingNumber } from "../utilities/density";
--- a/packages/fast-components-styles-msft/src/context-menu-item/index.ts
+++ b/packages/fast-components-styles-msft/src/context-menu-item/index.ts
@@ -61,11 +61,11 @@ const styles: ComponentStyles<ContextMenuItemClassNameContract, DesignSystem> = 
             },
         }),
         "&:hover": {
-            background: neutralFillStealthHover(neutralLayerFloating),
+            background: neutralFillStealthHover,
             ...highContrastSelected,
         },
         "&:active": {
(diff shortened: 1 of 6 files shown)


chrisdholt:
@eljefe223 I think this is either a feature or a fix, we want the package to bump with this change. I'd advocate for fix here - while I know that this is seen as broken from a design standpoint, when the initial work was done the new recipe hadn't been created yet. This is net-new technically for the component, so I would suggest "feat"

eljefe223:
> This will make the menu container and menu item appear correct, but a context menu can have any type of child, which will not have the correct background context. I was looking for something like we had solved for the select control where the flyout portion has the intended design system context, so all child components will render correctly. For a specific example, menus also have divider children, and there's no single style update that will allow a divider to work in and out of a context menu.

Thanks, Brian. We chatted about this offline. Was trying to avoid this as it was a but of extra work but I think it makes sense. Took care of this with the latest commit.

packages/fast-components-styles-msft/src/context-menu-item/index.ts
```diff
@@ -8,6 +8,7 @@ import {
     neutralFillStealthHover,
     neutralFocus,
     neutralForegroundRest,
+    neutralLayerFloating,
```
bheston: Still needed?

packages/fast-components-styles-msft/src/context-menu/index.ts
```diff
@@ -3,17 +3,18 @@ import { ComponentStyles } from "@microsoft/fast-jss-manager";
...
     HighContrastColor,
     highContrastOptOutProperty,
     highContrastSelector,
 } from "../utilities/high-contrast";
+import { neutralLayerFloating } from "../utilities/color";
 
 const styles: ComponentStyles<ContextMenuClassNameContract, DesignSystem> = {
     contextMenu: {
-        background: backgroundColor,
+        background: neutralLayerFloating,
```
bheston: I think this can go back to `backgroundColor` since it's wrapped in the correct color <Background> now.

eljefe223: This needs to remain since the styles get applied outside the context of the background change.

PR #2587, opened 2020-01-15
feat: adds background fill, text fill, and stroke fill capabilities to Figma plugin

nicholasrice:
# Description
Enables setting background fills, stroke fills, and text fills as color recipes to Figma nodes. Setting a recipe will assign the color to the node, and the node will react to any upstream changes to fills. 

This means child nodes can react to color changes of their parent, enabling adaptive color.


packages/fast-figma-plugin-msft/src/interface/plugin-ui.tsx
```diff
@@ -0,0 +1,200 @@
...
+        active: string;
+        action: any;
+    }): JSX.Element {
+        return (
+            <React.Fragment>
+                <Label htmlFor={options.id} jssStyleSheet={recipeLabelStyleOverrides}>
+                    {options.label}
+                </Label>
+                <Select
+                    id={options.id}
```
janechu: wondering if we should worry about id conflicts here, should we use `uniqueId` from `lodash-es` and have this as a prefix?

packages/fast-figma-plugin-msft/src/messaging/common.ts
```diff
@@ -0,0 +1,44 @@
...
+            isString(e.data.pluginId) &&
+            JSON.parse(e.data.pluginMessage)
+        );
+    } catch (e) {
+        return false;
+    }
+}
+
+/**
+ * Expand uppon the browser's MessageEvent object to define
```
janechu: nit: upon

Also, not too sure what you're trying to say here, did you mean "Expand on the browser's MessageEvent object to define Figma plugin data"?

packages/fast-figma-plugin-msft/src/interface/plugin-ui.tsx
nicholasrice: we control this currently, so no need right now.

packages/fast-figma-plugin-msft/src/messaging/common.ts
nicholasrice: I'll re-phrase - just means that the "data" property in Figma follows a strict format that isn't enforced by the `MessageEvent` type - should probably be called a `FigmaMessageEvent`. I'll also see if they export something from their d.ts file for this, cause they *should*

PR #2603, opened 2020-01-22
feat: Add focusTargetOnClose prop to dialog component

scomea:
# Description
This change enables authors to specify an element to focus or callback to invoke when a dialog is unmounted.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2565

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/dialog/dialog.props.ts
```diff
@@ -45,6 +45,14 @@ export interface DialogHandledProps extends DialogManagedClasses {
      */
     modal?: boolean;
 
+    /**
+     * Defines where focus goes when the dialog is unmounted.
+     * This is useful because modal dialogs will take focus back until that point in their life cycle.
+     * The element or ref provided will be focused when the dialog unmounts.
+     * Alternatively, providing a callback function gives developers more flexibility if needed.
+     */
+    refocusTarget?: React.RefObject<any> | HTMLElement | (() => void);
```
chrisdholt: I think we should consider alternative names here so we don’t paint ourselves into a corner semantically. Refocus implies returning focus to a previously focused thing - there are valid scenarios where you may need to move focus to something which previously did not have focus. 

What about something along the lines of ‘focusTargetOnClose’, ‘focusTargetOnUnkount’, etc?

packages/fast-components-react-base/src/utilities/extractElement.ts
```diff
@@ -0,0 +1,26 @@
...
+
+        const foundNode: Element | Text | null = ReactDOM.findDOMNode(sourceRef.current);
+
+        if (foundNode instanceof HTMLElement) {
+            return foundNode;
+        }
+    }
+
+    return null;
+}
```
chrisdholt: Could this be useful elsewhere? It seems like since this is somewhat of a generic react utility that `fast-utilities-react` might be a better place for this. Would be great as well to get test coverage.

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
```diff
@@ -1645,49 +1645,4 @@ describe("viewport positioner", (): void => {
...
-                    managedClasses={managedClasses}
-                />
-            </div>
-        );
-
-        const positioner: any = rendered.find("BaseViewportPositioner");
-        expect(positioner.instance()["extractElementFromRef"](anchorElement)).toBe(
-            anchorElement.current
-        );
-    });
```
chrisdholt: Should we do a separate PR to extract this utility? It seems significant enough to merit it's own PR.

packages/fast-components-react-base/src/dialog/dialog.props.ts
nicholasrice: What's the reason the Ref.RefObject can be `any`? Don't we want that to be an HTMLElement?

packages/fast-components-react-base/src/utilities/extractElement.ts
```diff
@@ -0,0 +1,26 @@
+import ReactDOM from "react-dom";
+import { isNil } from "lodash-es";
+
+/**
+ * Attempts to extract an html element from a ref
+ */
+export function extractElementFromRef(
+    sourceRef: React.RefObject<any> | HTMLElement
```
nicholasrice: I think this should be `React.RefObject<HTMLElement>`

packages/fast-components-react-base/src/utilities/extractElement.ts
```diff
@@ -0,0 +1,26 @@
+import ReactDOM from "react-dom";
+import { isNil } from "lodash-es";
+
+/**
+ * Attempts to extract an html element from a ref
+ */
+export function extractElementFromRef(
```
nicholasrice: I'd prefer to see a name more inline with the function implementation. This name implies it accepts a ref and returns an element, but it actually accepts a ref *or* an element. Can you either update the name to be more comprehensive or limit the scope of the function?

packages/fast-components-react-base/src/utilities/extractElement.ts
scomea: See https://github.com/microsoft/fast-dna/pull/2613
Moved to the utilities in Base components and added tests,

packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
scomea: https://github.com/microsoft/fast-dna/pull/2613

packages/fast-components-react-base/src/dialog/dialog.props.ts
scomea: Updated, but moved to it's own pr
https://github.com/microsoft/fast-dna/pull/2613

packages/fast-components-react-base/src/utilities/extractElement.ts
scomea: yep.  Moved to its own pr
https://github.com/microsoft/fast-dna/pull/2613

packages/fast-components-react-base/src/utilities/extractElement.ts
scomea: Any suggestions?  I don't want to limit the scope as part of usefulness is in differentiating when a ref or element is passed in directly as we seem to be doing that in props with some regularity.

scomea: renamed to "extractHtmlElement" in the other PR

scomea: (if further discussion required let's do it there - https://github.com/microsoft/fast-dna/pull/2613)

packages/fast-components-react-base/src/dialog/dialog.props.ts
scomea: renamed to focusTargetOnClose

packages/fast-components-react-base/src/dialog/dialog.spec.tsx
```diff
@@ -263,4 +263,37 @@ describe("dialog", (): void => {
 
         document.body.removeChild(container);
     });
+
+    test("should invoke focusTargetOnClose function on unmount of non-modal dialogs", () => {
```
nicholasrice: Should we also test focus invocation of element / ref<htmlelement> props?

scomea: added one for elements

PR #2617, opened 2020-01-23
fix: prevent default on tabs component navigation

### LINKED ISSUE #2610: Tabs component does not block scroll on up/down arrow keys navigation
The [tabs](https://github.com/microsoft/fast-dna/blob/master/packages/fast-components-react-base/src/tabs) component seems to provide support for up/down arrow key navigation, however, it does not block scroll in the page.

Please follow these steps to be able to repro:
1. Go to https://explore.edgewebui.net/FastMSFT/components/pivot
2. See the 'pivot' component uses the 'tabs' component.
3. Either update the 'pivot' container height in the dom or make screen small enough to have scroll in the page.
4. Click on any tab, and then start navigating using up/down arrow keys.

Expected behavior: no scroll happens

chrisdholt:
# Description
closes #2610 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/tabs/tabs.tsx
+++ b/packages/fast-components-react-base/src/tabs/tabs.tsx
@@ -318,10 +318,13 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
         switch (e.keyCode) {
             case keyCodeArrowLeft:
             case keyCodeArrowUp:
+                e.preventDefault();
                 this.activateTab(TabLocation.previous);
                 break;
             case keyCodeArrowRight:
             case keyCodeArrowDown:
+                e.preventDefault();
+
                 this.activateTab(TabLocation.next);
                 break;
             case keyCodeHome:
--- a/packages/fast-components-react-base/src/tabs/tabs.spec.tsx
+++ b/packages/fast-components-react-base/src/tabs/tabs.spec.tsx
@@ -360,6 +360,35 @@ describe("tabs", (): void => {
         expect(renderedWithChildren.prop("children")).not.toBe(undefined);
     });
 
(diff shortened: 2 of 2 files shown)


packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -318,10 +318,13 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
...
+                e.preventDefault();
                 this.activateTab(TabLocation.previous);
                 break;
             case keyCodeArrowRight:
             case keyCodeArrowDown:
+                e.preventDefault();
+
                 this.activateTab(TabLocation.next);
                 break;
             case keyCodeHome:
```
prabashn: Didn't even know about the home/end buttons... that's neat. In windows you get scroll up down due to these as well, should you be doing the same for those as well?

chrisdholt: interesting - is it windows specific?

prabashn: Although, I'm not sure if taking over those two buttons even while within the tabs control is necessarily a good thing.

prabashn: > interesting - is it windows specific?

Don't know -- that's all I have access to now. By scroll up/down, I mean "home"=scroll to the top of the document, and "end"=scroll to the end of the document.

Since these are standard buttons we (I) use in windows to reach the top/end, maybe not take over these, as it could become annoying.

chrisdholt: Yep, that's my bent here too; thx @prabashn!

PR #2626, opened 2020-01-27
fix: select - add css class for scaleToFit mode

scomea:
# Description
Found that styling for menu needed to be able to differentiate styling for "scaleToFit" mode, so added a class for that and tweaked our styles.

## Motivation & context
Styles that worked for scale to fit mode needed adjustments for when that is disabled (bad margins).  

## Issue type checklist
- [] **Bug fix**: A change that fixes an issue, link to the issue above.

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-msft/src/select/select.stories.tsx
+++ b/packages/fast-components-react-msft/src/select/select.stories.tsx
@@ -5,7 +5,11 @@ import { Heading, HeadingSize } from "../heading";
 import { Select } from "./";
 import { uniqueId } from "lodash-es";
 import { action } from "@storybook/addon-actions";
-import { AxisPositioningMode } from "@microsoft/fast-components-react-base";
+import {
+    AxisPositioningMode,
+    ViewportPositionerVerticalPosition,
+    ViewportPositionerHorizontalPosition,
+} from "@microsoft/fast-components-react-base";
 import { Dialog } from "../dialog";
 
 const rootElement: React.RefObject<HTMLDivElement> = React.createRef<HTMLDivElement>();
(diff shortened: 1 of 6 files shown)


packages/fast-components-react-msft/src/select/select.stories.tsx
```diff
@@ -154,12 +158,14 @@ storiesOf("Select", module)
...
     ))
-    .add("Adjacent", () => (
+    .add("Right", () => (
         <Select
             placeholder="Select an option"
             onValueChange={action("onValueChange")}
             menuFlyoutConfig={{
                 horizontalPositioningMode: AxisPositioningMode.adjacent,
+                defaultHorizontalPosition: ViewportPositionerHorizontalPosition.right,
+                horizontalLockToDefault: true,
```
janechu: This test seems to be using horizontal positioning adjacent and a default horizontal position right, since this is locked to default, is there any point in passing the `horizontalPositioningMode` and if not can you remove and add back another test for the adjact test?

packages/fast-components-react-msft/src/select/select.stories.tsx
```diff
@@ -154,12 +158,14 @@ storiesOf("Select", module)
             />
         </Select>
     ))
-    .add("Adjacent", () => (
+    .add("Right", () => (
         <Select
             placeholder="Select an option"
             onValueChange={action("onValueChange")}
             menuFlyoutConfig={{
                 horizontalPositioningMode: AxisPositioningMode.adjacent,
```
janechu: does this mean that since it is locked to default `horizontalPositioningMode` does nothing?

scomea: As it's implemented now the "defaults" are the ones allowed by the current positioning mode settings.  I could see setting it up so that locking to a default would override the other settings but that would be a task unto itself.

PR #2650, opened 2020-01-30
fix: select "aria-labelledby" should never be"undefined" string, remove "aria-live" tag

scomea:
# Description
Avoid adding "undefined" to the aria tag rendered on the component when the labelledBy prop is not set.  Remove "aria-live" tag from trigger as it was incorrectly causing value to be read on page load.

## Motivation & context
- nothing good could come from an incorrect aria attribute
- narrator should not read values when select is not focused

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/select/select.tsx
+++ b/packages/fast-components-react-base/src/select/select.tsx
@@ -465,10 +465,9 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (props.multiselectable) {
             return null;
         }
-        const labelledByFromProp: string = isNil(this.props.labelledBy)
-            ? ""
-            : `${this.props.labelledBy} `;
-        const labelledBy: string = `${labelledByFromProp}${triggerId}`;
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -466,15 +465,17 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (props.multiselectable) {
             return null;
         }
-        const labelledBy: string = `${this.props.labelledBy} ${triggerId}`;
+        const labelledByFromProp: string = isNil(this.props.labelledBy)
+            ? ""
+            : `${this.props.labelledBy} `;
+        const labelledBy: string = `${labelledByFromProp}${triggerId}`;
```
janechu: This could just be:
```
const labelledBy: string = `${isNil(this.props.labelledBy) ? "" : `${this.props.labelledBy} `}${triggerId}`;
```

To prevent adding another const. Same for the one below.

PR #2654, opened 2020-01-31
feat: viewport positioner enable delayed instanciation

scomea:
# Description
Adds a Boolean prop "delayContentInstanciation" to viewport positioner which defaults to false. When true the contents of the positioner are not instanciated until after an initial rendering pass where the empty component is placed in the layout.  The reason for this is that partners have encountered issues where the contents of a positioner get focus before the component gets placed and as a result the browser's scroll into view behavior is triggered as an undesirable side effect.

This change also required the component to support resizing of the positioner due to changing content size which it did not previously.

Tests have also been improved and coverage increased.

## Motivation & context
Addresses unwanted scroll into view behavior.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
+++ b/packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
@@ -1341,7 +1341,7 @@ describe("viewport positioner", (): void => {
         expect(positioner.instance().state.currentVerticalPosition).toBe(
             ViewportPositionerVerticalPositionLabel.bottom
         );
-        positioner.instance()["updatePositionerOffset"]();
+        positioner.instance()["updatePositionerOffset"](positionerRectX70Y70);
         expect(positioner.instance().baseHorizontalOffset).toBe(0);
         expect(positioner.instance().baseVerticalOffset).toBe(0);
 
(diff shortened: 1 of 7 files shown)


packages/fast-components-react-base/src/viewport-positioner/viewport-positioner.spec.tsx
```diff
@@ -209,350 +207,354 @@ describe("viewport positioner", (): void => {
...
+                horizontalPositioningMode={AxisPositioningMode.adjacent}
+                defaultHorizontalPosition={ViewportPositionerHorizontalPosition.left}
+                verticalPositioningMode={AxisPositioningMode.adjacent}
+                defaultVerticalPosition={ViewportPositionerVerticalPosition.top}
+                anchor={anchorElement}
+                verticalLockToDefault={true}
+                horizontalLockToDefault={true}
+                managedClasses={managedClasses}
+            />,
+            { attachTo: container }
```
janechu: This is starting to look like it's potentially changing implementation of this component. Can you verify the existing method of using a react ref still works?

scomea: We recently removed the extract ref function from this component and moved it to a shared utility and we test extraction there (utilities/extract-element.spec.tsx):

 test("extractHtmlElement function returns element passed in as a ref", (): void => {
        const rendered: any = shallow(<TestClass />);

        const extracted: HTMLElement = extractHtmlElement(rendered.instance().testRef);
        expect(extracted).toEqual(rendered.instance().testRef.current);
    });

At that time I moved the tests for that function out of this spec file and into that one, as well adding the test above.  I can duplicate something like that here if desired, but seems somewhat redundant to me?  I could go either way.

PR #2660, opened 2020-02-05
chore: update to simplify getting started and improve documentation

awentzel:
# Description
Simplifies the getting started documentation and build process.
* Adds script `prepare` to automatically execute Lerna command as part of Install.
* Adds links for Lerna and Yarn to open in a new window.  We always want external references to open in a new window rather than redirect our users off our www.fast.design.

## Motivation & context
I experienced some confusion on a informational message from Lerna during the installation and prepare step.  I researched the [issue](https://remarkablemark.org/blog/2019/04/06/lerna-bootstrap-prepare-yarn-workspaces/) and looks like an improvement that could be made since `yarn run prepare` is invalid and that could confuse others.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -20,16 +20,21 @@ cd fast-dna
 yarn global add lerna
 ```
 
-Install and symbolically link dependencies across all workspaces:
+Installs all the dependencies:
 
 ```shell
 yarn install
 ```
 
-- `lerna run prepare` to symbolically link dependencies
-- `yarn tslint` or `yarn tslint:fix` runs tslint on typescript in a given package.
-- `yarn unit-tests` runs unit-tests for the package.
-- `yarn test` runs processes in a package required to pass prior to check-in and includes building, linting, and unit-testing.
+Builds all the packages:
+
+```shell
+lerna run prepare
+```
+
+- `yarn tslint` or `yarn tslint:fix` executes tslint on TypeScript in a given package.
+- `yarn unit-tests` execute unit testing for packages.
+- `yarn test` executes processes in a package required to pass before check-in and generally includes building, linting, and unit testing.
 
 To run these processes across *all* projects, substitute `yarn` for `lerna run ____`:
 
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -33,7 +33,7 @@ lerna run prepare
 ```
 
 - `yarn tslint` or `yarn tslint:fix` executes tslint on TypeScript in a given package.
-- `yarn unit-tests` execute unit testing for packages.
+- `yarn unit-tests` executes unit testing for packages.
 - `yarn test` executes processes in a package required to pass before check-in and generally includes building, linting, and unit testing.
 
 To run these processes across *all* projects, substitute `yarn` for `lerna run ____`:
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -23,7 +23,7 @@ yarn global add lerna
 Installs all the dependencies:
 
 ```shell
-yarn install
+yarn
 ```
 
 Builds all the packages:
--- a/docs/en/contributing/install.md
+++ b/docs/en/contributing/install.md
@@ -8,11 +8,11 @@ sidebar_label: Install
 
 ## Prerequisites
 
-Before setting up FAST-DNA, install Git and [yarn](https://yarnpkg.com/en/docs/install).
+Before setting up FAST-DNA, install Git and <a href="https://yarnpkg.com/en/docs/install" target="_blank">Yarn</a>.
 
 ## Setup the source repository
 
-Clone the repository, `cd` into the project, install [Lerna](https://github.com/lerna/lerna) and install dependencies:
+Clone the repository, `cd` into the project, install <a href="https://github.com/lerna/lerna" target="_blank">Lerna</a> and install dependencies:
 
 ```shell
 git clone https://github.com/Microsoft/fast-dna.git
@@ -20,21 +20,16 @@ cd fast-dna
 yarn global add lerna
 ```
 
-Install all dependencies:
+Install and symbolically link dependencies across all workspaces:
 
 ```shell
 yarn install
 ```
 
-Prepare the workspaces:
-
-```shell
-lerna run prepare
-```
-
+- `lerna run prepare` to symbolically link dependencies
 - `yarn tslint` or `yarn tslint:fix` runs tslint on typescript in a given package.
 - `yarn unit-tests` runs unit-tests for the package.
-- `yarn test` runs processes in a package required to pass prior to check-in. Generally includes building, linting, and unit-testing.
+- `yarn test` runs processes in a package required to pass prior to check-in and includes building, linting, and unit-testing.
 
 To run these processes across *all* projects, substitute `yarn` for `lerna run ____`:
 
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "integration-tests:alpha": "node build/testing/sauce-labs/test-browsers.js alpha",
     "integration-tests:beta": "node build/testing/sauce-labs/test-browsers.js beta",
     "integration-tests:release": "node build/testing/sauce-labs/test-browsers.js release",
+    "prepare": "lerna run prepare",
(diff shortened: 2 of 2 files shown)


docs/en/contributing/install.md
```diff
@@ -8,33 +8,28 @@ sidebar_label: Install
...
-
-```shell
-lerna run prepare
-```
-
+- `lerna run prepare` to symbolically link dependencies
 - `yarn tslint` or `yarn tslint:fix` runs tslint on typescript in a given package.
 - `yarn unit-tests` runs unit-tests for the package.
-- `yarn test` runs processes in a package required to pass prior to check-in. Generally includes building, linting, and unit-testing.
+- `yarn test` runs processes in a package required to pass prior to check-in and includes building, linting, and unit-testing.
```
chrisdholt: I think this was intentionally broad as not every package includes all three of those.

docs/en/contributing/install.md
```diff
@@ -8,33 +8,28 @@ sidebar_label: Install
 
 ## Prerequisites
 
-Before setting up FAST-DNA, install Git and [yarn](https://yarnpkg.com/en/docs/install).
+Before setting up FAST-DNA, install Git and <a href="https://yarnpkg.com/en/docs/install" target="_blank">Yarn</a>.
```
chrisdholt: What is the purpose of changing this to be an anchor here (and below) rather than leveraging the markdown syntax?

docs/en/contributing/install.md
```diff
@@ -8,33 +8,28 @@ sidebar_label: Install
...
 yarn install
 ```
 
-Prepare the workspaces:
-
-```shell
-lerna run prepare
-```
-
+- `lerna run prepare` to symbolically link dependencies
```
janechu: `lerna run prepare` does not symbolically link dependencies, it builds the packages

docs/en/contributing/install.md
awentzel: Linking external / 3rd party tools and resources via a new window is a best practice when writing content.

chrisdholt: I'm not going to block on this, but is there a source of truth we are using to determine that? This is changing the default behavior of the links on the page and setting a new default to **always** open in a new tab. Currently if a user wants to open in a new tab there are a couple ways they can do that - 1) context menu, and 2) keyboard shortcut. 

Letting the user choose how to manage their browsing workflow seems to be the ideal here and in most situations. There may be exceptions when you absolutely want to pick a different default as is being done here (a user is doing work on a page, form submission where work could be lost, etc) but I'm not sure this is the same.

docs/en/contributing/install.md
```diff
@@ -8,33 +8,33 @@ sidebar_label: Install
...
 
 ```shell
 lerna run prepare
 ```
 
-- `yarn tslint` or `yarn tslint:fix` runs tslint on typescript in a given package.
-- `yarn unit-tests` runs unit-tests for the package.
-- `yarn test` runs processes in a package required to pass prior to check-in. Generally includes building, linting, and unit-testing.
+- `yarn tslint` or `yarn tslint:fix` executes tslint on TypeScript in a given package.
+- `yarn unit-tests` execute unit testing for packages.
```
janechu: I would add an "s", making this executes

docs/en/contributing/install.md
chrisdholt: I don't think this was resolved, so I'm reopening this as unresolved. That does not stop the PR from going in with enough approvals, but my response wasn't addressed so I don't think we can mark the conversation as "resolved".

docs/en/contributing/install.md
```diff
@@ -8,33 +8,33 @@ sidebar_label: Install
...
+Clone the repository, `cd` into the project, install <a href="https://github.com/lerna/lerna" target="_blank">Lerna</a> and install dependencies:
 
 ```shell
 git clone https://github.com/Microsoft/fast-dna.git
 cd fast-dna
 yarn global add lerna
 ```
 
-Install all dependencies:
+Installs all the dependencies:
```
chrisdholt: Can you elaborate on the need for this change? This now reads as a description of what the command does rather than the declaration of what to do / how to setup the source repo. If the change here is meant simply as documentation of setting up the source repo, I would expect this to follow something along the lines of line 35.

docs/en/contributing/install.md
```diff
@@ -8,33 +8,33 @@ sidebar_label: Install
...
 
 ```shell
 yarn install
 ```
 
-Prepare the workspaces:
+Builds all the packages:
```
chrisdholt: Can you elaborate on the need for this change? This now reads as a description of what the command does rather than the declaration of what to do / how to setup the source repo. If the change here is meant simply as documentation of setting up the source repo, I would expect this to follow something along the lines of line 35.

package.json
```diff
@@ -28,6 +28,7 @@
     "integration-tests:alpha": "node build/testing/sauce-labs/test-browsers.js alpha",
     "integration-tests:beta": "node build/testing/sauce-labs/test-browsers.js beta",
     "integration-tests:release": "node build/testing/sauce-labs/test-browsers.js release",
+    "prepare": "lerna run prepare",
```
chrisdholt: What is the benefit of aliasing this in the root package and how does it align with this specific PR?

docs/en/contributing/install.md
awentzel: This doesn't change the defaults on the page, it only changes the link that uses HTML Anchor instead of MD.  Your points are valid.  Users still have the option to open in a new tab or new browser as you suggest.  We're not taking that from them.  However, we are attempting to keep users on our site for SEO purposes which better supports our goals.

awentzel: Here is the first resource I found https://www.bluecorona.com/blog/outbound-links-seo-best-practices/  If you still feel strongly about this topic you can raise it to the steering committee.

docs/en/contributing/install.md
awentzel: If the command says what it does, then why have a label at all?   I was attempting to create a more consistent, meaningful, and readable label using proper grammar.  You're free to suggest something else, but, I'm working to improve our documentation where I see opportunity as I run into things.

docs/en/contributing/install.md
awentzel: Same as above.  I intended it to read as a description and improve the wording and grammar. I didn't do that for lines 35+ because they are sub-commands which happen only beneath the hood.  I would argue they are not necessary on this page.

package.json
awentzel: Noted in the description, but, this removes the warning message in the CLI.

awentzel: This all goes towards getting started documentation as I got back into the code with a fresh perspective.

docs/en/contributing/install.md
awentzel: that was my intent. doesn't the command declare what I'm doing?

docs/en/contributing/install.md
chrisdholt: This seems to more closely match the pattern like below where the testing is outlined - essentially, "here are all the commands". I think the initial intent was how to get the packages up and running. I'm fine with either, but having these in their own blocks seems inconsistent now.

package.json
chrisdholt: I'm not aware of the CLI error - what was it?

awentzel: Actually, this also runs the prepare statements automatically saving time.  Basically, a user only has to run `yarn` to both install and prepare. reducing keystrokes.  Nick thought this made sense as well.

docs/en/contributing/install.md
awentzel: Can we just remove all the commands completely, after all, they everyone knows they are in the package.json

package.json
awentzel: There wasn't an error.  It was a warning.  but, after looking again it was just informational.  but, in my search to fix I found this trick to get prepare to autorun.

chrisdholt: Thanks for the FYI- I forgot prepare runs on install which is great.

docs/en/contributing/install.md
chrisdholt: I'm fine with that, but not going to block on this. I think this can go in as it is. The update to the formatting just confused me a bit.

PR #2663, opened 2020-02-06
fix: select calls preventDefault on trigger click

scomea:
# Description
added prevent default call on select trigger click events

## Motivation & context
can cause unwanted side effects, like submitting a form prematurely

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -506,6 +506,7 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
         if (this.props.disabled || e.defaultPrevented) {
             return;
         }
+        e.preventDefault();
```
chrisdholt: should we add a test for this?

scomea: done

PR #2665, opened 2020-02-08
feat: convert form to use the message system

### LINKED ISSUE #2483: Add the messaging system to the Form
The [messaging system](https://github.com/microsoft/fast-dna/issues/2482) should be added to the `Form` component as a property and should replace all data manipulation.

This should be merged into the `releases/tooling-breaking-changes-v2` branch.

Relies on #2482

### LINKED ISSUE #2523: Rename controlPlugins to controls in the Form properties and make it required, provide a default export of the current default type controls
# Abstract

In order to make the Form as light as possible, controls that make up standard JSON schema format should be exported as default, but must be added to the component. This will enable other users to still use the Form component out of the box, but not have the Form ship with unused controls if they intend to replace them.

## Implementation

The user should implement the Form in the following pattern:

```jsx
import { Form, defaultFormControls } from "@microsoft/fast-tooling-react";

<Form
    // other properties
    controls={defaultFormControls}
/>
```

## Requirements
1. Rename `controlPlugins` to `controls`
2. Remove the defaultFormControls as part of the Form and create a named array of them called `defaultFormControls`
3. The react children control is not part of JSON schema, it will be exported as another secondary export as noted in #2524

### LINKED ISSUE #2524: Remove childOptions from the Form
# Abstract

Child options should no longer be required by the Form as the only item that needs access to `childOptions` is the children control, which will soon be imported separately and not as a default control as noted in #2523. This means that `childOptions` should not need to be passed to the Form itself.

## Implementation

The new children control should be offered as an export and should be incorporated in the following way:

```jsx
import { ChildrenControl, Form } from "@microsoft/fast-tooling-react";

<Form
    // other properties
    controls={defaultFormControls.concat([
        new StandardControlPlugin({
            id: "react-children",
            control: (config) => {
                return <ChildrenControl {...config} options={childOptions} />;
            }
        })
    ]}
/>
```

JSON schema:
```json
"children": {
    "formControlId": "react-children",
}
```

janechu:
# Description

These changes update the Form component from @microsoft/fast-tooling-react to use the new tooling message system.

## Motivation & context

In an effort to sync the components in the tooling package, any components sharing data will be using the message system as a means of sending and receiving updates on data and navigation.

Closes #2483, Closes #2523, Closes #2524 

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/src/form/custom-controls/control.children.tsx
```diff
@@ -155,6 +155,19 @@ class ChildrenControl extends React.Component<
         }
     }
 
+    /**
+     * Get React's default children
+     */
+    // private getReactDefaultChildren(): FormChildOptionItem[] {
+    //     return [
```
nicholasrice: Needed?

packages/fast-tooling-react/src/form/form.props.ts
```diff
@@ -25,42 +25,16 @@ export type FormTag = "form" | "div";
  */
 export interface FormProps {
     /**
-     * The optional class name
+     * The message system
+     * used for sending data to the message system
      */
-    className?: string;
```
nicholasrice: You might consider abstracting this to a interface that your form knows how to interact with. That way this doesn't necessarily need to be a web-worker and is generally more extensible.

janechu: This is a really good point that I hadn't thought about, I've filed an issue to be tackled next: https://github.com/microsoft/fast-dna/issues/2668

PR #2671, opened 2020-02-11
feat: add base version of toolbar component

scomea:
# Description
This change adds a toolbar component to base React component set.  A MSFT styled version will follow.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2581

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/toolbar/README.md
+++ b/packages/fast-components-react-base/src/toolbar/README.md
@@ -1,0 +1,8 @@
+## Toolbar
+A generic menu component implementing the [menu](https://www.w3.org/TR/wai-aria-1.1/#menu) role and the keyboard navigation aligning to that role. The *context menu* will accept any child elements but is built to work with *context menu item* and *divider* components.
+
+### Usage
+To accurately manage document focus, the *context menu* assumes that all rendered children are either an element with a role of `"menuitem"`, `"menuitemradio"`, `"menuitemcheckbox"`, or `"separator"` (an `hr` element is also acceptable). It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *context menu*. Additionally, the *context menu* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
+
+### Accessibility
+The *context menu* considers a child element focusable when it has a role of `"menuitem"`, `"menuitemradio"`, or `"menuitemcheckbox"`, and will manage focus between these elements automatically. 
--- a/packages/fast-components-react-base/src/toolbar/README.md
+++ b/packages/fast-components-react-base/src/toolbar/README.md
@@ -1,8 +1,24 @@
 ## Toolbar
-A generic menu component implementing the [menu](https://www.w3.org/TR/wai-aria-1.1/#menu) role and the keyboard navigation aligning to that role. The *context menu* will accept any child elements but is built to work with *context menu item* and *divider* components.
+A generic toolbar component implementing the [toolbar](https://www.w3.org/TR/wai-aria-1.1/#toolbar) role and the keyboard navigation aligning to that role. The *toolbar* will accept any child elements but will only assign focus to children with roles specified in the component's "focusableRoles" prop.
 
 ### Usage
-To accurately manage document focus, the *context menu* assumes that all rendered children are either an element with a role of `"menuitem"`, `"menuitemradio"`, `"menuitemcheckbox"`, or `"separator"` (an `hr` element is also acceptable). It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *context menu*. Additionally, the *context menu* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
+To accurately manage document focus, the *toolbar* assumes that all only children with the appropriate roles should be focusable. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *toolbar*. Additionally, the *toolbar* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
+
+The default list of valid roles for toolbar items is as follows:
+    "button",
+    "checkbox",
+    "link",
+    "menuitem",
+    "menuitemradio",
+    "menuitemcheckbox",
+    "progressbar",
+    "radio",
+    "searchbox",
+    "slider",
+    "spinbutton",
+    "switch",
+    "textbox",
+
 
 ### Accessibility
-The *context menu* considers a child element focusable when it has a role of `"menuitem"`, `"menuitemradio"`, or `"menuitemcheckbox"`, and will manage focus between these elements automatically. 
+The *toolbar* considers a child element focusable when it has a role of that is included in the component's "focusableRoles" prop, and will manage focus between these elements automatically.  
--- a/packages/fast-components-react-base/src/toolbar/README.md
+++ b/packages/fast-components-react-base/src/toolbar/README.md
@@ -2,7 +2,7 @@
 A generic toolbar component implementing the [toolbar](https://www.w3.org/TR/wai-aria-1.1/#toolbar) role and the keyboard navigation aligning to that role. The *toolbar* will accept any child elements but will only assign focus to children with roles specified in the component's "focusableRoles" prop.
 
 ### Usage
-To accurately manage document focus, the *toolbar* assumes that all only children with the appropriate roles should be focusable. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *toolbar*. Additionally, the *toolbar* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
+To accurately manage document focus, the *toolbar* assumes that only children with the appropriate roles should be focusable. It also expects that these elements are passed as *direct* children. Wrapping all toolbar items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *toolbar*. Additionally, the *toolbar* expects that all focusable toolbar children implement the `tabIndex` property and apply that property to the underlying DOM.
 
 The default list of valid roles for toolbar items is as follows:
     "button",
--- a/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
+++ b/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
@@ -1,7 +1,7 @@
 import React from "react";
 import {
-    ToolbarClassNameContract,
     ManagedClasses,
+    ToolbarClassNameContract,
 } from "@microsoft/fast-components-class-name-contracts-base";
 import { Orientation } from "@microsoft/fast-web-utilities";
 
--- a/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
+++ b/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
@@ -20,6 +20,7 @@ export interface ToolbarHandledProps extends ToolbarManagedClasses {
 
     /**
      * The index of the child which gets focused when the toolbar initially gets focus
+     * If unset the toolbar will focus on the first valid focusable child.
      */
     initialFocusIndex?: number;
 
@@ -41,7 +42,7 @@ export interface ToolbarHandledProps extends ToolbarManagedClasses {
 
     /**
      * Whether child items are focusable when disabled.
-     * Detecting disabled elements is based onelements having a "aria-disabled" tag set to true when disabled
+     * Detecting disabled elements is based on elements having a "aria-disabled" tag set to true when disabled
      * default is true
      */
     allowFocusOnDisabledItems?: boolean;
--- a/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
+++ b/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
@@ -1,0 +1,50 @@
+import React from "react";
+import {
+    ToolbarClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-base";
+import { Orientation } from "@microsoft/fast-web-utilities";
+
+export interface ToolbarManagedClasses extends ManagedClasses<ToolbarClassNameContract> {}
+export interface ToolbarUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+export interface ToolbarHandledProps extends ToolbarManagedClasses {
+    /**
+     * The toolbar children
+     */
+    children?: React.ReactNode;
+
+    /**
+     * When true, the toolbar will immediately place focus on the appropriate focusable child when the toolbar mounts
+     */
+    enableAutoFocus?: boolean;
+
+    /**
+     * The index of the child which gets focused when the toolbar initially gets focus
+     */
+    initialFocusIndex?: number;
+
+    /**
+     * Whether the toolbar is horizontal or vertical.  Default is horizontal.
+     * Note that actual layout of the toolbar is controlled via css styling and
+     * this prop mostly affects whether navigation is controlled by the up/down arrows
+     * or the left/right arrows
+     */
+    orientation?: Orientation;
+
+    /**
+     * The toolbar component will only pass focus to children who have been assigned valid aria roles -
+     * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques
+     * The default list isToolbar.DefaultFocusableRoles and contains most "widget" types.
+     * Authors can pass their own list of valid roles instead.
+     */
+    focusableRoles?: string[];
+
+    /**
+     * Whether child items are focusable when disabled.
+     * Detecting disabled elements is based onelements having a "aria-disabled" tag set to true when disabled
+     * default is true
+     */
+    allowFocusOnDisabledItems?: boolean;
+}
+
+export type ToolbarProps = ToolbarHandledProps & ToolbarUnhandledProps;
--- a/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
+++ b/packages/fast-components-react-base/src/toolbar/toolbar.props.ts
@@ -42,7 +42,7 @@ export interface ToolbarHandledProps extends ToolbarManagedClasses {
 
     /**
      * Whether child items are focusable when disabled.
-     * Detecting disabled elements is based on elements having a "aria-disabled" tag set to true when disabled
+     * Detecting disabled elements is based on elements having a "aria-disabled" tag or disabled attribute set to true when disabled
      * default is true
      */
     allowFocusOnDisabledItems?: boolean;
--- a/packages/fast-components-react-base/src/toolbar/toolbar.schema.ts
+++ b/packages/fast-components-react-base/src/toolbar/toolbar.schema.ts
@@ -5,7 +5,35 @@ export default {
     type: "object",
     id: "@microsoft/fast-components-react-base/toolbar",
     formPluginId: "@microsoft/fast-components-react-base/toolbar",
-    properties: {},
+    properties: {
+        enableAutoFocus: {
+            title: "Enable auto focus",
+            type: "boolean",
+            default: "false",
+        },
+        initialFocusIndex: {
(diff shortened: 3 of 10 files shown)


packages/fast-components-react-base/src/toolbar/README.md
```diff
@@ -0,0 +1,24 @@
+## Toolbar
+A generic toolbar component implementing the [toolbar](https://www.w3.org/TR/wai-aria-1.1/#toolbar) role and the keyboard navigation aligning to that role. The *toolbar* will accept any child elements but will only assign focus to children with roles specified in the component's "focusableRoles" prop.
+
+### Usage
+To accurately manage document focus, the *toolbar* assumes that all only children with the appropriate roles should be focusable. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *toolbar*. Additionally, the *toolbar* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
```
chrisdholt:
```suggestion
To accurately manage document focus, the *toolbar* assumes that only children with the appropriate roles should be focusable. It also expects that these elements are passed as *direct* children. Wrapping all menu items inside of a `React.Fragment` (or some other single element) will throw off the internal focus tracking of the *toolbar*. Additionally, the *toolbar* expects that all menu item children implement the `tabIndex` property and apply that property to the underlying DOM.
```

chrisdholt: This seems to imply that the children of a toolbar are always menu items of some kind. Toolbar is fundamentally different than a menubar - have these been convoluted?

packages/fast-components-react-base/src/toolbar/README.md
```diff
@@ -0,0 +1,24 @@
...
+    "menuitem",
+    "menuitemradio",
+    "menuitemcheckbox",
+    "progressbar",
+    "radio",
+    "searchbox",
+    "slider",
+    "spinbutton",
+    "switch",
+    "textbox",
```
chrisdholt: Can you link to where this is coming from? I can't seem to find the reference here: https://w3c.github.io/aria-practices/#toolbar

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * When true, the toolbar will immediately place focus on the appropriate focusable child when the toolbar mounts
+     */
+    enableAutoFocus?: boolean;
+
+    /**
+     * The index of the child which gets focused when the toolbar initially gets focus
+     */
+    initialFocusIndex?: number;
```
chrisdholt: If there is a default, can we call it out here?

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,50 @@
...
+     * The toolbar component will only pass focus to children who have been assigned valid aria roles -
+     * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques
+     * The default list isToolbar.DefaultFocusableRoles and contains most "widget" types.
+     * Authors can pass their own list of valid roles instead.
+     */
+    focusableRoles?: string[];
+
+    /**
+     * Whether child items are focusable when disabled.
+     * Detecting disabled elements is based onelements having a "aria-disabled" tag set to true when disabled
```
chrisdholt: If I put a button into the toolbar, do I need to set `aria-disabled` on that instead of `disabled`?

chrisdholt:
```suggestion
     * Detecting disabled elements is based on elements having a "aria-disabled" tag set to true when disabled
```

packages/fast-components-react-base/src/toolbar/toolbar.schema.ts
```diff
@@ -0,0 +1,45 @@
...
+            title: "Allow focus on disabled items",
+            type: "boolean",
+            default: "true",
+        },
+    },
+    reactProperties: {
+        children: {
+            title: "Children",
+            type: "children",
+            formPluginId: "@microsoft/fast-components-react-base/context-menu/children",
```
chrisdholt: this is referencing context menu children, that seems incorrect

packages/fast-components-react-base/src/toolbar/toolbar.spec.tsx
```diff
@@ -0,0 +1,249 @@
...
+ * Configure Enzyme
+ */
+configure({ adapter: new Adapter() });
+
+describe("context menu", (): void => {
+    test("should have a displayName that matches the component name", () => {
+        expect(`${DisplayNamePrefix}${(Toolbar as any).name}`).toBe(Toolbar.displayName);
+    });
+
+    test("should have correct role attribute 'menu'", () => {
```
chrisdholt: this should have a role of toolbar.

packages/fast-components-react-base/src/toolbar/toolbar.spec.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+    test("should not place focus any child without a the proper role", (): void => {
+        const rendered: any = mount(
+            <Toolbar>
+                <button />
+                <button role="menuitem">two</button>
+                <button />
+                <button role="menuitem">three</button>
+                <button />
+                <button role="menuitem">four</button>
```
chrisdholt: Should this be role based? Is it possible that you could have an element with an implicit role that should receive focus?

packages/fast-components-react-base/src/toolbar/toolbar.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+            <button role="button">Item 1</button>
+            <button role="button">Item 2</button>
+            <button role="button">Item 3</button>
+        </Toolbar>
+    ))
+    .add("Ignore elements without proper role set", () => (
+        <Toolbar>
+            <button role="button">Item 1</button>
+            <button>Button without a role</button>
+            <button role="button">Item 3</button>
```
chrisdholt: I don't think we can implement this way. The implicit role of button is already button. This implementation goes outside the current recommended standards approach of including implicit roles on semantic elements. I think we need to reconsider the approach if this is the current thinking.

scomea: I was wondering about that.  Do we add checks for a number of semantic elements by checking their type and support those in addition to defined roles, or some other approach?

chrisdholt: I don't know that the current approach will work as is - with that said, I don't have an immediate resolution for how to go about it. This may be a good opportunity to gather feedback from collaborators and the community either in the Discord channels or via linking this with the spec, outlining the issue and pinging the collaborators.

packages/fast-components-react-base/src/toolbar/README.md
scomea: It's a first crack at a default list of widget roles that would make sense, not based on any toolbar spec but derived by looking at widget roles - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: "If unset the toolbar will focus on the first valid focusable child."

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: That's what we are detecting  Context menu detects disabled children in this way.  Can/should we try to detect the disabled attribute as well?

scomea: yep

packages/fast-components-react-base/src/toolbar/toolbar.schema.ts
scomea: copy paste error...

packages/fast-components-react-base/src/toolbar/toolbar.spec.tsx
scomea: yep

packages/fast-components-react-base/src/toolbar/README.md
scomea: yep

scomea: corrected

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
chrisdholt: Context Menu has a fixed set of things that can go in it, which is why I believe we are looking for the role. There are not semantic elements for `menuitem`, `menuitemcheckbox`, `menuitemradio` so the role would have to be present. This scenario requires a much more broad application.

scomea: Yes, it does.  But we do need to be able to differentiate between focusable and non-focusable elements in the toolbar.  Another approach could be to have authors specifically label or declare those elements in props somehow.

scomea: ...but this is about detecting disabled state.

chrisdholt: Sure - I just think we likely need to consider other options here. Requiring a specific implementation of aria, which is not required normally, goes against web standards and best practices for implementors. IE, adding a role of button on a button or adding an attribute of aria-disabled on an input or button which has a supported attribute of "disabled"

scomea: improved disabled state detection to include "disabled" attribute

packages/fast-components-react-base/src/toolbar/toolbar.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * Ensure we always validate our internal state on item focus events, otherwise
+     * the component can get out of sync from click events
+     */
+    private handleItemFocus = (e: React.FocusEvent<HTMLElement>): void => {
+        const target: Element = e.currentTarget;
+        const focusIndex: number = this.domChildren().indexOf(target);
+
+        if (focusIndex === -1) {
```
bheston: How could this happen? Do we need this check, or does it hide potentially some other tracking issues?

packages/fast-components-react-base/src/toolbar/toolbar.tsx
```diff
@@ -0,0 +1,274 @@
...
+            return;
+        }
+
+        if (this.isDisabledElement(target)) {
+            target.blur();
```
bheston: Would this cause focus and blur to fire on the control? Is it possible styling or logic might happen in the child that would be different from a normal disabled case where it never got focus anyway? Want to make sure we don't see flashes like styling into focus state and then not really fast.

packages/fast-components-react-base/src/toolbar/toolbar.tsx
```diff
@@ -0,0 +1,274 @@
...
+        });
+    };
+
+    /**
+     * Determines if a given element should be focusable by the menu
+     */
+    private isFocusableElement = (element: Element): element is HTMLElement => {
+        return (
+            element instanceof HTMLElement &&
+            this.props.focusableRoles.indexOf(element.getAttribute("role")) !== -1 &&
```
bheston: How is it possible that there's no way to get the default role for an element? That seems like a miss in the spec to me.

packages/fast-components-react-base/src/toolbar/toolbar.stories.tsx
eljefe223: I there a role that we could assign that would exclude an element. What about `role="none"`?

https://www.scottohara.me/blog/2018/05/05/hidden-vs-none.html

chrisdholt: I'm not sure we should leverage the roles model to remove here, there could potentially be issues with AT by doing so. If we use role="presentation" we'll need to ensure that we treat it as such for non-AT users as well. This would be the same with something like "aria-hidden" in this instance.

scomea: I think we can get away from the whole roles based approach by using tappable's isFocusable function or something similar.  Working on it.

packages/fast-components-react-base/src/toolbar/toolbar.tsx
scomea: Elements like "div" would have not any default role.  But my next iteration should remove the roles based checks.

packages/fast-components-react-base/src/toolbar/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+### Accessibility
```
janechu: When this is not the only toolbar in the application, how do you intend to add a label? This should also be documented as part of accessibility.

packages/fast-components-react-base/src/toolbar/README.md
janechu: From the WAI Aria spec lined above:
> Authors MAY manage focus of descendants for all instances of this role

We also need a way of opting out of focus trapping. The key word here is "may".

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface ToolbarHandledProps extends ToolbarManagedClasses {
+    /**
+     * The toolbar children
+     */
+    children?: React.ReactNode;
+
+    /**
+     * When true, the toolbar will immediately place focus on the appropriate focusable child when the toolbar mounts
+     */
+    enableAutoFocus?: boolean;
```
janechu: This is not necessary, should autofocus be needed on an item in the toolbar, the user can place it on the children themselves.

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,51 @@
...
+} from "@microsoft/fast-components-class-name-contracts-base";
+import { Orientation } from "@microsoft/fast-web-utilities";
+
+export interface ToolbarManagedClasses extends ManagedClasses<ToolbarClassNameContract> {}
+export interface ToolbarUnhandledProps extends React.HTMLAttributes<HTMLDivElement> {}
+export interface ToolbarHandledProps extends ToolbarManagedClasses {
+    /**
+     * The toolbar children
+     */
+    children?: React.ReactNode;
```
janechu: It is not necessary to enumerate children in the interface as it is part of the React.Component interface for props.

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * When true, the toolbar will immediately place focus on the appropriate focusable child when the toolbar mounts
+     */
+    enableAutoFocus?: boolean;
+
+    /**
+     * The index of the child which gets focused when the toolbar initially gets focus
+     * If unset the toolbar will focus on the first valid focusable child.
+     */
+    initialFocusIndex?: number;
```
janechu: How does the user know what the initial focus index should be?

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,51 @@
...
+     */
+    focusableRoles?: string[];
+
+    /**
+     * Whether child items are focusable when disabled.
+     * Detecting disabled elements is based on elements having a "aria-disabled" tag or disabled attribute set to true when disabled
+     * default is true
+     */
+    allowFocusOnDisabledItems?: boolean;
+}
```
janechu: I would like to see a shortcut key as well, as referenced in https://www.w3.org/TR/wai-aria-practices/#toolbar

> In applications where quick access to a toolbar is important, such as accessing an editor's toolbar from its text area, a documented shortcut key for moving focus from the relevant context to its corresponding toolbar is recommended.

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,51 @@
...
+     */
+    orientation?: Orientation;
+
+    /**
+     * The toolbar component will only pass focus to children who have been assigned valid aria roles -
+     * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques
+     * The default list isToolbar.DefaultFocusableRoles and contains most "widget" types.
+     * Authors can pass their own list of valid roles instead.
+     */
+    focusableRoles?: string[];
```
janechu: Does this mean that if I do:

```tsx
<Toolbar>
    <button>Click me</button>
</Toolbar>
```

Focus will not happen unless I include on the button `role={"button"}`

packages/fast-components-react-base/src/toolbar/README.md
scomea: the second iteration of the component takes a different approach, new pr coming soon

packages/fast-components-react-base/src/toolbar/toolbar.spec.tsx
scomea: next version not role based.

packages/fast-components-react-base/src/toolbar/toolbar.stories.tsx
scomea: new pr on second version coming shortly

packages/fast-components-react-base/src/toolbar/toolbar.spec.tsx
```diff
@@ -0,0 +1,249 @@
...
+                <div role="menuitem">two</div>
+            </Toolbar>
+        );
+
+        rendered.childAt(0).simulate("keydown", { keyCode: keyCodeArrowLeft });
+
+        expect(rendered.state("focusIndex")).toBe(1);
+    });
+
+    test("should place focus any child with the proper role", (): void => {
```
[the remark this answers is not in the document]
scomea: test no longer relevant

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Whether child items are focusable when disabled.
+     * Detecting disabled elements is based on elements having a "aria-disabled" tag or disabled attribute set to true when disabled
+     * default is true
```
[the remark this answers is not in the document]
scomea: this option will be removed in next iteration, disabled won't get focus

packages/fast-components-react-base/src/toolbar/toolbar.tsx
scomea: We manage focus through use of keystrokes, we have to react to focus changes from other methods (code or mouse click), this is where we do that.

packages/fast-components-react-base/src/toolbar/toolbar.tsx
scomea: not doing this in new version, pr soon

packages/fast-components-react-base/src/toolbar/toolbar.tsx
scomea: we no longer depend on roles in next iteration

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: Getting rid of it in next version, will be a new PR but going over feedback so it isn't lost.

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: removed

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: They are the ones populating the toolbard.  Will add some explanatory stuff in next version PR.

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: Added home and end to next version

packages/fast-components-react-base/src/toolbar/toolbar.props.ts
scomea: next version not role based.

PR #2675, opened 2020-02-13
feat(fast-element): add initial source from prototype

EisenbergEffect:
# Description

This PR includes the source from our prototype++ efforts to create a web components micro library for building the next version of fast components.

## Motivation & context

This code should serve as the starting point for creating web component versions of our fast components. This is an early version with many remaining todos, bugs, and perf optimizations. Tests are also lacking since this was a prototype originally. Over the next few weeks we'll put tests in place, address the major bugs and perf issues, and round out the feature set. However, the present code should be enough to get team members moving ahead with basic component implementation.

## Issue type checklist

## Process & policy checklist


packages/fast-element/src/ast.ts
```diff
@@ -0,0 +1,41 @@
+export interface IExpressionSource {
```
chrisdholt: I would love to pull the "I" from the interfaces here - it's a convention we've held to with FAST and one I'd prefer to keep if possible. Thoughts on that? I'm happy to create an issue to track that work.

EisenbergEffect: No objections from me. Do you want to do it as part of this PR or as a follow-up?

chrisdholt: I think we can do it as a follow-up. I'll write up an issue.

packages/fast-element/src/directives/repeat.ts
```diff
@@ -0,0 +1,378 @@
...
+            } else {
+                --j;
+            }
+
+            next = views[i + 1];
+            if (next !== void 0) {
+                // view.nodes!.link(next.nodes);
+            } else {
+                // view.nodes!.link(location);
+            }
```
chrisdholt: If we aren't performing work in these do we need to execute them? Could we just comment out the block for now?

packages/fast-element/src/directives/repeat.ts
```diff
@@ -0,0 +1,378 @@
...
+        let next: ISyntheticView;
+        let j = seqLen - 1;
+        let i = newLen - 1;
+        let view: ISyntheticView;
+        for (; i >= 0; --i) {
+            view = views[i];
+            if (indexMap[i] === -2) {
+                view.insertBefore(location);
+            } else if (j < 0 || seqLen === 1 || i !== seq[j]) {
+                // view.attach();
```
chrisdholt: Does the else if need to exist here for now? Could we comment the block out?

EisenbergEffect: We could but this is also placeholder for fixes for some of those "bugs" I mentioned 🤣

packages/fast-element/src/directives/repeat.ts
EisenbergEffect: See comment below.

packages/fast-element/src/directives/repeat.ts
chrisdholt: I assumed so, but I ask :)

packages/fast-element/src/directives/when.ts
```diff
@@ -0,0 +1,81 @@
...
+    public createPlaceholder(index: number) {
+        return DOM.createLocationPlaceholder(index);
+    }
+}
+
+export class WhenBehavior
+    implements IBehavior, IGetterInspector, IPropertyChangeListener {
+    private location: Node;
+    private view: ISyntheticView | null = null;
+    private chachedView?: ISyntheticView;
```
chrisdholt: Should this be `cachedView`?

EisenbergEffect: It's the past tense of the verb "to chacha".
Used in a sentence: "Did you see how @chrisdholt chached last night?"
When you see someone chacha, you have a "chached view". Also, if they are not latino, it's a "synthetic chached view".

packages/fast-element/src/ast.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export class InterpolationExpression implements IExpression {
+    constructor(private parts: (string | Getter)[]) {}
+
+    public evaluate(scope: unknown, context?: IEvaluationContext) {
+        return this.parts
+            .map(x => (typeof x === "string" ? x : x(scope, context!)))
+            .join("");
+    }
+}
```
janechu: Can we get a few comments on the file? I'm not sure what `ast` stands for.

EisenbergEffect: I removed the use of the "ast" term. This would normally stand for Abstract Syntax Tree, but we don't actually have one of those here, which makes it more confusing. I've renamed the file to "expression" which is more accurate. I've also removed some unnecessary code and added basic description docs to each item in the expression.ts file.

PR #2676, opened 2020-02-13
doc(specs): introduce spec template

EisenbergEffect:
# Description

This PR adds a new top-level `specs` folder for storing team specs. A spec template is provided as well as a readme to serve as an index of specs.

## Motivation & context

We'd like to increase transparency over the component design and implementation process, both inside and outside of our team. We'd also like to enable richer design discussions and greater consistency in our components. We believe spending some time writing specs for new components will enable these goals and result in higher-quality components in the end.

A secondary goal is to enable upstreaming our component specs to the open-ui project, helping to improve the open-ui process and content.

## Issue type checklist

- [x] **Documentation** A change that adds/fixes user docs or engineering notes.

## Process & policy checklist

### DIFF
--- a/specs/README.md
+++ b/specs/README.md
@@ -1,9 +1,7 @@
 # Specs
 
-Here you'll find specifications for fast-dna custom elements and other library features.
+Here you'll find specifications for custom elements and other library features.
 
 > **Note:** When authoring a new spec, please use [our standard spec template](./template.md). The PR that introduces your spec should also add a link to this readme in the specs section below.
 
 ## Specs
-
-Coming soon...
--- a/specs/README.md
+++ b/specs/README.md
@@ -1,0 +1,9 @@
+# Specs
+
+Here you'll find specifications for fast-dna custom elements and other library features.
+
+> **Note:** When authoring a new spec, please use [our standard spec template](./template.md). The PR that introduces your spec should also add a link to this readme in the specs section below.
+
+## Specs
+
+Coming soon...
--- a/specs/template.md
+++ b/specs/template.md
@@ -78,7 +78,7 @@
 
 - *Special RTL handling*
 - *Swapping of internal icons/visuals*
-- *Localization of text*
+- *Localization*
 
 ### Security
 
--- a/specs/template.md
+++ b/specs/template.md
@@ -22,7 +22,7 @@
 
 ### Prior Art/Examples
 
-*Screenshots and/or links to existing, canonical or exemplary implementations of the component.*
+*Screenshots and/or links to existing, canonical, or exemplary implementations of the component.*
 
 ---
 
--- a/specs/template.md
+++ b/specs/template.md
@@ -18,7 +18,7 @@
 
 ### Risks and Challenges
 
-*Notable risks or challenges associated with implementing the component. Would we need to make any breaking changes to other parts of fast-dna in order to achieve this component's goals?*
+*Notable risks or challenges associated with implementing the component. Would we need to make any breaking changes in order to achieve this component's goals?*
 
 ### Prior Art/Examples
 
@@ -31,7 +31,7 @@
 *Describe the design of the component, thinking through several perspectives:*
 
 - *A customer using the component on a web page.*
-- *A developer building an app with the component and interacting through HTML/CSS/JS.*
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
 - *A designer customizing the component.*
 
 ### API
@@ -43,7 +43,7 @@
 - *Methods*
 - *Events*
 
-*Consider high and low-level APIs. Attempt to design a powerful and extensible low-level API with a higher-level API for developer/designer ergonomics and simplicity.*
+*Consider high and low-level APIs. Attempt to design a powerful and extensible low-level API with a high-level API for developer/designer ergonomics and simplicity.*
 
 ### Anatomy and Appearance
 
@@ -58,7 +58,7 @@
 
 ## Implementation
 
-*Important aspects of the planned implementation, with careful consideration of web standards and integration.*
+*Important aspects of the planned implementation with careful consideration of web standards and integration.*
 
 ### States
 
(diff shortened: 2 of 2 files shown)


specs/template.md
```diff
@@ -0,0 +1,117 @@
...
+
+*A list of the key features unique to this component.*
+
+### Risks and Challenges
+
+*Notable risks or challenges associated with implementing the component. Would we need to make any breaking changes to other parts of fast-dna in order to achieve this component's goals?*
+
+### Prior Art/Examples
+
+*Screenshots and/or links to existing, canonical or exemplary implementations of the component.*
```
chrisdholt: Should we oxford comma here?

specs/template.md
```diff
@@ -0,0 +1,117 @@
...
+
+### States
+
+*Key component states, valid state transitions, and how interactions trigger a state transition.*
+
+### Accessibility
+
+*Consider the accessibility of the component, including:*
+
+- *Use with Assistive Technology*
```
chrisdholt: I wonder if an example might be helpful here. For instance, I don't think that we should build for specific AT solutions (NVDA, Narrator, Voice Over, JAWS, etc) - and I don't think this is what's being proposed, it's just how I'm reading it. Perhaps detailing that working in the shadow dom has implications on how roles are presented to AT or something?

specs/template.md
```diff
@@ -0,0 +1,117 @@
...
+- *Keyboard Navigation and Focus*
+- *Form Input*
+
+### Globalization
+
+*Consider whether the component has any special globalization needs such as:*
+
+- *Special RTL handling*
+- *Swapping of internal icons/visuals*
+- *Localization of text*
```
chrisdholt: I might suggest just keeping this to "localization".
```suggestion
- *Localization*
```

specs/template.md
EisenbergEffect: Good catch.

specs/template.md
EisenbergEffect: Added a quick note based on your recommendation.

specs/README.md
```diff
@@ -0,0 +1,9 @@
+# Specs
+
+Here you'll find specifications for fast-dna custom elements and other library features.
```
awentzel: Generally, we have always used  FAST-DNA in all caps.  I prefer the reading in lower case, but, as a name we should maintain casing.  I actually think we could just remove fast-dna all together.  "... find specifications for custom"

specs/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+> **Note:** When authoring a new spec, please use [our standard spec template](./template.md). The PR that introduces your spec should also add a link to this readme in the specs section below.
+
+## Specs
+
+Coming soon...
```
awentzel: for web content, generally, I'm not a fan of "coming soon".  I would rather than add it when it's ready.

specs/template.md
```diff
@@ -0,0 +1,118 @@
...
+
+*Primary use cases for this component.*
+  
+### Features
+
+*A list of the key features unique to this component.*
+
+### Risks and Challenges
+
+*Notable risks or challenges associated with implementing the component. Would we need to make any breaking changes to other parts of fast-dna in order to achieve this component's goals?*
```
awentzel: I think we can remove "of fast-dna"

specs/template.md
```diff
@@ -0,0 +1,118 @@
...
+*Screenshots and/or links to existing, canonical, or exemplary implementations of the component.*
+
+---
+
+## Design
+
+*Describe the design of the component, thinking through several perspectives:*
+
+- *A customer using the component on a web page.*
+- *A developer building an app with the component and interacting through HTML/CSS/JS.*
```
awentzel: is "JS" here short for JavaScript or JSS?  I've always spelled out JavaScript.

specs/template.md
```diff
@@ -0,0 +1,118 @@
...
+### API
+
+*The key elements of the component's public API surface:*
+
+- *Component Name*
+- *Props/Attrs*
+- *Methods*
+- *Events*
+
+*Consider high and low-level APIs. Attempt to design a powerful and extensible low-level API with a higher-level API for developer/designer ergonomics and simplicity.*
```
awentzel: The use of "high" and "low" are used, but, then you mention "higher-level".  Curious, if this should be "high-level" as well?  or is the intent that it could just be higher than the low.

specs/template.md
```diff
@@ -0,0 +1,118 @@
...
+- *Slot Names*
+- *Host Classes*
+- *Slotted Content/Slotted Classes*
+- *CSS Parts*
+
+---
+
+## Implementation
+
+*Important aspects of the planned implementation, with careful consideration of web standards and integration.*
```
awentzel: not sure the comma after "implementation," is needed.

specs/template.md
```diff
@@ -0,0 +1,118 @@
...
+*Key component states, valid state transitions, and how interactions trigger a state transition.*
+
+### Accessibility
+
+*Consider the accessibility of the component, including:*
+
+- *Keyboard Navigation and Focus*
+- *Form Input*
+- *Use with Assistive Technology*
+  - e.g. The implications shadow dom might have on how roles are presented to the AT.
```
awentzel: Should "dom" be capitalized as an acronym?  Also, if we use AT as an abbreviation, could we first do it with "Assistive Technology" so readers know from that point on what it stands for.  I had to do a second take on it.

specs/template.md
```diff
@@ -0,0 +1,118 @@
...
+
+*What is the plan for testing the component, if different from the normal path?*
+
+### Tooling
+
+*Are there any special considerations for tooling? Will tooling changes need to be made? Is there a special way to light up this component in our tooling that would be compelling for developers/designers?*
+
+### Documentation
+
+*What additions or changes would need to be made to our user documentation and demos? Are there any architectural/engineering docs we should create as well, perhaps due to some interesting technical challenge or design decisions related to this component?*
```
awentzel: can we replace "would need to be made to our user" with " are needed for user"

PR #2679, opened 2020-02-14
perf(fast-element): improve subscribe/unsubscribe by reducing alloc

EisenbergEffect:
# Description

This PR improves general performance, improves startup metrics, and reduces memory allocation.

## Motivation & context

Tracking observable subscribers previously involved creating a new array and pushing subscribers into that array for each source object. The constant creation of arrays along with the use of `Array#push` creates memory pressure, which has an affect on performance as well.

This PR takes an approach based on some techniques I've used previously in Aurelia. Instead of creating an `Array`, we create a `SubscriberCollection`. The subscriber collection is optimized for the most common scenario of 1 - 3 subscribers per observable property or array. With this in mind, it can store the subscriber in an internal field, allowing it to avoid `Array#push`. If the collection ever exceeds three subscribers, it will spill over into an array.

As part of this PR, I took the opportunity to refactor and improve the API design of all the notification-related classes and interfaces. The result is a clearer design that's more suitable for both property and array observation. Additionally, the amount of code needed to leverage the notification primitives has been reduced in some cases (e.g. `ArrayObserver` can now extend from `SubscriberCollection` and `Controller` can now extend from `PropertyChangeNotifier`).

## Issue type checklist

- [x] **Performance**: A change that improves performance or reduces memory usage.

## Process & policy checklist

### DIFF
--- a/packages/fast-element/src/behaviors/binding-base.ts
+++ b/packages/fast-element/src/behaviors/binding-base.ts
@@ -2,7 +2,7 @@ import { IBehavior } from "./behavior";
 import { IGetterInspector, Observable } from "../observation/observable";
 import { BindingDirective } from "../directives/bind";
 import { DOM } from "../dom";
-import { ISubscriber } from "src/observation/subscriber-collection";
+import { ISubscriber } from "../observation/subscriber-collection";
 
 class ObservationRecord {
     constructor(private source: any, private propertyName: string) {}
--- a/packages/fast-element/src/behaviors/binding-base.ts
+++ b/packages/fast-element/src/behaviors/binding-base.ts
@@ -1,32 +1,22 @@
 import { IBehavior } from "./behavior";
-import {
-    IGetterInspector,
-    IPropertyChangeListener,
-    Observable,
-} from "../observation/observable";
+import { IGetterInspector, Observable } from "../observation/observable";
 import { BindingDirective } from "../directives/bind";
 import { DOM } from "../dom";
+import { ISubscriber } from "src/observation/subscriber-collection";
 
 class ObservationRecord {
     constructor(private source: any, private propertyName: string) {}
 
-    subscribe(listener: IPropertyChangeListener) {
-        Observable.getNotifier(this.source).addPropertyChangeListener(
-            this.propertyName,
-            listener
-        );
+    subscribe(subscriber: ISubscriber) {
+        Observable.getNotifier(this.source).subscribe(subscriber, this.propertyName);
     }
 
-    unsubscribe(listener: IPropertyChangeListener) {
-        Observable.getNotifier(this.source).removePropertyChangeListener(
-            this.propertyName,
-            listener
-        );
+    unsubscribe(subscriber: ISubscriber) {
+        Observable.getNotifier(this.source).unsubscribe(subscriber, this.propertyName);
     }
 }
 
-export abstract class BindingBase
-    implements IBehavior, IGetterInspector, IPropertyChangeListener {
+export abstract class BindingBase implements IBehavior, IGetterInspector, ISubscriber {
     protected source: unknown;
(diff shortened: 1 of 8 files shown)


janechu:
@EisenbergEffect can we make sure to document the fact that using 1-3 subscribers is more performant in the developer documentation? I think this may be important when considering component implementations, especially if a component is expected to be created several times on a page.

packages/fast-element/src/behaviors/binding-base.ts
```diff
@@ -1,32 +1,22 @@
 import { IBehavior } from "./behavior";
-import {
-    IGetterInspector,
-    IPropertyChangeListener,
-    Observable,
-} from "../observation/observable";
+import { IGetterInspector, Observable } from "../observation/observable";
 import { BindingDirective } from "../directives/bind";
 import { DOM } from "../dom";
+import { ISubscriber } from "src/observation/subscriber-collection";
```
janechu: looks like the autocomplete on your text editor unfortunately did not autocomplete a relative path correctly, I think you want`src/rest/of/path` to be `../rest/of/path`

packages/fast-element/src/directives/when.ts
```diff
@@ -3,12 +3,9 @@ import { ITemplate, ICaptureType } from "../template";
...
 import { IExpression, AccessScopeExpression, Getter } from "../expression";
 import { IBehavior } from "../behaviors/behavior";
-import {
-    IPropertyChangeListener,
-    Observable,
-    IGetterInspector,
-} from "../observation/observable";
+import { Observable, IGetterInspector } from "../observation/observable";
 import { BindingDirective } from "./bind";
+import { ISubscriber } from "src/observation/subscriber-collection";
```
janechu: same pathing issue here

packages/fast-element/src/behaviors/binding-base.ts
EisenbergEffect: Fixed.

packages/fast-element/src/directives/when.ts
EisenbergEffect: Got it. Thanks!

packages/fast-element/src/observation/subscriber-collection.ts
```diff
@@ -0,0 +1,94 @@
+export interface ISubscriber {
+    handleChange(source: any, args: any): void;
+}
+
+export class SubscriberCollection {
+    private sub1: ISubscriber | undefined = void 0;
+    private sub2: ISubscriber | undefined = void 0;
+    private sub3: ISubscriber | undefined = void 0;
+    private spillover: ISubscriber[] | undefined = void 0;
```
janechu: might be nice to just stick a comment in here as to why we're doing the 3 subscriber then spillover as you have in the PR description

EisenbergEffect: Added a summary and remarks to the SubscriberCollection that captures this info.

packages/fast-element/src/observation/subscriber-collection.ts
```diff
@@ -0,0 +1,105 @@
...
+            return;
+        }
+
+        if (this.sub3 === subscriber) {
+            this.sub3 = void 0;
+            return;
+        }
+
+        if (this.spillover !== void 0) {
+            const index = this.spillover.indexOf(subscriber);
```
nicholasrice: Should this *also* set `spillover` to undefined if the there are no more items? It seems you could avoid re-entering both this if statement and the statement on line 99 un-necessarily, though I'm not sure how much that really matters.

EisenbergEffect: I think we'd have to do some specific benchmarking around that to determine if it would make a big difference or not. I'd prefer to leave it as is for now.

PR #2681, opened 2020-02-15
feat: adds design-system provider stub component

nicholasrice:
# Description
Adds a temporary implementation of the `DesignSystemProvider` to facilitate component development while the permanent mechanisms for providing recipe values -> custom css properties is being developed.

Adds a custom element that provides a number of recipe values to child components, as well as surfacing a subset of the existing design-system.  @bheston and I will be working to finalize an audit of the values which is why this only includes a subset.
## Issue type checklist

## Process & policy checklist


packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -0,0 +1,72 @@
...
+    cornerRadius: number;
+    disabledOpacity: number;
+    density: number;
+    designUnit: number;
+    elevatedCornerRadius: number;
+    focusOutlineWidth: number;
+    outlineWidth: number;
+}
+
+function camelToKebab(str: string): string {
```
janechu: we have a "spinalCase" function in fast-web-utilities, should we just improve that?

nicholasrice: I'd prefer to wait as this is a temporary implementation and I don't want to spend the time right now writing an extensive function w/ full tests.

PR #2684, opened 2020-02-19
perf(fast-element): new strategy for array observation and repeat

EisenbergEffect:
# Description

This PR re-writes the internal mechanism for array change observation and how the repeat directive works in conjunction with that.

## Motivation & context

When benchmarking the library, most benchmarks looked pretty solid, but a few were way out of range. Here is a snapshot of the perf before this PR:

3 - perf

Notice "swap rows" and "append rows to large table". After this PR, here's what the perf metrics look like:

5 - perf

"swap rows" is close to 10x faster and "append rows to large table" is almost 2x faster. Additionally, "create rows", "replace all rows", and "clear rows" also improved, enabling fast-element to now beat lit-html by a noticeable amount, being only .03 off from hand-optimized vanilla JS.

Additionally, memory allocation was improved. Here's the before:

3 - memory

And the after:

5 - memory

## Issue type checklist

- [x] **Performance**: Improvements to performance or memory consumption.

This is not a breaking change for element authors. It does change how array observation works internally and the data structures related to that. If someone were manually observing an array, they would now receive a collection of splices as opposed to the previous IndexMap.

## Process & policy checklist

## Next Steps

This is the last perf PR for a bit. Now that we're moving out of the prototyping phase, I'll be focusing on documentation, testing, and fixing bugs as they come up.

### DIFF
--- a/packages/fast-element/src/directives/repeat.ts
+++ b/packages/fast-element/src/directives/repeat.ts
@@ -5,14 +5,9 @@ import { DOM } from "../dom";
 import { Observable, IGetterInspector } from "../observation/observable";
 import { BindingDirective } from "./bind";
 import { ISyntheticView } from "../view";
-import {
-    synchronizeIndices,
-    applyMutationsToIndices,
-    ArrayObserver,
-    enableArrayObservation,
-    IndexMap,
-} from "../observation/array-observer";
 import { ISubscriber } from "../observation/subscriber-collection";
+import { ArrayObserver, enableArrayObservation } from "../observation/array-observer";
+import { Splice } from "../observation/array-change-records";
 
 export class RepeatDirective extends BindingDirective {
     behavior = RepeatBehavior;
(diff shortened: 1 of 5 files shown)


packages/fast-element/src/directives/repeat.ts
```diff
@@ -5,14 +5,9 @@ import { DOM } from "../dom";
...
 import { ISyntheticView } from "../view";
-import {
-    synchronizeIndices,
-    applyMutationsToIndices,
-    ArrayObserver,
-    enableArrayObservation,
-    IndexMap,
-} from "../observation/array-observer";
 import { ISubscriber } from "../observation/subscriber-collection";
+import { ArrayObserver, enableArrayObservation } from "../observation/array-observer";
```
awentzel: Curious to our naming conventions?  Is there a difference in types between ArrayObserver vs enableArrayObservation?  Also, I like how you alphabetized the parameters.

EisenbergEffect: In this case `ArrayObserver` is a class, while` enableArrayObservation` is a function. Since the repeater and array observation are a large-ish feature, I've designed it to be tree shakeable. So, before you can use it, you need to import and call `enableArrayObservation`. This will happen automatically any time a `repeat` directive is used, so no need to worry about it in practical use.

EisenbergEffect:
@awentzel I'll add some more code comments throughout. Most of what is covered in this PR is semi-private API space as it mostly pertains to the engine that is powering the `repeat` directive. I think the comments would definitely help contributors, less so element authors.

EisenbergEffect:
Ping @awentzel for CircleCI issue 😢

Issue #2682, opened 2020-02-18
Carousel sequence indicators have no visual affordance when focused


chrisdholt:
**Describe the bug; what happened?**
Carousel sequence indicators have no visual affordance when focused. This is an accessibility violation and fairly severe, especially when autoplay is enabled. While there is a visual indicator for which slide you are on, there is no visual indicator for when you are focused on the control itself.

**What are the steps to reproduce the issue?**
1. Go to https://explore.fast.design/components/carousel
2. Navigate through the view using the keyboard
3. Notice there is no VFI for sequence indicators

**What behavior did you expect?**
A VFI exists and is visible when sequence indicator has focus

**In what environment did you see the issue?**
All Environments

bheston:
Updated the design to include typical focus indicator. Note that spacing increased between items to allow for this.
![image](https://user-images.githubusercontent.com/47367562/74853599-00130b00-52f3-11ea-8c9b-66c25a8509af.png)

PR #2686, opened 2020-02-20
fix: update sequence indicators to have visual focus styling

eljefe223:
# Description
closes: #2682 

Added an `after` pseudo element to handle focus visual. Updated stories to use simple state handler to prevent autoplay when carousel is focued.

Before:
Screen Shot 2020-02-20 at 12 18 44 PM

After:
image

Screen Shot 2020-02-20 at 12 18 25 PM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/carousel/index.ts
+++ b/packages/fast-components-styles-msft/src/carousel/index.ts
@@ -1,11 +1,17 @@
 import { DesignSystem, DesignSystemResolver } from "../design-system";
 import { ComponentStyles, CSSRules } from "@microsoft/fast-jss-manager";
-import { directionSwitch, format, toPx } from "@microsoft/fast-jss-utilities";
+import {
+    directionSwitch,
+    format,
+    toPx,
+    applyFocusVisible,
+} from "@microsoft/fast-jss-utilities";
 import {
     neutralFillStealthHover,
     neutralFillStealthRest,
     neutralForegroundRest,
     neutralOutlineRest,
+    neutralFocus,
 } from "../utilities/color";
 import { CarouselClassNameContract } from "@microsoft/fast-components-class-name-contracts-msft";
 import { designUnit, outlineWidth } from "../utilities/design-system";
@@ -41,6 +47,13 @@ const lightModeNeutralOutlineRest: DesignSystemResolver<string> = neutralOutline
     (): string => white
 );
 
+const darkModeNneutralFocus: DesignSystemResolver<string> = neutralFocus(
+    (): string => black
+);
+const lightModeNneutralFocus: DesignSystemResolver<string> = neutralFocus(
+    (): string => white
+);
+
 function flipperStyles(): CSSRules<{}> {
     return {
         position: "absolute",
@@ -66,7 +79,7 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
     carousel_slides: {},
     carousel_sequenceIndicators: {
         position: "absolute",
-        bottom: "8px",
+        bottom: "4px",
         display: "inline-flex",
         padding: "0",
         "text-align": "center",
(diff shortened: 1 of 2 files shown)


packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -80,10 +93,26 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
...
         display: "inline-block",
-        padding: "0 2px",
+        padding: "4px",
+        position: "relative",
         "&:focus": {
             outline: "none",
         },
+        ...applyFocusVisible<DesignSystem>({
+            "&::after": {
+                opacity: "1",
```
chrisdholt: Is opacity needed here for some other state where it's hidden, or is this the result of copy/paste :)?

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -80,10 +93,26 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
...
+                position: "absolute",
+                border: "2px solid transparent",
```
chrisdholt: should this map to the design system? Is the border for the indicators mapped? If so, I think this should follow as well.

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -80,10 +93,26 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
...
+                "border-radius": "40px",
+                top: "0",
+                right: "0",
+                left: "0",
+                bottom: "0",
+                content: "''",
+                display: "block",
+                transition: "all 0.05s ease-in-out",
```
chrisdholt: Can we check with design on the transition value here, this seems barely perceivable at 50 milliseconds.

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -80,10 +93,26 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
     },
     carousel_sequenceIndicator: {
         display: "inline-block",
-        padding: "0 2px",
+        padding: "4px",
```
chrisdholt: Is this change related or unrelated?

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -66,7 +79,7 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
     carousel_slides: {},
     carousel_sequenceIndicators: {
         position: "absolute",
-        bottom: "8px",
+        bottom: "4px",
```
chrisdholt: Is this change related or unrelated?

packages/fast-components-styles-msft/src/carousel/index.ts
eljefe223: I wasn't sure entirely. @bheston The sequence indicators have a hard coded width and the height maps to `designUnit` since these elements are so close to them bottom what happens when density increases? Seems that if the height and focus border increase they will clip outside the bounds of the carousel or get very close.

Just bumped the design to 3 and carousel looks very wrong:
image

awentzel:
I would suggest updating the title to express what the fix is as opposed to what the problem is.  For example, fix: update sequence indicators to have visual focus styling or something to that effect.  this way when reading the changelog file, it's clear on exactly what change has been implemented.  Most people don't search for changelog's for problems and this would improve our capability in the future when we surface more dynamic meaningful changelog history.

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -41,6 +47,13 @@ const lightModeNeutralOutlineRest: DesignSystemResolver<string> = neutralOutline
     (): string => white
 );
 
+const darkModeNneutralFocus: DesignSystemResolver<string> = neutralFocus(
```
bheston: Can we call these `themeDark` and `themeLight` instead to match the style names below. Those names were meant to be based on the setting for the content of the slides rather than the overall UI color, which we have reserved "mode" for.

packages/fast-components-styles-msft/src/carousel/index.ts
```diff
@@ -172,11 +200,17 @@ const styles: ComponentStyles<CarouselClassNameContract, DesignSystem> = {
...
             },
+            "&::after": {
+                "border-color": darkModeNneutralFocus,
+            },
             "&$carousel_sequenceIndicator__active": {
                 "&::before": {
                     background: darkModeNeutralFillStealthRest,
                     ...highContrastHighlightColorIndicator,
                 },
+                "&::after": {
```
bheston: Isn't / shouldn't this be inherited? Any way to not declare twice?

PR #2688, opened 2020-02-20
docs(fast-element): building components

EisenbergEffect:
# Description

This PR adds basic documentation showing how to use the `fast-element` library to create web components.

## Motivation & context

The goal of the initial version of these docs is to help onboard our own team and accelerate building the next version of fast-components as web components. After the initial draft, I plan to make improvements so that this documentation can become part of our public documentation on our web site. We'll plan to have that ready to publish in sync with our release of the initial components.

## Issue type checklist

- [x] **Documentation**: New documentation or corrections.

## Process & policy checklist


packages/fast-element/docs/building-components.md
```diff
@@ -0,0 +1,75 @@
+# Building Components
+
```
nicholasrice: This is redundant with README.md - needed in this section?

packages/fast-element/docs/building-components.md
```diff
@@ -0,0 +1,75 @@
...
+export class NameTag extends FastElement {
+  @attr greeting: string = 'Hello';
+
+  greetingChanged() {
+    this.shadowRoot!.innerHTML = this.greeting;
+  }
+}
+```
+
+To add attributes to your HTML element, create properties decorated by the `@attr` decorator. All attributes defined this way will be automatically registered with the platform so that they can be updated through the browser's native `setAttribute` API as well as the property. You can optionally add a `$propertyName$Changed` method to your class, and it will be called whenever your property changes, whether it changes through the property or the attribute API.
```
nicholasrice: You're probably on it, but it would be good to have an example similar to the above for the `$propertyName$Changed` method

awentzel:
Should this PR be a draft?

EisenbergEffect:
@awentzel I forgot about GitHub's draft PR feature 😞 That definitely would have been a good choice for this. I can't see a way to change it to a draft after the fact. I'll be finished up with this PR by tomorrow, so it won't hang out too much longer at least.

packages/fast-element/docs/building-components.md
EisenbergEffect: I made some edits to make this clearer.

packages/fast-element/docs/building-components.md
EisenbergEffect: I'm writing this document so that it can also form part of the published docs. Since the readme won't be present in that context, I wanted to make sure there was a short intro in this document as well.

packages/fast-element/README.md
```diff
@@ -1 +1,5 @@
-# FAST Element
\ No newline at end of file
+# FAST Element
+
+The `fast-element` library is a lightweight solution for easily building performant, memory-efficient, standards-compliant web components. Fast elements work in every major browser and can be used in combination with any front-end framework or even without a framework.
```
eljefe223: We capitalize "Web Components" and "Fast Element" in packages/fast-element/docs/building-components.md.

packages/fast-element/docs/building-components.md
```diff
@@ -0,0 +1,734 @@
...
+```
+
+> **WARNING:** Avoid scenarios that require you to directly set HTML, especially when the content is coming from an external source. If you must do this, always sanitize the HTML content using a robust HTML sanitizer library, represented by the use of the `sanitize` function above.
+
+#### Dynamic Attributes
+
+Most HTML attributes have a corresponding property on the HTML element itself, so the method of property binding described above will be the most common mechanism used. However, there are some scenarios, such as with [ARIA attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA), where there is no corresponding element property. For these scenarios, prepend a `$` to the attribute name, and the value will be set with the `setAttribute` API instead of through a property. Additionally, some attributes are known as [boolean attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Boolean_Attributes)  (e.g. required, readonly, disabled). These attributes behave differently form normal attributes, and need special value handling. The templating engine will handle this for you if you prepend the attribute name with a `?`.
```
eljefe223: typo: form

These attributes behave differently **from** normal attributes

EisenbergEffect:
Ok, this one is ready for review 😄

scomea:
Can we write directly to css styles in template rendering? ie. <div style={{height: x.itemHeight}}

EisenbergEffect:
Yes, it would look like this:

```HTML
<div style="height: ${x => x.itemHeight}">...</div>
```

PR #2690, opened 2020-02-21
feat: add toolbar as base component (v0.2)

scomea:
# Description
Significantly reworked version from the previous PR.  This change adds toolbar as a base component.  The primary function of the component is to manage focus and keyboarding within a vertically or horizontally oriented toolbar.  Nested containers within the toolbar are supported by using the ToolbarItemGroup component also included in this PR.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2581

## Issue type checklist:

## Process & policy checklist


packages/fast-components-react-base/src/toolbar/toolbar.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+                <button>Item 5</button>
+            </ToolbarItemGroup>
+        </Toolbar>
+    ))
+    .add("Empty", () => <Toolbar />)
+    .add("Ignore disabled elements", () => (
+        <Toolbar>
+            <button>Item 1</button>
+            <button disabled={true}>Disabled attribute</button>
+            <button disabled={false}>Item 3</button>
```
eljefe223: iis `disabled={false}` necessary or just to explicitly show that this button is not disabled

scomea: Just to make the example explicit and cover all cases.  Item one has no disabled attribute at all.

packages/fast-components-react-base/src/utilities/extract-element.ts
```diff
@@ -5,7 +5,7 @@ import { isNil } from "lodash-es";
  * extracts the html element from React ref or simply returns html element passed in
  */
 export function extractHtmlElement(
-    sourceRef: React.RefObject<HTMLElement> | HTMLElement
+    sourceRef: React.RefObject<any> | HTMLElement
```
nicholasrice: what is necessitating this change?

packages/fast-components-react-base/src/toolbar/toolbar.tsx
```diff
@@ -0,0 +1,355 @@
...
+     *  Increments focus relative to the item with the provided path
+     */
+    private incrementFocus = (focusPath: string, adjustment: number): void => {
+        const focusableItems: HTMLElement[] = this.getFocusableItems();
+
+        if (focusableItems.length === 0) {
+            return;
+        }
+
+        let targetItemIndex: number = -1;
```
nicholasrice: Just as an FYI, there is a "findIndex" method on the array you can use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

This works too, but if you do want to make this more concise you could use it.

packages/fast-components-react-base/src/toolbar/toolbar.tsx
```diff
@@ -0,0 +1,355 @@
...
+                    break;
+
+                case keyCodeArrowUp:
+                    e.preventDefault();
+                    this.incrementFocus(this.state.focusItemPath, -1);
+                    break;
+            }
+        }
+
+        // shortcuts
```
nicholasrice: Is there a reason you initialize `focusableWidgets` to null? Both cases assign this to the product  of `this.getFocusableItems()`

packages/fast-components-react-base/src/toolbar-item-group/toolbar-item-group.props.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface ToolbarItemGroupUnhandledProps
+    extends React.HTMLAttributes<HTMLDivElement> {}
+export interface ToolbarItemGroupHandledProps extends ToolbarItemGroupManagedClasses {
+    /**
+     * The "address" of the group in the toolbar hierarchy.
+     * In toolbars with nested item groups the index can be described with an array
+     * indicating the the index of the child group(s) in succession
+     * ie. [1, 3] places focus on child at index 3 of a group at index 1 of the toolbar root.
+     */
+    itemPath?: number[];
```
nicholasrice: did you explore other API options for this requirement? This feels rather awkward but I'm not sure what to suggest in place of it.

packages/fast-components-react-base/src/toolbar-item-group/toolbar-item-group.tsx
```diff
@@ -0,0 +1,183 @@
...
+} from "./toolbar-item-group.props";
+import Toolbar from "../toolbar/toolbar";
+import { isNil } from "lodash-es";
+import { ToolbarContext, ToolbarContextType } from "../toolbar/toolbar-context";
+
+export interface ToolbarItemGroupState {
+    /**
+     *  the indexes of children of this component that can receive focus
+     */
+    focusableItemIndexes: number[] | null;
```
nicholasrice: Same question as above - are there alternative API's you explored for this functionality?

packages/fast-components-react-base/src/utilities/extract-element.ts
scomea: Because not all refs we'd want to extract a dom element from are necessarily of type HTMLElement.  In this case toolbar has a ref that points to a ToolbarItemGroup:

   private rootElement: React.RefObject<ToolbarItemGroup> = React.createRef<
        ToolbarItemGroup
    >();

packages/fast-components-react-base/src/toolbar/toolbar.tsx
scomea: Yes, much nicer.  Thx

packages/fast-components-react-base/src/toolbar/toolbar.tsx
scomea: redundant, removed.  And renamed to focusableItems (at one point I was referring to focusable things as "widgets" but decided "items" was better, this was an artifact).

packages/fast-components-react-base/src/toolbar-item-group/toolbar-item-group.tsx
scomea: I'm not really worried about this one as it is component internal implementation and not an exposed api.

packages/fast-components-react-base/src/toolbar-item-group/toolbar-item-group.props.ts
scomea: Actually the place for this question is with respect to toolbar props where we have:
initialFocusIndex?: number | number[];

In that case the author would actually need to write the notation.  So maybe in that case we could use id's or a ref?:
initialFocusItemId?: string;

initialFocusItemRef?: React.RefObject<any>

This particular prop should only ever be set programmatically by the parent and not by the author so I've added this to the comment
     * This prop is set programmatically by the parent toolbar/toolbar-item-group and should
     * not be set directly by authors.
What would be better here is a means to set this value in the parent's render fn without suggesting to authors that it's something they should modify.

scomea: I'm going to go ahead and change the initialFocusIndex to a ref based mechanism as that seems better.

nicholasrice: Cool sounds good - yea I saw the notation in a few places, the comment was contextual to all. I think refs would feel nicer, but could potentially have trade-offs. Thanks for looking into it

packages/fast-components-react-base/src/dialog/dialog.tsx
```diff
@@ -212,6 +212,7 @@ class Dialog extends Foundation<DialogHandledProps, DialogUnhandledProps, {}> {
             // only handle tab keystrokes when modal
             return;
         }
+
```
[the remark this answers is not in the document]
scomea: nope, undid the change even though it was just spacing.

marcusturewicz:
Is this still valid? I'd like to see a toolbar.

PR #2692, opened 2020-02-22
feat: add fixed height to horizontal overflow

JamesBurnside:
# Description
Add fixedHeight property to HorizontalOverflow.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2687
When server side rendering - the children heights are not calculated meaning height is always "auto". This feature allows a known fixed height to be used instead.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -207,7 +207,7 @@ describe("horizontal overflow", (): void => {
         expect(rendered.state("itemsHeight")).toBe(100);
     });
     test("should set `itemHeight` to the fixedHeight if children have been passed but a fixedHeight has been set", () => {
-        const rendered = mount(
+        const rendered: any = mount(
             <HorizontalOverflow managedClasses={managedClasses} fixedHeight={100}>
                 <img key="image1" src="https://placehold.it/200x200?text=1" />
             </HorizontalOverflow>
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -206,7 +206,7 @@ describe("horizontal overflow", (): void => {
 
         expect(rendered.state("itemsHeight")).toBe(100);
     });
-    test("should set `itemHeight` to the fixedHeight if children have been passed but a fixedHeight has been set", () => {
+    test("should set `itemHeight` to the fixedHeight if children have been passed and a fixedHeight has been set", () => {
         const rendered: any = mount(
             <HorizontalOverflow managedClasses={managedClasses} fixedHeight={100}>
                 <img key="image1" src="https://placehold.it/200x200?text=1" />
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
@@ -192,12 +192,28 @@ describe("horizontal overflow", (): void => {
 
         expect(renderedWithImages.find("img").length).toBe(6);
     });
-    test("should not set `itemHeight` if no children have been passed", () => {
+    test("should not set `itemHeight` if no children have been passed or fixedHeight set", () => {
         const rendered: any = mount(
             <HorizontalOverflow managedClasses={managedClasses} />
         );
 
         expect(rendered.state("itemsHeight")).toBe(null);
+    });
+    test("should set `itemHeight` to the fixedHeight if no children have been passed but a fixedHeight has been set", () => {
+        const rendered: any = mount(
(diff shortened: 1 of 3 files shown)


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.spec.tsx
```diff
@@ -192,13 +192,29 @@ describe("horizontal overflow", (): void => {
...
         expect(rendered.state("itemsHeight")).toBe(null);
     });
+    test("should set `itemHeight` to the fixedHeight if no children have been passed but a fixedHeight has been set", () => {
+        const rendered: any = mount(
+            <HorizontalOverflow managedClasses={managedClasses} fixedHeight={100} />
+        );
+
+        expect(rendered.state("itemsHeight")).toBe(100);
+    });
+    test("should set `itemHeight` to the fixedHeight if children have been passed but a fixedHeight has been set", () => {
```
eljefe223: NIT: should set `itemHeight` to the fixedHeight if children have been passed **and** a fixedHeight has been set

eljefe223: NIT: should set `itemHeight` to the fixedHeight if children have been passed **and** a fixedHeight has been set

JamesBurnside:
@chrisdholt @nicholasrice @janechu friendly ping for review :)

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -42,6 +42,7 @@ class HorizontalOverflow extends Foundation<
     public static defaultProps: Partial<HorizontalOverflowProps> = {
         managedClasses: {},
         nextItemPeek: 50,
+        fixedHeight: null,
```
nicholasrice: Does this really need to be set to null in the `defaultProps`?  Does omitting this and making the value undefined by default have the same result?

If for whatever reason we *do* need to initialize to null, I think the prop interface should reflect that it can be `null`

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -404,7 +410,9 @@ class HorizontalOverflow extends Foundation<
      * onLoad handler to make sure any children affecting height are accounted for
      */
     private itemsOnLoad = (): void => {
-        const itemsHeight: number = this.getItemMaxHeight();
+        const itemsHeight: number = this.props.fixedHeight
```
nicholasrice: You might want to do a `typeof this.props.fixedHeight === "number"` check rather than a truthy check since explicitly passing 0 will enter the fail case. I'm not sure why you *would* pass 0, but I think we should support it

JamesBurnside: Very good point, will update that

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
JamesBurnside: Having this set to null means it will never be of value `undefined`, I'll update the prop definition to be number | null

packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -192,11 +194,16 @@ class HorizontalOverflow extends Foundation<
             return;
         }
 
-        const itemsHeight: number = this.getItemMaxHeight();
+        const itemsHeight: number =
+            this.props.fixedHeight !== null
+                ? this.props.fixedHeight
+                : this.getItemMaxHeight();
```
janechu: You can just add this logic into `getItemMaxHeight` instead of implementing this same logic twice here and in` itemsOnLoad`

JamesBurnside: My concern with that was the function name: getItemMaxHeight(). This doesn't imply it was return the fixed height and someone may want to use it in the future to actually get the max height of the child items. I could rename it to something like getFixedHeightOrItemMaxHeight, or add another function called getHorizontalOverflowHeight that performs this ternary check

janechu: I think I'd prefer then that we go with your suggestion of an additional method just named `getItemHeight` where this ternary is performed and use that here.

JamesBurnside: Added 👍

JamesBurnside:
This should be ready to complete now 👍

JamesBurnside:
Is there a way to set auto complete after merge than having to update from master manually?

chrisdholt:
> Is there a way to set auto complete after merge than having to update from master manually?

I'll watch this for you and complete :)

JamesBurnside:
> > Is there a way to set auto complete after merge than having to update from master manually?
> 
> I'll watch this for you and complete :)

Thanks Chris! Is there an updated package likely generated soon?

PR #2693, opened 2020-02-24
chore: fast switch web component spec

eljefe223:
# Description
Added spec for `fast-toggle` web component

There is some debate about the proper accessibility of  `fast-toggle`. Our current `<Toggle/>` from `"fast-components-react-msft` uses a checkbox. However, aria has a more appropriate role of [`switch`](https://www.w3.org/TR/wai-aria-1.1/#switch) which is designed to be used for binary state toggles (on and off). However the support from certain screen readers like NVDA does not read role of `switch` according to [the spec](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role).

I created a quick prototype of the possible implementations of toggle switch to see how screen readers read them differently.

https://codepen.io/eljefe223/pen/mdJrozB

Narrator reads the above as follows:
**Mac:**
check / unchecked , Notify by email , checkbox
select / deselect, Notify by email,  Toggle Button
on / off, Notify by email, Switch

**PC:**
check / unchecked , Notify by email , checkbox
Notify by email, Button
on / off, Notify by email, Switch

**NVDA:**
Notify by email, checkbox,  not checked / checked
Notify by email, Toggle button, not pressed / pressed
Notify by email, Toggle button, not pressed / pressed

**Update:** 
It appears that NVDA merged in a change recently that changes how narrator announces the `switch` role. https://github.com/nvaccess/nvda/pull/10782

This would mean that `switch` will now be announced like a checkbox instead of a toggle button. 

NVDA seems to treat the role of `switch` just like a toggle button (a button with an `aria-pressed` attribute). To me either "not checked" / "checked" or "not pressed" / "pressed" are equally bad at describing a toggle switch where built in narrators seem to do the right thing using role of `switch`. With that said I'm proposing our web component for toggle to use a role of `switch`

There are open issues on NVDA to address this.
https://github.com/nvaccess/nvda/issues/9187

## Issue type checklist

## Process & policy checklist


eljefe223:
**Update:** 
It appears that NVDA merged in a change recently that changes how narrator announces the `switch` role. https://github.com/nvaccess/nvda/pull/10782

This would mean that `switch` will now be announced like a checkbox instead of a toggle button.

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+### API
+
+- fast-toggle
+- selected, unselected message, selected message, disabled, labelId, statusMessageId
+- onChange
+
+### Anatomy and Appearance
+
+<div>
```
nicholasrice: This is formatting incorrectly - can you wrap the code sample with code block syntax?

https://github.github.com/gfm/#fenced-code-blocks

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+- *A customer using the component on a web page.*
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+- *A designer customizing the component.*
+
+### API
+
+- fast-toggle
+- selected, unselected message, selected message, disabled, labelId, statusMessageId
+- onChange
```
nicholasrice: Do we also want to provide *slots* for selected and unselected messages, so that the HTML rendered is configurable?

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+*Describe the design of the component, thinking through several perspectives:*
+
+- *A customer using the component on a web page.*
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+- *A designer customizing the component.*
+
+### API
+
+- fast-toggle
```
nicholasrice: do we want to consider calling this "switch" instead of toggle to map to the ARIA name?

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+- selected, unselected message, selected message, disabled, labelId, statusMessageId
+- onChange
+
+### Anatomy and Appearance
+
+<div>
+    <label id="toggle-label">Toggle label</label>
+    <button role="switch" aria-checked="true" aria-labelledby="toggle-label" aria-describedby="status-message">
+        <!-- Status Indicator -->
+        <span/>
```
nicholasrice: How will an app author style these component parts? Which [`::part`](https://developer.mozilla.org/en-US/docs/Web/CSS/::part)s does it support?

specs/toggle.md
nicholasrice: Can you detail when the onChange should fire? What data will be exposed in this event? Is this a custom event?

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+    selectedMessage="Off"
+    labelId="label-id"
+    statusMessageId="status-id"
+>
+    Notify by Email
+</fast-toggle>
+
+### States
+
+*Toggle* can either be **on** or **off** clicking or keyboard interactions will flip it to its opposite state.
```
nicholasrice: *which* keyboard interactions cause it to change state? Will it always change state, or will it only change state when the value is explicitly set by an app author?

nicholasrice:
Have you run the various screen readers across the proposed implementation *in* a WC? It would be good to know that it doesn't interpret anything differently  when the elements exist in the Shadow DOM

specs/toggle.md
EisenbergEffect: Note: if/when we fire an event, we'll want to call it `change` for consistency with the platform. We also want to make sure we keep event names in lower-case, so that they work well with other front-end frameworks that attach events in HTML with certain patterns. I'll make a note on that in my documentation.

specs/toggle.md
eljefe223: We talked about that, I'm open to either but for the sake of getting started I worked with `Toggle`

specs/toggle.md
EisenbergEffect: @eljefe223 In case you need a reference on `::part`, here's some MDN docs: [CSS Shadow Parts](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) We can basically put a `part` attribute on any HTML element in our Shadow DOM and it will enable the user to select that part with a `::part` pseudo-element selector for more style customization.

EisenbergEffect: Here's another helpful link to the part explainer: https://meowni.ca/posts/part-theme-explainer/ I'll try to add some of these to the spec template for future reference.

specs/toggle.md
EisenbergEffect: I think using a slot for this is nice. We can provide fallback content for the slot with a default display. But if someone wants to override that with different text, or even with something like an icon, svg, etc. they can use the slot directly to get what they want.

specs/toggle.md
EisenbergEffect: Historically, have we included the labels inside the components or made that a separate component?

chrisdholt: Historically we have, though I have not seen a design in the past two years which has included these state labels. If this becomes a true switch, an argument could be made that they aren't necessarily needed. I know that @bheston has thoughts on this from a design standpoint.

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+A toggle should allow the following properties. Selected, whether the toggle is selected or not. Selected message, the message that displays when the toggle is selected. Unselected message, the message that displays when toggle is not selected. Disabled, whether the toggle is disabled or not.
+
+### Risks and Challenges
+
+Typically it's best to include a selected and unselected message but it is not required. It's common to see *Toggle* without a message but there are accessibility concerns since it's not necessarily obvious what is **on** and what is **off**. The visual representation of **on** tends to show the state indicator to the right with the background filled and **off** tends to show the state indicator to the left and the background unfilled. However, this pattern is not always consistent.
+
+## Design
+
+*Describe the design of the component, thinking through several perspectives:*
```
EisenbergEffect: Feel free to remove the boilerplate.

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+*Toggle* is a type of checkbox that represents on/off values, as opposed to checked/unchecked values. The *button* should have its role set to *aria-role="switch"* and *aria-checked* should be set to represent the current state of the *Toggle*. *aria-labelledby* should be set to the id of the *Toggle* label and *aria-describedby* should be set to the status message only if values other than "On" and "Off" are passed.
+
+### Globalization
+
+*Toggle* should mirror in RTL languages. 
+
+### Dependencies
+
+*Will implementing the component require taking on any dependencies?*
```
EisenbergEffect: Feel free to remove the boilerplate and just say something like "No dependencies outside of fast-element itself"

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+- *3rd party libraries*
+- *Upcoming standards we need to polyfill*
+- *Dependencies on other fast components or utilities*
+
+*Do any of these dependencies bring along an associated timeline?*
+
+### Documentation
```
EisenbergEffect: Maybe a few quick notes here about what we should put into our customer-facing docs for this component. What examples would be relevant? Any usage guidance? etc.

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+        <span/>
+    </button>
+    <span id="status-message">On</span>
+</div>
+
+## Implementation
+
+<fast-toggle
+    selected={true}
+    unSelectedMessage="On"
```
EisenbergEffect: An important difference to note when moving from React to HTML. HTML is case insensitive, so we'll need to make all attribute names lower-case. We could have `unselected-message` for example, but not `unSelectedMessage`.

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
+# fast-toggle
+
+## Overview
+
+*Toggle* mimics a physical switch that allows users to turn things **on** or **off**.
```
janechu: I think there could be a better word choice than things, maybe just "... a physical switch that can be turned **on**..."

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Use Cases
+
+*Toggle* is used to represent a binary state such and **on** or **off**. If any extra step is required for changes to be effective or an indeterminate state is needed, you should use *checkbox* and corresponding "Apply" button instead. Users shouldn’t need to do something else or go somewhere else in order to experience the *Toggle's* effect.
+
+Typical use cases include, but are not limited to turning a feature on and off or showing or hiding a piece of UI
```
janechu: there should be a ",' after "but are not limited to".

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+  
+### Features
+
+A toggle should allow the following properties. Selected, whether the toggle is selected or not. Selected message, the message that displays when the toggle is selected. Unselected message, the message that displays when toggle is not selected. Disabled, whether the toggle is disabled or not.
```
janechu: Can you make this a bullet list? It's a bit hard to follow as a paragraph.

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Risks and Challenges
+
+Typically it's best to include a selected and unselected message but it is not required. It's common to see *Toggle* without a message but there are accessibility concerns since it's not necessarily obvious what is **on** and what is **off**. The visual representation of **on** tends to show the state indicator to the right with the background filled and **off** tends to show the state indicator to the left and the background unfilled. However, this pattern is not always consistent.
```
janechu: only MSFT is doing background unfilled vs filled, it would seem like other design systems use a muted color instead of empty with an outline. I would make this more ambiguous perhaps "...to the left and the background unfilled *or as a muted or neutral color*."

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+### Accessibility
+https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role
+https://www.w3.org/TR/wai-aria-1.1/#switch
+https://github.com/nvaccess/nvda/issues/9187
+
+*Toggle* is a type of checkbox that represents on/off values, as opposed to checked/unchecked values. The *button* should have its role set to *aria-role="switch"* and *aria-checked* should be set to represent the current state of the *Toggle*. *aria-labelledby* should be set to the id of the *Toggle* label and *aria-describedby* should be set to the status message only if values other than "On" and "Off" are passed.
+
+### Globalization
+
+*Toggle* should mirror in RTL languages.
```
janechu: Can you give more details, such as "... by showing the label text on the other side of the switch and the state indicator aligned to the opposite side to the LTR checked side"

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+</fast-toggle>
+
+### States
+
+*Toggle* can either be **on** or **off** clicking or keyboard interactions will flip it to its opposite state.
+
+### Accessibility
+https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role
+https://www.w3.org/TR/wai-aria-1.1/#switch
+https://github.com/nvaccess/nvda/issues/9187
```
janechu: please make sure these are bullets in the rendered markdown

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
...
+### Documentation
+
+*What additions or changes are needed for user documentation and demos? Are there any architectural/engineering docs we should create as well, perhaps due to some interesting technical challenge or design decisions related to this component?*
+
+---
+
+## Resources
+https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role
+https://www.w3.org/TR/wai-aria-1.1/#switch
+https://github.com/nvaccess/nvda/issues/9187
```
janechu: Consider making this a readable list (needs to be bullet points) and should be words instead of links such as [Linked Item]()

specs/toggle.md
```diff
@@ -0,0 +1,92 @@
+# fast-toggle
```
janechu: Should this just read "Toggle" or if we change it, "Switch"?

specs/toggle.md
eljefe223: Yeah, I think I had this up in "Use Cases" I'll move it from there to here.

specs/toggle.md
eljefe223: @nicholasrice I assume similar concepts like controlled and uncontrolled carry over to WC? If so is it fair to use similar language in WC component specs?

specs/switch.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+## Implementation
+
+```
+<fast-switch
+    checked={true}
+    unchecked-message="On"
+    checked-message="Off"
+    label-id="label-id"
```
nicholasrice: do you need "label-id" and "statusMessageId" in the implementation example anymore?

specs/switch.md
```diff
@@ -0,0 +1,117 @@
...
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+- *A designer customizing the component.*
+
+### API
+
+*Component name:*
+- `fast-switch`
+
+*Attributes:*
+- `checked`: boolean
```
nicholasrice: It might be good to enumerate slots separately since they're not really an attribute and conceptually they're pretty different

specs/switch.md
eljefe223: Right, missed this when I went through and removed. Thanks.

specs/switch.md
```diff
@@ -0,0 +1,117 @@
...
+- `disabled`: boolean
+
+*Slots:*
+- `unchecked-message`
+- `checked-message`
```
nicholasrice: oops you need the default slot here too, which maps to the label

specs/switch.md
```diff
@@ -0,0 +1,118 @@
...
+
+---
+
+## Design
+
+*Describe the design of the component, thinking through several perspectives:*
+
+- *A customer using the component on a web page.*
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+- *A designer customizing the component.*
```
EisenbergEffect: You can remove this boilerplate from the final version of the spec.

specs/switch.md
```diff
@@ -0,0 +1,118 @@
...
+- `disabled`: boolean
+
+*Slots:*
+- `default`
+- `unchecked-message`
+- `checked-message`
+
+*Events:*
+- `change` - fires when component is clicked or "spacebar" is pressed. This is a standard `onChange` event from the `input`
+
```
EisenbergEffect: I really like the explicit sections for name, attrs, slots, events. Nice!

specs/switch.md
```diff
@@ -0,0 +1,110 @@
...
+Typical use cases include, but are not limited to, turning a feature on and off or showing or hiding a piece of UI
+  
+### Features
+
+A switch should allow the following properties:
+- `checked`, whether the switch is cheched or not.
+- `checked-message`, the message that displays when the switch is checked.
+- `unchecked-message`, the message that displays when switch is not checked.
+- `disbaled`, whether the switch is disabled or not.
```
marjonlynch: typo, disabled

PR #2695, opened 2020-02-25
chore: add tree view spec

chrisdholt:
# Description
This PR adds a component specification for Tree View (and subsequently Tree View Items).

## Issue type checklist

## Process & policy checklist


SethDonohue:
Don't forget to add the link to this spec in the specs/README.md
```
**Note:** When authoring a new spec, please use [our standard spec template](./template.md). The PR that introduces your spec should also add a link to this readme in the specs section below.
```

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+
+Selected:
+
+In a single select tree, only one item may be selected at a time. Selection should not follow focus. The following state transitions or interactions are all valid:
+
+- if none of the nodes are selected before the tree receives focus, focus is set on the first node.
+- if a node is selected before the tree receives focus, focus is set on the selected node.
+- clicking on a node selects it
+- clicking on the expand/collapse button selects and performs the expand/collapse behavior
+- when focus is on a node, enter selects the node
```
EisenbergEffect: If I programmatically select an item, does it expand the parent hierarchy?

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+Expanded
+![](./images/tree-item-expanded.png)
+
+Collapsed
+![](./images/tree-item-collapsed.png)
+
+Like tree view, the tree view item has a default slot for its children.
+
+Slots:
+- `content`
```
EisenbergEffect: Maybe a little more explanation of the purpose of each slot. I'm not sure what the `content` slot is. Is that the default slot?

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+
+Parts:
+- root
+- content
+- before-content
+- after-content
+- expand-collapse-glyph
+- child-nodes
+- expanded
+- selected
```
EisenbergEffect: Good call on `expanded` and `selected` as parts. I'm not sure I would have gone that route but it makes sense when you think about targeting styles at the HTML parts that are expanded/selected.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+- `after-content` 
+- `expand-collapse-glyph`
+
+Parts:
+- root
+- content
+- before-content
+- after-content
+- expand-collapse-glyph
+- child-nodes
```
EisenbergEffect: Is this intended to be the container that wraps all the child nodes? So, for example, if I wanted to change the layout of the children, I would target this part and apply flex/grid to it.

janechu:
The images for expanded and collapsed are also with the grey background which is not indicated that it is selected. Could you instead make the images half wide, two columns with the first column as **selected** the other being **unselected**? That way we can see the selected and unselected states for "collapsed", "expanded" and "end node"?

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+- [Atlassian](https://atlaskit.atlassian.com/packages/core/tree/example/static-tree)
+- [Windows (UWP)](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tree-view)
+
+---
+
+## Design
+
+*Describe the design of the component, thinking through several perspectives:*
+
+- *A customer using the component on a web page.*
```
EisenbergEffect: Feel free to remove boilerplate...

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+There are technically two components being presented as part of this specification - tree view, and tree view item. 
+
+As defined by the W3C:
+> A tree view widget presents a hierarchical list. Any item in the hierarchy may have child items, and items that have children may be expanded or collapsed to show or hide the children. For example, in a file system navigator that uses a tree view to display folders and files, an item representing a folder can be expanded to reveal the contents of the folder, which may be files, folders, or both.
+
+### Background
+
+We recently created a React implementation of Tree View and Tree View Item with [PR #2510](https://github.com/microsoft/fast-dna/pull/2510). 
+
+Currently, a react implementation exists which supports a `defaultExpanded` value which sets the default expanded or collapsed state. In the event that the `defaultExpanded` value is false, but there is a selected child the tree will expand automatically. [Issue #2597](https://github.com/microsoft/fast-dna/issues/2597) requests an expanded prop to fully control the expanded/collapsed state.
```
janechu: Are these specs intended to call out the React implementation? I'm not sure how useful this is in a spec.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+## Overview
+
+There are technically two components being presented as part of this specification - tree view, and tree view item. 
+
+As defined by the W3C:
+> A tree view widget presents a hierarchical list. Any item in the hierarchy may have child items, and items that have children may be expanded or collapsed to show or hide the children. For example, in a file system navigator that uses a tree view to display folders and files, an item representing a folder can be expanded to reveal the contents of the folder, which may be files, folders, or both.
+
+### Background
+
+We recently created a React implementation of Tree View and Tree View Item with [PR #2510](https://github.com/microsoft/fast-dna/pull/2510).
```
janechu: I don't think this line belongs in a spec.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
+# Tree View + Tree View Item
+
+## Overview
+
+There are technically two components being presented as part of this specification - tree view, and tree view item.
```
janechu: I think you can leave off "technically" or I would prefer there are two specs and they link to each other where appropriate.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Background
+
+We recently created a React implementation of Tree View and Tree View Item with [PR #2510](https://github.com/microsoft/fast-dna/pull/2510). 
+
+Currently, a react implementation exists which supports a `defaultExpanded` value which sets the default expanded or collapsed state. In the event that the `defaultExpanded` value is false, but there is a selected child the tree will expand automatically. [Issue #2597](https://github.com/microsoft/fast-dna/issues/2597) requests an expanded prop to fully control the expanded/collapsed state.
+
+### Use Cases
+
+Primarily used to present a hierarchical list such as a file system. A tree view can also be served as a flat list where the items will function as navigation.
```
janechu: I'm not sure there should be a call out to navigation, the flat list may not perform navigation.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+- **Selection:** A tree view can support either single select or multi-select behavior for nodes. **I would propose that for the initial spec we start with supporting single select only.*
+
+
+- **Drag & Drop support:** Often, as is in the case of a "file system" example, drag and drop support is desired. This does present [challenges](#drag-&-drop).
+
+### Risks and Challenges
+
+#### Nesting:
+
+Infinite nesting combined with the current visual design presents a styling challenge. Consider a scenario where we have the following tree in pseudo code:
```
janechu: I think you could leave off "pseudo code".

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+</fast-tree-view>
+```
+In this scenario, there is no root node with an expand/collapse glyph. The third item in the tree view is a parent node, which *would* have an expand/collapse glyph to the right. To ensure that the content of the parent node align with the rest of the root nodes, the entire tree will need to visually shift to ensure alignmnet of the tree content. This presents an interesting challenge for how we will communicate these changes to the parent to ensure the styling is appropriately applied. Ultimately, the parent just needs to know that *one* of the items below has child nodes.
+
+#### Drag & Drop
+
+Drag & Drop support poses a unique set of challenges, especially if our goal is to provide out of the box support via the [HTML Drag and Drop API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API). Alternatively, those using our components with React, Vue, Angular, etc may want to use libraries built to work with their specific framework flavor. 
+
+**Considering the challenge that drag & drop would present for any given component, I think it's worth holding on that feature, and taking it up seperately.*
```
janechu: I think it would be better to just file an issue for drag and drop and link it here.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+- extends HTML Element props
+
+**Tree View Item**
+
+- expanded - nodes with child-nodes can either be expanded or collapsed. 
+- selected - the node is selected
+- content - the content for the tree view
+- before-content - slot which precedes content, often a glyph or icon
+- after-content - slot which follows content, often a glyph or icon
+- expand-collapse-glyph - slot for svg, will include a default svg
```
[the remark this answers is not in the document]
janechu: Do we also want just `expand-glyph` and `collapse-glyph` then choose when to show them? It seems like we're assuming that we have a single glyph and will rotate it, but that makes a few assumptions about styling. If you have both then you could use CSS animations to rotate and fade in/out so it works for both scenarios.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+- on-expanded-change - callback for when the expanded state has changed
+- on-selected - callback for when an item has been selected
+
+### Anatomy and Appearance
+
+*Screenshots and/or description of the basic appearance of the component. Outline its structure with a diagram of its visual tree (shadow dom). Enumerate key areas of visual customization, such as:*
```
janechu: This looks like boilerplate.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+The tree view should align to the interaction model provided by the W3C: https://w3c.github.io/aria-practices/#TreeView
+
+### Globalization
+
+The component visuals should support RTL scenarios (per below):
+![](./images/tree-view-rtl.png)
+
+### Test Plan
+
+While testing is still TBD for our web components, I would expect this to align with the testing strategy and not require any additional test support.
```
janechu: I think we should determine this and write a document for component testing we can just link in this section for components.

janechu:
I'm not sure there's a reason to put the markdown in its own folder.

specs/tree-view/tree-view.md
chrisdholt: I think it’s useful in the context of this part of the template which is looking for background context to specific parts of the proposal. Currently if the control is instantiated with a nested child where the Parent tree is not expanded, we will expand the entire parent structure regardless of whether that has been requested. This control is independent of a react version, but I think in this specific case the requests serve to inform whoever builds the control which is beneficial.

specs/tree-view/tree-view.md
chrisdholt: I think that’s fair, it simply gives context to the feature request which follows which I think is helpful in informing the control decisions.

specs/tree-view/tree-view.md
chrisdholt: I think it would be difficult to understand the relationship between these two with two specs and it would make for difficult reading. I’ll pull the problematic word.

specs/tree-view/tree-view.md
chrisdholt: Sure, I’ll call this out more as an example.

specs/tree-view/tree-view.md
chrisdholt: I will pull it off. I did add here to be clear as the examples here don’t align to the spec directly (which was intended).

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+**Tree View Item**
+
+- expanded - nodes with child-nodes can either be expanded or collapsed. 
+- selected - the node is selected
+- content - the content for the tree view
+- before-content - slot which precedes content, often a glyph or icon
+- after-content - slot which follows content, often a glyph or icon
+- expand-collapse-glyph - slot for svg, will include a default svg
+- on-expanded-change - callback for when the expanded state has changed
+- on-selected - callback for when an item has been selected
```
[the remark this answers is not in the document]
nicholasrice: On a somewhat related note - I'm noticing many of [these libraries](https://custom-elements-everywhere.com/) only support lowercase or kebab-case event names. Is that something we want to standardize?

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+- clicking on the expand/collapse button selects and performs the expand/collapse behavior
+- when focus is on a node, enter selects the node
+
+### Accessibility
+
+The tree view should align to the interaction model provided by the W3C: https://w3c.github.io/aria-practices/#TreeView
+
+### Globalization
+
+The component visuals should support RTL scenarios (per below):
```
nicholasrice: Does document direction affect which arrow keys expand / collapse a parent node?

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+
+**Tree View**
+
+- extends HTML Element props
+
+**Tree View Item**
+
+- expanded - nodes with child-nodes can either be expanded or collapsed. 
+- selected - the node is selected
+- content - the content for the tree view
```
nicholasrice: what is the *type* of "content"? is it a string?

specs/tree-view/tree-view.md
nicholasrice: need that custom state API you were talking about @EisenbergEffect!

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+Like tree view, the tree view item has a default slot for its children.
+
+Slots:
+- `content`
+- `before-content`
+- `after-content` 
+- `expand-collapse-glyph`
+
+Parts:
+- root
```
nicholasrice: For clarity, are you purposing that the "default" slot maps to sub-menu items?

```html
<tree-view-item>
    <span slot="content">Tree item text</span>
    <tree-view-item>...</tree-view-item>
    <tree-view-item>...</tree-view-item>
    <tree-view-item>...</tree-view-item>
</tree-view-item>
```

specs/tree-view/tree-view.md
nicholasrice: Also this might be a little odd because it is the element itself that is expanded, collapsed, selected, or unselected, not a "part" of the element

```css
fast-tree-view-item::part(selected) {
  /* This actually targets some DOM node *inside* the tree-view-item, not the item itself
}
```

if we had custom states, I'd expect to be able to write:
```css
tree-view-item:selected { /* target item when selected */ }
tree-view-item:selected::part(content) { /* target a part in item when selected */ }
```

specs/tree-view/tree-view.md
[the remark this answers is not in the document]
chrisdholt: @janechu I intentionally went with *one* to not overload the contract. We have a single slot, but there are several ways you can style and support variations on the glyph. Here are a few ways this could be handled:
- You could use a single glyph and rotate using CSS which would be the default.
- With the expanded callback an implementor knows when a tree item is expanded or not. The glyph provided could be updated based on the expanded/collapsed state of the control. 
- The glyph could be overridden with CSS using the named `:part`. This can change based on the expanded state as well.

specs/tree-view/tree-view.md
chrisdholt: Our react implementation does this currently and my initial thought is defaulting to this behavior again out of convenience. The primary reason I added the references to the feature requests and background above is because we've had feedback that seems to want a fully controlled behavior and I think we should revisit this behavior again as part of this control.

specs/tree-view/tree-view.md
chrisdholt: Sure - I'm torn on content being the default slot vs. having the . If content were the default slot, `children` or `child-nodes` would likely be named. I think my initial intent was for the child nodes to be the default slot, but that disrupts the ease of a flat structure.

specs/tree-view/tree-view.md
chrisdholt: Yes. I think a class here is helpful, though they are all contained within a role of "group"

specs/tree-view/tree-view.md
chrisdholt: Thanks - I think we could add this to the top of the spec; I got a couple comments on here, but I was torn on a couple ways to go:
1. Leave boilerplate so people can know what's been ommitted
2. Omit the boilerplate as it's not used.

If our hope is the second, it might help to have that called out explicitly.

specs/tree-view/tree-view.md
chrisdholt: I thought about that, but I think it's helpful to include here for a couple of reasons. Drag and drop is a feature that we've been asked for previously as well as improvements to "other" (react) implementations. While it may not be initially included, I think it's important to list the feature as part of the spec to show our thought process and reasoning why it may not be included initially. 

I agree with filing an issue, but it may make more sense to do that when we create a PR for this control to close the issue. My primary reasoning is 1) we may end up building in support for D&D once we get further into development of this, and 2) the control will give context for that issue. I think we could create now, but there wouldn't be a ton of value to the issue outside a placeholder IMO.

specs/tree-view/tree-view.md
chrisdholt: Yep, passed a request to Rob above to update the template with guidance to delete BP when not used so it's explicit.

specs/tree-view/tree-view.md
chrisdholt: I'm on point for writing the testing spec - we can update and link later when it goes in I think, yeah?

chrisdholt:
> I'm not sure there's a reason to put the markdown in its own folder.

@janechu I asked @EisenbergEffect about this based on the fact that in many cases, we'll have images included as part of these specs. This is what we came to. I think individual folders make more sense and seem easier to manage than a flat structure of markdown files with an uber `images` folder.

I'm open to feedback here, but this was an intentional decision and not arbitrary.

specs/tree-view/tree-view.md
chrisdholt: I was, but I'm wondering if content should likely become the default slot as it makes the most sense for the broadest use cases (both nested trees and flat trees). Given that, I think I'm going to update the slots to represent `content` as default and `child-nodes` as named.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+![](./images/tree-item-selected.png)
+
+Expanded
+![](./images/tree-item-expanded.png)
+
+Collapsed
+![](./images/tree-item-collapsed.png)
+
+Like tree view, the tree view item has a default slot for its children.
+
```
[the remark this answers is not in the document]
chrisdholt: This could be helpful in solving the nested depth challenge above. Good idea

specs/tree-view/tree-view.md
chrisdholt: I think content is the slot, and likely the default slot. For users which want to add something like type ahead  behavior, they could do so using complex HTML. Alternatively, as a default slot for the item, this could also be a string which would help with the simple/straight forward tree like below:
```html
<tree>
    <tree-item>Foo</tree-item>
    <tree-item>Bar</tree-item>
    <tree-item>Baz</tree-item>
</tree>
```
Alternatively, default slot with complex:
```html
<tree>
    <tree-item><span>Foo</span></tree-item>
    <tree-item><span>Bar</tree-item>
    <tree-item><span>Baz</span></tree-item>
</tree>
```

specs/tree-view/tree-view.md
chrisdholt: I don't think it does; from everything I can find the keyboarding persists in both LTR and RTL locales.

specs/tree-view/tree-view.md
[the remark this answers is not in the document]
chrisdholt: I think we should standardize. My gut says we should go with kebab for readability. Some of these in lowercase would be difficult to parse and I can only forsee more complexity:
`beforecontent`
`aftercontent`

With that said, I updated these names based on feedback I saw on the toggle PR :)

specs/tree-view/tree-view.md
nicholasrice: *is it a slot?!*

jk sorry for the double-type. I'd agree with content for the default slot and named slot for items

specs/tree-view/tree-view.md
EisenbergEffect: @nicholasrice We should definitely stick to lowercase event names. Any framework whose template compiler is based in DOM is constrained by the case normalization of HTML. I should mention that the FastElement template engine also has this constraint currently. There are ways we could solve that, but it involves adding more code and affecting startup time, so I opted to keep things simple.

specs/tree-view/tree-view.md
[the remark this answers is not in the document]
EisenbergEffect: Left some details on the casing below. We should atleast go with lowercase for max compat. Kebab should be fine as well.

specs/tree-view/tree-view.md
janechu: I'm adding a section in the template that might help with where this lives see: https://github.com/microsoft/fast-dna/pull/2712

My concern is that other components may need drag and drop and we should have an issue, even if it's mostly blank and marked as an investigation, that multiple specs can link back to. I agree with including the reasoning of "why" drag and drop should be included for this component. 

Just another note is that when drag and drop is eventually done, we have links from the drag and drop issue to this spec (and others) thanks to the cross linking so we can go through and implement it without digging around for what our intentions were.

specs/tree-view/tree-view.md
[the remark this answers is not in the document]
janechu: @chrisdholt That makes sense, can we include that information somewhere? I think it should be part of the documentation when this is implemented.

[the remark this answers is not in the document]
chrisdholt: Yeah, I agree it will be helpful to include this information as it's currently just inferred.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,195 @@
...
+**Considering the challenge that drag & drop would present for any given component, I think it's worth holding on that feature, and taking it up seperately.*
+
+### Prior Art/Examples
+- [FAST-DNA (React)](https://explore.fast.design/components/tree-view)
+- [Material UI](https://material-ui.com/components/tree-view/)
+- [Lightning Design](https://www.lightningdesignsystem.com/components/trees/)
+- [Ant Design](https://ant.design/components/tree/)
+- [Atlassian](https://atlaskit.atlassian.com/packages/core/tree/example/static-tree)
+- [Windows (UWP)](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tree-view)
```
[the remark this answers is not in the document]
eljefe223: You inspire me Chris... ;) Think I'm going to add something similar to `Switch`

specs/README.md
```diff
@@ -4,4 +4,5 @@ Here you'll find specifications for custom elements and other library features.
 
 > **Note:** When authoring a new spec, please use [our standard spec template](./template.md). The PR that introduces your spec should also add a link to this readme in the specs section below.
 
-## Specs
\ No newline at end of file
+## Specs
+[Tree View](./tree-view/tree-view.md)
```
EisenbergEffect: Thank you!

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,186 @@
...
+
+Infinite nesting combined with the current visual design presents a styling challenge. Consider a scenario where we have the following tree:
+```html
+<fast-tree-view>
+    <fast-tree-item>Root node one</fast-tree-item>
+    <fast-tree-item>Root node two</fast-tree-item>
+    <fast-tree-item>
+        Parent Node && Root node
+        <fast-tree-item slot="child-nodes">Nested node one</fast-tree-item slot="child-nodes">
+        <fast-tree-item slot="child-nodes">Nested node two</fast-tree-item>
```
EisenbergEffect: `child-nodes` or `children` ?

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,186 @@
...
+## Design
+
+### API
+
+*The key elements of the component's public API surface:*
+
+**Tree View**
+
+- extends HTML Element props
+- custom CSS property for `depth` to support the infinite nesting scenario
```
EisenbergEffect: Is the idea here that `depth` would be incremented at each level and then used to calculate some indentation in styles?

specs/tree-view/tree-view.md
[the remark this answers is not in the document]
EisenbergEffect: The example above shows a `child-nodes` slot being used but it's not listed here.

specs/tree-view/tree-view.md
EisenbergEffect: Maybe even better as `items` since it's not really child nodes, but tree view items.

EisenbergEffect: Idea: In the constructor of the `fast-tree-item` we can write the `slot` value so that consumers do not have to assign it. This will allow them to leave that out of their html and ensure that tree items always project to the correct slot.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,186 @@
...
+| selected/collapsed | ![](./images/tree-item-selected-expanded.png)
+
+Like tree view, the tree view item has a default slot for its children.
+
+Slots:
+- `content` - the default slot.
+- `child-nodes` - named slot for child nodes.
+- `before-content` - named slot to optionally include content (often a glyph) before the `content` slot
+- `after-content` - named slot to optionally include content (often a glyph) after the `content` slot
+- `expand-collapse-glyph` - named slot which can be used to provide an element which will serve as the expand/collapse glyph
```
EisenbergEffect: Seems a little duplicative with the slot list above.

specs/tree-view/tree-view.md
EisenbergEffect: See my note above on how we can allow tree-view-item to auto-project to the correct slot without the consumer needing to assign the slot.

specs/tree-view/tree-view.md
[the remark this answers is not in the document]
chrisdholt: yeah, there's a few places where it *can* be listed and looks like it was just missed here. It's in a few places lower down the spec. I'll add here as well.

specs/tree-view/tree-view.md
chrisdholt: I'd be fine with either here in terms of naming. I do think it depends on what we want to land universally for *most* things here. I think both `fast-tree-view` and `fast-tree-item` should use the same naming here. Technically, I envision that a user could pass a <div> with the role of "treeitem" to either of these and it would still work. As long as it's in the slot and has the role of `treeitem` I'd expect everything to continue working.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,187 @@
...
+
+Infinite nesting combined with the current visual design presents a styling challenge. Consider a scenario where we have the following tree:
+```html
+<fast-tree-view>
+    <fast-tree-item>Root node one</fast-tree-item>
+    <fast-tree-item>Root node two</fast-tree-item>
+    <fast-tree-item>
+        Parent Node && Root node
+        <fast-tree-item slot="tree-item">Nested node one</fast-tree-item slot="tree-item">
+        <fast-tree-item slot="tree-item">Nested node two</fast-tree-item>
```
chrisdholt: @nicholasrice I've updated to note as `tree-item` for the time being.

specs/tree-view/tree-view.md
chrisdholt: Yes, that's the overall concept here. Thoguhts/concerns?

specs/tree-view/tree-view.md
chrisdholt: @EisenbergEffect the spec seems to provide multiple places for us to detail out slots. Slots, API, etc. Is one place preferred over the other?

specs/tree-view/tree-view.md
EisenbergEffect: The only trick would be the need to read the depth property before calculating the new depth, which would require a getComputedStyles call...which would force layout, etc. 

Another approach might be to use the connectedCallback of `tree-view-item` to have it look at it's parent element and check a property that specifies depth. Then it could calc and set the same property on itself. That property could then be used to apply an inline style...or just set a css property as well. We can experiment and see what ends up being better.

chrisdholt: Sure - I don't think we'll need the getComputedStyles here as I think there's a better way, but we will need to experiment.

specs/tree-view/tree-view.md
chrisdholt: re-updated to `item`

specs/tree-view/tree-view.md
nicholasrice: Wow. I can't believe css vars can't self-reference, I thought I'd seen this in the wild and thought for sure this would work:

`style="--depth: calc(var(--depth) + 1)`.

Here is a less-optimal solution: https://codesandbox.io/s/stoic-brook-73o1t?fontsize=14&hidenavigation=1&theme=dark

It uses two elements and two depth variables. It works but isn't ideal. Might be better then manually tracking a property in JS though.

specs/tree-view/tree-view.md
```diff
@@ -0,0 +1,187 @@
...
+*The key elements of the component's public API surface:*
+
+**Tree View**
+
+- extends HTML Element props
+- custom CSS property for `depth` to support the infinite nesting scenario
+
+**Tree View Item**
+
+- expanded - nodes with child nodes (tree items) can either be expanded or collapsed.
```
nicholasrice: Can you categorize these in a similar way that @eljefe223 did? I think it adds clarity to organize by properties, attributes, events, and slots

chrisdholt: Sure - can we add these preferences to the template itself?

chrisdholt: nvm, I see how this is inferred

Issue #2696, opened 2020-02-25
Stealth button with href has no hover affordance on explore.fast.design 


chrisdholt:
1. Go to https://explore.fast.design/components/stealth-button
2. In the form to the left, add a value for the href attribute
3. Hover over the stealth button
4. Notice there is no visual affordance for hover

**What behavior did you expect?**
The hover state should not change.

eljefe223:
This appears to be because of the `:enabled` selector on `hover` and and `active` since `<a/>` do not respond to `:enabled` we either need. to remove the `:enabled` selector from the base button styles or add individual `hover` and `active` selector to target the a tag.

PR #2698, opened 2020-02-25
fix: button anchors have no hover or active states

eljefe223:
# Description
closes: #2696 

This change makes sure that when a `<Button/>` is passed a `href` and renders an `<a/>` that the hover and active states remain the same. Since `:enabled` selectors don't apply to `<a/>` we need to specify and `hover` and `active` state for `<a/>`.

this change applies to all appearance variations of `<Button/>`, `<AccentButton/>`, `<LightweightButton/>`, `<NeutralButton/>`, `<OutlineButton/>`, `<StealthButton/>`, all appearance variations of `<ActionTrigger/>`, and all appearance variations of `<CallToAction/>` so please test accordingly.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-styles-msft/src/accent-button/index.ts
+++ b/packages/fast-components-styles-msft/src/accent-button/index.ts
@@ -32,14 +32,14 @@ const styles: ComponentStyles<AccentButtonClassNameContract, DesignSystem> = {
         color: accentForegroundCut,
         fill: accentForegroundCut,
         background: accentFillRest,
-        "&:hover:enabled": {
+        "&:hover:enabled, a&:not($button__disabled):hover": {
             background: accentFillHover,
             ...highContrastSelectedOutline,
             "& $button_beforeContent, & $button_afterContent": {
                 ...highContrastHighlightForeground,
             },
         },
-        "&:active:enabled": {
+        "&:active:enabled, a&:not($button__disabled):active": {
             background: accentFillActive,
         },
         ...applyFocusVisible<DesignSystem>({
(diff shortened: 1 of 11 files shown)


packages/fast-components-styles-msft/src/accent-button/index.ts
```diff
@@ -32,14 +32,14 @@ const styles: ComponentStyles<AccentButtonClassNameContract, DesignSystem> = {
         color: accentForegroundCut,
         fill: accentForegroundCut,
         background: accentFillRest,
-        "&:hover:enabled": {
+        "&:hover:enabled, a&:hover": {
```
bheston: Per our chat, it looks like these need inclusion of `button__disabled` in some way.

chrisdholt: can you elaborate here so we have online context? :)

bheston: The existing selector includes `:enabled` so we don't get hover and active styling on disabled buttons. Since <a> elements don't support the actual `disabled` attribute, we use the class instead. We need the class name in the selector so we don't end up with styles changing on hover and active.

eljefe223: Yeah something we caught yesterday. Spent sometime experimenting last night and this morning. Ended up using `:not()` css selector. This allows us to just replace and instance of `:enabled` selector across all buttons.

packages/fast-components-styles-msft/src/outline-button/index.ts
```diff
@@ -65,18 +64,13 @@ const styles: ComponentStyles<LightweightButtonClassNameContract, DesignSystem>
             ...highContrastDisabledBorder,
         },
         ...highContrastOutline,
-        "a&": {
+        "a&:not($button__disabled)": {
```
bheston: Something about this nesting looks odd. It duplicates the `not` portion and also includes a selector for when it _is_ disabled. Is this working correctly?

eljefe223: yes it is, Kham experimented removing the nested not selector and ran into the specificity issues again because we are specifically targeting `a&:not($button__disabled):hover` simply having `:hover` here gets overwrote.

PR #2699, opened 2020-02-25
fix: auto suggest menu opens with new suggestions and closes on blur

scomea:
# Description
This PR revives a previous one that went stale and adds another small fix:
- auto-suggest menu opens when component has focus and new suggestions are provided.  This is to better support scenarios where there is latency in getting suggestions such as waiting for a server response.
- auto suggest menu closes when focus is no longer within the component.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2655

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/auto-suggest/README.md
```diff
@@ -10,5 +10,7 @@ The expected usage pattern of the component is for authors to watch the "onValue
 
 The input region of the component can be customized using the component's "inputRegion" render prop.
 
+The "isMenuOpen" prop enables authors to take control of when the menu is open.  Setting "isMenuOpen" to false keeps the menu closed, and setting it to true keeps it open (unless there are no children, in which case it is always closed).
```
janechu: I think you can remove this since it is not a prop but internal state.

scomea: It is actually both.  The prop makes the menu state controllable.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -87,9 +87,19 @@ class AutoSuggest extends Foundation<
     }
 
     public componentDidUpdate(prevProps: AutoSuggestProps): void {
-        const updatedMenuVisibility: boolean = this.validateMenuState(
```
janechu: Does this mean that if you set the prop `isMenuOpen` since you are no longer calling `validateMenuState` you will override the prop?

scomea: No, the prop always wins in the validate function.

scomea: This check actually enables setting a prop to override existing state.

janechu: I see, what would help next time is making it clear that the `toggleMenu` method already runs a check using the `validateMenuState` before setting the state.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -274,6 +287,20 @@ class AutoSuggest extends Foundation<
         );
     };
 
+    /**
+     * Handle blur events
+     */
+    private handleBlurCapture = (event: React.FocusEvent): void => {
+        if (
+            this.state.isMenuOpen &&
+            this.rootElement.current !== null &&
```
paucortescosta: nit: just for consistency, I think it would be good to check the same way as above using isNil

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -87,9 +87,19 @@ class AutoSuggest extends Foundation<
     }
 
     public componentDidUpdate(prevProps: AutoSuggestProps): void {
-        const updatedMenuVisibility: boolean = this.validateMenuState(
-            this.state.isMenuOpen
-        );
+        let updatedMenuVisibility: boolean = this.state.isMenuOpen;
+
+        if (
```
nicholasrice: These two if statements seem needlessly complex:

The second statement will never pass unless the first one passes because `updatedMenuVisibility` is initialized to `this.state.isMenuOpen`. You should be able to do all the checking you need to in one if statement.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -213,11 +224,13 @@ class AutoSuggest extends Foundation<
...
-    private renderChildren(): React.ReactNode {
+    private renderChildren(alternateChildren?: React.ReactNode): React.ReactNode {
+        const children: React.ReactNode = isNil(alternateChildren)
+            ? this.props.children
+            : alternateChildren;
         if (this.props.filterSuggestions) {
-            const children: React.ReactNode = this.props.children;
             return React.Children.map(
-                children,
+                children as React.ReactNode,
```
nicholasrice: isn't this var typed as a `React.ReactNode` above? Is this type coercing necessary?

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: yep

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: yep, and the toggle function checks against previous state too.

packages/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
scomea: Build complains -TS2345: Argument of type '{}' is not assignable to parameter of type 'ReactElement…

nicholasrice: huh.... okay.

PR #2710, opened 2020-02-28
feat: add linked data control to update dictionary data

janechu:
# Description

Add linked data control to the `Form`. This makes adding and removing linked data a default behavior.

Relies on #2710 and #2705, #2704 to pass the build

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/form/controls/control.linked-data.tsx
+++ b/packages/fast-tooling-react/src/form/controls/control.linked-data.tsx
@@ -1,0 +1,275 @@
+import { ManagedClasses } from "@microsoft/fast-components-class-name-contracts-base";
+import React from "react";
+import { keyCodeEnter } from "@microsoft/fast-web-utilities";
+import { getDataFromSchema } from "../../data-utilities";
+import manageJss, { ManagedJSSProps } from "@microsoft/fast-jss-manager-react";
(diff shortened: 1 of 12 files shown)


packages/fast-tooling-react/src/form/controls/control.linked-data.tsx
```diff
@@ -0,0 +1,275 @@
...
+        );
+    }
+
+    private renderFilteredLinkedDataOptions(): React.ReactNode {
+        return Object.keys(this.props.schemaDictionary).map(
+            (schemaDictionaryKey: string) => {
+                return (
+                    <option
+                        key={schemaDictionaryKey}
+                        value={this.props.schemaDictionary[schemaDictionaryKey].title}
```
nicholasrice: FYI using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries would mean you don't have to retrieve data from `this.props`:

```jsx
return Object.entries(this.props.schemaDictionary).map((entry: [key, value]) => {
     return    <option
                        key={entry.key}
                        value={entry.value.title}
                     />
})
```

janechu: Nice, I like that.

PR #2711, opened 2020-02-28
refactor(fast-element): simplify by removing projector abstraction

EisenbergEffect:
# Description

A simple internal refactor that removes an unnecessary abstraction.

## Motivation & context

During the initial prototype, I thought we would need a full abstraction to represent custom elements that do or do not use shadow dom. After getting the initial work done, it turns out that there's only one or two if statements that are needed. So, rather than creating an interface and two separate classes that implement, just to run one or two lines of code different, I removed the entire abstraction and replaced it with a simpler if/else block.

The benchmarks that we have don't yet test the speed of component instantiation as a subset of overall rendering. While not the main point of this refactor, I suspect that it will have a net positive effect on component creation time in scenarios with many components, since it removes polymorphism.

## Issue type checklist

- [x] **Refactor**: An internal re-working that doesn't affect features.

## Process & policy checklist


EisenbergEffect:
@janechu @chrisdholt Need a second review in order to move forward with next steps. If you get a chance, this is a relatively small change.

PR #2713, opened 2020-03-02
chore: export extract element function

scomea:
# Description
Make extractHtmlElement function available for external use from the base components package.

## Motivation & context
Encountered a situation where it would be useful outside of the base components package.

## Issue type checklist

## Process & policy checklist


janechu:
Can you update the wording on this PRs description? It's unclear from the way it's written what you're trying to do.

packages/fast-components-react-base/src/index.ts
```diff
@@ -114,7 +114,11 @@ import Typography from "./typography";
 export { Typography };
 export * from "./typography";
 
-export { WindowWithIntersectionObserver, WindowWithResizeObserver } from "./utilities";
+export {
+    WindowWithIntersectionObserver,
+    WindowWithResizeObserver,
+    extractHtmlElement,
```
nicholasrice: We do have a `@microsoft/fast-react-utilities` package this may be better suited for.

scomea: Oh yeah, I didn't want to put it in generic web utilities and missed that we had a react specific one.  I'll move it there.

scomea: Moved, and also changed the ref prop to "any" as you noted in the other PR:
export function extractHtmlElement(
    sourceRef: React.RefObject<any> | HTMLElement
): HTMLElement | null {...

scomea:
Re-setting reviews as I needed to add tests to make coverage and that revealed that some better handling of text nodes in VP was appropriate.

PR #2718, opened 2020-03-03
feat: convert the Navigation component API to use the message system

### LINKED ISSUE #2484: Add the messaging system to the Navigation
The [messaging system](https://github.com/microsoft/fast-dna/issues/2482) should be added to the `Navigation` component as a property and should replace all data manipulation.

This should be merged into the `releases/tooling-breaking-changes-v2` branch.

Relies on #2482

janechu:
# Description

Similar to the `Form` component the `Navigation` component now takes the message system as a prop instead of data, schema and onChange callbacks.

## Motivation & context

This allows for syncing data and navigation updates via the message system.

Closes #2484

## Issue type checklist

## Process & policy checklist


packages/fast-tooling-react/app/pages/navigation.tsx
```diff
@@ -29,28 +48,29 @@ class NavigationTestPage extends React.Component<{}, NavigationTestPageState> {
...
-    };
-
     private handleUpdateDragAndDropReordering = (): void => {
         this.setState({
             dragAndDropReordering: !this.state.dragAndDropReordering,
         });
     };
+
+    private handleMessageSystem = (e: MessageEvent): void => {
+        if (e.data) {
```
nicholasrice: Don't think you need two if statements

packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
```diff
@@ -1,255 +1,216 @@
...
-            const clientOffset: XYCoord | null = monitor.getClientOffset(); // mouse position
-            const hoverClientY: number = clientOffset.y - hoverBoundingRect.top; // pixels from the top
-            const verticalOffset: number = hoverBoundingRect.height / 4;
-            const direction: VerticalDragDirection =
-                hoverClientY < hoverMiddleY - verticalOffset
-                    ? VerticalDragDirection.up
-                    : hoverClientY > hoverMiddleY + verticalOffset
-                        ? VerticalDragDirection.down
-                        : VerticalDragDirection.center;
+function treeItemEndLeaf(
```
nicholasrice: Any reason you're not making these functional components? I think you could simplify all of these function arguments into a prop interface and use this like a component, which will be more extensible over time and is more idiomatic:

```
<TreeItemEndLeaf
  ...props
/>
```

janechu: Because, oddly, once you do that the drag and drop functionality ceases to work. I haven't nailed down why that would be but it seems that drag and drop is expecting a single component to work with.

PR #2720, opened 2020-03-03
chore: rename plugin and add placeholder art required by Figma publishing

nicholasrice:
# Description
Removes working name in place of landed name. Also adds placeholder art required during the figma plugin publishing process.

## Issue type checklist

## Process & policy checklist


packages/fast-figma-plugin-msft/manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "fast-figma-plugin-msft",
+  "name": "(Alpha) Theme Designer",
```
awentzel: What is our naming policy for an alpha, beta, release, vnext etc?  What would we name beta?  Does the package folder name always match the manifest name?  If so, should the package be renamed?  I might also, rather see "Theme Designers - Alpha" in the case there are multiple builds they will group together.  I prefer to read what it is, then read package attribution info like alpha, beta, etc.

bheston: I like the idea of the release status at the end as well. Not sure what the significance is, if any, aside from if it shows in the UI. I suspect this is the menu title.

nicholasrice:
Going to roll this into some other work.

PR #2723, opened 2020-03-03
chore: adds form-associated-custom-element spec

nicholasrice:
# Description
This specification is designed to outline a base class for implementing custom form elements in web components.

(living) prototype implementation of this spec exists here: https://github.com/microsoft/fast-dna/blob/users/nirice/wc-checkbox-prototype/packages/fast-components/src/checkbox/checkbox.ts

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/form-associated-custom-element.md
+++ b/specs/form-associated-custom-element.md
@@ -1,53 +1,56 @@
 # Form Associated Custom Elements
 
 ## Overview
-This spec regards a general implementation to facilitate form communication and connection for components that should be form-associated, including checkboxes, radios, text inputs, and other components which store user input and should be captured during form submission.
+This spec regards a general implementation for components that should be form-associated, including checkboxes, radios, text inputs, and other components which store user input and should be captured during form submission.
+
+### Defintions and acronyms:
+- FACE: form-associated custom elements
 
 ### Background
-Components that replace or behave like native form elements (input, textarea, select) should generally behave like native like their native counterpart. One key aspect to this is associating the element with the parent form. To do this, we will expose a mechanism to expose form-association to custom elements that can be shared across implementations with maximum re-use.
+Components that are intended to replace a native form element (input, textarea, select) should generally behave like their native counterpart. One key aspect to this is associating the element with the parent form. To do this, we will expose a mechanism to expose form-association to custom elements that can be shared across implementations with maximum re-use.
 
 ### Use Cases
 Any custom element that should associate a value to a form.
 
 ## Design
-The implementation will be an *abstract class* that will extend `FastElement`. The class will expose implementations for all of the common form element capabilities. It will also expose and manage an implementation for when the form-associated custom element APIs are not available.
+The implementation will be an *abstract class* that will extend `FastElement`. The class will expose implementations for all of the common form element capabilities. It will also expose and manage an implementation for when the FACE APIs are not available.
 
-Another option I looked at is using a decorator, but there are a few challenges with that approach:
+Another possible implementation is a decorator, but there are a few challenges with that approach:
   1. decorators cannot augment the class type directly, so we would need to use a pattern similar to https://www.typescriptlang.org/docs/handbook/mixins.html to gain accurate type definitions
   2. Decorators cannot (easily?) provide a *default* implementation that can be overridden by the extending class. Using an abstract class allows straight overrides of implementation or merging of implementation with super
 
-The implementation will standardize an interface between two common cases: browsers with form-associated-custom-element support and browsers without.
+The implementation will standardize an interface between two common cases: browsers with FACE API support and browsers without.
 
 ### Browsers with FACE support
 For browsers *with* support, methods and properties will generally proxy to the `ElementInternals` implementation.
 
 ### Browsers with FACE support
-The implementation will manage a "proxy" element that will be appended to the light-dom. This proxy element will server as the custom-element's association to the form. All relevant properties will be forwarded to this element, and certain API calls will retrieve data from it.
+The implementation will manage a "proxy" element that will be appended to the light-dom. This proxy element will serve as the custom-element's association to the form. All relevant properties will be forwarded to this element, and certain API calls will retrieve data from it.
 
 ### API
 #### IDL attributes
 - `public static formAssociated: boolean`
-  - Required for form-associated custom element API. This will feature-detect the API and resolve a value corresponding to feature-support
+  - Requirement of FACE API. This will feature-detect the API and resolve a value corresponding to feature-support.
 - `public readonly validity: ValidityState`
-  - Returns the [validity](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) of the component
+  - Returns the [validity](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) of the component.
 - `public readonly form: HTMLFormElement | null`
-  - The associated form element
+  - The associated form element.
 - `public readonly validationMessage: string`
-  - The current validation message of the element
+  - The current validation message of the element.
 - `public readonly labels: Node[]`
-  - Labels associated to the element. See [risks and challenges](#retrieving-labels)
+  - Labels associated to the element. See [risks and challenges](#retrieving-labels).
 - `protected abstract proxy: HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement`
-  - A proxy element that will be appended to the DOM when form-association APIs are unavailable. It will server as the custom-elements connection to the form.
+  - A proxy element that will be appended to the DOM when FACE APIs are unavailable. It will server as the custom-elements connection to the form.
 - `protected elementInternals: ElementInternals`
   - Provides form-association through dedicated APIs.
 - `protected setFormValue(value: File | string | FormData | null, state?: File | string | FormData): void`
-  - When using `elementInternals`, this will set the value in the form. With no form-association support, this will noop because the value will automatically be associated by the proxy element. This can be overridden as necessary by components with more advanced behavior.
+  - When using `elementInternals`, this will set the value in the form. With no FACE support, this will do nothing because the value will automatically be associated by the proxy element. This can be overridden as necessary by components with more advanced behavior.
 - `protected onLabelClick(): void`: **Discussion**
   - see [risks and challenges](#clicking-labels)
 - `protected handleKeyPress(): void`
   - Will submit the form when `enter` key is pressed to match standard input behavior. Can be overridden or extended.
 - `protected setValidity(flags: ValidityStateFlags, message?: string, anchor?: HTMLElement): void;`
-  - With form association support, will set the validity of the component. With no form association, this will noop unless a message is passed. If a message is passed, the proxy element's `setCustomValidity` method will be invoked with the value of the message.
+  - With form association support, will set the validity of the component. With no form association, this will do nothing unless a message is passed. If a message is passed, the proxy element's `setCustomValidity` method will be invoked with the value of the message.
 
 #### Content attributes
 - `id: string`
@@ -66,7 +69,7 @@ The implementation will manage a "proxy" element that will be appended to the li
 ### Risks and Challenges
 
 #### Retrieving labels
-Accessing the [`labels`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels) property of a native input returns a `NodeList`. In cases where the form-association APIs are available, the implementation from `ElementInternals.labels` works as expected. In cases where we implement a proxy element though, we need to construct the label set from the following:
+Accessing the [`labels`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels) property of a native input returns a `NodeList`. In cases where the FACE APIs are available, the implementation from `ElementInternals.labels` works as expected. In cases where we implement a proxy element though, we need to construct the label set from the following:
   1. any parent `<label>`element. Can be retrieved from the proxy element's `labels` property
   2. any `<label>` element with `[for="$id"]`, where $id is the `id` attribute *of the custom element*. Because the custom element does not reflect it's `id` attribute to the proxy element (that would result in two elements with the same ID in the DOM), the labels are not automatically associated to the `labels` property of the proxy.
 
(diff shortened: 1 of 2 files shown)


specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,115 @@
+# Form Associated Custom Elements
+
+## Overview
+This spec regards a general implementation to facilitate form communication and connection for components that should be form-associated, including checkboxes,
+radios, text inputs, and other components which store user input and should be captured during form submission.
```
janechu: Looks like there is a return here between checkboxes and radios.

nicholasrice: fixed.

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
+# Form Associated Custom Elements
+
+## Overview
+This spec regards a general implementation to facilitate form communication and connection for components that should be form-associated, including checkboxes, radios, text inputs, and other components which store user input and should be captured during form submission.
```
janechu: Wording seems a little tough to understand, maybe just "This spec defines a general implementation of form-associated components with special attention paid to form specific behaviors."

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Background
+Components that replace or behave like native form elements (input, textarea, select) should generally behave like native like their native counterpart. One key aspect to this is associating the element with the parent form. To do this, we will expose a mechanism to expose form-association to custom elements that can be shared across implementations with maximum re-use.
```
janechu: I think you mean "...should generally behave like their native counterpart."

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Use Cases
+Any custom element that should associate a value to a form.
+
+## Design
+The implementation will be an *abstract class* that will extend `FastElement`. The class will expose implementations for all of the common form element capabilities. It will also expose and manage an implementation for when the form-associated custom element APIs are not available.
+
+Another option I looked at is using a decorator, but there are a few challenges with that approach:
```
janechu: Can we make disassociate this wording with any references to the author? Perhaps the wording might work as follows:

```markdown
### Other implementations

Another approach that was considered is the use of a decorator. This was determined to be a less ideal option due to:
...
```

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+  1. decorators cannot augment the class type directly, so we would need to use a pattern similar to https://www.typescriptlang.org/docs/handbook/mixins.html to gain accurate type definitions
+  2. Decorators cannot (easily?) provide a *default* implementation that can be overridden by the extending class. Using an abstract class allows straight overrides of implementation or merging of implementation with super
+
+The implementation will standardize an interface between two common cases: browsers with form-associated-custom-element support and browsers without.
+
+### Browsers with FACE support
+For browsers *with* support, methods and properties will generally proxy to the `ElementInternals` implementation.
+
+### Browsers with FACE support
```
janechu: This title is duplicated, also, can we get a link to what is FACE support?

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+The implementation will manage a "proxy" element that will be appended to the light-dom. This proxy element will server as the custom-element's association to the form. All relevant properties will be forwarded to this element, and certain API calls will retrieve data from it.
+
+### API
+#### IDL attributes
```
janechu: Can we get a link and a reference link to IDL attributes? I'm thinking just https://developer.mozilla.org/en-US/docs/Glossary/IDL

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+- `public readonly validity: ValidityState`
+  - Returns the [validity](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) of the component
+- `public readonly form: HTMLFormElement | null`
+  - The associated form element
+- `public readonly validationMessage: string`
+  - The current validation message of the element
+- `public readonly labels: Node[]`
+  - Labels associated to the element. See [risks and challenges](#retrieving-labels)
+- `protected abstract proxy: HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement`
+  - A proxy element that will be appended to the DOM when form-association APIs are unavailable. It will server as the custom-elements connection to the form.
```
janechu: server = serve?

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+The implementation will manage a "proxy" element that will be appended to the light-dom. This proxy element will server as the custom-element's association to the form. All relevant properties will be forwarded to this element, and certain API calls will retrieve data from it.
+
+### API
+#### IDL attributes
+- `public static formAssociated: boolean`
+  - Required for form-associated custom element API. This will feature-detect the API and resolve a value corresponding to feature-support
+- `public readonly validity: ValidityState`
+  - Returns the [validity](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState) of the component
+- `public readonly form: HTMLFormElement | null`
+  - The associated form element
```
janechu: Should these all end in "."? The list seems inconsistent in its formatting.

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+- `public readonly validationMessage: string`
+  - The current validation message of the element
+- `public readonly labels: Node[]`
+  - Labels associated to the element. See [risks and challenges](#retrieving-labels)
+- `protected abstract proxy: HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement`
+  - A proxy element that will be appended to the DOM when form-association APIs are unavailable. It will server as the custom-elements connection to the form.
+- `protected elementInternals: ElementInternals`
+  - Provides form-association through dedicated APIs.
+- `protected setFormValue(value: File | string | FormData | null, state?: File | string | FormData): void`
+  - When using `elementInternals`, this will set the value in the form. With no form-association support, this will noop because the value will automatically be associated by the proxy element. This can be overridden as necessary by components with more advanced behavior.
```
janechu: Can we get a link to noop?

nicholasrice: it just means "no operation". I will try to clarify.

specs/form-associated-custom-element.md
nicholasrice: I can define - "form associated custom elements"

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+- `protected onLabelClick(): void`: **Discussion**
+  - see [risks and challenges](#clicking-labels)
+- `protected handleKeyPress(): void`
+  - Will submit the form when `enter` key is pressed to match standard input behavior. Can be overridden or extended.
+- `protected setValidity(flags: ValidityStateFlags, message?: string, anchor?: HTMLElement): void;`
+  - With form association support, will set the validity of the component. With no form association, this will noop unless a message is passed. If a message is passed, the proxy element's `setCustomValidity` method will be invoked with the value of the message.
```
janechu: I see "form association support" and "form-association support", which one should we use in the document?

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+  - Boolean attribute. Disables the form control, ensuring it doesn't participate in form submission
+- `name: string`
+  - The name of the element. Allows access by name from the associated form.
+- `required: boolean`
+  - Boolean attribute.
+
+### Risks and Challenges
+
+#### Retrieving labels
+Accessing the [`labels`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels) property of a native input returns a `NodeList`. In cases where the form-association APIs are available, the implementation from `ElementInternals.labels` works as expected. In cases where we implement a proxy element though, we need to construct the label set from the following:
```
janechu: I would change this wording to "...In cases where a proxy element is implemented, a label must be constructed from the following:"

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+  1. any parent `<label>`element. Can be retrieved from the proxy element's `labels` property
+  2. any `<label>` element with `[for="$id"]`, where $id is the `id` attribute *of the custom element*. Because the custom element does not reflect it's `id` attribute to the proxy element (that would result in two elements with the same ID in the DOM), the labels are not automatically associated to the `labels` property of the proxy.
+
+`NodeList`s are not constructable in JavaScript, and because we need to construct the label set in cases where we're using a proxy element, I'm purposing standardizing this property to `Node[]` , which AFAIK is as close to `NodeList` as we can get.
```
janechu: I might change this to:
"\`NodeList\`s are not constructable in JavaScript, thus this property is standardized to \`Node[]\` due to the need to construct the label when implementing with a proxy element. The use of `Node[]` is intended to map as close to \`NodeList\` as possible."

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Clicking labels
+Clicking a label of a native input element can have several side effects. First (and in general) it will focus the element. In cases like radio and checkbox, it changes the value of the input. There are likely other side-effects.
+
+How this works under the hood seems to be that clicking the label *also* invokes a click event on the element the label *labels*.
+
+However, because (in proxy element cases) the label's aren't *actually* associated to the element, the click event on the custom element isn't ever fired. This is the case for all label's using the `for` attribute to make the association. *Wrapping* labels (`<label><custom-element></label>`) *do* seem to fire the click event on the custom element.
+
+In order to *truly* support this scenario, we need to attach click event handlers to all labels for the elements. It *also* means that we need to attach new listeners when *new* labels get added to the DOM. This is the tricky part. We could use mutation observer, but my fear is that is heavy-handed for an edge-case scenario.
```
janechu: I would change first line to "In order to \*truly\* support this scenario, click event handlers must be attached to all labels for the elements."

janechu: I would change second line to "It *also* means that new listeners must be attached when *new* labels get added to the DOM. This is the tricky part."

janechu: I would change the third line to "A mutation observer could be used, but this may be heavy-handed for an edge-case scenario."

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Test Plan
+
+TBD. Form association APIs are very new so JSDOM is not likely to expose the feature. It might make sense to do in-browser tests against browsers both with and without support to ensure parity.
+
+
+## Resources
+- https://docs.google.com/document/d/1JO8puctCSpW-ZYGU8lF-h4FWRIDQNDVexzHoOQ2iQmY/edit?pli=1
+- https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements-face-example
+- https://web.dev/more-capable-form-controls/
```
janechu: Can these be friendly named links?

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+How this works under the hood seems to be that clicking the label *also* invokes a click event on the element the label *labels*.
+
+However, because (in proxy element cases) the label's aren't *actually* associated to the element, the click event on the custom element isn't ever fired. This is the case for all label's using the `for` attribute to make the association. *Wrapping* labels (`<label><custom-element></label>`) *do* seem to fire the click event on the custom element.
+
+In order to *truly* support this scenario, we need to attach click event handlers to all labels for the elements. It *also* means that we need to attach new listeners when *new* labels get added to the DOM. This is the tricky part. We could use mutation observer, but my fear is that is heavy-handed for an edge-case scenario. 
+
+Another options would be to document the gap and not support this scenario, instead only attaching listeners to elements that are in the DOM when the component connects. Any other thoughts here?
```
janechu: I think we should document the gap and not support this, I can't think of a real use case.

specs/form-associated-custom-element.md
EisenbergEffect: Is this something that we can lazily evaluate?

specs/form-associated-custom-element.md
EisenbergEffect: I see that the answer to my previous question...looks like we would need to eagerly evaluate the labels in order to hook up the events in this case.

What if we instead attached an event on the form that the element was within, caught it there, and then used that as a central place to handle labels? Then we don't need to worry about observing for new labels.

specs/form-associated-custom-element.md
EisenbergEffect: See idea above about using event delegation.

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+No globalization concerns.
+
+### Security
+
+No new security concerns. Form data sanitation will be the responsibility of the consuming application.
+
+### Performance
+
+Potentially using a MutationObserver to attach new click handlers to label elements. TBD on if we want to do this.
```
EisenbergEffect: I think we should avoid this for the first version at least. It's probably and edge case and we could easily add it later if we get a customer request. Hard to take it out once it's in. Also, see note above about event delegation. Maybe we can handle labels in another way entirely.

specs/form-associated-custom-element.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Dependencies
+
+No Dependencies
+
+### Test Plan
+
+TBD. Form association APIs are very new so JSDOM is not likely to expose the feature. It might make sense to do in-browser tests against browsers both with and without support to ensure parity.
```
EisenbergEffect: Agree. I'd like to get real cross-browser testing in place.

specs/form-associated-custom-element.md
EisenbergEffect: I'm definitely interested in how the implementations might look for the above. If there's a prototype, I'd love to have you walk me through it.

specs/form-associated-custom-element.md
nicholasrice: what should the friendly name be?

specs/form-associated-custom-element.md
nicholasrice: Yea we only have to constuct it when when it's accessed. It is a readonly property so I have it implemented in the prototype as a `get` accessor.

nicholasrice: Ahh but yes, if we want to attach listeners to labels we would always evaluate after connected.

specs/form-associated-custom-element.md
nicholasrice: That would probably catch *most* cases. It does remind me of a case I missed though (I'll add it) - the `form` content attribute can be assigned to the id of a form to be associated with, meaning the input doesn't *need* to be in a form. Probably a very edge-case but it is possible.

I guess we *could* say this component doesn't support that, in which case I think your suggestion would work.

specs/form-associated-custom-element.md
janechu: Maybe just their titles, so first "Form Participation API Explained", second "Creating a form-associated custom element", third "More capable form controls".

specs/form-associated-custom-element.md
nicholasrice: This is just what they're called in the web platform - I'd rather not link to the definition when it is easily found through a search if it is unfamiliar.

specs/form-associated-custom-element.md
janechu: If you feel like it, I was breezing through here and didn't read it thoroughly, I see what you're saying now.

specs/form-associated-custom-element.md
nicholasrice: I've posted a link to the prototype branch in the PR description. Not all of this is represented yet but I'll continue to flesh that prototype out.

nicholasrice: Cleaned this up

nicholasrice:
I've added a "next steps" section to address label clicking for non FACE elements. I think we can take this as a follow task when appropriate.

PR #2724, opened 2020-03-03
chore: add dialog spec

chrisdholt:
# Description
This PR adds a specification for a dialog web component.

## Issue type checklist

## Process & policy checklist


specs/dialog/dialog.md
```diff
@@ -0,0 +1,113 @@
...
+
+---
+
+## Design
+
+### API
+
+*The key elements of the component's public API surface:*
+
+**Tree View**
```
nicholasrice: copy + pasta

chrisdholt: gggrrrrrr....

specs/dialog/dialog.md
```diff
@@ -0,0 +1,113 @@
...
+### Anatomy and Appearance
+**Structure:**
+```
+<div class="root">
+    <div class="positioning-region">
+        <div class="overlay">
+        <div
+            role="dialog"
+            aria-modal="true"
+            tabIndex={-1}
```
janechu: should this be tab-index="-1"?

chrisdholt: yes, updated - good catch!

specs/dialog/dialog.md
```diff
@@ -0,0 +1,113 @@
...
+## Design
+
+### API
+
+*Component name:*
+- `fast-dialog`
+
+*Attributes:*
+- `modal` - boolean value for modal functionality. Sets aria-modal to true.
+- `visible` - boolean value to toggle the visibility of the dialog
```
scomea: Is this necessary?  Why can't authors simply not include their dialog in the dom when they don't need it?

nicholasrice: [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) is a built-in attribute that I believe has the same function here. It might be worth considering adopting a more standard attribute for this purpose.

specs/dialog/dialog.md
```diff
@@ -0,0 +1,113 @@
...
+
+*Custom properties:*
+- width
+- height
+
+*Slots:*
+- default slot for content
+
+*Events*
+- dismiss - event fired when a user presses escape or clicks outside the dialog content
```
nicholasrice: do want an event for opening also?

specs/dialog/dialog.md
chrisdholt: That requires a prescriptive implementation; what if I want to just stick it in the DOM always and then toggle the visible with a simple js snippet in my html? This API provides a more flexible API IMO.

chrisdholt: I think we could adopt this - would the hidden selector work here if I wanted to style via CSS? In the event that a user changes the display property where the hidden attribute is, that will override the hidden behavior. In this case, I would want to be sure that the following would work:

HTML:
```html
<fast-dialog hidden>Foo bar</fast-dialog>
```
CSS:
```css
fast-dialog[hidden] {
    display: none;
}
```

specs/dialog/dialog.md
chrisdholt: Would we consider "opening" the same as visible in this case? I'm open to adding an additional event here, but am curious about the use case considering app authors would be managing the visibility and instantiation of the dialog fully.

specs/dialog/dialog.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Features
+
+- **Soft dismiss (light dismiss):** In certain implementations, app authors want a dialog to close when a user clicks outside the content region or presses the `esc` key with their keyboard.
+
+- **Focus management:** In all circumstances, focus should move to an element contained in the dialog. Per the W3C, "Unless a condition where doing otherwise is advisable, focus is initially set on the *first* focusable element" (emphasis mine). If an element within the dialog is set to be focused by an app author, our control should not override that behavior. In both modal and non-modal dialogs, tab focus should remain within the dialog content region and not move outside it. When a dialog the dialog is closed, users may also want a way to return focus to the element which invoked the dialog.
```
EisenbergEffect: A little typo "When a dialog the dialog"

specs/dialog/dialog.md
EisenbergEffect: In my understanding, that should work. We'll provide a default host style in the dialog css as follows:

```css
:host([hidden]) { 
  display: none;
}
```

The consumer of the dialog can always provide their own css to override the host styles.

EisenbergEffect: I wonder if it would be nice to have a convenience API for showing dialogs. For example:

```TypeScript
const result = await FastDialog.show(MyDialogContentComponent, settings);
```

The `show` method could create an instance of `FastDialog` and create an instance of the provided component. It would add the component to the content of the dialog. Then it would append the dialog to the body of the document. After the dialog was dismissed, it would resolve the promise and return the instance of the provided content component to the caller so that they could get any data they needed out of that.

Not necessary for a v1, and requires a good bit more design, but something that could be a nice down the road. I've built some things like this and people seem to have found it useful.

EisenbergEffect: It might be worth mentioning what the default values of the above attributes should be. For example, I'm assuming the default is `model: true` and `visible: true`.

specs/dialog/dialog.md
```diff
@@ -0,0 +1,113 @@
...
+| default | ![](./images/dialog.png) |
+| dark mode | ![](./images/dialog-dark.png) |
+
+Parts:
+- root - the root of the dialog
+- positioning-region - ensuring the dialog is centered correctly requires a certain structure. Centering with absolute positioning can cause blurry content within the dialg itself. After investigating several implementations, the best and most common way of ensuring the dialog can be centered is to include a div to position the actual dialog itself using something like flexbox or css-grid. This ensures that the blurry content issue does not happen and the dialog centers easily within the screen.
+- modal-overlay - the modal overlay
+- content-region - the region where content is actually rendered. This part is where the `role="dialog"` will actually exist
+
+---
```
EisenbergEffect: Is it worth talking a bit about how animation might work for those who want to animate their dialogs in and out of view?

specs/dialog/dialog.md
chrisdholt: oops, fixed

specs/dialog/dialog.md
chrisdholt: I think we can look at that as a feature request @EisenbergEffect. I've added the defaults here, but as an FYI, I've changed `visible` to `hidden` based on @nicholasrice's comment below.

specs/dialog/dialog.md
chrisdholt: Do we want configurable animations if our initial work is to provide the MSFT implementation? I think the initial thinking from design was to bake the animation in here.

chrisdholt: My working assumption is that animation is something that will be considered part of the "adaptive ui" design system at some point which will provide configurability. If someone wants something completely different than can always create their own instance.

EisenbergEffect: Just worth mentioning in the spec I think. Just take your comments and put them in the doc for reference.

I imagine that configurability wouldn't be too hard though. For example, if we put a part attribute on the overlay and a part attribute on the dialog and then we internally apply classes for enter/leave, that would allow the consumer to target those parts and provide their own enter/leave animations. There's a bit more to it, but something to think about.

If we build the animation in, we can do that as a separate set of `ElementStyles` as well, so that consumers could easily re-compose a dialog using all our styles but replacing the animation piece without having to re-write anything.

Just thinking...

chrisdholt: Sure

chrisdholt: updated with some additional details

specs/dialog/dialog.md
```diff
@@ -0,0 +1,125 @@
...
+| dark mode | ![](./images/dialog-dark.png) |
+
+Parts:
+- root - the root of the dialog
+- positioning-region - ensuring the dialog is centered correctly requires a certain structure. Centering with absolute positioning can cause blurry content within the dialg itself. After investigating several implementations, the best and most common way of ensuring the dialog can be centered is to include a div to position the actual dialog itself using something like flexbox or css-grid. This ensures that the blurry content issue does not happen and the dialog centers easily within the screen.
+- modal-overlay - the modal overlay
+- content-region - the region where content is actually rendered. This part is where the `role="dialog"` will actually exist
+
+Animation:
+The current working model (assumption) is that animation iwill be taken up as part of the adaptive ui story in the design system at some point. This will provide configurability to users. From a dialog standpoint, I think we'll have a default animation baked in, though this would be configurable through the design system.
```
EisenbergEffect: Typo "iwill"

specs/dialog/dialog.md
```diff
@@ -0,0 +1,125 @@
...
+    styles: css`${dialogBaseStyles}${myAnimationStyles}`
+})
+```
+---
+
+## Implementation
+
+### States
+
+Visible - The dialog is visible. This state is managed solely by the app author via the visible attribute.
```
EisenbergEffect: Need to update this to align with the note on `hidden` above.

specs/dialog/dialog.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Accessibility
+
+The tree view should align to the interaction model provided by the W3C: https://w3c.github.io/aria-practices/#keyboard-interaction-7
```
EisenbergEffect: copy paste

specs/dialog/dialog.md
chrisdholt: DOH

specs/dialog/dialog.md
chrisdholt: Ruh-roh...fixed! Thanks!

PR #2726, opened 2020-03-04
chore: add card spec

chrisdholt:
# Description
This PR proposes a new spec for a card component.

## Issue type checklist

## Process & policy checklist


janechu:
Should we include some screenshots?

chrisdholt:
> Should we include some screenshots?

I thought about this; I can add them, though I think I'll give the caveat that the examples are "composed" with the card. Does that sound acceptable?

specs/card/card.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Design
+
+### API
+
+*Component name:*
+- `fast-card`
+
+*Attributes:*
+- extends HTMLDivElement
```
EisenbergEffect: Just reminding you that we can't do this in a cross-browser way.

specs/card/card.md
```diff
@@ -0,0 +1,71 @@
...
+
+*Custom properties:*
```
EisenbergEffect: Just to make sure I'm understanding correctly, this is referring to CSS properties, yes?

specs/card/card.md
chrisdholt: Yes, though I believe that problem exists with CSS parts as well, correct? It seems like extending from the div element here makes the most sense long term. We can avoid doing that if it's too problematic.

specs/card/card.md
chrisdholt: Correct

specs/card/card.md
EisenbergEffect: The problem is that the card wouldn't be usable at all in Safari when extending from div. However, adding parts only limits customization. The component will still function.

chrisdholt: Updated to remove this note.

specs/card/card.md
chrisdholt: Updated this to better represent CSS custom properties.

PR #2729, opened 2020-03-04
chore: migrate message system to the fast-tooling package with data utilities dependencies

janechu:
# Description

Moves non-react specific message system and message system data utility dependencies into the `fast-tooling` package.

## Motivation & context

This migration is part of a larger effort to make sure that the react specific tooling package only contains react specific exports. This will be crucial in the future when these components are converted to web components.

Contributes to #2482

## Issue type checklist

## Process & policy checklist


packages/fast-tooling/src/data-utilities/relocate.ts
```diff
@@ -3,40 +3,34 @@
...
 import { DataType } from "./types";
 
-export enum TargetPosition {
-    prepend,
-    append,
-    insert,
-}
-
-export interface UpdateDataWithSourceConfig {
+interface UpdateDataWithSourceConfig {
```
scomea: You don't need to export these?

janechu: The type should be inferred instead of forcibly cast, you can see its use in the `message-system.utilities.ts` file on line 174. It's a good point though, I'll have to keep an eye out to see if there are requests for it, I'm trying to limit the amount of exports that are available for brevity.

scomea: I thought it odd that there are functions exported that use params of types that are not also exported.

PR #2730, opened 2020-03-04
feat(fast-element): simpler element definition without decorators

EisenbergEffect:
# Description

Some minor changes to fast-element definition APIs to improve consistency and make it easier to define elements in vanilla JS.

## Motivation & context

Not everyone who uses the fast-element library will be able to use decorators. We want to make sure that we're enabling folks who use vanilla JS or aren't ready to leverage TypeScript decorators. To enable this the `CustomElement.define` API now can take a type only, without a definition. If no definition is provided, it will look for a static `definition` property on the class and use that. This enables defining an element with a vanilla JS class and no decorators pretty easy. Here's an example:

```JavaScript
export const NameTag = CustomElement.define(class extends FastElement {
  static definition = {
    name: 'name-tag',
    template: html`...`,
    styles: css`...`,
    bindables: [
      'greeting'
    ]
  };

  greeting;
});
```

The definition can also be passed independently:

```JavaScript
const definition = {
  name: 'name-tag',
  template: html`...`,
  styles: css`...`,
  bindables: [
    'greeting'
  ]
};

export class NameTag extends FastElement {
  greeting;
}

CustomElement.define(NameTag, definition);
```

## Open questions

* Instead of `CustomElement.define`, should it be `FastElement.define`? Benefits include:
  * Less confusion with `window.customeElements.define`
  * One less import in non-decorator scenarios since `FastElement` can serve as the base class and the host for the static `define` method.
* I am not particularly happy with the name `bindables`. This property defines properties on the object that are surfaced as attributes on the HTML element. Since there are potentially other properties that are not attributes, this doesn't make sense to call `properties`. It's also awkward to call it `attributes` since when fully configured without defaults, it is a map from property name to config object. Also, there are other attributes on an element which are not represented in this array (those from the platform). Any recommendations on a better name are welcome. Until then, we can stick with this. I borrowed it from Aurelia, where it's worked for quite a few years without any complaint. It just seems a bit out of place in this context.

## Issue type checklist

This does change the imperative APIs for element definition functionality, but that is unlikley to be used by anyone at this time. These APIs are mostly called by the `@customElement` decorator and other internal procedures.

## Process & policy checklist

## Next Steps

* See discussion below for a follow-up PR to address the open questions above.
* Add an `$emit` helper for simplified custom events, guarded by controller connected state.
* Revisit the DI system and see if that can be made more robust.
* See if removing polymorphism in bindings improves perf and reduces library size.
* Tests - Waiting on the API and contracts to settle down a bit more, in order to reduce redoing a lot of work. Should be done this week. Then lots and lots of tests. In the mean time I've been doing simple manual testing using our test elements and benchmarks.


nicholasrice:
@EisenbergEffect the platform seems to call those "content attributes" - we could potentially take a que form them: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Content_versus_IDL_attributes

EisenbergEffect:
@nicholasrice The current structure is a map from property name to config object, like this:

```JavaScript
{
  'myProperty': { property: 'myProperty', attribute: 'my-property' }
}
```

So, it seems strange to call it "attributes" since it's a map from property name to a config object that defines the property name and the attribute name (along with any other config data we need in the future).

Maybe we could call the config type an `AttributeDescription` and call the property on the custom element description just `idl`?

```TypeScript
interface AttributeDescription {
  property: string;
  attribute: string;
}

interface CustomElementDescription {
  idl: Record<string, AttributeDescription>;
}
```

EisenbergEffect:
We could also call it `members`. That might be more friendly.

nicholasrice:
It seems to me that the term "Attributes" is intended to include *both* content attributes and IDL attributes. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input from an enumerated list of "Attributes" that includes both content attributes and IDL attributes. My preference would be to map to existing language than introduce something new

nicholasrice:
Hmm well actually that list is I believe only enumerating the *content attributes* (based on casing), so maybe disregard that list? I've personally found "content" vs "IDL" to be helpful in differentiating. The terms "property" and "attribute" get thrown around interchangeably I've found.

EisenbergEffect:
@nicholasrice My concern is that it might be confusing to have an object called `attributes` where the keys were property names on the class, not the HTML attribute names (since they can be different). Also, the final normalized element definition already has a property called `attributes` that maps the attribute name back to the same configuration object. We need this in order to look up in both directions quickly.

EisenbergEffect:
So, to be clear, I'm talking about the shape of the custom element configuration object that the consumer writes above, when they are not using decorators. The final representation, after normalization, needs lookups by both attr name and property name. However, during configuration, I think it makes most sense to provide the data in terms of the property name, since that's what is present in the class code. So, calling that "attributes" seems strange. We could call it "properties" I guess.

nicholasrice:
Yea I can see that being awkward too. "properties" is fine by me, but I'm also okay leaving as-is if we're not passionate about a more accurate description.

EisenbergEffect:
@nicholasrice Any thoughts on changing `CustomElement.define` to `FastElement.define`?

nicholasrice:
would the decorator change to `@fastElement` as well?

EisenbergEffect:
If we changed the decorator too, I worry that would be confusing since it's just a single letter case different. I believe litelement has customElement as its decorator and LitElement as its base class. So, if we had customElement and FastElement we would at least be aligning with that pattern.

nicholasrice:
Cool - I'm good with it

EisenbergEffect:
Ok, I'll do all these changes in a follow-up PR so that this one doesn't get too out of hand. Just need one more approval on this one @janechu @chrisdholt 🤣

PR #2732, opened 2020-03-04
chore: add badge web component spec

eljefe223:
# Description
Adds spec for `<fast-badge/>`

This assumes we want similar functionality that we have in the current react version of badge.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -67,6 +67,10 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 </fast-badge>
 ```
 
+### Accessibility
+
+There are currently no accessibility requirements for badge.
+
 ### Globalization
 
 *Badge* should mirror in RTL languages, meaning the *badge* should flip to the other side of the item it is labeling.
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -31,9 +31,7 @@ A badge should allow the following properties:
 
 *Attributes:*
 - `fill`: string
-- `size`: enum
-    - `small`
-    - `large`
+- `size`: string
 
 *Slots:*
 - `default`
@@ -42,16 +40,25 @@ A badge should allow the following properties:
 
 *Template:*
 ```
-<span style={{background: #FFD800}}>
-    <slot></slot>
-</span>
+<slot></slot>
 ```
 
 ## Implementation
 
 ```
+fast-badge {
+   --fill-primary: rgba(255, 0, 0, 1);
+   --fill-secondary: #00FF00;
+   --fill-transparent: transparent;
+   --size-small: 12px;
+   --size-medium: 16px;
+   --size-large: 20px;
+}
+```
+
+```
 <fast-badge
-    fill="#FFD800"
+    fill="primary"
     size="large"
 >
     New
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -47,7 +47,7 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 
 *Template:*
 ```
-<span>
+<span style={{background: #FFD800}}>
     <slot></slot>
 </span>
 ```
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -12,6 +12,7 @@ Typical use cases include, but are not limited to, denoting a sale or new item, 
 
 A badge should allow the following properties:
 - `fill`, a string that passes the color value for the backplate. Defaults to "transparent".
+- `circular`, if circular styling is desired.
 - `size`, the size the badge is rendered.
 
 ### Prior Art/Examples
@@ -31,7 +32,11 @@ A badge should allow the following properties:
 
 *Attributes:*
 - `fill`: string
-- `size`: string
+- `circular`: boolean
+- `size`: enum
+  - small
+  - medium
+  - large
 
 *Slots:*
 - `default`
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -12,7 +12,7 @@ Typical use cases include, but are not limited to, denoting a sale or new item a
 ### Features
 
 A badge should allow the following properties:
-- `filled`, whether the badge draws a backplate or not.
+- `fill`, a string that passes the color value for the backplate. Defaults to "transparent".
 - `size`, the size the badge is rendered.
 
 ### Risks and Challenges
@@ -35,8 +35,10 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 - `fast-badge`
 
 *Attributes:*
-- `filled`: boolean
+- `fill`: string
 - `size`: enum
+    - `small`
+    - `large`
 
 *Slots:*
 - `default`
@@ -45,7 +47,7 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 
 *Template:*
 ```
-<span>
+<span style={{backgtound: "#FFD800"}}>
     <slot></slot>
 </span>
 ```
@@ -54,8 +56,8 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 
 ```
 <fast-badge
-    filled={true}
-    size="Large"
+    fill="#FFD800"
+    size="large"
 >
     New
 </fast-badge>
@@ -75,4 +77,4 @@ No dependencies outside of fast-element itself.
 
 ### Documentation
 
-*Badge* component is a short inline [`<span>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) element highlight an item and attract attention or flag status, such as “New”, “Sale”, or other short phrases. It is not interactive, however, it generally overlays interactive content or is positioned close to a commanding component.
+*Badge* component is an inline element meant to highlight an item / attract attention or flag status, such as “New”, “Sale”, or other short phrases. It is not interactive, however, it generally overlays interactive content or is positioned close to a commanding component.
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -2,22 +2,17 @@
 
 ## Overview
 
-*Badge* component is used to highlight an item and attract attention or flag status, such as “New”, “Sale”, or other short phrases.
+*Badge* component is used to highlight an item and attract attention or flag status, such as “New”, “Sale”, or other short phrases or a number to represent unread messages, updates available, etc..
 
 ### Use Cases
-Keep the *badge* label short and succinct &ndash; it should not be a sentence or a long phrase (recommend 2-3 words). Short phrases work well, such as “App of the day,” but keep in mind that they will generally become longer when localized. Keep the text to nouns and adjectives if possible (e.g. “New” or “New today”), so the user won’t think it is an actionable element.
 
-Typical use cases include, but are not limited to, denoting a sale or new item and flagging an item as part of a category
+Typical use cases include, but are not limited to, denoting a sale or new item, flagging an item as part of a category or representing a value of unread messages.
   
 ### Features
 
 A badge should allow the following properties:
 - `fill`, a string that passes the color value for the backplate. Defaults to "transparent".
 - `size`, the size the badge is rendered.
-
-### Risks and Challenges
-
-It's import to not over dominate the users screen with the use of *badge*. They should be used sparingly to denote special status or draw attention to a particular item.
 
 ### Prior Art/Examples
 - [FAST-DNA (React)](https://explore.fast.design/components/badge)
@@ -65,7 +60,7 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 
 ### Accessibility
 
-There are currently no accessibility requirements for badge.
+Ensure text meets WCAG 2.1 color contrast against background.
 
 ### Globalization
 
@@ -74,7 +69,3 @@ There are currently no accessibility requirements for badge.
 ### Dependencies
 
 No dependencies outside of fast-element itself.
-
-### Documentation
-
-*Badge* component is an inline element meant to highlight an item / attract attention or flag status, such as “New”, “Sale”, or other short phrases. It is not interactive, however, it generally overlays interactive content or is positioned close to a commanding component.
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -43,12 +43,6 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 
 ### Anatomy and Appearance
 
-*Parts:*
-- label
-- checkbox
-- status-indicator
-- status-message
-
 *Template:*
 ```
 <span>
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -47,7 +47,7 @@ It's import to not over dominate the users screen with the use of *badge*. They 
 
 *Template:*
 ```
-<span style={{backgtound: "#FFD800"}}>
+<span>
     <slot></slot>
 </span>
 ```
--- a/specs/badge.md
+++ b/specs/badge.md
@@ -5,7 +5,7 @@
 *Badge* component is used to highlight an item and attract attention or flag status, such as “New”, “Sale”, or other short phrases.
 
 ### Use Cases
-Keep the *badge* label short and succinct &ndash; it should not be a sentence or a long phrase (recommend 2-3 words). Short phrases are work well, such as “App of the day,” but keep in mind that they will generally become longer when localized. Keep the text to nouns and adjectives if possible (e.g. “New” or “New today”), so the user won’t think it is an actionable element.
+Keep the *badge* label short and succinct &ndash; it should not be a sentence or a long phrase (recommend 2-3 words). Short phrases work well, such as “App of the day,” but keep in mind that they will generally become longer when localized. Keep the text to nouns and adjectives if possible (e.g. “New” or “New today”), so the user won’t think it is an actionable element.
 
 Typical use cases include, but are not limited to, denoting a sale or new item and flagging an item as part of a category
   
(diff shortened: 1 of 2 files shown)


specs/badge.md
```diff
@@ -0,0 +1,84 @@
...
+*Attributes:*
+- `filled`: boolean
+- `size`: enum
+
+*Slots:*
+- `default`
+
+### Anatomy and Appearance
+
+*Parts:*
```
bheston: Copy / paste

specs/badge.md
```diff
@@ -0,0 +1,78 @@
...
+
+*Badge* should mirror in RTL languages, meaning the *badge* should flip to the other side of the item it is labeling.
+
+### Dependencies
+
+No dependencies outside of fast-element itself.
+
+### Documentation
+
+*Badge* component is a short inline [`<span>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) element highlight an item and attract attention or flag status, such as “New”, “Sale”, or other short phrases. It is not interactive, however, it generally overlays interactive content or is positioned close to a commanding component.
```
nicholasrice: > *Badge* component is a short inline [`<span>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) element

This doesn't quite make sense because in a custom-element world it is actually a "fast-badge" element, not a span. I think just noting it as inline is sufficient.

specs/badge.md
```diff
@@ -0,0 +1,78 @@
...
+---
+
+### API
+
+*Component name:*
+- `fast-badge`
+
+*Attributes:*
+- `filled`: boolean
+- `size`: enum
```
nicholasrice: what are the values of this enum?

specs/badge.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Use Cases
+Keep the *badge* label short and succinct &ndash; it should not be a sentence or a long phrase (recommend 2-3 words). Short phrases work well, such as “App of the day,” but keep in mind that they will generally become longer when localized. Keep the text to nouns and adjectives if possible (e.g. “New” or “New today”), so the user won’t think it is an actionable element.
+
+Typical use cases include, but are not limited to, denoting a sale or new item and flagging an item as part of a category
+  
+### Features
+
+A badge should allow the following properties:
+- `filled`, whether the badge draws a backplate or not.
```
nicholasrice: This just controls a background color right? I wonder if it makes sense to change this to "fill" and have it accept a string, defaulting to "transparent" (or whatever the default typically is)

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+
+---
+
+### API
+
+*Component name:*
+- `fast-badge`
+
+*Attributes:*
+- `fill`: string
```
EisenbergEffect: Should this be handled with a CSS custom property rather than an HTML attribute? If handled with a CSS prop then the consumer can style these like any other. Not sure how we would handle size that way, if the values are a string enum. Fill should be easy enough. If we don't use a CSS prop here, I think we should have a strong justification for that.

nicholasrice: I was chatting offline with @eljefe223  about this yesterday. I'd agree - I don't think a "fill" property that accepts a color string provides much utility. As you say, you could just as easily style it:
```css
fast-badge { --fill: red; }
/* or */
fast-badge { background-color: red }
```
or
```html
<fast-badge style="background: red;" />
```

The struggle we've always had around badge is how to make common app scenarios easy to apply but also make it intuitively extensible. One idea is to have a "fill" property map to author-defined custom properties:

```css
fast-badge {
   --fill-foo: red;
   --fill-bar: green;
   --fill-transparent: transparent
}
```

Then in HTML, those could easily used:
```html
<fast-badge fill="transparent" />
<fast-badge fill="foo" />
```

You could still use any of the above css override techniques to create custom instances.

specs/badge.md
```diff
@@ -0,0 +1,80 @@
+# fast-badge
+
+## Overview
+
+*Badge* component is used to highlight an item and attract attention or flag status, such as “New”, “Sale”, or other short phrases.
```
chrisdholt: I'm not sure that the examples from Material or Ant fall into this overview. I do think we should consider them when we're building out this instance though. Can we expand the overview to include these?

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Use Cases
+Keep the *badge* label short and succinct &ndash; it should not be a sentence or a long phrase (recommend 2-3 words). Short phrases work well, such as “App of the day,” but keep in mind that they will generally become longer when localized. Keep the text to nouns and adjectives if possible (e.g. “New” or “New today”), so the user won’t think it is an actionable element.
```
chrisdholt: This seems more like general design guidance as opposed to use cases; is there a better place for this to live?

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+
+Typical use cases include, but are not limited to, denoting a sale or new item and flagging an item as part of a category
```
chrisdholt: Could we expand on these a bit to cover the various examples you have below? Ant & Material's initial examples don't particularly fit into the use cases outlined.

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+  
+### Features
+
+A badge should allow the following properties:
+- `fill`, a string that passes the color value for the backplate. Defaults to "transparent".
+- `size`, the size the badge is rendered.
```
chrisdholt: When thinking about size, how should we quantify that? I'm not sure small and large would be scalable. Could you expand on if these are CSS properties and if so, how they might support the variations in size across design systems?

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Risks and Challenges
+
+It's import to not over dominate the users screen with the use of *badge*. They should be used sparingly to denote special status or draw attention to a particular item.
```
chrisdholt: Is this a risk or challenge?

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+### API
+
+*Component name:*
+- `fast-badge`
+
+*Attributes:*
+- `fill`: string
+- `size`: enum
+    - `small`
+    - `large`
```
chrisdholt: I'm not sure small & large will scale. If we want to introduce a larger badge or a smaller badge then our sizes need to change. Is there a way we could approach this that might allow the system to scale more?

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+
+*Slots:*
+- `default`
+
+### Anatomy and Appearance
+
+*Template:*
+```
+<span style={{background: #FFD800}}>
```
chrisdholt: Is this still accurate based on the above comments regarding `fill`?

chrisdholt: Also, should we consider removing the span here? Relevant conversation [here](https://github.com/microsoft/fast-dna/pull/2743/files/984788cdf64fd400f3527dbb1c7e30c96b6b5524#r388507785)

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+    fill="#FFD800"
+    size="large"
+>
+    New
+</fast-badge>
+```
+
+### Accessibility
+
+There are currently no accessibility requirements for badge.
```
chrisdholt: I believe WCAG 2.1 color contrast would be a requirement here for the background vs text.

specs/badge.md
```diff
@@ -0,0 +1,80 @@
...
+
+*Badge* should mirror in RTL languages, meaning the *badge* should flip to the other side of the item it is labeling.
+
+### Dependencies
+
+No dependencies outside of fast-element itself.
+
+### Documentation
+
+*Badge* component is an inline element meant to highlight an item / attract attention or flag status, such as “New”, “Sale”, or other short phrases. It is not interactive, however, it generally overlays interactive content or is positioned close to a commanding component.
```
chrisdholt: I think we can remove this additional documentation down here. It reads as specific design guidance which I'm not sure would scale outside our specific guidance.

specs/badge.md
eljefe223: I assume you are talking about a badge that adorns and item that displays a count?

specs/badge.md
eljefe223: I think it could be both. Its a risk to overuse than the special status looses its meaning. Its a challenge because there may be a temptation to over use it.

chrisdholt: I guess my point is that this seems very specific to *how* you use the badge, rather than risks or challenges in building out this component

specs/badge.md
chrisdholt: correct

specs/badge.md
eljefe223: you mean changing too `<div part="root">` or just removing all together? Don't we need a containing element independent of text to style?

specs/badge.md
eljefe223: Maybe we just leave out size. We can default to values mapped to our designSystem and app authors can just override if they need to. Or we do Nicks suggestion for `fill`.

`size` property could map to author-defined custom properties.

```
fast-badge {
   --size-small: 12px;
   --size-medium: 16px;
   --size-large: 20px;
}
```
Then used like:

```
<fast-badge size="small" />
<fast-badge fill="large" />
```

specs/badge.md
eljefe223: The purpose of this section was not clear to me then. I guess we can remove it.

specs/badge.md
eljefe223: I'll remove for now. If add in some design guidance at a later date this could potentially live there.

specs/badge.md
eljefe223: see above comment

specs/badge.md
```diff
@@ -0,0 +1,71 @@
...
+    fill="#FFD800"
+    size="large"
+>
+    New
+</fast-badge>
+```
+
+### Accessibility
+
+Ensure text meets WCAG 2.1 color contrast against background.
```
chrisdholt: This is a good callout that I didn't see prior. Have you thought about how we might support this from an API standpoint? If I consider that I can set any particular value as the background of the badge, how do we support or ensure this?

specs/badge.md
chrisdholt: I think the concept of a map here similar to fill would add value 👍

specs/badge.md
chrisdholt: Sorry - There's two things here:

1. Inline style doesn't seem necessary given the conversation above on fill.
2. Removing the span altogether as we can simply style the host, which can act as the containing element.

specs/badge.md
eljefe223: I presumption is we will use whatever utility @EisenbergEffect comes up with that fills the role that our `<Background/>` react component did. So whatever fill value is passed the foreground color `--neutral-foreground-rest` will adjust to ensure proper contrast.

specs/badge.md
```diff
@@ -0,0 +1,78 @@
...
+## Implementation
+
+```
+fast-badge {
+   --fill-primary: rgba(255, 0, 0, 1);
+   --fill-secondary: #00FF00;
+   --fill-transparent: transparent;
+   --size-small: 12px;
+   --size-medium: 16px;
+   --size-large: 20px;
```
chrisdholt: Do these set the text, width, height, etc? How does the overall badge change based on these values?

eljefe223: Yeah, @nicholasrice chatted about this. Using users defined variables makes less sense for size. Since we would need to change multiple properties as you mentioned above (text, width, height). We should either go back to an `enum` for small, medium or large or have a better story for updating density for individual components (chatted with @bheston earlier about this) and decoupling sizing specific `attr` from any of our web components. I send out a PR for the badge web component to work through these very issues and currently using the `enum` approach.

chrisdholt: I'm not sure density would work to solve the size problem here. I may need a large badge in something like a "hero" or marketing carousel, but a smaller badge in a product card or something like that. The badge size there is relative to the control size in that scenario, but I would expect the density to stay the same on the page. I wouldn't want to change the density just around the badge implementation to get a larger badge in that scenario.

eljefe223: I agree here I just wanted to mention it because it was something @bheston was thinking about.

specs/badge.md
chrisdholt: @eljefe223 unresolving this for the time being to bring attention back to the naming / methodology of setting "size". I'm still concerned that small, medium, large will not scale. What happens if we want to or need to add a new size between any of these? Alternatively, what happens if we want a larger or smaller size to be added? Large no longer is large, or we need to add "larger"

specs/badge.md
```diff
@@ -0,0 +1,83 @@
...
+<slot></slot>
+```
+
+## Implementation
+
+```
+fast-badge {
+   --fill-primary: rgba(255, 0, 0, 1);
+   --fill-secondary: #00FF00;
+   --fill-transparent: transparent;
```
chrisdholt: We don't use primary anywhere as of now - what would this map to? Typically we have avoided "primary" and "secondary" (see button). I would expect some alignment across the system; is this a recommendation for a system wide change? If so, I think it's worthy of its own thread.

specs/badge.md
eljefe223: What do you propose? 1,2,3?

specs/badge.md
eljefe223: These are app author defined variables, they can be named whatever they choose.

chrisdholt: Can you elaborate on how that works from a technical standpoint? For instance, if I want three states with my own semantics, how does that get set on our end? 

In addition to the above, if these are defined by the app author, can we make sure that's clear in the spec? I was reading it as though those are set by us, as they appear to have the values being defined and not both the property and the values.

eljefe223: In the template:
```
<div style="background-color: var(--badge-fill-${x => x.fill})">
   ...
</div>
```

Implementation
```
<style>
  fast-badge {
    --badge-fill-foo: red;
  }
</style>

<fast-badge fill="foo"/>

```
I can make this clearer in the spec.

specs/badge.md
nicholasrice: @eljefe223,  I think @chrisdholt is proposing (correct me if I'm wrong) using the same implementation that "fill" does for this component, transforming the value of this attribute into CSS Custom Properties used by the component.

chrisdholt: Yes, I think that by providing a methodology for applying *any* size measurement with preferred semantics allows me to have "one size fits all" badge, or "small, medium, large", or anything.

eljefe223: @nicholasrice I'm confused... We talked about this remember? Originally I proposed this, but the issue is that size effects more than one property. `font-size` and `line-height` for the standard badge and `min-width` and `height` for the circular badge. Where `fill` can map to a variable since it only changes one property.

eljefe223: I don't think we want to create `attr` for each of those properties.

nicholasrice: Ahh yes. Perhaps we just make badge support these size classes (instead of a size attribute):

```html
<fast-badge class="small"></fast-badge>
```
```css
:host(.small) { /* small styles */ }
```

That is extensible by authors to define new "sizes" or styles as needed.

Thoughts?

chrisdholt: I think the semantics of how we are defining size size is a primary problem here. If we define small, medium and large while I can add my own new sizes using classes, they may not be intuitively extensible. For instance, if I want a smaller size, I could use "smaller" as a class name. Similarly if I want something larger than large, that seems reasonable as well. However we have always seen problems when we want something between small and medium or between medium and large. The naming of small, medium, large is all relative to one another. If my design system, or even if our own system calls for a size between any of these, we now need to come up with something less-semantic or intuitive.

nicholasrice: Do you have a proposal?

chrisdholt: I just walked through the examples below in prior art/examples and I actually don't see a primary use case for size in the way we are using it outside of the existing FAST implementation. Others do have the ability to have more simple badging with just a dot or indicator, which *could* correlate to size, but it's not their preferred method of identifying that variation. I'm wondering if size common enough to merit first class support over something more common like the dot variation of the others below.

In terms of what it means for size, I would expect that the visual size of the badge would respond based on the content - such that if I applied a larger instance of the type ramp the badge would grow to accommodate it. I do wonder if there is a way to drop "size" and build this in a way that the badge visuals respond well to different text sizes and manage "size" that way.

eljefe223: Resolving based on our conversation we had offline. Removing `size` from api

specs/badge.md
bheston: Opening this back up because I think it represents the same question about how we support different appearances on any component. We've talked about "style sets" or "appearances", essentially a grouping of styling properties. It's not as simple as just setting a fill because the foreground needs to respond. Also, if setting a fill, it's probable that the value is used elsewhere.

The closest parallel I think would be a good example is how we describe an input to either be filled or outlined. Is that an "appearance" property, or is it just in css? Do we encourage people to make two templates and derive components like FilledInput and OutlinedInput?

Or perhaps eventually it's more nuanced than that and the appearance might be determined by the context, like on L1 it's outlined and on L2 and above it's filled. Obviously very specific example, but the general idea being that some logic in the component or styling can determine the default set. Of course someone could override and customize as necessary with css, etc.

PR #2737, opened 2020-03-04
feat: add card as a new web component

chrisdholt:
# Description
This PR adds `card` as a new web component based on [the spec](https://github.com/microsoft/fast-dna/blob/master/specs/card/card.md).

Outstanding issues:
- While we have a stub of the design system provider in, we have not yet included the concept of `elevation`. This PR includes an initial stub for elevation but does not resolve the overall design system implementation.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/card/card.styles.ts
+++ b/packages/fast-components/src/card/card.styles.ts
@@ -1,6 +1,10 @@
 import { css } from "@microsoft/fast-element";
 
 export const CardStyles = css`
+    :host([hidden]) {
+        display: none;
+    }
+
     :host {
         --elevation: 4;
         height: var(--card-height, 200px);
--- a/packages/fast-components/src/card/card.styles.ts
+++ b/packages/fast-components/src/card/card.styles.ts
@@ -1,0 +1,15 @@
+import { css } from "@microsoft/fast-element";
+
+export const CardStyles = css`
+    :host {
+        --elevation: 4;
+        --height: "100%";
+        --width: "100%";
+        display: block;
+        background: var(--neutral-layer-card);
+        border-radius: calc(var(--elevated-corner-radius));
+        width: var(--width);
+        height: var(--height);
+        box-shadow: 0 0 calc(var(--elevation) * 1px) rgba(0, 0, 0, 0.5);
+    }
+`;
--- a/packages/fast-components/src/card/card.styles.ts
+++ b/packages/fast-components/src/card/card.styles.ts
@@ -3,13 +3,12 @@ import { css } from "@microsoft/fast-element";
 export const CardStyles = css`
     :host {
         --elevation: 4;
-        --height: "100%";
-        --width: "100%";
         display: block;
+        box-sizing: border-box;
+        height: "100%";
+        width: "100%";
         background: var(--neutral-layer-card);
-        border-radius: calc(var(--elevated-corner-radius));
-        width: var(--width);
-        height: var(--height);
+        border-radius: var(--elevated-corner-radius);
         box-shadow: 0 0 calc(var(--elevation) * 1px) rgba(0, 0, 0, 0.5);
     }
 `;
--- a/packages/fast-components/src/card/card.styles.ts
+++ b/packages/fast-components/src/card/card.styles.ts
@@ -3,10 +3,10 @@ import { css } from "@microsoft/fast-element";
 export const CardStyles = css`
     :host {
         --elevation: 4;
+        height: var(--card-height, 200px);
+        width: var(--card-width, 200px);
         display: block;
         box-sizing: border-box;
-        height: "100%";
-        width: "100%";
         background: var(--neutral-layer-card);
         border-radius: var(--elevated-corner-radius);
         box-shadow: 0 0 calc(var(--elevation) * 1px) rgba(0, 0, 0, 0.5);
--- a/packages/fast-components/src/card/card.styles.ts
+++ b/packages/fast-components/src/card/card.styles.ts
@@ -7,9 +7,10 @@ export const CardStyles = css`
 
     :host {
         --elevation: 4;
+        display: block;
+        contain: content;
         height: var(--card-height, 200px);
         width: var(--card-width, 200px);
-        display: block;
         box-sizing: border-box;
         background: var(--neutral-layer-card);
         border-radius: var(--elevated-corner-radius);
--- a/packages/fast-components/src/card/card.styles.ts
+++ b/packages/fast-components/src/card/card.styles.ts
@@ -9,8 +9,8 @@ export const CardStyles = css`
         --elevation: 4;
         display: block;
         contain: content;
-        height: var(--card-height, 200px);
-        width: var(--card-width, 200px);
+        height: var(--card-height, 100%);
+        width: var(--card-width, 100%);
         box-sizing: border-box;
         background: var(--neutral-layer-card);
         border-radius: var(--elevated-corner-radius);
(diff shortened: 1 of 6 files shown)


packages/fast-components/src/card/card.styles.ts
```diff
@@ -0,0 +1,15 @@
...
+export const CardStyles = css`
+    :host {
+        --elevation: 4;
+        --height: "100%";
+        --width: "100%";
+        display: block;
+        background: var(--neutral-layer-card);
+        border-radius: calc(var(--elevated-corner-radius));
+        width: var(--width);
+        height: var(--height);
```
eljefe223: Just curious, are these values coming from the design system and you are reassigning the variables here? Why not just assign the height and width 100%?

nicholasrice: I agree - I'm not sure we really need to surface custom properties for this. An author can just as easily style the width / height properties directly

packages/fast-components/src/card/card.styles.ts
```diff
@@ -0,0 +1,15 @@
+import { css } from "@microsoft/fast-element";
+
+export const CardStyles = css`
+    :host {
+        --elevation: 4;
+        --height: "100%";
+        --width: "100%";
+        display: block;
+        background: var(--neutral-layer-card);
+        border-radius: calc(var(--elevated-corner-radius));
```
nicholasrice: I don't think you need to calc this.

packages/fast-components/src/card/card.styles.ts
chrisdholt: I think there's 2 questions here with the underlying question of why is this here? :)

@eljefe223 these values aren't coming from the design system. They are custom properties I was exposing for the card. 

@nicholasrice I had a thought that we might be able to set the custom property from once from a parent and have all children respond accordingly, but that's not actually possible. If it were, I think it would've been a helpful use case. Since it doesn't do what I expected, it's terrible.
```html
    <div style="--width: 300px; --height: 163px"> // doesn't work, bad idea :)
        <fast-card>Foo</fast-card>
        <br>
        <fast-card>Bar</fast-card>
        <br>
        <fast-card>Baz</fast-card>
    </div>
```

nicholasrice: yea you're defining the property in the host. You might try:

```css
:host {
   height: var(--hard-height, 200px);
   width: var(--hard-width, 200px);
}
```

nicholasrice: That does provide a nice utility using the style attribute over creating a real css selector:

```css
.foo fast-card { height: x; width: y }
```

chrisdholt: I do like it - I'm open to exploring that if you are. I don't think it's a ton of overhead.

chrisdholt: Took a shot at using the custom variables here. Allows quite a bit of flexibility in terms of setting width/height. You can now use the vars in the parent, or you can define via class override, etc. Lots of utility to set width/height which I think if a goal for cards.

packages/fast-components/src/card/card.styles.ts
```diff
@@ -0,0 +1,14 @@
+import { css } from "@microsoft/fast-element";
+
+export const CardStyles = css`
+    :host {
+        --elevation: 4;
```
EisenbergEffect: Should we prefix our CSS variable names? @chrisdholt @nicholasrice has there been any thought about this regarding unification with other MS teams around variable names?

EisenbergEffect: We'll want to include the default style for `hidden` for the host. See the docs I wrote on styling in shadow dom. Also, consider adding `contain`.

packages/fast-components/src/card/card.template.ts
```diff
@@ -0,0 +1,8 @@
+import { html } from "@microsoft/fast-element";
+import { Card } from "./card";
+
+export const CardTemplate = html<Card>`
+  <div part="root">
+    <slot></slot>
+  </div>
```
chrisdholt: @nicholasrice depending on where we land with the conversation on your PR, do you think we even need a root class on the card here? Would it be unintuitive to just remove it and leave it off? Ultimately you just target the host or the stuff you put inside the card. It may be an outlier from our other components in terms of consistency, but it seems logical to me to remove it...

packages/fast-components/src/card/card.styles.ts
chrisdholt: @nicholasrice has been working on the design tokens side of things, so he likely has more insight here. `elevation` and other variables I've used for now are just stubs in my mind. I do think we need to land a story here, but wanted to get things rolling while those convos are happening.

@EisenbergEffect can you elaborate on `contain` or point me to the relevant docs?

EisenbergEffect: https://github.com/microsoft/fast-dna/blob/master/packages/fast-element/docs/building-components.md#shadow-dom-styling

chrisdholt: Updated - very cool.

packages/fast-components/src/card/card.template.ts
EisenbergEffect: I agree, it's probably not needed in this case.

packages/fast-components/src/card/card.styles.ts
nicholasrice: I'll sync with Levi on whether they have landed a token prefix or not. I've noticed some "-ms-" prefixes in some of the prototyping they're doing, but I don't know if that is something they have actually landed.

My gut reaction is to *also* prefix, but if the intent is to arrive at a common token naming and providing theme interop that may not actually be desired.

packages/fast-components/src/card/card.template.ts
nicholasrice: checkbox has no "root" element. I actually think *most* components (if any) will need a "root" element, since the host essentially serves that function.

packages/fast-components/src/card/card.styles.ts
nicholasrice: They have not landed that implementation detail so I think this is fine for now. I can write up a discussion issue to propose a decision tree for this, but we'll likely want to differentiate by whether the property used is a global design system token, or a more proprietary concept.

packages/fast-components/src/card/card.styles.ts
nicholasrice: Yea I'm open to it *I think* - my concern is using very generic terms like "height" and "width", and specifically I'm worried about un-intentional propagation of those properties to child elements that might use the same names.

For instance, both this card and the dialog PR implement the `--width` and `--height` custom properties. If I were to nest them and set the `--height` property on the dialog, the card would also implement that new value, which is likely undesriable.

chrisdholt: @nicholasrice should we prefix here with the component name? Seems safer and avoids the potential for the above conflict.

nicholasrice: It does - it also opens the door for the "why isn't *padding* parameterized, why isn't *insert property name here* parameterized" questions and were in the game of managing a bunch of component-specific custom property APIs.

While not 100% parallel, [this discussion](https://github.com/microsoft/fast-dna/issues/2685) does address managing component custom property APIs. It might be worth exploring this in a discussion format before setting the precedent in code. Where do we draw the line on what we want to expose as CSS custom properties?

packages/fast-components/src/card/card.template.ts
chrisdholt: updated this

PR #2741, opened 2020-03-04
feat: add form association component

nicholasrice:
# Description
Adds the initial implementation of the utility specced out here: https://github.com/microsoft/fast-dna/blob/master/specs/form-associated-custom-element.md

There are a few follow ups that will be tracked, including adding boolean attribute support and label click handling.

## Motivation & context
This utility provides infrastructure to support custom elements with form association so that the custom element can participate in form submission, have form association, and form validation features. It uses the ElementInternals API when it is available, but will route API calls through a proxy element when the ElementInternals API is not supported.

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/element-internals.d.ts
```diff
@@ -0,0 +1,83 @@
...
+        state?: File | string | FormData | null
+    ): void;
+    /**
+     * Marks internals target element as suffering from the constraints indicated by the flags argument,
+     * and sets the element's validation message to message.
+     * If anchor is specified, the user agent might use
+     * it to indicate problems with the constraints of internals target
+     * element when the form owner is validated interactively or reportValidity() is called.
+     */
+    setValidity(flags: ValidityStateFlags, message?: string, anchor?: HTMLElement): void;
```
janechu: I think the message is required, from reading https://html.spec.whatwg.org/multipage/custom-elements.html#elementinternals

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,263 @@
...
+
+                break;
+        }
+    }
+
+    /**
+     * Used to stop propagation of proxy element events
+     * @param e Event object
+     */
+    private stopPropogation(e: Event): void {
```
janechu: *stopPropagation

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,263 @@
...
+ */
+/* tslint:disable:member-ordering */
+export abstract class FormAssociated<
+    T extends HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement
+> extends FastElement {
+    /**
+     * Must evaluate to true to enable elementInternals.
+     * Feature detects API support and resolve respectively
+     */
+    public static get formAssociated(): boolean {
```
EisenbergEffect: We should perform this check once and cache the result so it can be used across all form elements and all properties that branch based on this condition.

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,263 @@
...
+        } else if (
+            this.proxy instanceof HTMLElement &&
+            this.proxy.ownerDocument &&
+            this.id
+        ) {
+            // Labels associated by wraping the element: <label><custom-element></custom-element></label>
+            const parentLabels = this.proxy.labels;
+            // Labels associated using the `for` attribute
+            const forLabels = Array.from(
+                this.proxy.ownerDocument.querySelectorAll(`[for='${this.id}']`)
```
EisenbergEffect: Rather than querying the ownerDocument, I think we need to call `getRootNode()` on `proxy`. This will either return the document or the nearest shadow dom root. This is important because if the component is used inside of another shadow dom, the query selector against document will not find the labels.

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,263 @@
...
+            );
+
+            return !!parentLabels
+                ? forLabels.concat(Array.from(parentLabels))
+                : forLabels;
+        } else {
+            return [];
```
EisenbergEffect: Inside of the fast-element library, there's now a const named `emptyArray`. This allows us to share the same frozen empty array instance. You could use that here. Saves a bit of memory as we don't have to instantiate new arrays for every API that returns an empty, readonly array.

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,263 @@
...
+    /**
+     * Whether the element should be autofocused
+     */
+    @attr
+    public autofocus: boolean; // Map to proxy element
+
+    @attr
+    public disabled: boolean = false;
+    private disabledChanged(): void {
+        if (this.proxy instanceof HTMLElement) {
```
EisenbergEffect: In what case will proxy not bee an HTMLElement? Is this the case where we are using element internals instead of the proxy?

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,263 @@
...
+     * The proxy implementation should be transparent to
+     * the app author, so block these events from emitting.
+     */
+    private proxyEventsToBlock = ["change", "click"];
+
+    constructor() {
+        super();
+
+        if (FormAssociated.formAssociated) {
+            this.elementInternals = (this as any).attachInternals();
```
EisenbergEffect: As a follow-up PR, I wonder if we could create an API like `controller.attachInternals(this)` which would attach and return the internals or create the proxy and a fake internals implementation that has many of the properties above on it.

packages/fast-components/src/form-associated/index.ts
nicholasrice: No, its during initialization. I think it is a side effect of the `@attr` decorator - none of the member properties are available the first time this cb is called.

packages/fast-components/src/form-associated/index.ts
nicholasrice: I think that is likely the best long-term solution since the ElementInternals object is likely to expand over time.

EisenbergEffect: Ah. If the proxy isn't created until after they are set...yeah that makes sense.

packages/fast-components/src/form-associated/index.ts
nicholasrice: added a cached property - let me know if there was a more efficient way to do that though.

packages/fast-components/src/form-associated/index.ts
nicholasrice: good call - updated.

packages/fast-components/src/form-associated/index.ts
nicholasrice: cool - because that object is readonly I had to update the typing of the `labels` property to a `ReadonlyArray<Node>` but I think that is desired anyway. All return values are now the product of `Object.freeze`

nicholasrice:
@EisenbergEffect Thanks Rob - comments addressed.

packages/fast-components/src/form-associated/index.ts
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * Must evaluate to true to enable elementInternals.
+     * Feature detects API support and resolve respectively
+     */
+    private static cachedFormAssociated: boolean | null = null;
+    public static get formAssociated(): boolean {
+        return FormAssociated.cachedFormAssociated === null
+            ? (FormAssociated.cachedFormAssociated = "ElementInternals" in window)
+            : FormAssociated.cachedFormAssociated;
+    }
```
EisenbergEffect: I'd consider pulling this out of the class instead. Something like this:

```TypeScript
const supportsElementInternals = (function() {
  return "ElementInternals" in window;
})();
```

Put that at the module level and then reference the const value thoughout.

PR #2743, opened 2020-03-05
chore: add checkbox component spec

nicholasrice:
# Description
Adds spec document for checkbox component. This component builds off the [form-assocation](https://github.com/microsoft/fast-dna/blob/master/specs/form-associated-custom-element.md) component spec.

## Issue type checklist

## Process & policy checklist


specs/checkbox.md
```diff
@@ -0,0 +1,128 @@
+# FAST-Checkbox
```
janechu: Can we just call this "Checkbox"? We seem to be diverging among us on how to title these specs.

specs/checkbox.md
```diff
@@ -0,0 +1,128 @@
...
+- `checked: boolean`
+  - The current checked state of the checkbox
+- `indeterminate: boolean`
+  - The indeterminate state. Independent of checked
+
+*Content attributes*
+- `readonly`
+  - The checkbox should be submitted with the form but should not be editable.
+- `value`
+  - Defaults to "on" to match `input[type="checkbox"]`
```
janechu: "on"? I thought it was true/false?

specs/checkbox.md
```diff
@@ -0,0 +1,128 @@
...
+- `change: CustomEvent`
+  - no custom data
+  - bubbles
+
+### Anatomy and Appearance
+
+```HTML
+<!-- shadow root -->
+<label part="label"><slot></slot></label>
+<div part="root">
```
janechu: It's interesting that the "root" in this case is not the wrapper that includes the label. Is our convention going forward that root does not necessarily include other pieces of UI? I would assume that switch would have a root so I could refer to another component with a similar concept of labelling but there's no root on there (I think maybe that was a miss)

specs/checkbox.md
```diff
@@ -0,0 +1,128 @@
...
+*Slotted Content/Slotted Classes*
+*CSS Parts*
+- root
+- label
+- state-indicator
+
+### States
+**checked**: `true` or `false`
+The checked state can be toggled by:
+- Clicking the checkbox (or any of it's labels
```
janechu: You dropped this ")"

specs/checkbox.md
nicholasrice: nope - the `checked` property is true/false. Value is "on"

specs/checkbox.md
nicholasrice: Yea this was meant to spark conversation :)

the "root" of this component is the custom element, `:host` - in this instance we don't really need a wrapping element for the whole component, but the going convention seems to be to have a "root" part. Really, this is just the "checkbox" portion, but `fast-checkbox::part(checkbox)` also feels strange.

chrisdholt: interesting - if we don't need a wrapping element, I'm all in favor of this approach. I'm going to update my two PR's with this. One question this does bring up though is "how do we decide what is the 'root'"?

janechu: I think since this is primarily for styling there should be a visual block aspect to it, similar to what's going on here. When we refer to the checkbox we refer to the control that looks like a box with a check in it, not necessarily any extra text or popups etc that may come along with it. That said, if there comes a need to refactor and add a wrapper to the component and we've already used the word "root", then what should that wrapper be called? Perhaps "root" should refer to the wrapping instance we should use the word "control" should refer to the actual interactive piece.

nicholasrice: See changes - this changed a bit while working through the implementation.

nicholasrice:
Merging this because the initial component is already merged

PR #2746, opened 2020-03-05
chore: add progress web component spec

eljefe223:
# Description

Adds spec for `<fast-progress/>`

This assumes we want similar functionality that we have in the current react version of progress.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/progress.md
+++ b/specs/progress.md
@@ -1,20 +1,19 @@
-# fast-progress
+# Progress
 
 ## Overview
 
-*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either be in an **indeterminate** or **determinate** state. This is determined by whether `value` attribute is passed to the component. 
+*Progress* is used to display the length of time a process will take (referred to as a **determinate** state) or to represent an unspecified wait time (referred to as an **indeterminate** state). *Progress* is typically visually represented by a circular or linear animation. When the `value` attribute is passed the state is **determinate**, otherwise it is **indeterminate**.
 
 ### Use Cases
 
-The *progress* component should be used for an element that displays the progress status for a task that takes a long time or consists of several steps. Typical use cases include loading elements or status of a download. 
+The *progress* component should be used for an element that displays the progress status for a task that takes a more than a few seconds or consists of several steps. Typical use cases include loading elements or status of a download. 
   
 ### Features
 
-A progress should allow the following properties:
+A progress should allow the following attributes:
 - `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
 - `min-value`, the minimum value.
 - `max-value`, the maximum value.
-- `circular`, whether the progress renders circular or not.
 - `paused`, whether the progress is paused or not.
 
 ### Prior Art/Examples
@@ -38,7 +37,6 @@ A progress should allow the following properties:
 - `value`: number
 - `min-value`: number
 - `max-value`: number
-- `circular`: boolean
 - `paused`: boolean
 
 ### Anatomy and Appearance
--- a/specs/progress.md
+++ b/specs/progress.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either me in an **indeterminate** state or **determinate** state this is determined by whether `value` attribute is passed to the component. 
+*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either be in an **indeterminate** or **determinate** state. This is determined by whether `value` attribute is passed to the component. 
 
 ### Use Cases
 
--- a/specs/progress.md
+++ b/specs/progress.md
@@ -12,8 +12,9 @@ The *progress* component should be used for an element that displays the progres
 
 A progress should allow the following attributes:
 - `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
-- `min-value`, the minimum value.
-- `max-value`, the maximum value.
+- `min`, the minimum value.
+- `max`, the maximum value.
+- `circular`, whether the progress renders circular or not.
 - `paused`, whether the progress is paused or not.
 
 ### Prior Art/Examples
@@ -35,9 +36,14 @@ A progress should allow the following attributes:
 
 *Attributes:*
 - `value`: number
-- `min-value`: number
-- `max-value`: number
+- `min`: number
+- `max`: number
+- `circular`: boolean
 - `paused`: boolean
+
+*Slots:*
+- `indetermimate-linear`
+- `indetermimate-circular`
 
 ### Anatomy and Appearance
 
@@ -73,10 +79,12 @@ A progress should allow the following attributes:
 ```
 <div role="progressbar">
     <div part="indeterminate" slot={"indeterminate"}>
-        <svg part="circular-container" viewBox="0 0 16 16">
-            <circle part="circular-background" cx="8px" cy="8px" r="7px" />
-            <circle part="circular-indeterminate" cx="8px" cy="8px" r="7px" />
-        </svg>
+        <slot name="indetermimate-circular"></slot>
+            <svg part="circular-container" viewBox="0 0 16 16">
+                <circle part="circular-background" cx="8px" cy="8px" r="7px" />
+                <circle part="circular-indeterminate" cx="8px" cy="8px" r="7px" />
+            </svg>
+        </slot>
     </div>
 </div>
 ```
@@ -99,8 +107,10 @@ A progress should allow the following attributes:
 ```
 <div role="progressbar">
     <div part="indeterminate" slot={"indeterminate"}>
-        <span part="linear-indeterminate-indicator-1"/>
-        <span part="linear-indeterminate-indicator-2"/>
+        <slot name="indetermimate-linear">
+            <span part="linear-indeterminate-indicator-1"/>
+            <span part="linear-indeterminate-indicator-2"/>
+        </slot>
     </div>
 </div>
 ```
@@ -111,8 +121,8 @@ A progress should allow the following attributes:
 ```
 <fast-progress
     value={50}
-    min-value={0}
-    max-value={100}
+    min={0}
+    max={100}
     circular={true}
 />
 ```
--- a/specs/progress.md
+++ b/specs/progress.md
@@ -1,0 +1,142 @@
+# fast-progress
+
+## Overview
+
+*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either me in an **indeterminate** state or **determinate** state this is determined by whether `value` attribute is passed to the component. 
+
+### Use Cases
+
+The *progress* component should be used for an element that displays the progress status for a task that takes a long time or consists of several steps. Typical use cases include loading elements or status of a download. 
+  
+### Features
+
+A progress should allow the following properties:
+- `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
+- `min-value`, the minimum value.
+- `max-value`, the maximum value.
+- `circular`, whether the progress renders circular or not.
+- `paused`, whether the progress is paused or not.
+
+### Prior Art/Examples
+- [FAST-DNA (React)](https://explore.fast.design/components/progress)
+- [Material UI](https://material-ui.com/components/progress/)
+- [Lightning Design / Linear](https://www.lightningdesignsystem.com/components/progress-bar/)
+- [Lightning Design / Circular](https://www.lightningdesignsystem.com/components/progress-ring/)
+- [Ant Design](https://ant.design/components/progress/)
+- [Atlassian / Linear](https://atlaskit.atlassian.com/packages/server/progress-bar)
+- [Atlassian / Circular](https://atlaskit.atlassian.com/packages/core/spinner)
+- [Windows (UWP)](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/progress-controls)
+
+---
+
+### API
+
+*Component name:*
+- `fast-progress`
+
+*Attributes:*
+- `value`: number
+- `min-value`: number
+- `max-value`: number
+- `circular`: boolean
+- `paused`: boolean
+
+### Anatomy and Appearance
+
+*Parts:*
+- determinate
+- indeterminate
+- circular-container
+- circular-background
+- circular-determinate
+- circular-indeterminate
+- linear-determinate
+- linear-indeterminate-indicator-1
+- linear-indeterminate-indicator-2
+
+*Circular Determinate Template:*
+```
+<div
+    role="progressbar"
+    aria-valuenow={50}
+    aria-valuemin={0}
+    aria-valuemax={100}
+>
+    <div part="determinate" slot={"determinate"}>
+        <svg part="circular-container" viewBox="0 0 16 16">
+            <circle part="circular-background" cx="8px" cy="8px" r="7px" />
+            <circle part="circular-determinate" style={{stroke-dasharray: "22px 44px"}} cx="8px" cy="8px" r="7px" />
+        </svg>
+    </div>
+</div>
+```
+
+*Circular Indeterminate Template:*
+```
+<div role="progressbar">
+    <div part="indeterminate" slot={"indeterminate"}>
+        <svg part="circular-container" viewBox="0 0 16 16">
+            <circle part="circular-background" cx="8px" cy="8px" r="7px" />
+            <circle part="circular-indeterminate" cx="8px" cy="8px" r="7px" />
+        </svg>
+    </div>
+</div>
+```
+
+*Linear Determinate Template:*
+```
+<div
+    role="progressbar"
+    aria-valuenow={50}
+    aria-valuemin={0}
+    aria-valuemax={100}
+>
+    <div part="determinate" slot={"determinate"}>
+        <div part="linear-determinate" style={{ width: 50% }}/>
+    </div>
+</div>
+```
+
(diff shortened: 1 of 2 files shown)


specs/progress.md
```diff
@@ -0,0 +1,142 @@
+# fast-progress
```
janechu: I know this was done on switch, but I think we need to standardize this to just "Progress", fast-progress is the tag name.

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Overview
+
+*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either be in an **indeterminate** or **determinate** state. This is determined by whether `value` attribute is passed to the component.
```
janechu: I would suggest this be re-written a little for clarity to:

*Progress* is used to display the length of time a process will take (referred to as a **determinate** state) or to represent an unspecified wait time (referred to as an **indeterminate** state). *Progress* is typically visually represented by a circular or linear animation. When the `value` attribute is passed the state is **determinate**, otherwise it is **indeterminate**.

specs/progress.md
```diff
@@ -0,0 +1,142 @@
+# fast-progress
+
+## Overview
+
+*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either be in an **indeterminate** or **determinate** state. This is determined by whether `value` attribute is passed to the component. 
+
+### Use Cases
+
+The *progress* component should be used for an element that displays the progress status for a task that takes a long time or consists of several steps. Typical use cases include loading elements or status of a download.
```
janechu: Can you further define what a "long time" is?

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+
+The *progress* component should be used for an element that displays the progress status for a task that takes a long time or consists of several steps. Typical use cases include loading elements or status of a download. 
+  
+### Features
+
+A progress should allow the following properties:
+- `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
+- `min-value`, the minimum value.
+- `max-value`, the maximum value.
+- `circular`, whether the progress renders circular or not.
```
janechu: I would think the styles change but there would be no need for a `circular` attribute.

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+
+*Progress* is used to display the length of a process or to represent an unspecified wait time. *Progress* can either be circular or linear. *Progress* can either be in an **indeterminate** or **determinate** state. This is determined by whether `value` attribute is passed to the component. 
+
+### Use Cases
+
+The *progress* component should be used for an element that displays the progress status for a task that takes a long time or consists of several steps. Typical use cases include loading elements or status of a download. 
+  
+### Features
+
+A progress should allow the following properties:
```
janechu: I would change properties to attributes.

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+- `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
+- `min-value`, the minimum value.
+- `max-value`, the maximum value.
+- `circular`, whether the progress renders circular or not.
+- `paused`, whether the progress is paused or not.
```
janechu: What's the use case for paused? In the determinate state the value would just not be changing, in the indeterminate state I can only assume the animation would freeze which would be odd behavior.

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+- circular-determinate
+- circular-indeterminate
+- linear-determinate
+- linear-indeterminate-indicator-1
+- linear-indeterminate-indicator-2
+
+*Circular Determinate Template:*
+```
+<div
+    role="progressbar"
```
janechu: I vaguely remember there was a reason, but why not use the progress element here instead?

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+    aria-valuenow={50}
+    aria-valuemin={0}
+    aria-valuemax={100}
+>
+    <div part="determinate" slot={"determinate"}>
+        <svg part="circular-container" viewBox="0 0 16 16">
+            <circle part="circular-background" cx="8px" cy="8px" r="7px" />
+            <circle part="circular-determinate" style={{stroke-dasharray: "22px 44px"}} cx="8px" cy="8px" r="7px" />
+        </svg>
```
janechu: We seem to be boxing this style in via these SVGs, @bheston can you give an opinion on how we should want to allow users to style this component?

specs/progress.md
```diff
@@ -0,0 +1,142 @@
...
+        <div part="linear-determinate" style={{ width: 50% }}/>
+    </div>
+</div>
+```
+
+*Linear Indeterminate Template:*
+```
+<div role="progressbar">
+    <div part="indeterminate" slot={"indeterminate"}>
+        <span part="linear-indeterminate-indicator-1"/>
```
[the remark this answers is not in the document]
eljefe223: Other idea would be to provide a slot and default the html for out implementation of the indeterminate animation.

specs/progress.md
EisenbergEffect: An important note about the aria attributes (and I can't remember if I saw this in other specs or not): we'll want the aria attributes to exist on the custom element host itself, rather than the nodes inside the shadow dom.

specs/progress.md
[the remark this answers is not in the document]
EisenbergEffect: I think we could have a single slot for the animation, and then inside of that have our own conditional rendering that uses the different default animations based on circular, linear, determinate, etc. We can do the conditional rendering using the `when` directive and that can all be fallback content inside of a slot.

specs/progress.md
```diff
@@ -0,0 +1,140 @@
...
+    value={50}
+    min-value={0}
+    max-value={100}
+    circular={true}
+/>
+```
+
+### States
+
+*Progress* can either be **determinate** or **indeterminate** depending on whether the `value` attribute is passed. *Progress* can render as a linear or circular progress depending on whether `true` or `false` is passed to the the `circular` attribute.
```
EisenbergEffect: Once we get the new feature in, we can have circular function as a true boolean attribute. If this gets implemented before that work is done, just note that you'll have to check for string values of true/false as a work around.

specs/progress.md
eljefe223: I resolved this in Code but leaving the comment open for discussion, But I believe was should keep it based on @EisenbergEffect comments.

specs/progress.md
```diff
@@ -0,0 +1,140 @@
...
+### Use Cases
+
+The *progress* component should be used for an element that displays the progress status for a task that takes a more than a few seconds or consists of several steps. Typical use cases include loading elements or status of a download. 
+  
+### Features
+
+A progress should allow the following attributes:
+- `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
+- `min-value`, the minimum value.
+- `max-value`, the maximum value.
```
EisenbergEffect: I think we might want to just call these `min` and `max` to match what native HTML does.

specs/progress.md
eljefe223: Had an in depth conversation with @EisenbergEffect on this topic so I thought I would summarize here:

We will do work in our components to write aria values to the host element based on attributes that come from the app author. So an app author would implement as you would expect:

```
<fast-progress
    value={50}
    min-value={0}
    max-value={100}
/>
```

But the DOM will show the component as:

```
<fast-progress
    value={50}
    min-value={0}
    max-value={100}
    aria-valuenow={50}
    aria-valuemin={0}
    aria-valuemax={100}
/>
```

and the snapshot of the html inside the shadow root as:

```
<div
    role="progressbar"
    aria-valuenow={50}
    aria-valuemin={0}
    aria-valuemax={100}
>
    <div part="determinate" slot={"determinate"}>
        <div part="linear-determinate" style={{ width: 50% }}/>
    </div>
</div>
```

This spec is correct as written because I'm showing "snap shots" of the DOM under **Anatomy and Appearance** and showing author implementation under the **Implementation** section.

specs/progress.md
```diff
@@ -0,0 +1,150 @@
+# Progress
+
+## Overview
+
+*Progress* is used to display the length of time a process will take (referred to as a **determinate** state) or to represent an unspecified wait time (referred to as an **indeterminate** state). *Progress* is typically visually represented by a circular or linear animation. When the `value` attribute is passed the state is **determinate**, otherwise it is **indeterminate**.
+
+### Use Cases
+
+The *progress* component should be used for an element that displays the progress status for a task that takes a more than a few seconds or consists of several steps. Typical use cases include loading elements or status of a download.
```
chrisdholt: Rather than what it shouldn't do, can we add some use cases for when someone may be trying to use it and what they are doing with it? I added a few for Dialog like this.

specs/progress.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Anatomy and Appearance
+
+*Parts:*
+- determinate
+- indeterminate
+- circular-container
+- circular-background
+- circular-determinate
+- circular-indeterminate
```
chrisdholt: Is circular a completely different implementation? For something like a determinate progress bar, Ant has good examples of both linear and circular instances: https://ant.design/components/progress/

specs/progress.md
```diff
@@ -0,0 +1,150 @@
...
+```
+<div role="progressbar">
+    <div part="indeterminate" slot={"indeterminate"}>
+        <slot name="indetermimate-linear">
+            <span part="linear-indeterminate-indicator-1"/>
+            <span part="linear-indeterminate-indicator-2"/>
+        </slot>
+    </div>
+</div>
+```
```
chrisdholt: Are these all independent?

eljefe223: Depending on the properties are passed to the `<Progress/>` component these scenarios would be conditionally rendered. This is simply meant to show the DOM structure of each potential scenario

specs/progress.md
janechu: In that case, can you investigate and add the reasoning into the spec?

specs/progress.md
eljefe223: I included that example above. I was wondering if it would simplify things if we just have 2 separate components `<fast-circular-progress/>` and `<fast-progress>`. Some libraries keep it as one others separate them out (Lightning Design - [Circular](https://www.lightningdesignsystem.com/components/progress-ring/) | [Linear](https://www.lightningdesignsystem.com/components/progress-bar/), Atlassian - [Circular](https://atlaskit.atlassian.com/packages/core/spinner) | [Linear](https://atlaskit.atlassian.com/packages/server/progress-bar))

specs/progress.md
eljefe223: Sorry, I'm confused by this. Where do I say what authors **should not** do with this component?

chrisdholt: Updated, I must have read this wrong 🤦‍♂

specs/progress.md
```diff
@@ -0,0 +1,150 @@
...
+*Attributes:*
+- `value`: number
+- `min`: number
+- `max`: number
+- `circular`: boolean
+- `paused`: boolean
+
+*Slots:*
+- `indetermimate-linear`
+- `indetermimate-circular`
```
janechu: Spelling, *indeterminate

specs/progress.md
janechu: I'm actually inclined towards splitting these, the Lighting Design and Atlassian showcase a more extensible model which I think is preferable. At that point we could do Progress Bar and Progress Spinner or Progress Ring, both are being used by Fluent and UWP respectively.

specs/progress.md
chrisdholt: thanks for clarifying!

specs/progress.md
nicholasrice: I believe that is why also, but it's been a while.

specs/progress.md
scomea: IMO we should be designing this component so that the visualization of progress can be extremely flexible at the template/styling level without changing the base code and in that sense any notion of "circular" should ideally be confined there.  Isn't circular just a style decision?

specs/progress.md
scomea: I agree with Jane, we should not embed specific style concepts in base component code if we can avoid it.

eljefe223: Circular is not a design language specific concept it is a control in its own. You simply can't restyle a linear progress to be circular, it requires different html elements and structure. Unless we are saying we don't won't a circular progress, we will need to write the template in a way that can support it.

specs/progress.md
eljefe223: See my above comment, circular is not just styling.

specs/progress.md
scomea: Can we do this with template changes only though?  I"m not yet clear on how we'd manage/pack that in our system (ie. the single component vs. multiple component).

specs/progress.md
bheston: With our current separation of component code, template, and styling, could this be one base `progress` class that deals with value and in/determinate mode and two templates, one for circular and one for linear? That seems most extensible as a consumer could create a dashboard style, a custom circular style, an illustration, etc.

I'm not sure if templates currently have the capability for this, but it seems ideal if the base component renders the container with the appropriate aria attributes, and the templates render the visual content.

This split seems like the most reuse of the basics without the need to duplicate many things like the parts or slots.

specs/progress.md
bheston: Cool, the aria attributes sounds like what I suggested above.

I do think it's reasonable that people may want to restyle the svg, but it sounds like that will be possible with a custom template. This would be for someone who wants to do something more graphical or formatted slightly differently. They would still want to respect the value attribute (probably), but that would be up to them.

specs/progress.md
eljefe223: I think the plan is to create a base component and then different templates. The folder structure will look something like this:

```
progress >
  circular-progress >
    circular-progress.stories.ts
    circular-progress.styles.ts
    circular-progress.template.ts
    index.ts
  radial-progress >
    linear-progress.stories.ts
    linear-progress.styles.ts
    linear-progress.template.ts
    index.ts
  base-progress.ts
  index.ts
```

specs/progress.md
```diff
@@ -0,0 +1,163 @@
...
+- Jim is building a dashboard to track fullfilment for his online store he uses progress ring to visualize various metrics.
+
+- Susan is downloading a large collection of research music onto her harddrive. She sees a linear progress displaying the percentage complete for the currently downloading song and indeterminate linear progress for those songs in queue.
+
+- Fin is scrolling a large list of continuously loading data as he scrolls he sees a indeterminate progress ring to let him know that data is loading.
+
+### Types
+*Progress* can be visually represented in 2 ways.
+- Linear - A straight horizontal line. `fast-progress`
+- Circular - A cicular ring shape. `fast-progress-ring`
```
nicholasrice: are you saying this is a separate HTML element?

specs/progress.md
```diff
@@ -0,0 +1,163 @@
...
+
+*Linear Determinate Template:*
+```
+<div
+    role="progressbar"
+    aria-valuenow={50}
+    aria-valuemin={0}
+    aria-valuemax={100}
+>
+    <div part="determinate" slot={"determinate"}>
```
nicholasrice: minor nit, but you're mixing JSX into this - is that intentional?

specs/progress.md
```diff
@@ -0,0 +1,163 @@
...
+
+*Slots:*
+- `indeterminate-linear`
+- `indeterminate-circular`
+
+### Anatomy and Appearance
+
+*Parts:*
+Linear parts:
+- determinate
```
nicholasrice: The part names seem to be applied more as states than as physical parts of the component. Is there a way we can name these element parts so that we can re-use structure?

specs/progress.md
eljefe223: Yes. We will have progress-base and then swap out templates and styles.

specs/progress.md
eljefe223: Yeah your right... Still getting used to this new "normal" word of css and straight html. ;)

specs/progress.md
nicholasrice: Sorry I'm still confused. Do I, as an app author *using* this library, would write the following HTML?

```html
<!-- when I want a determinate progress -->
<fast-progress value="2" min="0" max="100"></fast-progress>

<!-- and this when I want a loading spinner-->
<fast-progress-ring></fast-progress-ring>
```

specs/progress.md
```diff
@@ -0,0 +1,152 @@
...
+    aria-valuemax=100
+>
+    <svg class="progress" part="progress" viewBox="0 0 16 16" slot="determinate">
+        <circle class="background" part="background" cx="8px" cy="8px" r="7px" />
+        <circle class="determinate" part="determinate" style="stroke-dasharray: 22px 44px" cx="8px" cy="8px" r="7px" />
+    </svg>
+</fast-progress-ring>
+```
+
+*Orogress Ring Indeterminate Template:*
```
nicholasrice:
```suggestion
*Progress Ring Indeterminate Template:*
```

specs/progress.md
```diff
@@ -0,0 +1,152 @@
+# Progress
+
+## Overview
+
+*Progress* and *progress ring* is used to display the length of time a process will take or to visualize percentage value (referred to as a **determinate** state) and to represent an unspecified wait time (referred to as an **indeterminate** state). *Progress* components are typically visually represented by a circular or linear animation. When the `value` attribute is passed the state is **determinate**, otherwise it is **indeterminate**.
```
nicholasrice:
```suggestion
*Progress* and *progress ring* are used to display the length of time a process will take or to visualize percentage value (referred to as a **determinate** state) and to represent an unspecified wait time (referred to as an **indeterminate** state). *Progress* components are typically visually represented by a circular or linear animation. When the `value` attribute is passed the state is **determinate**, otherwise it is **indeterminate**.
```

PR #2747, opened 2020-03-05
refactor(fast-element): improve design of fast element definitions

EisenbergEffect:
# Description

Improve the design of the custom element definition, configuration, and associated APIs.

## Motivation & context

This change clarifies several APIs, improves naming, simplifies a few areas, and reorganizes some code. All the affected code relates to the configuration and definition objects associated with FastElement.

Here is a summary of the changes, mostly based on conversations from my previous PR:

* The `CustomElement.define` API was changed to `FastElement.define` in order to indicate that this is `FastElement`-specific and disambiguate it from the standard `customElements.define` API.
* The `createFastElement` API for creating fast elements derived from elements other than `HTMLElement` is now exposed as `FastElement.from`.
* The `CustomElement.getDefinition` API is now `FastElement.getDefinition`
* `CustomElementDefinition` was renamed to `FastElementDefinition` and is now only a data structure. Its previous functionality, entirely around creating definitions, was moved into the `FastElement.define` method so that various aspects of creation could be optimized (less loops over the same data).
* The `bindable` terminology was removed and the `bindable` decorator was simplified to better handle our uses cases. It is now named `@attr` rather than `@attr` simply being an alias for `@bindable`.
* The `bindables` property of the partial definition object, which was a Record structure, was renamed to `attributes` and is an array of `AttributeConfiguration` objects.
* The `bindables` property of the full definition object, which was a Record structure, was renamed to `attributes` and is an array of `AttributeDefinition` objects. Additionally,  `attributeLookup` and `propertyLookup` properties were added to enable quick lookup of the definition by either the attribute or property name.

## Issue type checklist

- [x] **Refactor**: An improvement to the design of existing code.

Technically, this does change the shape of various APIs and data structures related to `FastElement`. However, they are not the main use APIs and are not currently documented, so no one is likely to be affected by them at this time.

## Process & policy checklist


packages/fast-element/src/interfaces.ts
```diff
@@ -1,7 +1,9 @@
...
-
 export type Callable = typeof Function.prototype.call | { call(): void };
 
 export const emptyArray = Object.freeze([]);
+
+export type CustomElementConstructor = {
+    new (): HTMLElement;
+};
+
+export type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
```
nicholasrice: Are you unable to use the `Omit` language feature?

https://www.typescriptlang.org/docs/handbook/utility-types.html#omittk

EisenbergEffect: I thought it wasn't available in the version of TypeScript we were using. I'll try removing this and see if it compiles.

EisenbergEffect: Seems to build now that I've removed it. Good call!

nicholasrice: Yea the packages manage their versions independently - this and fast-components target latest I believe

PR #2751, opened 2020-03-05
feat: add dialog as a new web component

chrisdholt:
# Description
This PR adds `dialog` as a new web component.

Spec: https://github.com/microsoft/fast-dna/blob/master/specs/dialog/dialog.md

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        document.removeEventListener("keydown", this.handleDocumentKeydown);
+
+        if (this.shouldDialogTrapFocus()) {
+            document.removeEventListener("focusin", this.handleDocumentFocus);
+        }
+    }
+
+    private handleDocumentKeydown = (e: KeyboardEvent): void => {
+        // remove as any here when we have boolean support
```
janechu: Is this comment accurate? Seems like it might have been misplaced.

packages/fast-components/src/dialog/dialog.template.ts
```diff
@@ -0,0 +1,34 @@
+import { html } from "@microsoft/fast-element";
+import { ref, when } from "@microsoft/fast-element/dist/directives/";
```
janechu: Do we need to export these from the root of this package?

packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,145 @@
...
+import tabbable from "tabbable";
+
+export class Dialog extends FastElement {
+    @attr
+    public modal: boolean = true;
+
+    @attr
+    public hidden: boolean = false;
+
+    @attr({ attribute: "trap-focus" })
```
nicholasrice: is focus trapping independent of the dialog's modal value?

packages/fast-components/src/dialog/dialog.template.ts
chrisdholt: Rob was doing some investigation on that. I believe we've updated 'when' (I did that yesterday) so that can come in.

packages/fast-components/src/dialog/dialog.ts
chrisdholt: It use to be accurate. :) I thought I got most of them.

packages/fast-components/src/dialog/dialog.ts
chrisdholt: Right now it is. The thought was to provide a means for people to opt-out, even if that means in modal scenarios. The intent was based on this from the W3C:

>Like non-modal dialogs, modal dialogs contain their tab sequence. That is, Tab and Shift + Tab do not move focus outside the dialog. However, unlike most non-modal dialogs, modal dialogs do not provide means for moving keyboard focus outside the dialog window without closing the dialog.

Currently the coded behavior is that both contain their tab sequence (first two sentences refer to this). If we want to ensure that never happens with a modal, I'm good with that, but this was intentional and meant to somewhat spark the conversation on "how flexible" we want to be.

packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,145 @@
...
+                composed: true,
+            })
+        );
+    }
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+        document.addEventListener("keydown", this.handleDocumentKeydown);
+
+        if (this.shouldDialogTrapFocus()) {
```
nicholasrice: Adding and removing handlers should probably be a reaction to the `trapFocus` property. With this implementation, if `trapFocus` is changed during the component then lifecycle handlers may not be attached when they should be, or may be attached when they aren't needed.

You would still want to clean up in disconnectedCallback if handlers are attached

packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,145 @@
...
+            e.preventDefault();
+        }
+    };
+
+    private handleTabKeyDown = (e: KeyboardEvent): void => {
+        if (!this.shouldDialogTrapFocus()) {
+            return;
+        }
+
+        const tabbableElements: HTMLElement[] = tabbable(this);
```
nicholasrice: Invoking `tabbable` every keyDown is going to end up querying the DOM a lot more than is likely necessary. Perhaps we should store the tabbable element references and then use a mutation observer to update the list when elements are added / removed?

packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,145 @@
...
+        const tabbableElementCount: number = tabbableElements.length;
+
+        if (tabbableElementCount === 0) {
+            this.dialog.focus();
+            e.preventDefault();
+            return;
+        }
+
+        if (e.shiftKey && e.target === tabbableElements[0]) {
```
nicholasrice: Can you add a comment about *why* you're preventing default?

packages/fast-components/src/dialog/dialog.ts
chrisdholt: Yeah I was thinking about that but decided to push this out and get feedback (like this).

packages/fast-components/src/dialog/dialog.ts
chrisdholt: I took a shot at this if. Resolving this only as it's outdated now.

packages/fast-components/src/dialog/dialog.ts
chrisdholt: taken another go at this based on feedback. Resolving this only as it's outdated.

packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,178 @@
...
+
+        this.observer = new MutationObserver(this.onChildListChange);
+        // only observe if nodes are added or removed
+        this.observer.observe(this, { childList: true });
+
+        document.addEventListener("keydown", this.handleDocumentKeydown);
+
+        // Ensure the DOM is updated
+        // This helps avoid a delay with `autofocus` elements recieving focus
+        DOM.queueUpdate(this.trapFocusChanged);
```
chrisdholt: @EisenbergEffect just an FYI that DOM.queueUpdate solved the issue seen with autofocus elements being delayed. Thx!

EisenbergEffect: Groovy. I'm coming up on reviewing this code soon. I'll take a look at the internal details and make sure my guess on why this is, is correct and see if we need to make any changes in the library or not.

packages/fast-components/src/dialog/dialog.ts
```diff
@@ -0,0 +1,178 @@
...
+    public trapFocus: boolean = true;
+
+    @attr({ attribute: "aria-describedby" })
+    public ariaDescribedby: string;
+
+    @attr({ attribute: "aria-labelledby" })
+    public ariaLabelledby: string;
+
+    @attr({ attribute: "aria-label" })
+    public ariaLabel: string;
```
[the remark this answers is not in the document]
chrisdholt: Yeah, per the W3C specification, the label (or labelledby, describedby) needs to be on the element with a role of dialog. That element is nested within the component, so we need to pass it down.

Is this working...I'm trying to find out. :) I'm using voiceover on Mac which is not always reliable.

packages/fast-components/src/dialog/dialog.template.ts
chrisdholt: resolving this for `when` - unsure when/if ref will get exported.

packages/fast-components/src/dialog/dialog.ts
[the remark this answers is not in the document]
chrisdholt: I *think* this is working as expected. I was able to get `aria-label` and `aria-labelledby` both working as they do with a straight HTML instance. Keep in mind, that is with voiceover, but this seems to do what I would expect. Might be good to get someone to test on narrator though.

packages/fast-components/src/dialog/dialog.styles.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const DialogStyles = css`
+    :host([hidden]) {
+        display: none;
+    }
+
+    :host {
+        --elevation: 8;
+        --height: 480px;
+        --width: 640px;
```
chrisdholt: @nicholasrice do you have any concerns about using custom properties here for the dialog?

nicholasrice: I don't really understand what utility they add

nicholasrice: nevermind I see now

packages/fast-components/src/dialog/dialog.template.ts
```diff
@@ -0,0 +1,33 @@
+import { html } from "@microsoft/fast-element";
+import { when } from "@microsoft/fast-element";
+import { ref } from "@microsoft/fast-element/dist/directives/";
+import { Dialog } from "./dialog";
```
nicholasrice: nit: consolidate imports from `@microsoft/fast-element`

packages/fast-components/src/dialog/dialog.ts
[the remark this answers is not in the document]
nicholasrice: I pulled it up on a narrator - it does not seem to interpret `aria-labelledby` correctly. A slight aside - can you add the `aria-labelledby` as a storybook example?

packages/fast-components/src/dialog/dialog.template.ts
chrisdholt: These aren’t all exported from the root - @EisenbergEffect on his thinking there

PR #2753, opened 2020-03-05
chore: add slider and slider-label specs

marjonlynch:
# Description
Adds spec document for slider and slider-label component. This component builds off the form-assocation component spec.

## Process & policy checklist


specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+### Anatomy and Appearance
+**Structure:**
+
+```
+  <div
+    role="slider"
+    $aria-valuenow="${x => x.aria-valuenow}"
+    $aria-valuemin="${x => x.aria-valuemin}"
+    $aria-valuemax="${x => x.aria-valuemax}"
+  >
```
EisenbergEffect: We'll want these aria attributes on the host element, rather than on the root element of the div. As mentioned above, no need to have aria properties as well. We can derive the aria attribute value from the normal property values and replicate that to the host.

specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+
+## Implementation
+
+```
+<fast-slider
+    id="slider1"
+    initialValue={32}
+    ariaminvalue={"0"}
+    ariamaxvalue={"40"}
+    ariavaluenow="32"}
```
EisenbergEffect: See notes above on value and aria attrs.

specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+    min={0}
+    max={44}
+    step={1}
+>
+    <h2 slot="label"><b>Density</b></h2>
+    <fast-slider-label
+      label="32"
+      showTickMark="true"
```
EisenbergEffect: Just a reminder that HTML attributes are all lowercase.

specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+If they use <slider-label> elements those will be styled a specific way internally and the dev can override any part of that style they want.
+
+### API
+
+*Component Name*
+- `slider-label`
+
+*Attributes*
+- `showTickmark` - boolean to show/hide the tick mark. Default is show.
+- `valuePosition` - used to position the label on the track relative to the min and max values
```
EisenbergEffect: Maybe we should simplify these names. `mark` and `position` ?

specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+- `mode`
+The slider component can be set to different modes using the `mode` prop.  The default is `singleValue`.  The modes are as follows:
+ > - `singleValue` -  A single thumb, no selected range is displayed.
+ > - `adjustUpperValue` -  A single thumb associated with the upper value of slider's selected range.  The *slider*'s front bar element displays and is anchored to the low end of the slider's current range as the user adjusts the maximum value.
+ > - `adjustLowerValue` -  A single thumb associated with the lower value of slider's selected range.  The *slider*'s front bar element displays and is anchored to the high end of the slider's current range as the user adjusts the minimum value.
+ > - `adjustBoth` - in which case the selected range starts at 40% to 60% of the total range
+
+ - `initialValue` - The `initialValue` prop allows authors to specify the initial selected range of the `slider`.  It defaults to a (step constrained) value at the midpoint on the `slider`'s total range for all modes except `adjustBoth`.
```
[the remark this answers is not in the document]
scomea: A react implementation artifact IMO.

specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+
+*Events*
+- `onValueChange`
+
+*Slots*
+- `backgroundTrack`
+- `foregroundTrack`
+- `sliderTrack`
+- `lowerThumb`
+- `upperThumb`
```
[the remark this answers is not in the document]
marjonlynch: like foreground-track or foregroundtrack?

specs/slider/slider.md
marjonlynch: You don't think it should be show-mark? I'm fine with it, just mark seemed non-intuitive to me as a boolean. position makes sense to me

specs/slider/slider.md
```diff
@@ -0,0 +1,209 @@
...
+
+ - `initialValue` - The `initialValue` prop allows authors to specify the initial selected range of the `slider`.  It defaults to a (step constrained) value at the midpoint on the `slider`'s total range for all modes except `adjustBoth`.
+
+The aria values by default will be the min, max and value number but can be overridden via these props
+- `aria-valuemin` -
+- `aria-valuemax` - 
+- `aria-valuenow` - 
+
+*Events*
+- `onValueChange`
```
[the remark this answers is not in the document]
marjonlynch: I've updated this, still not sure after our conversation if this is correct or not but this is what I'm thinking. `change` event raised to external components, `valueChanged` for dealing with the value changing internally and to invoke the `change` event

specs/slider/slider.md
```diff
@@ -0,0 +1,201 @@
...
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+A developer can inject labels on the slider as children, if they use <slider-label> elements those will be styled a specific way internally and the dev can override any part of that style they want. Alternatively the dev can put whatever they want in children and style as they wish.
+
+- *A designer customizing the component.*
+A designer can override any internal styling applied to the slotted labels, the background track, the slider track and the thumb(s) if appropriate.
+  
+### Features
+- Implement your own lower and upper thumb UI, default provided.
+- Implement your own background and slider track UI, default provided.
+- Developers can hook up to the `slider`'s `onValueChange` callback to be notified when the value has changed.
```
chrisdholt: This is nitpicky, but for the first two this reads like you are speaking to a developer personally and informally, while the third sentence simply says "Developers". Since I think the target audience here is the same can we align to one?

specs/slider/slider.md
```diff
@@ -0,0 +1,201 @@
...
+
+*Methods*
+- `onValueChanged` - handling the internal framework call when the value attribute is changed.
+
+*Slots*
+- `background-track`
+- `foreground-track`
+- `slider-track`
+- `lower-thumb`
+- `upper-thumb`
```
chrisdholt: how do I change the slot for just a single thumb?

specs/slider/slider.md
```diff
@@ -0,0 +1,201 @@
...
+*Events*
+- `change` - raise the change event for external parties to be informed of the value change.
+
+*Methods*
+- `onValueChanged` - handling the internal framework call when the value attribute is changed.
+
+*Slots*
+- `background-track`
+- `foreground-track`
+- `slider-track`
```
chrisdholt: Can you add details regarding how these are each different?

specs/slider/slider.md
marjonlynch: yep will do

specs/slider/slider.md
marjonlynch: So I think this is a leading question to say it's confusing to know which thumb you would be slotting if you're just using one, and I agree. this is based as you can probably tell on the old slider with the thinking that we want to match the functionality there. In the old component singleValue mode used the lowerThumb. I also see that if you wanted to go with the default scenario you'd end up with fallback content for upper-thumb which is not desirable so I'm updating that now. It might be more intuitive to have the slots be thumb and upper-thumb?

specs/slider/slider.md
marjonlynch: I simplified the interface to handle just the default case. I'm pretty sure it's the only case we were using seeing as the mode enum wasn't exported outside of fast. Added descriptions to what remains

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
+# slider + slider-label
```
janechu: Prefer we stick to titles formatted in uppercase so "Slider + Slider Label" - see tree view spec.

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+
+** see [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range)
+
+### Use Cases
+
+- *A customer using the component on a web page.*
+On a web page the customer drags the slider from min to max values ultimately setting the value of the component to a value constrained by step and min and max values.
+
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+A developer can inject labels on the slider as children, if they use <slider-label> elements those will be styled a specific way internally and the dev can override any part of that style they want. Alternatively the dev can put whatever they want in children and style as they wish.
```
janechu: Can we change "as children" to "into the default slot", I think it might be best to try and stick with web component terminology. Also would change the wording a little, "...and the dev can override..." to just "...and any part of the style can be overridden" and likewise "Alternatively the dev can put..." to "Styles and items intending to be placed in the default slot can be replaced as needed."

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+
+*Component Name*:
+- `fast-slider`
+
+*Props/Attrs*
+- `min` - minimum allowed value for the slider
+- `max` - maximum allowed value for the slider
+- `step` - limits the values of the `slider` to increments of the `step` value added to the minimum value of the 
+`slider`'s total range.  The default value is 1. The minimum and maximum values of a `slider`'s range are always valid results regardless of the `step` prop. The `step` prop is used as the value for incrementing the thumb by pressing the arrow keys.
+
```
janechu: extra line?

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+    <slot></slot>
+  </div>
+```
+
+## Implementation
+
+```
+<fast-slider
+    id="slider1"
+    value={32}
```
janechu: I think you can just use "" instead of {} since this is the implementation.

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+- aria-valuenow will be set as the slider value changes
+- right and up arrow move the slider in increasing value and left and down arrows move it in decreasing increments
+- *Form Input*
+
+
+### Globalization
+
+*Consider whether the component has any special globalization needs such as:*
+
+- *Special RTL handling*
```
janechu: I think these lines can be eliminated

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+    <slot name="thumb">
+      <div></div>
+    </slot>
+    <slot></slot>
+  </div>
+```
+
+## Implementation
+
+```
```
janechu: I think you could put "html" after the backticks to get nicer syntax highlighting.

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+- `label`
+
+### Anatomy and Appearance
+**Structure:**
+
+```
+  <div>
+    ${when(x => x.show-mark, html<slider-label>`
+      <slot name="mark"><div>|</div></slot>
+    ')}
```
janechu: I think this ' is meant to be a backtick

specs/slider/slider.md
```diff
@@ -0,0 +1,189 @@
...
+If they use <slider-label> elements those will be styled a specific way internally and the dev can override any part of that style they want.
+
+### API
+
+*Component Name*
+- `slider-label`
+
+*Attributes*
+- `show-mark` - boolean to show/hide the mark. Default is show.
+- `position` - used to position the label on the track relative to the min and max values
```
scomea: "relative to min and max" doesn't really make sense to me.  Maybe more like "used to position the label at the corresponding value on the track".

specs/slider/slider.md
```diff
@@ -0,0 +1,184 @@
...
+- `min` - minimum allowed value for the slider
+- `max` - maximum allowed value for the slider
+- `step` - limits the values of the `slider` to increments of the `step` value added to the minimum value of the 
+`slider`'s total range.  The default value is 1. The minimum and maximum values of a `slider`'s range are always valid results regardless of the `step` prop. The `step` prop is used as the value for incrementing the thumb by pressing the arrow keys.
+ - `value` - Allows authors to specify the initial selected range of the `slider`.  It defaults to a (step constrained) value at the midpoint on the `slider`'s total range. 
+
+*Events*
+- `change` - raise the change event for external parties to be informed of the value change.
+
+*Methods*
```
nicholasrice: Can you help me understand the `onValueChanged` method? When would this get used instead of the change event?

scomea: Feels like an artifact from supporting React "controlled" modes.

marjonlynch: change event was meant to be for external parties interested in the value changing. valueChanged was me thinking I needed to handle when something like the hidden input control changed and then raising the event for external parties. As it turns out I do have a valueChanged method but it is private so maybe it doesn't need to be documented here.

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- `fast-slider`
+
+*Props/Attrs*
+- `min` - minimum allowed value for the slider
+- `max` - maximum allowed value for the slider
+- `step` - limits the values of the `slider` to increments of the `step` value added to the minimum value of the 
+`slider`'s total range.  The default value is 1. The minimum and maximum values of a `slider`'s range are always valid results regardless of the `step` prop. The `step` prop is used as the value for incrementing the thumb by pressing the arrow keys.
+ - `value` - Allows authors to specify the initial selected range of the `slider`.  It defaults to a (step constrained) value at the midpoint on the `slider`'s total range. 
+
+*Events*
```
scomea: No orientation option for vertical sliders?

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- `background-track` - background area encompassing the track and labels
+- `slider-track` - the horizontal or vertical track along which the thumb slides
+- `thumb` - the control the user drags along the track to change the value of the slider
+
+### Anatomy and Appearance
+**Structure:**
+
+```html
+  <div
+    role="slider"
```
EisenbergEffect: Note: We'll want the `role` and appropriate aria attributes to appear on the host element, `fast-slider` itself. We've got a new feature in `fast-element` to make this easy. Not an issue for this spec per se. We can handle it when we get into the implementation itself.

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+
+*Parts*
+- `label`
+
+### Anatomy and Appearance
+**Structure:**
+
+```html
+  <div>
+    ${when(x => x.show-mark, html<slider-label>`
```
EisenbergEffect: Should probably be `x => x.showMark`

specs/slider/slider.md
marjonlynch: Oops yea I planned to put that in, will add that. No mode for now though.

specs/slider/slider.md
marjonlynch: ok, will update here when I find out the new way to do this

specs/slider/slider.md
marjonlynch: yes, I just found that out in the actual component. will change here as well

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
+# Slider + Slider-Label
+
+## Overview
+
+The `slider` component allows the user to move one or two `thumb` elements along a vertical or horizontal axis to select a value or range. `slider` is also considered an input field and must work within a `<form>` as other input fields do.
+
+### Background
+
+Just styling `<input type="range"/>` and using `<datalist>` for the tickmarks was considered but the standards aren't supported enough to allow for the level customization we want to provide.
```
chrisdholt: Just for clarity, is this that standards are not supported by each browser, or the current standard elements don't support the level of customization we want?

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+### Features
+- Implement your own lower and upper thumb UI, default provided
+- Implement your own background and slider track UI, default provided
+- Handle the `slider`'s `onValueChange` callback to be notified when the value has changed.
+- Customize the tick marks along the track
+- Include slider in form based operations
+
+### Prior Art/Examples
+- [FAST-DNA (React)](https://explore.fast.design/components/slider)
+- [Material UI](https://material-ui.com/components/slider/)
```
chrisdholt: Are these the only two examples that we can hunt down? Would be ideal to have a few more if we can find them.

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+  <div
+    role="slider"
+  >
+    <input type="range" 
+      id="${x => x.id}"
+      min="${x => x.min}"
+      max="${x => x.max}"
+      step="${x => x.step}"
+      value="${x => x.value}"
+    >
```
chrisdholt: Do you plan on using form association here? The input being rendered here is throwing me off. I think you'd want the input here if you planned on delegating focus to it; however, since we can't style the element well enough, I think that's an unlikely scenario. I think we'll want this to be form associated instead. You can see an example of the differences by comparing the checkbox web component (form association) to the text field on I have in [review](https://github.com/microsoft/fast-dna/pull/2785) (focus delegation).

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- aria-valuenow will be set as the slider value changes
+- right and up arrow move the slider in increasing value and left and down arrows move it in decreasing increments
+- *Form Input*
+
+
+### Globalization
+Slider should render a mirror view in rtl for example:
+![](./images/slider-rtl.png) 
+
+- Strings passed into the component should already be localized
```
chrisdholt: Could you detail the expectation for how the upper/lower thumbs respond in terms of positioning and keyboarding?

specs/slider/slider.md
chrisdholt: @EisenbergEffect or @nicholasrice can correct me if I'm wrong here, btw :)

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+A developer can inject labels on the slider into the default slot, if they use <slider-label> elements those will be styled a specific way internally and any part of that style can be overridden. Styles and items intending to be placed in the default slot can be replaced as needed.
+
+- *A designer customizing the component.*
+A designer can override any internal styling applied to the slotted labels, the background track, the slider track and the thumb(s) if appropriate.
+  
+### Features
+- Implement your own lower and upper thumb UI, default provided
+- Implement your own background and slider track UI, default provided
+- Handle the `slider`'s `onValueChange` callback to be notified when the value has changed.
```
nicholasrice: What is the reason for exposing an `onValueChange` *callback* instead of firing a `change` event? For other input elements we will be simply firing a `change` event - is this intentionally different?

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- `max` - maximum allowed value for the slider
+- `step` - limits the values of the `slider` to increments of the `step` value added to the minimum value of the 
+`slider`'s total range.  The default value is 1. The minimum and maximum values of a `slider`'s range are always valid results regardless of the `step` prop. The `step` prop is used as the value for incrementing the thumb by pressing the arrow keys.
+ - `value` - Allows authors to specify the initial selected range of the `slider`.  It defaults to a (step constrained) value at the midpoint on the `slider`'s total range. 
+
+*Events*
+- `change` - raise the change event for external parties to be informed of the value change.
+
+*Slots*
+- `background-track` - background area encompassing the track and labels
```
nicholasrice: Is this "background-track" actually necessary, or can the host element serve this purpose?

specs/slider/slider.md
nicholasrice: Yes, this would use form-association.

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- `slider-track` - the horizontal or vertical track along which the thumb slides
+- `thumb` - the control the user drags along the track to change the value of the slider
```
nicholasrice: How are is "slider-label" provided to this component? Is there also a "default" slot?

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- *A designer customizing the component.*
+If they use <slider-label> elements those will be styled a specific way internally and the dev can override any part of that style they want.
+
+### API
+
+*Component Name*
+- `slider-label`
+
+*Attributes*
+- `show-mark` - boolean to show/hide the mark. Default is show.
```
nicholasrice: If the default is to "show" we might consider changing this to "hide-mark". "show-mark" implies that, when that attribute is on the element, it should show the mark. That conflicts with the desired behavior of being able to write `<slider-label></slider-label>` (no attribute) and having the mark be visible. We *could* resolve this by automatically writing the attribute to the host if it doesn't exist, but that requires more JavaScript and element attributes.

specs/slider/slider.md
marjonlynch: Removed that from template example.

specs/slider/slider.md
marjonlynch: Ok, keeping this out for the time being until I discuss with Chris and others how much functionality we want from the start here. Aware of the issue.

specs/slider/slider.md
marjonlynch: Yea, I've been building the component at the same time and it's using Form association and working. So I removed the input from the template in this spec and the actual code.

specs/slider/slider.md
marjonlynch: I only have one "track" slot now, I left it called background-track but maybe it should be track.

specs/slider/slider.md
marjonlynch: yes, added that to documentation

specs/slider/slider.md
marjonlynch: that makes sense, I'll change it to hide-mark

specs/slider/slider.md
marjonlynch: changed this to be just change

specs/slider/slider.md
marjonlynch: each browser has different level of support, no browser seems to have full support for things like the labels. I have no idea how far off they are to implementing this as suggested/documented.

specs/slider/slider.md
marjonlynch: ok will look for a couple more

specs/slider/slider.md
marjonlynch: changed to hideMark

specs/slider/slider.md
marjonlynch: changed this to just refer to thumb. We can maybe talk offline but thinking this component should go in as simple as possible and can be expanded upon in subsequent versions.

specs/slider/slider.md
marjonlynch: added Atlassian and Lightning Design System. Which both do not incorporate upper and lower thumbs or vertical/horizontal orientation

specs/slider/slider.md
```diff
@@ -0,0 +1,192 @@
...
+
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+A developer can inject labels on the slider into the default slot, if they use <slider-label> elements those will be styled a specific way internally and any part of that style can be overridden. Styles and items intending to be placed in the default slot can be replaced as needed.
+
+- *A designer customizing the component.*
+A designer can override any internal styling applied to the slotted labels, the background track, the slider track and the thumb(s) if appropriate.
+  
+### Features
+- Implement your own thumb UI, default provided
+- Implement your own background and slider track UI, default provided
```
chrisdholt: Is there a reason that the background or track is not just the :host element?

specs/slider/slider.md
chrisdholt: Unresolving this - can we add this detail or a bit more here? There is a standard progress element but it doesn't support the full level of detail, styling, etc that most design systems need or require.

chrisdholt: For context: https://caniuse.com/#feat=input-range

specs/slider/slider.md
chrisdholt: Did this get deleted all up? If we're planning on including upper and lower values via thumbs I'd like that represented in the spec. We can still implement that as "next steps" and not include in the initial version, but I think it's important to capture and understand the full functionality so we don't paint ourselves into a corner with an API.

chrisdholt: Examples of this can be seen in something like tree view where we detail that we are *not* initially implementing certain functionality.

specs/slider/slider.md
marjonlynch: I've renamed this to be just track and with dual mode we will have something else like foregroundTrack. Track is the actual line we draw horizontally or vertically and is swappable using a slot. I mislead with the backgroundTrack and track before but simplified since then. So yes they host would be the thing to style the background (not the track line) now

marjonlynch: but i haven't updated the spec yet, will do that shortly

specs/slider/slider.md
marjonlynch: Yes, I will be adding dual thumb back in per our conversation

specs/slider/slider.md
```diff
@@ -0,0 +1,181 @@
...
+- [Atlassian UI](https://atlaskit.atlassian.com/packages/core/range)
+- [Lightning Design System](https://www.lightningdesignsystem.com/components/slider/)
+---
+
+### API
+
+*Component Name*:
+- `fast-slider`
+
+*Props/Attrs*
```
eljefe223: do we want to use the language props?

marjonlynch: we do not, fixed

specs/slider/slider.md
marjonlynch: I think I resolved this with my updated text (using what you said mostly). matching roughly what is mentioned on atlassian and lightning design slider documentation.

specs/slider/slider.md
marjonlynch: I added back in modes but changed the names slightly. I also added an * to state we should only implement single-value to start.

PR #2755, opened 2020-03-06
perf(fast-element): simplify expressions

EisenbergEffect:
# Description

This PR removes the set of interfaces and classes that were meant to model expressions in templates and replaces them with a simple function.

## Motivation & context

When the early work on fast-element began, I experimented with a couple of different approaches to templating. As part of that work, an abstraction around expressions was built with multiple implementations. Now that the final design is stabilizing, these abstractions are no longer needed and could be replaced with a simple function. The end result is less abstraction, less polymorphism, less code, and a little bit of a perf improvement.

## Issue type checklist

- [x] **Performance**: Improved the performance or memory consumption.
- [x] **Refactor**: Improved the design of existing code.

Technically this is a breaking change, but it's unlikely to affect anyone since the way that templates are written did not change, just some internal representation of various details.

## Process & policy checklist


packages/fast-element/src/index.ts
```diff
@@ -13,3 +12,4 @@ export * from "./observation/observable";
 export * from "./behaviors/index";
 export * from "./directives/directive";
 export * from "./directives/bind";
+export * from "./dom";
```
EisenbergEffect: After talking with @chrisdholt I realized we weren't exporting the DOM helpers. It's not really related to the main purpose of this PR. I just sort of squeezed this change in since it was so small.

PR #2758, opened 2020-03-06
chore: update name-tag example to include hidden host styling

chrisdholt:
# Description
This PR updates the name-tag example component in `fast-components` to include hidden styling for the :host selector.

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/name-tag/name-tag.styles.ts
```diff
@@ -1,6 +1,10 @@
 import { css } from "@microsoft/fast-element";
 
 export const NameTagStyles = css`
```
nicholasrice: Looks like you have another PR to simply remove this component - should we just close this one?

chrisdholt: We can - I was simply removing the export in the other; should we drop the example all up?

nicholasrice: Oh sorry - I thought it removed everything. That's fine - we can remove the actual component when we get a few components through.

PR #2759, opened 2020-03-06
chore: update fast docs to remove `on` from attr and observable changed method

chrisdholt:
# Description
This PR updates the method names in fast-element docs so there is a clearly separation of *propertyName*Changed methods vs other methods.

## Issue type checklist

## Process & policy checklist


packages/fast-element/docs/building-components.md
```diff
@@ -362,7 +362,7 @@ export class FriendList extends FastElement {
     this.name = '';
   }
 
-  onNameChanged(event: Event) {
+  nameChanged(event: Event) {
```
EisenbergEffect: Because `name` is observable, I believe this will create an infinite loop when the name is changed through the input.

* input change => invoke nameChange
* nameChange => update name property
* name property => invoke NameChange
* nameChange => update name property
* ...

Perhaps it's better to rename it to something like `handleNameInput`?

chrisdholt: sure - so in terms of observable, we wouldn't want to leverage *propertyName*Changed, but in terms of @attr it's 👍, correct?

EisenbergEffect: Basically, both attr and observable support the *propertyName*Changed method pattern. However, in this example, the input element is directly calling a change method when it receives input, so that it can update the property. Basically, there's no two-way binding, so to keep the property in sync with the input, we need to wire for the input event and then use that to update our own property. But if we name the method we're using to handle the input event the same as the *propertyName*Changed method pattern, then the observable/attr will also call that when the property changes.

chrisdholt: Sounds good - I've updated with your suggestion. I think it'll help clear up this confusion a bit...even if just for me.

EisenbergEffect: Agree. Thanks for the improvement!

PR #2760, opened 2020-03-06
feat(fast-element): enable host instructions

EisenbergEffect:
# Description

Enable a template to declare attributes and bindings that should be placed on the host HTML element of a custom element.

## Motivation & context

Previously, all dynamic template expressions had to be on HTML within the ShadowDOM of a component. With this update, the component developer can now specify attributes that should be propagated to the host HTML element itself. This is done by wrapping the template in a `template` tag and placing any attributes you want to appear on the host, on the `template` tag itself. The template compiler will parse those and when the element's view is created, it will attach them to the host element.

Here's an example template:

```TypeScript
const template = html<FastProgress>`
  <template role="progressbar"
      $aria-valuenow={x => x.value}
      $aria-valuemin={x => x.min}
      $aria-valuemax={x => x.max}>
    ..rest of template targeted at shadow dom goes here...
  </template>
`;
```

When this renders, the template system will write `role`, `aria-valuenow`, `aria-valuemin`, and `aria-valuemax` on the `<fast-progress>` element itself and it will keep the aria attributes up to date as the value, min, and max properties change on the component. So, you would see something like this in the DOM:

```HTML
<fast-progress min="0"
    max="100"
    value="50"
    role="progressbar"
    aria-valuenow="50" 
    aria-valuemin="0" 
    aria-valuemax="100">
</fast-progress>
```

> **NOTE:** A root template tag is not required for views. You only need one if you intend to provide host instructions.

## Issue type checklist

No

## Process & policy checklist

I'll be adding a section to the documentation shortly as an additional commit to this PR.


janechu:
Very nice!

PR #2762, opened 2020-03-06
feat: add checkbox web component

nicholasrice:
# Description
Adds the fast-checkbox web component. The spec is still out in code-review awaiting sign-off - I will update this PR accordingly.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/checkbox/checkbox.styles.ts
+++ b/packages/fast-components/src/checkbox/checkbox.styles.ts
@@ -27,7 +27,7 @@ export const CheckboxStyles = css`
     .label {
         font-family: var(--body-font);
         color: var(--neutral-foreground-rest);
-        ${/* Need to discuss with Brian how HorizontalSpacingNumber can work */ ""} padding-inline-start: calc(var(--design-unit) * 2px + 2px);
+        ${/* Need to discuss with Brian how HorizontalSpacingNumber can work. https://github.com/microsoft/fast-dna/issues/2766 */ ""} padding-inline-start: calc(var(--design-unit) * 2px + 2px);
         margin-inline-end: calc(var(--design-unit) * 2px + 2px);
         cursor: pointer;
         ${/* Font size is temporary - replace when adaptive typography is figured out */ ""} font-size: calc(1rem + (var(--density) * 2px));
--- a/packages/fast-components/src/checkbox/checkbox.styles.ts
+++ b/packages/fast-components/src/checkbox/checkbox.styles.ts
@@ -1,6 +1,84 @@
 import { css } from "@microsoft/fast-element";
+import { disabled, display } from "../styles";
 
 export const CheckboxStyles = css`
-    :host {
+    ${display("inline-flex")} :host {
+        align-items: center;
+        outline: none;
+        margin: calc(var(--design-unit) * 1px) 0;
+        ${/*
+           * Chromium likes to select label text or the default slot when
+           * the checkbox is clicked. Maybe there is a better solution here?
+           */ ""} user-select: none;
+    }
+
+    .checkbox {
+        position: relative;
+        width: calc((var(--height-number) / 2 + var(--design-unit)) * 1px);
+        height: calc((var(--height-number) / 2 + var(--design-unit)) * 1px);
+        box-sizing: border-box;
+        border-radius: calc(var(--corner-radius) * 1px);
+        border: calc(var(--outline-width) * 1px) solid var(--neutral-outline-rest);
+        background: var(--neutral-fill-input-rest);
+        outline: none;
+        cursor: pointer;
+    }
+
+    :host(.disabled) .checkbox,
+    :host(.readonly) .checkbox {
+        cursor: not-allowed;
+    }
+
+    .checkbox:hover {
+        background: var(--neutral-fill-input-hover);
+        border-color: var(--neutral-outline-hover);
+    }
+
+    .checkbox:focus {
+        box-shadow: 0 0 0 1px var(--neutral-focus) inset;
+        border-color: var(--neutral-focus);
+    }
+
+    .label {
+        font-family: var(--body-font);
+        color: var(--neutral-foreground-rest);
+        ${/* Need to discuss with Brian how HorizontalSpacingNumber can work */ ""} padding-inline-start: calc(var(--design-unit) * 2px + 2px);
+        margin-inline-end: calc(var(--design-unit) * 2px + 2px);
+        cursor: pointer;
+        ${/* Font size is temporary - replace when adaptive typography is figured out */ ""} font-size: calc(1rem + (var(--density) * 2px));
+    }
+
+    :host(.disabled) .label,
+    :host(.readonly) .label {
+        cursor: not-allowed;
+    }
+
+    .checked-indicator {
+        width: 100%;
+        height: 100%;
+        display: block;
+        fill: var(--neutral-foreground-rest);
+        opacity: 0;
+        pointer-events: none;
+    }
+
+    .indeterminate-indicator {
+        border-radius: calc(var(--corner-radius) * 1px);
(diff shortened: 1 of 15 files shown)


packages/fast-element/src/index.ts
```diff
@@ -1,13 +1,18 @@
-export * from "./template";
-export * from "./fast-element";
 export * from "./attributes";
+export * from "./behaviors/index";
 export * from "./controller";
-export * from "./interfaces";
```
nicholasrice: This was part of a merge conflict - doesn't hurt to alphabetize though I guess

packages/fast-components/src/checkbox/checkbox.template.ts
```diff
@@ -0,0 +1,54 @@
+import { html, when } from "@microsoft/fast-element";
+import { Checkbox } from "./checkbox";
+import { bool } from "../utilities";
+
+export const CheckboxTemplate = html<Checkbox>`
+    <div
+        part="checkbox"
+        role="checkbox"
+        class="checkbox"
+        $aria-checked="${x => bool(x.checked)}"
```
EisenbergEffect: I think we should put these aria attributes on the host element. We can circle back and do that after my PR for host instructions lands.

packages/fast-components/src/checkbox/checkbox.template.ts
```diff
@@ -0,0 +1,54 @@
...
+import { bool } from "../utilities";
+
+export const CheckboxTemplate = html<Checkbox>`
+    <div
+        part="checkbox"
+        role="checkbox"
+        class="checkbox"
+        $aria-checked="${x => bool(x.checked)}"
+        $aria-required="${x => bool(x.required)}"
+        $aria-disabled="${x => bool(x.disabled)}"
```
chrisdholt: Should these use the latest work Rob built to pass to the host?

packages/fast-components/src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,141 @@
...
+
+/* tslint:disable:member-ordering */
+export class Checkbox extends FormAssociated<HTMLInputElement> {
+    @attr({ attribute: "readonly" })
+    public readOnly: boolean; // Map to proxy element
+    private readOnlyChanged(): void {
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.readOnly = this.readOnly;
+        }
+    }
```
chrisdholt: I like how you've set this up, but it obviously requires disabling member ordering. Should we create an issue to disable that all-up in our tslint or in this package? I think this convention is super intuitive and easy to read - I'd like us to adopt it.

packages/fast-components/src/styles/display.ts
```diff
@@ -0,0 +1,56 @@
...
+    | "table-footer-group"
+    | "table-row-group"
+    | "table-cell"
+    | "table-column"
+    | "table-row"
+    | "none"
+    | "initial"
+    | "inherit";
+
+const hidden = css`
```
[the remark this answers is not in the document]
janechu: I like that, I think we should have some kind of documented write up though so contributors can know why we decide to use different methods for creating css.

packages/fast-components/src/styles/display.ts
```diff
@@ -0,0 +1,56 @@
...
+    | "table-column-group"
+    | "table-header-group"
+    | "table-footer-group"
+    | "table-row-group"
+    | "table-cell"
+    | "table-column"
+    | "table-row"
+    | "none"
+    | "initial"
+    | "inherit";
```
janechu: This could be alphabetized, also I think additional valid values are:
- flow-root
- unset

packages/fast-components/src/checkbox/checkbox.ts
janechu: I agree, I think we could start adding some style conventions to our contributor documentation because we definitely will get contributions that are all over the place.

packages/fast-components/src/checkbox/checkbox.styles.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    .checkbox:focus {
+        box-shadow: 0 0 0 1px var(--neutral-focus) inset;
+        border-color: var(--neutral-focus);
+    }
+
+    .label {
+        font-family: var(--body-font);
+        color: var(--neutral-foreground-rest);
+        ${/* Need to discuss with Brian how HorizontalSpacingNumber can work */ ""} padding-inline-start: calc(var(--design-unit) * 2px + 2px);
```
janechu: Should this be a link to an issue where the discussion can happen?

packages/fast-components/src/styles/display.ts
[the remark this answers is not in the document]
nicholasrice: My thought here was that *every* component should have this, so associating this sheet with every component rather than every component enumerating the same CSS would be more efficient. If that isn't the case I'm happy to just make it a string - just thought the re-use of the adopted stylesheet would be beneficial

packages/fast-components/src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,141 @@
...
+        this.constructed = true;
+    }
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+
+        this.updateForm();
+
+        this.addEventListener("keypress", this.keypressHandler);
+        this.addEventListener("click", this.clickHandler);
```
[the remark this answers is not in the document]
nicholasrice: Oops good call - meant to do that

packages/fast-components/src/checkbox/checkbox.template.ts
nicholasrice: I'm a little surprised narrator read the right thing when I tried this because the label isn't associated with the host element - but it seems to read any child content so I'm good with this. I think the full scope of this change will be not deferring focus on this control, making the host the focus target and boosting relevant attributes to the host - does that sound right. We also probably don't need to add "checked" or "disabled" classes, but we still could if that is a good utility.

packages/fast-components/src/checkbox/checkbox.template.ts
nicholasrice: I don't think he has added the functionality for that, but I've hoisted these up to the host with imperative APIs

packages/fast-components/src/styles/display.ts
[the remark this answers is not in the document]
nicholasrice: Changed to be a string type

packages/fast-components/src/checkbox/checkbox.template.ts
nicholasrice: hoisted to the parent. Left class-names in place so that styling overrides capabilities are separated from component internal implementation.

packages/fast-components/src/checkbox/checkbox.ts
nicholasrice: I'll put up a PR.

packages/fast-components/src/checkbox/checkbox.ts
```diff
@@ -0,0 +1,167 @@
...
+    public readOnly: boolean; // Map to proxy element
+    private readOnlyChanged(): void {
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.readOnly = this.readOnly;
+        }
+
+        this.readOnly
+            ? this.classList.add("readonly")
+            : this.classList.remove("readonly");
+        this.setAttribute("aria-readonly", bool(this.readOnly).toString());
```
EisenbergEffect: As a follow-up we can use the new host directives feature to bind these attributes on the host element itself, which should remove all the manual aria attr code. Not required to change it now, but a nice incremental improvement.

nicholasrice: Implemented

PR #2764, opened 2020-03-06
chore: remove name tag as export

chrisdholt:
# Description
I noticed that we were exporting the name tag example, which I don't think we want to actually ship. Considering our goal with fast-components is to export "all the things" that help us compose the components, I thought we should export the form-associated abstract class.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/index.ts
+++ b/packages/fast-components/src/index.ts
@@ -1,2 +1,2 @@
-export * from "./name-tag";
 export * from "./design-system-provider";
+export * from "./form-associated";
--- a/packages/fast-components/src/index.ts
+++ b/packages/fast-components/src/index.ts
@@ -1,2 +1,1 @@
 export * from "./design-system-provider";
-export * from "./form-associated";


janechu:
@nicholasrice @chrisdholt should we be exporting the form-associated logic, or was there some reason it was not being exported?

chrisdholt:
> @nicholasrice @chrisdholt should we be exporting the form-associated logic, or was there some reason it was not being exported?

I think we should export - I mentioned this in the description, but I believe the goal is to enable folks to compose their own stuff as well. I don't see harm in exporting this, but if there was reason to not do so I can close this.

nicholasrice:
It is likely to change pretty significantly and maybe even go away so I wanted to keep it internal until our API stabilized.

packages/fast-components/src/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from "./name-tag";
 export * from "./design-system-provider";
+export * from "./form-associated";
```
nicholasrice: I don't think we should export this - at least yet - for the reasons I outlined previously.

chrisdholt: Pulled this export based on our convo

PR #2768, opened 2020-03-09
chore: add anchored region spec

scomea:
# Description
Adds a spec for an "anchored region" web component.

## Motivation & context
Needed base component for others (select, flyout, etc...)

## Issue type checklist

## Process & policy checklist


specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+
+For a more in-depth understanding of how this component works under the covers please refer to the [intersection observer api](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API). 
+
+### Use Cases
+
+It is envisioned that this component would be used as a building block for other components in this library (select, flyout, tooltip, etc) as well as being available for standalone use in responsive layouts.
+
+### Features
+
+- **Relative positioning:** Users can use to it position an element relative to another another element directly, like enabling a menu to open above or below a trigger button.
```
nicholasrice: nit: "use it to..." rather than "use to it"

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+
+A region that always renders above the anchor element.
+```
+<div id="viewport">
+    ...stuff...
+    <button id="anchor">
+        Button is an anchor
+    </button>
+    <anchored-region
+        anchorid="anchor"
```
nicholasrice: many attributes expect an ID value without explicitly having "id" in the name - `for="id"`, `aria-labelledby="id"`, `aria-describedby="id"`, etc. Should we just use "anchor" instead of "anchorid"?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+
+The dimensions of the anchored region will match the dimensions of the content unless scaling is enabled on a particular axis (verticalscalingenabled & horizontalscalingenabled) in which case it will fill all available space between the anchor and viewport.
+
+### API
+
+*Component name:*
+- `fast-anchored-region`
+
+*Attributes:*
+- anchorId - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
```
nicholasrice: see note above on naming for "anchorId"

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+
+- viewportId - The html id of the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element.  If unset the immediate parent element of the positioner is used.
```
nicholasrice: same comment about "viewportId" - maybe just "viewport"?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+- default slot for content
+
+*Events:*
+- positionchanged - event is thrown whenever the placement of the region relative to the anchor changes (top/bottom and left/right). 
+
+
+### Anatomy and Appearance
+**Structure:**
+```
+<div class="root">
```
nicholasrice: You probably don't need the "root" element - the root is just the `anchor-region` element

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+*Component name:*
+- `fast-anchored-region`
+
+*Attributes:*
+- anchorId - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
+
+- viewportId - The html id of the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element.  If unset the immediate parent element of the positioner is used.
+
+- horizontalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
+- horizontalDefaultPosition - Can be 'left', 'right' or 'unset'.  Default is 'unset'
```
nicholasrice: do we *also* want to support "start" and "end" which adjust depending on document-direction, similarly to the `*-inline-start', `*-inline-end`, and flex properties?

specs/anchored-region/anchored-region.md
nicholasrice: can the viewport context be the window?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+
+*Component name:*
+- `fast-anchored-region`
+
+*Attributes:*
+- anchorId - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
+
+- viewportId - The html id of the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element.  If unset the immediate parent element of the positioner is used.
+
+- horizontalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
```
nicholasrice: Do you describe what the values 'uncontrolled', 'locktodefault', 'dynamic' and 'onetime' do? It's not super clear to me what effect these will have.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+- verticalScalingEnabled - The region is sized from code to match available space, in other scenarios the region gets sized via content size.
+
+- horizontalPosition - read only, the current horizontal position of the component. Possible values are 'left', 'right' or 'unset'.
+- verticalPosition - read only, the current vertical position of the component. Possible values are 'top', 'bottom' or 'unset'.
+
+*Slots:*
+- default slot for content
+
+*Events:*
+- positionchanged - event is thrown whenever the placement of the region relative to the anchor changes (top/bottom and left/right).
```
nicholasrice: minor nit - HTML convention uses present tense event names (unless there are edge cases I don't know about?) - should we align here? "positionchange" vs "positionchang*ed*"?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,254 @@
...
+
+- horizontalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
+- horizontalDefaultPosition - Can be 'left', 'right' or 'unset'.  Default is 'unset'
+- horizontalInset - Boolean that indicates whether the region should overlap the anchor on the horizontal axis. Default is false which places the region adjacent to the anchor element.
+- horizontalThreshold - Numeric value that defines how small in pixels the region must be to the edge of the viewport to switch to the opposite side of the anchor. The component favors the default position until this value is crossed.  When there is not enough space on either side or the value is unset the side with the most space is chosen.
+- horizontalScalingEnabled - The region is sized from code to match available space, in other scenarios the region gets sized via content size.
+
+- verticalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
+- verticalDefaultPosition - Can be 'top', 'bottom' or 'unset'. Default is unset.
```
nicholasrice: A pattern we see in some HTML components (notably checkbox and radio) is that a property is used simply to set the initial value. For example, `<input type="checkbox" checked />` only means the *initial* value is checked. Should we adopt a similar model here? `<fast-anchor-region vertical-position="top"` simply sets the default?

Also, might be good to add a short description for this; I'm inferring what this is doing and could be mistaken

specs/anchored-region/anchored-region.md
scomea: I don't think so.  Authors should be able to apply custom css the scaling region and/or its internals to achieve whatever layout logic they want to the region contents.

specs/anchored-region/anchored-region.md
scomea: I don't believe window is a valid viewport for the purposes of the intersectionObserver api, so document is the upper limit.

specs/anchored-region/anchored-region.md
nicholasrice: I mean to control the default position, not the internals. The case is when I want the `horizontalDefaultPosition` to be "left" in LTR but "right" in RTL

specs/anchored-region/anchored-region.md
nicholasrice: Let me rephrase - can the component position itself relative to an element so that the position is always on the screen? (think native select behavior, where it will draw upwards if there is not enough *screen* real-estate to paint below).

And as an aside - if this supports using the document as the viewport, how do I instruct this component to *use* the document as the viewport?

specs/anchored-region/anchored-region.md
scomea: Sure.

specs/anchored-region/anchored-region.md
scomea: Oh, sorry, yeah I think we could do that.  I will add it.  Do we have somewhere to put component shared resources?  I think we'll want to put a getCurrentDirection() like thing there.

specs/anchored-region/anchored-region.md
scomea: I had not thought of the document id issue.  The previous version of the component used the document as the default viewport, but I am thinking using the immediate parent as the viewport was more "out of the box" usable as that almost certainly a valid viewport but in complex pages the document frequently isn't.  Let me think on it and propose something.

scomea: As for the positioning question, yes you can do that provided that the document root is a valid viewport - no intervening "fixed" elements or extra scrolling surfaces between the anchor/anchored region and the viewport.  The scrolling case is fungible though as a the initial positioning should work and many of these flyout constructs could be closed on scroll.

scomea: I'm debating whether or not we should have a Boolean "always on screen" prop as we did in the previous version which would keep the region in the viewport even as the anchor is resized/scrolled out as I think developers could do that with the scaling behavior and a css grid.  Want to see how easy that is though.

specs/anchored-region/anchored-region.md
scomea: sure

specs/anchored-region/anchored-region.md
scomea: Adding this to the "Design" section:

The component allows users to set a "Positioning Mode" on each axis which defines how the component will behave:
- 'uncontrolled':  The anchored region will appear as it normally would in document flow.
- 'locktodefault': The anchored region will always be placed in the specified default position regardless of available space.  For example a menu that always opens upwards.
- 'dynamic': The anchored region is placed relative to the anchor element based on how much space is available between it and the viewport.  For example a menu that opens opens upwards when it is near the bottom of the page, and downwards when near the top.
- 'onetime': As the 'dynamic' option, but position is not changed after initial placement.

specs/anchored-region/anchored-region.md
scomea: I'm thinking there would be read only props for "horizontalPosition" and "verticalPosition" that would allow access to the current state of the component.  In that context I think it's clearer to have this be labelled as a default.  Note that the default value is used to decide position through the lifetime of the component.

specs/anchored-region/anchored-region.md
scomea: So thinking that the viewport attribute description changes to 
"viewport - Used to identify the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element. Possible values are the id of the desired viewport element, '#document' or '#parent'.  The default value is '#parent'."

scomea: Does "#" work as a prefix do differentiate between string id's and presets?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
+# Anchored region
+
+## Overview
+
+An *anchored region* is a container component which enables authors to create layouts where the contents of the anchored region can be positioned relative to another "anchor" element.  Additionally, the *anchored region* can react to the available space between the anchor and a parent ["viewport"](https://developer.mozilla.org/en-US/docs/Glossary/viewport) element such that the region is placed on the side of the anchor with the most available space, or even resize itself based on that space.
+
+### Background
+
+This component is inspired by the ["Viewport positioner"](https://github.com/microsoft/fast-dna/issues/2597)  component in the React component set.  It is used as a building block in other components to enable responsive flyouts, or positionable/scaling menus in the [select](https://explore.fast.design/components/select) component.
```
janechu: Your viewport positioner link goes to an issue marked as tree view item.

janechu: I would not link to select in the explore site as that link may break in the future as we prepare to launch the web components, instead could you refer to selects and context menus as a general group of components?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+It is envisioned that this component would be used as a building block for other components in this library (select, flyout, tooltip, etc) as well as being available for standalone use in responsive layouts.
+
+### Features
+
+- **Relative positioning:** Users can use it to position an element relative to another another element directly, like enabling a menu to open above or below a trigger button.
+
+- **Responsive positioning:** Users can use it to position an element relative to another element based on available space, for example a menu could open upwards if the trigger button is near the bottom of the page, and downwards if it is nearer the top.
+
+- **Responsive scaling:** Users can use it to create a layout region that dynamically sizes depending on space between the anchor and the viewport elements.
```
janechu: I fear we are putting too many features into this component from the get-go. Can you identify the most basic features needed for this and place any potential future asks in under "Next steps"? See the spec template for details.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+Note that the "always in view" functionality that existed in Viewport Positioner has been removed as this can be replicated through other means (such as a scaling region with grids, for example).
+
+### Risks and Challenges
+
+This component depends on getting accurate positioning data from the dom outside of the component and it is possible for authors to break the relationship in advanced scenarios.
```
janechu: DOM should be in all caps as it is an acronym

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+For example can likely only guarantee positioning if there are no other scrolling containers between the viewport and the anchor/anchored-region.  Authors should avoid nested scrolling containers.
```
janechu: This might read a little better if you re-worded to: "Positioning may only be garaunteed if there are no other..."

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+Also, some css settings can interfere with the data returned by intersection observer and break the positioning functionality.  For example an element with position="fixed" in the dom hierarchy would be a no-no:
```
janechu: *DOM

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+```
+<div id="viewport">
+    <div style="{{position:fixed}}">
+        <button id="anchor">
+            Button is an anchor
+        </button>
+        <anchored-region
```
janechu: All components should be prefixed with `fast-`.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+A region that always renders above the anchor element.
+```
+<div id="viewport">
+    ...stuff...
+    <button id="anchor">
+        Button is an anchor
+    </button>
+    <anchored-region
+        anchor="anchor"
+        verticalPositioningMode="locktodefault"
```
janechu: Use dash case for attributes.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+- 'dynamic': The anchored region is placed relative to the anchor element based on how much space is available between it and the viewport.  For example a menu that opens opens upwards when it is near the bottom of the page, and downwards when near the top.
+- 'onetime': As the 'dynamic' option, but position is not changed after initial placement.
+
+### API
+
+*Component name:*
+- `fast-anchored-region`
+
+*Attributes:*
+- anchor - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
```
janechu: Should this be an ID or should it be a slot? I could forsee an implementation that was more encapsulated looking like this:
```
<fast-anchored-region>
    <button slot="anchor">My Anchor</button>
    <div>The content of the anchored region</div>
</fast-anchored-region>
```

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+- viewport - Used to identify the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element. Possible values are the id of the desired viewport element, '#document' or '#parent'.  The default value is '#parent'.
```
janechu: Does this rely on the implementation including IDs in the DOM?

specs/anchored-region/anchored-region.md
janechu: Can you add them as a list under this bullet point so we can directly refer to them?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+- horizontalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
+- horizontalDefaultPosition - Can be 'start', 'end', 'left', 'right' or 'unset'.  Default is 'unset'
+- horizontalInset - Boolean that indicates whether the region should overlap the anchor on the horizontal axis. Default is false which places the region adjacent to the anchor element.
+- horizontalThreshold - Numeric value that defines how small in pixels the region must be to the edge of the viewport to switch to the opposite side of the anchor. The component favors the default position until this value is crossed.  When there is not enough space on either side or the value is unset the side with the most space is chosen.
```
janechu: Is this necessary? I would think you could do this by adding margin.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+- `fast-anchored-region`
+
+*Attributes:*
+- anchor - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
+
+- viewport - Used to identify the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element. Possible values are the id of the desired viewport element, '#document' or '#parent'.  The default value is '#parent'.
+
+- horizontalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
+- horizontalDefaultPosition - Can be 'start', 'end', 'left', 'right' or 'unset'.  Default is 'unset'
+- horizontalInset - Boolean that indicates whether the region should overlap the anchor on the horizontal axis. Default is false which places the region adjacent to the anchor element.
```
janechu: Can you provide some visuals as to what this would accomplish? I also believe it might go under "Next steps".

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+*Component name:*
+- `fast-anchored-region`
+
+*Attributes:*
+- anchor - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
+
+- viewport - Used to identify the HTMLElement used as the viewport the component uses to determine available layout space around the anchor element. Possible values are the id of the desired viewport element, '#document' or '#parent'.  The default value is '#parent'.
+
+- horizontalPositioningMode - Can be 'uncontrolled', 'locktodefault', 'dynamic' or 'onetime'.  Default is 'uncontrolled'.
+- horizontalDefaultPosition - Can be 'start', 'end', 'left', 'right' or 'unset'.  Default is 'unset'
```
janechu: I would also characterize this as "Next steps" because I believe default browser behavior can be reflected here, same as the verticalDefaultPosition where it is assumed to be a dropdown menu, aligned to the left of the anchor.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+- horizontalInset - Boolean that indicates whether the region should overlap the anchor on the horizontal axis. Default is false which places the region adjacent to the anchor element.
+- horizontalThreshold - Numeric value that defines how small in pixels the region must be to the edge of the viewport to switch to the opposite side of the anchor. The component favors the default position until this value is crossed.  When there is not enough space on either side or the value is unset the side with the most space is chosen.
+- horizontalScalingEnabled - The region is sized from code to match available space, in other scenarios the region gets sized via content size.
```
janechu: Also feels like an opportunity for "Next steps".

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Implementation
+
+### States
+Layout update checks in the component happen when:
+- the viewport scrolls
+- intersection observer reports a collision with the viewport
+- resize observer reports a resize event on the anchor, the viewport or the component itself.
+
+These layout checks analyse the dom geometry based on callbacks and repositions the anchored region appropriately: top/bottom/unset for the vertical axis and left/right/unset for the horizontal axis.
```
janechu: nit: *DOM

specs/anchored-region/anchored-region.md
scomea: I think this is the basic set of features based on what functionality that got used in the previous React version.  I have removed the "always on screen" behavior that we had there as it didn't seem to get used.  My reference for what this thing should do is the "position  menu" scenario (select) which the list you flag here is table stakes - menu always opens in same position, menu opens where there is most room, and menu scales with viewport.  There might be some small things we can do without, like maybe the option to only run layout once for example, but I see the proposed feature set as a logical starting point.  We should even be able to steal existing code for a lot of it.

specs/anchored-region/anchored-region.md
scomea: There are two "positioning mode" attributes so I'd rather not duplicate the detailed description.  It immediately precedes the api section of the doc so it is ver close

specs/anchored-region/anchored-region.md
scomea: Pointing to the code file for React Select rather than the explore site. I think as "background" pointing to previous implementations make sense as a tecnical reference.

specs/anchored-region/anchored-region.md
scomea: No, the anchor by definition can't be a child of the anchored region.  The pattern you show would likely be what you would see for a select like component that wraps the anchor (ie the trigger button that opens the select) as well as an anchored-region (the menu).

specs/anchored-region/anchored-region.md
scomea: Yes.  I'm open to other ideas.  The previous iteration used refs but that's a React thing.  I think we'd want to add the ability to pass references through JS as well, maybe not in first iteration tho. I'll add a mention of that.

specs/anchored-region/anchored-region.md
scomea: Yes, because if we have a scaling menu we still want to be able to have it flip to the other side at some point and can't use the dimensions of the content to figure that out in that case.

specs/anchored-region/anchored-region.md
scomea: Added explanations for 'inset' vs 'adjacent' to the design section.  I'd argue this is core functionality.

specs/anchored-region/anchored-region.md
scomea: I'd argue this is core functionality that should be part of the first iteration.

specs/anchored-region/anchored-region.md
scomea: I'd argue this is core functionality that should be part of the first iteration.

specs/anchored-region/anchored-region.md
janechu: So, good point on select, I think it's a good parallel. I'm thinking that select which accepts no options is a good example of default behavior where it merely tries to fit where it can in the available space. Should the anchored region be able to do the same? As in, be able to take no attributes and still have good default behavior, because that is what I think we should be inclined to start with. I agree with including default behavior for what you've listed, "menu opens in same position as default" (like browser select), "menu opens where there is the most room" (like browser select) and I would actually leave the scaling off for now, we could assume it's as big as the user determines it to be since they have full control over the content and styling. In this way we can start off with a smaller component that works out of the box and includes some default behavior and then re-examine every additional feature and add them incrementally.

specs/anchored-region/anchored-region.md
janechu: But the region we're talking about could be the slotted region, there doesn't need to be a wrapper if you don't want to have one. Correct me if I'm wrong but couldn't you do something like this:

```
<slot name="anchor">
    <button>My default anchor</button>
</slot>
<div part="content">
    <slot>
        <div>My default content</div>
    </slot>
</div>
```

Others might have a better idea of how to do this but you can see a similar concept in the checkbox spec https://github.com/microsoft/fast-dna/blob/master/specs/checkbox.md

scomea: The pattern you describe is pretty close to what I'd expect for a Select component that used this anchor to place its menu.  However I don't think it makes sense to implement that at the level of this component because there are common scenarios, like a flyout menu for example, where a single menu instance exists at the page level and the developer has it appear next to different anchor elements as the user interacts.  In the proposed API this would be done by changing the anchor id, but doesn't really work with the slot scenario.

specs/anchored-region/anchored-region.md
scomea: I think we know we want scaling, and we already have working code that I'm feeling reasonably good about being able to transfer (know more soon hopefully) and I want to be sure we support it well so I'd like to see it the initial feature set to be sure it integrates well (I'd want to have scaling prototyped before publishing the initial version at least).  If anything I was thinking that maybe the region *always* scales and authors use css to position items within the scaling region, but I've tended to back away from that as it feels like it would make base implementations more complex because you'd need to worry about css.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,285 @@
...
+
+For a more in-depth understanding of how this component works under the covers please refer to the [intersection observer api](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API). 
+
+### Use Cases
+
+It is envisioned that this component would be used as a building block for other components in this library (select, flyout, tooltip, etc) as well as being available for standalone use in responsive layouts.
+
+### Features
+
+- **Relative positioning:** Users can use it to position an element relative to another another element directly, like enabling a menu to open above or below a trigger button.
```
chrisdholt: It could be the language, or the way it's being described but technically in CSS I can already position something relative to another element. The value in the component here seems to be that it will ensure it doesn't render outside the viewport in the event of collision.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **Responsive scaling:** Users can use it to create a layout region that dynamically sizes depending on space between the anchor and the viewport elements.
+
+Note that the "always in view" functionality that existed in Viewport Positioner has been removed as this can be replicated through other means (such as a scaling region with grids, for example).
+
+### Risks and Challenges
+
+This component depends on getting accurate positioning data from the DOM outside of the component and it is possible for authors to break the relationship in advanced scenarios.
+
+Positioning may only be guaranteed if there are no other scrolling containers between the viewport and the anchor/anchored-region.  Authors should avoid nested scrolling containers.
```
chrisdholt: Can you elaborate on why this is the case? Without an explanation on why this is a challenge it's hard to consider it.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **Responsive positioning:** Users can use it to position an element relative to another element based on available space, for example a menu could open upwards if the trigger button is near the bottom of the page, and downwards if it is nearer the top.
+
+- **Responsive scaling:** Users can use it to create a layout region that dynamically sizes depending on space between the anchor and the viewport elements.
+
+Note that the "always in view" functionality that existed in Viewport Positioner has been removed as this can be replicated through other means (such as a scaling region with grids, for example).
+
+### Risks and Challenges
+
+This component depends on getting accurate positioning data from the DOM outside of the component and it is possible for authors to break the relationship in advanced scenarios.
```
chrisdholt: Can you elaborate on why this is the case? Without an explanation on why this is a challenge it's hard to consider it.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,285 @@
...
+
+Positioning may only be guaranteed if there are no other scrolling containers between the viewport and the anchor/anchored-region.  Authors should avoid nested scrolling containers.
+
+Also, some css settings can interfere with the data returned by intersection observer and break the positioning functionality.  For example an element with position="fixed" in the DOM hierarchy would be a no-no:
```
chrisdholt: Can you elaborate on why this is the case?

specs/anchored-region/anchored-region.md
chrisdholt: I agree with the idea of this having a good low-level default that can support the most common scenario. It would be helpful to understand how each of these break down in terms of both high and low-level API. If I have a simple app, I don't want to have to define any properties for this as part of adding a select on my page. For flyout, I may want more configuration, but I think we would similarly support a relatively simple API to start ( top / left) while allowing for more complexity. 

Is there a way that this could be a JS-only utility (or utilities) to support high and low-level API's?

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,285 @@
...
+
+![](./images/adjacent-inset.png)
+
+Inset on both axis positions the region so that it overlaps the anchor:
+
+![](./images/inset-inset.png)
+
+Adjacent on both axis positions the region diagonally to the anchor:
+
+![](./images/adjacent-adjacent.png)
```
chrisdholt: Can these be organized into a table?

specs/anchored-region/anchored-region.md
chrisdholt: @scomea I think there could be templating solutions though to handle what element is projected or assigned to the slot. If this region is generating DOM and the element needs to be on the same level as the region - it seems like the safest way to ensure the markup is as expected would be to leverage slots. I may also be misunderstanding something though.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,285 @@
...
+- vertical-scaling-enabled - The region is sized from code to match available space, in other scenarios the region gets sized via content size.
+
+- horizontal-position - read only, the current horizontal position of the component. Possible values are 'left', 'right' or 'unset'.
+- vertical-position - read only, the current vertical position of the component. Possible values are 'top', 'bottom' or 'unset'.
+
+*Slots:*
+- default slot for content
+
+*Events:*
+- positionchange - event is thrown whenever the placement of the region relative to the anchor changes (top/bottom and left/right).
```
chrisdholt: should we just have a change event?

specs/anchored-region/anchored-region.md
scomea: I see this one component as a building block for other components and those components would have a lot of flexibility as to how much.  So yes, a select could opt to offer a very basic set of options to the author and the select component would configure its internal anchored region appropriately.  Conversly a flyout could offer more flexibility.  I'm sure this could be reimagined a number of ways, but I think we'd just be moving the complexity around.

specs/anchored-region/anchored-region.md
scomea: The basic relative positioning has a convenience factor in that if you have a select component that responds to layout changes you could choose to turn off that functionality (ie. always open up) without having to have a different layout mechanism to support that.  But more importantly it allows the region to be anchored dynamically to any other element within its viewport regardless of whether they are relative to each other in the dom.

specs/anchored-region/anchored-region.md
scomea: The issue is that the component recalculates layout based on events - collisions, resizes and viewport scrolls.  Some implementations like a simple flyout that switches position when it collides with the edge of the viewport should work fine as we'll detect the collision regardless of intervening scrolling containers. Other implementations, like the scaling region, really need to recalculate layout with scroll events to track the edge of the viewport.  I think this is something we could work to improve on going forward withing needing to worry about it too much in an initial version.

specs/anchored-region/anchored-region.md
scomea: yes, actually I believe so.

specs/anchored-region/anchored-region.md
chrisdholt: Can you add that nuance to the document? I think it's important context.

specs/anchored-region/anchored-region.md
scomea: reworked section text

specs/anchored-region/anchored-region.md
scomea: reworked section text

specs/anchored-region/anchored-region.md
scomea: reworked section text

specs/anchored-region/anchored-region.md
scomea: Added "Additonally, the same anchored region can change which element it is anchored to dynamically.  For example a single tooltip instance in a page could be positioned next to any other element on the page by switching the anchor property of the anchored region that contains it."

specs/anchored-region/anchored-region.md
nicholasrice: I believe in HTML5 any string is valid as the value of the ID, so I don't think that using a '#' prefix gives us enough information: https://html.spec.whatwg.org/multipage/dom.html#attributes
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

I suppose we could see if that ID *does* exist and use it if it does, otherwise default to the desired behavior... I don't really like that behavior discrepancy though.

specs/anchored-region/anchored-region.md
janechu: I think complexity being opted into is desirable, there could be an opportunity for abstraction based on the dependency injections, which I believe is available in some form in the fast-element.

specs/anchored-region/anchored-region.md
janechu: Can you make a case for all of these as to why they are core functionality? I don't have a lot to go on thus far.

scomea: Default position (left right) and Inset gives us the ability to define placement relative to the anchor and cover the base arrangement for how one might want to place a menu relative to an anchor button (see images of arrangements in the Design section).

We want to support scaling regions.  I believe that having this as a bool attribute is easier to understand and implement than having all regions scale and using css enable various scaling/non-scaling scenarios.

If we support scaling regions we need to provide a way to specify a threshold value where the scaling region flips to the other side of the anchor.

The "mode" attribute seems essential in that users should be able to turn off positioning logic on an axis, or force a position.  The "one time" choice in this set we could cut, seems useful and cheap to me but it is an optimization that we could introduce later.

When I say "core functionality" I base it on what was needed/requested in initial uses of the React component.  I have cut a relatively complex feature (always on screen) as it was not used AFAIK and I believe there are ways to support this functionality with a scaling region + css.

specs/anchored-region/anchored-region.md
nicholasrice: I think using slots for this too prescriptive.  What if I want to attach a flyout to an element that I don't have control over rendering? Why should this element care *where* the anchor is? What is the *benefit* or using a slot vs an id reference?

I don't think an anchored region should care what it is anchored to and how that anchor is rendered - all it really should care about is if that element exists.

specs/anchored-region/anchored-region.md
nicholasrice: using IDs is a pretty idiomatic way to approach this I think - many native HTML elements rely on id associations.

janechu: I'm fine with it, I just don't know if we should be very prescriptive about what those IDs should be or have defaults for them.

nicholasrice: Agreed - I noted above but "#parent" is a valid ID which *could* have unintentional effects. We probably want to avoid *any* default value here.

specs/anchored-region/anchored-region.md
scomea: How about we go to a single default - if an author does not define a viewport ID we use the parent element.  If an author wants to use document root they need to do so with an id?

specs/anchored-region/anchored-region.md
scomea: Not sure exactly what you mean.  I don't believe the api proposed here adds complexity compared to the fast version.

specs/anchored-region/anchored-region.md
nicholasrice: I like that - does it cover all the cases?

specs/anchored-region/anchored-region.md
nicholasrice: I think that can go in the `@microsoft/fast-web-utilities` package

specs/anchored-region/anchored-region.md
janechu: The API is only a small part of my concern, it's the complexity internal to the component (the amount of code we ship with). What I'd like to see is this component being as light as possible out of the box. If there is a way to add responsive scaling, as an example, via dependency injection, that would mean that anyone who is not opting in to responsive scaling doesn't have to carry around the extra bloat. Instead we can create a utility they can inject into the component to provide that functionality if they need it.

specs/anchored-region/anchored-region.md
scomea: Presumably an author could assign an id to any element they wanted to use as a viewport

specs/anchored-region/anchored-region.md
nicholasrice: My primary concern with continuing to churn on this is that it is blocking Select and Auto Suggest components.

If we can't come to a resolution on this quickly I'd like to propose we table this control and prioritize a private implementation for those components so we can get those out the door.

specs/anchored-region/anchored-region.md
scomea: One of the issues we had with the previous version was that resizing the region when it was rendered would cause the host page layout to jiggle so the universal solution was to wrap the region in a fixed size 0x0 div.  I'm proposing that we should do this by default which requires an fixed size wrapping div.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,280 @@
...
+- 'dynamic': The anchored region is placed relative to the anchor element based on how much space is available between it and the viewport.  For example a menu that opens opens upwards when it is near the bottom of the page, and downwards when near the top.
+- 'onetime': As the 'dynamic' option, but position is not changed after initial placement.
+
+### API
+
+*Component name:*
+- `fast-anchored-region`
+
+*Attributes:*
+- anchor - The html id of the HTMLElement used as the anchor around which the positioning region is placed.  This must be set for the component's positioning logic to be active.
```
EisenbergEffect: For both `anchor` and `viewport` I think we should also have `anchorControl` and `viewportControl` properties (not attributes) which hold a direct reference to the anchor and viewport elements directly. When `anchor` or `viewport` changes, it will use the id to lookup the control and then set the appropriate control property. The rest can be driven off the control properties. This enables templates to actually directly bind to the control properties using refs and skip the id lookup if we want. So, you get an API that can work based off of id as well as control reference.

specs/anchored-region/anchored-region.md
```diff
@@ -0,0 +1,280 @@
...
+### Globalization
+Authors may want to change default position from left to right or vice versa based on rtl settings, but that can't be predicted by the component itself.
+
+### Performance
+Layout logic could called frequently during scrolling and resize operations so efficient code is a priority here. At a minimum layout recalculations should be limited to once per frame.
+
+### Dependencies
+[IntersectionObserver api](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) is unsupported on IE, and [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) is unsupported on IE, Safari and Firefox.  Both are required by the component.
+
+Question - do we support a fallback mode that relies on getBoundingClientRect() if the observers are not present?
```
EisenbergEffect: Maybe hold off on this for now. I think there are polyfills for the above, so maybe we can leverage them instead.

specs/anchored-region/anchored-region.md
scomea: Yes. I agree, will add that.

specs/anchored-region/anchored-region.md
scomea: Yup

specs/anchored-region/anchored-region.md
scomea: Although I'm thinking anchorElement and viewportElement might be better than "Control"?

EisenbergEffect: I agree. I think the built-in label uses the "control" name but it seems more accurate to use "element" terminology.

scomea:
Here is the working branch for the prototype component.  Not ready for review but it is at the "basically works" stage if folks want to take a look and give high level feedback.  Still needs a LOT of cleanup but feeling good about it. 
https://github.com/microsoft/fast-dna/tree/users/scomea/wc-positioner

PR #2773, opened 2020-03-10
chore: add text field spec

chrisdholt:
# Description
This PR adds a spec for a text field component

## Issue type checklist

## Process & policy checklist


specs/text-field/text-field.md
```diff
@@ -0,0 +1,142 @@
...
+Used anywhere an author might otherwise use:
+- input[type="text"]
+- input[type="email"]
+- input[type="password"]
+- input[type="tel"]
+- input[type="url"]
+
+### Features
+
+1. Form Association where supported
```
nicholasrice: I'm curious if you'll really want to use the form-association here because you'll actually be rendering an input element into the shadow DOM

specs/text-field/text-field.md
```diff
@@ -0,0 +1,142 @@
...
+### States
+
+**disabled**: `true` or `false` - when disabled, the value will not be changeable through user interaction. It should also not expose it's value to a form submission.
+
+**readonly**: `true` or `false` - when readonly, the value will not be changeable through user interaction. The value will still be exposed to forms on submission.
+
+### Accessibility
+
+The input element inside the shadow-dom of the checkbox will be a focusable element with the following accessability content attributes:
+- role: [textbox](https://www.w3.org/WAI/PF/aria/roles#textbox)
```
nicholasrice: I don't think you'll these attributes if you're bringing focus to the input itself - AT will recognize that focus is on an input element and respond accordingly

specs/text-field/text-field.md
```diff
@@ -0,0 +1,142 @@
...
+---
+
+### API
+
+Extends form associated custom element
+
+*Component Name*
+- `fast-text-field`
+
+*Attrs* - Adapted from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text)
```
EisenbergEffect: How many of these are just forwarding to the internal input element? Should I create a decorator similar to `@attr` that handles forwarding to another element inside the template?

chrisdholt: I would anticipate everything down to appearance would essentially be forwarded to the input. I'm trying to think of other scenarios like this where we would forward a significant amount of attrs down inside the template. I *think* this is a rare case, but I could be wrong.

specs/text-field/text-field.md
chrisdholt: Gotcha - so if we have an input being rendered within the shadow DOM it will still function as an input and submit with the form?

chrisdholt: no longer using form association

PR #2776, opened 2020-03-10
chore: add typography spec

bheston:
# Description

Added typography spec for web components

## Motivation & context

Initial structure for adaptive typography styles

## Issue type checklist

## Process & policy checklist


specs/typography.md
```diff
@@ -0,0 +1,177 @@
...
+
+Most other component frameworks either maintain the base font size or set it smaller to 14px. They also typically maintain heading element sizes based on `em` or `rem`, though some use individual specific values. Typically heading sizes are close to the browser default, though Material Design goes a lot larger at the top.
+
+---
+
+## Design
+
+The type styles published are consistently named so they are easy to identify and consume. A clear separation is made between the semantic naming and the raw type ramp.
+
+Default styles are applied to standard elements such that headings and other base typography pick up the adaptive styling without needing to add class names.
```
nicholasrice: This can have interop issues with components and will often lead to significant overrides of base elements within components - historically we've avoided doing this.

specs/typography.md
```diff
@@ -0,0 +1,177 @@
...
+*~~The key elements of the component's public API surface~~*
+
+### Adaptive Styling Properties (**Editorial**: Added this section)
+
+- `type-ramp-base-size` = "14px"  
+The base size for primary body text. Other type ramp sizes may eventually be determined from this starting value.
+
+### Anatomy and Appearance
+
+The type ramp is constructed as css variables that go plus and minus from the base. This allows a consumer to extend the ramp if necessary.
```
nicholasrice: This is a deviation from what we have today, where the ramp simply extends 1-9. I'd like to understand the benefits this approach provides - can you expand on the *why* of this +- scale?

specs/typography.md
```diff
@@ -0,0 +1,177 @@
...
+
+This ramp is currently based on fixed values according to the current visual design, but could be updated from an algorithm if desired.
+
+- `fast-type-ramp-base-size` = `type-ramp-base-size`
```
nicholasrice: I don't see line-height addressed here - is this spec non-inclusive of that?

specs/typography.md
```diff
@@ -0,0 +1,177 @@
...
+| Element    | Relative | Size   |
+|------------|----------|--------|
+| ```<h1>``` | 2em      | 32px   |
+| ```<h2>``` | 1.5em    | 24px   |
+| ```<h3>``` | 1.17em   | 18.7px |
+| ```<h4>``` | 1em      | 16px   |
+| ```<h5>``` | 0.83em   | 13.2px |
+| ```<h6>``` | 0.67em   | 10.7px |
+
+Any of these elements may be used with FAST-DNA styles applied.
```
chrisdholt: Can you elaborate on the emphasis with heading here? I'm not sure I understand the overall call out with headings, user agent styling values and how those relate to our typography story.

specs/typography.md
chrisdholt: I would like to understand the principle that's driving this part of the proposal and how it relates to our goals in FAST-DNA.  As mentioned above, traditionally we have avoided this and I think there is often good reason for doing so.

specs/typography.md
```diff
@@ -0,0 +1,177 @@
...
+- `fast-subheader-type-size` = `fast-type-ramp-plus-4-size`
+- `fast-page-title-type-size` = `fast-type-ramp-plus-3-size`
+- `fast-section-title-type-size` = `fast-type-ramp-plus-2-size`
+- `fast-paragraph-larger-type-size` = `fast-type-ramp-plus-1-size`
+- `fast-paragraph-type-size` = `fast-type-ramp-base-size`
+- `fast-paragraph-smaller-type-size` = `fast-type-ramp-minus-1-size`
+- `fast-label-type-size` = `fast-type-ramp-base-size`
+- `fast-button-type-size` = `fast-type-ramp-base-size`
+- `fast-caption-type-size` = `fast-type-ramp-minus-1-size`
+- `fast-caption-smaller-type-size` = `fast-type-ramp-minus-2-size`
```
chrisdholt: Can you provide comments as to when these each would be used? Is there a pattern that is being proposed for how this might extend in the future if need be?

chrisdholt: I'm also curious how this scales beyond our own internal principles? At first glance, this seems very opinionated compared to the current implementation of typography.

specs/typography.md
```diff
@@ -0,0 +1,177 @@
...
+
+*~~Important aspects of the planned implementation with careful consideration of web standards and integration.~~*
+
+### ~~States~~
+
+*~~Key component states, valid state transitions, and how interactions trigger a state transition.~~*
+
+### Accessibility
+
+The benefit of enabling standard elements to be styled is that if used properly they will already have the appropriate default aria roles and attributes. The downside is potential misuse and missing attributes. For instance ```<p class="fast-heading">My Heading</p>``` won't be properly identified as a heading by assistive technology.
```
chrisdholt: I'm not sure this assumption is accurate. Just because I want the visual treatment of a fast-heading doesn't mean I want a semantic treatment of that in a given scenario. The same could be said with subheading, paragraph, etc.

chrisdholt: This *is* accurate if my intent here is that this would be identified as a heading, but I kind of feel like this is a non-goal here. This reads to me as though we're trying to enforce a specific standard rather than support and empower.

EisenbergEffect:
I'm going to refrain from commenting on this spec for now and wait until the typography-specific details are worked through by others. At that time, I can review again with any eye towards how we implement in light of FastElement and our CSS library capabilities. I think that's likely where I can provide the most value.

janechu:
Some images to illustrate these concepts would be very helpful.

specs/typography.md
nicholasrice: One note on this that I forgot about - because of the isolation provided by web components we won't really have a leaking problem here. It likely isn't as big of an issue as it has been previously

EisenbergEffect:
How are we doing on this? I think we'll need to get typography moving ahead pretty soon. Are we close to having a final recommendation?

bheston:
@EisenbergEffect yes still working on it, continuing today. Perhaps we should chat about options for incorporating it. Maybe you, me, and @nicholasrice hopping on a voice channel later today.

chrisdholt:
@bheston please include me on the conversation if you decide to have it. I’d echo the concern here with typography for v1 specifically as we’re exploring net-new principles that have yet to be validated across the system.

EisenbergEffect:
I'm available this afternoon. Any time should work as long as we're wrapped up by 5pm pacific.

specs/typography.md
chrisdholt: I agree on empowering developers to easily create accessible markup. My experience has been that separating the semantic markup from the visual treatment of these better empowers developers to do the right thing and create accessible experiences. At the core of this, if the intent is to support and empower accessibility easily I would suggest that our best method of doing that is to keep it simple and simply provide classes as a way to style. This would allow us to point to the standards and recommendations while staying completely out of the way and keeping a singular way of applying the visual system. Doing that does make preventing abuse a non-goal, but I do think it creates clarity.

1. Markup should be semantic and accessible per the existing standards, you can use any elements you need to accomplish that in your app/page/scenario.

2. Classes are the methodology of applying visual styles to your semantic markup.

Again, this does leave room for what we might think is abusing the system (`<p class="my-heading">Foo</p>`), but I don't think we can assume that others would share that exact guidance. If the intended approach is to add more guardrails to prevent misuse I think we are working against our principles.

chrisdholt:
> I'm available this afternoon. Any time should work as long as we're wrapped up by 5pm pacific.

I have a few meetings this afternoon, but would like to be part of the discussion here. If we can look for a time on the calendar that would be helpful so I can move things if necessary.

specs/typography/typography.md
```diff
@@ -0,0 +1,143 @@
...
+- `typeRamp()`: A default type ramp as illustrated above
+- `typeRamp({'12px', '14px', '24px', '36px'}, 1)`: A type ramp with fixed type size values and default line height, index 1 is base
+- `typeRamp({{'12px', '16px'}, {'14px', '20px'}, {'24px', '28px'}, {'36px', '44px'}}, 1)`: A type ramp with fixed type size values and fixed individual line heights
+- `typeRamp(16, 1.25, 3, 5)`: A type ramp based at `16px` on a major third [modular scale](https://type-scale.com/) of `1.25` with `3` sizes below base and `5` sizes above.
+
+### Anatomy and Appearance
+
+The type ramp is presented as css variables that extend plus and minus from the base.
+
+- `fast-type-ramp-base-size` = `"14px"`
```
nicholasrice, line 70: Where do these properties come from? Are they defined by the design-system-provider component?

specs/typography/typography.md
```diff
@@ -0,0 +1,143 @@
...
+- [Material Design](https://material.io/design/typography/the-type-system.html)
+- [Bootstrap](https://getbootstrap.com/docs/4.4/content/typography)
+- [Ant Design](https://ant.design/components/typography)
+- [Mineral UI](https://mineral-ui.netlify.com/typography)
+
+---
+
+## Design
+
+The default type ramp uses a selection of values from [Microsoft Fluent Design](https://www.microsoft.com/design/fluent).
```
chrisdholt, line 45: I'm not seeing values after the jump here and I end up down a rabbit trail trying to find them. Do we need the link here or will the image do?

specs/typography/typography.md
```diff
@@ -0,0 +1,143 @@
...
+- `fast-type-ramp-plus-4-size` = `"34px"`
+- `fast-type-ramp-plus-4-height` = `"44px"`
+- `fast-type-ramp-plus-5-size` = `"46px"`
+- `fast-type-ramp-plus-5-height` = `"56px"`
+- `fast-type-ramp-plus-6-size` = `"60px"`
+- `fast-type-ramp-plus-6-height` = `"72px"`
+- `fast-type-ramp-minus-1-size` = `"12px"`
+- `fast-type-ramp-minus-1-height` = `"16px"`
+- `fast-type-ramp-minus-2-size` = `"10px"`
+- `fast-type-ramp-minus-2-height` = `"16px"`
```
chrisdholt, line 87: nit and more of an organizational change; based on the idea that these values scale out from the middle, it might help people visualize the "ramp" part of it by mimicking that here.

```
minus
base
plus
```

chrisdholt: As an aside - does this start from the "middle"? It looks like there are more plus values than minus values.

specs/typography/typography.md
```diff
@@ -0,0 +1,143 @@
...
+---
+
+## Implementation
+
+*~~Important aspects of the planned implementation with careful consideration of web standards and integration.~~*
+
+### ~~States~~
+
+*~~Key component states, valid state transitions, and how interactions trigger a state transition.~~*
+
```
chrisdholt, line 98: We can pull additional boilerplate

specs/typography/typography.md
```diff
@@ -0,0 +1,143 @@
...
+*~~Consider whether the component has any special globalization needs:~~*
+
+### ~~Security~~
+
+*~~Are there any security implications surrounding the component?~~*
+
+### ~~Performance~~
+
+*~~Are there any performance pitfalls or challenges with implementing the component?~~*
+
```
chrisdholt, line 114: Boilerplate can be pulled

specs/typography.md
nicholasrice: Cool thanks for explaining - I think that makes sense.

specs/typography/typography.md
```diff
@@ -0,0 +1,143 @@
...
+*~~Important aspects of the planned implementation with careful consideration of web standards and integration.~~*
+
+### ~~States~~
+
+*~~Key component states, valid state transitions, and how interactions trigger a state transition.~~*
+
+### Accessibility
+
+Since this spec only deals with type styles there are not traditional accessibility concerns, however this might be a point of potential future extension to allow for type sizing or capping of the smallest size.
```
chrisdholt, line 101: Not necessary, but it might be good to add a bit here about our principle in not directly tying our underlying typography architecture to HTML semantics.

specs/typography/typography.md
bheston: Good point. Link removed.

specs/typography/typography.md
bheston: No, the base is not necessarily the middle of the ramp, in fact probably rarely. I described it as extending up and down from the base, but does it read as "middle"?

bheston: Ah, I see "centered". I'll update that.

chrisdholt: I think the primary thing I saw was that there seems to be "plus six" but I only saw "minus two". It could just be wording, but it makes it seem like the ramp looks like this:

```
-2
-1
Base
+1
+2
+3
+4
+5
+5
```

specs/typography/typography.md
bheston: I could find a way to mention that here. I kind of mention it above about not having usage prescribed by the ramp, but that specifically follows the heading examples. I do plan on sticking with that idea, at least optionally, in the separate spec that describes heading styles though.

specs/typography/typography.md
bheston: Good question that made me realize I was essentially thinking of this as a fragment of the design system specifically for type sizes. I suspect there are plenty of reasons that we'd want or need to keep at least the minimal values used by the default FAST components in the main design system provider though.

I'll consider some simplification here, but also I'm now reconsidering the significance of the underlying ramp outside of the ways it's used or mapped to components. I'm going to get back to a test page real quick and review the usage in the main components.

specs/typography.md
bheston: I still believe we can support both, and will include this with examples in the headings spec.

specs/typography/typography.md
bheston: Right, that's the intent. The base size is not in the middle, and generally there will probably be more larger sizes than smaller.

PR #2785, opened 2020-03-11
feat: add text field web component

chrisdholt:
# Description
This PR adds text field as a new component

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/text-field/text-field.template.ts
+++ b/packages/fast-components/src/text-field/text-field.template.ts
@@ -1,0 +1,55 @@
+import { html, when } from "@microsoft/fast-element";
+import { ref } from "@microsoft/fast-element/dist/directives";
+import { TextField } from "./text-field";
+
+export const TextFieldTemplate = html<TextField>`
+    ${when(
+        x => x.childNodes.length,
+        html`
+        <label
+            part="label"
+            class="label"
+            for="control"
+        ><slot></slot></label>
+    `
+    )}
+    <div class="root" part="root">
+        <span
+            part="before-content"
+            ${ref("beforeContentContainer")}
+        >
+            <slot
+                name="before-content"
+                @slotchange=${x => x.handleBeforeContentChange()}
+                ${ref("beforeContent")}
+            ></slot>
+        </span>
+        <input
+            class="control"
+            part="control"
+            id=${x => (x.childNodes.length ? "control" : void 0)}
+            $type=${x => x.type}
+            $disabled=${x => x.disabled}
+            $placeholder=${x => x.placeholder}
+            $list=${x => x.list}
+            $maxlength=${x => x.maxlength}
+            $minlength=${x => x.minlength}
+            $pattern=${x => x.pattern}
+            $readonly=${x => x.readonly}
+            $required=${x => x.required}
+            $size=${x => x.size}
+            $spellcheck=${x => x.spellcheck}
+            $value=${x => x.value}
+        />
+        <span
+            part="after-content"
+            ${ref("afterContentContainer")}
+        >
+            <slot
+                name="after-content"
+                @slotchange=${x => x.handleAfterContentChange()}
+                ${ref("afterContent")}
+            ></slot>
+        </span>
+    </div>
+`;
--- a/packages/fast-components/src/text-field/text-field.template.ts
+++ b/packages/fast-components/src/text-field/text-field.template.ts
@@ -1,5 +1,4 @@
-import { html, when } from "@microsoft/fast-element";
-import { ref } from "@microsoft/fast-element/dist/directives";
+import { html, ref, when } from "@microsoft/fast-element";
 import { TextField } from "./text-field";
 
 export const TextFieldTemplate = html<TextField>`
--- a/packages/fast-components/src/text-field/text-field.template.ts
+++ b/packages/fast-components/src/text-field/text-field.template.ts
@@ -29,6 +29,7 @@ export const TextFieldTemplate = html<TextField>`
             part="control"
             id=${x => (x.childNodes.length ? "control" : void 0)}
             $type=${x => x.type}
+            $autofocus=${x => x.autofocus}
             $disabled=${x => x.disabled}
             $placeholder=${x => x.placeholder}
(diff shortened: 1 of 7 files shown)


packages/fast-components/src/text-field/text-field.template.ts
```diff
@@ -0,0 +1,57 @@
+import { html, when } from "@microsoft/fast-element";
+import { ref } from "@microsoft/fast-element/dist/directives";
```
janechu: I think this is now exported from the root so you should be able to leave off the `dist/directives`.

packages/fast-components/src/text-field/text-field.template.ts
```diff
@@ -0,0 +1,57 @@
...
+    <div class="root" part="root">
+        <span
+            part="before-content"
+            ${ref("beforeContentContainer")}
+        >
+            <slot
+                name="before-content"
+                @slotchange=${x => x.handleBeforeContentChange()}
+                ${ref("beforeContent")}
+            ></slot>
```
janechu: I wonder, should the span for before content (and after content) go in their respective slots?

packages/fast-components/src/text-field/text-field.ts
```diff
@@ -0,0 +1,95 @@
...
+    email = "email",
+    password = "password",
+    tel = "tel",
+    text = "text",
+    url = "url",
+}
+
+export class TextField extends FastElement {
+    @attr
+    public appearance: TextFieldAppearance = TextFieldAppearance.outline;
```
janechu: Sort of seems like the enum could be "default" otherwise we have to go back through the code to understand that outline is the default.

packages/fast-components/src/text-field/text-field.template.ts
```diff
@@ -0,0 +1,57 @@
+import { html, when } from "@microsoft/fast-element";
+import { ref } from "@microsoft/fast-element/dist/directives";
+import { TextField } from "./text-field";
+
+export const TextFieldTemplate = html<TextField>`
+    ${when(
+        x => x.childNodes.length,
```
EisenbergEffect: I've seen `childNodes.length` being used with `when` in some new PRs. I apparently missed calling this out in an earlier component. This won't work if the child nodes change after the component is initially rendered because child nodes is an exotic, non-array thing coming from native DOM land. Basically, we have no way to automatically observe it for changes. To truly make this work, you would need to base `when` off of a boolean observable and then create a mutation observer to watch the child nodes and update that observable whenever the nodes changed.

I added a new feature to our roadmap to create a `children` decorator that will create a property of observable child nodes and allow you to have a simple callback to respond when they change. That should make this easy in the future.

packages/fast-components/src/text-field/text-field.ts
```diff
@@ -0,0 +1,95 @@
...
+    private appearanceChanged(): void {
+        this.appearance === TextFieldAppearance.filled
```
EisenbergEffect: This is a perfect use for host directives...but I think there would be negative side-effects with my current implementation as it would blow away any end-user classes, when in reality we want to merge them. I think after I get some of my binding refactoring done, we may able to look at improving this so it could be done declaratively instead. No need to change anything now.

packages/fast-components/src/text-field/text-field.template.ts
EisenbergEffect: If it went into the slot, then user provided slot content would replace that, since content inside a slot is treated as fallback content, rendered only when the user doesn't provide actual content. I this case, I think the span is needed in either case for layout reasons.

packages/fast-components/src/text-field/text-field.template.ts
EisenbergEffect: Yep, this one is now available directly from `@microsoft/fast-element` as of the PR I merged this morning.

packages/fast-components/src/text-field/text-field.ts
chrisdholt: @JaneChu I'm hesitant to change from this convention as the default could change as the system or design language changes. If the default were to change we would additionally need to add an appearance for `outline`.

packages/fast-components/src/text-field/text-field.template.ts
chrisdholt: Correct - we want some basic CSS on the before and after content. Some of which needs to be specific to the left or right of the control. Unfortunately there isn't a methodology for applying the styling to the slot itself via a class so if we want the control to support before and after content without looking broken we need to include the spans to do so

packages/fast-components/src/text-field/text-field.template.ts
chrisdholt: I'll update now that it's merged.

packages/fast-components/src/text-field/text-field.template.ts
chrisdholt: I think label tends to not be dynamic. With that said, we should update this based on the roadmap - I'd prefer to handle these all at once though as it will be a common scenario for all these form elements.

packages/fast-components/src/text-field/text-field.ts
janechu: I suppose what we're asking is if you wanted the default style and it was changed from the outline style to something else, would that be OK. Seems like a question for @bheston

chrisdholt: Outline and filled are the appearances, it just so happens that outline is what we default to. If a user wanted whatever our default appearance was and our default appearance changed either to filled or something new, they would automatically get that by not setting it. If they do set it to outline or filled, we respect those.

I think the question above is, if someone wants the outline appearance and that is what they get out of the box, they may not set it. If they do not set it and then we change the default to filled that would break their expectation. One potential fix to that would be an additional enum of "default" which then sets to outline; however, if we set that as the default we still have the same problem. If I don't set anything for the appearance it gets set to a default and then that changes. The only consistent way to ensure an explicit style would be to explicitly set it, which I think would be our preference.

janechu: I think that's good reasoning.

PR #2786, opened 2020-03-11
chore: add text area spec

chrisdholt:
# Description
This PR adds a new spec for a text area component.

## Issue type checklist

## Process & policy checklist


specs/text-area/text-area.md
```diff
@@ -0,0 +1,130 @@
...
+*Screenshots below are of the basic appearance of the component and are not exhaustive.*
+
+| State | Image |
+| ----- | ----- |
+| default | ![](./images/text-area.png) |
+| focus | ![](./images/text-area-focus.png)
+| placeholder | ![](./images/text-area-placeholder.png)
+
+*Slot Names*
+- default - the content of the text area. Unlike input where content is a value, the content of a text area is predefined children.
```
janechu: Could you re-word to "Unlike an input which stores a value and has no content, text area can have content which is treated as the value. This content must always be text."

specs/text-area/text-area.md
```diff
@@ -0,0 +1,130 @@
...
+- label - the label content
```
janechu: Should the images show a one with a label?

chrisdholt: We can - I don't technically have an image for that or filled...so I didn't include those. Happy to though.

specs/text-area/text-area.md
chrisdholt: sure.

specs/text-area/text-area.md
chrisdholt: added

specs/text-area/text-area.md
chrisdholt: updated!

chrisdholt: I changed this slightly as it seems we don't need slot projection to pass the children here

chrisdholt:
> Approving this, but I'm wondering about another approach. What if we took our existing text field WC and gave it a `multiline` attribute and when that was true we used a text area instead of an input? (or we could add a new type value of textarea) Would that reduce duplication? Seems there's a lot of similarities.

I agree that there is some duplication and similarities in terms of attr's, but I think this might end up making the control a bit complex. In the text field scenario, I think a goal is to support easy addition of things like glyphs, or perhaps even a button (TBD). I'm not sure how that would work with a textarea from a design standpoint - though it could be explored.

@EisenbergEffect Perhaps this can be something to track and gather convergence on as part of Open-UI?

PR #2787, opened 2020-03-11
feat: add text area web component

chrisdholt:
# Description
This PR adds Text Area as a web component.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/text-area/fixtures/text-area.html
+++ b/packages/fast-components/src/text-area/fixtures/text-area.html
@@ -23,11 +23,8 @@
 
     <!-- Read only -->
     <h4>Read only</h4>
-    <fast-text-area readonly>Readonly text area</fast-text-area>
-    <fast-text-area readonly>
-        <span slot="label">label</span>
-        Readonly text area
-    </fast-text-area>
+    <fast-text-area readonly value="Readonly text area"></fast-text-area>
+    <fast-text-area readonly value="Readonly text area">label</fast-text-area>
 
     <!-- Read only -->
     <h4>Autofocus</h4>
@@ -68,11 +65,8 @@
 
     <!-- Read only -->
     <h5>Read only</h5>
-    <fast-text-area appearance="filled" readonly>Readonly text area</fast-text-area>
-    <fast-text-area appearance="filled" readonly>
-        <span slot="label">label</span>
-        Readonly text area
-    </fast-text-area>
+    <fast-text-area appearance="filled" value="Readonly text area">label</fast-text-area>
+    <fast-text-area appearance="filled" value="Readonly text area">label</fast-text-area>
 
     <!-- With label -->
     <h4>Visual vs audio label</h4>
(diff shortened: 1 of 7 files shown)


packages/fast-components/src/text-area/fixtures/text-area.html
```diff
@@ -0,0 +1,83 @@
...
+    <h5>Disabled</h5>
+    <fast-text-area appearance="filled" disabled></fast-text-area>
+    <fast-text-area appearance="filled" disabled>
+        <span slot="label">label</span>
+    </fast-text-area>
+    <fast-text-area appearance="filled" disabled placeholder="placeholder"></fast-text-area>
+
+    <!-- Read only -->
+    <h5>Read only</h5>
+    <fast-text-area appearance="filled" value="Readonly text area">label</fast-text-area>
```
eljefe223: I seem to be able to edit the filled read only fields

chrisdholt: interesting, I'll check that out

chrisdholt: ah -specifically the filled ones

chrisdholt: oh, that's because it doesn't have the attr! Updating :)

chrisdholt: @eljefe223 resolved!

packages/fast-components/src/text-area/text-area.template.ts
```diff
@@ -0,0 +1,38 @@
+import { html, ref, when } from "@microsoft/fast-element";
+import { TextArea } from "./text-area";
+
+export const TextAreaTemplate = html<TextArea>`
+    ${when(
```
EisenbergEffect: This will work for initial render, but not if the child nodes change later. We've already discussed so we'll leave this as is for now. I've got a new feature coming, hopefully next week, that will enable us to do this in a more robust way.

packages/fast-components/src/text-area/text-area.template.ts
```diff
@@ -0,0 +1,38 @@
...
+            for="control"
+        >
+            <slot name="label"></slot>
+        </label>
+    `
+    )}
+    <textarea
+        part="control"
+        class="control"
+        id="${x => (x.childNodes.length ? "control" : void 0)}"
```
EisenbergEffect: Any reason not to just give it an id either way?

packages/fast-components/src/text-area/text-area.ts
```diff
@@ -0,0 +1,102 @@
...
+    public connectedCallback(): void {
+        super.connectedCallback();
+
+        if (this.value) {
+            this.textarea.value = this.value;
+        }
+    }
+
+    public handleTextInput = (): void => {
+        this.$emit("change", this.textarea.value);
```
EisenbergEffect: Yay! Nice to see the new helper being used 😄

EisenbergEffect: I think handling the `input` event is fine for now, but in a second pass over our form controls, we may want to surface both change and input. See https://stackoverflow.com/questions/17047497/difference-between-change-and-input-event-for-an-input-element Something to think about. /cc @nicholasrice

packages/fast-components/src/text-area/text-area.template.ts
chrisdholt: No, just thinking if it's not used it's not needed, but we can add either way.

PR #2789, opened 2020-03-13
feat: add switch as new web component

eljefe223:
# Description

This PR adds Switch as a web component.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/switch/switch.styles.ts
+++ b/packages/fast-components/src/switch/switch.styles.ts
@@ -99,13 +99,4 @@ export const SwitchStyles = css`
     :host(.checked) .checked-message {
         display: block;
     }
-
-    .root {
-        margin-bottom: auto;
-        border-radius: calc(var(--elevated-corner-radius));
-        width: var(--dialog-width);
-        height: var(--dialog-height);
-        background: var(--background-color);
-        z-index: 1;
-    }
 `;
--- a/packages/fast-components/src/switch/switch.styles.ts
+++ b/packages/fast-components/src/switch/switch.styles.ts
@@ -1,16 +1,19 @@
 import { css } from "@microsoft/fast-element";
-import { disabledCursor } from "../styles";
+import { disabledCursor, display } from "../styles";
 
 export const SwitchStyles = css`
     :host([hidden]) {
         display: none;
     }
 
-    :host {
-        position: relative;
-        display: inline-flex;
+    ${display("inline-flex")} :host {
         align-items: center;
         outline: none;
+        margin: calc(var(--design-unit) * 1px) 0;
+        ${/*
+           * Chromium likes to select label text or the default slot when
+           * the checkbox is clicked. Maybe there is a better solution here?
+           */ ""} user-select: none;
     }
 
     :host(.disabled) {
@@ -49,7 +52,7 @@ export const SwitchStyles = css`
         border-color: var(--neutral-focus);
     }
 
-    .status-indicator {
+    .checked-indicator {
         position: absolute;
         height: calc((var(--height-number) - (var(--design-unit) * 2)) * 1px);
         width: calc((var(--height-number) - (var(--design-unit) * 2)) * 1px);
@@ -71,11 +74,11 @@ export const SwitchStyles = css`
         ${/* Need to discuss with Brian how HorizontalSpacingNumber can work. https://github.com/microsoft/fast-dna/issues/2766 */ ""} margin-inline-end: calc(var(--design-unit) * 2px + 2px);
     }
 
-    ::slotted(.content) {
+    ::slotted(*) {
         ${/* Need to discuss with Brian how HorizontalSpacingNumber can work. https://github.com/microsoft/fast-dna/issues/2766 */ ""} margin-inline-start: calc(var(--design-unit) * 2px + 2px);
     }
 
-    :host(.checked) .status-indicator {
+    :host(.checked) .checked-indicator {
         left: calc((var(--height-number) + var(--design-unit)) * 1px);
         background: var(--accent-foreground-cut-rest);
     }
--- a/packages/fast-components/src/switch/switch.styles.ts
+++ b/packages/fast-components/src/switch/switch.styles.ts
@@ -1,0 +1,111 @@
+import { css } from "@microsoft/fast-element";
+import { disabledCursor } from "../styles";
+
+export const SwitchStyles = css`
+    :host([hidden]) {
+        display: none;
+    }
+
+    :host {
+        position: relative;
+        display: inline-flex;
+        align-items: center;
+        outline: none;
+    }
+
+    :host(.disabled) {
+        opacity: var(--disabled-opacity);
+    }
+
+    :host(.disabled) .label,
+    :host(.readonly) .label,
+    :host(.readonly) .switch,
+    :host(.disabled) .switch {
+        cursor: ${disabledCursor};
+    }
+
+    .switch {
+        position: relative;
+        outline: none;
+        width: calc(var(--height-number) * 2px);
+        height: calc(var(--height-number) * 1px);
+        background: var(--neutral-fill-input-rest);
+        border-radius: calc(var(--height-number) * 1px);
+        border: calc(var(--outline-width) * 1px) solid var(--neutral-outline-rest);
+    }
+
+    .switch:hover {
+        background: var(--neutral-fill-input-hover);
+        border-color: var(--neutral-outline-hover);
+    }
+
+    .switch:active {
+        background: var(--neutral-fill-input-active);
+        border-color: var(--neutral-outline-active);
+    }
+
+    :host(:focus) .switch {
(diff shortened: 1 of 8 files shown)


packages/fast-components/src/switch/switch.styles.ts
```diff
@@ -0,0 +1,102 @@
...
+        color: var(--neutral-foreground-rest);
+        cursor: pointer;
+        ${/* Font size is temporary - replace when adaptive typography is figured out */ ""} font-size: calc(1rem + (var(--density) * 2px));
+    }
+
+    .label {
+        ${/* Need to discuss with Brian how HorizontalSpacingNumber can work. https://github.com/microsoft/fast-dna/issues/2766 */ ""} margin-inline-end: calc(var(--design-unit) * 2px + 2px);
+    }
+
+    ::slotted(.content) {
```
nicholasrice: I think we want to find an approach that doesn't require the author to put a class on slotted content to achieve the correct styling.

packages/fast-components/src/switch/switch.styles.ts
```diff
@@ -0,0 +1,102 @@
...
+        ${/* Need to discuss with Brian how HorizontalSpacingNumber can work. https://github.com/microsoft/fast-dna/issues/2766 */ ""} margin-inline-start: calc(var(--design-unit) * 2px + 2px);
+    }
+
+    :host(.checked) .status-indicator {
```
nicholasrice: checkbox uses "checked-indicator" - should this be consistent?

packages/fast-components/src/switch/switch.template.ts
```diff
@@ -0,0 +1,41 @@
+import { html, when } from "@microsoft/fast-element";
+import { Switch } from "./switch";
+import { bool } from "../utilities";
+
+export const SwitchTemplate = html<Switch>`
+    <template
+        role="switch"
+        $aria-checked="${x => bool(x.checked)}"
+        $aria-disabled="${x => bool(x.disabled)}"
```
nicholasrice: I think you'll also want to reflect aria-required here.

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        this.updateForm();
+
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.checked = this.checked;
+        }
+
+        if (this.constructed) {
+            this.dispatchEvent(
+                new CustomEvent("change", { bubbles: true, composed: true })
```
nicholasrice: This should use `this.$emit`

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,100 @@
...
+            );
+        }
+
+        this.checked ? this.classList.add("checked") : this.classList.remove("checked");
+    }
+
+    protected proxy = document.createElement("input");
```
nicholasrice: You'll want to make sure this input has it's `type` attribute set to `"checkbox"`

packages/fast-components/src/switch/switch.styles.ts
```diff
@@ -0,0 +1,102 @@
...
+    :host([hidden]) {
+        display: none;
+    }
+
+    :host {
+        position: relative;
+        display: inline-flex;
+        align-items: center;
+        outline: none;
+    }
```
chrisdholt: You could use the `display()` helper that @nicholasrice put together for this.

packages/fast-components/src/switch/switch.styles.ts
chrisdholt: Would it be fair to say that regardless of what "content" is, you'd want to have the spacing there? If so, you could just use:
```css
::slotted(*) { }
```

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Tracks whether the "checked" property has been changed.
+     * This is necessary to provide consistent behavior with
+     * normal input checkboxes
+     */
```
chrisdholt: I actually am not sure that we want the same behavior as normal input checkboxes here, but I do think that's we have here for some of this.

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,100 @@
+import { attr, observable } from "@microsoft/fast-element";
+import { FormAssociated } from "../form-associated";
+import { keyCodeSpace } from "@microsoft/fast-web-utilities";
+import { bool } from "../utilities";
+
+export class Switch extends FormAssociated<HTMLInputElement> {
```
chrisdholt: Is this a form associated element? Do we want to base our switch off a checkbox?

specs/switch.md
```diff
@@ -42,8 +42,8 @@ Typically it's best to include a checked and unchecked message but it is not req
 
 *Slots:*
 - `default`
-- `unchecked-message`
-- `checked-message`
+- `unchecked-message` pass "content" class for default styling
+- `checked-message` pass "content" class for default styling
```
chrisdholt: I mentioned above, but I think we should just apply this for whatever gets passed here.

specs/switch.md
```diff
@@ -77,10 +81,10 @@ Typically it's best to include a checked and unchecked message but it is not req
 ```
 <fast-switch
     checked={true}
-    unchecked-message="On"
-    checked-message="Off"
 >
     Notify by Email
+    <span class="content" slot="checked-message">On</span>
+    <span class="content" slot="unchecked-message">Off</span>
```
chrisdholt: Can you remove the classes as part of the change?

packages/fast-components/src/switch/switch.ts
eljefe223: It is a checkbox just without an indeterminate state. I went over this with @nicholasrice earlier today so he may have some feedback.

specs/switch.md
eljefe223: I like your idea, thanks for the suggestion.

packages/fast-components/src/switch/switch.template.ts
```diff
@@ -0,0 +1,43 @@
...
+export const SwitchTemplate = html<Switch>`
+    <template
+        role="switch"
+        $aria-checked="${x => bool(x.checked)}"
+        $aria-required="${x => bool(x.required)}"
+        $aria-disabled="${x => bool(x.disabled)}"
+        $aria-readonly="${x => bool(x.readOnly)}"
+        $tabindex="${x => (bool(x.disabled) ? null : 0)}"
+        @keypress="${(x, c) => x.keypressHandler(c.event as KeyboardEvent)}"
+        @click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
```
EisenbergEffect: Really awesome to see the new host directives in use!

packages/fast-components/src/switch/switch.template.ts
```diff
@@ -0,0 +1,43 @@
...
+    >
+        ${when(
+            x => x.childNodes.length,
```
EisenbergEffect: We'll leave this as is for now, but change it once we get some new features. Just note that this will only work for initial render but if the children change later, the label won't be added/removed. Nothing to change for now.

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,112 @@
...
+            this.dirtyChecked = true;
+        }
+
+        this.updateForm();
+
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.checked = this.checked;
+        }
+
+        if (this.constructed) {
```
EisenbergEffect: You can remove all the `if (this.constructed)` checks around calles to `$emit`. One of the advantages of `$emit` is that it won't fire the event if the node isn't fully connected.

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,112 @@
...
+     * Tracks whether the "checked" property has been changed.
+     * This is necessary to provide consistent behavior with
+     * normal input checkboxes
+     */
+    private dirtyChecked: boolean = false;
+
+    /**
+     * Set to true when the component has constructed
+     */
+    private constructed: boolean = false;
```
EisenbergEffect: We can remove this bool. See note above.

packages/fast-components/src/switch/switch.ts
```diff
@@ -0,0 +1,104 @@
+import { attr, observable } from "@microsoft/fast-element";
+import { FormAssociated } from "../form-associated";
+import { keyCodeSpace } from "@microsoft/fast-web-utilities";
+import { bool } from "../utilities";
+import { type } from "os";
```
janechu: where is this coming from?

eljefe223: Yeah that was one of those VS Code auto import situations... Thanks!

PR #2791, opened 2020-03-13
refactor(fast-element): improve template compilation, directives, behaviors, and view instantiation

EisenbergEffect:
# Description

A pretty hefty refactor of the template compiler, directive API, behaviors, and how view instantiation hooks things up. If you want to see how a modern templating engine works, this PR touches almost all the areas of interest.

## Motivation & context

This PR fixes some bugs, corrects some awkward APIs, makes things more consistent and simpler (esp. around extensibility), removes those "fa" classes from everything, improves performance in a few specific cases, etc.

### Items of Interest

* The template compiler and view instantiation logic has been moved over to use a `TreeWalker`. This actually enables a lot of things to be simplified, since the compiler no longer needs to create fake elements for things like when/if/text just so a query selector can find them later. This also means that during view instantiation we no longer have to remove those fake elements, so the compile and creation process gets more efficient as a result.
* The `Directive` API was a bit awkward and is now re-designed to have a simple interface to implement for custom directives.
* The `BindingBehavior` class used to have lots of sub-classes. It's now a single class with just some per-directive method swapping.

In the process of making these changes, various related items were cleaned up and a number of bugs were discovered and fixed.

## Issue type checklist

- [x] **Performance**: A change that improves performance or reduces memory allocation.
- [x] **Refactor**: A change that improves existing code without adding/removing features.

This is technically a breaking change. But like many of the changes I've been making, it is unlikely to affect anyone since this changes internals and lower-level APIs than most folks are using right now.

## Process & policy checklist


packages/fast-element/src/directives/binding.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    this.source = null;
+}
+
+function triggerUnbind(this: BindingBehavior) {
+    this.target.removeEventListener(this.targetName!, this, true);
+    this.source = null;
+}
+
+function updateAttributeTarget(this: BindingBehavior, value: unknown): void {
```
EisenbergEffect: Note to self: this implementation is incorrect. Need to fix this before merging.

packages/fast-element/src/directives/binding.ts
```diff
@@ -0,0 +1,168 @@
...
+    (value as boolean)
+        ? (this.target as HTMLElement).setAttribute(this.targetName!, "")
```
nicholasrice: Reminder to fix :) these look the same

packages/fast-element/src/directives/binding.ts
```diff
@@ -0,0 +1,168 @@
...
+            case BindingType.attribute:
+                this.updateTarget = updateAttributeTarget;
+                break;
+            case BindingType.booleanAttribute:
+                this.updateTarget = updateBooleanAttributeTarget;
+                break;
+            case BindingType.text:
+                this.updateTarget = updateTextTarget;
+                break;
+            case BindingType.text:
```
janechu: both of these cases are the same?

EisenbergEffect: Good catch! Fixed now.

packages/fast-element/src/directives/binding.ts
EisenbergEffect: Fixed. Thanks!

PR #2795, opened 2020-03-16
feat: add badge web component

eljefe223:
# Description

Adds `fast-badge` as a web component.

image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/badge/badge.styles.ts
+++ b/packages/fast-components/src/badge/badge.styles.ts
@@ -65,6 +65,6 @@ export const BadgeStyles = css`
     :host(.circular.large) .badge {
         ${/* Need to work with Brian on width and height here */ ""} height: calc(var(--height-number) * 1px);
         min-width: calc(var(--height-number) * 1px);
-        padding: 0 calc(var(--horizontal-spacing) * 1px);
+        padding: 0 calc(var(--design-unit) * 1px);
     }
 `;
--- a/packages/fast-components/src/badge/badge.styles.ts
+++ b/packages/fast-components/src/badge/badge.styles.ts
@@ -18,50 +18,43 @@ export const BadgeStyles = css`
 
     :host(.circular) .badge {
         border-radius: 100px;
-        padding: 0;
+        padding: 0 calc(var(--design-unit) * 1px);
         display: flex;
         align-items: center;
         justify-content: center;
-        overflow: hidden;
         box-sizing: border-box;
     }
 
     :host(.small) .badge {
         ${/* Font size, weight, and line height are temporary - 
             replace when adaptive typography is figured out */ ""} font-size: 10px;
-        font-weight: 400px;
         line-height: 16px;
-    }
-
-    :host(.circular.small) .badge {
-        ${/* Need to work with Brian on width and height here */ ""} height: 20px;
-        min-width: 20px;
-        padding: 0 calc(var(--design-unit) * 1px);
     }
 
     :host(.medium)) .badge {
         ${/* Font size, weight, and line height are temporary - 
             replace when adaptive typography is figured out */ ""} font-size: 12px;
-        font-weight: 400px;
         line-height: 18px;
-    }
-
-    :host(.circular.medium) .badge {
-        ${/* Need to work with Brian on width and height here */ ""} height: 26px;
-        min-width: 26px;
-        padding: 0 calc(var(--design-unit) * 1px);
     }
 
     :host(.large) .badge {
         ${/* Font size, weight, and line height are temporary - 
             replace when adaptive typography is figured out */ ""} font-size: 14px;
-        font-weight: 400px;
         line-height: 20px;
+    }
+
+    :host(.circular.small) .badge {
+        ${/* Need to work with Brian on width and height here */ ""} height: calc((var(--height-number) - (var(--design-unit) * 3)) * 1px);
+        min-width: calc((var(--height-number) - (var(--design-unit) * 3)) * 1px);
+    }
+
+    :host(.circular.medium) .badge {
+        ${/* Need to work with Brian on width and height here */ ""} height: calc((var(--height-number) - (var(--design-unit) * 1.5)) * 1px);
+        min-width: calc((var(--height-number) - (var(--design-unit) * 1.5)) * 1px);
     }
 
     :host(.circular.large) .badge {
         ${/* Need to work with Brian on width and height here */ ""} height: calc(var(--height-number) * 1px);
         min-width: calc(var(--height-number) * 1px);
-        padding: 0 calc(var(--design-unit) * 1px);
     }
 `;
--- a/packages/fast-components/src/badge/badge.styles.ts
+++ b/packages/fast-components/src/badge/badge.styles.ts
@@ -1,0 +1,69 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "../styles";
+
+export const BadgeStyles = css`
+    ${display("inline-block")} :host {
+        box-sizing: border-box;
+        font-family: var(--body-font);
+    }
+
+    .badge {
+        border-radius: calc(var(--elevated-corner-radius) * 1px);
+        padding: calc(var(--design-unit) * 1.5px);
+        ${/* Font size, weight, and line height are temporary - 
+            replace when adaptive typography is figured out */ ""} font-size: 12px;
+        font-weight: 400px;
+        line-height: 18px;
+    }
+
+    :host(.circular) .badge {
+        border-radius: 100%;
+        padding: 0;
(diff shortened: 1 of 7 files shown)


packages/fast-components/src/badge/badge.ts
```diff
@@ -0,0 +1,45 @@
...
+    // Revisit this once we have a better
+    // story for ensuring proper contrast
+    // from author defined `fill`
+    @attr({ attribute: "color" })
+    public color: string;
+
+    @attr
+    public size: BadgeSize;
+    private sizeChanged(): void {
+        switch (this.size) {
```
EisenbergEffect: I think you will also need to remove the previous size class.

eljefe223: Yeah, I think what we want to do here is just remove the current size before we enter the switch.

```
this.classList.remove(this.size);
```

packages/fast-components/src/badge/badge.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    // Revisit this once we have a better
+    // story for ensuring proper contrast
+    // from author defined `fill`
+    @attr({ attribute: "color" })
+    public color: string;
+
+    @attr
+    public size: BadgeSize;
+    private sizeChanged(): void {
```
nicholasrice: @EisenbergEffect looking at observable it seems like it would be trivial to invoke this change handler with a previous value (and "current" value, if we wanted) - that would help in a situation like this where we need to know what the value *was* - do you have concerns with that? If not I think it would be a useful feature.

EisenbergEffect:
@nicholasrice I think we can do that. I've added it to the list of things to address as part of the attr improvements. I'm not sure if we can do it with attr or not. So, if I can, we'll do it in both places, otherwise we may need to fork the codepaths. I'm doing this together because I may need to fork the codepaths anyway. I hope that makes sense...

nicholasrice:
Sounds good, thanks @EisenbergEffect!

packages/fast-components/src/badge/badge.ts
eljefe223: @nicholasrice is there something actionable here?

nicholasrice: not in the interim :) Was just hijacking your code to make Rob aware of a potential API improvement.

EisenbergEffect:
My latest PR adds oldValue and newValue to the sig.

packages/fast-components/src/badge/fixtures/base.html
```diff
@@ -0,0 +1,71 @@
...
+            --badge-fill-tertiary: #DFDFDF;
+            --badge-fill-foo: #EE1010;
+            --badge-fill-bar: rgba(238,6,16,0.7);
+            --badge-color-foo: white;
+            --badge-color-bar: black;
+        }
+    </style>
+    <h1>Badge</h1>
+    <h4>Default</h4>
+    <fast-badge fill="primary" style="font-size: 10px; line-height: 16px;" color="foo" id="foo" aria-label="Simple badge" modal="true">
```
chrisdholt, line 15: modal true?

chrisdholt: looks like turtles all the way down

chrisdholt, line 15: I'm not sure I'm fully understanding the purpose of setting all the font properties as inline styles here.

packages/fast-components/src/badge/fixtures/base.html
```diff
@@ -0,0 +1,71 @@
...
+        Badge
+    </fast-badge>
+    <fast-badge fill="secondary" color="foo" id="foo" aria-label="Simple badge" modal="true">
```
chrisdholt, line 18: What's the benefit of the aria-labels here?

packages/fast-components/src/badge/fixtures/base.html
```diff
@@ -0,0 +1,71 @@
...
+            <fast-badge style="position: absolute; right: -6px; top: -6px;" circular="true" fill="bar" color="foo" id="foo" aria-label="Simple badge" modal="true">
+                999
+            </fast-badge>
+        </div>
+        <div style="position: relative; display: inline-flex; width: 50px; height: 50px; background: #DFDFDF; margin: 12px;">
+            <fast-badge style="position: absolute; right: -6px; top: -6px;" circular="true" fill="bar" color="foo" id="foo" aria-label="Simple badge" modal="true">
+                15
+            </fast-badge>
+        </div>
+    </div>
```
chrisdholt, line 70: Without context I'm not fully understanding the scope of the examples here. If these are configurations to accomplish what was outlined in the spec it seems fairly complex to create different variations of badges.

packages/fast-components/src/badge/badge.template.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const BadgeTemplate = html<Badge>`
+    <template>
+        <div
+            class="badge"
+            style="background-color: var(--badge-fill-${x =>
+                x.fill}); color: var(--badge-color-${x => x.color})"
+        >
+            <slot></slot>
+        </div>
```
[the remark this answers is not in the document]
eljefe223: It is because we can't assign styling inline in the template and pass styling to the component implementation. For example:

```
<fast-badge style="font-size: 20px"/>
```

packages/fast-components/src/badge/fixtures/base.html
eljefe223: Just to show that its possible and that the badge properly handles it

packages/fast-components/src/badge/fixtures/base.html
eljefe223: I'm just showing a badge overlaid on top of a another element. If you don't think it's relevant I can remove it.

image

packages/fast-components/src/badge/fixtures/base.html
eljefe223: Yeah, I don't know what happened there.

packages/fast-components/src/badge/badge.styles.ts
```diff
@@ -0,0 +1,29 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "../styles";
+
+export const BadgeStyles = css`
+    ${display("inline-block")} :host {
+        box-sizing: border-box;
+        font-family: var(--body-font);
+        ${/* Font size, weight, and line height are temporary - 
+            replace when adaptive typography is figured out */ ""} font-size: 12px;
+        font-weight: 400px;
```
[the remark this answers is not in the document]
eljefe223: Yeah, I copied this form Text Area

packages/fast-components/src/badge/fixtures/base.html
```diff
@@ -0,0 +1,71 @@
...
+            </fast-badge>
+        </div>
+        <div style="position: relative; display: inline-flex;">
+            <img src="https://placehold.it/140x140/DFDFDF/171717">
+            <fast-badge style="position: absolute; right: 4px; top: 4px;" fill="secondary" color="foo" id="foo">
+                App of the day
+            </fast-badge>
+        </div>
+    </div>
+    <div>
```
nicholasrice, line 49: Can you make some of these long repeated inline styles a css class in your style element so it reduces clutter?

packages/fast-components/src/badge/badge.template.ts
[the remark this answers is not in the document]
chrisdholt: @eljefe223 @EisenbergEffect are we not able to check for and merge the style object within the model itself? Setting the fill and color seems to be the primary purpose of the markup here.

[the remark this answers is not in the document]
eljefe223: We aren't at least based on conversations I had with @nicholasrice . I tried a couple different methods none seemed to work. Would be a good feature request for `fast-element` it is possible.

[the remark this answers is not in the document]
EisenbergEffect: Currently, we can't leverage host directives for `style` or `class` because it would clobber anything the customer put on the host themself. So, either we need a wrapper that we can bind to, or we need to execute code to manually add/remove classes and styles from the host.

It's possible we'll be able to have smarter class and style binding in the future. I've been thinking about it, but it hasn't bubbled up to the top of my work list yet.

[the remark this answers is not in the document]
EisenbergEffect: Note: I think we can definitely have smarter class binding and possibly have smarter style binding. The style part is harder. I think I've got some prior art lying around but haven't spent the time to research all that yet.

PR #2805, opened 2020-03-18
fix: ensure null values are removed from tabs

eljefe223:
# Description
Preact does not filter out `null` results in `children.map` added a function to filter for truthyness.

This PR is form 2018 and never got merged in:
https://github.com/preactjs/preact-compat/pull/461

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/tabs/tabs.tsx
+++ b/packages/fast-components-react-base/src/tabs/tabs.tsx
@@ -431,9 +431,11 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
      * Can remove if below gets merged in.
      * https://github.com/preactjs/preact-compat/pull/461
      */
-    private filterChildren(nodes: any): React.ReactNode {
-        if (!isNil(nodes)) {
+    private filterChildren(nodes: React.ReactNode): React.ReactNode {
+        if (Array.isArray(nodes)) {
             return nodes.filter(Boolean);
+        } else {
+            return undefined;
         }
     }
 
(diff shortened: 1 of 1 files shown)


packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -423,6 +425,17 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
               );
     }
 
+    /**
+     * Need to filter out none truthy results for Preact.
+     * Can remove if below gets merged in.
+     * https://github.com/preactjs/preact-compat/pull/461
+     */
+    private filterChildren(nodes: any): React.ReactNode {
```
nicholasrice: Can you explain why you're only doing this filter if there are props or items? Don't you only want to filter when `nodes` is an array?

eljefe223: yup your right, updated

packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -423,6 +426,19 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
...
+    /**
+     * Need to filter out none truthy results for Preact.
+     * Can remove if below gets merged in.
+     * https://github.com/preactjs/preact-compat/pull/461
+     */
+    private filterChildren(nodes: React.ReactNode): React.ReactNode {
+        if (Array.isArray(nodes)) {
+            return nodes.filter(Boolean);
+        } else {
+            return undefined;
```
nicholasrice: Sorry we had a misunderstanding - I think you just want to return `nodes` here because there is nothing to filter.

PR #2807, opened 2020-03-18
feat: wire recipes to design-system-provider

nicholasrice:
# Description

design-system driven custom properties will now be regenerated if any design system properties change.

## Motivation & context
The previous implementation of recipe properties was static. While still a stub implementation, this will allow us to re-generate recipe values when source data changes.

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/design-system-provider/design-system-provider.template.ts
```diff
@@ -1,6 +1,144 @@
...
+    neutralFillToggle,
+    neutralFillInput,
+    accentFillLarge,
+    neutralOutline,
+    neutralDividerRest,
+    neutralFocusInnerAccent,
+} from "@microsoft/fast-components-styles-msft";
 
 export const DesignSystemProviderTemplate = html<DesignSystemProvider>`
+  <template style="
```
EisenbergEffect: It is pretty cool that we can do this. It's probably not great perf-wise...but who knows? In my perf work recently, I've often been surprised by what I thought would be slow and ultimately didn't end up mattering.

nicholasrice: yep! It is super useful. This will likely have a performance hit just because it's generating *every* color regardless of use and not doing any property inheritance, but it will work well enough for the time being.

PR #2809, opened 2020-03-19
chore: update to allow contributors to run code coverage

awentzel:
# Description
GitHub Secrets should not be used for code coverage reporting if we want contributor builds to succeed.  

> Secrets are environment variables that are encrypted and only exposed to selected actions. Anyone with **collaborator** access to this repository can use these secrets in a workflow.

Code Climate has confirmed as a best practices to not use as a secret and to expose directly into our the pipeline configuration.  

> Yes, we generally recommend exposing the ID. It's a low-risk decision.
> 
> Here's one of ours: https://github.com/codeclimate/codeclimate/blob/master/.circleci/config.yml
> 
> At most, you risk having someone upload a coverage payload with your id, resulting in incorrect coverage shown on your Code Climate results. That incorrect data would be replaced with your own coverage data the next time you push a commit to your default branch.

## Motivation & context
It has been discovered that Code Climate will not run for contributors because it was stored as a secret token on GitHub.  Removing the secret and exposing directly will resolve this issue.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/.github/workflows/ci-daily.yml
+++ b/.github/workflows/ci-daily.yml
@@ -14,7 +14,7 @@ jobs:
   build_linux:
     runs-on: ubuntu-latest
     env:
-        CC_TEST_REPORTER_ID: ${{ secrets.CODE_CLIMATE_TEST_REPORTER_ID }}
+        CC_TEST_REPORTER_ID: 08a773cb4ea5811add5a45e12873e5cd2634c005568705cc37abfd5217617a32
 
     steps:
     - uses: actions/checkout@v2


.github/workflows/ci-daily.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   build_linux:
     runs-on: ubuntu-latest
     env:
-        CC_TEST_REPORTER_ID: ${{ secrets.CODE_CLIMATE_TEST_REPORTER_ID }}
+        CC_TEST_REPORTER_ID: 08a773cb4ea5811add5a45e12873e5cd2634c005568705cc37abfd5217617a32
```
nicholasrice: should this be private?

awentzel: This is covered in the description.

PR #2817, opened 2020-03-23
chore: add menu and menu item spec

chrisdholt:
# Description
This PR adds a spec for menu and menu item components.

## Issue type checklist

## Process & policy checklist


specs/menu/menu.md
```diff
@@ -0,0 +1,213 @@
...
+- role - an enum representing the menu items' role
+    - menuitem
+    - menuitemcheckbox
+    - menuitemradio
+- disabled - the menu item is disabled
+- checked - sets the checked value for menuitemcheckbox or menuitemradio items
+
+*Slots:*
+- content - slot for the content (the default slot for the item)
+- item - the optional children
```
janechu: I'm a little confused about this, it appears to be how you're determining nesting but it seems unintuitive.  Can you go into a little more detail?

specs/menu/menu.md
```diff
@@ -0,0 +1,213 @@
...
+### Anatomy and Appearance
+
+**Menu**
+
+| State | Image |
+| ----- | ----- |
+| default | ![](./images/menu.png)
+| with glyph | ![](./images/menu-glyph.png)
+
+The menu has no named slots or parts - it has a default slot for tree items.
```
janechu: tree items?

specs/menu/menu.md
chrisdholt: Can you elaborate on what is unintuitive about using a slot for creating a nested menu? I'm following a previous implementation and precedent with tree items where the default slot is for the content of the menu item and slots would be nested items.

specs/menu/menu.md
```diff
@@ -0,0 +1,213 @@
...
+    <fast-menu-item>Menu item four</fast-tree-item>
+</fast-context-menu>
+```
+
+#### Icon support:
+
+Consider a scenario where we have the following menu - some menu items have glyphs, others do not:
+```html
+<fast-menu>
+    <fast-menu-item slot="item">Menu item one</fast-menu-item>
```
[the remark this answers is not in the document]
chrisdholt: this instance was copy/paste issue. These would be default slot in the menu.

specs/menu/menu.md
chrisdholt: I see how this was a bit confusing.

specs/menu/menu.md
chrisdholt: updated

specs/menu/menu.md
```diff
@@ -0,0 +1,213 @@
+# Menu
+
+## Overview
+
+There are two components being presented as part of this specification - menu, and menu item.
```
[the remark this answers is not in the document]
chrisdholt: updated

specs/menu/menu.md
```diff
@@ -0,0 +1,210 @@
...
+<fast-menu>
+    <fast-menu-item>Menu item one</fast-menu-item>
+    <fast-menu-item>Menu item two</fast-menu-item>
+    <fast-menu-item>
+        <fancy-icon slot="icon"></fany-icon>
+        Menu item three
+    </fast-menu-item>
+</fast-context-menu>
+```
+The third item in the menu has a glyph, which would inset the text from the other menu items. To ensure that the content of this menu item aligns with the rest of the items, all items need to visually shift to ensure alignmnet of the menu content. This presents an interesting challenge for how we will communicate these changes to ensure the styling is appropriately applied. Ultimately, the parent just needs to know that *one* of the items below has a menu item with the glyph.
```
nicholasrice:
```suggestion
The third item in the menu has a glyph, which would inset the text from the other menu items. To ensure that the content of this menu item aligns with the rest of the items, all items need to visually shift to ensure alignment of the menu content. This presents an interesting challenge for how we will communicate these changes to ensure the styling is appropriately applied. Ultimately, the parent just needs to know that *one* of the items below has a menu item with the glyph.
```

specs/menu/menu.md
nicholasrice: What is the expectation? Does it expect a menu element w/ items?

specs/menu/menu.md
```diff
@@ -0,0 +1,210 @@
...
+- after-content - slot which follows content, often a glyph or icon
+
+*CSS Parts*
+- content
+- before-content
+- after-content
+- item
+
+*Events:*
+- invoke (event) - event for when the item has been clicked or invoked via keyboard
```
nicholasrice: a `<button>` element fires a "click" event on mouse click or Enter. Did you consider mapping to that behavior?

specs/menu/menu.md
chrisdholt: I think so, I don't think I'd expect individual items without a menu wrapping them.

nicholasrice: nevermind - your examples below answered that question. I think "submenu" would work, though we could also align to "menu" which might provide some consistency when we create toolbar components that create slots for *top level* menus

specs/menu/menu.md
chrisdholt: As in mapping the behavior of a menuitem to a click event? We could, though the behavior of what happens on enter/space are slightly different depending on the role.

nicholasrice: Yea reusing the "click" event name. Checkbox and radio inputs both *also* fire a click event on space / mouse click (as well as an input and change event). My thought is that since there is already precedent for the behavior we desire from native elements that we should align rather than making up new event name.

specs/menu/menu.md
```diff
@@ -0,0 +1,210 @@
...
+
+*Events*
+- none
+
+**Menu Item**
+*Component name:*
+- `fast-menu-item`
+
+*Attributes:*
+- role - an enum representing the menu items' role
```
EisenbergEffect: Do we need to build in radio and checkbox? or is there simply a slot that this or other components could be put into? Does it affect the menu item itself in any particular way?

EisenbergEffect: Do we need a menu item separator?

specs/menu/menu.md
```diff
@@ -0,0 +1,210 @@
...
+    - menuitemcheckbox
+    - menuitemradio
+- disabled - the menu item is disabled
+- checked - sets the checked value for menuitemcheckbox or menuitemradio items
+
+*Slots:*
+- content - slot for the content (the default slot for the item)
+- item - the optional children
+- before-content - slot which precedes content, often a glyph or icon
+- after-content - slot which follows content, often a glyph or icon
```
EisenbergEffect: I know we had some of this for tree-view-item as well...but to remind me: was this because we needed to align the before, content, and after in a columnar layout? If not, it seems we could just let folks put whatever they want in content.

specs/menu/menu.md
chrisdholt: The difference in behavior is spelled out when we reference the W3C keyboard interaction. I don't think we need different events though to manage it. The primary differences depend on the role of the item and state of the menu and any submenus. It seems reasonable to reuse the click event for keyboard here.

specs/menu/menu.md
chrisdholt: My thought is that it's up to users to manage their markup in some sense. The menu takes items with the roles of `menuitem`, `menuitemcheckbox`, `menuitemradio` and I think it would also take the a simple `<hr />` or div with a role of separator to manage the separation - at least that is the current working model. Menu items with a checkbox role can stand up next to standard menu items, but menu items with a radio role should be contained in a group. Menu items with checkbox and radio functionality both use aria-checked and visually they would look the same (with a checkmark). The primary difference is that in a group of items with `menuitemradio` only one can be selected.

specs/menu/menu.md
chrisdholt: We synced on this offline - I'm going to look and see if there is an abstract utility to support simply passing content and having a class create this kind of grid.

chrisdholt: I've removed references to this here for now. If we can come up with an abstract css utility to manage these I think we can do without. If not, the slots will reappear.

specs/menu/menu.md
EisenbergEffect: Maybe we have a special slotted content style for hr?

chrisdholt: yeah, this seems reasonable.

specs/menu/menu.md
chrisdholt: @EisenbergEffect @nicholasrice @janechu I think we'll need these slots. I'd like to have a quick sync sometime today to discuss my very abbreviated and short/initial test with not having these named slots.

PR #2828, opened 2020-03-24
feat: add progress web component

eljefe223:
# Description
Adds progress web component:
image

image

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/progress/progress-ring/fixtures/circular.html
```diff
@@ -0,0 +1,45 @@
+<fast-design-system-provider style="font-family: var(--body-font)">
+    <h1>Progress ring</h1>
+    <h4>Default</h4>
+    <fast-progress-ring min=0 max=100 value=75></fast-progress-ring>
```
janechu: should these not have quotes surrounding the values? eg. `min="0"`

packages/fast-components/src/progress/progress-ring/progress-ring.styles.ts
```diff
@@ -0,0 +1,72 @@
...
+        stroke: var(--accent-fill-rest);
+        fill: none;
+        stroke-width: 2px;
+        stroke-linecap: round;
+        transform-origin: 50% 50%;
+        transform: rotate(-90deg);
+        transition: all 0.2s ease-in-out;
+    }
+
+    .indeterminate-indicator-1 {
```
janechu: should we use the "-1" on "indeterminate-indicator-1" if there is only one?

packages/fast-components/src/progress/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./base-progress";
```
janechu: should we export this?

packages/fast-components/src/progress/progress-ring/progress-ring.styles.ts
eljefe223: i did this to keep the structure similar across each progress component and not have duplicated part names.

packages/fast-components/src/progress/index.ts
eljefe223: Is there a reason not to? I don't see the harm in doing so. This way someone could build their own custom element off of base progress.

packages/fast-components/src/progress/base-progress.ts
```diff
@@ -0,0 +1,18 @@
+import { attr, FastElement } from "@microsoft/fast-element";
+
+export class BaseProgress extends FastElement {
+    @attr
+    public value: number;
```
EisenbergEffect: Before we merge, we could use the new attr converter feature to set these up with a number converter, ensuring that the values are always numbers (and not strings) when coming from HTML.

packages/fast-components/src/progress/base-progress.ts
```diff
@@ -0,0 +1,18 @@
...
+
+    @attr
+    public min: number;
+
+    @attr
+    public max: number;
+
+    @attr
+    public paused: boolean;
```
EisenbergEffect: We could set this to use the new boolean attr behavior as well. Currently this will have a bug if paused is the string value "false", for example. This feature isn't merged yet, but should be merging shortly.

EisenbergEffect: If we don't want it to be a boolean attr, we can independently leverage the boolean converter to ensure the type is correct when coming from HTML as well.

packages/fast-components/src/progress/progress-ring/progress-ring.template.ts
```diff
@@ -0,0 +1,34 @@
...
+        $aria-valuemin="${x => x.min}"
+        $aria-valuemax="${x => x.max}"
+    >
+        ${when(
+            x => x.value,
+            html<BaseProgress>`
+            <svg class="progress" part="progress" viewBox="0 0 16 16" slot="determinate">
+                <circle class="background" part="background" cx="8px" cy="8px" r="7px"></circle>
+                <circle class="determinate" part="determinate" style="stroke-dasharray: ${x =>
+                    (44 * x.value) / 100}px 44px" cx="8px" cy="8px" r="7px"></circle>
```
EisenbergEffect: Nice!

packages/fast-components/src/progress/progress-ring/progress-ring.styles.ts
```diff
@@ -0,0 +1,72 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "../../styles";
+
+export const ProgressRingStyles = css`
+    :host([hidden]) {
```
nicholasrice: you don't need the hidden here - that is handled by `display("flex")`

packages/fast-components/src/progress/progress-ring/progress-ring.styles.ts
```diff
@@ -0,0 +1,68 @@
...
+        stroke: var(--neutral-fill-rest);
+    }
+
+    :host(.paused) .determinate {
+        stroke: var(--neutral-foreground-hint);
+    }
+
+    @keyframes spin-infinite {
+        0% {
+            stroke-dasharray: 0.01px 43.97px;
```
nicholasrice: I just put in some work to enable high-contrast scenarios. Can you make sure this does the right think in high-contrast mode?

nicholasrice: Tested - HC looks good!

Issue #2742, opened 2020-03-04
Enable Boolean Attributes and Type Conversion for FastElement


nicholasrice:
### HTML Standard 2.4.2

A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.

The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.

[Link](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes)

### Current Behavior in FastElement

The `@attr` decorator has some un-predictable behavior. Take the following:

```TypeScript
@attr
required: boolean = false;
```

For some element *implementing* that property, invoking `element.setAttribute("required", "")` will add the attribute as expected, but will also assign the IDL property to the value `""`. Likewise, invoking `element.removeAttribute("required")` assigns the IDL attribute to the value `null`.

Additionally, setting the property to true or false does not update the attribute by adding or removing it.

### Proposed Solution

Let's add a `mode` property to the configuration for attributes, supporting the following values:

* `reflect` - Reflects properties to the attribute as is.
* `boolean` - Reflects properties as boolean attributes.
* `none` - Does not reflect properties.

By default, the mode will be `reflect`. 

**Example 1: The Default Reflect Behavior**

```TypeScript
class MyButton extends FastElement {
  @attr foo;
}
```

**Example 2: Boolean Attribute**

```TypeScript
class MyButton extends FastElement {
  @attr({ mode: 'boolean' }) foo;
}
```

## Type Conversion

In HTML, all attribute values are strings. However, in IDL they can be other primitive types. For example, an attribute might have the value "5" but the IDL should have the number 5. Or, an attribute might have the value "" which should be interpreted as IDL true, or null, which should be interpreted as IDL false (the case with boolean attributes).

### Proposed Solution

Introduce an interface `ValueConverter` as follows:

```TypeScript
interface ValueConverter {
  toView(value: any): string;
  fromView(value: string): any;
}
```

Extend the `attr` configuration with a `converter` property  to allow providing a converter.

**Example 3: Custom Converter**

```TypeScript
class MyButton extends FastElement {
  @attr({ converter: numberConverter }) foo: number = 5;
}
```

[...]

PR #2829, opened 2020-03-24
feat(fast-element): boolean attributes and type conversion

EisenbergEffect:
# Description

This PR adds support for type conversion of attribute values as well as boolean attribute behavior.

## Motivation & context

Specifically, this PR implements #2742 by adding new APIs to the AttributeDefinition class, which is now used by attribute getters and setters to change values. The APIs are also used to respond to the onAttributeChangedCallback.

The feature enables three modes (`reflect`, `boolean`, `none`) per the above issue, and also enables a custom converter, per the above issue. A `booleanConverter` is provided out of the box and is automatically set when in `boolean` mode.

Here's how one would create a custom element with a boolean attribute behavior:

```TypeScript
@customElement(...)
export class Switch extends FastElement {
  @attr({ mode: 'boolean' }) disabled = false;
}
```

This PR also updates change handlers for attr and observable to pass the old and new values.

## Issue type checklist

## Process & policy checklist

> **NOTE:** I'll be updating the documentation on attributes to explain these new capabilities. I'll add it as an additional commit to this PR and update the checklist above, but have not done this work yet. That should not block review of the code in this PR.

### DIFF
--- a/packages/fast-element/src/attributes.ts
+++ b/packages/fast-element/src/attributes.ts
@@ -1,20 +1,138 @@
+import { Observable } from "./observation/observable";
+
+export interface ValueConverter {
+    toView(value: any): string;
+    fromView(value: string): any;
+}
+
+export type AttributeMode = "reflect" | "boolean" | "none";
+
 export type AttributeConfiguration = {
     property: string;
     attribute?: string;
+    mode?: AttributeMode;
+    converter?: ValueConverter;
 };
 
 export type DecoratorAttributeConfiguration = Omit<AttributeConfiguration, "property">;
 
+export const booleanConverter: ValueConverter = {
+    toView(value: any): string {
+        return value ? "true" : "false";
+    },
+
+    fromView(value: any): any {
+        if (
+            value === null ||
+            value === void 0 ||
+            value === "" ||
+            value === "false" ||
+            value === false ||
+            value === 0
+        ) {
+            return false;
+        }
+
+        return true;
+    },
+};
+
 export class AttributeDefinition {
+    private readonly fieldName: string;
+    private readonly callbackName: string;
+    private readonly hasCallback: boolean;
+    private readonly guards = new Set();
+
     public constructor(
+        public readonly Owner: Function,
         public readonly property: string,
-        public readonly attribute: string = property.toLowerCase()
-    ) {}
+        public readonly attribute: string = property.toLowerCase(),
+        public readonly mode: AttributeMode = "reflect",
+        public readonly converter?: ValueConverter
+    ) {
+        this.fieldName = `_${property}`;
+        this.callbackName = `${property}Changed`;
+        this.hasCallback = this.callbackName in Owner.prototype;
+
+        if (mode === "boolean" && converter === void 0) {
+            this.converter = booleanConverter;
+        }
+    }
+
+    public setValue(object: any, newValue: any) {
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
@nicholasrice With this implementation, to implement the design system decorator, you will have your decorator take the same config as the `attr` decorator, plus any additional properties you want. Your decorator will then simply call the `attr` decorator manually internally. Before calling that `attr` decorator, you will add a callback method to the prototype or wrap an existing callback so that you can run the custom design system property change behavior.

packages/fast-element/src/attributes.ts
```diff
@@ -1,21 +1,139 @@
...
 };
 
 export type DecoratorAttributeConfiguration = Omit<AttributeConfiguration, "property">;
 
+export const booleanConverter: ValueConverter = {
+    toView(value: any): string {
+        return value ? "true" : "false";
+    },
+
+    fromView(value: any): any {
```
nicholasrice: HTML5 allows for boolean attrbiutes to be set to the *attribute name*: https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes

eg `disabled="disabled"` 

Does this converter cover that case?

nicholasrice:
@EisenbergEffect awesome thanks for adding that extension feature!

packages/fast-element/src/attributes.ts
EisenbergEffect: No it doesn't cover that scenario. Is there some advantage to doing it that way?

EisenbergEffect: Well, technically, the way the converter is implemented, it will interpret any string besides "false"  and empty string as being a true value. Is that ok?

janechu: Well, if we followed what could be true, empty string would actually be true. I'm not sure how much we mind but `<input disabled />` is `<input disabled="disabled" />` is `<input disabled="" />`.

EisenbergEffect: Merged this. If there's a big issue with the above behavior, I can make a future change.

EisenbergEffect: Currently, empty string will result in false in the way the converter is implemented.

nicholasrice: I don't think there is an advantage to setting `disabled="disabled"` - was more commenting because it is a native language feature that we might want to mimic.

Here is a little pen of how button behaves: https://codepen.io/rice119/pen/oNXwqpP

It seems if the attribute exists then the IDL is true, otherwise it is false.

EisenbergEffect: Hmm. Minus the empty string scenario, our implementation works just like this. I need to play with how web components report the attribute changed value when the attribute is removed. If it is reported as an empty string, for example, then that would cause us a problem in terms of matching implementation exactly. If it returns null or undefined, then we can make this match with no trouble. I've got another PR coming with the docs I forgot to add with this. If I can make this change, I'll include it in that PR.

nicholasrice: awesome thanks @EisenbergEffect!

EisenbergEffect: Ok, I can fix this 😄 New PR incoming with the fix and the docs. Thanks for digging in on that and catching the inconsistency!

PR #2835, opened 2020-03-25
fix: horizontal overflow - add initial value to reduce fn to avoid throw

scomea:
# Description
We were using a reduce function without an inial value so it could throw on an empty array.  Added an initial value of 0.

## Motivation & context
Avoid throwing error

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
+++ b/packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
@@ -440,7 +440,10 @@ class HorizontalOverflow extends Foundation<
     private isOverflow(): boolean {
         const availableWidth: number = this.getAvailableWidth();
         const itemWidths: number[] = this.getItemWidths();
-        const totalItemWidth: number = itemWidths.reduce((a: number, b: number) => a + b);
+        const totalItemWidth: number = itemWidths.reduce(
+            (a: number, b: number) => a + b,
+            0
+        );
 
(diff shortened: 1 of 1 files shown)


packages/fast-components-react-base/src/horizontal-overflow/horizontal-overflow.tsx
```diff
@@ -440,7 +440,10 @@ class HorizontalOverflow extends Foundation<
     private isOverflow(): boolean {
         const availableWidth: number = this.getAvailableWidth();
         const itemWidths: number[] = this.getItemWidths();
-        const totalItemWidth: number = itemWidths.reduce((a: number, b: number) => a + b);
+        const totalItemWidth: number = itemWidths.reduce(
+            (a: number, b: number) => a + b,
+            0
+        );
```
chrisdholt: Should we add a test for this?

scomea: Not sure how to fake it "for realz", but I could add a direct function test with staged values.

chrisdholt: I wouldn't worry about it here as now we handle the empty array. I guess we could say, "empty array shouldn't throw"?

scomea: added

PR #2836, opened 2020-03-26
feat: add a utility for validation as a service using ajv

### LINKED ISSUE #2823: Create a utility to transform ajv error messages and send them to the message system
# Background

After the completion of #2822 a mapping utility must be written to convert/transform error messages generated from `ajv` to the new interface.

## Requirements

- Create a utility in `@microsoft/fast-tooling` that will subscribe to the message system and during data updates will update validation using the message system

janechu:
# Description

Creates a utility for ajv error messages to be transformed and set to the message system. These are intended to be used by the `Form` so that any validation method can be used as long as it submits the validation via the message system.

## Motivation & context

Resolves #2823 

## Issue type checklist

## Process & policy checklist


packages/fast-tooling/package.json
```diff
@@ -74,7 +74,8 @@
         "typescript": "^3.3.3"
     },
     "peerDependencies": {
-        "lodash-es": "^4.0.0"
+        "lodash-es": "^4.0.0",
+        "ajv": "^6.10.0"
```
janechu: So, I think making ajv a dependency here is actually an issue, I'm considering adding another package just for this utility called `@microsoft/fast-tooling-ajv-validator` what do people think?

nicholasrice: What is the concern?

janechu: That we have an additional dependency added to the package. We're arbitrarily deciding to use `ajv` as the validator here, but since this is a stand-alone utility, where we are inviting others to create their own. It might make more sense to have this exist as part of another package, but it could also be tree shaken.

janechu: I've moved it from peer to dependency.

janechu: That will create the opportunity for users to tree-shake if they decide to use it and be less cumbersome.

PR #2842, opened 2020-03-26
fix: udpate text tield as form connected

chrisdholt:
# Description
The initial text field work was unfortunately not being exposed as a form element. This PR updates the control to be a form associated element.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/text-field/text-field.ts
+++ b/packages/fast-components/src/text-field/text-field.ts
@@ -1,4 +1,5 @@
-import { attr, FastElement } from "@microsoft/fast-element";
+import { attr } from "@microsoft/fast-element";
+import { FormAssociated } from "../form-associated";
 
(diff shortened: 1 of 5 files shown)


packages/fast-components/src/text-field/text-field.ts
```diff
@@ -22,61 +23,127 @@ export class TextField extends FastElement {
...
-        this.required
-            ? this.classList.add("required")
-            : this.classList.remove("required");
-    }
-
     @attr({ mode: "boolean" })
-    public readonly: boolean;
-    private readonlyChanged(): void {
-        this.readonly
+    @attr({ attribute: "readonly", mode: "boolean" })
```
EisenbergEffect: If you don't provide an attribute name, it will be derived from lowercasing the property name. So, technically you don't need to specify this one.

Issue #2834, opened 2020-03-25
[Proposal] FastElement Property and Attribute Binding Changes


EisenbergEffect:
## Description

As the features and usage of FastElement have advanced, some early decisions related to the templating syntax and behavior of property and attribute bindings seem less correct. I'm proposing that we make a breaking change to how bindings work in order to create greater consistency, reduce library size, and enable critical scenarios.

## Today

Today we have the following syntax:

* `attr=${x => x.value}` - Uses the expression to update the value of the *property* associated with the `attr` via `element.attr = value`.
* `$attr=${x => x.value}` - Uses the expression to update the value of the *attribute* via `element.setAttribute('attr', value);`
* `?attr=${x => x.value}` - Uses the expression to update the value of the *attribute* using boolean attribute behavior.

### Problems

* Some have found that plain `attr=` behavior to be unintuitive. They expect this to set the attribute, rather than the property.
* Some custom element attributes names don't work correctly with `attr=`. For example, if the attribute is `my-attribute`, the binding `my-expression=${x => x.value}` will result in an attempt to set the `my-expression` property, which doesn't exist. Most likely the property name is `myExpression`. There's no way for the template compiler to know this. All attr name conversion is based on knowledge of built-ins. So, there's no way to perform a property binding in this case. An explicit attribute binding is needed. This is unintuitive. 
* If an element exposes some settings via properties with no attribute, then there's no way for the binding engine to set those if the property has mixed case property names.

## Proposal

* `attr=${x => x.value}` - Uses the expression to update the value of the *attribute* via `element.setAttribute('attr', value);`
* `:attr=${x => x.value}` - Uses the expression to update the value of the *property* via `element.attr = value`.
* `?attr=${x => x.value}` - Uses the expression to update the value of the *attribute* using boolean attribute behavior.

1. I propose that a binding with no prefix should set the attribute with no name alteration or mapping. It should be set exactly as specified using the `setAttribute` API.
2. I propose that the prefix `:` designate a property binding and that we add support for mixed-case property names.
3. I propose that we remove the `$` prefix, as it will no longer be needed.
4. I propose that we remove all attribute to property name conversion code, as it will no longer be needed.

**Note:** The `:` syntax was chosen for property binding because it is used in other popular libraries like `Vue` and `Aurelia 2`. This will be familiar to many people who may adopt the library.

I believe these changes will be more intuitive as well as fix the shortcomings with attributes that don't match property names, and mixed case property names. 

## Level of Effort

[...]

PR #2843, opened 2020-03-26
feat(fast-element): attribute/property binding revamp

EisenbergEffect:
# Description

This PR makes attribute the default binding mode and removes the need for the `$` prefix for this case. Property binding now requires a `:` prefix. Both property and event bindings now support mixed case names.

## Motivation & context

For the full background of these changes, see issue #2834.

In addition to what is described here, a small improvement was made to event bindings. These bindings call `preventDefault()` by default, which is what is typically desired. With this PR, a developer can now opt-out of this by returning `true` from their handling function.

Finally, I made a change to the new modes code, that changes mode "none" to "fromView" which is more accurate to what is actually happening. A "none" mode would be accomplished by using `@observable` instead of `@attr`.

## Issue type checklist

This is a major breaking change to the templating syntax and behavior. All existing components will need to be updated in order to function properly.

## Process & policy checklist


packages/fast-element/docs/building-components.md
```diff
@@ -261,37 +261,45 @@ You can also use an expression to set a property on an HTML element. Simply plac
...
 
-**Example: Boolean Attributes**
+**Example: Inner HTML**
 
 ```HTML
-<button type="submit" ?disabled="${x => !x.enabled}">Submit</button>
+<div :innerHTML=${x => sanitize(x.someDangerousHTMLContent)}></div>
 ```
 
+> **WARNING:** Avoid scenarios that require you to directly set HTML, especially when the content is coming from an external source. If you must do this, always sanitize the HTML content using a robust HTML sanitizer library, represented by the use of the `sanitize` function above.
```
EisenbergEffect: FYI I'm looking into taking advantage of new Trusted Types APIs so that anytime someone tries to bind the innerHTML property, we force the update to go through our HTML policy (or one they provide). Our policy would throw by default, forcing end users to provide a sanitization policy.

EisenbergEffect:
@chrisdholt I've got enough approval to merge now. Doing so will break all our components. How should we proceed? Should we have component authors commit to this branch to adjust their components and then merge it all at once? Or just merge now, let it break, and fix the components separately?

chrisdholt:
> @chrisdholt I've got enough approval to merge now. Doing so will break all our components. How should we proceed? Should we have component authors commit to this branch to adjust their components and then merge it all at once? Or just merge now, let it break, and fix the components separately?

Since we haven't published, let's just merge and break. In the future we can do this in a feature branch and version - for now, I think we can just take a pass to fix once this is in.

PR #2845, opened 2020-03-27
feat: allow custom messages to be sent to the message system which will return them

janechu:
# Description

This allows any arbitrary message to be sent via a custom message which can contain any content sufficing that it includes the "custom" type. This will allow components opted in to the message system to pass arbitrary messages that the message system may not need to store or act on.

## Motivation & context

There is a requirement to validate a oneOf if data is provided inside the `Form` component so as to show which oneOf item is selected. This should be done via the validation system, however the message system does not need to know the outcome, therefore a custom message is required to allow the `Form` and the validation to communicate separately.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling/src/message-system/message-system.utilities.props.ts
+++ b/packages/fast-tooling/src/message-system/message-system.utilities.props.ts
@@ -401,8 +401,6 @@ export interface GetNavigationMessageOutgoing {
     navigation: NavigationConfig;
 }
 
-export type CustomMessageContent<T> = { [P in keyof T]?: T[P] };
-
 export interface CustomMessageIncomingOutgoing {
     type: MessageSystemType.custom;
 }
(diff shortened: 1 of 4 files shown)


packages/fast-tooling/src/message-system/message-system.utilities.props.ts
```diff
@@ -401,6 +401,17 @@ export interface GetNavigationMessageOutgoing {
     navigation: NavigationConfig;
 }
 
+export type CustomMessageContent<T> = { [P in keyof T]?: T[P] };
```
nicholasrice: Isn't this just `Partial`?

janechu: Whoops, yes. Fixed.

PR #2851, opened 2020-03-28
chore: add carousel spec

SethDonohue:
# Description

Adds web component spec for Carousel component

## Issue type checklist

## Process & policy checklist


specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+            slot="play-toggle"
+            @click="${x => x.togglePlay()}"
+            class="previous-flipper"
+        >
+            <svg class="play-svg">
+            <path d="example path..."></path>
+            </svg>
+        </button>
+
+        <div role="tablist" class="tablist" aria-label="A carousel of items">
```
eljefe223: I just started looking into a web component version of Tabs might be good to sync up here.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+
+*Slots:*
+- `default` - slot for slides content // TODO: IS this the right way?
+- `play-toggle` - the control to toggle if the slides rotate or not
+- `next-flipper` - the control used to rotate to the next slide
+- `previous-flipper` - the control used to rotate to the previous slide
+- `tablist` - the control to select a certain slide
+
+*Events*
+- `on-pause` - callback fired when the rotation is stopped, such as when keyboard focus or hover is on any element of the carousel, or the *play-toggle* is set to off.
```
nicholasrice: we typically don't use the "on-" convention for event names; I think the first would just be "pause". The [ARIA carousel spec](https://www.w3.org/TR/wai-aria-practices/#carousel) uses the term "rotate" a lot - thoughts on using "rotate" for the event name instead of "active-id-update"?

nicholasrice: We might also want a "play" event also? Or maybe this is all surfaced under a general "autorotate" event?

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+                class="carousel-item ${x => x.getActiveIndex() === index ? 'active' : ''}"
+                role="group"
+                aria-roledescription="slide"
+                aria-label="${x => x.getActiveIndex()} of ${x => x.getItemCount()}"
+            >
+                ${item}
+            </div>`
+            }).join('')
+        }
+        <!-- Solution 2 -->
```
nicholasrice: Only having one item in the DOM might make slide transitions difficult or impossible to achieve.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+        </div>
+    </div>
+    <slot name="previous-flipper"> SEE DEFAULT </slot>
+    <slot name="tab-list">SEE DEFAULT </slot>
+    </div>
+</template>
+```
+
+**Default Flipper**
+```html
```
nicholasrice: Having default buttons poses an interesting question. These elements will need localized text labels to communicate to AT what they are for. Typically we don't bake localized strings into elements though.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+    <div tabindex="-1" role="tab" class="tablist-tab"></div>
+    <div tabindex="-1" role="tab" class="tablist-tab"></div>
+</div>
+```
+
+**Implementation**
+```html
+<body>
+    <fast-carousel
+        // TODO: CHECK is this a way to apply the items?
```
nicholasrice: One option for providing the items would be to use slots. `` The control can then determine what to do with all of the slotted content.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+        item-0='<img src="...image1.src">'
+        item-1='<img src="...image2.src">'
+        item-2='<img src="...image3.src">'
+    >
+        <button
+            slot="next-flipper"
+            @click="${x => x.next()}" // TODO: How do I attach the carousel.next() function to a slotted item?
```
nicholasrice: You could use event delegation for this. The host element would listen for click events and then dispatch an action based on which element was clicked. For example, if the composed path comes from an element that has the `[slot="next-flipper"]` attribute then we do work to render the next slide.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+
+---
+
+## Implementation
+
+### States
+
+*Key component states, valid state transitions, and how interactions trigger a state transition.*
+
+- `isPaused` - if the carousel slide rotation is paused
```
nicholasrice: you could probably shorten this to just `paused` because it is a boolean.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+- `activeId` - the id of the active slide
```
nicholasrice: It might be useful to just store a reference to the active slide element itself.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+
+The carousel should align to the design pattern and interaction model provided by the W3C: https://w3c.github.io/aria-practices/#carousel
+
+### Globalization
+
+The component may have to change the order of rotation for slides as an RTL users sequence expectations is reversed from LTR. This would require that the flippers and tablist be functionally reversed. The tablist would also have to be visually reversed with the first tab item starting on the right of the tablist.
+
+### Test Plan
+
+
```
nicholasrice: See note about using slots to provide the items regarding this perf comment.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+*Component name:*
+- `fast-carousel`
+
+*Attributes:*
+- `autoplay` - boolean value to toggle if the carousel plays on load or not (defaults to false)
+- `autoplay-interval` - a number representing the autoplay interval in milliseconds
+- `loop` - boolean value to toggle if the carousel loops slides at the beginning or end (defaults to true)
+- `paused` - whether the rotation is paused or not
+- `aria-roledescription` - set to *carousel* for overall component, and set to *slide* for each slide
```
eljefe223: This shounds like it does not need to be an attribute if it always needs to be set to carousel or slide

specs/carousel/carousel.md
```diff
@@ -0,0 +1,258 @@
...
+>
+    <svg class="play-svg">
+        <path d="example path..."></path>
+    </svg>
+</button>
+```
+
+**Default Tab List**
+```html
+<div role="tablist" class="tablist" aria-label="A carousel of items">
```
eljefe223: We will want to use whatever we decide for `fast-tabs` here. I just started looking into that spec. Maybe we should sync up on this.

chrisdholt: I think a sync might be helpful here, but I'm not 100% sure we need to map to `tabs`. There is a working draft for carousel in the W3C as a control: https://w3c.github.io/aria-practices/#carousel

I do have some thoughts on the above spec though as it seems to include several interaction models that conflict and could make implementing difficult or confusing for end users.

specs/carousel/carousel.md
SethDonohue: Yeah I need to put much more thought into this part. I like your ideas and will add them in.

specs/carousel/carousel.md
SethDonohue: Yes, a typical carousel has a few so I'll look at updating this.

specs/carousel/carousel.md
SethDonohue: Good thoughts. I assume we want implementors/devs to be able to use their own flippers, so I'll add this concern to the list and see if I can come up with something.

chrisdholt: I think one way to approach this is how we default the control. For instance, the default carousel would like quite a bit like our current implementation with sequence indicators and flippers which are hidden from AT. This way, the default flippers are hidden from AT by design. To unhide them you would need to pass your own flippers. We could also easily ensure that buttons match the styling by using the slotted pseudo selector to assign the styles to any button passed - `::slotted(button)`. I think there's a lot of room to creatively solve this.

specs/carousel/carousel.md
SethDonohue: Yes, I saw that the old one uses tabs, but feel like it may not be necessary. I'll address some of these other issues and then sync with you both on the above.

specs/carousel/carousel.md
SethDonohue: Updated Solution 2 to have 3 slides

specs/carousel/carousel.md
SethDonohue: Like using a querySelector([slot="play-toggle"]) to grab all the items. Sounds good!

specs/carousel/carousel.md
SethDonohue: Updated to have the click handler on the host element.

specs/carousel/carousel.md
SethDonohue: Added a reference prop

specs/carousel/carousel.md
SethDonohue: Do we want buttons to have a certain styling, or do we want the author/implementor to pass all that in?
Maybe a sync on this idea and others is a good idea. I'll get that going soon

specs/carousel/carousel.md
SethDonohue: Updated to use "item" slot and then the controller can handle all the items given.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,282 @@
...
+
+** See the [W3C Specification](https://w3c.github.io/aria-practices/#wai-aria-roles-states-and-properties-4) for requirements and details.
+
+*CSS custom properties:*
+- ???
+
+*Slots:*
+- `default` - slot for slides content // TODO: IS this the right way?
+- `play-toggle` - the control to toggle if the slides rotate or not
+- `next-flipper` - the control used to rotate to the next slide
```
nicholasrice, line 60: I kind of think "flipper" isn't the right term here. That is the name of the control *we* use for this, but it's really just a button. Thoughts on changing this to somthing like `next-button` / `previous-button`?

specs/carousel/carousel.md
SethDonohue: @chrisdholt Does your fast-flipper address all of this? I took the defaults out for the spec and actually have the **Implementation** using fast-flipper.

chrisdholt: There's likely a couple ways to handle it:

1) Take a dependency on Flipper.

2) Support default buttons in the slot which use the fast-flipper styles. Doing this would allow reuse if someone wanted to pass the fast-flipper themselves and handle telemetry, etc.

3) Some other way?

SethDonohue: > @chrisdholt Does your fast-flipper address all of this? I took the defaults out for the spec and actually have the **Implementation** using fast-flipper.

Correction: I added fast-flipper as the default for the next-button and previous-button slots in the template.

specs/carousel/carousel.md
SethDonohue: I have added a simple tablist example in the **implementation** of the spec.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,228 @@
...
+            <div tabindex="0" role="tab" class="tab-list-tab"></div>
+            <div tabindex="-1" role="tab" class="tab-list-tab"></div>
+            <div tabindex="-1" role="tab" class="tab-list-tab"></div>
+            <div tabindex="-1" role="tab" class="tab-list-tab"></div>
+            <div tabindex="-1" role="tab" class="tab-list-tab"></div>
+        </div>
+    >
+    </fast-carousel>
+</body>
+```
```
EisenbergEffect, line 177: It doesn't look like we're using the default slot. Can we put the items in the default slot?

Also, is it possible to generate the tab-list if one isn't provided? This would need to be kept in sync with the actual slotted items, but that's not too hard (I've got an idea for a library feature to help with this). These could be generated as fallback content for the tab-list slot.

SethDonohue: Changed to use the default.

Yeah I think we are still discussing how tab-list will be used per the W3 Carousel spec as there are some different interaction models that could be used.
I still assume it'd be best to have a default. What's your idea for the library feature to help?

EisenbergEffect: I'm thinking of having a `@slotted` decorator that would give you the list of html elements that exist in a particular slot, keeping it up to date automatically as slotted content changes:

```TypeScript
export class FastCarousel extends FastElement {
  @slotted items: HTMLElement[]; // all the html elements in the default slot 
}
````

With this in place you could repeater over `items` and generate the tab list into the default slot. Any time the slotted items change, the repeat would automatically update the tab list for you.

The `@slotted` decorator would take an optional slot name and an optional element filter for more advances scenarios. It would default to the default slot and a filter that selected all nodes.

Comment (EisenbergEffect) [reply specs/carousel/carousel.md:177]: Another option is to do something like ref where you put a directive on your slot and it gives you a property with its content:

```
const template = html`
  <slot ${slotted('items')}></slot>
`;

specs/carousel/carousel.md
```diff
@@ -0,0 +1,228 @@
...
+                </div>
+                <div
+                    class="carousel-item-next"
+                    role="group"
+                    aria-roledescription="slide"
+                    aria-label="${x.getNextIndex} of ${x.getItemCount}"
+                >
+                    ${x.getNextItem()}
+                </div>`
+        }
```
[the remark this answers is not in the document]
EisenbergEffect: I don't think the code you have above is going to work. I'm not quite following why you have the nested lambda expression there.

I'd probably go with rendering all the items and using the classes to activate them. We do have a new directive called `compose` that's planned to handle more advanced dynamic and conditional rendering of different templates.

specs/carousel/carousel.md
SethDonohue: To be clear the items are the slide content, and in this template the tab-list is for the slide picker controls. Maybe I should name that something different that tab-list?

SethDonohue: > I'm thinking of having a `@slotted` decorator that would give you the list of html elements that exist in a particular slot, keeping it up to date automatically as slotted content changes:
> 
> ```ts
> export class FastCarousel extends FastElement {
>   @slotted items: HTMLElement[]; // all the html elements in the default slot 
> }
> ```
> 
> With this in place you could repeater over `items` and generate the tab list into the fallback content for tab list. Any time the slotted items change, the repeat would automatically update the tab list for you.
> 
> The `@slotted` decorator would take an optional slot name and an optional element filter for more advances scenarios. It would default to the default slot and a filter that selected all nodes.

I think this would be handy for what I am trying to do. I am currently having difficulty getting the nodes/html elements that are in the default slot, and then using them to render the slides, but I believe I am missing a key concept.

EisenbergEffect: You might be accessing assignedNodes to early and need to also register for the slotchanged event. What I'm proposing would do all that work for you so you just have a nice array of nodes in the end, which you could repeat over to generate other UI. I can probably get to to work on this Monday or Tuesday of next week if this sounds good.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,228 @@
...
+- `autoplay-interval` - a number representing the autoplay interval in milliseconds
+- `loop` - boolean value to toggle if the carousel loops slides at the beginning or end (defaults to true)
+- `aria-labelledby` - optional based on implementation**
+- `aria-label` - optional, based on implementation**
+- `paused` - whether the rotation is paused or not
+
+** See the [W3C Specification](https://w3c.github.io/aria-practices/#wai-aria-roles-states-and-properties-4) for requirements and details.
+
+*CSS custom properties:*
+- ???
```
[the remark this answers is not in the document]
nicholasrice: If we don't need any then we can just omit this section

specs/carousel/carousel.md
SethDonohue: Sorry on the delay, this definitely sounds good.

specs/carousel/carousel.md
```diff
@@ -0,0 +1,200 @@
...
+---
+
+## Design
+
+### API
+
+*Component name:*
+- `fast-carousel`
+
+*Attributes & Properties:*
```
scomea, line 40: Should we expose the current slide index?  And somehow enable authors to navigate to a specific slide somehow?

specs/carousel/carousel.md
```diff
@@ -0,0 +1,206 @@
...
+
+*Slots:*
+- `default` - use default slot for slide content
+- `play-toggle` - the control to toggle if the slides rotate or not
+- `next-button` - the control used to rotate to the next slide
+- `previous-button` - the control used to rotate to the previous slide
+- `tab-list` - the control to select a certain slide
+
+*Events*
+- `change` - callback fired when the slide changes. This updates the `activeslideid` and passes a reference to the active slide. The play/pause state and `activeslideid` can then be read by the author from the reference element, e.g. event.target.paused, event.target.activeslideid
```
eljefe223, line 64: In tabs we pass an reference to the `activetab` and then allow the app author to set the `activetabid`

specs/carousel/carousel.md
```diff
@@ -0,0 +1,206 @@
...
+Notes:
+- Since the host element has the handleClick(), it would use event delegation to check what element was clicked on. Example: If the target was an element with a `[slot="next-button"]` it would trigger the controller method for nextSlide(). This way each element passed in via slot would not needs it's own onClick.
+
+```html
+<!-- shadow dom -->
+<div
+    class="carousel"
+    aria-roledescription="carousel"
+    tabindex="-1"
+    @click="${x => x.handleClick()}"
```
eljefe223, line 78: If this is the shadow dom we wouldn't see the click handler here.

EisenbergEffect: I think this will be solved by attaching the click handler to the host itself. That should let it see the clicks from the light dom child nodes. Good catch @eljefe223 !

specs/carousel/carousel.md
EisenbergEffect: FYI, the new `slotted` directive is in. Right now you'll have to filter the nodes to ensure you only get element nodes, which is a pain. I'll see if I can add the filter feature in soon.

Issue #2850, opened 2020-03-27
[Proposal] innerHTML Security


EisenbergEffect:
## Description

In FastElement, it's possible to bind to `innerHTML` using the following template syntax:

```HTML
<div :innerHTML=${x => x.someDangerousHTMLContent}></div>
```

This opens up the door to various security issues. We can rely on developers to do the right thing, by sanitizing input, as strongly suggested in our documentation:

```HTML
<div :innerHTML=${x => sanitize(x.someDangerousHTMLContent)}></div>
```

However, this still relies on developers to know that they need to do this, and it's quite possible that they will make a mistake or forget. To address this issue, I suggest that we detect bindings against `innerHTML` during template compilation, and wrap the expression that retrieves the content with the new TrustedTypes policy feature. By default, the FastElement library would create a policy that throws an error, requesting that customers configure a real policy with their sanitization rules. In this way, we can guarantee that XSS attacks are not possible through `innerHTML` bindings and we also can hook into the browser's CSP features (which allow policy names to be locked down).

## Examples

Looking back at the first, problematic example:

```HTML
<div :innerHTML=${x => x.someDangerousHTMLContent}></div>
```

During template compilation, the expression would be transformed from this:

```TypeScript
x => x.someDangerousHTMLContent
```

To this:

```TypeScript
x => fastHTMLPolicy.createHTML(x.someDangerousHTMLContent)
```

By making this transform, we ensure that all html bindings go through our policy. Again, by default, the policy would throw, requesting the developer to configure a real policy.

To configure a policy, we'll Likely have an API like this:

```TypeScript
const htmlPolicy = trustedTypes.createPolicy('my-policy', {
  createHTML: (html) => {
    // TODO: call some sanitization library on the html
    return html;
  }
});

BindingDirective.configureHTMLPolicy(htmlPolicy);
```

Under the hood, the BindingDirective would have a default implementation like this:

```TypeScript
let fastHTMLPolicy = trustedTypes.createPolicy('fast-html-policy', {
  createHTML: (html) => {
    throw new Error('Please configure an HTML policy..blah..blah..');
  }
});
```

## Dependencies

Implementing the feature in this way relies on the TrustedTypes feature being available in browsers. It is not yet ready, but is behind a flag in Chromium. However, a tiny polyfill, which enables the API to work can be accomplished with this code:

```TypeScript
if(typeof trustedTypes == 'undefined')trustedTypes={createPolicy:(n, rules) => rules};
```

[...]

PR #2855, opened 2020-03-31
feat(fast-element): add trusted types for HTML

EisenbergEffect:
# Description

This PR integrates the new Trusted Types web standard at all locations where user HTML could come through the system.

## Motivation & context

See #2850 for the full background.

As part of this PR a new `createHTML` method was added to the `DOM` helper. Both the template compilation process and `innerHTML` bindings now force all HTML to pass though this method. Internally, the method uses an HTML Trusted Types Policy against the HTML. The default policy is a noop. However, a custom policy that sanitizes HTML can be configured through the new `setHTMLPolicy` method on the `DOM` helper. This method can only be called once, for security reasons. Documentation has been updated with guidance around trusted type policies.

In order to implement this in such a way that the policy could be configured before templates are compiled, I had to refactor a few pieces to make template compilation lazy. This allows templates to be declared without forcing immediate compilation. The compilation now happens just-in-time for the first view instantiation. Besides addressing the timing issues related to policy configuration, this also yields a startup performance benefit in the case where there are many components or templates defined, but only a few of them are actually needed for the initial render.

## Issue type checklist

- [x] **Performance** Improves performance or memory usage.

Yes, this is technically a breaking change to the low-level APIs related to template compilation. However, this is unlikely to affect anyone at this point in time.

## Process & policy checklist


packages/fast-element/src/directives/directive.ts
```diff
@@ -19,7 +19,7 @@ export class AttachedBehaviorDirective<T = any> extends Directive {
     }
 
     public createPlaceholder(index: number) {
-        return `${this.name}="${DOM.createInterpolationPlaceholder(index)}"`;
+        return DOM.createCustomAttributePlaceholder(this.name, index);
```
EisenbergEffect: This change is unrelated to the main purpose of this PR. I just moved this code into the `DOM` helper in order to facilitate reuse and participate in the DOM's frozen characteristic.

packages/fast-element/src/dom.ts
```diff
@@ -4,7 +4,35 @@ const markerClass = `fast-${Math.random()
...
+const fastHTMLPolicy: TrustedTypesPolicy = globalThis.trustedTypes.createPolicy(
+    "fast-html",
+    {
+        createHTML: html => html,
+    }
+);
+
+let htmlPolicy: TrustedTypesPolicy = fastHTMLPolicy;
+
+export const DOM = Object.freeze({
```
EisenbergEffect: The `DOM` object is nor frozen so that it's `createHTML` method cannot be overwritten by attackers.

packages/fast-element/src/template.ts
```diff
@@ -14,38 +14,54 @@ export interface SyntheticViewTemplate {
     create(): SyntheticView;
 }
 
-export class HTMLTemplate extends Directive
+export class ViewTemplate extends Directive
```
EisenbergEffect: Renamed for greater clarity (related to interfaces) and to disambiguate from platform template primitives.

packages/fast-element/src/dom.ts
nicholasrice: nice!

PR #2856, opened 2020-03-31
chore: add number field field spec

chrisdholt:
# Description
Adds a new specification for number field. 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/number-field.md
+++ b/specs/number-field.md
@@ -1,0 +1,128 @@
+# Number Field
+
+## Overview
+
+An implementation of a generic input for numerical values as a form-connected web-component.
+
+### Use Cases
+
+Used anywhere an author might otherwise use input[type="number"].
+
+### Features
+
+1. Form association
+
+2. Conversion to/from number/string between the view and the control
+
+### Prior Art/Examples
+
+- [FAST-DNA (React)](https://explore.fast.design/components/number-field)
+- [Carbon Design](https://www.carbondesignsystem.com/components/number-input/code)
+- [Ant Design](https://ant.design/components/input-number/)
+- [Office Fabric](https://developer.microsoft.com/en-us/fabric#/controls/web/spinbutton)
+- [Duet](https://www.duetds.com/components/number-input/)
+
+---
+
+### API
+
+Extends Form associated component
+
+*Component Name*
+- `fast-number-field`
+
+*Attrs* - Adapted from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number)
+- `autofocus` - automatically focuses the control
+- `readonly` - the number field should be submitted with the form but should not be editable.
+- `placeholder` - an exemplar value to display in the input field whenever it is empty
+- `required` - boolean value that sets the field as required
+- `value` - string value of the number field, can be an empty string
+- `disabled` - disables the control
+- `list` - the id of the `<datalist>` element that contains the optional pre-defined autocomplete options
+- `max`	- the maximum value to accept for this input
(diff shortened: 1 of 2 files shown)


specs/number-field.md
```diff
@@ -0,0 +1,123 @@
...
+- `list` - the id of the `<datalist>` element that contains the optional pre-defined autocomplete options
+- `max`	- the maximum value to accept for this input
+- `min` -	the minimum number to accept for this input
+- `maxlength`	- the maximum number of characters the input should accept
+- `minlength` -	the minimum number of characters long the input can be and still be considered valid
+- `name` - the name of the control
+- `size` - a number indicating how many characters wide the input field should be
+- `appearance` - enum
+  - outline (default)
+  - filled
```
EisenbergEffect: Should we also include whether or not the number can contain a decimal and how many decimal places it supports?

EisenbergEffect: Should we support a `step` property that indicates how much to increment/decrement by? As an example, imagine an audio app that wants to represent tuning in KHz. They might want to have an increment of .1 KHz.

specs/number-field.md
```diff
@@ -0,0 +1,123 @@
...
+
+### States
+
+**disabled**: `true` or `false` - when disabled, the value will not be changeable through user interaction. It should also not expose it's value to a form submission.
+
+**readonly**: `true` or `false` - when readonly, the value will not be changeable through user interaction. The value will still be exposed to forms on submission.
+
+### Accessibility
+
+The host element will be the focusable element and should handle providing the role of textbox to assistive technology (AT) as well as reporting any state based aria attributes (disabled, readonly, etc) to AT.
```
EisenbergEffect: Are there any keyboard interactions for increment/decrement of the number?

specs/number-field.md
```diff
@@ -0,0 +1,123 @@
...
+
+*Screenshots below are of the basic appearance of the component and are not exhaustive.*
+
+**The visual treatment of this control should visually match the number field component**
+
+| State | Image |
+| ----- | ----- |
+| default | ![](../text-field/images/text-field.png) |
+| focus | ![](../text-field/images/text-field-focus.png)
+| filled | ![](../text-field/images/text-field-filled.png)
```
EisenbergEffect: Images didn't seem to show up for me.

specs/number-field.md
```diff
@@ -0,0 +1,123 @@
...
+- disabled
+- required
+- readonly
+
+*CSS Parts*
+- label
+- root
+- start
+- end
+- control
```
EisenbergEffect: Would we provide custom spin buttons with the ability to style them and replace their icon?

specs/number-field.md
chrisdholt: I think step for sure should be added, I'll update with that.

specs/number-field.md
chrisdholt: Sure, I can articulate that.

specs/number-field.md
chrisdholt: Odd, I'll see if I can figure out why...or I'll dupe and rename :)

specs/number-field.md
chrisdholt: I think we'll provide slots for it, but I think the buttons themselves might be a different spec and PR.

chrisdholt: Initially I think the spin buttons could be passed into the existing `end` slot. I'll add next steps for creating those and passing them by default.

specs/number-field.md
chrisdholt: added step 👍

PR #2857, opened 2020-03-31
chore: add tabs web component spec

eljefe223:
# Description
Adds spec for `fast-tabs`

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/README.md
+++ b/specs/README.md
@@ -19,6 +19,7 @@ Here you'll find specifications for custom elements and other library features.
 | [Progress](../packages/web-components/fast-components/src/progress/progress.spec.md) | :white_check_mark: |
 | [Slider](../packages/web-components/fast-components/src/slider/slider.spec.md) | :white_check_mark: |
 | [Switch](../packages/web-components/fast-components/src/switch/switch.spec.md) | :white_check_mark: |
+| [Tabs](../packages/web-components/fast-components/src/tabs/tabs.md) | :white_check_mark: |
 | [Text Area](../packages/web-components/fast-components/src/text-area/text-area.spec.md) | :white_check_mark: |
 | [Text Field](../packages/web-components/fast-components/src/text-field/text-field.spec.md) | :white_check_mark: |
 | [Tooltip](./tooltip.md) | -- |
--- a/specs/README.md
+++ b/specs/README.md
@@ -19,7 +19,7 @@ Here you'll find specifications for custom elements and other library features.
 | [Progress](../packages/web-components/fast-components/src/progress/progress.spec.md) | :white_check_mark: |
 | [Slider](../packages/web-components/fast-components/src/slider/slider.spec.md) | :white_check_mark: |
 | [Switch](../packages/web-components/fast-components/src/switch/switch.spec.md) | :white_check_mark: |
-| [Tabs](../packages/web-components/fast-components/src/tabs/tabs.md) | :white_check_mark: |
+| [Tabs](./tabs/tabs.md) | :white_check_mark: |
 | [Text Area](../packages/web-components/fast-components/src/text-area/text-area.spec.md) | :white_check_mark: |
 | [Text Field](../packages/web-components/fast-components/src/text-field/text-field.spec.md) | :white_check_mark: |
 | [Tooltip](./tooltip.md) | -- |
--- a/specs/tabs/tabs.md
+++ b/specs/tabs/tabs.md
@@ -1,4 +1,4 @@
-# fast-tabs
+# Tabs
 
 ## Overview
 
@@ -41,19 +41,19 @@
   - vertical
 - `activeId` - string
 - `tab-activation` - enum 
-  - focus - default
-  - key-press
+  - auto - default
+  - manual
 
 *Events:*
-- `onChange` - fires when component `activeId` updates
+- `change` - fires when component `activeId` updates
 
 ### Anatomy and Appearance
 
 *Parts:*
 - tabs
-- tab-list
+- tablist
 - tab
-- tab-panel
+- tabpanel
 
 *Slot Names*
 - before-content
@@ -62,8 +62,8 @@
 *Template:*
 ```HTML
 <div class="tabs" class="tabs">
-    <div class="tab-list" part="tab-list" role="tablist" aria-label="Sample Tabs">
-        <slot name="before-content" part="before-content"></slot>
+    <slot class="start" name="start" part="start"></slot>
+    <div class="tablist" part="tablist" role="tablist" aria-label="Sample Tabs">
         <button class="tab" part="tab" role="tab" aria-selected="true" aria-controls="panel-1" id="tab-1" tabindex="0">
             Tab One
         </button>
@@ -73,15 +73,15 @@
         <button class="tab" part="tab" role="tab" aria-selected="false" aria-controls="panel-3" id="tab-3" tabindex="-1">
             Tab Three
         </button>
-        <slot name="after-content" part="after-content"></slot>
     </div>
-    <div class="tab-panel" part="tab-panel" id="panel-1" role="tabpanel" tabindex="0" aria-labelledby="tab-1">
+    <slot class="end" name="end" part="end"></slot>
+    <div class="tabpanel" part="tabpanel" id="panel-1" role="tabpanel" tabindex="0" aria-labelledby="tab-1">
         Content of the first tab
     </div>
-    <div class="tab-panel" part="tab-panel" id="panel-2" role="tabpanel" tabindex="0" aria-labelledby="tab-2" hidden>
+    <div class="tabpanel" part="tabpanel" id="panel-2" role="tabpanel" tabindex="0" aria-labelledby="tab-2" hidden>
         Content of the second tab
     </div>
-    <div class="tab-panel" part="tab-panel" id="panel-3" role="tabpanel" tabindex="0" aria-labelledby="tab-3" hidden>
+    <div class="tabpanel" part="tabpanel" id="panel-3" role="tabpanel" tabindex="0" aria-labelledby="tab-3" hidden>
         Content of the third tab
     </div>
 </div>
@@ -156,7 +156,7 @@
 
 ### States
 
-*Tabs* can either be controlled or uncontrolled, meaning if `activeId` is passed the app author is taking control of the selected tab. The `onChange` event fires differently depending on the `tab-activation` attribute, on tab focus if `tab-activation` is set to `"focus"` or on mouse click or "spacebar" press if `tab-activation` is set to `"key-press"`.
+*Tabs* can either be controlled or uncontrolled, meaning if `activeId` is passed the app author is taking control of the selected tab. The `change` event fires differently depending on the `tab-activation` attribute, on tab focus if `tab-activation` is set to `"focus"` or on mouse click or "spacebar" press if `tab-activation` is set to `"key-press"`.
 
 ### Globalization
 
--- a/specs/tabs/tabs.md
+++ b/specs/tabs/tabs.md
@@ -46,11 +46,11 @@ Some scenarios require an indicator that highlights the currently active tab the
 - `orientation` - enum
   - horizontal - default
   - vertical
-- `activeId` - string
-- `activeIndicator` - boolean, defaults to true
+- `activeid` - string
+- `activeindicator` - boolean, defaults to true
 
 *Events:*
-- `change` - fires when component `activeTab` updates
+- `change` - fires when component `activetab` updates
 
 *Parts:*
 - tabs
@@ -133,7 +133,7 @@ Some scenarios require an indicator that highlights the currently active tab the
 
 ### States
 
-*Tabs* can either be controlled or uncontrolled, meaning if `activeId` is passed the app author is taking control of the selected tab. When the `change` event fires it updates the `activeId` and pass a refrence to the `activeTab`.
+*Tabs* can either be controlled or uncontrolled, meaning if `activeid` is passed the app author is taking control of the selected tab. When the `change` event fires it updates the `activeid` and pass a refrence to the `activetab`.
 
 ### Globalization
 
--- a/specs/tabs/tabs.md
+++ b/specs/tabs/tabs.md
@@ -1,0 +1,173 @@
+# fast-tabs
+
+## Overview
+
+*Tabs* are a set of layered sections of content, known as tab panels, that display one panel of content at a time. Each tab panel has an associated tab element, that when activated, displays the panel. The list of tab elements is arranged along one edge of the currently displayed panel.
+
+### Use Cases
+
+- Steve is researching for an upcoming vacation for him and his family. He opens his web browser and opens a new tab to make a rental car reservation and another for his flight information. Steve can quickly switch back and fourth between tabs to get the information he needs for the rental car reservation.
+
+- Monika visits a website to learn how to fix her leaky sink. The website displays a vertical tab interface with each step. Monika can jump ahead steps or go back to previous steps by selecting various tabs.
+  
+### Features
+
+- **Orientation:** Allows the tab list to be oriented vertically to the left or right (depending on language region) of the tab content or horizontally above the content.
+- **Tab activation:** Offers a way to activate a tab and it's content by either focusing a tab or by key press (Space or Enter).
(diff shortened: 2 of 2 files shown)


specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+### API
+
+*Component name:*
+- `fast-tabs`
+
+*Attributes:*
+- `orientation` - enum
+  - horizontal - default
+  - vertical
+- `activeId` - string
```
EisenbergEffect, line 42: For this, I think it might be better to have this be a reference to the tab element that is active, rather than an id, if possible.

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+- tab-panel
+
+*Slot Names*
+- before-content
+- after-content
+
+*Template:*
+```HTML
+<div class="tabs" class="tabs">
+    <div class="tab-list" part="tab-list" role="tablist" aria-label="Sample Tabs">
```
EisenbergEffect, line 65: I can't remember if I noticed this for previous components, but I wonder if there's value to making the class and part name the same as the role in situations where an element has all three.

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+        <fast-tab>Tab Three</fast-tab>
+        <fast-tab-panel>Content of the Third tab</fast-tab-panel>
+    </fast-tab-item>
+    <div slot="before-content">
+        Before content
+    </div>
+    <div slot="after-content">
+        Before content
+    </div>
+</fast-tabs>
```
EisenbergEffect, line 134: What about something like this:

```html
<fast-tabs>
  <fast-tab-item>
    <span slot="tab">Tab One</span>
    The rest of the content is the tab panel content...
  <fast-tab-item>
</fast-tabs>
```

EisenbergEffect: Though, I'm not sure how we could easily enable the DOM grouping below with this logical grouping. If the tab was only text, it could be done without too much trouble. But making that slottable (which is very nice from a composition perspective) makes it tricky since you can't really take the slotted content of one element and project it back to its parent. We'd have to do some very strange voodoo to make that happen.

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+
+*Slot Names*
+- before-content
+- after-content
+
+*Template:*
+```HTML
+<div class="tabs" class="tabs">
+    <div class="tab-list" part="tab-list" role="tablist" aria-label="Sample Tabs">
+        <slot name="before-content" part="before-content"></slot>
```
chrisdholt, line 66: How does this work being nested within the role of tablist? Have we tested that this would actually be accessible? As it's being presented here,  it appears as though *anything* can be a `chlid` (or "permitted content" in spec language) which I don't think is the case. Are there examples we can point to in the spec or where this is supported  or permitted?

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+    <div class="tab-panel" part="tab-panel" id="panel-2" role="tabpanel" tabindex="0" aria-labelledby="tab-2" hidden>
+        Content of the second tab
+    </div>
+    <div class="tab-panel" part="tab-panel" id="panel-3" role="tabpanel" tabindex="0" aria-labelledby="tab-3" hidden>
+        Content of the third tab
+    </div>
+</div>
+```
+
+## Implementation
```
chrisdholt, line 90: Do you have a proposal for which implementation would work best?

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+
+*Parts:*
+- tabs
+- tab-list
+- tab
+- tab-panel
+
+*Slot Names*
+- before-content
+- after-content
```
chrisdholt, line 60: We have an existing issue looking at naming for these. I think we're close to aligning on "start" and "end". If there's no objection with those I would suggest updating to match.

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+- `fast-tabs`
+
+*Attributes:*
+- `orientation` - enum
+  - horizontal - default
+  - vertical
+- `activeId` - string
+- `tab-activation` - enum 
+  - focus - default
+  - key-press
```
chrisdholt, line 45: Can you elaborate a bit more about the `tab-activation` attribute? What purpose does it serve and how is it used?

specs/README.md
```diff
@@ -13,7 +13,8 @@ Here you'll find specifications for custom elements and other library features.
 - [Menu](./menu/menu.md)
 - [Progress](./progress.md)
 - [Switch](./switch.md)
+- [Tabs](./tabs/tabs.md)
 - [Text Area](./text-area/text-area.md)
 - [Text Field](./text-field/text-field.md)
-- [Tree View](./tree-view/tree-view.md)
+- [Tree View](./tree-view/tree-view.md)git
```
chrisdholt, line 19: oops :)

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
+# fast-tabs
```
chrisdholt, line 1:
```suggestion
# Tabs
```

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+---
+
+### API
+
+*Component name:*
+- `fast-tabs`
+
+*Attributes:*
+- `orientation` - enum
+  - horizontal - default
```
chrisdholt, line 40: YES!

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+  - vertical
+- `activeId` - string
+- `tab-activation` - enum 
+  - focus - default
+  - key-press
+
+*Events:*
+- `onChange` - fires when component `activeId` updates
```
nicholasrice, line 48: this would just be 'change' - the "on" is for event *handlers*

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+    </div>
+    <div slot="after-content">
+        Before content
+    </div>
+</fast-tabs>
+```
+
+### States
+
+*Tabs* can either be controlled or uncontrolled, meaning if `activeId` is passed the app author is taking control of the selected tab. The `onChange` event fires differently depending on the `tab-activation` attribute, on tab focus if `tab-activation` is set to `"focus"` or on mouse click or "spacebar" press if `tab-activation` is set to `"key-press"`.
```
nicholasrice, line 159: Can you provide a code-sample on how this `activeId` would be used? I'm struggling to visualize how an end-user would use this and where the IDs come from.

specs/tabs/tabs.md
eljefe223: `activeTab`?

specs/tabs/tabs.md
eljefe223: Good point would be easier just to love the slotted content outside the tab list

specs/tabs/tabs.md
eljefe223: I'm not sure we want to support both but its outlined in the WC3:

Automatic (Focus) 
https://w3c.github.io/aria-practices/examples/tabs/tabs-1/tabs.html

Manual (Key Press)
https://w3c.github.io/aria-practices/examples/tabs/tabs-2/tabs.html

IF we do think we should support these maybe should match the naming they use here, `auto` and `manual`

specs/tabs/tabs.md
eljefe223: An app author could set the `activeId` on the tabs.

```
<fast-tabs activeId="tab2">
    <fast-tab-item id="tab1">
        <fast-tab>Tab One</fast-tab>
        <fast-tab-panel>Content of the first tab</fast-tab-panel>
    </fast-tab-item>
    <fast-tab-item id="tab2">
        <fast-tab>Tab Two</fast-tab>
        <fast-tab-panel>Content of the Second tab</fast-tab-panel>
    </fast-tab-item>
    <fast-tab-item id="tab3">
        <fast-tab>Tab Three</fast-tab>
        <fast-tab-panel>Content of the Third tab</fast-tab-panel>
    </fast-tab-item>
</fast-tabs>
```

specs/tabs/tabs.md
eljefe223: I personal like logical grouping option 2, as @EisenbergEffect 
noted above we could just have `<fast-tabs>` and `<fast-tab-item>` where the tab and tab content are slotted elements on `<fast-tab-item>` something like:

```
<fast-tabs>
    <fast-tab-item>
        <span slot="tab">Tab One</span>
        Content of the first tab
    </fast-tab-item>
    <fast-tab-item>
        <span  slot="tab">Tab Two</span>
        Content of the Second tab
    </fast-tab-item>
    <fast-tab-item>
        <span  slot="tab">Tab Three</span>
        Content of the Third tab
    </fast-tab-item>
</fast-tabs>
```

specs/tabs/tabs.md
eljefe223: Could we potentially look for elements inside `<fast-tabs/>` that have a certain slot and place all the `"tab"` inside a `tab-list` and the default slotted content of `<fast-tab-item>` as a `tabpanel`?

The template of `<fast-tab-item>` would look something like:

```
<template>
    <div class="tab" part="tab" role="tab" slot="tab">
         <slot name="tab"><slot>
     <div>
     <div class="tabpanel" part="tabpanel" role="tabpanel" slot="tabPanel">
          <slot></slot>
     </div>
</template>
```

specs/tabs/tabs.md
EisenbergEffect: That works for me.

specs/tabs/tabs.md
eljefe223: Another option is to abstract tabs further and have an `items` attribute on `fast-tabs` that would be and array of objects that contain the data for each tab. This is what we currently do in react but this doesn't seem like a  "web-component" approach. I don't know of any native elements that take complex objects.

EisenbergEffect: Ideally, I think we want the imperative JS slot API for this, so we can take the tabs directly from one element and then manually add them to the slot of another element. However, that's not available today. I think we can accomplish what you're saying still though. I don't know for sure. My recommendation would be to do a PoC before we continue with this spec, so that we can know for sure. Here's how I think it *could* work:

Assume the following markup that the customer would write:

```HTML
<fast-tabs>
  <fast-tab-item>
    <span slot="tab">Tab One</span>
    The rest of the content is the tab panel content...
  <fast-tab-item>
</fast-tabs>
```

The `fast-tab-item` would have no slot named `tab` so that it's tab would not display where it's at. However, the `fast-tabs` would have a `tab` slot.

All `fast-tab-item` would be projected into the `default` slot of `fast-tabs`. We then have a `slotchanged` event handler on the default slot of `fast-tabs`. Whenever that event fires, we look at the slotted `fast-tab-item` elements, grab their child nodes and find their nodes marked with slot of `tab`. Those nodes then get removed from the child nodes of the item and added to the child nodes of `fast-tabs` itself.

I think this should work. It does change the light dom structure from what the consumer writes, so that if they write the above, the end result would be something like this:

```
<fast-tabs>
  <span slot="tab">Tab One</span>
  <fast-tab-item>  
    The rest of the content is the tab panel content...
  <fast-tab-item>
</fast-tabs>
```

The trick with this would be tracking the tab slots back to the items they were removed from, so that if tabs are re-ordered, the tabs order can be updated to match. This would also need to handle dynamic adding and removing of tabs as well. The `slotchanged` event will fire on all of these things, so we'll know when they happen. We'd just have to do the bookeeping ourselves to keep it in sync.

**Note:** We can also wrap the tab content in arbitrary html before moving it to the tabs element as well.

There may be other ways to accomplish this, but I don't think there's an out-of-the-box way to use slots to get the projection we want and the authoring experience we want. Do we want to do a PoC and see if something similar to this outline works?

specs/tabs/tabs.md
eljefe223: I'm struggling to see why a reference to the active tab is better than an ID? how would an app author set the `activeTab`? With `activeId` it would be something like this:

```HTML
<fast-tabs activeId="tab2">
    <fast-tab-item id="tab1">
        <span slot="tab">Tab One</span>
        Content of the first tab
    </fast-tab-item>
    <fast-tab-item id="tab2">
        <span slot="tab">Tab Two</span>
        Content of the Second tab
    </fast-tab-item>
    <fast-tab-item id="tab3">
        <span slot="tab">Tab Three</span>
        Content of the Third tab
    </fast-tab-item>
</fast-tabs>
```

specs/tabs/tabs.md
eljefe223: I think if there is any question we should do a PoC first.

specs/tabs/tabs.md
EisenbergEffect: If the customer is setting the component up in JS, this allows them to create the tab items and set the active item without needing to set up an id. This also enables templating libraries that support the "ref" concept to let their users assign the active tab using a ref.

So, those are a couple of use cases. Beyond that, for some of our other web components, we've been following a pattern of using a control ref property with a corresponding id property. So, I'd like to do this for consistency both with that and the native platform, which uses the same pattern for controls like label.

specs/tabs/tabs.md
EisenbergEffect: I agree. I think it's worth taking some time to do that. @eljefe223 are you up for working on that. @chrisdholt Is it ok if we take some time to explore through a PoC here?

specs/tabs/tabs.md
EisenbergEffect: If the customer is setting the component up in JS, this allows them to create the tab items and set the active item without needing to set up an id. This also enables templating libraries that support the "ref" concept to let their users assign the active tab using a ref.

So, those are a couple of use cases. Beyond that, for some of our other web components, we've been following a pattern of using a control ref property with a corresponding id property. So, I'd like to do this for consistency both with that and the native platform, which uses the same pattern for controls like label.

specs/tabs/tabs.md
EisenbergEffect: Another option for the html could be to move the tab and panel out.

```html
<fast-tabs>
        <fast-tab>Tab One</fast-tab>
        <fast-tab-panel>Content of the first tab</fast-tab-panel>

        <fast-tab>Tab Two</fast-tab>
        <fast-tab-panel>Content of the Second tab</fast-tab-panel>

        <fast-tab>Tab Three</fast-tab>
        <fast-tab-panel>Content of the Third tab</fast-tab-panel>
</fast-tabs>
```

EisenbergEffect: With this configuration, we don't need to do any shuffling ourselfs. We can make the fast-tab go to the "tab" slot and fast-tab-panel go to the default slot.

eljefe223: @chrisdholt @nicholasrice do you guys have thoughts here?

specs/tabs/tabs.md
chrisdholt: @EisenbergEffect I missed this, but I think it's actually preferred. I think the added complexity and potential variations in use case require us actually testing out with a proof of concept first.

specs/tabs/tabs.md
chrisdholt: Can we try this in the proof of concept suggested above? I think that'll give us the best lead on any potential drawbacks or limitations regarding implementation and API

specs/README.md
```diff
@@ -19,6 +19,7 @@ Here you'll find specifications for custom elements and other library features.
 | [Progress](../packages/web-components/fast-components/src/progress/progress.spec.md) | :white_check_mark: |
 | [Slider](../packages/web-components/fast-components/src/slider/slider.spec.md) | :white_check_mark: |
 | [Switch](../packages/web-components/fast-components/src/switch/switch.spec.md) | :white_check_mark: |
+| [Tabs](./tabs/tabs.md) | :white_check_mark: |
```
chrisdholt, line 22: Technically the check mark would go in when the component is complete - I plan on pushing a change for the table again today based on some feedback.

PR #2858, opened 2020-04-01
fix: correct tabs interaction model to map to the w3c specification and add default orientation

chrisdholt:
# Description
This change fixes the interaction model to align with the W3C specification.

## Motivation & context
It was recently brought to my attention that the MSFT carousel and pivot components both prevent page scroll behavior on arrow up/down events. While that behavior is expected for vertically oriented tabs, it is unexpected for horizontally oriented tabs. The [W3C specification](https://w3c.github.io/aria-practices/#keyboard-interaction-20) specifically details that horizontal tabs do not listen for up/down events to enable the default browser behavior:
> If the tab list is horizontal, it does not listen for Down Arrow or Up Arrow so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.

With this in mind, I've updated the control to behave as expected. In addition, this change reveals a need to align to a single orientation by default. Given that the spec reads as though the default is horizontal and our two first class "styled" implementations are horizontal, this PR proposes a default value of "horizontal". 

While this change technically breaks existing behavior, that behavior is incorrect. I believe this should be classified as a fix without the designation of a "breaking change".

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components-react-base/src/tabs/tabs.tsx
+++ b/packages/fast-components-react-base/src/tabs/tabs.tsx
@@ -8,6 +8,7 @@ import {
     keyCodeArrowUp,
     keyCodeEnd,
     keyCodeHome,
+    Orientation,
 } from "@microsoft/fast-web-utilities";
 import { get } from "lodash-es";
 import React from "react";
@@ -37,6 +38,7 @@ export interface TabsState {
 
 class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
     public static defaultProps: Partial<TabsProps> = {
+        orientation: Orientation.horizontal,
         disableTabFocus: false,
         managedClasses: {},
     };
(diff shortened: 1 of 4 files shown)


packages/fast-components-react-base/src/tabs/tabs.tsx
```diff
@@ -316,17 +318,32 @@ class Tabs extends Foundation<TabsHandledProps, TabsUnhandledProps, TabsState> {
...
+                }
+                break;
             case keyCodeArrowDown:
-                e.preventDefault();
-
-                this.activateTab(TabLocation.next);
+                if (!isHorizontal) {
+                    e.preventDefault();
+                    this.activateTab(TabLocation.next);
+                }
```
janechu, line 346: I wonder if it might be easier to have an abstraction more like this:
```typescript
// horizontal orientation utilizes the left and right arrows
// while vertical orientation utilizes the up and down arrows
if (this.props.orientation === Orientation.horizontal) {
    switch (e.keyCode) {
        case keyCodeArrowLeft:
            e.preventDefault();
            this.activateTab(TabLocation.previous);
            return;
        case keyCodeArrowRight:
            e.preventDefault();
            this.activateTab(TabLocation.next);
            return;
    }
} else {
    switch (e.keyCode) {
        case keyCodeArrowUp:
            e.preventDefault();
            this.activateTab(TabLocation.previous);
            return;
        case keyCodeArrowDown:
            e.preventDefault();
            this.activateTab(TabLocation.next);
            return;
    }
}

switch (e.keyCode) {
    case keyCodeHome:
        this.activeTab(TabLocation.first);
        return;
    case keyCodeEnd:
        this.activeTab(TabLocation.last);
        return;
}
```

chrisdholt: sure, good suggestion. Updated.

PR #2860, opened 2020-04-01
feat: design system provider and recipe implementation

nicholasrice:
# Description
Adds a system for mapping design system values to css custom properties and for allowing consumers to register CSS custom property setters on the design system.

## Motivation & context
Many theming values can be described as simple css custom properties. To expose a mechanism for *managing* those properties, the design-system-provider allows the setting of content attributes & IDL properties and reflecting those values to css custom properties. These values are intended to be used directly in component stylesheets.

More *advanced* theming values require some calculation and are a product *of* the design system. To meet that goal, the system allows declaring a custom element as a design system consumer and then propagating dependent custom property calculator up to the design system provider. Whenever the design system changes, these property calculators will be re-evalutated.

## Issue type checklist

## Process & policy checklist


nicholasrice:
Still writing up some documentation but wanted to get the code out to start getting comments on it.

packages/fast-components/src/design-system-consumer/design-system-consumer.ts
```diff
@@ -0,0 +1,76 @@
...
+            public provider: DesignSystemProvider | null = null;
+
+            /**
+             * Find the parent DesignSystem provider.
+             */
+            public findProvider(): DesignSystemProvider | null {
+                let parent = composedParent(this as any);
+
+                while (parent !== null) {
+                    if ((parent as any).isDesignSystemProvider) {
```
nicholasrice, line 38: One way I think we could enhance this is to check also check if the parent is a *consumer* - if it is, then we can just attach to it's provider. I was also considering checking *siblings* if they are a consumer but thought that may start to get inefficient - thoughts?

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
+    }
 
-    @designSystemProperty
-    public disabledOpacity = 0.3;
+    /**
+     * Invoked when the provider observable property defined by the consumer is changed
+     * @param prev the previous value
+     * @param next the next value
+     */
+    private providerChanged(
```
nicholasrice, line 197: This feels a bit wonky to me because it is actually the *consumer* that defines the observable property. If we were to add a providerChanged callback to the consumer then this wouldn't fire. We need the prev/next values though to unsubscribe. Happy to hear thoughts on an alternative way to approach this.

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
+    }
+}
 
-    @designSystemProperty
-    public designUnit = 4;
+/**
+ * Simple proxy object to notifiy observers on
+ * object mutation.
+ * @param source The object to fire notifications
+ * @param key The observable property name to fire a notification on
```
nicholasrice, line 216: @EisenbergEffect this function is primarily a work-around for `@observable` because observable won't notify on property mutation. You mentioned something earlier on making observable support mutation I think, perhaps you have thoughts here.

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
-    public outlineWidth = 1;
+        if (ticking) {
+            return;
+        } else {
+            ticking = true;
 
-    @observable
-    public designSystem: DesignSystem = DesignSystemDefaults;
+            window.requestAnimationFrame(() => {
+                ticking = false;
```
nicholasrice, line 255: @EisenbergEffect This seemed like an opportunity to hook into the rendering engine but I'm not sure how to do that. My thought here was that setting individual property values was more efficient than concating a giant list of properties into a style attribute string and binding to that in the templating engine but I could be off-base on that.

packages/fast-components/src/styles/recipes.ts
```diff
@@ -0,0 +1,339 @@
...
+    neutralLayerL3,
+    neutralLayerL4,
+    neutralOutline,
+} from "@microsoft/fast-components-styles-msft";
+
+export interface DesignSystemResolverEntry {
+    name: string;
+    resolver: (designSystem: DesignSystem) => string | number;
+}
+export const neutralforegroundrest: DesignSystemResolverEntry = {
```
bheston, line 36: Could these be camel case? It's pretty hard to read where they are referenced.

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
+    customProperty?: boolean;
+}
 
 /**
- * The following interface is an implementation of
- * design-system.schema.json in TypeScript with
- * as much accuracy as TypeScript supports.
+ * Decorator to declare a property as a design-system property.
+ * Accepts an optional config to customize whether a css custom property
+ * wll be written and if so, what the name of that property is.
```
bheston, line 17: typo "will"

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
-    focusOutlineWidth: number;
-    outlineWidth: number;
+export function designSystemProperty<T extends DesignSystemProvider>(
+    config: DesignSystemPropertyDeclarationConfig
+): (source: T, property: string) => void;
+export function designSystemProperty<T extends DesignSystemProvider>(
+    source: T,
+    property: string
+): void;
+export function designSystemProperty<T extends DesignSystemProvider>(
```
bheston, line 26: I find this style of functions really hard to read and understand. Is there no better / new way to achieve this, or should we create named variants instead or something? The logic is often hard to follow.

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
+        } else if (source === this.provider && key === designSystemKey) {
+            // If our provider's design system has changed
+            this.syncDesignSystemWithProvider();
+            this.consumers.forEach(this.writeConsumerRecipeData);
         }
     }
 
-    @designSystemProperty
-    public backgroundColor: string = "#FFFFFF";
+    public suscribe(consumer: DesignSystemConsumer): void {
```
bheston, line 152: "subscribe" *cough* https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker ;)

packages/fast-components/src/card/fixtures/card.html
```diff
@@ -1,4 +1,4 @@
-<fast-design-system-provider>
+<fast-design-system-provider use-defaults>
```
bheston, line 1: "use-defaults" seems to be the default way to consume this. That is, it's added here a number of times, though places it wasn't added obviously wouldn't show as a diff. Should / could this be the default behavior?

packages/fast-components/src/checkbox/index.ts
```diff
@@ -2,12 +2,29 @@ import { customElement } from "@microsoft/fast-element";
...
+    neutraloutlinehover,
+    neutraloutlinerest,
+} from "../styles/recipes";
 
 @customElement({
     name: "fast-checkbox",
     template,
     styles,
 })
+@designSystemConsumer({
```
bheston, line 19: If I'm reading this correctly, it looks like I would need to import and associate a recipe like neutralOutlineRest in the component, then I could reference `var(--neutral-outline-rest)` in the styles. Having this disconnected seems difficult to keep in sync, and doesn't seem like it supports swapping styles on existing components. I'm sure I don't know most of the internals of the model, but is there a way to swap or chain up this relationship so the declarative part comes in the `css` function instead of the component?

packages/fast-components/src/design-system-provider/design-system-provider.ts
nicholasrice: You mean the overload syntax? https://www.typescriptlang.org/docs/handbook/functions.html#overloads

There isn't really a better alternative that I know of.

packages/fast-components/src/styles/recipes.ts
nicholasrice: oops - yes.

packages/fast-components/src/card/fixtures/card.html
nicholasrice: Yea I just had a thought about this while writing the documentation. We only want to provide custom property values when the value is being explicitly set on the element because otherwise we would clobber the cascade and not inherit upstream values. `use-defaults` is essentially a short-hand to apply all the values. We could *also* apply the defaults if there is no upstream provider that was found - this would handle the root design system case in a way that doesn't require setting all of the attributes.

packages/fast-components/src/checkbox/index.ts
nicholasrice: Yea I agree. Part of this was to get *an* implementation up and see where the gaps are and what API changes we might need to make to fast-element to get a good experience with recipes. I don't think we have infrastructure to associate recipes with a stylesheet instance at the moment.

packages/fast-components/src/design-system-provider/design-system-provider.ts
nicholasrice, line 17:
```suggestion
 * will be written and if so, what the name of that property is.

```

packages/fast-components/src/design-system-provider/design-system-provider.ts
nicholasrice: =D

packages/fast-components/src/checkbox/index.ts
EisenbergEffect: To make sure I understand, you would want to associate recipes with a stylesheet, and then when you apply the styles, also apply the recipes. Is that correct? If so, I think we could do that without a ton of trouble. That could get pretty interesting considering also the idea of registering things by name so that an appearance change could add and remove the styles dynamically (and now recipes).

Let me know if my understanding is correct on that.

packages/fast-components/src/design-system-consumer/design-system-consumer.ts
EisenbergEffect: I don't think we should check siblings, but attaching to the parent makes sense to me. I wouldn't block on implementing that optimization though.

packages/fast-components/src/design-system-consumer/design-system-consumer.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    if (typeof argsOrConstructor === "function") {
+        return decorator(argsOrConstructor, { recipes: [] });
+    } else {
+        return (constructor: T): T => {
+            return decorator(constructor, argsOrConstructor);
+        };
+    }
+}
```
EisenbergEffect, line 76: Looking at the consumer work, I think we can eventually factor this to work through DI, so we don't need the extra decorator or inheritance. I don't want to block on that because DI just isn't close to ready. So, we should keep this implementation. But, I think we could have something like:

```TypeScript
@customElement({
  name:'foo-bar',
  template,
  styles,
  dependencies: [designSystemConsumer({
    recipes: [
        neutralOutlineRest,
        neutralFillInputRest,
        neutralForegroundRest,
        neutralFillInputHover,
        neutralOutlineHover,
      ]
  })]
})
```

Then FastElement could pull out all Behaviors from the container, one of which would be the consumer. That enables it to instantiate and connect/disconnect along with the element. Basically, this gives Fast Element a sort of mixin system, which can be tapped into through the dependencies configuration.

That's one way we can go. Another way is that if we associate these recipes instead with the styles themsleves, like I think was suggested above. Then since the plumbing is already in place to add styles, and we planned to enable remove with the registry feature...we could also add this logic into that.

packages/fast-components/src/design-system-provider/design-system-provider.ts
EisenbergEffect: Each property of the object itself needs and observable decorator. Is it possible to change the source so that each property is an observable or an attr?

packages/fast-components/src/design-system-provider/design-system-provider.ts
EisenbergEffect: There's a method `DOM.queueUpdate` that you can use to hook into the render batch. That will let you remove the rAF code. You will still probably want to use your `ticking` guard to dedupe queuing things up.

packages/fast-components/src/checkbox/index.ts
nicholasrice: Yep, exactly. Only the stylesheet should *really* care about it's dependent recipes, not the component.

packages/fast-components/src/design-system-consumer/design-system-consumer.ts
nicholasrice: I think the second option here is ideal. Is that something we have the support for today or do we need to do work in the stylesheet registry to get that to work?

packages/fast-components/src/design-system-provider/design-system-provider.ts
nicholasrice: I think it adds more overhead than it's worth. At this point we need to react when *any* change is made, it doesn't matter what it is. 

In the future I'd like to get us to a place where recipes depend on individual arguments and not the whole DS object so when we get there it might make sense to refactor this, but for now I'm going to keep this implementation to avoid complicating this component further.

packages/fast-components/src/card/fixtures/card.html
nicholasrice: Took a stab at this but it isn't as straight forward as I thought. Its a bit tricky to determine exactly when design system should use default values so I think we should just keep this opt-in for at the time being. This component is going to churn quite a bit I expect so I think this is something we can address as it hardens.

packages/fast-components/src/checkbox/index.ts
EisenbergEffect: I think we can make that happen. So, if we combine this with the registry idea, would we want something like this:

```TypeScript
const styles = css`
  ..some css here..
`.withRecipes([
    neutralOutlineRest,
    neutralFillInputRest,
    neutralForegroundRest,
    neutralFillInputHover,
    neutralOutlineHover,
].register('foo');
```

packages/fast-components/src/design-system-provider/design-system-provider.ts
EisenbergEffect: I think we can keep it for now. I bet we can come up for something in the future. I've been trying to avoid proxies so that we can at least say this can be polyfilled for IE11. In any case, let's go with this, since it does work, and then we'll circle back around later.

nicholasrice: Ah good point - okay I'll take another quick stab at this

packages/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,98 +1,271 @@
...
+     * or when any property defined as a design-system property is changed.
+     * Will update the designSystem object
+     * @param source The source object changing
+     * @param key The property of the source object that changed
+     */
+    public handleChange(source: any, key: string): void {
+        if (source === this && this.designSystemProperties.hasOwnProperty(key)) {
+            // If a property on *this* object that is declared as a design system property
+            this.designSystem[key] = this[key];
+            const desginSystemProperty = this.designSystemProperties[key];
```
janechu, line 137: nit: spelling on const

packages/fast-components/src/name-tag/fixtures/mark.html
```diff
@@ -1,4 +1,6 @@
 <fast-design-system-provider>
-    <fast-name-tag>Mark!</fast-name-tag>
-    <fast-name-tag style="--background-color: #00F">Blue Mark!</fast-name-tag>
+    <fast-design-system-provider>
```
janechu, line 2: does this need to be nested twice?

packages/fast-components/src/card/fixtures/card.html
janechu: I'd say if we're already checking for an upstream provider we should apply the defaults.

packages/fast-components/src/design-system-provider/design-system-provider.ts
janechu: That might not be so strange, I could see it as the consumer having to define the recipes it is interested in consuming.

packages/fast-components/src/name-tag/fixtures/mark.html
nicholasrice: I was testing inheritance. It's not a "real" component but I can revert

packages/fast-components/src/card/fixtures/card.html
nicholasrice: Again - its not very straight forward. We can take as a follow up.

packages/fast-components/src/design-system-provider/design-system-provider.ts
nicholasrice: Got it to work

packages/fast-components/src/card/fixtures/card.html
nicholasrice: https://github.com/microsoft/fast-dna/issues/2886

PR #2861, opened 2020-04-01
feat: add slider web component

marjonlynch:
# Description

Adding slider web component

slider-wc-vertical

slider-wc-forced-colors

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/slider-label/slider-label.styles.ts
```diff
@@ -0,0 +1,58 @@
+import { css } from "@microsoft/fast-element";
+import { SystemColors } from "../styles/system-colors";
+
+export const SliderLabelStyles = css`
+    :host {
```
nicholasrice, line 5: this should use the `display` function from `../styles/display` so it can benefit from the "hidden" styles and any other universal styles we want to add in the future.

packages/fast-components/src/slider/slider-utilities.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Converts a pixel coordinate on the track to a percent of the track's range
+ */
+export function convertPixelToPercent(
+    pixelPos: number,
+    minPosition: number,
+    maxPosition: number,
+    direction?: string
+): number {
+    let pct: number = 0;
```
nicholasrice, line 10: there is a `limit` function in `@microsoft/fast-web-utilities` you can leverage: 
```const pct = limit(0, 1, (pixelPos - minPosition) / maxPosition);```

File: packages/fast-components/src/slider/slider.ts [hunk 3]
```diff
@@ -0,0 +1,327 @@
...
+    private valueChanged(): void {
+        if (Number(this.value) === Number.NaN) {
+            this.value = "0";
+        }
+
+        if (this.proxy instanceof HTMLElement) {
+            this.updateForm();
+        }
+
+        const percentage: number =
```
nicholasrice, line 86: should this use that pixelToPercent function above?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+        this.trackMinWidth = this.track.getBoundingClientRect().left;
+    };
+
+    private setupListeners = (): void => {
+        this.addEventListener("keydown", this.keypressHandler);
+        this.addEventListener("mousedown", this.clickHandler);
+        this.thumb.addEventListener("mousedown", this.handleThumbMouseDown);
+    };
+
+    private setupClassForOrientation = (): void => {
```
nicholasrice, line 200: Is there a reason you cannot bind the  class to the root element in the template itself?

Does this not work?
```
<div class="root ${x => x.orientation === SliderOrientation.horizontal ? "slider-horizontal" : "slider-vertical" }" />
```

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+     * Set to true when the component has constructed
+     */
+    private constructed: boolean = false;
+    protected proxy = document.createElement("input");
+
+    constructor() {
+        super();
+
+        this.proxy.setAttribute("type", "range");
+        this.setAttribute("role", "slider");
```
nicholasrice, line 150: You should be able to set these in the template:

````
<template role="slider" tabindex="0">
   // other stuff
</template>

File: packages/fast-components/src/slider/slider.ts [hunk 6]
```diff
@@ -0,0 +1,327 @@
...
+
+    private setupTrackConstraints = (): void => {
+        this.trackWidth = this.track.clientWidth;
+        this.trackMinWidth = this.track.getBoundingClientRect().left;
+    };
+
+    private setupListeners = (): void => {
+        this.addEventListener("keydown", this.keypressHandler);
+        this.addEventListener("mousedown", this.clickHandler);
+        this.thumb.addEventListener("mousedown", this.handleThumbMouseDown);
```
nicholasrice, line 197: You don't remove the thumb listener in teardown but probably should

packages/fast-components/src/slider-label/slider-label.template.ts
```diff
@@ -0,0 +1,28 @@
+import { html, when, ref } from "@microsoft/fast-element";
+import { SliderLabel } from "./slider-label";
+
+export const SliderLabelTemplate = html<SliderLabel>`
+    <div
+        ${ref("root")}
+        part="slider-label"
+        class="slider-label"
```
chrisdholt, line 8: does this need a prefix of slider? For part, the css selector would look like:

```
fast-slider::part(slider-label) {}
```

I think we can leave the prefix off here and just go with `label`:
```
fast-slider::part(label) {
    color: pink;
}
```

packages/fast-components/src/slider-label/slider-label.template.ts
```diff
@@ -0,0 +1,28 @@
+import { html, when, ref } from "@microsoft/fast-element";
+import { SliderLabel } from "./slider-label";
+
+export const SliderLabelTemplate = html<SliderLabel>`
+    <div
```
chrisdholt, line 5: Could this div be the host element instead?

packages/fast-components/src/slider-label/slider-label.template.ts
```diff
@@ -0,0 +1,28 @@
...
+                html`
+                <div class="mark">
+                </div>    
+            `
+            )}
+            <div class="label-positioner">
+                <slot name="label">
+                    <span class="label"}>
+                        ${x => x.label}
+                    </span>
```
chrisdholt, line 23: Should we have the slot wrapping the span or the span wrapping the slot?

packages/fast-components/src/slider-label/slider-label.template.ts
chrisdholt: Additional thought - this seems like quite a bit of markup for the label. Is it all necessary?Do we need a containier, positioner and a nested span to create a slider label?

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,58 @@
+import { attr, FastElement, observable } from "@microsoft/fast-element";
+import { FASTSlider } from "../slider";
+import { SliderOrientation } from "../slider/slider";
+
+export class SliderLabel extends FastElement {
+    @observable
+    public positionStyle: string;
+
+    @observable
+    public root: HTMLDivElement;
```
chrisdholt, line 10: I don't think you need to observe refs (@eisenbergEffect to keep me honest)

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    @attr
+    public max;
+
+    @attr
+    public position: string;
+
+    @attr
+    public label: string;
```
chrisdholt, line 19: Do any of these need to have `changed` events?

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    @attr({ attribute: "hide-mark", mode: "boolean" })
+    public hideMark: boolean = false;
```
chrisdholt, line 22: what happens if this changes after instantiation?

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    private sliderMax: number;
+    private sliderMin: number;
+    private sliderOrientation: SliderOrientation;
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.sliderMax = (this.parentNode as FASTSlider).max;
+        this.sliderMin = (this.parentNode as FASTSlider).min;
```
chrisdholt, line 31: I think this is a clever implementation, but what happens if I use this with my own slider which is not FASTSlider?

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,58 @@
...
+    private styleForOrientation = (): void => {
+        if (this.sliderOrientation === SliderOrientation.horizontal) {
+            this.root.classList.add("slider-label-horizontal");
+        } else {
+            this.root.classList.add("slider-label-vertical");
+        }
+    };
+
+    private positionAsStyle = (): any => {
+        const direction: string = (this.parentNode as FASTSlider).direction;
```
chrisdholt, line 46: is this guaranteed to have a direction property?

packages/fast-components/src/slider/slider.stories.ts
```diff
@@ -0,0 +1,14 @@
...
+import addons from "@storybook/addons";
+
+// Prevent tree-shaking
+FASTSlider;
+FASTDesignSystemProvider;
+
+export default {
+    title: "Slider",
+};
+export const Base = () => Examples;
```
chrisdholt, line 14: Is `Base` a safe const here to be exporting?

packages/fast-components/src/slider/slider.styles.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export const SliderStyles = css`
+    :host {
+        --thumb-size: calc(var(--height-number) * 0.5);
+        --thumb-translate: calc(var(--thumb-size) * 0.5);
+        --track-overhang: calc((var(--design-unit) / 2) * -1);
+    }
+    :host([hidden]) {
+        display: none;
+    }
```
chrisdholt, line 14: These should be able to drop under the host implementation on line 16. The display none behavior is inherit with the display helper.

packages/fast-components/src/slider/slider.styles.ts
```diff
@@ -0,0 +1,135 @@
...
+    ${display("inline-grid")} :host {
+        align-items: center;
+        outline: none;
+        width: 100%;
+        margin: calc(var(--design-unit) * 1px) 0;
+        ${
+            /*
+           * Chromium likes to select label text or the default slot when
+           * the checkbox is clicked. Maybe there is a better solution here?
+           */ ""
```
chrisdholt, line 25: copy paste - is this an issue here?

packages/fast-components/src/slider/slider.styles.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+    .thumb-cursor:active {
+        background: var(--neutral-foreground-active);
+    }
+    .slider-horizontal .thumb-container {
+        transform: translateX(calc(var(--thumb-translate) * 1px));
+    }
+    .slider-vertical .thumb-container {
+        transform: translateY(calc(var(--thumb-translate) * 1px));
+    }
```
chrisdholt, line 70: Is this from a previous implementation or are we also supporting vertical with this first rev?

packages/fast-components/src/slider/slider.template.ts
```diff
@@ -0,0 +1,45 @@
...
+    role="slider"
+    aria-valuenow="${x => x.value}"
+    aria-valuemin="${x => x.min}"
+    aria-valuemax="${x => x.max}"
+  >
+    <div
+          ${ref("root")}
+          part="slider"
+          class="slider"
+    >
```
chrisdholt, line 16: Do we need this or could this be the host element?

packages/fast-components/src/slider/slider.template.ts
```diff
@@ -0,0 +1,45 @@
...
+      <div part="layout-region" class="layout-region">
```
chrisdholt, line 17: we use `positioning region` in dialog, should we align?

packages/fast-components/src/slider/slider.template.ts
```diff
@@ -0,0 +1,45 @@
...
+          <div ${ref("track")} part="track-container" class="track">
+            <slot name="track"></slot>
+          </div>
+          <div ${ref("foregroundTrack")} part="foreground-track" class="foreground-track">
+            <slot name="foreground-track"></slot>
+          </div>
+          <div style="height: 1px"></div>
```
chrisdholt, line 24: This is always and forever 1px?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+    keyCodeArrowDown,
+    keyCodeArrowUp,
+} from "@microsoft/fast-web-utilities";
+import { convertPixelToPercent } from "./slider-utilities";
+
+export enum SliderMode {
+    singleValue = "single-value",
+    adjustFromLower = "adjust-from-lower",
+    adjustFromUpper = "adjust-from-upper",
+    dualThumb = "dual-thumb",
```
chrisdholt, line 15: If we aren't supporting all of this out of the box it may be best to not include it until it's supported. Slider mode is simply "singleValue" which is the default.

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+    public thumb: HTMLDivElement;
+
+    @observable
+    public upperThumb: HTMLDivElement;
+
+    @observable
+    public lowerThumb: HTMLDivElement;
+
+    @observable
+    public direction: string = "ltr";
```
chrisdholt, line 56: Should this be a string or map to a global enum? I believe we have direction in `fast-web-utilities`

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+    protected proxy = document.createElement("input");
+
+    constructor() {
+        super();
+
+        this.proxy.setAttribute("type", "range");
+        this.setAttribute("role", "slider");
+        this.setAttribute("tabindex", "0");
+        this.direction = this.getDirection();
+        this.constructed = true;
```
chrisdholt, line 153: depending on what you're using this for, you may not need it. Per @EisenbergEffect the $emit helper only runs after construction.

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+                break;
+            case keyCodeArrowLeft:
+            case keyCodeArrowDown:
+                this.decrement();
+                break;
+        }
+    };
+
+    private getDirection = (): string => {
+        const dirNode: any = this.parentElement!.closest("div[dir]");
```
chrisdholt, line 185: A couple of questions here:

1. Is this `any`? I think we'd have a typing for this.
2. You're looking for a div, that seems like a strong assumption. I think you likely want to traverse the DOM to check for a dir attribute.

chrisdholt: Ultimately, what happens if I set the dir attribute on the `body` and leave it. I think this would fail.

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+        return dirNode !== null && dirNode["dir"] === "rtl" ? "rtl" : "ltr";
+    };
+
+    private setupTrackConstraints = (): void => {
+        this.trackWidth = this.track.clientWidth;
+        this.trackMinWidth = this.track.getBoundingClientRect().left;
```
chrisdholt, line 191: is minwidth always equal to the left value of the bounding client rect? RTL?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+
+    private stopDragging = (): void => {
+        this.isDragging = false;
+        window.removeEventListener("mouseup", this.handleWindowMouseUp);
+        window.removeEventListener("mousemove", this.handleMouseMove);
+    };
+
+    private clickHandler = (e: MouseEvent) => {
+        if (!this.disabled && !this.readOnly) {
+            let trackElement: any = this.shadowRoot!.querySelector(".track");
```
chrisdholt, line 272: Question: could you use a ref for this?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+            this.trackWidth = trackElement.clientWidth;
+            if (this.trackWidth === 0) {
+                this.trackWidth = 1;
+            }
+            this.trackMinWidth = trackElement.getBoundingClientRect().left;
```
chrisdholt, line 277: is this different than what is being stored while setting up constraints (has it changed?)?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,327 @@
...
+
+            e.preventDefault();
+            (e.target as HTMLElement).focus();
+            window.addEventListener("mouseup", this.handleWindowMouseUp);
+            window.addEventListener("mousemove", this.handleMouseMove);
+
+            this.value = `${this.calculateNewValue(e.pageX)}`;
```
chrisdholt, line 284: It looks like pageX was redefined as part of the standard, does it work as expected in Firefox? Also, is X consistent in RTL?

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: I think this is why I have the extra dom you mentioned so I can control the positioning of whatever is slotted here, the span is the default content, it has no positioning, thus the thing wrapping the slot and its default content

packages/fast-components/src/slider/slider.ts
marjonlynch: This is how it worked before, I'm not sure it would work any less here

packages/fast-components/src/slider-label/slider-label.ts
EisenbergEffect: No need to observe refs, since the template engine is just setting that property, rather than needing to update based on when it changes.

packages/fast-components/src/slider/slider.ts
marjonlynch: yes, I think I did this before I started using refs. will change

packages/fast-components/src/slider-label/slider-label.ts
EisenbergEffect: Maybe worth extracting an interface that defines the minimal properties needed to use the label with a slider. I think it's just min and max. Things like orientation and direction could be optional on that interface and this code could assume defaults if not present.

packages/fast-components/src/slider/slider.ts
EisenbergEffect: Per above, you can remove `@observable` from anything that is a `ref`.

packages/fast-components/src/slider/slider.ts
EisenbergEffect: Yep, I think `constructed` and checks can probably be removed now.

packages/fast-components/src/slider/slider.ts
EisenbergEffect: It should work on a div. We don't want to do this on `template` though since our class binding isn't smart enough not to conflict with user classes. But if this component fully controls the div and only changes classes through the binding, this should work.

packages/fast-components/src/slider-label/slider-label.template.ts
chrisdholt: Actually, I think this is going to act like the default value for the slot. If I pass my own slotted content I think it'll take the place here.

packages/fast-components/src/slider/slider.ts
marjonlynch: done

packages/fast-components/src/slider/slider.ts
marjonlynch: done

packages/fast-components/src/slider/slider.ts
marjonlynch: It does work, I just didn't think of doing it that way. checked it in as such, I have no idea which is more efficient or better or what but it's this way now

packages/fast-components/src/slider/slider-utilities.ts
marjonlynch: done, thanks!

packages/fast-components/src/slider-label/slider-label.styles.ts
marjonlynch: I think I did this correctly, might want to double check it. I imported display and set it as I thought it should be for slider-label

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: done, I believe as we discussed.

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: done, it does not need to be there as suggested.

packages/fast-components/src/slider/slider.ts
marjonlynch: good catch! the answer is it has not changed, so removed.

packages/fast-components/src/slider/slider.ts
marjonlynch: ok, so it turns out I didn't need to put div[dir], I think just [dir] will traverse up as far as you can traverse manually.

marjonlynch: Also if that is the case I think that justifies using any, unless I can use HTMLElement or something

chrisdholt: Looks like `Element | null` may be the best typing: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest.

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: max shouldn't have been there, position now has a changed event handler.

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: I've fixed this to deal with there not being a parent fast-slider, but now I'll do the interface thing.

packages/fast-components/src/slider/slider.styles.ts
marjonlynch: this is from a previous iteration but, I was planning to have the styles and other things in place to support vertical orientation but create an issue for it not working and track that in this PR or separately.

packages/fast-components/src/slider/slider.template.ts
marjonlynch: Sounds good, we shall align.

packages/fast-components/src/slider/slider.ts
marjonlynch: ok that makes sense

packages/fast-components/src/slider/slider.styles.ts
chrisdholt: My only concern with this is that, without an implementation to map this too - we can't validate that it's a good solution for the problem. I'd prefer to implement it as part of the vertical orientation work because the code being here kind of implies that it's _good_ and _correct_.

packages/fast-components/src/slider/slider.ts
marjonlynch: I got rid of constructed and the setAttrs moved to the template

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: interface in place

packages/fast-components/src/slider/slider.stories.ts
marjonlynch: exporting with an appropriate name now

packages/fast-components/src/slider/slider.styles.ts
marjonlynch: done and done

packages/fast-components/src/slider/slider.template.ts
marjonlynch: done

packages/fast-components/src/slider/slider.ts
marjonlynch: done

packages/fast-components/src/slider/slider.ts
marjonlynch: done

packages/fast-components/src/slider/slider.template.ts
marjonlynch: removed

packages/fast-components/src/slider/slider.ts
marjonlynch: The track doesn't actually change in rtl or ltr, so this should always work for horizontal anyway.

packages/fast-components/src/slider/slider.styles.ts
marjonlynch: removed

packages/fast-components/src/slider/slider.ts
marjonlynch: the pageX should not be an issue in RTL, that should be working and accounted for. With regards to firefox there is currently another bug being fixed in the framework to account for something firefox does causing connectedCallback to happen before the constructor is called. So I can't test in firefox yet.

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: If you are talking about position that has been fixed, if you are talking about hideMark I'm not sure if anything needs to be done.

packages/fast-components/src/slider/slider.styles.ts
```diff
@@ -0,0 +1,129 @@
...
+        user-select: none;
+    }
+    .slider {
+        box-sizing: border-box;
+        border-radius: calc(var(--corner-radius) * 1px);
+        outline: none;
+        cursor: pointer;
+    }
+    .foreground-track {
+        display: none;
```
eljefe223, line 29: Is this mean't to always be hidden?

marjonlynch: this is for the other modes that aren't being implemented in this first version so I should maybe just get rid of it

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: moved that root div up into host, cleaned up a bit more

packages/fast-components/src/slider/slider.styles.ts
marjonlynch: removed it and other stuff related to dual mode which won't be in this pr

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: I believe this has been changed in the code accordingly. part(label) will refer to just the label part of this component, all of what was "slider-label..." has been moved up into :host

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: modified to use interface and have defaults if parent does not implement said interface

packages/fast-components/src/slider-label/slider-label.styles.ts
```diff
@@ -0,0 +1,58 @@
+import { css } from "@microsoft/fast-element";
+import { disabledCursor, display } from "../styles";
+import { SystemColors } from "../styles/system-colors";
+import { heightNumber } from "../styles/size";
+
+export const SliderLabelStyles = css`
+    ${display("block")} :host {
+        --half-thumb-size: calc(${heightNumber} * 0.25);
+    }
+    .slider-label {
```
nicholasrice, line 10: needed?

packages/fast-components/src/slider-label/slider-label.template.ts
```diff
@@ -0,0 +1,27 @@
+import { html, when } from "@microsoft/fast-element";
+import { SliderLabel } from "./slider-label";
+import { SliderOrientation } from "../slider";
+
+export const SliderLabelTemplate = html<SliderLabel>`
+    <template
+        class=${x =>
```
nicholasrice, line 7: I think *this* is what Rob was referring to - we don't want to reflect to the "class" attribute here because it will clobber any class attribute applied by an app author.

packages/fast-components/src/slider-label/slider-label.template.ts
```diff
@@ -0,0 +1,27 @@
...
+            x.config.orientation === SliderOrientation.horizontal
+                ? "slider-label-horizontal"
+                : "slider-label-vertical"}
+        style=${x => x.positionStyle}
```
nicholasrice, line 11: Hmm style also - this will clobber any inline styles applied by the app author.

packages/fast-components/src/slider-label/slider-label.template.ts
EisenbergEffect: Yes. Hope to fix that but for now, don't set class on the template itself otherwise it will clobber any user classes that they place on the element.

packages/fast-components/src/slider-label/slider-label.template.ts
EisenbergEffect: Affirmative.

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: dang it I had it in code and moved it here because of another comment. Will put it back

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: Will put this one back in class

packages/fast-components/src/slider-label/slider-label.styles.ts
marjonlynch: removed, it was used but only once, if we are talking about --half-thumb-size

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: well wait this one is a problem. the reason I moved it up was to get rid of what was referred to as superfluous dom/div, so I guess it was not superfluous after all? I need to position this control using style.

EisenbergEffect: Technically, you should be able to use JS to add/remove the styles. You just cant use a style binding on the host element. The binding isn't smart enough to add/remove only the styles you provide without mucking with user styles. But....this still might be something that's better with a wrapping div. I can see there still be issues with user-land conflicts.

packages/fast-components/src/slider/slider.ts
marjonlynch: pageX and pageY seem to be working fine in Edge and FireFox. RTL is also not an issue here.

packages/fast-components/src/slider/slider.styles.ts
marjonlynch: vertical orientation is in place now so this shall stay

packages/fast-components/src/slider/slider.template.ts
marjonlynch: moved

packages/fast-components/src/slider-label/slider-label.template.ts
marjonlynch: as you know from offline conversation I got it to work, so style out of host template now.

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,96 @@
...
+    public config: SliderConfiguration = {
+        min: 0,
+        max: 0,
+        direction: Direction.ltr,
+        orientation: SliderOrientation.horizontal,
+    };
+
+    constructor() {
+        super();
+        this.getSliderConfiguration();
```
EisenbergEffect, line 36: This needs to be moved to the connected callback because someone can instantiate the control in JS and then later add it to the DOM.

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,320 @@
...
+    @observable
+    public trackHeight: number = 0;
+    @observable
+    public trackMinHeight: number = 0;
+
+    /**
+     * The element's value to be included in form submission changed.
+     */
+    @attr
+    public value: string; // Map to proxy element.
```
EisenbergEffect, line 58: Should value have a converter to ensure that the property is always a number?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,320 @@
...
+     */
+    @attr
+    public mode: SliderMode = SliderMode.singleValue;
+
+    protected proxy = document.createElement("input");
+
+    constructor() {
+        super();
+        this.proxy.setAttribute("type", "range");
+        this.direction = this.getDirection();
```
EisenbergEffect, line 145: The `getDirection()` function uses parentElement, which might be null if someone creates this in JS and then later add it to the DOM. This is something you'll want to do in the connected callback.

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,320 @@
...
+
+        if (constrainedVal < this.min || constrainedVal > this.max) {
+            // TODO here until we figure out how this happens
+            return Number(this.value);
+        } else {
+            return constrainedVal;
+        }
+    };
+
+    private increment = (): void => {
```
EisenbergEffect, line 295: The `increment` and `decrement` methods seems like nice candidates for public methods that could be invoked in JS.

marjonlynch: done

packages/fast-components/src/slider/slider.ts
marjonlynch: done and done nullableNumberConverter used

packages/fast-components/src/slider/slider.ts
marjonlynch: moved to connectedCallback and if rtl trigger setting value again to get the right calculation result

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: done

packages/fast-components/src/slider-label/fixtures/base.html
```diff
@@ -0,0 +1,8 @@
+<fast-design-system-provider use-defaults>
+    <!-- Basic -->
+    <div style="margin: 20px">
+        <fast-slider-label>
+            test
+        </fast-slider-label>
+    </div>
+</fast-design-system-provider>
```
chrisdholt, line 8: is this example exhaustive enough to ensure we don't regress the control in the future when making changes? Should we have examples passing position and hide-mark?

chrisdholt: Should this include vertical examples as well since you have that included?

marjonlynch: Seems like the slider-label is somewhat exhaustively tested in the slider fixtures. Does it make sense to do that again here? I can try and add a few things, for sure hide-mark.

chrisdholt: I think that's a fair point but it assumes that people will test across controls. I think we should include examples under the assumption that people sometimes develop in isolation and do not do what is right :).

marjonlynch: I added position, vertical and hide-mark examples

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.setStyleForOrientation();
+        this.positionStyle = this.positionAsStyle();
+    }
+
+    private setStyleForOrientation = (): void => {
+        if (this.config.orientation === SliderOrientation.horizontal) {
+            this.classList.add("slider-label-horizontal");
+        } else {
+            this.classList.add("slider-label-vertical");
```
chrisdholt, line 48: do these classes need to be this verbose? Elsewhere we aren't prefixing with the control name. I'd prefer `horizontal` and `vertical` here (elsewhere regarding the prefix as well)

packages/fast-components/src/slider-label/slider-label.ts
```diff
@@ -0,0 +1,97 @@
...
+        }
+    };
+
+    private isSliderConfig(node: any): node is SliderConfiguration {
```
chrisdholt, line 52: nit: is this typing actually `any` ?

packages/fast-components/src/slider/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    name: "fast-slider",
+    template,
+    styles,
+})
+@designSystemConsumer({
+    recipes: [
+        neutralForegroundActive,
+        neutralForegroundHover,
+        neutralForegroundRest,
+        neutralOutlineRest,
```
chrisdholt, line 40: Is neutralOutlineHover missing?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,313 @@
...
+     */
+    @attr
+    public orientation: SliderOrientation = SliderOrientation.horizontal;
+    private orientationChanged(): void {
+        if (this.orientation === SliderOrientation.horizontal) {
+            this.classList.remove("slider-vertical");
+            this.classList.add("slider-horizontal");
+        } else {
+            this.classList.remove("slider-horizontal");
+            this.classList.add("slider-vertical");
```
chrisdholt, line 120: can we drop the prefix on these names?

packages/fast-components/src/slider/slider.ts
```diff
@@ -0,0 +1,313 @@
...
+            case keyCodeArrowRight:
+            case keyCodeArrowUp:
+                this.increment();
+                break;
+            case keyCodeArrowLeft:
+            case keyCodeArrowDown:
+                this.decrement();
+                break;
+        }
+    };
```
chrisdholt, line 193: Should `shift + arrowRight` and others increment or decrement based on the step property?

packages/fast-components/src/slider/slider.ts
marjonlynch: yes, will do

packages/fast-components/src/slider-label/slider-label.ts
marjonlynch: I think this needs to be any seeing as I'm not looking for a type and thus the attribute check inside the function, the node could be anything. At least that's my story

packages/fast-components/src/slider/index.ts
marjonlynch: done

packages/fast-components/src/slider/slider.ts
marjonlynch: apparently not, fixed and added  `home` and `end` key processing

PR #2862, opened 2020-04-01
chore: add flipper spec

chrisdholt:
# Description
This PR proposes adding a new spec for a flipper component.

## Issue type checklist

## Process & policy checklist


specs/flipper/flipper.md
```diff
@@ -0,0 +1,105 @@
...
+
+### API
+
+Extends FAST Element
+
+*Component Name*
+- `fast-flipper`
+
+*Attrs*
+- `hiddenFromAT` - The control is hidden from assistive technology. Defaults to true
```
chrisdholt, line 33: Feedback is welcome on a better name here :). Per previous comments by @EisenbergEffect, something like `hiddenFromAssistiveTechnologies` is a bit verbose. As an FYI, this convention was used here as an initial stub because the same language is often used by the W3C spec to denote when something is not exposed to AT users.

specs/flipper/flipper.md
```diff
@@ -0,0 +1,105 @@
...
+- `disabled` - the control is disabled. There are scenarios where a flipper would be visible and disabled
+- `label` - the accessible label for the control, exposed to AT when not hidden
```
chrisdholt, line 35: I think a first class prop here makes sense for label as most of the time flippers do not use text content but visual representations only. Alternatively, we could remove this and just support users passing an aria-label and document that somewhere.

specs/flipper/flipper.md
nicholasrice: We could maybe do something like: 

```
@attr({attribute: "aria-hidden"})
public hiddenFromAT: boolean = true;
```

Then then user could simply write `<fast-flipper aria-hidden="false" />` to override. Note that this doesn't seem to actually work - aria-hidden is overridden by the default property value but I think that might be a bug. You might also need to do write a custom converter.

specs/flipper/flipper.md
nicholasrice: I'd +1 for just using an `aria-label`. I don't know that exposing a distinct attribute that then maps the value to the aria-label attribute is necessary. It is a scenario we should cover in documentation though for sure.

specs/flipper/flipper.md
EisenbergEffect: Hmm. Overwriting built-ins may not be something we can support. I'm not sure. It would be nice to use aria-hidden if we can. Do we need to do anything in our own code in response to this changing?

specs/flipper/flipper.md
```diff
@@ -0,0 +1,103 @@
...
+
+*Events*
+- `click` - mimics the click event of traditional buttons
+
+### Anatomy and Appearance
+
+```
+<template
+    role="button" 
+    aria-hidden=${x => x.hiddenFromAT}
```
EisenbergEffect, line 46: Can we just let the end user decide by adding aria-hidden themself or is there something we need to do internally in response to the hiddenFromAT property changing?

chrisdholt: Some of this depends on how we want to default it. If you default it to being exposed to assistive tech, then you need to drop the tabindex and add aria hidden in the event you want to hide it. If you default to hidden from assistive tech, the end user needs to add tabindex, override the internal aria-hidden behavior, and add the role of button. It's shown as persisting above, but I don't think we would add that if it were not exposed to AT. So, in a sense it's a swing of 3 properties on the implementors side to successfully transition between these states. I see a single prop to toggle this as a value add for empowering accessibility; however, we'd technically be empowering it as well by making someone handle it themselves.

specs/flipper/flipper.md
nicholasrice: I think it just needs to be hidden *by default* - perhaps when connected we could just check to see if the attribute exists, and if not set it imperatively.

EisenbergEffect: That's an interesting idea. What do you think @chrisdholt ?

chrisdholt: So, if `aria-hidden` exists on the host, use that value and respond accordingly. If it doesn't exist, set aria-hidden in connectedCallback?

chrisdholt: @nicholasrice @EisenbergEffect just want to confirm I'm understanding the proposed implementation (above)

chrisdholt: Confirmed offline, I like this idea and think we should pursue it in implementation. 👍

PR #2863, opened 2020-04-02
chore: add divider spec

chrisdholt:
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/divider.md
+++ b/specs/divider.md
@@ -1,0 +1,56 @@
+# Divider
+
+## Overview
+
+An implementation of a horizontal rule as a web-component. 
+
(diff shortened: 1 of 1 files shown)


specs/divider.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+
+### Accessibility
+
+The divider should default to having a role of `separator`, just as a typical horizontal rule would. When changing the look, functional, interactive, or structural relevance implied by the `hr` element a role of `presentation` may be applied. Since the role of [separator](https://w3c.github.io/aria/#separator) conveys meaning to assistive technology, the visual representation should likewise convey the same meaning. With that in mind, the visual treatment of a horizontal rule with a role of `separator` should meet non-text contrast requirements. A role of `presentation` implies that the divider is a visual treatment only so the contrast requirement does not apply in that case.
+
+
+## Next Steps
+- Monitor the status of [`is` global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is)
```
nicholasrice, line 56: Can you add some rational for *why* we're monitoring the status of that API so we have some context looking back at this?

PR #2866, opened 2020-04-02
feat: more precise drag and drop allowed before and after linked data

janechu:
# Description

This change allows linked data to be dragged before and after other linked data. It also greatly reduces the complexity of the dragging elements, limiting them to only the single `<span>` or `<a>` elements that represent the "triggers". This assists in reducing the overlap that was happening  between elements as drop targets via the `react-dnd` package.

Additionally in order to allow an item to be dragged before or after another item, an additional property was added to the message systems `addLinkedData` action to specify the index that the linked data could be spliced into.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
+++ b/packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
@@ -118,6 +118,7 @@ export const DraggableNavigationTreeItem: React.FC<NavigationTreeItemProps> = ({
         },
         end(): void {
             // TODO: investigate why when not dropped on a drop target this takes extra time to respond
+            // see issue: https://github.com/microsoft/fast-dna/issues/2867
             dragEnd();
         },
     });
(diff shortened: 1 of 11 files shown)


packages/fast-tooling-react/src/navigation/navigation-tree-item.tsx
```diff
@@ -151,6 +117,7 @@ export const DraggableNavigationTreeItem: React.FC<NavigationTreeItemProps> = ({
             dragStart(dictionaryId);
         },
         end(): void {
+            // TODO: investigate why when not dropped on a drop target this takes extra time to respond
```
chrisdholt, line 120: nittiest of nits (worst of the worst comments): Should we have a task for this, or are you already tracking in a branch?

janechu: No, we should have a task, I forgot I put that in there, fixed!

PR #2868, opened 2020-04-02
chore: add tooltip spec

chrisdholt:
# Description
This PR proposes adding a new spec for `tooltip`. This work is dependent upon the anchored region.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/tooltip.md
+++ b/specs/tooltip.md
@@ -63,7 +63,7 @@ Tooltip widgets do not receive focus. A hover that contains focusable elements c
 
 ### Accessibility
 
-The interaction model for the tooltip should map to the [W3C interactioni model](https://w3c.github.io/aria-practices/#tooltip).
+The interaction model for the tooltip should map to the [W3C interaction model](https://w3c.github.io/aria-practices/#tooltip).
 
 Two important things to note:
 1. Focus stays on the triggering element while the tooltip is displayed.
--- a/specs/tooltip.md
+++ b/specs/tooltip.md
@@ -1,0 +1,74 @@
+# Tooltip
+
+## Overview
+
+A tooltip is a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It typically appears after a small delay and disappears when Escape is pressed or on mouse out.
+
+### Use Cases
+
+Tooltips are often used to supplement UI elements and provide additional information. Tooltips will often provide accessible names or add additional context to other UI components (such as icon buttons). While tooltips are natively provided to elements via the title attribute, they are notoriously difficult to style. Additionally, they are only invoked on hover and not on focus.
+
+### Non-goals
+Tooltip widgets do not receive focus. A hover that contains focusable elements can be made using a non-modal dialog (flyout).
+
+### Prior Art/Examples
+- [Lightning Design](https://www.lightningdesignsystem.com/components/tooltips/)
+- [Office Fabric](https://developer.microsoft.com/en-us/fabric#/controls/web/tooltip)
+- [Evergreen](https://evergreen.segment.com/components/tooltip/)
+- [Carbon Design](https://www.carbondesignsystem.com/components/tooltip/code/)
+- [Ant Design](https://ant.design/components/tooltip/)
+- [Atlassian](https://atlaskit.atlassian.com/packages/core/tooltip)
+- [Bootstrap](https://getbootstrap.com/docs/4.3/components/tooltips/)
+
+---
+
+### API
+
+*Component name:*
+- `fast-tooltip`
+
+*Attributes:*
+- `trigger` - the html id of the HTMLElement which triggers the tooltip. This element references the tooltip element with aria-describedby.
+- `delay` - time in milliseconds to wait before showing and hiding the tooltip. Defaults to 300.
+- `visible` - the visiblity of the tooltip
+- `position` - enum; where the tooltip should appear relative to its target.
+    - bottom
+    - left
+    - right
+    - top
+
+*Slots:*
+- `default`
+
+*Events*
+- dismiss - event fired when a user presses escape
+
+### Anatomy and Appearance
+
+*Template:*
+```
+<template
+    role="tooltip"
+>
+    <slot></slot>
+</template>
+```
+
+## Implementation
+
+```
+<button id="foo" aria-describedby="tooltip">Foo</button>
+<fast-tooltip trigger="foo" id="tooltip">This is a tooltip. It does not recieve focus.</fast-tooltip>
+```
+
+### Accessibility
+
+The interaction model for the tooltip should map to the [W3C interactioni model](https://w3c.github.io/aria-practices/#tooltip).
+
+Two important things to note:
+1. Focus stays on the triggering element while the tooltip is displayed.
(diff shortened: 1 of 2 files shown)


specs/tooltip.md
```diff
@@ -0,0 +1,74 @@
...
+
+---
+
+### API
+
+*Component name:*
+- `fast-tooltip`
+
+*Attributes:*
+- `trigger` - the html id of the HTMLElement which triggers the tooltip. This element references the tooltip element with aria-describedby.
```
nicholasrice, line 31: The anchored-region uses the "anchor" keyword for this same purpose - perhaps we should align?

specs/tooltip.md
```diff
@@ -0,0 +1,74 @@
...
+    role="tooltip"
+>
+    <slot></slot>
+</template>
+```
+
+## Implementation
+
+```
+<button id="foo" aria-describedby="tooltip">Foo</button>
```
nicholasrice, line 60: That's such a bummer that these need to cross-reference each other - I wish there was an `aria-describes` attribute.

specs/tooltip.md
```diff
@@ -0,0 +1,74 @@
...
+<fast-tooltip trigger="foo" id="tooltip">This is a tooltip. It does not recieve focus.</fast-tooltip>
+```
+
+### Accessibility
+
+The interaction model for the tooltip should map to the [W3C interactioni model](https://w3c.github.io/aria-practices/#tooltip).
```
nicholasrice, line 66:
```suggestion
The interaction model for the tooltip should map to the [W3C interaction model](https://w3c.github.io/aria-practices/#tooltip).
```

specs/tooltip.md
chrisdholt: Same - I wonder if anything like that has ever been floated?

specs/tooltip.md
chrisdholt: I think we can; I used the language from a few other design systems that have tooltips here, but we can align with our terminology used in anchored region.

specs/tooltip.md
```diff
@@ -0,0 +1,74 @@
...
+
+---
+
+### API
+
+*Component name:*
+- `fast-tooltip`
+
+*Attributes:*
+- `anchor` - the html id of the HTMLElement which triggers the tooltip. This element references the tooltip element with aria-describedby.
```
EisenbergEffect, line 31: I think we'll want to follow the same pattern here as @scomea is using with the positioner element. We'll want a property that references the associated element as well as an attribute that takes an id. The attribute would look up by id and set the property. This also allows folks to set the associated element directly in code if they have a reference (or through a templating engine that can do that).

specs/tooltip.md
EisenbergEffect: Could we do this automatically in the tooltip? So, when the tooltip's associated control changes, if that control doesn't have the described-by attribute, we set it up. (we can generate a unique id if there isn't an id set as well)

chrisdholt: @EisenbergEffect it seems reasonable that we could use the reference provided to check for the existence of the attribute and then assign it if does not exist.

specs/tooltip.md
```diff
@@ -0,0 +1,74 @@
...
+- `delay` - time in milliseconds to wait before showing and hiding the tooltip. Defaults to 300.
+- `visible` - the visiblity of the tooltip
+- `position` - enum; where the tooltip should appear relative to its target.
```
bheston, line 34: Can this be based on the viewport positioner as well? It seems like a desired position would be good, but that we adjust if it's out of bounds.

chrisdholt: Per the PR description this work is dependent on Anchored Region :). We will not expose everything from Anchored Region though; this will start with very high-level and basic support to keep this positioned within the viewport.

specs/tooltip.md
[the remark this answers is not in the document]
chrisdholt: Do we need to elevate that as part of the API or is there a default viewport we can trust/use? I'd like to start by modeling a super simple API that expects this is attached to a "element" within the natural viewport. I'd like to shadow the added complexity of Anchored Region at the outset here. We can always expose more functionality, but once we release the API it's harder to bring things back.

specs/tooltip.md
```diff
@@ -0,0 +1,74 @@
...
+
+### API
+
+*Component name:*
+- `fast-tooltip`
+
+*Attributes:*
+- `anchor` - the html id of the HTMLElement which triggers the tooltip. This element references the tooltip element with aria-describedby.
+- `delay` - time in milliseconds to wait before showing and hiding the tooltip. Defaults to 300.
+- `visible` - the visiblity of the tooltip
```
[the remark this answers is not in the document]
chrisdholt: It may or may not.

specs/tooltip.md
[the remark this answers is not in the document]
scomea: By default the anchored region will use it's parent element as the viewport which is actually a reasonable base api I think.

specs/tooltip.md
scomea: Should we have some migration plan to ensure that we can get to our richer api without needing to blow up the initial attributes?  Specifically - if we go to a tooltip that can be configured to always be in the same relative position to it's anchor or to choose where there is the most room what happens to position?

specs/tooltip.md
scomea, line 31: nit: I don't think "the html id of the HTMLElement which triggers the tooltip" is the right verbiage.  I think ""the html id of the HTMLElement which the tooltip is attached to" would be more accurate.

specs/tooltip.md
chrisdholt: I may be missing your question, but by not exposing those initially the API is additive and we won't blow up attrs or break folks.

specs/tooltip.md
chrisdholt: good call out, I'll update

specs/tooltip.md
scomea: I think the scenario where you have a single tooltip instance and change it's anchor dynamically would be fairly common, so tooltip should be able to attach/detach from an anchor element.

chrisdholt: Thanks @scomea, I think that would simply be handled by a changed event when the anchor or what not changes which seems implementation heavy to include in the spec - but I agree with the premise outlined above.

specs/tooltip.md
scomea: I guess my question is how do you plan to add to the API without needing to change the initial parts?  In this case how do we get to a place where there is a "position" attribute that is sometimes absolute (ie. always render above) and sometimes a preference (ie. render above, unless there isn't enough room)? For example in anchored region we use "default" position to communicate that the selection isn't always forced.

chrisdholt: I guess I'm not seeing an issue where adding more to a components API is problematic. In my mind, we start with scoped features. We know that a tooltip needs to be relative to its anchor or trigger so we offer options on where you'd like us to try and place it. If we think the positioning logic to keep it visible within the viewport is something that is valuable then that behavior starts as just baked. You can give us a position and we'll do our best to accommodate it so long as there is enough space. I don't think we need to bake every possible positioning feature in as part of an alpha release - we'll get a better idea of what people actually find helpful and useful when they start using it and filing issues they may come across.

specs/tooltip.md
chrisdholt: matched anchored region here

PR #2870, opened 2020-04-02
feat: add flipper as a web component

chrisdholt:
# Description
This PR adds Flipper as a web component.

This is a DRAFT PR which is based on #2862 and depends on that PR being accepted and merged.

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/flipper/flipper.styles.ts
```diff
@@ -0,0 +1,99 @@
...
+export const FlipperStyles = css`
+    ${display("inline-flex")} :host {
+        width: calc(var(--height-number) * 1px);
+        height: calc(var(--height-number) * 1px);
+        justify-content: center;
+        align-items: center;
+        margin: 0;
+        position: relative;
+        fill: var(--neutral-foreground-rest);
+        color: var--neutral-foreground-rest);
```
janechu, line 14: looks like you forgot this "("

chrisdholt: Nice catch - I'll update this once the spec is approved 👍

nicholasrice, line 14:
```suggestion
        color: var(--neutral-foreground-rest);
```

packages/fast-components/src/flipper/flipper.stories.ts
```diff
@@ -0,0 +1,13 @@
+import { FASTFlipper } from "./";
```
awentzel, line 1: As a naming standard, we already know in these stories, classes, etc, that we're dealing with FAST.  Do we need to replicate FAST in the named imports, interfaces, and class names?  Does this help the consumer?  Could we eliminate the prefixed naming for FAST?  We may be able to reduce file size while increasing maintainability in that it's easy to scan the file and understand what's going on.  Just from my fresh perspective.

packages/fast-components/src/flipper/flipper.styles.ts
```diff
@@ -0,0 +1,99 @@
...
+    :host(.disabled) {
+        opacity: var(--disabled-opacity);
+        cursor: ${disabledCursor};
+    }
+
+    :host(:hover)::before {
+        background: var(--neutral-fill-stealth-hover);
+        border-color: var(--neutral-outline-hover);
+    }
+
```
awentzel, line 53: nit: extra CRLF

packages/fast-components/src/flipper/index.ts
```diff
@@ -0,0 +1,14 @@
+import { customElement } from "@microsoft/fast-element";
```
awentzel, line 1: when do we choose "customElement" (lowercase 'c') vs "Flipper" (upper case 'F') in our naming standards?  Have we documented our naming expectations?

packages/fast-components/src/flipper/flipper.stories.ts
chrisdholt: This is importing what is exported to the consumer which I think we do want name spaced. If we want to revisit naming of what is being exported I'd like to raise that as an issue in the repo so we can track independently of this PR.

packages/fast-components/src/flipper/flipper.template.ts
```diff
@@ -0,0 +1,49 @@
+import { html, when } from "@microsoft/fast-element";
+import { Flipper, FlipperDirection } from "./flipper";
+
+export const FlipperTemplate = html<Flipper>`
+<template
+    role="button"
+    tabindex=${x => (x.hiddenFromAT ? -1 : 0)}
+    aria-hidden=${x => (x.hiddenFromAT ? true : void 0)}
```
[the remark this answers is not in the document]
chrisdholt: this implementation is changing based on feedback from the spec. This won't be here after the fact :).

packages/fast-components/src/flipper/index.ts
chrisdholt: These are two different things and both rooted existing standards and common practices. `@customElement` is a decorator which defines our element as a custom element - it's standards based. `Flipper` and similarly `FASTFlipper` are classes which is why they are capitalized.

chrisdholt:
> A suggestion, as a fixture, it may not matter, however, shouldn't we conform to accessibility standards? For example, should we be using an H1 followed by H4 elements? If, we write fixtures as we'd expect in production adhering to accessibility standards we'd get feedback sooner.

I think that's fair; I'm following previous convention that exists within our repo. We can also align the naming of our fixture files for cleanliness. If you'd like to create an issue we can pri making these updates based on your recommendations.

packages/fast-components/src/flipper/flipper.ts
```diff
@@ -0,0 +1,38 @@
...
+            : this.classList.remove("disabled");
+    }
+
+    @attr({ attribute: "aria-hidden", mode: "fromView", converter: booleanConverter })
+    public hiddenFromAT: boolean = true;
+    public hiddenFromATChanged(): void {
+        this.hiddenFromAT === false
+            ? this.setAttribute("tabindex", "0")
+            : this.setAttribute("tabindex", "-1");
+    }
```
EisenbergEffect, line 23: Very cool use of the features here!

packages/fast-components/src/flipper/flipper.template.ts
```diff
@@ -0,0 +1,46 @@
+import { html, when } from "@microsoft/fast-element";
+import { Flipper, FlipperDirection } from "./flipper";
+
+export const FlipperTemplate = html<Flipper>`
+<template
+    role="button"
+    aria-disabled=${x => (x.disabled ? true : void 0)}
```
EisenbergEffect, line 7: Nice!

PR #2888, opened 2020-04-05
feat: add button and anchor as web component

chrisdholt:
# Description
This PR adds `fast-button` and `fast-anchor` as web components.

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/button/anchor.ts
```diff
@@ -0,0 +1,112 @@
...
+                @slotchange=${x => x.handleEndContentChange()}
+            ></slot>
+        </span>
+    </a>
+`;
+
+export class Anchor extends FastElement {
+    @attr
+    public appearance: AnchorAppearance;
+    public appearanceChanged(): void {
```
nicholasrice, line 49: this method gets prev and next values: `public appearanceChanged(prev: string, next: string)` so I don't think you need to track a private currentAppearance

packages/fast-components/src/button/button.ts
```diff
@@ -0,0 +1,126 @@
...
+                name="end"
+                ${ref("end")}
+                @slotchange=${x => x.handleEndContentChange()}
+            ></slot>
+        </span>
+    </button>
+`;
+
+export class Button extends FastElement {
+    @attr
```
nicholasrice, line 52: Buttons are technically form-associated elements - spec has details on implementation with that in mind.

packages/fast-components/src/button/anchor.ts
chrisdholt: Sweet, I’ll update to use the API.

packages/fast-components/src/button/button.ts
chrisdholt: Sounds good, I likely looked right past that. I’ll update.

packages/fast-components/src/styles/patterns/button.ts
```diff
@@ -0,0 +1,235 @@
...
+    ${display("inline-block")} :host {
+        font-family: var(--body-font);
+        outline: none;
+
+        --fast-button-corner-radius: calc(var(--corner-radius) * 1px);
+        --fast-button-fill-rest: var(--neutral-fill-rest);
+        --fast-button-fill-hover: var(--neutral-fill-hover);
+        --fast-button-fill-active: var(--neutral-fill-active);
+
+        --fast-button-foreground-rest: var(--neutral-foreground-rest);
```
nicholasrice, line 15: Are these custom properties carry-over from the initial private implementation? I thought we landed that we didn't want to get heavily involved with component-specific custom properties as an extensibility point, instead preferring selector based overrides -`fast-button.accent {}`, `fast-button::part(end) {}`

packages/fast-components/src/styles/patterns/button.ts
```diff
@@ -0,0 +1,235 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "../display";
+import { focusVisible } from "../focus";
+
+export const BaseButtonStyles = css`
+    ${display("inline-block")} :host {
+        font-family: var(--body-font);
+        outline: none;
```
nicholasrice, line 8: You probably need to rebase - the design-system implementation went in so you'll need to declare which recipes this component depends on.

chrisdholt:
Closing to reimplement based on latest design system updates.

PR #2905, opened 2020-04-09
feat: add button and anchor as web components

chrisdholt:
# Description
This PR adds new web components for `fast-button` and `fast-anchor` to support the visual styling of "Buttons" and "Hypertext".

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/styles/patterns/button.ts
```diff
@@ -0,0 +1,199 @@
...
+        line-height: 20px;
+        line-height: 1;
+        box-sizing: border-box;
+        display: inline-flex;
+        justify-content: center;
+        align-items: center;
+        padding: 0 calc((6 + (var(--design-unit) * 2 * var(--density))) * 1px);
+        height: calc(${heightNumber} * 1px);
+        min-width: calc(${heightNumber} * 1px);
+        max-width: calc(var(--design-unit) * 94.5px));
```
bheston, line 26: I think we should remove the max width. It doesn't add anything and it's an arbitrary value. Basic guidance should say don't make really long buttons, but I don't think there's value in enforcing it.

packages/fast-components/src/styles/patterns/button.ts
```diff
@@ -0,0 +1,199 @@
...
+        font-family: var(--body-font);
+        outline: none;
+    }
+
+    .control {
+        ${
+            /* Font size is temporary - 
+            replace when adaptive typography is figured out */ ""
+        } font-size: 14px;
+        line-height: 20px;
```
bheston, line 17: Probably remove the fixed line-height value and keep the "1".

packages/fast-components/src/button/index.ts
```diff
@@ -0,0 +1,103 @@
...
+        neutralForegroundRest,
+        neutralOutlineActive,
+        neutralOutlineHover,
+        neutralOutlineRest,
+    ],
+})
+export class FASTButton extends Button {}
+export * from "./button";
+
+// Anchor
```
bheston, line 65: I get that these are very similar, but why is this in the same file as a different export name? Seems hard to track down and somewhat against an extensibility model.

packages/fast-components/src/button/button.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @attr({ mode: "boolean" })
+    public autofocus: boolean;
+
+    @attr({ attribute: "form" })
+    public formId: string;
+
+    @attr
+    public formaction: string;
```
bheston, line 72: Can these be camel case? This is hard to read.

bheston: Actually, these all seem to shadow native properties. Is there any way we can promote those or declare them in one place, similarly to what I think we had from React's native HTML elements. I can see benefits to proxying the underlying attributes, but it also seems potentially limiting if new features are added or if we don't add an uncommon one and someone wants to use it. Do they have to create a custom override component?

packages/fast-components/src/styles/patterns/button.ts
chrisdholt: I'm for this

packages/fast-components/src/styles/patterns/button.ts
chrisdholt: sounds good

packages/fast-components/src/button/button.ts
chrisdholt: Yes, these shadow native properties. We are not at this time able to extend from the interface of an HTMLButtonElement because it will not work across all browsers. It's possible that we'll have that support in the future, but currently we need to declare and shadow the native API.

packages/fast-components/src/button/anchor.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export type AnchorAppearance = ButtonAppearance | "hypertext";
+
+export const anchorTemplate = html<Anchor>`
+    <a
+        class="control"
+        download="${x => x.download}"
+        href="${x => x.href}"
+        hreflang="${x => x.hreflang}"
+        ping="${x => x.lang}"
```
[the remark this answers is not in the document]
chrisdholt: ping is incorrect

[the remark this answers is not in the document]
chrisdholt: @EisenbergEffect this has been resolved.

packages/fast-components/src/button/anchor.ts
```diff
@@ -0,0 +1,90 @@
...
+    public referrerpolicy: string;
+
+    @attr
+    public rel: string;
+
+    @attr
+    public target: "_self" | "_blank" | "_parent" | "_top";
+
+    @attr
+    public type: string;
```
nicholasrice, line 73: One thought I had for the start / end work: we could implement them as a mixin so the boilerplate isn't duplicated across components: https://www.typescriptlang.org/docs/handbook/mixins.html

packages/fast-components/src/button/index.ts
nicholasrice: +1 - I'd be for moving this to a separate "anchor" directory

chrisdholt: resolved

packages/fast-components/src/button/anchor.ts
chrisdholt: Can we land this change in a separate PR?

packages/fast-components/src/anchor/anchor.styles.ts
```diff
@@ -0,0 +1,18 @@
...
+    StealthButtonStyles,
+} from "../styles";
+
+export const AnchorStyles = css`
+    ${BaseButtonStyles}
+    ${AccentButtonStyles}
+    ${HypertextStyles}
+    ${LightweightButtonStyles}
+    ${OutlineButtonStyles}
+    ${StealthButtonStyles}
```
janechu, line 17: I thought we were trying to avoid CSS bloat per the spec, did we decide this was no longer necessary?

packages/fast-components/src/anchor/anchor.ts
```diff
@@ -0,0 +1,58 @@
...
+export class Anchor extends FastElement {
+    @attr
+    public appearance: AnchorAppearance = "neutral";
+    public appearanceChanged(
+        oldValue: AnchorAppearance,
+        newValue: AnchorAppearance
+    ): void {
+        if (oldValue !== newValue) {
+            this.classList.add(`${newValue}`);
+            this.classList.remove(`${oldValue}`);
```
janechu, line 15: seems like you could just skip the backtick templating and just do `this.classList.add(newValue)`, etc.

packages/fast-components/src/anchor/anchor.styles.ts
chrisdholt: @EisenbergEffect proposed a registration methodology which would support minimizing bloat incurred by this by being able to register and unregister styles when used so we can avoid having 5 different components. The thread is here: https://github.com/microsoft/fast-dna/pull/2844#discussion_r398941666

packages/fast-components/src/anchor/anchor.ts
chrisdholt: updated

Issue #2893, opened 2020-04-07
`disabled` attributes in fast-components have regressed


chrisdholt:
**Describe the bug; what happened?**
It appears that disabled attributes across form controls which extend from  Form Associated element have regressed and the attribute is no longer being applied.

**What are the steps to reproduce the issue?**

Example:
1. Boot the `fast-components` package
2. Click on `checkbox`
3. Scroll down to disabled examples
4. See that disabled checkboxes are not disabled

**What behavior did you expect?**
When the disabled attribute is applied to the DOM or given a value of `true`, the disabled attribute should be applied. 

```html
<fast-checkbox disabled></fast-checkbox> // should be true

<fast-checkbox disabled="true"></fast-checkbox> // should be true
```

**Is there any additional context?**
All form associated elements seem to have this issue. Disabled is set in Form Associated as a boolean attr with a default value of `false`.

Removing the default value of `false` which is set in Form Associated fixes the issue, but I would expect the behavior of a default value to be the same as a value coming from the view.

EisenbergEffect:
Fixed in #2903

PR #2906, opened 2020-04-09
fix: switch size

eljefe223:
# Description
This fixes `fast-switch` size and disabled state. Disabled was also broken in `fast-checkbox` so fixing it here. Also there was an extra closing tag in `fast-badge` base.html so I removed it.

After:
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/fast-components/src/badge/fixtures/base.html
+++ b/packages/fast-components/src/badge/fixtures/base.html
@@ -54,7 +54,7 @@
     <fast-badge fill="secondary" color="foo" id="foo" circular="true">
         99
     </fast-badge>
-    <fast-badge fill="tertiary" color="bar" id="foo" circular="true"></fast-badge>
+    <fast-badge fill="tertiary" color="bar" id="foo" circular="true">
         9999999
     </fast-badge>
     <h4>Examples</h4>
--- a/packages/fast-components/src/badge/fixtures/base.html
+++ b/packages/fast-components/src/badge/fixtures/base.html
@@ -54,7 +54,7 @@
     <fast-badge fill="secondary" color="foo" id="foo" circular="true">
         99
     </fast-badge>
-    <fast-badge fill="tertiary" color="bar" id="foo" circular="true">
+    <fast-badge fill="tertiary" color="bar" id="foo" circular="true"></fast-badge>
         9999999
     </fast-badge>
     <h4>Examples</h4>
--- a/packages/fast-components/src/checkbox/checkbox.ts
+++ b/packages/fast-components/src/checkbox/checkbox.ts
@@ -38,9 +38,6 @@ export class Checkbox extends FormAssociated<HTMLInputElement> {
         this.defaultChecked = this.checkedAttribute;
     }
 
-    @attr
-    public disabled: boolean;
-
     /**
      * Initialized to the value of the checked attribute. Can be changed independently of the "checked" attribute,
      * but changing the "checked" attribute always additionally sets this value.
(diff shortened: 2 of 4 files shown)


packages/fast-components/src/checkbox/checkbox.ts
```diff
@@ -38,6 +38,9 @@ export class Checkbox extends FormAssociated<HTMLInputElement> {
         this.defaultChecked = this.checkedAttribute;
     }
 
+    @attr
+    public disabled: boolean;
+
```
chrisdholt, line 43: @eljefe223 this seems unrelated; can we remove?

chrisdholt: Also, this is inheriting from form associated - if you were seeing an issue with disabled it was related to #2893 which has since been fixed.

packages/fast-components/src/checkbox/checkbox.ts
```diff
@@ -38,6 +38,9 @@ export class Checkbox extends FormAssociated<HTMLInputElement> {
         this.defaultChecked = this.checkedAttribute;
     }
 
+    @attr
+    public disabled: boolean;
```
EisenbergEffect, line 42: Do we need to make this `@attr({ mode: 'boolean' })` or `@attr({mode: 'fromView', converter: booleanConverter })` Ping @chrisdholt for the latest recommendation based on new components.

packages/fast-components/src/switch/switch.ts
```diff
@@ -33,6 +33,9 @@ export class Switch extends FormAssociated<HTMLInputElement> {
         this.defaultChecked = this.checkedAttribute;
     }
 
+    @attr
+    public disabled: boolean;
```
EisenbergEffect, line 37: Same question as above.

packages/fast-components/src/checkbox/checkbox.ts
eljefe223: Ok, Didn't realize their was a bug tracking, adding this attribute fixed this but I can remove.

packages/fast-components/src/badge/fixtures/base.html
```diff
@@ -54,7 +54,7 @@ <h4>Circular</h4>
     <fast-badge fill="secondary" color="foo" id="foo" circular="true">
         99
     </fast-badge>
-    <fast-badge fill="tertiary" color="bar" id="foo" circular="true"></fast-badge>
+    <fast-badge fill="tertiary" color="bar" id="foo" circular="true">
         9999999
     </fast-badge>
```
chrisdholt, line 59: I missed this one - but we can probably do in a different PR

packages/fast-components/src/switch/switch.styles.ts
```diff
@@ -63,8 +64,8 @@ export const SwitchStyles = css`
 
     .checked-indicator {
         position: absolute;
-        height: calc((${heightNumber} - (var(--design-unit) * 2)) * 1px);
-        width: calc((${heightNumber} - (var(--design-unit) * 2)) * 1px);
+        height: calc((${heightNumber} - (var(--design-unit) * 5.5)) * 1px);
```
bheston, line 67: I'm not going to block on this since I'm behind in working on the spec for handling density, but I'd like to see us calculate sizes by something that provides some declarative way to understand what's going on and avoid magic number or fractions that mostly happen to produce the desired value. I'll look into this more holistically though.

PR #2909, opened 2020-04-09
feat: add anchored region as web component

scomea:
# Description
Adds anchored region to the web component collection.

## Motivation & context
https://github.com/microsoft/fast-dna/blob/master/specs/anchored-region/anchored-region.md

## Issue type checklist

## Process & policy checklist


packages/fast-components/src/anchored-region/anchored-region.template.ts
```diff
@@ -0,0 +1,14 @@
+import { html, ref, when } from "@microsoft/fast-element";
+import { AnchoredRegion } from "./anchored-region";
+
+export const AnchoredRegionTemplate = html<AnchoredRegion>`
+  <template>
```
nicholasrice, line 5: You don't need the template tag here if you're not adding any attributes to it.

packages/fast-components/src/anchored-region/anchored-region.template.ts
```diff
@@ -0,0 +1,14 @@
...
+      <div
+          class="region"
```
nicholasrice, line 7: Should region be able to be styled externally? If so, it might be worth adding a `part="region"` to it

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
...
+
+        this.updateLayout();
+    }
+
+    public disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.disconnectObservers();
+    }
+
+    adoptedCallback() {
```
nicholasrice, line 239: I dunno, *do* you?

scomea: The "to do" items are why it's still draft.  I believe we should do a reset but need to revisit to make sure.

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
+import { attr, FastElement, observable, ref } from "@microsoft/fast-element";
+import { isNil } from "lodash-es";
+import rafThrottle from "raf-throttle";
```
nicholasrice, line 3: I don't think raf-throttle is an enumerated dependency. Neither is lodash-es. This is likely only working because of Yarn workspaces resolving to the root node_modules dir.

I think we can avoid taking the dep on raf-throttle though and hook into `fast-element`'s `DOM.queueUpdate` method

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
...
+    adoptedCallback() {
+        // TODO: do we need to handle this?
+    }
+
+    /**
+     * event thrown when the region's position changes
+     */
+    positionChanged() {
+        this.dispatchEvent(
+            new CustomEvent("changed", {
```
nicholasrice, line 248: so close to native "change" event name - is there a reason to use "changed" and not "change"?

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
...
+        this.disconnectObservers();
+        this.setInitialState();
+        this.connectObservers();
+    }
+
+    /**
+     * sets the starting configuration for component internal values
+     */
+    private setInitialState = (): void => {
+        this.initialLayoutComplete = false;
```
nicholasrice, line 277: Any particular reason why you're not initializing these when they're being defined?

scomea: Because we need to be able to reset the component to its initial state under certain circumstances, like when the anchor element changes, and it seemed best to define these in only one place.

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
...
+                    break;
+            }
+        }
+    };
+
+    /**
+     *  Handle region resize events
+     */
+    private handleRegionResize = (): void => {
+        const regionHeight: number = this.region !== null ? this.region.clientHeight : 0;
```
nicholasrice, line 483: Doesn't seem like you need to do this work unless you enter the if cases - consider moving these inside those statements.

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
...
+    private handleScroll = (): void => {
+        this.requestLayoutUpdate();
+    };
+
+    /**
+     * Request's an animation frame if there are currently no open animation frame requests
+     */
+    private requestLayoutUpdate = (): void => {
+        if (this.openRequestAnimationFrame === null) {
+            this.openRequestAnimationFrame = window.requestAnimationFrame(
```
nicholasrice, line 541: consider `DOM.queueUpdate` here

packages/fast-components/src/anchored-region/anchored-region.ts
```diff
@@ -0,0 +1,978 @@
...
+            this.positionChanged();
+        }
+    };
+
+    /**
+     *  Updates the style string applied to the region element as well as the css classes attached
+     *  to the root element
+     */
+    private updateRegionStyle = (): void => {
+        this.classList.toggle("horizontalInset", this.horizontalInset);
```
nicholasrice, line 670: cool - I didn't know `classList` had a toggle method!

scomea: Yes, it's handy.

packages/fast-components/src/anchored-region/anchored-region.template.ts
scomea: apparently I don't

packages/fast-components/src/anchored-region/anchored-region.ts
scomea: I think queueUpdate addresses a different concern.  In any case I took out the throttles for now as the functions invoked by resize merely store values and queue an update so while we could get a lot of updating of values we don't trigger excess renders.  We will need some generic throttle functionality tho.

packages/fast-components/src/anchored-region/anchored-region.ts
scomea: done

packages/fast-components/src/anchored-region/anchored-region.ts
scomea: done

packages/fast-components/src/anchored-region/anchored-region.ts
scomea: Added a reset.

packages/fast-components/src/anchored-region/anchored-region.ts
scomea: We decided to use "change" in the spec so revised code to emit that event rather than a custom one.

packages/fast-components/src/anchored-region/anchored-region.template.ts
scomea: yep

packages/fast-components/src/anchored-region/anchored-region.ts
EisenbergEffect: Basically, queueUpdate manages a single queue of tasks to run on the next animation frame. When you queue something, if the raf isn't scheduled, it will schedule one and then add your task to the queue. If it's already schedule, it just adds your task to the current queue. It's good to use our queueUpdate method because it ensures all our work is scheduled in the same way.

If you need to throttle the rate at which an individual component schedules things in tehe queue, that would need to be handled on a per-component basis. Individual bindings do this, so they don't schedule more than one update per frame.

On a related note, what 3rd party libraries are we using in our web components. This is the first that I've noticed a dependency outside of our own. Are we using lodash elsewhere or any other library? I'd like to avoid deps if we can.

PR #2910, opened 2020-04-09
fix: update select to pass onClick and onKeydown events

scomea:
# Description
User provided onClick and onKeyDown handlers were being ignored.  This change invokes them appropriately when provided.

## Motivation & context
User reported issue.

## Issue type checklist

## Process & policy checklist


packages/fast-components-react-base/src/select/select.tsx
```diff
@@ -503,7 +503,11 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
     /**
      * Handles clicks
      */
-    private handleClick = (e: React.MouseEvent): void => {
+    private handleClick = (e: React.MouseEvent<HTMLDivElement>): void => {
+        if (typeof this.props.onClick === "function") {
+            this.props.onClick(e);
```
janechu, line 508: Do you think there should be an assumption that if you're passing these props that the default behavior below should be ignored? Namely, that if they pass an onClick, they can change the `isMenuOpen` prop themselves, making this a controlled component?

Or since there is duplicate behavior for `onClick` and `onKeyDown` should there be an `onInvoke` whereby anything that has been determined to trigger the opening and closing of the dropdown should fire that instead?

scomea: I think the guidance should be that if you want to short circuit the default behavior when you grab the event you need to mark it by calling preventDefault().  We check for defaultPrevented and bail if it has been,

janechu: That makes sense but might be unintuitive, can you make sure that it's noted in the `README`?

scomea: Added a blurb in the to the docs

Issue #2849, opened 2020-03-27
[Proposal] ElementStyles Registration


EisenbergEffect:
## Description

This proposal is to add a `register` method to `ElementStyles`, which would allow the styles object to be registered into a global lookup by namespace and name. These styles could then later be retrieved by name and applied dynamically to the `adoptedStyleSheets` of a component.

The primary use case for this is in cases such as button, where we have the same component, but with several different styling options. If the customer only uses one of those options, we don't want them to have to pay for all of the options. We could handle this by creating different components, but since the only difference is styles, that would seem overly complicated. Simply being able to switch out styles seems a nicer solution. Another example of this is high-contrast only applying high-contrast styles if needed.

### Registering Styles

```TypeScript
import { css } from '@microsoft/fast-element';

export const styles = css`
  .foo { color: red; }
`.register('fast-button', 'lightweight')
```

The `css` tagged template helper returns an instance of `ElementStyles` to which we add a simple `register` method, which returns the element styles (enabling chaining). The `register` method will add the style object to a global registry, which can then be accessed elsewhere. 

### Finding Registered Styles

```TypeScript
import { StyleRegistry } from '@microsoft/fast-element';

const styles = StyleRegistry.find('fast-button', 'lightweight');
```

Any other component can lookup styles based on a known namespace and name.

### Adding/Removing Element Styles

```TypeScript
import { StyleRegistry, customElement, FastElement } from '@microsoft/fast-element';

@customElement({ ... })
export class FastButton extends FastElement {
  private appearanceStyles: ElementStyles | null = null;

  @attr appearance: string | null = null;
  appearanceChanged() {
    if (this.appearanceStyles !== null) {
      this.removeStyles(this.appearanceStyles);
    }

    this.appearanceStyles = StyleRegistry.find(this.tagName.toLowerCase(), 'lightweight');

    if (this.appearanceStyles !== null) {
      this.addStyles(this.appearanceStyles);
    }
  }
}
```

In this case, we'll add two methods to `FastElement`, implemented via `Controller` to enable adding and removing `ElementStyles` instances. As part of this, we'll also make some related changes to the `ElementStyles` interface to accommodate removal scenarios.

One final note is that with this implementation, end users would need to import any used optional styles in their JS code, to ensure that they are not tree-shaken out. There needs to be some static analyzable connection for bundlers. So, this allows us to decouple styles internally, in order to not force customers into using everything, but they do need to opt-in to the styles by importing them in their own code.

[...]

PR #2925, opened 2020-04-13
feat(fast-element): style registration and associated behaviors

EisenbergEffect:
# Description

This PR adds the ability to register a set of styles under a global key so that it can be looked up and dynamically applied by key later. Additionally, styles can now have `Behavior` instances associated with them, allowing them the ability to add arbitrary behavior to the custom element to support the added styles.

## Motivation & context

For the full context see issue #2849.

The final API is slightly different than the issue. Here's an example:

```TypeScript
const styles = css`
  ...css goes here...
`.withBehaviors(
    behaviorOne,
    behaviorTwo
  )
  .withKey('fast-button:lightweight');
```

In addition to this, the `Controller` that supports `FastElement` had APIs added for dynamically applying styles and behaviors. For example:

```TypeScript
import { ElementStyles, customElement, FastElement } from '@microsoft/fast-element';

@customElement({ ... })
export class FastButton extends FastElement {
  private appearanceStyles: ElementStyles | null = null;

  @attr appearance: string | null = null;
  appearanceChanged() {
    if (this.appearanceStyles !== null) {
      this.$fastController.removeStyles(this.appearanceStyles);
    }

    this.appearanceStyles = ElementStyles.find(`fast-button:${this.appearance}`);

    if (this.appearanceStyles !== null) {
      this.$fastController.addStyles(this.appearanceStyles);
    }
  }
}
```

This PR also fixes a latent bug in the `css` tagged template helper which would occur if composed styles were anywhere but the beginning or end.

## Issue type checklist

This is a non-breaking change.

## Process & policy checklist


EisenbergEffect:
@nicholasrice I've enabled instances of `Behavior` to be associated with styles. The behavior interfaces looks like this:

```TypeScript
interface Behavior {
  bind(source: unknown): void;
  unbind(): void;
}
```

When associating behaviors with styles, the `source` for the binding will be the custom element itself. This allows these behaviors to act upon the element, leverage the underlying controller, access the shadow dom, etc. They also participate in the lifecycle, with `bind` being called during `connectedCallback` and `unbind` being called in `disconnectedCallback`. `Behavior`s are a fundamental primitive of the templating engine, so this fits together nicely without having to add recipe-specific concepts to the core. My assumption is that we could wrap recipes up inside behavior instances automatically so that they are easy to apply this way.

Let me know if you think this will work.

nicholasrice:
Yea I think that API will work quite nicely - looking great!

chris-phelps:
@EisenbergEffect : I was trying to incorporate fast-card into a new card type and noticed that there is a very small amount of padding being added. Can this be changed as my design requirements require the image to extend all the way to the border end:

image
image

chrisdholt:
@chris-phelps can you create a new issue to track this, I don't think it's directly related to this PR :)

PR #2935, opened 2020-04-15
feat(fast-element): add functioning execution context

EisenbergEffect:
# Description

The "execution context" is contextual data that is present when binding a behavior and invoking an expression. Previously, there was only a stub implementation in place. This PR implements the feature fully.

## Motivation & context

In various templating scenarios, contextual information can be critical. For example, in an event handler, you may need access to the `event` object, or inside of a repeat block, you may need access the `parent` scope or even bind to the `index` of the item. Since these pieces of data are not directly part of the user's element or data model the runtime provides this information where contextually appropriate.

Most binding expressions look like this:

```TypeScript
x => x.someProperty
```

However, a second parameter is available, which provides the context.

```TypeScript
(x, c) => x.onClick(c.event)
```

In the example above `c` is an instance of `ExecutionContext`. There is always a context available, but its properties are only meaningful in particular scenarios. Here is a list of properties found on the context:

* `event` - The event object when inside an event handler.
* `parent` - The parent scope when inside a `repeat` block.
* `index` - The index of the current item when inside a `repeat` block (opt in).
* `length` - The length of the array when inside a `repeat` block (opt in).
* `even` - True if the index of the current item is even when inside a `repeat` block (opt in).
* `odd` - True if the index of the current item is odd when inside a `repeat` block (opt in).
* `first` - True if the current item is first in the array inside a `repeat` block (opt in).
* `middle` - True if the current item is somewhere in the middle of the the array inside a `repeat` block (opt in).
* `last` - True if the current item is last in the array inside a `repeat` block (opt in).

Some pieces of context array item info are more costly to compute. So, for performance reasons, they are not available by default. You must opt in. To opt in, pass options to the repeat directive, with the setting `positioning: true`. Here's an example:

```TypeScript
const template = html`
<ul>
  ${repeat(x => x.items, html<TodoItem>`
    <li>
      ${(x, c) => c.index + 1} <input type="checkbox" ?checked=${x => x.isDone} @change=${x => x.toggleDone()}>
      <span class=${x => x.isDone ? 'done' : ''}>${x => x.description}</span>
      <fast-neutral-button @click=${(x, c) => c.parent.removeTodo(x)}>&times;</fast-neutral-button>
    </li>
  `, { positioning: true })}
</ul>
`
```

## Issue type checklist

This is technically a breaking change, but is not likely to affect anyone.

## Process & policy checklist


janechu:
Love the opt-in for perf.

EisenbergEffect:
Added the relevant content from the PR to docs now.

EisenbergEffect:
Also need to wait to publish until the DI stuff is removed. It's a small PR coming as soon as my current one is merged.

PR #2947, opened 2020-04-16
feat: refactor recipe implementation to use behaviors

nicholasrice:
# Description
Recent work from `@microsoft/fast-element` has provided a better way to associate recipes with stylesheets. This PR refactors the initial implementation in favor of using those APIs.

Recipes will now be associate with a stylesheet and not an element, which makes both the stylesheet and the element more portable. The `@designSystemConsumer` decorator is no longer needed and has been replaced by a light weight behavior to define a provider on an element.

This change also updates the design-system properties themselves to reflect from the view, which means less DOM manipulation in startup cases where defaults are being applied.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/design-system-consumer/design-system-consumer.ts
```diff
@@ -1,91 +1,38 @@
...
+    bind(source: T) {
+        source.provider = findProvider(source);
+    }
 
-                if (!!provider) {
-                    this.provider = provider;
-                    this.provider.subscribe(this);
-                }
-            }
+    /* eslint-disable-next-line */
```
nicholasrice, line 18: @EisenbergEffect It seems like the `bind` method is required but may not always be needed - perhaps this could be optional?

packages/web-components/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -71,16 +70,69 @@ export function isDesignSystemProvider(
...
+            });
+
+            this.syncDesignSystemWithProvider();
+        }
+    }
+
+    constructor() {
+        super();
+
+        this.$fastController.addBehaviors([new DesignSystemConsumerBehavior()]);
```
nicholasrice, line 134: @EisenbergEffect I want to make sure I wasn't circumventing anything here by adding behaviors to the controller - does this look right?

packages/web-components/fast-components/src/design-system-consumer/design-system-consumer.ts
EisenbergEffect: It's tricky to make it optional because of the perf hit in checking for its existence. I'm planning some experiments around instantiate and bind related to perf. As part of that work, I could try out some ideas to remove unbind and maybe allow behaviors to add themselves to an unbind list via their bind implementation. I'm not sure. It's a big change to make. So for now, I think we need to stick with an empty unbind and then circle back around to see if we can improve that later.

packages/web-components/fast-components/src/design-system-provider/design-system-provider.ts
EisenbergEffect: Looks good. I tried to design that so that arbitrary behaviors could be added independent of the style association part.

packages/web-components/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -1,20 +1,19 @@
...
+import { FASTElement, Observable, observable } from "@microsoft/fast-element";
+import {
+    DesignSystemConsumer,
+    DesignSystemConsumerBehavior,
+} from "../design-system-consumer";
+import {
+    CSSCustomPropertyDefinition,
+    CSSCustomPropertyTarget,
+} from "../custom-properties";
+import { CSSCustomPropertyManager } from "../custom-properties";
```
janechu, line 10: nit: you could just add this import to the ones above

PR #2949, opened 2020-04-16
fix: update high contrast colors on switch component

khamudom:
# Description

Updated the background of the switch, when selected, to be Highlight color. When disabled, the labels are also set to GrayText color. There is also a double border on tab focus, in the selected state. 

before
image

after
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/switch/switch.styles.ts
+++ b/packages/web-components/fast-components/src/switch/switch.styles.ts
@@ -145,11 +145,12 @@ export const SwitchStyles = css`
         }
 
         :host(.checked) .switch {
-            background: ${SystemColors.FieldText};
+            background: ${SystemColors.Highlight};
+            border-color: ${SystemColors.Highlight};
         }
 
         :host(.checked) .checked-indicator {
-            background: ${SystemColors.Field};
+            background: ${SystemColors.HighlightText};
         }
 
         :host(.disabled) {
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-components/src/switch/switch.styles.ts
```diff
@@ -156,5 +163,12 @@ export const SwitchStyles = css`
             background: ${SystemColors.Field};
             border-color: ${SystemColors.GrayText};
         }
+
+        :host(.disabled) .label,
+        :host(.disabled) .unchecked-message,
+        :host(.disabled.checked) .checked-message {
+            forced-color-adjust: none;
```
nicholasrice, line 170: I know we've typically colored the label of a disabled control as disabled but I want to point out that is a deviation from the default behavior of form elements and labels. See https://codepen.io/rice119/pen/LYpNGVY w/ high-contrast. The text does not change color, only the checkbox. Have we gotten guidance one way or the other on this?

chrisdholt: We can certainly get feedback on this - I think it's fair. I'd suggest we either keep this behavior for now and remove later with an issue, or remove here and track whether we should add this behavior. @nicholasrice do you have a preference one way or another while we land this?

nicholasrice: My general preference is platform alignment unless we have good reasons to deviate.

khamudom: This is generally what Windows does throughout the system, setting color to both input and label.  Not saying we have to follow what Windows is doing, but if I high contrast is Windows specific, then maybe we match it.

chrisdholt: @khamudom let's follow the platform here and remove the change to label text color when disabled

khamudom: screenshot also updated

PR #2951, opened 2020-04-16
fix: addresses issue causing mismanaged disabled state

nicholasrice:
# Description
Disabled state for text-field and text-area were not being applied correctly.

## Motivation & context
TextField clobbered the disabled attr defined by FromAssociated with a different config, which clobbered the converter and resulted in un-expected disabled values.

Also updates several other components which could have similar issues due to re-defining the attr on an extended property.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/text-area/text-area.ts
```diff
@@ -56,25 +56,17 @@ export class TextArea extends FormAssociated<HTMLTextAreaElement> {
...
+    @attr({ mode: "boolean" })
     public autofocus: boolean;
     private autofocusChanged(): void {
         if (this.proxy instanceof HTMLElement) {
             this.proxy.autofocus = this.autofocus;
         }
     }
 
-    @attr
+    @attr({ converter: nullableNumberConverter, mode: "fromView" })
```
chrisdholt, line 55: are these not needed on textfield as well?

packages/web-components/fast-components/src/text-field/text-field.ts
```diff
@@ -108,8 +107,6 @@ export class TextField extends FormAssociated<HTMLInputElement> {
         }
     }
 
-    @attr
-    public value: string;
```
chrisdholt, line 112: is the duplication of these from the inherited abstract class causing the issue?

nicholasrice: it was caused by defining a different converter for the inherited property, yes - multiple `@attr` on the property is a no-no.

I've cleaned them all up except for slider which converts the `value` prop to a number. We'll want to address that but I think it is a larger change we should isloate.

packages/web-components/fast-components/src/text-area/text-area.ts
nicholasrice: fixed.

packages/web-components/fast-components/src/text-field/text-field.ts
chrisdholt: gotcha - assuming the issue would persist if the `@attr` had a different default value here which may be something that needs to be addressed as well?

nicholasrice: nm pulling it off slider seems to work for the examples so pulled it off. There might be some bugs but I think we can harden it in a second pass.

packages/web-components/fast-components/src/slider/slider.ts
```diff
@@ -50,7 +50,7 @@ export class Slider extends FormAssociated<HTMLInputElement>
     /**
      * The element's value to be included in form submission changed.
      */
-    @attr({ converter: nullableNumberConverter })
+    // @attr({ converter: nullableNumberConverter })
```
janechu, line 53: should we just remove this line?

chrisdholt: looks like an artifact from removing from slider :)

nicholasrice, line 53:
```suggestion
```

PR #2952, opened 2020-04-16
chore: add radio spec

marjonlynch:
Add radio component spec

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/radio/radio.md
+++ b/packages/web-components/fast-components/src/radio/radio.md
@@ -1,7 +1,7 @@
 # Radio
 
 ## Overview
-An implementation of a radio button as a form-connected web-component.
+An implementation of a radio button as a form-connected web-component. Facilitating single select from a group of visible choices.
 
 ### Use Cases
 Used anywhere an author might otherwise use an `input[type="radio"]`. Used to facilitate choice where only one choice is acceptable.
@@ -25,7 +25,7 @@ We want general feature-parity between this component and an `input[type="radio"
 ---
 
 ### API
-Extends [form associated custom element](./form-associated-custom-element.md).
+Extends [form associated custom element](../form-associated/form-associated-custom-element.md).
 
 *Component Name*
 `fast-radio`
@@ -39,8 +39,9 @@ Extends [form associated custom element](./form-associated-custom-element.md).
   - The radio should be submitted with the form but should not be editable.
 - `disabled`
   - The radio should be disabled from user interaction and will not be submitted with the form data.
-- `value`
-  - is not visible to the user, it's used to determine which radio is selected in a radio group
+- `value` - Not visible to the user, it's used for form data and to distinguish between other radio buttons of the same name attribute value.
+- `name` - All radio buttons with the same name will operate as a group allowing only 
+one radio within the group to be selected at a time.
 - `checked`
   - The initial checked value. 
   
@@ -53,10 +54,14 @@ Extends [form associated custom element](./form-associated-custom-element.md).
 
 ```HTML
 <!-- shadow root -->
-<label part="label"><slot></slot></label>
-<div part="radio">
-  <div part="checked-indicator"></div>
-</div>
+<template 
+  role="radio"
+>
+  <label part="label"><slot></slot></label>
+  <div part="radio">
+    <div part="checked-indicator"></div>
+  </div>
+</template>
 <!-- end shadow root -->
 ```
 
--- a/packages/web-components/fast-components/src/radio/radio.md
+++ b/packages/web-components/fast-components/src/radio/radio.md
@@ -1,0 +1,104 @@
+# Radio
+
+## Overview
+An implementation of a radio button as a form-connected web-component.
+
+### Use Cases
+Used anywhere an author might otherwise use an `input[type="radio"]`. Used to facilitate choice where only one choice is acceptable.
+  
+### Features
+- form association
+- focus delegation
+
+### Risks and Challenges
+We want general feature-parity between this component and an `input[type="radio"]`.
+
+### Prior Art/Examples
+- [FAST-DNA (React)](https://explore.fast.design/components/radio)
+- [Material UI](https://material-ui.com/components/radio-buttons/)
+- [Lightning Design](https://www.lightningdesignsystem.com/components/radio-group/)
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+- `change: CustomEvent`
+  - no custom data
+  - bubbles
+
+### Anatomy and Appearance
+
+```HTML
+<!-- shadow root -->
+<template 
+  role="radio"
```
chrisdholt, line 58: you'll likely need disabled to map to aria-disabled here.

packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+>
+  <label part="label"><slot></slot></label>
+  <div part="radio">
```
chrisdholt, line 61: I'd suggest using `control` here rather than `radio` as we've aligned a bit on that terminology for consistency.

packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+
+*Host Classes*
+- checked
+- disabled
+- required
+- readonly
+
+*Slotted Content/Slotted Classes*
+*CSS Parts*
+- radio
```
chrisdholt, line 79:
```suggestion
- control
```

packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+- label
+- checked-indicator
+
+### States
+**checked**: `true` or `false`
+The checked state can be toggled by:
+- Clicking the radio button (or any of it's labels)
+- Pressing the space-bar while focus is placed on the radio button
```
chrisdholt, line 87: If I'm in a group, do I need to press spacebar or does arrow keys move selection since this is single selection?

packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+When readonly, the value will not be changeable through user interaction. The value will still be exposed to forms on submission.
+
+### Accessibility
+The root element inside the shadow-dom of the radio will be a focusable element with the following accessibility content attributes:
+[MDN Web docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_radio_role)
+- role: radio
+- aria-checked: the checked state of the component
+- aria-required: the required state of the component
+- aria-disabled: the disabled state of the component
+- tabindex: 0
```
chrisdholt, line 104: nice

packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+
+Because we're delegating focus to this element, the relevant information will be presented to AT users when they focus this element.
```
chrisdholt, line 106: are we delegating focus? I think we're actually focusing this element rather than delegating focus to an element within the markup.

packages/web-components/fast-components/src/radio/radio.md
```diff
@@ -0,0 +1,109 @@
...
+### Anatomy and Appearance
+
+```HTML
+<!-- shadow root -->
+<template 
+  role="radio"
+>
+  <label part="label"><slot></slot></label>
+  <div part="radio">
+    <div part="checked-indicator"></div>
```
EisenbergEffect, line 62: Do we want a slot for the indicator?

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: All the aria stuff was documented in the accessibility section but I suppose if I put role here I should put the rest or just take out the template part here and leave the aria documentation in the other section?

chrisdholt: I think it can exist in both - they serve different purposes. With it left off here it seems to be missing. If I skip this and go down to accessibility I'd expect what you have outlined there. Since you're showing an example of the actual template, I'd expect it to be called out here specifically I think.

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: removed

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: removed template from here and left documentation in accessibility section. Hopefully these days it's assumed that would go in the <template/> bit but I can change if you disagree

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: done

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: Updated, pressing space bar should toggle on the radio if it wasn't already, using arrow keys also does this along with moving focus

packages/web-components/fast-components/src/radio/radio.md
chrisdholt: I'm not going to push on it - I just tend to write these as though I'm not going to be actually doing the work (which is what happened with tooltip) :).

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: hmm seems like we might, I'm my defense I did ponder this for a minute. I think we'd need this already for EZ settings. So I will add it.

packages/web-components/fast-components/src/radio/radio.md
marjonlynch: done. I just realized I could click a button and have your suggestion committed, but too late now

packages/web-components/fast-components/src/radio/radio.md
chrisdholt: I think we will, similar to checkbox I believe. The checked indicator here can likely be an SVG as well "defaulted" into the slot.

marjonlynch: done and done

PR #2957, opened 2020-04-16
feat(fast-element): safe class binding

EisenbergEffect:
# Description

Bind to an element's `class` without overwriting classes added through other mechanisms.

## Motivation & context

Particularly with host element bindings to class, there's a risk of stomping on end-user classes. The new implementation defaults to adding classes non-destructively and only removing classes that were previously added by the binding.

This update is slightly less performant than before, but a dev can always opt into more performant, destructive binding by using a property binding to `className` instead.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-element/src/directives/binding.ts
```diff
@@ -57,6 +57,45 @@ function updatePropertyTarget(this: BindingBehavior, value: unknown): void {
     this.target[this.targetName!] = value;
 }
 
+function updateClassTarget(this: BindingBehavior, value: string): void {
```
nicholasrice, line 60: I didn't know you could prescribe `this` like this - that's cool

packages/web-components/fast-element/src/directives/binding.ts
```diff
@@ -57,6 +57,45 @@ function updatePropertyTarget(this: BindingBehavior, value: unknown): void {
...
+    const classVersions = this.classVersions || Object.create(null);
+    const target = this.target;
+    let version = this.version || 0;
+
+    // Add the classes, tracking the version at which they were added.
+    if (value !== null && value !== undefined && value.length) {
```
nicholasrice, line 66: I'm curious why you're checking against null and undefined rather than if value is `typeof string`  - is this more performant?

EisenbergEffect: That's actually a good question. I'm not sure if a typeof check would be more performant in this case. I can do some experimentation with that and update this in the future.

PR #2967, opened 2020-04-20
fix: high contrast color on slider and slider label

khamudom:
# Description
Taking a high contrast color pass in slider and slider label.

## Motivation & context
The disabled color on the slider and slider label is not being set to the disabled color.
image
I was able to add a fix to the slider.
image
Instead of adding the fix for `slider-label` inside the `slider` component, I added a disabled attribute to`slider-label` and also set the disabled styles inside the label component. This will also not have the `slider-label` style dependent on `slider`.
Added disabled to the SliderConfiguration and if the `slider` is disabled, the `slider-label` style will get set.
image
Also added a disabled example for `slider-label`.
image

When hovering over the slider thumb, it is now set to Highlight.
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/slider-label/slider-label.ts
+++ b/packages/web-components/fast-components/src/slider-label/slider-label.ts
@@ -38,7 +38,6 @@ export class SliderLabel extends FASTElement {
         this.disabled
             ? this.classList.add("disabled")
             : this.classList.remove("disabled");
-        this.setAttribute("aria-disabled", this.disabled.toString());
     }
 
     public connectedCallback(): void {
--- a/packages/web-components/fast-components/src/slider-label/slider-label.ts
+++ b/packages/web-components/fast-components/src/slider-label/slider-label.ts
@@ -8,6 +8,7 @@ const defaultConfig: SliderConfiguration = {
     max: 0,
     direction: Direction.ltr,
     orientation: SliderOrientation.horizontal,
+    disabled: false,
 };
 
 export class SliderLabel extends FASTElement {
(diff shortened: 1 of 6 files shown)


packages/web-components/fast-components/src/slider-label/slider-label.ts
```diff
@@ -31,6 +32,15 @@ export class SliderLabel extends FASTElement {
         orientation: SliderOrientation.horizontal,
     };
 
+    @attr({ attribute: "disabled", mode: "boolean" })
+    public disabled: boolean; // Map to proxy element
+    private disabledChanged(): void {
+        this.disabled
+            ? this.classList.add("disabled")
+            : this.classList.remove("disabled");
+        this.setAttribute("aria-disabled", this.disabled.toString());
```
EisenbergEffect, line 41: I'm not sure that you need the setAttribute call here. The binding you have on `template` should handle it.

packages/web-components/fast-components/src/slider-label/slider-label.ts
```diff
@@ -54,13 +64,17 @@ export class SliderLabel extends FASTElement {
         if (!this.isSliderConfig(this.parentNode)) {
             this.config = defaultConfig;
         } else {
-            const { min, max, direction, orientation } = this
+            const { min, max, direction, orientation, disabled } = this
```
EisenbergEffect, line 67: @marjonlynch We still plan to follow up PR with some observation of the parent so that labels will update as orientation, direction, disabled, etc. change right?

packages/web-components/fast-components/src/slider-label/slider-label.ts
khamudom: removed setAttribute

packages/web-components/fast-components/src/slider-label/slider-label.ts
marjonlynch: @EisenbergEffect yes, I needed to get on radio component before finishing this one up. Radio is almost ready for PR and then I will get back on slider unless priorities dictate otherwise.

PR #2971, opened 2020-04-21
feat: add tab web component

eljefe223:
# Description
Adds `fast-tabs` web component.

image

## Issue type checklist

## Process & policy checklist


specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+    </div>
+    <div slot="after-content">
+        Before content
+    </div>
+</fast-tabs>
+```
+
+### States
+
+*Tabs* can either be controlled or uncontrolled, meaning if `activeId` is passed the app author is taking control of the selected tab. The `change` event fires differently depending on the `tab-activation` attribute, on tab focus if `tab-activation` is set to `"focus"` or on mouse click or "spacebar" press if `tab-activation` is set to `"key-press"`.
```
nicholasrice, line 159: I wonder if "click" would be more intuitive than "key-press" - it is shorter and is a "click" event can be fired from keystrokes like enter and spacebar (depending on the element)

specs/tabs/tabs.md
```diff
@@ -0,0 +1,173 @@
...
+### API
+
+*Component name:*
+- `fast-tabs`
+
+*Attributes:*
+- `orientation` - enum
+  - horizontal - default
+  - vertical
+- `activeId` - string
```
nicholasrice, line 42: You're mixing casing styles here (camel & kebab). I'd prefer attributes be lower case - either `active-id` or `activeid`

nicholasrice: Also - did you consider an "active" boolean attribute on either the tab or tab-panel? That may be a more intuitive API than pointing to the id of a child element.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    constructor() {
+        super();
+
+        if (this.connected) {
+            this.tabIDs = this.getTabIds();
+            this.tabPanelIDs = this.getTabPanelIds();
+            this.activeTabIndex = this.getActiveIndex();
+        }
```
nicholasrice, line 322: is this case ever true? It seems `connected` always gets asigned true in the connectedCallback which should always be called after the constructor.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+        if (this.activeTabIndex !== tabsLength - 1) {
+            this.activeTabIndex = this.activeTabIndex + 1;
+        } else {
+            this.activeTabIndex = 0;
+        }
+        this.setTabs();
+    }
+
+    private focusTab(): void {
+        const tb = this.tabs as HTMLElement[];
```
nicholasrice, line 306: Doesn't seem like you really need to store `this.tabs` as a variable if you're only referencing it once.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+                this.activeIndicatorRef.style.gridRow = `${this.activeTabIndex + 1}`;
+                this.activeIndicatorRef.style.transform = "translateX(0px)";
+                this.activeIndicatorRef.classList.remove("activeIndicatorTransition");
+            });
+        }
+    }
+
+    private decrement(tabsLength: number): void {
+        this.prevActiveTabIndex = this.activeTabIndex;
+        if (this.activeTabIndex !== 0) {
```
nicholasrice, line 287: there is a function in `fast-web-utilities` that will do this - `wrapInBounds`:

```
this.activeTabIndex = wrapInBounds(0, tabsLength, this.activeTabIndex - 1)

// and in increment()
this.activeTabIndex = wrapInBounds(0, tabsLength, this.activeTabIndex + 1)
```

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+            this.activeIndicatorRef.style.gridRow = `${this.activeTabIndex + 1}`;
+            const next: number = this.activeIndicatorRef.offsetTop;
+            this.activeIndicatorRef.style.gridRow = `${this.prevActiveTabIndex + 1}`;
+            const dif: number = next - prev;
+            this.activeIndicatorRef.style.transform = `translateY(${dif}px)`;
+            this.activeIndicatorRef.classList.add("activeIndicatorTransition");
+            this.activeIndicatorRef.addEventListener("transitionend", () => {
+                this.ticking = false;
+                this.activeIndicatorRef.style.gridRow = `${this.activeTabIndex + 1}`;
+                this.activeIndicatorRef.style.transform = "translateX(0px)";
```
nicholasrice, line 279: should this be `translateY`?

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+                this.ticking = false;
+                this.activeIndicatorRef.style.gridColumn = `${this.activeTabIndex + 1}`;
+                this.activeIndicatorRef.style.transform = "translateX(0px)";
+                this.activeIndicatorRef.classList.remove("activeIndicatorTransition");
+            });
+        } else {
+            const prev: number = this.activeIndicatorRef.offsetTop;
+            this.activeIndicatorRef.style.gridRow = `${this.activeTabIndex + 1}`;
+            const next: number = this.activeIndicatorRef.offsetTop;
+            this.activeIndicatorRef.style.gridRow = `${this.prevActiveTabIndex + 1}`;
```
nicholasrice, line 272: It seems like this logic is the same, it's just the properties that you're using that are different. Maybe instead of duplicating the logic you could conditionalize the properties into variables, and then use those variables in property accessors

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    private handleTabKeyDown = (event: KeyboardEvent): void => {
+        const keyCode: number = event.keyCode;
+        const tabsLength: number = this.tabs.length;
+        if (this.isHorizontal()) {
+            switch (keyCode) {
+                case keyCodeArrowLeft:
+                    event.preventDefault();
+                    this.decrement(tabsLength);
```
nicholasrice, line 199: I'm not sure it's helpful to pass the tabLength to these functions - it's always the same and always retrieved from `this.tabs.length` - perhaps these functions could just access the property internally?

I could also see this being simplified by replacing `increment` and `decrement` with a single function that accepts a delta to adjust by:

```
this.adjust(-1)

//... 

this.adjust(1)
```

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+        this.handleActiveIndicatorPosition();
+        this.getTabPanels();
+        this.focusTab();
+        this.change();
+    }
+
+    private handleTabClick = (event): void => {
+        const selectedTab = event.srcElement as Element;
+        this.prevActiveTabIndex = this.activeTabIndex;
+        this.activeTabIndex = Array.from(this.tabs).indexOf(event.target);
```
nicholasrice, line 182: I think you want to use `event.currentTarget` here - it is possible that an element *inside* the tab will trigger the event, in which case this will evaluate to -1

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+        this.change();
+        this.getTabs();
+        this.handleActiveIndicatorPosition();
+        this.getTabPanels();
+        this.focusTab();
+        this.change();
+    }
+
+    private handleTabClick = (event): void => {
+        const selectedTab = event.srcElement as Element;
```
nicholasrice, line 180: you shouldn't use `srcElement` - it is deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+                    ? `tab-${index + 1}`
+                    : this.tabIDs[index]
+            );
+            this.activeTabIndex !== index
+                ? tabpanel.setAttribute("hidden", "")
+                : tabpanel.removeAttribute("hidden");
+        });
+    };
+
+    private getTabIds(): string[] {
```
nicholasrice, line 157: Is there a case where these ID's wont exist? If so, I'd prefer to see an accurate type signature for these functions and not coercing with `as string`.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+
+    public end: HTMLSlotElement;
+    public endContainer: HTMLDivElement;
+    public handleEndContentChange(): void {
+        this.end.assignedNodes().length > 0
+            ? this.endContainer.classList.add("end")
+            : this.endContainer.classList.remove("end");
+    }
+
+    public change = (): void => {
```
nicholasrice, line 64: should this really be public?

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,324 @@
...
+                !this.isHorizontal()
+                    ? tab.classList.add("vertical")
+                    : tab.classList.remove("vertical");
+            }
+        });
+    };
+
+    private getTabPanels = (): void => {
+        this.tabIDs = this.getTabIds();
+        this.tabPanelIDs = this.getTabPanelIds();
```
nicholasrice, line 137: Can you rename `getTabPanels` and `getTabs` to something more descriptive of what they're doing? They don't actually return anything and seem to have numerous side-effects

specs/tabs/tabs.md
eljefe223: Something weird happened here, this spec is in a separate review. I'll remove this from this PR. If you could review that one here: https://github.com/microsoft/fast-dna/pull/2857

specs/tabs/tabs.md
eljefe223: i'll fix this in other review.

packages/web-components/fast-components/src/tabs/tabs.ts
eljefe223: Its `translateX` if `orientation` is set to `horizontal`, `translateY` is orientation is set to `vertical`

packages/web-components/fast-components/src/tabs/tabs.ts
eljefe223: Don't callbacks need to be public?

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+    private tabpanelIds: Array<string | null>;
+    private connected: boolean = false;
+
+    private change = (): void => {
+        this.$emit("change", this.activetab);
+    };
+
+    private getActiveIndex(): number {
+        if (this.activeid !== undefined) {
+            return this.tabIds.indexOf(this.activeid) === -1
```
EisenbergEffect, line 85: I'd recommend a minor optimization here to get the index just once.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+
+    private setTabs = (): void => {
+        this.tabIds = this.getTabIds();
+        this.tabpanelIds = this.getTabPanelIds();
+        this.activeTabIndex = this.getActiveIndex();
+        this.tabs.forEach((tab: HTMLElement, index: number) => {
+            if (tab.slot === "tab") {
+                tab.setAttribute(
+                    "id",
+                    this.tabIds[index] === undefined || this.tabIds[index] === null
```
EisenbergEffect, line 101: Optimize to get index only once.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+                        ? `tab-${index + 1}`
+                        : (this.tabIds[index] as string)
+                );
+                tab.setAttribute(
+                    "aria-selected",
+                    this.activeTabIndex === index ? "true" : "false"
+                );
+                tab.setAttribute(
+                    "aria-controls",
+                    this.tabpanelIds[index] === undefined ||
```
EisenbergEffect, line 111: Same as above. We can do the index just once.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+        if (this.connected) {
+            this.tabIds = this.getTabIds();
+            this.tabpanelIds = this.getTabPanelIds();
+            this.activeTabIndex = this.getActiveIndex();
+        }
+    }
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.connected = true;
```
EisenbergEffect, line 299: You can remove the custom property and do `this.$fastController.isConnected`.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+                        : (this.tabpanelIds[index] as string)
+                );
+                tab.setAttribute(
+                    "style",
+                    this.isHorizontal()
+                        ? `grid-column: ${index + 1};`
+                        : `grid-row: ${index + 1};`
+                );
+                tab.addEventListener("click", this.handleTabClick);
+                tab.addEventListener("keydown", this.handleTabKeyDown);
```
EisenbergEffect, line 123: Should we do something to prevent multiple even listener attach if this method is called more than once?

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+                "aria-labeledby",
+                this.tabIds[index] === undefined || this.tabIds[index] === null
+                    ? `tab-${index + 1}`
+                    : (this.tabIds[index] as string)
+            );
+            this.activeTabIndex !== index
+                ? tabpanel.setAttribute("hidden", "")
+                : tabpanel.removeAttribute("hidden");
+        });
+    };
```
EisenbergEffect, line 155: Like above, let's do the index lookups only once.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    private focusTab(): void {
+        this.tabs[this.activeTabIndex].focus();
+    }
+
+    constructor() {
+        super();
+
+        if (this.connected) {
```
EisenbergEffect, line 290: Shouldn't have connected called until after constructor.

packages/web-components/fast-components/src/tabs/tabs.ts
EisenbergEffect: Another way to handle this might be to have each tab handle its own click and keydown event, and then simply emit a custom event that the tab component can handle to control what happens. If we did that, you can add all event handlers in the constructor. This won't work if you want to handle any type of element being a tab though. We'd have to keep it as is for that scenario. In which case, you'll want to tag the elements to make sure you don't add duplicate handlers as mentioned above.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+        this.activeIndicatorRef.classList.add("activeIndicatorTransition");
+        this.activeIndicatorRef.addEventListener("transitionend", () => {
+            this.ticking = false;
+            this.activeIndicatorRef.style[gridProperty] = `${this.activeTabIndex + 1}`;
+            this.activeIndicatorRef.style.transform = `${translateProperty}(0px)`;
+            this.activeIndicatorRef.classList.remove("activeIndicatorTransition");
+        });
+    }
+
+    private adjust(adjustment: number): void {
```
EisenbergEffect, line 273: I wonder if we could come up with a more descriptive name for this method?

packages/web-components/fast-components/src/tabs/fixtures/base.html
```diff
@@ -0,0 +1,115 @@
...
+        </fast-tab-panel>
+        <fast-tab-panel id="TabPanelTwo">
+            Tab two content. This is for testing.
+        </fast-tab-panel>
+        <fast-tab-panel id="TabPanelThree">
+            Tab three content. This is for testing.
+        </fast-tab-panel>
+    </fast-tabs>
+    <h4>No active indicator - Vertical</h4>
+    <fast-tabs activeIndicator="false" orientation="vertical" activeIndicator>
```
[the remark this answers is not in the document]
eljefe223: Ahh, good catch there was a need for 2 at one time but no longer!

packages/web-components/fast-components/src/tabs/tabs.ts
eljefe223: Rob and I chatted about this offline. If you give it the same exact instance for the same event, it won't get registered twice. So we are good.

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+    vertical = "vertical",
+    horizontal = "horizontal",
+}
+
+export class Tabs extends FASTElement {
+    @attr
+    public orientation: TabsOrientation = TabsOrientation.horizontal;
+    private orientationChanged(): void {
+        this.orientation === TabsOrientation.vertical
+            ? this.classList.add("vertical")
```
nicholasrice, line 22: You could be able to bind this class in the template

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+            this.setTabs();
+            this.setTabPanels();
+            this.handleActiveIndicatorPosition();
+        }
+    }
+
+    public start: HTMLSlotElement;
+    public startContainer: HTMLDivElement;
+    public handleStartContentChange(): void {
+        this.start.assignedNodes().length > 0
```
nicholasrice, line 58: These classes also look like they could be delegated to the template

packages/web-components/fast-components/src/tabs/tabs.ts
```diff
@@ -0,0 +1,301 @@
...
+            const tabpanelId: string | null = this.tabpanelIds[index];
+            tabpanel.setAttribute(
+                "id",
+                tabpanelId === undefined || tabpanelId === null
+                    ? `panel-${index + 1}`
+                    : tabpanelId
+            );
+            tabpanel.setAttribute(
+                "aria-labeledby",
+                tabID === undefined || tabID === null ? `tab-${index + 1}` : tabID
```
nicholasrice, line 154: minor nit here, but you could do this with only one comparison:

`typeof tabID === "string"`

packages/web-components/fast-components/src/tabs/tabs.ts
eljefe223: This would require the tab to have an `orientation` attr. I'd like to avoid that

PR #2973, opened 2020-04-21
fix: improve performance of finding contextual design-system-providers

nicholasrice:
# Description
Improves the performance of the algorithm for finding an element's contextual design-system-provider element.

## Motivation & context
A previous change adjusted the recipe system to associate with stylesheets instead of components. In introducing that change, a side effect was that each recipe was resolving the design-system-provider independently, leading to a significant amount of duplicated work.

This change adjust the `findProvider` algorithm used by recipes to first check to see if a provider has been stored on the source object. If it has, a reference to that provider will be returned. Additionally, when a provider is found, a reference top the provider is stored on the source object for future invocations of the function. This means each element is only finding the provider once.

This change also introduces a check to determine if the parent element *has* a design system provider context. If it does, a reference to that provider will be returned. This will improved resolution speed of nested scenarios.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/custom-properties/behavior.ts
+++ b/packages/web-components/fast-components/src/custom-properties/behavior.ts
@@ -37,14 +37,14 @@ export function cssCustomPropertyBehaviorFactory(
         bind(source: typeof FASTElement): void {
             const target = this.host(source);
 
-            if (target) {
+            if (target !== null) {
                 target.registerCSSCustomProperty(this);
             }
         },
         unbind(source: typeof FASTElement): void {
             const target = this.host(source);
 
-            if (target) {
+            if (target !== null) {
                 target.unregisterCSSCustomProperty(this);
             }
         },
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-components/src/custom-properties/behavior.ts
```diff
@@ -34,14 +34,14 @@ export function cssCustomPropertyBehaviorFactory(
         name,
         value,
         host,
-        bind(source: typeof FASTElement | HTMLElement): void {
+        bind(source: typeof FASTElement): void {
             const target = this.host(source);
 
             if (target) {
```
EisenbergEffect, line 40: Nit: Since this is perf-sensitive code, we should change `if(target)` to `if (target !== void 0`) or `null` or whatever the "empty" value is, as that explicit, non-coerced check will be slightly faster.

PR #2980, opened 2020-04-21
fix: select root should not be focusable

scomea:
# Description
Select root had a tab index of -1 which meant it could be focused on as well as the trigger.  This caused the select to be focused twice and narrated twice in narrator scan mode.

## Motivation & context
Fixes narrator issues encountered by partners.

## Issue type checklist

## Process & policy checklist


packages/react/fast-components-react-base/src/select/select.tsx
```diff
@@ -158,7 +158,7 @@ class Select extends Foundation<SelectHandledProps, SelectUnhandledProps, Select
                 className={this.generateClassNames()}
                 onKeyDown={this.handleKeydown}
                 onClick={this.handleClick}
-                tabIndex={-1}
```
nicholasrice, line 161: won't this hide the element entirely from screen readers?

scomea: No, the actual focusable element is the trigger button contained within the component.

nicholasrice: I don't think this is the correct usage of aria-hidden, MDN explicitly recommends against doing exactly this (focusable element as child of `aria-hidden="true"`): https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

It's also not clear to me why this helps address focusability  of the element - what problem is adding this attribute solving?

scomea: The point of the fix was to stop double narration but just removing the tab index seems to do it, that aria-hidden was redundant so removed it.

PR #2985, opened 2020-04-21
feat: add radio component

marjonlynch:
Add radio web component

## Process & policy checklist


packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,212 @@
...
+    private checkedChanged(): void {
+        if (!this.dirtyChecked) {
+            this.dirtyChecked = true;
+        }
+
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.checked = this.checked;
+        }
+
+        this.dispatchEvent(new CustomEvent("change", { bubbles: true, composed: true }));
```
EisenbergEffect, line 86: You can use `this.$emit('change');`

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (this.name !== undefined) {
+            const radioGroup = radioGroups.get(this.name);
+            if (radioGroup) {
+                radioGroup.push(this);
+            } else {
+                radioGroups.set(this.name, [this]);
+            }
+        }
+
+        this.addEventListener("keydown", this.keydownHandler);
```
EisenbergEffect, line 134: I've been thinking about event bindings that are internal to elements lately...and I think you can move this line to the constructor and then completely remove the removeEventListener call in disconnect. I don't think there's a need to tear it down since the handler and the element are part of the single unit that will be GC'd.

marjonlynch: ok will do

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: dangit, will do

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -0,0 +1,138 @@
...
+import {
+    neutralFillInputHoverBehavior,
+    neutralFillInputRestBehavior,
+    neutralForegroundRestBehavior,
+    neutralOutlineHoverBehavior,
+    neutralOutlineRestBehavior,
+} from "../styles/recipes";
+
+export const RadioStyles = css`
+    --var
```
nicholasrice, line 15: probably don't need this `--var`

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,212 @@
...
+        this.checked ? this.classList.add("checked") : this.classList.remove("checked");
+        this.checkedAttribute = this.checked;
+
+        if (this.checked) {
+            this.updateOtherGroupRadios();
+        }
+    }
+
+    private updateOtherGroupRadios(): void {
+        if (this.name !== undefined && this.parentNode) {
```
chrisdholt, line 96: Is the expectation here that these are direct descendants of a radiogroup?

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,212 @@
...
+        this.dispatchEvent(new CustomEvent("change", { bubbles: true, composed: true }));
+        this.checked ? this.classList.add("checked") : this.classList.remove("checked");
+        this.checkedAttribute = this.checked;
+
+        if (this.checked) {
+            this.updateOtherGroupRadios();
+        }
+    }
+
+    private updateOtherGroupRadios(): void {
```
chrisdholt, line 95: I'm a bit torn here - should a radio be responsible for changing the attributes of other radios or should a radiogroup handle this? I think it's clever, but would a tab manage other tabs, or would the tabs component do that? Is this a principle we'd align to elsewhere?

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,212 @@
...
+                                break;
+                            } else {
+                                index -= 1;
+                            }
+                        }
+                    }
+                }
+                break;
+        }
+    };
```
chrisdholt, line 195: Should this belong in a radiogroup component? Would we expect other components to map to this API?

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,212 @@
+import { attr, observable } from "@microsoft/fast-element";
+import {
+    keyCodeArrowDown,
+    keyCodeArrowLeft,
+    keyCodeArrowRight,
+    keyCodeArrowUp,
+    keyCodeSpace,
+} from "@microsoft/fast-web-utilities";
+import { FormAssociated } from "../form-associated";
+import { FASTRadio } from ".";
```
chrisdholt, line 10: I think this will need to path to `./`

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: dangit, that was left in before the radioGroups map, not needed, removing.

packages/web-components/fast-components/src/radio/radio.styles.ts
marjonlynch: that's a global var, surprised you didn't know that Nick. Removing.

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: using the name attribute to group radio button is the standard AFAIK. I started to create a radiogroup and got push back on whether it was necessary at all. I couldn't come up with a good reason to have the radiogroup component so this is a solution. If there are other components that have this standard of grouping by name attribute then maybe they should use this method as well? Hopefully I'm not misunderstanding your question here.

chrisdholt: Where did the pushback happen? I think I missed a conversation...

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: tabs don't have this standard that radio has without our components, people can use the name attribute to group radio buttons and get this behavior by default. Even though I have a name attribute that is something that is normally built in.

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: done

packages/web-components/fast-components/src/radio/radio.ts
chrisdholt: I think the primary concern here is ensuring that we have a principle that we're applying. Menu items can support several internal roles such as "menuitem", "menuitemradio", "menuitemcheckbox". I think it falls down in that example as the menu items need to be a child of a menu. 

As I said, I'm a bit torn but I understand the intent. All that said, it would've been nice to have this detailed out in the spec (or added after the fact once it was decided). We should document this somewhere - for now, I'd likely recommend adding the details to the spec as we don't have a full story for the Readme files yet.

marjonlynch: It was stated in the spec that the name attribute was used to group controls. I was under the impression we also wanted to create components that would adhere to html standards as they were, in this case if you required the use of a radiogroup component to get that behavior it wouldn't be the standard way radios are grouped in html. I understand wanting to have consistency, I guess there's a trade off decision to be made there.

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,211 @@
...
+                            if (!group[index].disabled) {
+                                group[index].checked = true;
+                                group[index].focus();
+                                break;
+                            } else {
+                                index += 1;
+                            }
+                        }
+                    }
+                }
```
chrisdholt, line 174: It seems like this could be extracted out to a separate function to set focus in a specific direction as everything but movement of the index is the same. Essentially having a separate function where you pass a value (+1 or -1) with the direction you want to focus.

packages/web-components/fast-components/src/radio/radio.ts
chrisdholt: @marjonlynch - one thing that we'll need to consider here is how these will work when they are within a toolbar: https://w3c.github.io/aria-practices/#for-radio-group-contained-in-a-toolbar

Does the current implementation support that behavior as outlined in the W3C specification or should we / will we need to add that?

packages/web-components/fast-components/src/radio/radio.ts
chrisdholt: I think the issue is more ensuring that we cover all bases when we add something like this which seems to be a valuable convenience. The radio behavior is slightly different when inside a toolbar: https://github.com/microsoft/fast-dna/pull/2985/files#r413181301. As the W3C is a standard, we need to consider if we're going to bake the convenience in for that as well or if we expect authors to do something different.

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: I had separate conversations with @nicholasrice and @EisenbergEffect around this topic and those conversations led me to this solution (the map was Rob's idea). Again, I started with the same idea you have, as you know I asked about doing the radio group. However given the way the <input type="radio"/> works out of the box this seemed like the better solution.  Actually I see this code could be made a few lines smaller as well now

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: yep, refactoring now, will do

packages/web-components/fast-components/src/radio/radio.template.ts
```diff
@@ -0,0 +1,25 @@
...
+    >
+        <div part="control" class="control">
+            <slot name="checked-indicator">
+                <div part="checked-indicator" class="checked-indicator"></div>
+            </slot>
+        </div>
+        ${when(
+            x => x.childNodes.length,
+            html` <label part="label" class="label"><slot></slot></label> `
+        )}
```
chrisdholt, line 23: Have we tested and validated and tested that this behaves as expected? Just curious based off my exchange here in the W3C repo: https://github.com/w3c/aria-practices/issues/1286#issuecomment-566372426

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: I'm going to resolve this conversation as it's similar to the other discussion which I'll leave open until properly addressed.

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: group stuff removed, just a humble radio left

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+        outline: none;
+        margin: calc(var(--design-unit) * 1px) 0;
+        ${
+            /*
+             * Chromium likes to select label text or the default slot when
+             * the radio button is clicked. Maybe there is a better solution here?
+             */ ""
+        } user-select: none;
+        position: "relative";
+        "flex-direction": "row";
```
eljefe223, line 27: This shouldn't be wrapped in quotes ;)

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+        align-items: center;
+        outline: none;
+        margin: calc(var(--design-unit) * 1px) 0;
+        ${
+            /*
+             * Chromium likes to select label text or the default slot when
+             * the radio button is clicked. Maybe there is a better solution here?
+             */ ""
+        } user-select: none;
+        position: "relative";
```
eljefe223, line 26: same here

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+        "flex-direction": "row";
+        transition: "all 0.2s ease-in-out";
```
eljefe223, line 28: and here

packages/web-components/fast-components/src/radio/radio.template.ts
marjonlynch: Chris looking at that bug it doesn't seem like the same scenario as the template code used here and in checkbox, that refers to a label outside the radio tree/branch and using the for attribute. in this case it's just content of the radio. At least the voice over on Mac is reading it fine. I don't have JAWS on my machine though.

packages/web-components/fast-components/src/radio/radio.template.ts
```diff
@@ -0,0 +1,29 @@
+import { html, when } from "@microsoft/fast-element";
+import { Radio } from "./radio";
+
+export const RadioTemplate = html<Radio>`
+    <template
+        role="radio"
+        aria-checked="${x => x.checked}"
```
eljefe223, line 7: These should all be prefixed, except `aria-checked` with `?` so they don't show in the DOM  when they are false.
image

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -0,0 +1,137 @@
...
+            /* Need to discuss with Brian how HorizontalSpacingNumber can work. https://github.com/microsoft/fast-dna/issues/2766 */ ""
+        } padding-inline-start: calc(var(--design-unit) * 2px + 2px);
+        margin-inline-end: calc(var(--design-unit) * 2px + 2px);
+        cursor: pointer;
+        ${
+            /* Font size is temporary - replace when adaptive typography is figured out */ ""
+        } font-size: calc(1rem + (var(--density) * 2px));
+    }
+
+    .checked-indicator {
```
eljefe223, line 56: It seems that this `checked-indicator` is not centered.
image

packages/web-components/fast-components/src/radio/radio.template.ts
chrisdholt: I'd recommend testing and ensuring that the following works as expected:

- I can associate a label and have it work as the template suggests
- I can use aria-labelledby with a visible label which is not the internal label
- I can successfully set aria-label for a non-visible label

Pulling from: https://w3c.github.io/aria-practices/#radiobutton
> Each radio element is labelled by its content, has a visible label referenced by aria-labelledby, or has a label specified with aria-label.

packages/web-components/fast-components/src/radio/radio.template.ts
marjonlynch: Oh boy that's a new one for me, hadn't seen that anywhere, but found it in button, will do

packages/web-components/fast-components/src/radio/radio.styles.ts
marjonlynch: prolly the errant " in the styles, looks centered now

marjonlynch: Screen Shot 2020-04-23 at 10 31 07 AM

packages/web-components/fast-components/src/radio/radio.template.ts
marjonlynch: I believe with the latest fixture html all 3 are present and I went through them with VoiceOver

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export class Radio extends FormAssociated<HTMLInputElement> {
+    @attr({ attribute: "readonly", mode: "boolean" })
+    public readOnly: boolean; // Map to proxy element
+    private readOnlyChanged(): void {
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.readOnly = this.readOnly;
+        }
+
+        this.readOnly
```
nicholasrice, line 13: You should bind this in the template

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -0,0 +1,124 @@
...
+        if (!this.dirtyChecked) {
+            this.dirtyChecked = true;
+        }
+
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.checked = this.checked;
+        }
+
+        this.$emit("change");
+        this.checked ? this.classList.add("checked") : this.classList.remove("checked");
```
nicholasrice, line 78: You should bind this class in the template.

marjonlynch: ok, will do

packages/web-components/fast-components/src/radio/fixtures/base.html
```diff
@@ -0,0 +1,47 @@
...
+    <fast-radio disabled>label</fast-radio>
+    <fast-radio disabled checked>checked</fast-radio>
+
+    <h4>Inline</h4>
+    <fast-radio name="fruit-inline" value="apples" checked>Apples</fast-radio>
+    <fast-radio name="fruit-inline" disabled value="bananas">Bananas</fast-radio>
+    <fast-radio name="fruit-inline" value="honeydew">Honeydew</fast-radio>
+
+    <h4>Vertical</h4>
+    <fieldset style="display: flex; flex-direction: column; align-items: start;">
```
chrisdholt, line 29: I wonder if now that we're adding a `radiogroup` component these examples may be more beneficial there? We can likely limit this file to single instances of radio and the variations.

marjonlynch: I'll trim this one up before I push the next commit

PR #2991, opened 2020-04-23
fix: add mechanism to register custom properties prior to construction

nicholasrice:
# Description
Fixes an issue where, depending on project configuration, `CSSCustomPropertyDefinitions` were not getting attached by the `DesignSystemProvider`.

## Motivation & context
The previous implementation of the `DesignSystemProvider` made an assumption that `connectedCallback` lifecycle callbacks were always executed in document order. This is an unsafe assumption and order can differ dramatically depending on:
1. Whether the DOM w/ custom elements is parsed before / after the script defining those custom elements
2. The order in which the custom elements are *defined*

This changes adds a mechanism to attach `CSSCustomPropertyDefinitions` to a `DesignSystemProvider` *before* the component is actually constructed. Because the usual public API to register these definitions aren't defined until construction, logic is being introduced to treat an array property as a primitive registration vessel if the APIs don't exist at the time when we're registering the definitions. During the `connectedCallback`, any `CSSCustomPropertyDefinitions` in that collection will be registered.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -182,8 +184,28 @@ export class DesignSystemProvider extends FASTElement
                 });
             }
         });
+
+        // Register all properties that may have been attached before construction
+        if (Array.isArray(this.disconnectedCSSCustomPropertyRegistry)) {
+            for (let i = 0; i < this.disconnectedCSSCustomPropertyRegistry.length; i++) {
```
janechu, line 190: nit: you could store the length of this

nicholasrice: it's actually not faster! https://jsperf.com/loops/33

EisenbergEffect: Length cache with pre-increment was actually fastest for me. It may depend on the browser.

janechu: strange, runs slightly faster for me, but it's really negligible.

PR #2995, opened 2020-04-23
chore: update prettier to latest version

awentzel:
# Description
Upgrades Prettier and Husky.

## Motivation & context
Upgrades and improved maintainability and developer performance. Upgrade to newer versions while bringing packages into consistency and removing code duplication/bloat.  

1.  Upgrade Prettier
2. Upgrade Husky, significantly older, and remove from packages to eliminate unnecessary packages and usage.

## Issue type checklist

## Process & policy checklist


awentzel:
@chrisdholt link to failing GH Action for Windows https://github.com/microsoft/fast-dna/runs/607662272?check_suite_focus=true

awentzel:
Prettier is giving more issues and so I created a new bug #3022

awentzel:
> Holding on this until we know it solves the problem. I don't see anything related to this as a version bug in the changelog: https://github.com/prettier/prettier/blob/master/CHANGELOG.md#205
> 
> Is there a link to the failing build you can share?

After tons of testing, I did not find prettier or husky to be the culprit of broken local CI.  However, there are inconsistencies and old versions that I think it would be great to upgrade to.  I have done significant testing through this process and found no issues with these upgrades on CI, on local CI, or when running locally.

packages/react/fast-components-react-msft/package.json
```diff
@@ -106,7 +106,7 @@
     "focus-visible": "^4.1.5",
     "jest": "^25.4.0",
     "lodash-es": "^4.0.0",
-    "prettier": "2.0.2",
+    "prettier": "2.0.5",
```
nicholasrice, line 109: Is there a reason we're targeting a specific version and not latest?

chrisdholt: @nicholasrice prettier should always be fixed or we'll hit build breaks as they change/update/fix formatting issues as they see fit.

chrisdholt: It'll require a code change just for formatting which seems unnecessary. I'm not sure updating to latest here is necessary even as there are no changes from the version we are currently locked to.

awentzel: I think it's prudent to maintain up-to-date packages. I also disagree that we should fix any package.  It's much easier to troubleshoot issues when they happen as a result of a release, otherwise, it's like finding a needle in a haystack.

chrisdholt: To be super clear - we will NOT have issues if we fix this package. We 100% have issues with we Semver this package.

chrisdholt: Prettier calls this practice out specifically in their docs: https://prettier.io/docs/en/install.html

awentzel: By design.  We have always operated on the premise that keeping up-to-date is preferred and build breaks as a result of package dependencies is an acceptable risk.

chrisdholt: In that case, I'm going to hold on the PR. I fundamentally believe that the current build break is unrelated to this. This PR will introduce build breaks that are 100% avoidable by following previous precedence with prettier prior to the 2.0 update as well as prettier's own guidance.

awentzel: I am not suggesting this fixes the build break. I'm am suggesting that "100% avoidable" is not accurate, given the other bug filed on build break related to prettier.

awentzel:
> 2. How this PR solves an existing problem with prettier - there are no security risks associated, there are no changelog items for JS or TS files between the current pinned version and this pinned version being proposed.

This PR does NOT solve any existing problem.  Please see the description for details on this PR.

chrisdholt:
> > Holding on this until we know it solves the problem. I don't see anything related to this as a version bug in the changelog: https://github.com/prettier/prettier/blob/master/CHANGELOG.md#205
> > Is there a link to the failing build you can share?
> 
> After tons of testing, I did not find prettier or husky to be the culprit of broken local CI. However, there are inconsistencies and old versions that I think it would be great to upgrade to. I have done significant testing through this process and found no issues with these upgrades on CI, on local CI, or when running locally.

@awentzel, to my point - this PR does not change anything between the previous prettier version and the current version. The changelog for prettier details no pertinent changes for prettier that impact our repo. The version change doesn't have any effect. I'm fine investigating a change to Husky, my hold is not in reference to that.

I want to make sure that we agree that this dependency should be fixed in the package.json and not arbitrarily updated when new versions are available. That is the primary reason I'm holding here. The intent with prettier is formatting consistency that does not change over time and that we do not need to care about. Per prettier:

> We recommend pinning an exact version of prettier in your package.json as we introduce stylistic changes in patch releases.

The above quote details exactly why we should avoid updating this regularly. We should not be bothered by how the formatter decides to change spacing in a patch version and we should fix the versioning and "forget it" barring a API or significant issue unrelated to styling/formatting.

awentzel:
Closing issue and moving to the Steering committee to reach consensus on policy regarding package versions.

PR #2999, opened 2020-04-23
fix: CSS custom properties not defined when a component first renders

nicholasrice:
# Description
Fixes a rendering bug related to the design-system-provider where CSS custom properties have not been written by the time a component first paints.

I also organized the methods / property order and added more code comments, so apologies if this looks more complicated than it should be.

## Motivation & context
This change changes the behavior of the design-system-provider to synchronously create
CSS custom properties instead of batching them asynchronously. This is being done to prevent flashes of un-styled content and un-desired transitions that result from the CSS custom property not being defined when the component using the CSS custom property first paints.

This change also introduces an optimization where it will use a `CSSStyleSheet` attached via `adoptedStyleSheets` to create the custom properties when that API is available. There are significant performance benefits to this approach over using the `HTMLElement.style.setProperty` API: https://jsbench.me/hck9d4noez/1. Inline styles will be the fallback behavior if adopted stylesheets are not supported.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/design-system-provider/design-system-provider.ts
```diff
@@ -206,6 +141,20 @@ export class DesignSystemProvider extends FASTElement
...
 
+    /**
+     * The target of CSSCustomPropertyDefinitions registered
+     * with the provider. This will be #1 when adoptedStyleSheets are supported
+     * and #2 when they are not.
+     *
+     * 1. The `style` property of a CSSStyleRule on an adoptedStyleSheet
+     * 2. The `style` property of the element, resulting in inline styles
+     */
+    public customPropertyTarget: CSSStyleDeclaration;
```
nicholasrice, line 152:
```suggestion
    private customPropertyTarget: CSSStyleDeclaration;
```

PR #3006, opened 2020-04-24
fix: standardization and refactor

nicholasrice:
# Description
This change standardizes implementations of classes, event dispatches, and several other clean up tasks.

Note - the following are **technically breaking**:
1. `readonly` -> `readOnly` on `Textarea` component
2. `before-content` & `after-content` -> `start` & `end` in `Textfield`

## Motivation & context
This change combs through the components to leverage some of the newer features out of `@microsoft/fast-element`. It also implements a mixin pattern for any component leveraging `start`/`end` patterns.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/text-field/text-field.template.ts
```diff
@@ -9,19 +10,16 @@ export const TextFieldTemplate = html<TextField>`
         aria-disabled="${x => x.disabled}"
         aria-readonly="${x => x.readOnly}"
         tabindex="${x => (x.disabled ? null : 0)}"
+        class="
+            ${x => x.appearance}
+            ${x => (x.readOnly ? "readonly" : "")}
```
janechu, line 15: seems like you're missing an "

packages/web-components/fast-components/src/utilities/apply-mixins.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Apply mixins to a constructor.
+ *
+ * Shamelessly stolen from https://www.typescriptlang.org/docs/handbook/mixins.html
```
EisenbergEffect, line 4: We should add an acknowledgements.md file to fast-components and call this out. I've got one in fast-element that calls out influences, etc.

packages/web-components/fast-components/src/text-field/text-field.template.ts
nicholasrice: good eye - fixed!

PR #3008, opened 2020-04-24
chore: add toolbar spec

chrisdholt:
# Description
This PR adds a new spec for a toolbar component.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/toolbar.md
+++ b/specs/toolbar.md
@@ -1,0 +1,127 @@
+# Toolbar
+
+## Overview
+As defined by the W3C:
+
+> A toolbar is a container for grouping a set of controls, such as buttons, menubuttons, or checkboxes.
+> 
+> When a set of controls is visually presented as a group, the toolbar role can be used to communicate the presence and purpose of the grouping to screen reader users. Grouping controls into toolbars can also be an effective way of reducing the number of tab stops in the keyboard interface.
+
+Based on the above definition, it appears that as it currently stands, all elements within a toolbar _should_ be interactive controls. This includes support for [disabled controls](#disabled-controls).
+
+### Use Cases
+
+Used anywhere someone may want to visually and structurally group related interactive controls.
+
+- Tom is reading his favorite internet newspaper. Upon finishing an article, he notices that it has accrued nearly one thousand comments. Tom decides to read the comments and he ends up writing a response to someone he disagrees with. As he gets ready to submit, he realizes it will be easier for others to read his comment if he formats it. Tom sees a toolbar with formatting options and is able to bold all the words he deems important. He submits his comment and waits to see what happens next.
+
+### Prior Art/Examples
+- [Fluent UI](https://fluentsite.z22.web.core.windows.net/components/toolbar/definition)
+- [Material UI](https://material-ui.com/api/toolbar/)
+- [WAI-ARIA](https://w3c.github.io/aria-practices/examples/toolbar/toolbar.html)
+
+---
+
+### API
+
+*Component name:*
+- `fast-toolbar`
+
+*Attributes:*
+- `orientation`: an enum
+    - horizontal (default)
+    - vertical
+
+*Slots:*
+- `default`
+
+### Anatomy and Appearance
+
+*Template:*
+```html
+<template
+    role="toolbar"
+    orientation="${x => x.orientation}"
+>
+    <slot></slot>
+</template>
+```
+
+## Implementation
+
+**With visible label:**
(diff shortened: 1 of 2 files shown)


chrisdholt:
> Are there any implications on styling of controls that are slotted into the toolbar? For example, do our button's change style when inside a toolbar?

Good question - my initial thought is that the toolbar would likely need to set the background context. Outside of that, ideally we would support flexibility where if I wanted to put `stealth` and `accent` buttons inside the toolbar, both would be accepted. Thoughts @bheston?

specs/toolbar.md
```diff
@@ -0,0 +1,127 @@
...
+
+> Typically, disabled elements are not focusable when navigating with a keyboard. However, in circumstances where discoverability of a function is crucial, it may be helpful if disabled controls are focusable so screen reader users are more likely to be aware of their presence.
+
+Following the "flexible" principle of FAST-DNA, I think we should support an API which enables exposing disabled controls to users. This allows implementors to decide if their circumstances warrant exposing that functionality or not. A simple solution here is that if an author wants to expose a disabled control, they would simply add it to the tabindex:
+
+```html
+<fast-button disabled tabindex="0">Disabled</fast-button>
+```
+
+#### Labels
```
nicholasrice, line 125: Would it make sense to surface a "label" slot for this?

```
<fast-toolbar>
   <h3 slot="label">This is a toolbar</h3>
   ....
</fast-toolbar>
```

specs/toolbar.md
```diff
@@ -0,0 +1,127 @@
...
+- [WAI-ARIA](https://w3c.github.io/aria-practices/examples/toolbar/toolbar.html)
+
+---
+
+### API
+
+*Component name:*
+- `fast-toolbar`
+
+*Attributes:*
```
scomea, line 30: Do we need a prop to specify where initial focus goes if not the first item?

specs/toolbar.md
```diff
@@ -0,0 +1,127 @@
...
+    <select>
+        <option>Option 1</option>
+        <option>Option 2</option>
+        <option>Option 2</option>
+    </select>
+    <div>
+        <fast-button>More</fast-button>
+    </div>
+</fast-toolbar>
+```
```
scomea, line 95: Do we need to support "groups" of elements within the toolbar to create visual subgroups within a toolbar but keep the children of the group as part of the toolbar's linear focus queue?
 ```
<Toolbar>
            <Button>Item 1</Button>
            <button>Item 2</button>
            <button>Item 3</button>
            <ToolbarItemGroup>
                <button>Item 4</button>
                <button>Item 5</button>
            </ToolbarItemGroup>
</Toolbar>
```

File: specs/toolbar.md [hunk 4]
```diff
@@ -0,0 +1,127 @@
...
+- `orientation`: an enum
+    - horizontal (default)
+    - vertical
+
+*Slots:*
+- `default`
+
+### Anatomy and Appearance
+
+*Template:*
```
eljefe223, line 40: Do we want to show the template or the shadow DOM here? Some specs do the shadow root and some the template. To me the template is implementation specific which is not what the spec is meant to show. To me its more valuable to show the DOM structure you want the component to generate. But open to peoples opinions, just think we should be consistent.

chrisdholt: I think both can be helpful - in the case where children are passed to a default slot, the shadow DOM is largely dependent on what is being passed. One of the things I like about showing the template in certain cases is that it illustrates what is coming from attrs and what is hard-coded to the control.

specs/toolbar.md
chrisdholt: I think the ideal scenario here would be to place focus only if focus has not already been placed within the toolbar. In that scenario, implementors can use existing methods (`.focus()`) on their HTMLElement, which seems more intuitive than a specialized API.

specs/toolbar.md
chrisdholt: It would be ideal if we don't have to and I think the initial intent should be to attempt without this. My intent with the divs shown in the example above was to specifically illustrate that the nesting structure within the toolbar shouldn't really matter.

specs/toolbar.md
chrisdholt: We certainly could as a convenience - I don't know that it gives a huge amount of value here compared to other places, but it doesn't hurt.

nicholasrice: yea just consistency I think

PR #3010, opened 2020-04-24
fix(fast-element): attr bindings preserved during upgrade

EisenbergEffect:
# Description

Fixes a bug in custom element upgrade when attributes are bound using a property binding.

## Motivation & context

When an element has attributes bound with a property binding, those properties get set on the element instance before the browser's custom element upgrade process kicks in. As a result, the getter/setter needed for observation is not yet defined because those live on the prototype of the not yet upgraded element. This causes the property values to shadow the getter/setter when the prototype is upgraded.

This bug was fixed for observable properties, but not for attributes. This PR fixes the issue by bringing further unification to observable and attr properties by introducing the concept of an "accessor" and providing a way to get all accessors for an object.

There appeared to be a minor perf regression as part of this work, however I couldn't be sure. So, I did a little perf work in related areas to ensure that the overall performance didn't regress as a result of the new feature (extensible accessors).

## Issue type checklist

There are technically some low-level breaking changes but they shouldn't affect anyone.

## Process & policy checklist


EisenbergEffect:
@chrisdholt This PR is a fix for the Graph folks and might affect the Blazor folks as well. I'd like to see if we can get this in for a release.

PR #3013, opened 2020-04-24
chore: add heading spec

bheston:
# Description

Adds a heading spec based on current typography spec.

Wanted to post an initial pass with most features per discussion with Nick around implementation options.

## Motivation & context

Initial structure for adaptive heading styles.

## Issue type checklist

## Process & policy checklist


specs/README.md
```diff
@@ -20,12 +20,13 @@ Here you'll find specifications for custom elements and other library features.
...
 | [Flipper](../packages/web-components/fast-components/src/flipper/flipper.spec.md) | :white_check_mark: | :white_check_mark: |
 | [Form associated custom elements](../packages/web-components/fast-components/src/form-associated/form-associated-custom-element.spec.md) | :white_check_mark: | :white_check_mark: |
+| [Heading](./typography/heading.md) | :white_check_mark: | -- |
 | Listbox | -- | -- |
 | [Menu](./menu/menu.md) | :white_check_mark: | -- |
 | [Number field](./number-field.md) | :white_check_mark: | -- |
 | [Progress](../packages/web-components/fast-components/src/progress/progress.spec.md) | :white_check_mark: | :white_check_mark: |
 | [Radio](../packages/web-components/fast-components/src/radio/radio.spec.md) | :white_check_mark: | :white_check_mark: |
- RadioGroup | -- | -- |
+| RadioGroup | -- | -- |
```
chrisdholt, line 29: If we're making a change here can it include changing this to:
```suggestion
Radio group
```

specs/typography/heading.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Scoped standard heading element map
+
+The easiest option for accessible content when you know how many levels you have.
+
+```TypeScript
+headingStyles.standardElementMap(levels: 3, includeBase: false)
+```
+
+Applies the styles to the standard heading elements for the desired number of levels so the adaptive styling is used without needing to add and maintain class names. Starts directly above the base type ramp size or optionally includes that size depending on stylistic preference.
```
chrisdholt, line 100: Can you elaborate on how these would apply to standard headings? Is this a utility rather than a component? It seems like this is applying values directly to standard elements, but with our current implementations it seems like this would be scoped to within the Shadow DOM and I can't quite think of a place where we would hard code a heading level in our components.

specs/typography/heading.md
```diff
@@ -0,0 +1,241 @@
...
+h3 {
+    font-size: var(--fast-type-ramp-plus-1-size);
+    line-height: var(--fast-type-ramp-plus-1-height);
+}
+
+h4, h5, h6 { /* This rule may not be necessary, but it helps identify usage or config issues. */
+    font-size: var(--fast-type-ramp-base-size);
+    line-height: var(--fast-type-ramp-base-height);
+}
+```
```
chrisdholt, line 135: Is this something we would set, or is this something someone would implement based on guidance?

specs/typography/heading.md
bheston: Yes, I think under the current idea of using standard markup it's more of a utility than a component. I believe this is the direction we were considering for elements like this where there wasn't much else being added by a component.

Per what I recalled from a discussion a while back, we could set styles and let them chain down the same way we're doing with the css variables that represent the design system properties. That is what I was thinking here, that whatever the final interface is for this, you would do it in conjunction with your root design system setup. I wasn't exactly sure what that would look like, which is why I commented these as pseudo interfaces.

Let's setup a quick chat if my assumptions are incorrect.

specs/typography/heading.md
bheston: Is this referring to the whole css block? If so I was intending that it's the output of the `standardElementMap` interface. Per my above comment, I thought this would most likely be configured at the base of your app, and inherited throughout the dom. Could also be done for a particular section, say with different density or a different type ramp.

chrisdholt: Yeah, the whole block, I think my primary question was how we're exposing this as a distributed set of code. Is this particularly guidance and naming?

specs/typography/heading.md
chrisdholt: Yes, I don't think these are components, so that assumption is correct. As a utility I assume we're providing a methodology that someone could use to create generate this. Could this be configured to use classes if I wanted instead of the standard element?

bheston: Yes, you could use `basicRamp` or `simplePageStyleMap` or write your own use case specific version of the latter.

specs/typography/heading.md
bheston: We just tweaked the naming of the underlying ramp in Nick's PR, but otherwise the intent was that logic in that function generated this block or equivalent according to the parameters. The logic may not be clear though, so for instance if you asked for 4 levels of headings we would add a definition for h4 and shift all the values (+x) by one. Similarly if you set the option to have your smallest heading be the same as your body we would shift by one the other way.

The idea is quick mapping to the standard elements you configure that you're going to use, but sized according to the design system and progressing from your body text.

This is my preferred and recommended usage pattern, so I want to make sure it's defined clearly. Maybe it needs another example.

chrisdholt: > The idea is quick mapping to the standard elements you configure that you're going to use, but sized according to the design system and progressing from your body text.

We had discussed having other methods of applying these (not just affixing to the principle of mapping directly to standard elements). Is that not the case with the proposal here? I think I'm still trying to understand what is guidance vs. implementation. If our utility only supports generating a font ramp for standard elements I think it may be too prescriptive. If the utility generates a ramp and ties it to either a list of values (my own classes) or semantic elements then I think that's fine. 

I like the intent of using semantic elements, but I still get the feeling that our utility here assumes that most documents flow like standard markdown or documentation sites, whereas the visual hierarchy for many marketing sites and apps varies based on context as well. The visual representation of the elements may not always be directly tied to their DOM hierarchy. If this utility could support and enable *both* of those scenarios I think it adds a lot of value.

Just to be clear, not trying to be difficult here :). While it's fine to have our default implementations be more prescriptive and default towards our preference/guidance, I think we need to be careful to not impose that guidance through implementation. Just trying to parse what the export is and whether the code enforces this principle of styling the semantic elements as it directly relates to hierarchy.

specs/typography/heading.md
```diff
@@ -0,0 +1,241 @@
...
+
+![](./images/standard-map-sample.png)
+
+#### Semantically named use case based styles
+
+An option for applying names to common scenarios. Serves as a common example as a guide for how to create your own semantic styles.
+
+```TypeScript
+headingStyles.simplePageStyleMap()
```
nicholasrice, line 160: I think what we'll want to do for these style-sets is expose them as:
1. Individual stylesheets that can be composed into larger stylesheets
2. Expose the "set" as a stylesheet that can be exposed globally or through a root control

bheston: I was reconsidering the underlying approach here. Based on some comment from Chris, I think the idea of mapping a ramp value to a use case name should exist at the general type ramp level. Like a function or whatever that would take name / value pairs where like:
page-title: 3,
section-title: 2,
item-title: 1,
caption: -1,
random-large-text: 6

The heading example here would use that and essentially be an example of usage, or could be brought in as you mention.

specs/typography/heading.md
bheston: My comment a little further down might address your concern. It at least allows a structured way for someone to flexibly define whatever style mappings they need. I still like the idea of having a basic one that we think is useful, but we could provide examples in the docs for someone to use instead as well.

PR #3015, opened 2020-04-25
chore: add accordion web component spec

eljefe223:
# Description
This adds the spec for `<fast-accordion/>`, `<fast-accordion-header/>` and `<fast-accordion-panel/>`

## Issue type checklist

## Process & policy checklist


specs/accordion.md
```diff
@@ -0,0 +1,173 @@
...
+
+*Attributes:*
+- `id` - string
+
+### Anatomy and Appearance
+
+```HTML
+<!-- shadow root -->
+<div>
+    <h3>
```
nicholasrice, line 75: I don't think we want to hard-code the heading level here, because different use-cases could have different requirements for what level it should be. Perhaps the heading is created by the user if / when they need it?

specs/accordion.md
```diff
@@ -0,0 +1,173 @@
...
+        <button
```
nicholasrice, line 76: Is there a case here where someone might want to swap out the whole button implementation?

specs/accordion.md
```diff
@@ -0,0 +1,173 @@
...
+<!-- end shadow root -->
+```
+
+---
+
+## Implementation
+
+```HTML
+<fast-accordion>
+    <fast-accordion-header>Panel one</fast-accordion-header>
```
nicholasrice, line 151: Did you consider a template structure where the header is a slot of the panel?

```
<fast-accordion>
   <fast-accordion-panel>
      <h3 slot="header">panel header</h3>
      <p>I'm some content</p>
    </fast-accordion-panel>
</fast-accordion>

File: specs/accordion.md [hunk 4]
```diff
@@ -0,0 +1,173 @@
...
+
+### Use Cases
+
+- Frank finds a spelling error in the FAST-DNA documentation. He submits a pull request (PR) to fix the error and goes to eat lunch while waiting for the maintainers to review his PR. Upon returning from lunch he notices a widget on the screen saying certain status checks failed. He clicks on the headline of the widget and it reveals additional information stating that he needs to run prettier to ensure proper formatting. He clicks the headline again and the additional information collapses. Frank runs prettier and his PR is completed successfully.
+
+- On April 16th Randy wakes up in a cold sweat because he realized he made a mistake in filing his taxes with the IRS. He opens his laptop and goes to the IRS website and sees a link to "frequently asked questions". Randy sees a list of questions on the page with an arrow next to it indicating that answers are hidden. He sees a question titled, "What should I do if I made a mistake on my federal turn that I've already filed?" Randy clicks the question and sees a section detailing out the exact steps to take. With his anxiety eased, Randy goes back to bed and falls asleep.
```
EisenbergEffect, line 12: 🤣

specs/accordion.md
```diff
@@ -0,0 +1,173 @@
...
+  - multi
+
+*Events*
+- `change: CustomEvent`
+  - no custom data
+  - bubbles
+
+**Accordion Header**
+*Component names:*
+- `fast-accordion-header`
```
EisenbergEffect, line 51: I'm wondering if we should update the tabs component with this naming. In tabs we have `fast-tabs`, `fast-tab`, and `fast-tab-panel` and with accordion we have `fast-accordion`, `fast-accordion-header`, and `fast-accordion-panel`. If we change `fast-tab` to `fast-tab-header` that would create more symmetry across our components and I think reduce chance of accidental typos/confusion between `fast-tabs` and `fast-tab`. Thoughts? @eljefe223 @chrisdholt

specs/accordion.md
eljefe223: Yeah, I think we have to do something like this. I realized that having the `header` and `panel` separated would not be possible in order to get the alternating stacked DOM structure we need. For example:

```
<header/>
<panel/>
<header/>
<panel/>
<header/>
<panel/>
```

eljefe223: I think we will still need `<fast-accordion-header/>` which will be inside the something like `<fast-accordion-item/>` like so:

```HTML
<fast-accordion>
    <fast-accordion-item>
        <fast-accordion-header headingtag="h2">Header One</fast-accordion-header>
        Panel one Content
    </fast-accordion-item>
    <fast-accordion-item expanded>
        <fast-accordion-header headingtag="h2" expanded>Header Two</fast-accordion-header>
        Panel two content
    </fast-accordion-item>
    <fast-accordion-item>
        <fast-accordion-header headingtag="h2">Header Three</fast-accordion-header>
        Panel three content
    </fast-accordion-item>
</fast-accordion>
```

specs/accordion.md
eljefe223: Thats all @chrisdholt , our resident creative use case author.

specs/accordion.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+---
+
+## Implementation
+
+```HTML
+<fast-accordion>
+    <fast-accordion-item>
+        <fast-accordion-header headingtag="h2">Header One</fast-accordion-header>
```
EisenbergEffect, line 167: Was there a reason this couldn't actually be an `h2` tag with a `heading` slot?

specs/accordion.md
eljefe223: I'm not sure it should be `<fast-tabs-header/>` as header doesn't really fit the terminology of tabs. I think if we were going to follow the same pattern as mentioned about it would be:

`<fast-tabs-tab/>` and `<fast-tab-panel/>`

EisenbergEffect: `fast-tabs-tab` is pretty awkward. Better to keep it as is in that case.

specs/accordion.md
eljefe223: because we need to know if its expanded or not to display the proper glyph

eljefe223: image

EisenbergEffect: Does the glyph need to be inside of the header tag or could it be beside it?

eljefe223: I think it should be inside... That's how the w3c example shows it.

https://w3c.github.io/aria-practices/examples/accordion/accordion.html

EisenbergEffect: Ok, that makes sense.

Note that making the tag swaps right now may be a bit difficult with the templating engine today. I've got features planned that will make that easier but doubtful they'll get in soon based on my current workload.

nicholasrice: We can probably circumvent tag-switching by using `role="heading"` and `aria-lavel="1|2|3|4|5|6"`

nicholasrice: I'm still not entirely convinced we need a separate `header` item:

```
<fast-accordion-item>
  <span slot="heading">I'm the heading</slot>
</fast-accordion-item>

<fast-accordion-item heading-level="3">
  <span slot="heading">I'm the heading</slot>
</fast-accordion-item>

<fast-accordion-item heading-level="3">
  <span slot="heading">I'm the heading</slot>
  <span slot="expanded-indicator">Some glyph</span>
  <span slot="collapsed-indicator">Some other glyph</span>
</fast-accordion-item>
```

EisenbergEffect: If we do have a separate header item, we definitely want to use the role and level aria attrs instead of switching the tag. That will be more performant even if I had the new template features in place.

eljefe223: The only down side I see here is the the `<button/>` needs to be inside the `<h1/>` , `<h2/>`, ... etc. Maybe the template could look something like:

```HTML
    <div role="heading" aria-level=${x => x.headinglevel}>
        <button
            type="button"
            part="button"
            class="button"
            aria-expanded=${x => x.expanded}
            aria-controls=`${x => x.id}-panel`
            id=${x => x.id}
        >
            <slot name="heading"></slot>
            <slot name="expanded-glyph" part="expanded-glyph"></slot>
            <slot name="collapsed-glyph" part="collapsed-glyph"></slot>
       </button>
    </div>
    <div
        id=`${x => x.id}-panel`
        role="region"
        aria-labelledby=${x => x.id}
    >
        <slot></slot>
    </div>
```

specs/accordion.md
chrisdholt: It was the first page I found when looking that included one and it seemed like a good opportunity :)

specs/accordion.md
```diff
@@ -0,0 +1,184 @@
...
+- default
+- heading
+- glyph
+
+### Anatomy and Appearance
+
+```HTML
+<!-- shadow root -->
+<div>
+    <div role="heading" aria-level="3">
```
chrisdholt, line 78: we likely don't want to fix the heading level here.

specs/accordion.md
```diff
@@ -0,0 +1,184 @@
...
+        <button
+            type="button"
+            class="button"
+            part="button"
+            aria-expanded="true"
+            aria-controls="accordion1-panel"
+            id="accordion1"
+        >
+            <slot name="heading" part="heading">Panel one</slot>
+            <slot name="glyph" part="glyph"></slot>
```
chrisdholt, line 88: Accordians support (and we have had several designs within MSFT) additional interactive elements (more menu, checkbox) within the header. How can we light up support for those if I want a checkbox before the header content and a more menu after?

specs/accordion.md
eljefe223: It's not fixed, its showing what the shadow root would render. Here is the template:
```HTML
 <div role="heading" aria-level=${x => x.headinglevel}>
        <button
            type="button"
            part="button"
            class="button"
            aria-expanded=${x => x.expanded}
            aria-controls=`${x => x.id}-panel`
            id=${x => x.id}
        >
            <slot name="heading"></slot>
            <slot name="expanded-glyph" part="expanded-glyph"></slot>
            <slot name="collapsed-glyph" part="collapsed-glyph"></slot>
       </button>
    </div>
    <div
        id=`${x => x.id}-panel`
        role="region"
        aria-labelledby=${x => x.id}
    >
        <slot></slot>
    </div>
```

specs/accordion.md
```diff
@@ -0,0 +1,184 @@
...
+        <span slot="glyph">^</span>
+        Panel three content
+    </fast-accordion-item>
+</fast-accordion>
+```
+
+### States
+
+
+### Accessibility
```
chrisdholt, line 174: Can we link to the interaction model here at the very least?

specs/accordion.md
```diff
@@ -0,0 +1,184 @@
...
+    </fast-accordion-item>
+    <fast-accordion-item>
+        <span slot="heading">Panel three</span>
+        <span slot="glyph">^</span>
+        Panel three content
+    </fast-accordion-item>
+</fast-accordion>
+```
+
+### States
```
chrisdholt, line 171: It would be great to have a summary of the various states and how one might get into them?

specs/accordion.md
chrisdholt: 👍

specs/accordion.md
eljefe223: The heading slot can take any elements you want to pass to it

specs/accordion.md
eljefe223: The `button` is now a `div` so you can now pass whatever elements you want to the `heading` slot.

```HTML
<div role="heading" aria-level="3">
        <div
            class="header"
            part="header"
            aria-expanded="true"
            aria-controls="accordion1-panel"
            id="accordion1"
        >
            <slot name="heading" part="heading">Panel one</slot>
            <slot name="glyph" part="glyph"></slot>
       </div>
    </div>
```

chrisdholt: Is that div conveyed to AT as a button or actionable element?

specs/accordion.md
```diff
@@ -0,0 +1,193 @@
...
+
+---
+
+### API
+
+**Accordion**
+*Component Name*
+- `fast-accordion`
+
+*Attributes:*
```
nicholasrice, line 39: It might be good to land a convention regarding how we name attributes with multiple words (`"expandmode"` vs `"expand-mode"`, `"headinglevel"` vs `"heading-level"`) so that we don't deviate around the project. For instance, the DesignSystemProvider uses dashes to delineate word-breaks: `"background-color"`.

@microsoft/fast-dna-collaborators thoughts and preferences?

specs/accordion.md
eljefe223: Did some re structuring based on team discussions. The start, end and glyph slots are now on the outside of the button. We will use a pseudo element on the button to make the entire header visually appear hovered / active. While allowing the start and end content to live outside the button.

specs/accordion.md
eljefe223: Personally I think to be more inline with HTML attributes we **should not** use the hyphens.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

Interesting there is only one multiword attribute in the about list that uses a hyphen `accept-charset`.

radium-v: For `dataset` and data attributes, dashes (after the initial `data-`) suggest arrays or lists of properties.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset

nicholasrice: > ...suggest arrays or lists of properties

I'm not follow this, can you clarify? Is there an example somewhere?

My takeaway from the above dataset docs is that a data *attribute name* (eg `data-foo-bar`) would be accessible via the dataset property by the *property name* `fooBar` whereas `data-foobar` would be accessible on the property `foobar`. I think it would follow from that then that a camel-cased property would reflect as a hyphen separated attribute name. While I prefer that convention, there are native properties in HTML that *don't* follow that pattern (`readonly` vs `readOnly`) 

I personally would prefer to see hyphen separated simply because it is easier to read, especially for longer attribute names that concatenate 3+ words (eg `"basehorizontalspacingmultiplier"` vs `"base-horizontal-spacing-multiplier"`)

radium-v: > I'm not follow this, can you clarify? Is there an example somewhere?

My bad, that part was from a different thing entirely. I confused the functionality of a custom implementation with the expected behavior for kebab-to-camelCase conversion.

eljefe223: I'm ok with adding hyphens, Is this something we should talk about retroactively fixing in other components?

nicholasrice: @eljefe223 Probably, though I was hoping others would chime in. I'll open an issue to discuss and we can address components as needed based on where that shakes out. Thanks!

specs/accordion.md
```diff
@@ -0,0 +1,193 @@
...
+- `expanded` - boolean
+- `id` - string
+- `heading-level` - enum
+  - 1
+  - 2
+  - 3 - default
+  - 4
+  - 5
+  - 6
```
chrisdholt, line 62: Is this always a heading?

eljefe223: Thats my interpretation.
"Each accordion header element contains a button that controls the visibility of its content panel."

https://w3c.github.io/aria-practices/examples/accordion/accordion.html

PR #3018, opened 2020-04-27
feat(fast-element): add a render method to templates for pure templating

EisenbergEffect:
# Description

Adds a `render` method to `ViewTemplate` which takes a data source (to bind to) and an element or element id to render the template into.

## Motivation & context

I realized that our templating engine isn't dependent on the browser supporting web components and can be used independently to render with a little bit of boilerplate code. Other teams have expressed interest in this capability esp. since it would allow them an upgrade path from IE11 to Edge by starting with basic templates and then moving to web components, all based on the same library. So, I added a simple method to enable pure templating that gets rid of the boilerplate code.

#### Example

```TypeScript
export class GreetingViewModel {
  @observable name: string = 'World';
}

const template = html<GreetingViewModel>`
  <h1>Hello ${x => x.name}!</h1>
`;

const vm = new GreetingViewModel(); // create a view model
const view = template.render(vm, 'some-div'); // render the template to some-div using the vm

// some time later
vm.name = 'FAST-DNA'; // causes async batched update to DOM

// some time later
view.dispose(); // removes view from DOM and unbinds all
```

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-element/src/template.ts
+++ b/packages/web-components/fast-element/src/template.ts
@@ -4,13 +4,17 @@ import { DOM } from "./dom";
 import { Behavior, BehaviorFactory } from "./directives/behavior";
 import { Directive } from "./directives/directive";
 import { BindingDirective } from "./directives/binding";
-import { ExecutionContext, Expression } from "./observation/observable";
+import {
+    ExecutionContext,
+    Expression,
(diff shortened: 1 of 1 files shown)


janechu:
This is very cool

PR #3027, opened 2020-04-28
feat: add radio-group component

marjonlynch:
# Description
* Add radio group component
* tabindex support for tab navigation
* arrow key navigation

radio-group component allows children radio buttons to be treated as a single select group.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/radio-group/README.md
+++ b/packages/web-components/fast-components/src/radio-group/README.md
@@ -1,4 +1,4 @@
 # fast-radio-group
-An implementation of a [radio](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) as a form-connected web-component.
+An implementation of a [radio-group](https://w3c.github.io/aria-practices/#radiobutton).
 
 For more information view the [component specification](./radio-group.spec.md).
--- a/packages/web-components/fast-components/src/radio-group/README.md
+++ b/packages/web-components/fast-components/src/radio-group/README.md
@@ -1,0 +1,4 @@
+# fast-radio-group
+An implementation of a [radio](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) as a form-connected web-component.
+
+For more information view the [component specification](./radio-group.spec.md).
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -32,13 +32,15 @@
     </div>
 
     <div role="toolbar" style="display: flex; flex-direction: column;margin-top: 12px;">
-        <button>go</button>
-        <fast-radio-group required="true" name="navigation">
-            <fast-radio value="back">back</fast-radio>
-            <fast-radio value="forward">forward</fast-radio>
-            <fast-radio value="refresh">refresh</fast-radio>
-        </fast-radio-group>
-        <button>stop</button>
+        <div style="display: flex; flex-direction: row; align-items: stretch; width: 100%">
+            <button style="width: 40px;margin-right: 4px;">go</button>
+            <fast-radio-group required="true" name="navigation">
+                <fast-radio value="back">back</fast-radio>
+                <fast-radio value="forward">forward</fast-radio>
+                <fast-radio value="refresh">refresh</fast-radio>
+            </fast-radio-group>
+            <button style="width: 40px;margin-left: 4px;">stop</button>            
+        </div>
     </div>
 
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -4,6 +4,7 @@
     <h4>Defaults</h4>
     <div>
         <fast-radio-group name="numbers">
+            <label slot="label">Numbers</label>
             <fast-radio value="one">One</fast-radio>
             <fast-radio value="two">Two</fast-radio>
         </fast-radio-group>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -32,4 +32,12 @@
             <fast-radio value="excel">Excel</fast-radio>
         </fast-radio-group>
     </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label4">Preset selected-value</label>
+        <fast-radio-group selected-value="maverick" aria-labelledby="label4" name="fruits">
+            <fast-radio value="iceman">Ice Man</fast-radio>
+            <fast-radio value="maverick">Maverick</fast-radio>
+        </fast-radio-group>
+    </div>
 </fast-design-system-provider>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -14,6 +14,7 @@
         <fast-radio-group aria-labelledby="label1" name="fruits">
             <fast-radio value="apples">Apples</fast-radio>
             <fast-radio value="oranges">Oranges</fast-radio>
+            <fast-radio value="bananas">Bananas</fast-radio>
         </fast-radio-group>
     </div>
 
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -16,7 +16,7 @@
 
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
         <label id="label1">Outside label</label>
-        <fast-radio-group aria-labelledby="label1" name="fruits">
+        <fast-radio-group required="true" aria-labelledby="label1" name="fruits">
             <fast-radio value="apples">Apples</fast-radio>
             <fast-radio value="oranges">Oranges</fast-radio>
             <fast-radio value="bananas">Bananas</fast-radio>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -1,0 +1,19 @@
+<fast-design-system-provider use-defaults>
+    <h1>Radio Group</h1>
+
+    <h4>Defaults</h4>
+    <div>
+        <fast-radio-group name="numbers">
+            <fast-radio value="one">One</fast-radio>
+            <fast-radio value="two">Two</fast-radio>
+        </fast-radio-group>
+    </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label1">Outside label</label>
+        <fast-radio-group aria-labelledby="label1" name="fruits">
+            <fast-radio value="apples">Apples</fast-radio>
+            <fast-radio value="oranges">Oranges</fast-radio>
+        </fast-radio-group>
+    </div>
+</fast-design-system-provider>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -41,9 +41,11 @@
 
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
         <label id="label4">Preset selected-value</label>
-        <fast-radio-group selected-value="maverick" aria-labelledby="label4" name="fruits">
-            <fast-radio value="iceman">Ice Man</fast-radio>
+        <fast-radio-group value="maverick" aria-labelledby="label4" name="best-pilot">
+            <fast-radio value="ice-man">Ice Man</fast-radio>
             <fast-radio value="maverick">Maverick</fast-radio>
+            <fast-radio value="viper">Viper</fast-radio>
+            <fast-radio value="jester">Jester</fast-radio>
         </fast-radio-group>
     </div>
 </fast-design-system-provider>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -31,6 +31,16 @@
         </fast-radio-group>
     </div>
 
+    <div role="toolbar" style="display: flex; flex-direction: column;margin-top: 12px;">
+        <button>go</button>
+        <fast-radio-group required="true" name="navigation">
+            <fast-radio value="back">back</fast-radio>
+            <fast-radio value="forward">forward</fast-radio>
+            <fast-radio value="refresh">refresh</fast-radio>
+        </fast-radio-group>
+        <button>stop</button>
+    </div>
+
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
         <label id="label2">Disabled radio group</label>
         <fast-radio-group disabled aria-labelledby="label2" name="cars">
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -11,6 +11,13 @@
     </div>
 
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <fast-radio-group name="players">
+            <label slot="label">Best basketball players</label>
+            <fast-radio value="airjordan">Michael Jordan</fast-radio>
+        </fast-radio-group>
+    </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
         <label>Single radio (no group)</label>
         <fast-radio value="single-life">Single life</fast-radio>
     </div>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -10,6 +10,11 @@
     </div>
 
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label>Single radio (no group)</label>
+        <fast-radio value="single-life">Single life</fast-radio>
+    </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
         <label id="label1">Outside label</label>
         <fast-radio-group aria-labelledby="label1" name="fruits">
             <fast-radio value="apples">Apples</fast-radio>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -16,4 +16,20 @@
             <fast-radio value="oranges">Oranges</fast-radio>
         </fast-radio-group>
     </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label2">Disabled radio group</label>
+        <fast-radio-group disabled aria-labelledby="label2" name="cars">
+            <fast-radio value="lambo">Lamborghini</fast-radio>
+            <fast-radio value="ferari">Ferari</fast-radio>
+        </fast-radio-group>
+    </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label3">readonly radio group</label>
+        <fast-radio-group readonly aria-labelledby="label3" name="office">
+            <fast-radio value="word">Word</fast-radio>
+            <fast-radio value="excel">Excel</fast-radio>
+        </fast-radio-group>
+    </div>
 </fast-design-system-provider>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -28,6 +28,12 @@
             <fast-radio value="apples">Apples</fast-radio>
             <fast-radio value="oranges">Oranges</fast-radio>
             <fast-radio value="bananas">Bananas</fast-radio>
+            <fast-radio value="kiwi">Kiwi</fast-radio>
+            <fast-radio value="grapefruit">Grapefruit</fast-radio>
+            <fast-radio value="mango">Mango</fast-radio>
+            <fast-radio value="blueberries">Blueberries</fast-radio>
+            <fast-radio value="strawberries">Strawberries</fast-radio>
+            <fast-radio value="pineapple">Pineapple</fast-radio>
         </fast-radio-group>
     </div>
 
--- a/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
@@ -49,7 +49,6 @@ Radio group allows the user to be presented with a list of all the options visib
 
 ### Anatomy and Appearance
 *Parts:*
-- label
 - control
 
 *Template:*
--- a/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
@@ -39,6 +39,10 @@ Radio group allows the user to be presented with a list of all the options visib
 
 *Attributes:*
 - `selected-value` - allows user to set the initial radio selected within the group independent of value.
+- `readonly`
+  - If there is a selected radio button in the group it should be submitted with the form but none of the radio buttons should be editable.
+- `disabled`
+  - All of the radio buttons should be disabled from user interaction and will not be submitted with the form data.
 
 *Slots:*
 - default slot for radio items
--- a/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
@@ -56,7 +56,10 @@ Radio group allows the user to be presented with a list of all the options visib
 <template
     role="radiogroup"
 >
-    <slot></slot>
+    <slot name="label"></slot>
+    <div class="control" part="control">
+        <slot ${slotted("slottedRadioButtons")}> </slot>
+    </div>
 </template>
 ```
 
--- a/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
@@ -48,6 +48,21 @@ Radio group allows the user to be presented with a list of all the options visib
 - `change` - raise the change event for external parties to be informed of the selected value change.
 
 ### Anatomy and Appearance
+*Parts:*
+- label
+- control
+
+*Template:*
+```
+<template
+    role="radiogroup"
+>
+    <div part="control" class="control">
+        <slot></slot>
+    </div>
+</template>
+```
+
 
 ### Accessibility
 
--- a/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.spec.md
@@ -5,9 +5,6 @@
 As defined by the W3C:
 > A radio group is a set of checkable buttons, known as radio buttons, where no more than one of the buttons can be checked at a time. Some implementations may initialize the set with all buttons in the unchecked state in order to force the user to check one of the buttons before moving past a certain point in the workflow.
 
-### Background
-
-
 ### Use Cases
 
 Radio group allows the user to be presented with a list of all the options visible which can facilitate the comparison of choice.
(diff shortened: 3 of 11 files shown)


packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,174 @@
...
+                radio.setAttribute("tabindex", "-1");
+            }
+        });
+
+        if (this.selectedValue === undefined) {
+            radioButtons[0].setAttribute("tabindex", "0");
+        }
+    }
+
+    private getFilteredRadioButtons = (): any[] => {
```
EisenbergEffect, line 98: I promise to make this easier in the future through the core library. Thanks for working it out in the meantime and for your patience!

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
```diff
@@ -0,0 +1,54 @@
...
+    :host(:${focusVisible}) .control {
+        box-shadow: 0 0 0 1px var(--neutral-focus) inset;
+        border-color: var(--neutral-focus);
+    }
+
+    :host(.disabled) {
+        opacity: var(--disabled-opacity);
+    }
+
+    @media (forced-colors: active) {
```
EisenbergEffect, line 40: Check with @nicholasrice since we've got a new way to conditionally include the high contrast styles.

packages/web-components/fast-components/src/radio-group/README.md
```diff
@@ -0,0 +1,4 @@
+# fast-radio-group
+An implementation of a [radio](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) as a form-connected web-component.
+
```
chrisdholt, line 3: Copy/paste? Can we replace this with a link pointing to the W3C specification as there is no standard element for radiogroup?

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
```diff
@@ -0,0 +1,76 @@
+# Radio group
+
+## Overview
+
+As defined by the W3C:
+> A radio group is a set of checkable buttons, known as radio buttons, where no more than one of the buttons can be checked at a time. Some implementations may initialize the set with all buttons in the unchecked state in order to force the user to check one of the buttons before moving past a certain point in the workflow.
+
+### Background
+
+
```
chrisdholt, line 10: Can we delete this if it's not used?

packages/web-components/fast-components/src/radio-group/fixtures/base.html
```diff
@@ -0,0 +1,49 @@
...
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label3">readonly radio group</label>
+        <fast-radio-group readonly aria-labelledby="label3" name="office">
+            <fast-radio value="word">Word</fast-radio>
+            <fast-radio value="excel">Excel</fast-radio>
+        </fast-radio-group>
+    </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label4">Preset selected-value</label>
```
chrisdholt, line 43: Should we surface a label slot?

packages/web-components/fast-components/src/radio-group/fixtures/base.html
```diff
@@ -0,0 +1,49 @@
...
+        <fast-radio-group selected-value="maverick" aria-labelledby="label4" name="fruits">
+            <fast-radio value="iceman">Ice Man</fast-radio>
+            <fast-radio value="maverick">Maverick</fast-radio>
+        </fast-radio-group>
```
chrisdholt, line 47: Also fruits? Hollywood and Viper would be great additions here :)

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
```diff
@@ -0,0 +1,76 @@
...
+### API
+
+*The key elements of the component's public API surface:*
+
+**Radio Group**
+*Component name:*
+- `fast-radio-group`
+
+*Attributes:*
+- `selected-value` - allows user to set the initial radio selected within the group independent of value.
```
chrisdholt, line 41: Should this just be value? `selected-value` seems somewhat redundant.

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
```diff
@@ -0,0 +1,14 @@
+import { html, slotted } from "@microsoft/fast-element";
+import { RadioGroup } from "./radio-group";
+
+export const RadioGroupTemplate = html<RadioGroup>`
+    <template
+        role="radiogroup"
+        ?aria-disabled="${x => x.disabled}"
+        ?aria-readonly="${x => x.readOnly}"
+    >
+        <div part="control" class="control">
```
chrisdholt, line 10: Do we need this div? It seems like the `:host` can serve this purpose based on the current markup and model.

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -91,6 +91,9 @@ export class Radio extends FormAssociated<HTMLInputElement> {
 
     public connectedCallback(): void {
         super.connectedCallback();
+        if (this.parentElement?.tagName !== "FAST-RADIO-GROUP") {
+            this.setAttribute("tabindex", "0");
+        }
```
chrisdholt, line 96: What happens if I want to set tabindex myself as well in my own implementation? Should we check for the attr before setting it here?

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
```diff
@@ -0,0 +1,76 @@
...
+- `disabled`
+  - All of the radio buttons should be disabled from user interaction and will not be submitted with the form data.
+
+*Slots:*
+- default slot for radio items
+
+*Events*
+- `change` - raise the change event for external parties to be informed of the selected value change.
+
+### Anatomy and Appearance
```
chrisdholt, line 53: Are we leaving this blank?

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,174 @@
...
+                    } else if (index - 1 < 0) {
+                        index = group.length - 1;
+                    } else {
+                        index -= 1;
+                    }
+                }
+                break;
+        }
+    };
+}
```
chrisdholt, line 174: Are we intentionally not supporting the differing keyboard behavior for toolbars as described in the WAI-ARIA spec referenced in the spec? https://w3c.github.io/aria-practices/#for-radio-group-contained-in-a-toolbar

packages/web-components/fast-components/src/radio-group/fixtures/base.html
nicholasrice: +1

packages/web-components/fast-components/src/radio-group/fixtures/base.html
```diff
@@ -0,0 +1,49 @@
...
+        <fast-radio-group readonly aria-labelledby="label3" name="office">
+            <fast-radio value="word">Word</fast-radio>
+            <fast-radio value="excel">Excel</fast-radio>
+        </fast-radio-group>
+    </div>
+
+    <div style="display: flex; flex-direction: column;margin-top: 12px;">
+        <label id="label4">Preset selected-value</label>
+        <fast-radio-group selected-value="maverick" aria-labelledby="label4" name="fruits">
+            <fast-radio value="iceman">Ice Man</fast-radio>
```
nicholasrice, line 45: I think "iceman" would be hyphenated here...

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
marjonlynch: oh yea I saw that on tabs, I'll change it to be like that

packages/web-components/fast-components/src/radio-group/README.md
marjonlynch: sometimes the link also describes radio group but I'll double check this one and the others

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
marjonlynch: yup, deleting

packages/web-components/fast-components/src/radio-group/fixtures/base.html
marjonlynch: oh if you'll let me keep this in, I'll fix it up good, "I feel the need, the need for speed!"

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
marjonlynch: will fix

packages/web-components/fast-components/src/radio-group/fixtures/base.html
chrisdholt: I'd be disappointed if you removed it altogether!

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: yep will fix

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: Ok I'm assuming, through some research, that I can check to see if the parent has a role of "toolbar", is this how you would expect radio-group to know it's in this toolbar context?

chrisdholt: I'm not sure without digging deeper into this implementation. It's possible that this would be nested within some divs so it may not be the immediate parent...

packages/web-components/fast-components/src/radio-group/README.md
marjonlynch: fixed, it wasn't a copy/paste error as that link described mostly radio-group, but I changed it to be the w3c description of radio-group

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
```diff
@@ -0,0 +1,52 @@
...
+        --input-size: calc((${heightNumber} / 2) + var(--design-unit));
+        align-items: center;
+        outline: none;
+        margin: calc(var(--design-unit) * 1px) 0;
+        user-select: none;
+        position: relative;
+        flex-direction: row;
+        transition: all 0.2s ease-in-out;
+    }
+
```
nicholasrice, line 19: Is all this styling really necessary? It seems this control is primarily concerned with *behavior* and that styling should be left to the radio's themselves.

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
```diff
@@ -0,0 +1,14 @@
+import { html, slotted } from "@microsoft/fast-element";
+import { RadioGroup } from "./radio-group";
+
+export const RadioGroupTemplate = html<RadioGroup>`
+    <template
+        role="radiogroup"
+        ?aria-disabled="${x => x.disabled}"
```
nicholasrice, line 7: these two aren't true boolean attributes - they need to either not exist or be set to `"true"` or `"false"`

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: so this is like looking for direction, it can be all the way up the chain, seems like.

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,171 @@
...
+    keyCodeArrowDown,
+    keyCodeArrowLeft,
+    keyCodeArrowRight,
+    keyCodeArrowUp,
+} from "@microsoft/fast-web-utilities";
+import { FASTRadio } from "../radio";
+
+export class RadioGroup extends FASTElement {
+    @attr({ attribute: "readonly", mode: "boolean" })
+    public readOnly: boolean; // Map to proxy element
```
nicholasrice, line 12: this isn't proxy connected - same for all comments below

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
chrisdholt: Only thing I would point to here is if there is value in supporting appearance/orientation from the radiogroup itself so folks don't need to wrap themselves, add a class, etc. But agree it seems like we have a lot of styling here.

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
marjonlynch: added template and parts

packages/web-components/fast-components/src/radio/radio.ts
```diff
@@ -91,6 +91,14 @@ export class Radio extends FormAssociated<HTMLInputElement> {
 
     public connectedCallback(): void {
         super.connectedCallback();
+        if (
```
nicholasrice, line 94: I'm wondering if this ties the radio too tightly to the radio-group. What if someone wanted to use a *different* mechanism to manage focus on the radio?

Is there a way to just make the control react correctly to change to the tabindex attribute?

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,171 @@
...
+                radio.setAttribute("tabindex", "-1");
+            }
+        });
+
+        if (this.value === undefined) {
+            radioButtons[0].setAttribute("tabindex", "0");
+        }
+    }
+
+    private getFilteredRadioButtons = (): any[] => {
```
nicholasrice, line 95: can this be `HTMLElement[]`?

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
marjonlynch: removed

packages/web-components/fast-components/src/radio/radio.ts
marjonlynch: This seems problematic because the connectedCallback or if something was in the template for radio happens after connectedCallback in the radio-group, thus overriding anything set by radio-group or any other container presumably. I'm not sure it should be set at all in FASTRadio

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: it should be FASTRadio[] now and I think better typing, there is still an any in the cast to push into a typed array, but now that's the only any anywhere

chrisdholt: @marjonlynch I don't know that it will always be FASTRadio. It could be any HTMLElement with a role of radio, yes?

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
```diff
@@ -0,0 +1,85 @@
...
+- `selected-value` - allows user to set the initial radio selected within the group independent of value.
+- `readonly`
+  - If there is a selected radio button in the group it should be submitted with the form but none of the radio buttons should be editable.
+- `disabled`
+  - All of the radio buttons should be disabled from user interaction and will not be submitted with the form data.
+
+*Slots:*
+- default slot for radio items
+
+*Events*
```
nicholasrice, line 47: hmmm actually - I'm not sure this element should raise a change-event. This is really just a semantic grouping

packages/web-components/fast-components/src/radio/radio.ts
nicholasrice: Okay lets leave this as-is then - we can always circle back if we need to and figure out a more robust solution.

nicholasrice: Thanks for investigating

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: ok that is a good point

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
marjonlynch: wouldn't that mean if someone wanted to listen when the user selection changes for the group they have to subscribe to each of the internal radio button change events?

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
nicholasrice: I think there is more that can be removed here. 

1. I don't think this should be setting `--input-size`
2. This doesn't get focus, so it probalby doesn't need the outline removal
3. user-select is handled by the inputs
4. The group doesn't get foucs, so it doesn't need focus-visible styles
5. The disabled style should be handled by the individual disabled controls - we shouldn't be setting the *region* to disabled opacity

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
nicholasrice: no they can still *listen* on the group

```js
document.querySelector('fast-radio-group')
    .addEventListener(
        "change", () => console.log("something inside of me fired a change event")
    )
```

packages/web-components/fast-components/src/radio-group/radio-group.ts
chrisdholt: I think that's fine, but I don't know that we can expect it to always be FASTRadio

packages/web-components/fast-components/src/radio-group/fixtures/base.html
marjonlynch: I'm adding a slot named "label" and adding back in a control div so I can control the vertical stacking of the label if slotted and the radios, it also means I can easily add vertical and horizontal classes to accommodate orientation attribute for radio group

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
marjonlynch: think I did all the removals this time

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: I will tidy this up but I think it is now doing the right thing when it's in a toolbar and not in a toolbar. I added a toolbar example in fixtures to test.

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
marjonlynch: removed the event from here and updated the for the label slot

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: this is fixed as far as the type but I think I might still need to put a check in to see that the item actually conforms to the RadioControl interface. Maybe that interface name needs to change as well, naming is hard

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
```diff
@@ -0,0 +1,87 @@
...
+### API
+
+*The key elements of the component's public API surface:*
+
+**Radio Group**
+*Component name:*
+- `fast-radio-group`
+
+*Attributes:*
+- `value` - allows user to set the initial radio selected within the group independent of value.
```
nicholasrice, line 38: What does "independent of value" mean in this context? I'm not following how this attribute should be used.

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
```diff
@@ -0,0 +1,15 @@
...
+import { RadioGroup } from "./radio-group";
+
+export const RadioGroupTemplate = html<RadioGroup>`
+    <template
+        role="radiogroup"
+        aria-disabled="${x => x.disabled}"
+        aria-readonly="${x => x.readOnly}"
+    >
+        <slot name="label"></slot>
+        <div class="control" part="control">
```
nicholasrice, line 11: "control" seems like an awkward name for the region containing the controls. Perhaps this should be "group" or "region" or "control-region" or something else to describe this as a container element?

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,217 @@
...
+export class RadioGroup extends FASTElement {
+    @attr({ attribute: "readonly", mode: "boolean" })
+    public readOnly: boolean;
+    private readOnlyChanged(): void {
+        const filteredRadios = this.getFilteredRadioButtons();
+        if (filteredRadios !== undefined) {
+            filteredRadios.forEach((radio: HTMLInputElement) => {
+                if (this.disabled) {
+                    radio.setAttribute("readonly", "");
+                } else {
```
nicholasrice, line 20: FYI - you generally just set the underlying *property* that the attribute is mapped to:
```
radio.readOnly = true;
// vs
radio.setAttribute('readonly', '');
```

Not only is it less verbose, but it is also generally significantly faster: https://jsperf.com/get-set-attribute-vs-direct https://jsbench.me/dbk9sookk1/1

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    @attr
+    public value: string;
+
+    @observable slottedRadioButtons: RadioControl[];
+    private selectedRadio: RadioControl | null;
+
+    constructor() {
+        super();
+        this.addEventListener("keydown", this.keydownHandler);
```
nicholasrice, line 58: Is there a reason this is set up in the constructor whereas the "change" handler is set up in the `connectedCallback`?

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.addEventListener("change", this.handleRadioChange);
```
nicholasrice, line 63: I think you'll want to remove these listeners in the `disconnectedCallback`

packages/web-components/fast-components/src/radio-group/radio-group.spec.md
marjonlynch: copy and paste error, fixing

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
marjonlynch: we used positioning-region for slider and I think another control, so I'll use that for consistency sake unless there are objections

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: Ok yea, Since it was typed HTMLInputElement that slipped my mind as a possibility, will fix

marjonlynch: also I thought I was adding the attribute and not setting it to true which further confused things.

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: no, the change handler should be up there as well, originally the change handler was on the individual radios but that is no longer the case, will move

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: rob said not to do that, looking up again why

marjonlynch: I moved this to the constructor so below now applies...originally the way I had it yes, It should have been removed there
from Rob, 
the constructor is only called once but connect and disconnect can be called multiple times
you can add the listener in your constructor and not worry about removing it as long as you are only adding it to your own DOM
if you are listening to other elements outside of your own DOM, then you would need to add them in connect and remove them in disconnect

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,254 @@
...
+                    } else if (index === group.indexOf(this.selectedRadio!)) {
+                        break;
+                    } else if (index - 1 < 0) {
+                        index = group.length - 1;
+                    } else {
+                        index -= 1;
+                    }
+                }
+                break;
+        }
```
scomea, line 252: I think you need to call preventDefault() on mouse/key events you handle.

packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+    };
+
+    public keydownHandler = (e: KeyboardEvent): void => {
+        const group: RadioControl[] = this.getFilteredRadioButtons();
+        let index: number = 0;
+        const toolbar: HTMLElement | null | undefined = this.parentElement?.closest(
+            '[role="toolbar"]'
+        );
```
scomea, line 182: Are there other circumstances where the radio group could be hosted in a container that also uses key navigation that we need to consider (ie. data grids, menus, etc...)?

scomea: ("key navigation" = using arrow keys)

marjonlynch: I'm going by https://w3c.github.io/aria-practices/#for-radio-group-contained-in-a-toolbar which only indicated a special scenario for toolbar, otherwise I think it is supposed to use the pattern described for Not in a toolbar

chrisdholt: +1 to the above - currently the only delta in implementation from a keyboarding standpoint is the one called out for toolbar in the spec

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: I didn't notice this in any other web components and frankly it messes things up if I do this in radio-group. I have a question in with Rob to see if I'm missing something here to validate. Will re-work things if necessary

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
```diff
@@ -0,0 +1,16 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "../styles";
+
+export const RadioGroupStyles = css`
+    ${display("flex")} :host {
+        align-items: flex-start;
```
chrisdholt, line 6: Clarifying question - is flex doing anything on the host if we have a positioning region under it which is "flexed"? Do we need the positioning region? It seems like one of these is redundant. Consider that I want the radios to wrap at small viewports or whatnot, am I applying that behavior to the host element or to the part of "positioning region"?

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
```diff
@@ -0,0 +1,16 @@
...
+        margin: calc(var(--design-unit) * 1px) 0;
+        position: relative;
```
chrisdholt, line 8: I'm not sure you need to set a position of relative here; you can likely remove.

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
marjonlynch: The host flex was to align the slotted label we wanted which could have a different alignment than the radios themselves. That was the thinking, if we didn't have the label slot it wouldn't be needed or only one flex layout would be needed.

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: confirmed with Rob, just calling preventDefault on the click, the rest of the handlers need to stay as is.

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
marjonlynch: removing

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: I believe all of these are now updated to the more efficient method

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: confirmed with Nick offline that the current RadioControl interface version satisfies this request.

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: This could be better but added more comments and cleaned it up

PR #3033, opened 2020-04-29
feat: add a web component definition and mapper to JSON schema for tooling

### LINKED ISSUE #2800: Create a definition and mapper for web components to interface with the tooling
# Background

The JSON schema intended to be used with the `<Form />` consist of easily editable props (in the case of React) or attributes (in the case of regular web components). This does not cover other items that may be necessary to expand on such as CSS properties, custom tag, and events to name a few.

## Requirements

1. JSON schema defined for a custom elements definition. An effort is already underway to conceptualize this [here](https://github.com/webcomponents/custom-elements-json/issues/3).
    - Should define the tag
    - Should define the attributes
    - Should define the slots (named and default)
    - Allows for validation when contributors are devising their own custom component definitions
    - Can be used to validate against any mappers that transform a custom element definition into a usable JSON schema for the `<Form />`
2. Mapper for any custom elements definition for the above JSON schema to transform it into a JSON schema for use in the `<Form />`

## Implementation

Example JSON schema for custom element definitions, note that the tag item definition should be imported as its own schema to map against individual definitions written by users in the `fast-comonents` library:

```ts
{
    $schema: "http://json-schema.org/schema#",
    id: "custom-element-definition",
    title: "A definition for web components",
    type: "object",
    version: 1,
    required: ["version"],
    properties: {
        version: {
            title: "The version of the schema the web components map to",
            type: "number",
            enum: [1], // update this when new schema versions are released
        },
        tags: {
            type: "array",
                items: {
                    type: "object",
                    properties: {
                        name: {
                            type: "string"
                        },
                        description: {
                            type: "string"
                        },
                        attributes: {
                            type: "array",
                            items: {
                                type: "object",
                                properties: {
                                    name: {
                                        type: "string",
                                    },
                                    type: {
                                        type: "string",
                                    },
                                    description: {
                                        type: "string",
                                    },
                                    default: {
                                        type: "string",
                                    },
                                    required: {
                                        type: "boolean",
                                    },
                                }
                            }
                        },
                        slots: {
                            type: "array",
                            items: {
                                type: "object",
                                properties: {
                                    name: {
                                        type: "string"
                                    },
                                    description: {
                                        type: "string"
                                    },
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

Example custom element definitions, note that these may be individual items in the `fast-components` library which are then compiled to this single definition:

```ts
{
    version: 1,
    tags: [
        {
            name: "fancy-button",
            description: "A fancier button",
            attributes: [
                {
                    name: "disabled",
                    type: "boolean",
                    description: "The disabled state",
                    default: "false",
                    required: false
                }
            ],
            slots: [
                {
                    name: "",
                    description: "The default slot, content in this slot displays directly inside the button"
                }
            ]
        }
    ]
}
```

Result of a mapper to make these definition usable in the message system, `<Form />`, and `<Navigation />`:

```ts
import { linkedDataSchema } from "@microsoft/fast-tooling";

{
    $schema: "http://json-schema.org/schema#",
    id: "fancy-button",
    mapsToTagName: "fancy-button",
    title: "The definition for fancy button custom element",
    type: "object",
    properties: {
        disabled: {
            mapsToAttribute: "disabled",
            title: "The disabled state",
            type: "boolean",
            default: false
        },
        // this should be generated in a way that avoids conflicts with attribute names
        // such as here, where attributes may not have uppercase in their naming
        Slot: {
            mapsToSlot: "", // the slot name this maps to
            title: "The default slot, content in this slot displays directly inside the button",
            ...linkedDataSchema
        }
    }
}
```

janechu:
# Description

Adds definitions and conversion tools for web components:
- A TypeScript definition for web component definitions
- A JSON schema definition for the web component definitions
- A converter to convert the data defining the web component to a JSON schema that the `@microsoft/fast-tooling` can use.

## Motivation & context

Resolves #2800 

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tooling/src/data-utilities/web-component.ts
```diff
@@ -0,0 +1,80 @@
...
+     * The name of the tag
+     */
+    name: string;
+
+    /**
+     * The description of this web component tag
+     */
+    description: string;
+
+    /**
```
nicholasrice, line 54: Does this definition purposefully omit a "property" vs "attribute" distinction?

janechu: For now properties are not defined, there is intended to be a distinction between "property" and "attribute". This definition was taken from an ongoing discussion into what this definition should look like.

See: https://github.com/webcomponents/custom-elements-json/issues/3#issuecomment-546750767

EisenbergEffect: @janechu Does this light up VS intellisense automatically if we put the generated json file into our package or do we need to do some other work for that?

janechu: There's some work involved, I haven't investigated fully but this is very close to the definition needed for VS Code web components plugin referenced in those issues. It is incomplete, I only included what the tooling needs for now with the intent to expand it later.

EisenbergEffect: @janechu What do you feel our chances are of getting that generated in time for release? I don't know how your priority list looks so wasn't sure if it would fit in. A friend of mine tried out the components yesterday and noted how much improved he felt the dev experience would be if we had that on launch. He had used some other libraries that had that, so he had a point of reference.

nicholasrice: Does this need to be generated? Could we write them manually for release?

Also - could you add the above link as a reference the the code-file? I think that is important context for these interfaces

EisenbergEffect: Definitely *could* be hand-written if needed and I'd be willing to do that if needed for sure.

janechu: I think it would be faster to write them, it shouldn't be that and I've got a pretty hefty backlog to get through for our release date.

PR #3036, opened 2020-04-29
chore: remove babel-jest given it's now comes with jest

awentzel:
# Description
Removes `babel-jest` because it now comes with `jest`.

## Motivation & context
Security alert for [`kind-of`](https://github.com/microsoft/fast-dna/network/alert/yarn.lock/kind-of/open) indicated a dependency inside `babel-jest`, as a result, I discovered `babel-jest` is now built into `jest` and so it's no longer a dev dependency we need. 

This does not fix the security alert, but, does reduce our footprint and move the security alert to `jest`.

## Issue type checklist

## Process & policy checklist


package.json
```diff
@@ -92,7 +92,7 @@
     "@types/lodash-es": "^4.17.0",
     "@types/node": "^9.4.7",
     "babel-core": "^7.0.0-bridge.0",
-    "babel-jest": "^25.4.0",
+
```
janechu, line 95: There shouldn't be empty lines in package.json files

nicholasrice:
Where are you reading that babel-jest "comes with" jest? Their setup documents mention installing babel-jest here: https://jestjs.io/docs/en/getting-started#using-babel

awentzel:
> Where are you reading that babel-jest "comes with" jest? Their setup documents mention installing babel-jest here: https://jestjs.io/docs/en/getting-started#using-babel

@nicholasrice https://github.com/babel/babel-jest

nicholasrice:
I suppose if tests pass then it works then it works. However, that page is claiming two things:
1. The source *moved* - the plugin still exists, the repo just lives somewhere else
2. That babel-jest is "automatically loaded" <-- this info is 3 years old and directly contradicted by Jest's own documentation.

PR #3050, opened 2020-05-01
feat: adding typeramp values

nicholasrice:
# Description
Adds typeramp to the design-system provider as configurable design-system properties.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/design-system-provider/index.ts
+++ b/packages/web-components/fast-components/src/design-system-provider/index.ts
@@ -153,6 +153,108 @@ export class FASTDesignSystemProvider extends DesignSystemProvider
     })
     public disabledOpacity: number;
 
+    @attr({ ...fromView, attribute: "type-ramp-minus-2-size" })
+    @designSystemProperty({
+        cssCustomProperty: "type-ramp-minus-2-size",
+        default: "10px",
+    })
+    public typeRampMinus2Size: string;
+
+    @attr({ ...fromView, attribute: "type-ramp-minus-2-height" })
+    @designSystemProperty({
+        cssCustomProperty: "type-ramp-minus-2-height",
+        default: "16px",
+    })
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-components/src/design-system-provider/index.ts
```diff
@@ -153,6 +153,108 @@ export class FASTDesignSystemProvider extends DesignSystemProvider
...
     public disabledOpacity: number;
 
+    @attr({ ...fromView, attribute: "type-ramp-minus-2-size" })
+    @designSystemProperty({
+        cssCustomProperty: "type-ramp-minus-2-size",
+        default: "10px",
+    })
+    public typeRampMinus2Size: string;
+
+    @attr({ ...fromView, attribute: "type-ramp-minus-2-height" })
```
chrisdholt, line 163: @bheston @nicholasrice do we want `line-height` here or just height? How important are the semantics of the `@attr`?

nicholasrice: If we do that I'd say we should also use "font-size" instead of "size" for consistency. I'm good with it, but curious on @bheston's feedback as this was the name that was outlined in the spec.

chrisdholt: I agree there - I don't think it's blocking but we use height elsewhere and it is a different property altogether. Similar with "size" to your point.

bheston: I see that point, and considered it when writing the spec. I could go for being more explicit, but I think I was trying to keep it shorter and more generic than actual attribute names. I can't think of a way where being more explicit hurts us though, so maybe that's best. I just double-checked some of the other typography taxonomy work that's going on and they seem to be leaning toward full attribute names, so let's make that change. 👍

PR #3053, opened 2020-05-04
chore(fast-element): configure production build and add test setup

EisenbergEffect:
# Description

Prepares the build for production release and sets up both Node.js and browser-based testing.

## Motivation & context

We need to fix our build output so that it works with modern browsers by itself as well as within Node.js and with popular bundlers. We also need tests 😄 

There are a few things this PR puts in place.

* All `imports` have been updated to use file extensions. This ensures that the entire library will work natively in the browser with `<script type="module">` lazy loading as well as with all major bundlers and Node.js in ESM mode.
* The build now outputs an ESM build to the `esm` folder, so that it's isolated from other build output.
* The build now outputs all individual `d.ts` files to a `dts` folder, so that they are isolated from other build output. A concatenated `.d.ts` for the library is output in the root of `dist`.
* The build now outputs a single script file build of the full library, so that it can be used with `<script type="module">` for efficient eager loading without a bundler.
* The build now outputs an `api-report.md` file that will help identify missing documentation and prevent us from unintentional breaking changes.
* The build now outputs a json file that contains all extracted documentation info, suitable as input into our documentation static generator.
* The build outputs statistics for minified and gzip sizes for the single-file build so that size can more easily be monitored across changes. The full library with all features and no tree shaking is approx. 8.88KB minified and gzipped.
* Mocha and Chai have been added for authoring tests.
* Tests run in Node.js using `mocha`, `jsdom`, and `jsdom-global`.
* Tests run in browser using Karma.
* Both Node.js and Browser tests can (and do) have custom test setup code configured with `setup-node` and `setup-browser` respectively.
* Istanbul coverage reporting is in place.
* Several commands have been added to the `package.json` to facilitate both Node and browser testing, including debugging and watching.
* The `package.json` is now set to `type: "module"` so that it runs as esm in Node.js.
* A singe cross-environment test was validated to ensure all build and setup work correctly. More tests will follow in successive PRs once this is merged.
* Improved the output of `npm pack` so that only dist and docs are part of the release (along with required readme, package.json, etc.)

## Issue type checklist

The `index.js` file of the `directives` subfolder was removed as unneeded. This may cause some dependents to break if they were importing through that path.

## Process & policy checklist


packages/web-components/fast-element/tsconfig.json
```diff
@@ -1,16 +1,20 @@
...
     "compilerOptions": {
-        "outDir": "dist",
-        "baseUrl": ".",
+        "declarationDir": "dist/dts",
+        "outDir": "dist/esm",
         "strictNullChecks": true,
-        "target": "es2015"
+        "target": "es2015",
+        "module": "ESNext",
+        "removeComments": true,
```
nicholasrice, line 9: Isn't this is going to remove jsDoc comments documenting the feature/API that is surfaced in IDEs?

packages/web-components/fast-element/tsconfig.json
```diff
@@ -1,16 +1,20 @@
...
+        "types": [
+            "mocha",
+            "webpack-env"
```
nicholasrice, line 12: I don't see `@types/webpack-env` enumerated as a dependency - where is that coming from?

packages/web-components/fast-element/tsconfig.json
```diff
@@ -1,16 +1,20 @@
...
-        "outDir": "dist",
-        "baseUrl": ".",
+        "declarationDir": "dist/dts",
+        "outDir": "dist/esm",
         "strictNullChecks": true,
-        "target": "es2015"
+        "target": "es2015",
+        "module": "ESNext",
+        "removeComments": true,
+        "types": [
```
nicholasrice, line 10: Curious why you're needing to manually enumerate the `types` field here - are these not found by default in the `@types` dir?

packages/web-components/fast-element/tsconfig.json
EisenbergEffect: Good point. How should we handle this? Would it be ok to have the comments in dts only but not in the distributed source? If so, we can use this plugin: https://github.com/Swatinem/rollup-plugin-dts to build a single dts file for the library. Or should we leave the comments in both the esm and dts builds but just remove them from the single script-file build?

packages/web-components/fast-element/tsconfig.json
EisenbergEffect: Oops. That shouldn't be needed. I'll remove it.

packages/web-components/fast-element/tsconfig.json
nicholasrice: If the goal was just to remove them from the script file I think the latter is the most straight-forward. I'd rather leave it to consuming apps to run their bundles through a minifier if they don't want comments.

packages/web-components/fast-element/tsconfig.json
EisenbergEffect: Double oops. It's actually needed in the browser-setup code. I'll add the dependency. Not sure why that wasn't there.

packages/web-components/fast-element/tsconfig.json
EisenbergEffect: Both of these types introduce globals, so they are needed here. Removing them causes build errors for either the tests or the test setup code.

packages/web-components/fast-element/tsconfig.json
EisenbergEffect: Ok fixed.

packages/web-components/fast-element/karma.conf.cjs
```diff
@@ -0,0 +1,162 @@
...
+  '--disable-background-timer-throttling',
+  '--disable-backing-store-limit',
+  '--disable-boot-animation',
+  '--disable-cloud-import',
+  '--disable-contextual-search',
+  '--disable-default-apps',
+  '--disable-extensions',
+  '--disable-infobars',
+  '--disable-translate',
+];
```
chrisdholt, line 19: should we run with experimental-web-platform-features?

packages/web-components/fast-element/package.json
```diff
@@ -15,19 +15,62 @@
...
+    "test-node": "mocha --reporter min --exit dist/esm/__test__/setup-node.js dist/esm/**/*.spec.js",
+    "test-node:verbose": "mocha --reporter spec --exit dist/esm/__test__/setup-node.js dist/esm/**/*.spec.js",
+    "test-chrome": "karma start karma.conf.cjs --browsers=ChromeHeadlessOpt --single-run --coverage",
+    "test-chrome:verbose": "karma start karma.conf.cjs --browsers=ChromeHeadlessOpt --single-run --coverage --reporter=mocha",
+    "test-chrome:watch": "karma start karma.conf.cjs --browsers=ChromeHeadlessOpt --coverage --watch-extensions js",
+    "test-chrome:debugger": "karma start karma.conf.cjs --browsers=ChromeDebugging --watch-extensions js",
+    "test-chrome:verbose:watch": "karma start karma.conf.cjs --browsers=ChromeHeadlessOpt --coverage --watch-extensions js --reporter=mocha",
+    "test-chrome:verbose:debugger": "karma start karma.conf.cjs --browsers=ChromeDebugging --watch-extensions js --reporter=mocha",
+    "test-firefox": "karma start karma.conf.cjs --browsers=FirefoxHeadless --single-run --coverage",
+    "test-firefox:verbose": "karma start karma.conf.cjs --browsers=FirefoxHeadless --single-run --coverage --reporter=mocha"
```
chrisdholt, line 38: Should we have a fixed `test` command we can run on CI? `lerna run test` will run all test scripts across all packages currently. Additionally, we can quickly leverage something similar with Yarn 2 when we move there (`yarn workspaces foreach test`)

EisenbergEffect: I can alias "test" to "test-node:verbose" for CI. Not a problem.

packages/web-components/fast-element/karma.conf.cjs
EisenbergEffect: I'm not sure to be honest. Are we using any experimental features? Maybe TrustedTypes?

chrisdholt: I believe focus-visible is still behind the flag, I'm not sure if we want to run with it on or off.

chrisdholt: Off is probably fine for now, I'll resolve

packages/web-components/fast-element/src/attributes.ts
```diff
@@ -1,6 +1,6 @@
-import { Accessor, Observable } from "./observation/observable";
-import { DOM } from "./dom";
-import { Notifier } from "./observation/notifier";
+import { Accessor, Observable } from "./observation/observable.js";
```
janechu, line 1: Is there any configuration option we can use in order to sidestep adding the filename extension?

EisenbergEffect: Not that I know of. There have been a couple of long discussion threads in TS with no clear action yet though.

EisenbergEffect:
@chrisdholt I'm still fixing a few issues but will run that before merging to be sure.

packages/web-components/fast-element/src/styles.ts
```diff
@@ -101,7 +101,7 @@ export class AdoptedStyleSheetsStyles extends ElementStyles {
     public removeStylesFrom(target: StyleTarget): void {
         const sourceSheets = this.styleSheets;
         target.adoptedStyleSheets = target.adoptedStyleSheets!.filter(
-            (x: CSSStyleSheet) => !sourceSheets.includes(x)
```
nicholasrice, line 104: Out of curiosity - what is the impetus for this change? Perf?

EisenbergEffect: When I was setting things up I was getting type errors on the old version. So, I changed this to avoid those errors. I think it was just wrong configuration at the time though, so it could be changed back. I just left it in. I'm not sure the perf implications...probably negligible for this scenario would be my guess.

EisenbergEffect:
Holding on merging this for now. I think I'm going to do a bit more work around dts generation and generating what we need for docs so I can integrate this with the upcoming documentation site.

EisenbergEffect:
Ok, I've added a new commit that adds the following:

* Generates a single d.ts. file for the library
* Generates an API report that shows us what is documented and can give us insight into when we are making breaking changes.
* Generates a json file with all extracted TSDoc info

All this is handled by `api-extractor`. With this in place, we can then use `api-documenter` against the json file as part of our Docusaurs setup, which will generate linked markdown pages for all APIs. Down the road, if we want to get fancy, we can write our own generator over the same json and plug it into Docusaurus, but I think the default is fine for initial release.

EisenbergEffect:
@nicholasrice @chrisdholt If you can take another look at this, I think it's ready to merge now.

PR #3059, opened 2020-05-04
chore: add stack panel to base component set

scomea:
# Description

Note:  this is resurrection of a previous PR that went stale (https://github.com/microsoft/fast-dna/pull/2578).  The functionality is needed for an in-progress data-grid component (https://github.com/microsoft/fast-dna/tree/users/scomea/data-grid).

The stack panel component enables authors to create a horizontally or vertically scrolling list of child items that can be virtualized (ie. not rendered to the dom) when not within the visible region of the scrolling container.  Authors provide the span (height or width depending on orientation) of each item, either a single value when all child items have identical spans, or as an array of values when children have variable spans. Because the stackpanel knows the space to allocate to each item it can create layouts without actually having to render each child which can be a significant performance gain.

The stack panel accepts children in the usual way for html markup.
`<StackPanel>`
`               <button/>`
`                <button/>`
`               <button/> `
`                etc...`
</StackPanel>`

The stackpanel's role is to manage virtualization at React node level and that other parent components, like a data grid or a feed for example, would handle more data driven inputs and handle converting those to nodes to be rendered  in an efficient manner (and would be the right place to manage paging data also).  

For example the prototype [data grid component](https://github.com/microsoft/fast-dna/tree/users/scomea/data-grid) takes in data and interacts with the stackpanel to only create nodes only for the rows that are in the region that is being rendered.  Browsing a large dataset in this scenario could end up with a stackpanel with children that look something like this:

`<StackPanel itemSpan={[500000, 100, ...more spans...100, 300000]}>`
`                <div>Filler child that represents first 5000 rows</div>`
`                <dataRow>Row index 5001</dataRow>`
`               ...more data rows...`
 `             <dataRow>Row index 6000</dataRow>`
 `           <div>Filler child that represents rows after 6000</div>`
`</StackPanel>`

As the user scrolls the data grid tracks the scroll position and updates the children as required.  The first and last child merely exist to take up virtual space and their spans change as the component is scrolled and which data rows are actually passed to the data grid changes accordingly.

## Motivation & context
https://github.com/microsoft/fast-dna/issues/2501

The primary motivation for this component is to enable more performant scrolling containers for authors who need it.  The trade off is that authors need to either specify that all children have the same span (height for a vertical panel, width for a horizontal one) or provide a fast callback to calculate height.  This is not a solution for authors that want items to be sized via html/css layout.

However the performance gain for this exchange can be dramatic with large collections of children.

Consider a stackpanel with 200 child items rendered with virtualization turned off which takes about 9 seconds of churn on the main thread when first rendered:
image

Turning on virtualization reduces that time to less than a second:
image

Further, upping the item count from 200 to 10000 does not have a significant impact:
image

## Issue type checklist

## Process & policy checklist


packages/react/fast-components-react-base/src/stack-panel/stack-panel.props.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * The default span of each item in pixels (width in horizontal orientation, height in vertical)
+     * Provide a single value for panels where all items have the same value, or an array in the case where
+     * items have varying spans.
+     * default is 100
+     */
+    itemSpan?: number | number[];
+
+    /**
```
nicholasrice, line 31: The description vs the property name seem divergent to me - is this expecting a *pixel value* or a number of items? If the latter, perhaps using "count" instead of "length" would be more clear.

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,580 @@
+import { isArray, isFunction } from "util";
```
nicholasrice, line 1: "util" looks like the wrong import source - that is a nodejs module. Did you mean "lodash-es"?

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,580 @@
...
+
+    /**
+     * Stores last scroll position from scroll events
+     */
+    private lastRecordedScroll: number = 0;
+
+    /**
+     * Stores last scroll position from scroll events
+     */
+    private scrollLayoutUpdateTimer: NodeJS.Timer;
```
nicholasrice, line 91: This type doesn't seem accurate to me - this won't always execute in a NodeJS environment

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,580 @@
...
+     * Render the items
+     */
+    private renderItems(): React.ReactFragment[] {
+        return React.Children.map(this.props.children, this.renderItem);
+    }
+
+    /**
+     * Test if a node is cloneable
+     */
+    private isClonableElement(node: React.ReactNode): node is React.ReactElement<any> {
```
nicholasrice, line 247: Do you really need to alias the React function? `React.isValidElement` is a type guard

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,580 @@
...
+    private updateItemPositions = (): void => {
+        const itemPositions: ItemPosition[] = [];
+        let currentItemStart: number = 0;
+        const childrenAsArray: React.ReactNode[] = React.Children.toArray(
+            this.props.children
+        );
+        const itemCount: number = childrenAsArray.length;
+
+        for (let i: number = 0; i < itemCount; i++) {
+            const itemSpan: number = isArray(this.props.itemSpan)
```
nicholasrice, line 302: could probably just use `Array.isArray`

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,580 @@
...
+            this.setState({
+                direction: newDirection,
+            });
+        }
+    };
+
+    /**
+     *  gets the current direction
+     */
+    private getDirection = (): Direction | null => {
```
nicholasrice, line 491: Does this ever return `null`?

packages/react/fast-components-react-base/src/stack-panel/stack-panel.props.ts
scomea: Updated with "count"

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: done

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: used the window timer instead

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: nah, removed

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: yup

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: nope, removed

packages/react/fast-components-react-base/src/index.ts
```diff
@@ -86,6 +86,11 @@ import SliderTrackItem from "./slider-track-item";
 export { SliderTrackItem };
 export * from "./slider-track-item";
 
+// holding off on exporting this until we are confident of the implementation
```
janechu, line 89: This is going to look strange in our changelog with the current title as it implies the stack panel is now available, can you change it to a chore instead and use `feat` once these lines are safe to uncomment?

packages/react/fast-components-react-base/src/stack-panel/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+### Usage
+
+Virtualization:  When virtualization is enabled the component does not render child items that are not in the visible area of the stackpanel's scrolling region or within the defined buffer zone (ie one or more items outside the viewport).
+
+Note that virtualized elements can't get focus so typically authors will want at least one out of view item instanciated otherwise it will not be possible to tab outside of the viewport.
+
+Authors should be aware that this component writes to the width/left or top/height css style attributes of children on render depending on the orientation of the stack panel.
+
+Keys: authors should strongly consider providing keys (https://reactjs.org/docs/lists-and-keys.html) to child elements to maintain stable identies for child items.
```
janechu, line 15: nit: *identities

packages/react/fast-components-react-base/src/stack-panel/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Note that virtualized elements can't get focus so typically authors will want at least one out of view item instanciated otherwise it will not be possible to tab outside of the viewport.
+
+Authors should be aware that this component writes to the width/left or top/height css style attributes of children on render depending on the orientation of the stack panel.
+
+Keys: authors should strongly consider providing keys (https://reactjs.org/docs/lists-and-keys.html) to child elements to maintain stable identies for child items. 
+
+### Props
+
+ enableVirtualization:  Whether child elements are virtualized or not. Defaults to true.
```
janechu, line 19: seems like the names of these props are now out of sync with the actual interface, can you update?

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,570 @@
...
+            scrollPos = this.rootElement.current.scrollTop;
+        } else {
+            scrollPos = RtlScrollConverter.getScrollLeft(
+                this.rootElement.current,
+                this.state.direction
+            );
+            scrollPos = this.state.direction === Direction.rtl ? -scrollPos : scrollPos;
+        }
+
+        return scrollPos;
```
janechu, line 516: nit, just a suggestion but you can avoid declaring a variable and do:
```ts
return this.props.orientation === Orientation.vertical
    ? this.rootElement.current.scrollTop
    : this.state.direction === Direction.rtl
    ? RtlScrollConverter.getScrollLeft(
        this.rootElement.current,
        this.state.direction
    )
    : -(
        RtlScrollConverter.getScrollLeft(
            this.rootElement.current,
            this.state.direction
        )
    );

```

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,570 @@
...
+        if (index === -1 || index >= this.itemPositions.length) {
+            return 0;
+        }
+
+        const itemPosition: ItemPosition = this.itemPositions[index];
+        const newScrollPosition: number = Math.max(
+            0,
+            Math.min(itemPosition.start, this.maxScroll)
+        );
+        return newScrollPosition;
```
janechu, line 564: Looks like you can just return:
```ts
return Math.max(
    0,
    Math.min(itemPosition.start, this.maxScroll)
);
```

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
```diff
@@ -0,0 +1,570 @@
...
+
+        let renderEndIndex: number = visibleRangeEndIndex + bufferLength;
+        if (renderEndIndex > lastIndex) {
+            renderEndIndex = lastIndex;
+        }
+
+        window.clearTimeout(this.scrollLayoutUpdateTimer);
+
+        this.setState({
+            renderedRangeStartIndex: renderStartIndex,
```
janechu, line 383: nit: suggest you eliminate the `renderStartIndex` variable and do something like:

```ts
(function(test: number, boundary: number) { return test > boundary ? test : boundary })(visibleRangeStartIndex - bufferLength, 0)
```

perhaps doing something similar for `renderEndIndex`.

packages/react/fast-components-react-base/src/index.ts
chrisdholt: Agree with this - we can still use the control internally without exporting it but if it goes into the changelog as though it was "added" it may confuse folks thinking it's available for export. Likely best to shift this to a chore.

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: yep

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: yep

packages/react/fast-components-react-base/src/stack-panel/stack-panel.tsx
scomea: Better:
 this.setState({
            renderedRangeStartIndex: Math.max(visibleRangeStartIndex - bufferLength, 0),
            renderedRangeEndIndex: Math.min(visibleRangeEndIndex + bufferLength, lastIndex),
            isScrollable: this.itemContainerSpan > this.viewportSpan,
        });

PR #3051, opened 2020-05-01
fix: better design-system-provider resolution

nicholasrice:
# Description
Adds a mechanism to register a tag-name as a design-system-provider so that design-system-provider resolution is configurable.

Changing the function to use the `DesignSystemProvider` reference caused a circular reference, so I also needed to re-organize the entities. 

closes #2992
## Issue type checklist

- [ ] **Chore**: A change that does not impact distributed packages.
- [x] **Bug fix**: A change that fixes an issue, link to the issue above.
- [ ] **New feature**: A change that adds functionality.

**Is this a breaking change?**
- [ ] This change causes current functionality to break.

## Process & policy checklist

- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/en/contributing/standards) for this project.

PR #3061, opened 2020-05-05
chore: add Design System Provider documentation

nicholasrice:
# Description
Adds documentation for the Design System Provider custom element.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,279 @@
+# Design System Provider
+
+## What is a Design System
+
+The _design system_ can generally be thought of as a collection named values that inform visual rendering. It quantifies concepts such as type-ramp, color palettes, design units, etc to be used in UI components. Other common names for this concept are _design variables_, _design tokens_ or _theme_.
```
EisenbergEffect, line 5: should be "a collection of" (add of)

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+These values are mutable throughout a UI tree. Where UI tree _A_ may see their contextual _design-unit_ as `4px`, section _B_ may have it's scale increased by providing a _design-unit_ of `6px`. Or, section _C_ may see a contextual _background color_ of `#FFF` where section _D_ may see a contextual _background color_ as `#000`.
```
EisenbergEffect, line 7: should be "its scale increased" (remove apostrophe)

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Design System Provider
+
+A _design system_ isn't much use if without convenient mechanisms to surface the _design system_ values to UI components and change values where desired. This is where the _Design System Provider_ comes in. `@microsoft/fast-components` exports the `FASTDesignSystemProvider` and `DesignSystemProvider` Web Components to help with:
```
EisenbergEffect, line 11: should be "much use without" (remove if)

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,279 @@
...
+```ts
+DesignSystemProvider.registerCSSCustomProperty({
+    name: "design-unit-12th",
+    value: designSystem => Math.pow(designSystem.designUnit, 12),
+});
+```
+
+In the above example, the value function will be re-evaluated if the _design system_ ever changes.
+
+The above API is made especially useful when defining a [component stylesheet](https://github.com/microsoft/fast-dna/blob/master/packages/web-components/fast-element/docs/building-components.md#defining-css) - individual stylesheets can declare dependencies on CSS custom properties that are functions of he element instance's *design system*
```
EisenbergEffect, line 70: should be "functions of the element" (missing "t" on the)

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+```ts
+// No CSS custom property will be created
+@designSystemProperty({default: "foo", cssCustomProperty: false})
+public fooBar;
+```
+
+The default value _must_ be set through the `default` property on the supplied config. This is the value that will be applied when the `use-defaults` attribute is used and no override property is provided.
+
+### FAST Design System Properties
```
EisenbergEffect, line 181: I wonder if we should turn this into a table so we can see the css property next to the provider property. And if there's not a correspondence, maybe specify that it comes from a recipe?

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,279 @@
...
+`.withBehaviors(
+    cssCustomPropertyBehaviorFactory(
+        "fancy-height",
+        designSystem => Math.pow(designSystem.designUnit, 12),
+        DesignSystemProvider.findProvider
+    )
+)
+```
+
+### FAST Design System Provider
```
EisenbergEffect, line 90: I wonder if we should move this up earlier in the doc so that it comes before any customization sections. It felt a bit out of place reading this after reading about creating recipes.

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice: My plan was to just link to the API docs when we got them but I can convert to a table for now.

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice: Yea I debated where to put it. I can move it up. My thought on having it lower is it didn't bury the other section after a long list of properties but contextually it belongs in the above section.

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,251 @@
...
+@designSystemProperty({default: "foo", cssCustomProperty: false})
+public fooBar;
+```
+
+The default value _must_ be set through the `default` property on the supplied config. This is the value that will be applied when the `use-defaults` attribute is used and no override property is provided.
+
+### FAST Design System Properties
+|Property Name|Type|Attribute Name|CSS Custom property|
+|---|---|---|---|
+|backgroundColor|string| background-color,  | background-color |
```
radium-v, line 184: comma

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,251 @@
...
+|accentBaseColor|string| accent-base-color | N/A |
```
radium-v, line 185: Should the `accentBaseColor` be provided as a CSS property?

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,251 @@
+# Design System Provider
```
janechu, line 1: a small table of contents at the start may be helpful for traversing this document.

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,251 @@
...
+```
+
+In the above example, the value function will be re-evaluated if the _design system_ ever changes.
+
+The above API is made especially useful when defining a [component stylesheet](https://github.com/microsoft/fast-dna/blob/master/packages/web-components/fast-element/docs/building-components.md#defining-css) - individual stylesheets can declare dependencies on CSS custom properties that are functions of the element instance's *design system*
+
+**EXAMPLE: creating a recipe dependency**
+```ts
+import { css } from "@microsoft/fast-element";
+import { cssCustomPropertyBehaviorFactory, DesignSystemProvider } from "@microsoft/fast-components";
```
radium-v, line 101: I'm not able to get `cssCustomPropertyBehaviorFactory` to pull in from the package (I'm testing locally by using `yarn link` in conjunction with #3051):

```
TS2305: Module '"../../../../../../packages/web-components/fast-components/dist"' has no exported member 'cssCustomPropertyBehaviorFactory'.
```

nicholasrice: Hmmm it might not be exported actually - good find.

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice: It generally should not be used directly and doesn't *currently* exist as a CSS custom property.

The reason for that is because the *base color* derives the accent palette, and UI colors are derived using recipes *from* that palette. There are cases where the color resolved by the recipe may match the `accentBaseColor` but also cases where it would not - resulting in contrast failures.

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice, line 184:
```suggestion
|backgroundColor|string| background-color | background-color |
```

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,251 @@
+# Design System Provider
+
+## What is a Design System
+
+The _design system_ can generally be thought of as a collection of named values that inform visual rendering. It quantifies concepts such as type-ramp, color palettes, design units, etc to be used in UI components. Other common names for this concept are _design variables_, _design tokens_ or _theme_.
+
+These values are mutable throughout a UI tree. Where UI tree _A_ may see their contextual _design-unit_ as `4px`, section _B_ may have its scale increased by providing a _design-unit_ of `6px`. Or, section _C_ may see a contextual _background color_ of `#FFF` where section _D_ may see a contextual _background color_ as `#000`.
```
marjonlynch, line 7: Wondering if UI tree A should be section A or all the others should be subtree B, C, D, or Tree B, C, D?

packages/web-components/fast-components/src/design-system-provider/README.md
radium-v: This should be resolved now by #3085

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice: yea they should have terminology aligned - I'll change them all to "tree"

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,260 @@
+# Design System Provider
+- [What is a Design System](#what-is-a-design-system)
```
EisenbergEffect, line 2: I'm not sure that we should put an inline ToC as that will make the final Docusaurus generated docs a bit odd. Docusaurus will generate the in-page ToC as a right sidebar.

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+const styles = css`
+    :host {
+        height: var(--fancy-height);
+    }
+`.withBehaviors(
+    cssCustomPropertyBehaviorFactory(
+        "fancy-height",
+        designSystem => Math.pow(designSystem.designUnit, 12),
+        DesignSystemProvider.findProvider
```
EisenbergEffect, line 120: Not a critique on the documentation, but would it be possible to make `DesignSystemProvider.findProvider` the default value for this param? It seems it has to be provided manually everywhere.

packages/web-components/fast-components/src/design-system-provider/README.md
```diff
@@ -0,0 +1,260 @@
...
+Building off the above, _design system_ properties can be declared using the `@designSystemProperty` decorator.
+
+**EXAMPLE: Creating a design system property declaration**
+
+```ts
+// ..
+@designSystemProvider("fancy-design-system-provider")
+class FancyDesignSystemProvider extends DesignSystemProvider {
+    @attr("fancy-property")
+    @designSystemProperty({ cssCustomProperty: "fancy-property", default: "red" })
```
EisenbergEffect, line 155: Again, not a critique of the docs, but it would be cool if the `designSystemProperty` could automatically pick up the `attr` name so that it didn't need to be provided (unless is was different).

packages/web-components/fast-components/src/design-system-provider/README.md
janechu: That's true, I forgot about docusaurus when I made the comment requesting it. Should we come to a consensus about how to deal with this? As of right now I have TOC on `fast-tooling` and `fast-tooling-react` packages because looking at the README by itself (on npm or in github) without it is cumbersome.

nicholasrice: okay - I'll remove.

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice: We could probably make a factory fn to supply that arg. Let me do some thinking on how best we could do this.

packages/web-components/fast-components/src/design-system-provider/README.md
nicholasrice: Good idea - I'll write up a proposal for that.

PR #3062, opened 2020-05-05
chore: load all the fast-components

radium-v:
# Description

Load fast-components for consumption on the website project.

## Motivation & context

This shoehorns all WCs available, we should probably reduce this from loading all to just the ones we need in the future.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/index.ts
+++ b/sites/fast-website/src/app/index.ts
@@ -1,1 +1,2 @@
+export * from "./components";
 export * from "@microsoft/fast-components";
--- a/sites/fast-website/src/app/index.ts
+++ b/sites/fast-website/src/app/index.ts
@@ -1,0 +1,1 @@
+export * from "@microsoft/fast-components";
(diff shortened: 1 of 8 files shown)


sites/fast-website/src/public/index.html
```diff
@@ -1,13 +1,13 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
nicholasrice, line 2: can you also add `dir="ltr"`? I think we have a few components that look for that

sites/fast-website/src/app/index.ts
```diff
@@ -0,0 +1 @@
+export * from "@microsoft/fast-components";
```
nicholasrice, line 1: Does this get tree shaken if webpack has the production mode set?

radium-v: I tested a prod build (see a450e0b) and confirm that webpack doesn't drop the exports and the components on the page load and work as expected.

nicholasrice: Awesome, thanks!

PR #3067, opened 2020-05-06
feat: add accordion web component

eljefe223:
# Description
Adds `fast-accordion` as new web component

image

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,141 @@
...
+        grid-column: 4;
+        z-index: 2;
+        pointer-events: none;
+    }
+
+    slot[name="collapsed-glyph"] {
+        display: flex;
+    }
+
+    :host(.expanded) slot[name="collapsed-glyph"] {
```
EisenbergEffect, line 100: I did not know we could target the slot itself in this way. Cool!

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,132 @@
...
+    public accordionItems: HTMLElement[];
+    public accordionItemsChanged(): void {
+        if (this.$fastController.isConnected) {
+            this.accordionIds = this.getHeaderIds();
+            this.setItems();
+        }
+    }
+
+    @observable
+    public activeIndicatorRef: HTMLElement;
```
EisenbergEffect, line 32: FYI It's not a problem if it's there, but you shouldn't need `@observable` on a basic ref.

packages/web-components/fast-components/src/patterns/start-end.ts
```diff
@@ -18,7 +18,7 @@ export class StartEnd {
 }
 
 export const endTemplate = html<StartEnd>`
-    <span name="end" part="end" ${ref("endContainer")}>
+    <span class="end" name="end" part="end" ${ref("endContainer")}>
```
EisenbergEffect, line 21: What's the point of the `name` attribute on the span in this case? I'm inclined to think that was supposed to be "class" begin with. Anyone know?

packages/web-components/fast-components/src/patterns/start-end.ts
```diff
@@ -28,7 +28,7 @@ export const endTemplate = html<StartEnd>`
 `;
 
 export const startTemplate = html<StartEnd>`
-    <span name="start" part="start" ${ref("startContainer")}>
+    <span class="start" name="start" part="start" ${ref("startContainer")}>
```
EisenbergEffect, line 31: Same as above.

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,132 @@
...
+    private accordionIds: Array<string | null>;
+
+    private change = (): void => {
+        this.$emit("change", this.activeid);
+    };
+
+    private setItems = (): void => {
+        this.accordionIds = this.getHeaderIds();
+        this.accordionItems.forEach((header: HTMLElement, index: number) => {
+            if (header instanceof FASTAccordionItem) {
```
EisenbergEffect, line 45: @nicholasrice Did we determine that this `instanceof` check can fail, even if the type is imported (and registered) first?

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,132 @@
...
+                header.addEventListener("change", this.activeHeaderChange);
```
EisenbergEffect, line 46: I think we'll need a corresponding `removeEventListener` call somewhere. If the header is removed and then added to another expander, then I think this would break.

packages/web-components/fast-components/src/accordion/accordion.ts
eljefe223: ugh this should not be here at all

packages/web-components/fast-components/src/accordion/accordion.ts
eljefe223: @nicholasrice and I chatted about this, I'll let him chime in. But think we should be good here.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    .button {
+        appearance: none;
+        border: none;
+        background: none;
+        grid-column: 2;
+        grid-row: 1;
+        outline: none;
+        padding: 0 calc((6 + (var(--design-unit) * 2 * var(--density))) * 1px);
```
radium-v, line 46: Would it make sense to hoist this repeated `calc` value to be its own css variable?

eljefe223: Yeah its used throughout our components, should be added to the style utilities, I would leave that for another PR though.

packages/web-components/fast-components/src/accordion/README.md
```diff
@@ -0,0 +1,4 @@
+# fast-accordion
+`fast-accordion` is a web component implementation of an [Accordion](https://w3c.github.io/aria-practices/#accordion).
+
+For more information view the [component specification](./accordion.spec.md).
```
chrisdholt, line 4: Can you move the file here and add a checkmark to the spec roadmap. You'll likely need to update the pathing in the spec readme as well.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    :host(.expanded) .region {
+        display: flex;
+    }
+
+    .heading {
+        z-index: 2;
+    }
+
+    .glyph {
```
chrisdholt, line 87: I think we might want to change all references to glyph to be "icon": https://github.com/WICG/open-ui/issues/27#issuecomment-602735148

I believe glyph is typically referring to a specific icon.

packages/web-components/fast-components/src/accordion/README.md
eljefe223: Already moved it just didn't update the spec roadmap

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
```diff
@@ -0,0 +1,45 @@
...
+            >
+                <span class="heading">
+                    <slot name="heading" part="heading"></slot>
+                </span>
+            </button>
+            ${startTemplate}
+            ${endTemplate}
+            <span class="glyph" part="glyph">
+                <slot name="expanded-glyph" part="expanded-glyph"></slot>
+                <slot name="collapsed-glyph" part="collapsed-glyph"></slot>
```
chrisdholt, line 33: Clever! Assuming this is toggled based on state. Do we think this is preferable as an implementor to dynamically changing/passing a different icon based on state?

packages/web-components/fast-components/src/patterns/start-end.ts
eljefe223: Yup I think your right, `name` is not necessary here, I noticed that but was reluctant to remove it. I'll update

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,34 @@
...
+    h5 = "5",
+    h6 = "6",
+}
+
+export class AccordionItem extends FASTElement {
+    @attr
+    public headinglevel: AccordionHeadingLevel = AccordionHeadingLevel.h2;
+
+    @attr({ mode: "boolean" })
+    public expanded: boolean = false;
```
chrisdholt, line 17: Do we want this to reflect to the view if not passed by the user or do we just want the view to show `aria-expanded`?

packages/web-components/fast-components/src/accordion/accordion.styles.ts
```diff
@@ -0,0 +1,20 @@
...
+    neutralForegroundRestBehavior,
+} from "../styles/recipes";
+
+export const AccordionStyles = css`
+    ${display("flex")} :host {
+        box-sizing: border-box;
+        flex-direction: column;
+        font-family: var(--body-font);
+        ${/* Font size, weight, and line height are temporary - 
+            replace when adaptive typography is figured out */ ""} font-size: 12px;
```
chrisdholt, line 14: Can you map this to the latest typography values in the design system?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export const AccordionItemStyles = css`
+    ${display("flex")} :host {
+        box-sizing: border-box;
+        font-family: var(--body-font);
+        flex-direction: column;
+        ${
+            /* Font size, weight, and line height are temporary - 
+            replace when adaptive typography is figured out */ ""
+        } font-size: 12px;
```
chrisdholt, line 22: Can we update this to use typography which is in now?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,141 @@
...
+        font-weight: 400;
+        line-height: 18px;
```
chrisdholt, line 24: Same note on using the typography value

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,129 @@
...
+    wrapInBounds,
+} from "@microsoft/fast-web-utilities";
+import { StartEnd } from "../patterns/start-end";
+import { applyMixins } from "../utilities";
+import { FASTAccordionItem } from "./accordion-item";
+
+export enum AccordionExpandMode {
+    single = "single",
+    multi = "multi",
+}
```
chrisdholt, line 16: Do we know what these are called in other design systems?

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export class Accordion extends FASTElement {
+    @attr
+    public expandmode: AccordionExpandMode = AccordionExpandMode.multi;
```
chrisdholt, line 20: It might keep things clean to have this be fromView only.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
eljefe223: its hidden and shown via css based on `:host(.expanded)`. Some cases require different Icons + or - for example some may just rotate a chevron > this allows for both since everything can be managed via css.

chrisdholt: sure - I understand the implementation here. My primary ask is whether we think this is preferred from an API standpoint vs. a single slot that changes based on state. For context, this is somewhat driving the question: https://github.com/WICG/open-ui/pull/54#discussion_r411660381

We don't need to land as part of this, but we should keep an eye on how things like these land in Open-UI and align where we can.

eljefe223: Curious how would a single slot work like the above, how would an author pass different icons?

chrisdholt: If I'm using SVG's I could likely conditionally render based on the expanded state of the component. If I want to use a CSS solution I can do that as well with that approach. If I want to use icon fonts I can swap a class name based on a single span sent into the slot.

Again, not prescribing that approach, just something to think about and keep an eye on (discussion above). I think this is a clever approach regardless of how that discussion pans out :)

eljefe223: cool I'll resolve this for now then.

packages/web-components/fast-components/src/accordion/accordion.ts
eljefe223: Some don't even offer the option, One system call the control `collapse` which allows multiple by default and if you set `accordion` too `true it only allows one. A few refer to it as multiple vs single selection

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
eljefe223: I tried using fromView mode but this was not changing `aria-expanded` on the accordion item. Not sure how we want to do here, the `aria-expanded` attr needs to update.

packages/web-components/fast-components/src/accordion/accordion.ts
nicholasrice: We determined that import-order matters in cases where DOM exists prior to script parsing / execution. In those cases, the elements are upgraded immediately upon definition. That's why we were seeing child components of the DS provider upgrading before the DS provider when the DS provider was imported and referenced *after* the child components. I *think* that case is covered here because the item is an explicit dependency of the accordion.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,34 @@
...
+    public button: HTMLElement;
+
+    /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+    public clickHandler = (e: MouseEvent) => {
+        this.expanded = !this.expanded;
+        this.change();
+    };
+
+    private change = (): void => {
+        this.$emit("change", this.id);
```
nicholasrice, line 32: Does this need to pass `this.id` as data? The event will have the element accessible from the target property of the event so this seems redundant.

packages/web-components/fast-components/src/accordion/accordion.ts
nicholasrice: FWIW aria uses "multi" in the `aria-multiselectable` attribute

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,137 @@
...
+            this.setItems();
+        }
+    }
+
+    private activeid: string;
+    private activeHeaderIndex: number = 0;
+    private accordionIds: Array<string | null>;
+
+    private change = (): void => {
+        this.$emit("change", this.activeid);
```
nicholasrice, line 37: same comment on the above here - this data is accessible from the event target so maybe it's unnecessary?

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,137 @@
...
+            if (header instanceof FASTAccordionItem) {
+                header.addEventListener("change", this.activeHeaderChange);
+                if (this.isSingleExpandMode()) {
+                    this.activeHeaderIndex !== index
+                        ? (header.expanded = false)
+                        : (header.expanded = true);
+                }
+            }
+            const headerId: string | null = this.accordionIds[index];
+            header.setAttribute(
```
nicholasrice, line 52: Does this `header` element exist in the shadow DOM or light DOM?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,138 @@
...
+import { forcedColorsStylesheetBehavior } from "../../styles/match-media-stylesheet-behavior";
+import { SystemColors } from "../../styles/system-colors";
+
+export const AccordionItemStyles = css`
+    ${display("flex")} :host {
+        box-sizing: border-box;
+        font-family: var(--body-font);
+        flex-direction: column;
+        font-size: var(--type-ramp-minus-1-font-size);
+        font-weight: 400;
```
nicholasrice, line 20: Do you need to set font-weight?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
```diff
@@ -0,0 +1,138 @@
...
+        border: calc(var(--outline-width) * 1px) solid var(--neutral-focus);
+        box-shadow: 0 0 0 calc((var(--focus-outline-width) - var(--outline-width)) * 1px)
+            var(--neutral-focus);
+    }
+
+    :host(.expanded) .region {
+        display: flex;
+    }
+
+    .heading {
```
nicholasrice, line 80: Can this be combined with the above `.heading` rule-set?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
```diff
@@ -0,0 +1,46 @@
...
+                aria-expanded="${x => x.expanded}"
+                aria-controls="${x => x.id}-panel"
+                id="${x => x.id}"
+                @click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
+            >
+                <span class="heading">
+                    <slot name="heading" part="heading"></slot>
+                </span>
+            </button>
+            ${startTemplate}
```
nicholasrice, line 30: It seems a *little* strange to me that the "start" and "end" are:
1. right next to each-other
2. not surrounding the heading slot

Can you help me understand why these two slots exist *here* in the template?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,34 @@
+import { attr, FASTElement, observable } from "@microsoft/fast-element";
+
+export enum AccordionHeadingLevel {
```
nicholasrice, line 3: An enum seems like overkill here and a bit of a disconnect  because you're forced to use the "h" before the level to satisfy language requirements.

Thoughts on just typing this as `1 | 2 | 3 | 4 | 5 | 6`?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,34 @@
...
+    h2 = "2",
+    h3 = "3",
+    h4 = "4",
+    h5 = "5",
+    h6 = "6",
+}
+
+export class AccordionItem extends FASTElement {
+    @attr
+    public headinglevel: AccordionHeadingLevel = AccordionHeadingLevel.h2;
```
nicholasrice, line 14: I think based on previous discussions this should use an attribute of `heading-level` and a property of `headingLevel`

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    @attr({ mode: "boolean" })
+    public expanded: boolean = false;
+
+    @attr
+    public id: string;
+
+    @observable
```
nicholasrice, line 22: Can we come up with somthing more semantic than "button" - maybe something along the lines of "expandButton", or "expansionControl"?

packages/web-components/fast-components/src/accordion/accordion.ts
eljefe223: light dom

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.styles.ts
eljefe223: probably not, I think I was just following patterns elsewhere

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
eljefe223: They are laid out in a grid and with column definitions.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
eljefe223: Good to know, I was following a pattern I used elsewhere.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
eljefe223: Since it makes no difference I'll move the end template before the button, This was originally to put the elements above the button but I ended up using `z-index` anyway

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
eljefe223: An enum member cannot have a numeric name. So I added the h in front. Is there a way around this?

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
eljefe223: Actually I take that back, It's for focus order, since the content in `start` and `end` can be focusable we want the button to focus first, If we change the order the `start` content will focus first.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
eljefe223: NM, I understand the suggestion now ;)

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
eljefe223: I chatted with @EisenbergEffect sounds like we would need to do some additional work that might over complicate things to have `aria-expanded` update but not `expanded`

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,25 @@
+import { attr, booleanConverter, FASTElement, observable } from "@microsoft/fast-element";
+
+export class AccordionItem extends FASTElement {
+    @attr({ attribute: "heading-level", mode: "fromView", converter: booleanConverter })
```
nicholasrice, line 4: Should this be using `booleanConverter`?

chrisdholt: heading-level likely should not

eljefe223: since its defaulted, we only want heading level to display if the author assigns one.

chrisdholt: fromView should be good for this - we'd use both here only if this were a boolean value which needs converting

eljefe223: I think I was confused on how the mode `fromView` works thinking it needed that converter, Removed.

packages/web-components/fast-components/src/accordion/accordion-item/accordion-item.template.ts
```diff
@@ -0,0 +1,46 @@
+import { html, ref } from "@microsoft/fast-element";
+import { endTemplate, startTemplate } from "../../patterns/start-end";
+import { AccordionItem } from "./accordion-item";
+
+export const AccordionItemTemplate = html<AccordionItem>`
+    <template
+        class="${x => (x.expanded ? "expanded" : "")}"
+        aria-expanded="${x => x.expanded}"
```
chrisdholt, line 8: @eljefe223 Is this aria-expanded redundant? I don't see this called out here: https://w3c.github.io/aria-practices/#accordion

packages/web-components/fast-components/src/accordion/accordion.ts
```diff
@@ -0,0 +1,139 @@
...
+                if (this.isSingleExpandMode()) {
+                    this.activeItemIndex !== index
+                        ? (item.expanded = false)
+                        : (item.expanded = true);
+                }
+            }
+            const itemId: string | null = this.accordionIds[index];
+            item.setAttribute(
+                "id",
+                typeof itemId !== "string" ? `accordion-${index + 1}` : itemId
```
scomea, line 55: Can't "`accordion-${index + 1}`" result in duplicate id's if you have multiple accordions?

eljefe223: They are in their own shadow DOMs. Duplicate ids across different shadow DOMs is acceptable.

scomea: Really? did not not know that.  So Query selector by id on the document won't find elements within web components then?

packages/web-components/fast-foundation/src/accordion/accordion-item/accordion-item.ts
```diff
@@ -0,0 +1,40 @@
...
+    })
+    public headinglevel: 1 | 2 | 3 | 4 | 5 | 6 = 2;
+
+    @attr({ mode: "boolean" })
+    public expanded: boolean = false;
+
+    @attr
+    public id: string;
+
+    @observable
```
EisenbergEffect, line 24: No need to put an `@observable` decorator on a ref property. Basically, you only need `@observable` if the template needs to update when your code changes the property. In the case of ref, it's the template that's setting you (other way around).

packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -0,0 +1,131 @@
...
+            item.setAttribute(
+                "id",
+                typeof itemId !== "string" ? `accordion-${index + 1}` : itemId
+            );
+            this.activeid = this.accordionIds[this.activeItemIndex] as string;
+            item.addEventListener("keydown", this.handleItemKeyDown);
+        });
+    };
+
+    private resetItems = (): void => {
```
EisenbergEffect, line 59: Can this be a regular method?

PR #3068, opened 2020-05-06
chore: add section layout to the website

khamudom:
# Description
Currently, we just have a blank page, so I created some placeholder sections for the components, by adding, header, section, and footer.

## Issue type checklist

- [x] **Chore**: A change that does not impact distributed packages.
- [ ] **Bug fix**: A change that fixes an issue, link to the issue above.
- [ ] **New feature**: A change that adds functionality.

**Is this a breaking change?**
- [ ] This change causes current functionality to break.

## Process & policy checklist

- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/en/contributing/standards) for this project.

PR #3070, opened 2020-05-06
feat: create section-header component for marketing site

khamudom:
# Description
This component is used in the hero section and in every other section down the page.

image

The templates consist of one default slot and three named slots (`badge`, `body` and `action`)

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -4,7 +4,7 @@ import { SectionHeader } from "./section-header";
 export const SectionHeaderTemplate = html<SectionHeader>`
     <template>
         <slot></slot>
-        <slot name="subheading"></slot>
+        <slot name="badge"></slot>
         <slot name="body"></slot>
         <slot name="action"></slot>
     </template>
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -3,8 +3,8 @@ import { SectionHeader } from "./section-header";
 
 export const SectionHeaderTemplate = html<SectionHeader>`
     <template>
-        <slot></slot>
         <slot name="badge"></slot>
+        <slot></slot>
         <slot name="body"></slot>
         <slot name="action"></slot>
     </template>
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -4,8 +4,8 @@ import { SectionHeader } from "./section-header";
 export const SectionHeaderTemplate = html<SectionHeader>`
     <template>
         <slot name="heading"></slot>
-        <h1 part="short">${x => x.short}</h1>
-        <p part="body" class="body">${x => x.body}</p>
+        <p part="short">${x => x.short}</p>
+        <p part="body">${x => x.body}</p>
         <slot name="button"></slot>
     </template>
 `;
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -1,0 +1,11 @@
+import { html } from "@microsoft/fast-element";
+import { SectionHeader } from "./section-header";
+
+export const SectionHeaderTemplate = html<SectionHeader>`
+    <div>
+        <h5 part="h5">${x => x.title}</h5>
+        <h1>${x => x.subTitle}</h1>
+        <p>${x => x.description}</p>
+        <slot name="button"></slot>
+    </div>
+`;
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -3,9 +3,9 @@ import { SectionHeader } from "./section-header";
 
 export const SectionHeaderTemplate = html<SectionHeader>`
     <div>
-        <h5 part="h5">${x => x.title}</h5>
-        <h1>${x => x.subTitle}</h1>
-        <p>${x => x.description}</p>
+        <h5 part="title">${x => x.title}</h5>
+        <h1 part="sub-title">${x => x.subTitle}</h1>
+        <p part="description">${x => x.description}</p>
         <slot name="button"></slot>
     </div>
 `;
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -3,9 +3,9 @@ import { SectionHeader } from "./section-header";
 
 export const SectionHeaderTemplate = html<SectionHeader>`
     <template>
-        <slot name="heading"></slot>
-        <p part="short">${x => x.short}</p>
-        <p part="body">${x => x.body}</p>
-        <slot name="button"></slot>
+        <slot></slot>
+        <slot name="subheading"></slot>
+        <slot name="body"></slot>
+        <slot name="action"></slot>
     </template>
 `;
--- a/sites/fast-website/src/app/components/section-header/section-header.template.ts
+++ b/sites/fast-website/src/app/components/section-header/section-header.template.ts
@@ -2,10 +2,10 @@ import { html } from "@microsoft/fast-element";
 import { SectionHeader } from "./section-header";
 
 export const SectionHeaderTemplate = html<SectionHeader>`
-    <div>
-        <h5 part="title">${x => x.title}</h5>
-        <h1 part="sub-title">${x => x.subTitle}</h1>
-        <p part="description">${x => x.description}</p>
+    <template>
+        <slot name="heading"></slot>
+        <h1 part="short">${x => x.short}</h1>
+        <p part="body" class="body">${x => x.body}</p>
         <slot name="button"></slot>
-    </div>
+    </template>
 `;
--- a/sites/fast-website/src/public/index.html
+++ b/sites/fast-website/src/public/index.html
@@ -108,16 +108,5 @@
         <!-- footer:end -->
       </div>
     </fast-design-system-provider>
-
-    <style>
-      fast-section-header h5 {
-        margin: 0;
-      }
-      fast-section-header::part(short) {
-        font-size: 46px;
-        line-height: 50px;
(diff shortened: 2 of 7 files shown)


sites/fast-website/src/public/index.html
```diff
@@ -9,6 +9,24 @@
...
+    <style>
+      fast-section-header h5 {
+        margin: 0;
+      }
+      fast-section-header::part(short) {
+        font-size: 46px;
+        line-height: 50px;
+        margin: 0;
+      }
+    </style>
```
chrisdholt, line 30: @radium-v @eljefe223 This is likely going to get fairly convoluted fairly quickly. Would it make sense to add storybook as a dev dependency so we can develop these in isolation more efficiently? It seems like some of these controls may have different states or variations and that might be difficult to keep track of in a monolithic html file over time. Just a thought.

sites/fast-website/src/app/components/section-header/section-header.template.ts
```diff
@@ -0,0 +1,11 @@
+import { html } from "@microsoft/fast-element";
+import { SectionHeader } from "./section-header";
+
+export const SectionHeaderTemplate = html<SectionHeader>`
+    <template>
+        <slot name="heading"></slot>
+        <p part="short">${x => x.short}</p>
```
chrisdholt, line 7: What is "short"?

sites/fast-website/src/app/components/section-header/section-header.template.ts
```diff
@@ -0,0 +1,11 @@
...
+        <p part="body">${x => x.body}</p>
```
chrisdholt, line 8: How is this different than "short"? Should this be a slot?

sites/fast-website/src/app/components/section-header/section-header.template.ts
```diff
@@ -0,0 +1,11 @@
+import { html } from "@microsoft/fast-element";
+import { SectionHeader } from "./section-header";
+
+export const SectionHeaderTemplate = html<SectionHeader>`
+    <template>
```
chrisdholt, line 5: Should this control have a default slot?

sites/fast-website/src/app/components/section-header/section-header.template.ts
khamudom: "short" is the attribute name @bheston had suggested. Another attribute we can also use  "summary" or "sub-header", if short is confusing. It is a new word for me too.

chrisdholt: I don't see short as being intuitive for this. It doesn't appear to be a sub header either. This seems to be kind-of "badging" the section, though it doesn't map directly to the badge guidance. It appears to be a 1:1 from a visual standpoint from what I can see.

sites/fast-website/src/app/components/section-header/section-header.template.ts
khamudom: The "body" is the "description" on this particular component. I am ok with making it a slot. This element looks to be the same whether used as a hero section or a page section, that was the reason I didn't initially go with a slot.

chrisdholt, line 8: Are we going to add styles for the body here? We'll want to add and target a class if so.

sites/fast-website/src/app/components/section-header/section-header.template.ts
khamudom: Since this text changes in size base where it is used, making it a slot does make sense.

chrisdholt: Sorry, for clarity - every slot here is named. Is there a default slot for the control? I might make the heading the default slot for the control as I could foresee adding sections that do not have body copy.

sites/fast-website/src/public/index.html
radium-v: There's an added stylesheet in #3068 and I plan on adding support for CSS splitting to the webpack build, so these style chunks can sit with their respective components.

chrisdholt: Should each component utilize the existing patterns we have in `fast-components`? We've built-in some fairly good performance optimizations there and we can leverage the actual typography as well. Super simple example: https://github.com/microsoft/fast-dna/tree/master/packages/web-components/fast-components/src/card

sites/fast-website/src/public/index.html
```diff
@@ -9,6 +9,24 @@
     <fast-design-system-provider use-defaults>
       <fast-button>Hello, World!</fast-button>
       <gradient-picker>stub</gradient-picker>
+      <fast-section-header
+        short="Features & Utilities"
+        body="At id vitae iaculis volutpat laoreet. Massa ut scelerisque ut ac, in aliquam odio. Semper sagittis blandit in sed turpis lorem at diam iaculis."
+      >
+        <h5 slot="heading">A COMPREHENSIVE TOOLKIT</h5>
```
radium-v, line 16: All-uppercase text should be Title Case in the markup and use `text-transform: uppercase`

sites/fast-website/src/app/components/section-header/section-header.template.ts
khamudom: That makes sense. I will add a default slot

sites/fast-website/src/app/components/section-header/section-header.template.ts
radium-v: I'd call this a `subheading`

sites/fast-website/src/public/index.html
chrisdholt: I might go further and suggest not using text transform in CSS at all due to the potential issues with localization in certain languages. Our principle has typically been that we don't transform text and we rely on the string itself.

sites/fast-website/src/app/components/section-header/section-header.template.ts
chrisdholt: My primary thought is that this appears to be a direct use of our Badge component. We should consider a slot here if that's the case. It may not be super intuitive since this visually precedes the heading. While that's solvable with CSS, I do wonder if it's intuitive to a user that want's to target that.

sites/fast-website/src/app/components/section-header/section-header.template.ts
chrisdholt: Not prescriptive here, just asking the questions :)

sites/fast-website/src/app/components/section-header/section-header.template.ts
EisenbergEffect: I think a named slot of "subheading" would be better than an attribute as it would allow the semantic content to remain in the light dom. You can then use the `::slotted` selector to target the slot and provide the styling you want.

sites/fast-website/src/app/components/section-header/section-header.template.ts
EisenbergEffect: Also, should be a slot for the same reason as above if possible.

sites/fast-website/src/app/components/section-header/section-header.template.ts
khamudom: yeah, I've updated to remove all attributes and will just have name slots and one default slot.

sites/fast-website/src/public/index.html
eljefe223: They definitely should follow the existing patterns.

sites/fast-website/src/public/index.html
```diff
@@ -76,7 +94,10 @@
 
         <section class="section">
           <div class="section-content">
-            Page section header
+            <fast-section-header>
+              <h5>LEARN, MODIFY AND FOLLOW</h5>
+              <h1 slot="subheading">Join the community</h1>
```
chrisdholt, line 99: These are definitely not h1's as an FYI. We'll want to make sure that the source is semantic and follows a logical outline.

chrisdholt: I think these are actually going to be something like a "badge"...I think.

sites/fast-website/src/public/index.html
```diff
@@ -25,14 +25,24 @@
...
           </div>
         </section>
         <!-- hero-section:end -->
 
         <section class="section">
           <div class="section-content">
-            Page section header
+            <fast-section-header>
+              <h5>FLEXIBLE AND ADAPTABLE</h5>
+              <h1 slot="badge">One component, many faces</h1>
```
chrisdholt, line 42: Can you pass the badge component here - or change this from being an h1?

eljefe223:
Don't want to drag this PR out but I think the slot naming is confusing I propose this:
image

chrisdholt:
> Don't want to drag this PR out but I think the slot naming is confusing I propose this:
> image

@eljefe223 is the proposal for everything to have a slot?

eljefe223:
> > Don't want to drag this PR out but I think the slot naming is confusing I propose this:
> > image
> 
> @eljefe223 is the proposal for everything to have a slot?

That's how it is currently. But the slots name don't match what I would expect them to do. Specifically the default slot being the top smaller heading and the badge slot being the large heading.

chrisdholt:
> > > Don't want to drag this PR out but I think the slot naming is confusing I propose this:
> > > image
> > 
> > 
> > @eljefe223 is the proposal for everything to have a slot?
> 
> That's how it is currently. But the slots name don't match what I would expect them to do. Specifically the default slot being the top smaller heading and the badge slot being the large heading.

Yeah in order I would say taking the visual:

named slots:
- badge
- body
- action

Default would be for Heading.

eljefe223:
> > > > Don't want to drag this PR out but I think the slot naming is confusing I propose this:
> > > > image
> > > 
> > > 
> > > @eljefe223 is the proposal for everything to have a slot?
> > 
> > 
> > That's how it is currently. But the slots name don't match what I would expect them to do. Specifically the default slot being the top smaller heading and the badge slot being the large heading.
> 
> My thought is that we have the following named slots:
> 
> * badge
> * body
> * action
> 
> Default slot would be for Heading.

This also works

sites/fast-website/src/public/index.html
khamudom: updated to be use badge

sites/fast-website/src/public/index.html
khamudom: updated

PR #3081, opened 2020-05-07
chore: add fast-website to the yarn and lerna workspaces

radium-v:
# Description

This PR adds the new `fast-website` and its `devDependencies` to the Yarn and Lerna workspaces.

## Motivation & context
Since we're temporarily using ultra-new additions from within the repo, we want to make sure our node_modules are pointing to the local neighbor instead of downloading the package independently.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,7 +3,8 @@
     "packages/**/*",
     "sites/fast-color-explorer",
     "sites/fast-component-explorer",
-    "sites/fast-creator"
+    "sites/fast-creator",
+    "sites/fast-website"
   ],
   "version": "independent",
   "npmClient": "yarn",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "license": "MIT",
   "private": true,
   "workspaces": {
-    "packages": ["packages/**/*", "sites/fast-color-explorer", "sites/fast-component-explorer", "sites/fast-creator", "sites/site-utilities"],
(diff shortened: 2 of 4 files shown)


radium-v:
I wasn't sure if commits from master should be rebased or done with a merge commit. I can force-push to redo it if a merge is desired over a rebase.

EisenbergEffect:
I think we're going to squash merge these in the end, so it may not matter as much. Anyone have opinions on this? /cc @nicholasrice @chrisdholt

janechu:
I've been maintaining my own feature branches and just rebasing master and force pushing them every once in a while to keep the history clean, then when the feature branch gets merged i temporarily enable a rebase with master to preserve commit history. This gets a little murkier when multiple contributors are on a branch so I'm not necessarily advocating for that here. However we shouldn't have these commits in the PR, it makes reviewing difficult. I do think that preserving commit history would be nice.

chrisdholt:
I agree with Jane's comments here. I think we'll run into trouble if with additional commits here. If we don't care too much about preserving history on the way back to master, we can leave it as is and squash, but I think it's valuable to have folks' commit history maintained.

radium-v:
Thanks @janechu @EisenbergEffect and @chrisdholt for the input. I'll remove the commits from this PR and we can discuss update strategy elsewhere.

Issue #3069, opened 2020-05-06
[Feature proposal] Add `orientation` attr to radio group component


chrisdholt:
This issue proposes adding an attribute of `orientation` to the API of `radio-group` to support common scenarios of having groups of radios which align horizontally as well as vertically.

I don't believe the list here is exhaustive, but below are a few design systems which expose an orientation for their radio group components:
- [Paste (Twilio)](https://paste.twilio.design/form-elements/radio-group/#horizontal-alignment)
- [Fluent UI](https://fluentsite.z22.web.core.windows.net/components/radio-group/definition?showCode=true&showRtl=false&showTransparent=false&showVariables=false#types-radio-group-vertical-example)
- [Andriod](https://developer.android.com/reference/android/widget/RadioGroup)
- [Veritas (Udacity)](https://veritas.udacity.design/#/Components/RadioGroup)

*NOTE: One of the above uses a boolean prop of `inline`.

## The case for adding an orientation attribute
**Ease of use**: Currently to change the orientation from vertically stacked to horizontally aligned a user would need to use their own CSS and target a part within the radio group component. I think this is OK and some of the design systems have examples which illustrate it. As a user, I could see how this might be convoluted or overly cumbersome compared to simply passing an attribute and having a CSS class added that takes care of this for me.

**Flexibility**: One of our principles for FAST is empowering developers through a flexible API and I believe this change supports that goal. While not exposing the attribute here, we are taking a prescriptive approach to the layout of this control. This leaves out some design systems and requires them to fend for themselves in implementing this. I would propose that adding this attr increases the flexibility of controls and allows all design systems to accomplish their goals in terms of "layout" within a radio group. If we do not include the attribute, teams need to extend our model to add their own support (and possibly our template if our suggested approach to adding classes is via the template). All that could be avoided by simply broadening the API to account for this.

**Why `orientation` as opposed to another name?**: I'm not opposed to using another name in place of orientation. Several other controls already include an `orientation` attribute, so there is precedence for the naming here. Additionally, orientation-like language is often used by the design systems which implement this behavior (horizontal/vertical). This would map 1:1 with other controls which have different visual orientations.

[...]

PR #3082, opened 2020-05-07
feat: add vertical and horizontal orientation for radio-group

marjonlynch:
# Description
In response to #3069 adding the orientation attribute to fast-radio-group to allow for vertical orientation and a default of horizontal.

## Motivation & context

Motivation described in proposal #3069.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -10,6 +10,15 @@
         </fast-radio-group>
     </div>
 
+    <div>
+        <fast-radio-group orientation="vertical" name="Bourbon">
+            <label slot="label">Bourbon</label>
+            <fast-radio value="elijahcraig">Elijah Craig</fast-radio>
+            <fast-radio value="woodford">Woodford Reserve</fast-radio>
+            <fast-radio value="elmerlee">Elmer T Lee</fast-radio>
+        </fast-radio-group>
+    </div>
+
     <div style="display: flex; flex-direction: column;margin-top: 12px;">
         <fast-radio-group name="players">
             <label slot="label">Best basketball players</label>
--- a/packages/web-components/fast-components/src/radio-group/fixtures/base.html
+++ b/packages/web-components/fast-components/src/radio-group/fixtures/base.html
@@ -11,11 +11,12 @@
     </div>
 
     <div>
-        <fast-radio-group orientation="vertical" name="Bourbon">
-            <label slot="label">Bourbon</label>
-            <fast-radio value="elijahcraig">Elijah Craig</fast-radio>
-            <fast-radio value="woodford">Woodford Reserve</fast-radio>
-            <fast-radio value="elmerlee">Elmer T Lee</fast-radio>
+        <fast-radio-group orientation="vertical" name="trees">
+            <label slot="label">Trees</label>
+            <fast-radio value="fir">Fir</fast-radio>
+            <fast-radio value="juniper">Juniper</fast-radio>
+            <fast-radio value="hemlock">Hemlock</fast-radio>
+            <fast-radio value="pine">Pine</fast-radio>
         </fast-radio-group>
     </div>
 
--- a/packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
@@ -7,9 +7,14 @@ export const RadioGroupStyles = css`
         margin: calc(var(--design-unit) * 1px) 0;
         flex-direction: column;
     }
-
     .positioning-region {
         display: flex;
         flex-wrap: wrap;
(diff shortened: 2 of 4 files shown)


packages/web-components/fast-components/src/radio-group/radio-group.ts
```diff
@@ -8,6 +8,11 @@ import {
 } from "@microsoft/fast-web-utilities";
 import { RadioControl } from "../radio";
 
+export enum RadioOrientation {
```
janechu, line 11: Don't we have some kind of orientation enum in web-utilities?

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
```diff
@@ -8,7 +8,13 @@ export const RadioGroupTemplate = html<RadioGroup>`
...
     >
         <slot name="label"></slot>
-        <div class="positioning-region" part="positioning-region">
+        <div
+            class="positioning-region ${x =>
+                x.orientation === RadioOrientation.horizontal
+                    ? "horizontal"
+                    : "vertical"}"
+            part="positioning-region"
+        >
```
chrisdholt, line 17: I think we'll want this class on the host element.

packages/web-components/fast-components/src/radio-group/fixtures/base.html
```diff
@@ -10,6 +10,15 @@ <h4>Defaults</h4>
         </fast-radio-group>
     </div>
 
+    <div>
+        <fast-radio-group orientation="vertical" name="Bourbon">
+            <label slot="label">Bourbon</label>
```
janechu, line 15: Not to be a literal buzz kill, but could we change this to a non-alcoholic drink or colors or something? :)

packages/web-components/fast-components/src/radio-group/radio-group.ts
EisenbergEffect: It would be nice to have a single Orientation used everywhere. If we don't have it, let's create that and use it here and in the other web components. Good call @janechu !

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
```diff
@@ -7,9 +7,14 @@ export const RadioGroupStyles = css`
...
     .positioning-region {
         display: flex;
         flex-wrap: wrap;
     }
+    .vertical {
+        flex-direction: column;
+    }
+    .horizontal {
+        flex-direction: row;
+    }
```
EisenbergEffect, line 19: Got to love the simplicity of this.

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
marjonlynch: I can do that but it would potentially mess up the label slot

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: Yea I did mention this on another PR or in discord but I'm not sure if we do or not, I will use the one in web-utilities or make it if it doesn't exist

packages/web-components/fast-components/src/radio-group/fixtures/base.html
marjonlynch: O'douls it is

packages/web-components/fast-components/src/radio-group/radio-group.ts
marjonlynch: yea I don't see one or any existing module that makes sense to put it in.

marjonlynch: oh we do, it's in aria.ts, ooook

marjonlynch: done

packages/web-components/fast-components/src/radio-group/fixtures/base.html
marjonlynch: done, tree types

packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
marjonlynch: done and done

packages/web-components/fast-components/src/radio-group/radio-group.template.ts
marjonlynch: I think if we have the label slot we don't necessarily want it to fall in the flex of the radios so the orientation is in a good place where it's at now. Unless you always want the label to be inline but in the horizontal (default) case that doesn't seem like what we'd want?

marjonlynch: going to resolve this, I don't see any objection

PR #3094, opened 2020-05-09
feat: add a gradient accent fill recipe based on hue shifting

bheston:
# Description

Added a gradient fill recipe based on accent color. Works by shifting the hue a currently fixed amount.

image

## Motivation & context

For the new FAST site.

Will need another version or modifier for less saturated version. Looking into an algorithm for that based on designs.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/react/fast-components-styles-msft/src/accent-button/index.ts
+++ b/packages/react/fast-components-styles-msft/src/accent-button/index.ts
@@ -24,6 +24,7 @@ import {
     highContrastSelectedOutline,
     highContrastSelector,
 } from "../utilities/high-contrast";
+import { accentFillGradientRest } from "../utilities/color/accent-fill-gradient";
 
 const styles: ComponentStyles<AccentButtonClassNameContract, DesignSystem> = {
     ...baseButton,
@@ -31,7 +32,8 @@ const styles: ComponentStyles<AccentButtonClassNameContract, DesignSystem> = {
         ...buttonStyles(),
         color: accentForegroundCut,
         fill: accentForegroundCut,
-        background: accentFillRest,
+        // background: accentFillRest,
+        background: accentFillGradientRest,
         "&:hover:enabled, a&:not($button__disabled):hover": {
             background: accentFillHover,
             ...highContrastSelectedOutline,
--- a/packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
+++ b/packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
@@ -1,6 +1,6 @@
-import { DesignSystem } from "src";
 import { ColorHSL, hslToRGB, parseColor, rgbToHSL } from "@microsoft/fast-colors";
 import { format } from "@microsoft/fast-jss-utilities";
+import { DesignSystem } from "../../design-system";
 import {
     FillSwatchFamily,
     Swatch,
@@ -11,7 +11,7 @@ import {
     SwatchRecipe,
     SwatchResolver,
 } from "./common";
-import { accentFill } from ".";
+import { accentFill } from "./index";
 
 function gradientSwatchRecipe<T extends SwatchFamily>(
     type: keyof T,
@@ -22,21 +22,23 @@ function gradientSwatchRecipe<T extends SwatchFamily>(
             type,
             callback
         );
-        const color: Swatch = baseRecipe(designSystem);
-        const colorHsl: ColorHSL = rgbToHSL(parseColor(color));
-        let hShifted: number = colorHsl.h + 28;
-        if (hShifted > 360) {
-            hShifted -= 360;
+        const baseColor: Swatch = baseRecipe(designSystem);
+        const baseColorHSL: ColorHSL = rgbToHSL(parseColor(baseColor));
+        let shiftedHue: number = baseColorHSL.h + 28;
+        if (shiftedHue > 360) {
+            shiftedHue -= 360;
         }
-        const secondaryHsl: ColorHSL = ColorHSL.fromObject({
-            h: hShifted,
-            s: colorHsl.s,
-            l: colorHsl.l,
-        });
+        const shiftedColor: Swatch = hslToRGB(
+            ColorHSL.fromObject({
+                h: shiftedHue,
+                s: baseColorHSL.s,
+                l: baseColorHSL.l,
+            })
+        ).toStringHexRGB();
         return format(
             "linear-gradient(0deg, {0} 0%, {1} 100%)",
-            (a: any): string => color,
-            (a: any): string => hslToRGB(secondaryHsl).toStringHexRGB()
+            (a: any): string => baseColor,
+            (a: any): string => shiftedColor
         )(designSystem);
     };
 }
--- a/packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
+++ b/packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
@@ -1,0 +1,55 @@
+import { DesignSystem } from "src";
+import { ColorHSL, hslToRGB, parseColor, rgbToHSL } from "@microsoft/fast-colors";
+import { format } from "@microsoft/fast-jss-utilities";
+import {
+    FillSwatchFamily,
+    Swatch,
+    SwatchFamily,
+    SwatchFamilyResolver,
+    swatchFamilyToSwatchRecipeFactory,
+    SwatchFamilyType,
+    SwatchRecipe,
+    SwatchResolver,
+} from "./common";
+import { accentFill } from ".";
+
+function gradientSwatchRecipe<T extends SwatchFamily>(
+    type: keyof T,
+    callback: SwatchFamilyResolver<T>
+): SwatchResolver {
+    return (designSystem: DesignSystem): Swatch => {
+        const baseRecipe: SwatchRecipe = swatchFamilyToSwatchRecipeFactory<T>(
+            type,
+            callback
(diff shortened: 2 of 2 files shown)


packages/react/fast-components-styles-msft/src/accent-button/index.ts
```diff
@@ -24,14 +24,16 @@ import {
...
 
 const styles: ComponentStyles<AccentButtonClassNameContract, DesignSystem> = {
     ...baseButton,
     button: {
         ...buttonStyles(),
         color: accentForegroundCut,
         fill: accentForegroundCut,
-        background: accentFillRest,
+        // background: accentFillRest,
+        background: accentFillGradientRest,
```
chrisdholt, line 36: Do we want to actually implement this and export with this package? My assumption is no as it would ship to those leveraging the packages. If the intent here was to validate, can we revert this change?

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
...
+>(SwatchFamilyType.rest, accentFill);
+export const accentFillGradientHover: SwatchResolver = gradientSwatchRecipe<
+    FillSwatchFamily
+>(SwatchFamilyType.hover, accentFill);
+export const accentFillGradientActive: SwatchResolver = gradientSwatchRecipe<
+    FillSwatchFamily
+>(SwatchFamilyType.active, accentFill);
+export const accentFillGradientFocus: SwatchResolver = gradientSwatchRecipe<
+    FillSwatchFamily
+>(SwatchFamilyType.focus, accentFill);
```
chrisdholt, line 55: Should we add these to the fast-components design system provider? If this is for the site, we'll want these recipes to be reflected there so we can leverage them.

nicholasrice:
I'm concerned that this recipe will not work with any other recipe and breaks the assumptions of the rest of the color system. How can any other recipe use the product of this algorithmically? How is this this gradient configured? How is a text color calculate for this recipe?

I'd prefer to see this private for the time being if it needs to be utilized today, but I think if gradients is a requirement then a broader look at how we make the entire system accommodate that would be prudent.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
+import { DesignSystem } from "src";
```
janechu, line 1: This pathing is going to break on the distributed package.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
...
+    FillSwatchFamily,
+    Swatch,
+    SwatchFamily,
+    SwatchFamilyResolver,
+    swatchFamilyToSwatchRecipeFactory,
+    SwatchFamilyType,
+    SwatchRecipe,
+    SwatchResolver,
+} from "./common";
+import { accentFill } from ".";
```
janechu, line 14: Can you make this "./index"? There's been some issue with potential path resolution between different OS's

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
...
+        const baseRecipe: SwatchRecipe = swatchFamilyToSwatchRecipeFactory<T>(
+            type,
+            callback
+        );
+        const color: Swatch = baseRecipe(designSystem);
+        const colorHsl: ColorHSL = rgbToHSL(parseColor(color));
+        let hShifted: number = colorHsl.h + 28;
+        if (hShifted > 360) {
+            hShifted -= 360;
+        }
```
janechu, line 30: Can this be:
```
const maxColorH: number = colorHsl.h > 332
    ? colorHsl.h - 332
    : colorHsl.h + 28;
```

I think (if I understand this correctly) that we should treat the linear gradient as having a range and that you are defining the `min` and `max` values of that range so the naming should follow that logic.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
...
+        const secondaryHsl: ColorHSL = ColorHSL.fromObject({
```
janechu, line 31: So here instead of `secondaryHsl` maybe `maxHSL`. Since it's also used only once you could either avoid creating the const or do:
```
const maxColor: Swatch = hslToRGB(
    ColorHSL.fromObject({
        h: hShifted,
        s: colorHsl.s,
        l: colorHsl.l,
    })
).toStringHexRGB()
```

This way below you can do:
```
return format(
    "linear-gradient(0deg, {0} 0%, {1} 100%)",
    (a: any): string => minColor,
    (a: any): string => maxColor
)(designSystem);
```

I might have some of these return types wrong.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
...
+    type: keyof T,
+    callback: SwatchFamilyResolver<T>
+): SwatchResolver {
+    return (designSystem: DesignSystem): Swatch => {
+        const baseRecipe: SwatchRecipe = swatchFamilyToSwatchRecipeFactory<T>(
+            type,
+            callback
+        );
+        const color: Swatch = baseRecipe(designSystem);
+        const colorHsl: ColorHSL = rgbToHSL(parseColor(color));
```
janechu, line 26: Could this be `minColorHSL` instead of `colorHSL`? (see my note below)

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
```diff
@@ -0,0 +1,55 @@
...
+function gradientSwatchRecipe<T extends SwatchFamily>(
+    type: keyof T,
+    callback: SwatchFamilyResolver<T>
+): SwatchResolver {
+    return (designSystem: DesignSystem): Swatch => {
+        const baseRecipe: SwatchRecipe = swatchFamilyToSwatchRecipeFactory<T>(
+            type,
+            callback
+        );
+        const color: Swatch = baseRecipe(designSystem);
```
janechu, line 25: Could this be `minColor` instead of `color`? (see my note below)

packages/react/fast-components-styles-msft/src/accent-button/index.ts
bheston: Yeah, this was mostly for validation. It could definitely go into another package more specifically for the site or FAST Frame.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
bheston: The design system provider is for the properties used to generate the recipes. In this case they build from the base accent recipe. We could potentially add properties for the hue offset and gradient parameters like start / end % and angle.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
bheston: 360 is significant because it's the hue range for colors. 28 is the variable here, and 332 is just 360 - 28. I'd actually prefer this logic be in the color package, as hue is often thought of as a repeating circle, especially for wrapping like this. We could update this here, but 332 is not significant, so I wouldn't want to hard-code that. Actually, 28 could possibly be a design system property if building this for wider use.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
bheston: I took this mostly. I went with 'shifted' instead of 'max' though. I don't think it's as much min / max as it is original / shifted (or something along those lines). For instance, if the shift were -28 instead of +28, what would be min / max at that point?

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
bheston: Made it baseColor.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
bheston: Made it baseColorHSL.

bheston:
@nicholasrice your points occurred to me as I was working on this. I'm not sure the right way to calculate the contrast over a gradient, if it's possible to use the strongest difference between A and B, or if it's possible the colors in between could be less contrast than A or B. In the current usage it's not being used for the color calculation, but letting the base accent color stand in. As you point out, that may not be safe, and not the way the recipes are intended to chain. To fully support gradients (or perhaps acrylic) I think we'll need to extend the system. The color 'value' is going to have to be provided differently than the actual style value.

packages/react/fast-components-styles-msft/src/utilities/color/accent-fill-gradient.ts
janechu: Good point

bheston:
This will likely live somewhere else, once we address how to handle it within the recipe infrastructure and for a11y.

PR #3096, opened 2020-05-09
feat: add fast-components-msft as a new package

### LINKED ISSUE #3088: [Intent to implement] Change default styles for fast-components and create a new MSFT package
This issue exists to raise awareness of our intent to implement the following (significant) changes:

1. Change the default design system and styles for the `fast-components` package.
2. Create a new package for Microsoft styled components: `fast-components-msft`

## Background
In our previous implementation of FAST, we were building for both a variety of frameworks and design systems. This led us to make certain decisions about how we structured the project and how packages were broken up. As we've created our web component packages, we elected to include all core functionality (class, template, default styles) as part of a single package. The reasoning here was that we are building as an extension of the web platform and not individual frameworks. Our goal of being able to use FAST with any framework is more easily accomplished by building with web standards. 

Once we had decided to include all the parts and pieces in a single tree-shakeable package, it made sense to also include a default design system and a default set of components as the primary exports. This allows users to leverage components directly from the `fast-components` package or leverage just the parts they need and recompose their own components. At the time we had started building out our web component proof of concept, the only style and design system investment the team had fully completed was the Microsoft styles (`fast-components-styles-msft`) so the decision was made to have the default export visually match the Microsoft package for the time being.

## FAST is a foundation, MSFT is an implementation
As we move forward and look to shift out of "beta" and into a more stable release, we feel it's important to make a change to the default component styles we are exporting from `fast-components`. While FAST is organizationally under Microsoft, the goals of our project are industry focused. Additionally, FAST has its own brand and its own visual design language. Our goals with the FAST design system and the `fast-components` package is to provide flexibility at scale, a good foundation for individuals and teams to build with, as well as the ability to create their own design systems by mixing the component templates and classes with styles specific to their design language. 

With this in mind, and considering that we now have a UI system for FAST-DNA (currently named "fast-frame"), shifting our default exports to align with these goals makes sense. Additionally, we hold a strong belief that goals and components which are specific to Microsoft should not be elevated above those of other design systems and organizations. This is a good thing for the community and for the goals which are specific to Microsoft. By providing a MSFT specific package, we can better support the needs of that community. In many cases, this may be as simple as providing convenient patterns which are specific to the MSFT design system but not prevalent in other systems (such as [this example](https://github.com/microsoft/fast-dna/pull/2732#discussion_r397338684)).

## Next Steps / Implementation
We've created a feature branch to manage this change without disrupting other in-progress work (`features/fast-components-msft`). Within a week we'll raise a PR to merge these changes from the feature branch to master. As mentioned above, the change will establish the following:

1. The default exports of `fast-components` will be FAST styled components leveraging the default FAST design system values. The components exported here will follow the convention of being prefixed with `fast`:
```html
<fast-button>Button</fast-button>
```
2. A new MSFT package which implements the template and classes from `fast-components` with MSFT specific styles and a MSFT specific implementation of the design system. This will serve as an example and pattern for other design systems wishing to leverage the system. The components exported here will follow the convention of being prefixed with `msft`:
```html
<msft-button>Button</msft-button>
```

This change is breaking, and I believe a good reflection of what should be our v1 implementation of `fast-components` and `fast-components-msft`. As always, we welcome your feedback and thoughts!

chrisdholt:
# Description
This PR is <s>the initial work for</s> closes #3088. This PR creates a new package for `fast-components-msft`. The rewrite of existing styles for the FAST design system <s>is not included as part of this PR.</s> has been completed and a new package for `fast-foundation` was created to hold the templates, classes, and utilities needed to compose web components with FAST.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components-msft/src/badge/index.ts
```diff
@@ -0,0 +1,26 @@
...
+            this.classList.add(newValue);
+            this.classList.remove(oldValue);
+        }
+    }
+
+    // Can we remove this and circular as attrs if we don't want to leverage them?
+    // Can I set this internally only in this implementation?
+    @attr({ mode: "fromView" })
+    public color: string = "value";
+}
```
chrisdholt, line 26: @microsoft/fast-dna-collaborators my thoughts on the implementation of Badge here was to tighten up the implementation to better align with the MSFT badge requirements. Ideally I'd like to not provide the color value here  and just have things work as expected. Since we cannot remove or delete attributes from the classes here, I'm trying to consider how to manage the mappings. 

It seems like having defaults for both `fill` and `color` make sense; while not illustrated above we could also ensure the extensibility by also supporting a string:
```ts
export type BadgeFill = "accent" | "lightweight" | "neutral" | string;
```
I'm not sure what to do with the color attribute in those scenarios though if we elect to handle them in the stylesheet.

packages/web-components/fast-components-msft/src/divider/fixtures/divider.html
```diff
@@ -0,0 +1,8 @@
+<fast-design-system-provider use-defaults>
+    <h1>Divider</h1>
+    <h4>Default</h4>
+    <fast-divider></fast-divider>
+
+    <h4>Presentation</h4>
+    <fast-divider role="presentation"></fast-divider>
```
chrisdholt, line 7: @microsoft/fast-dna-collaborators We've talked about having two different recipes based on whether this needs to be accessible to a screen reader. In the event that the divider is only for presentation, the recipe would not need to meet contrast requirements. In the event that it is exposed to AT, the recipe would meet contrast requirements. I think we should consider adding this for the MSFT implementation here. @bheston - thoughts on the recipe to use for the contrast requirement implementation?

packages/web-components/fast-components-msft/src/design-system-provider/index.ts
```diff
@@ -0,0 +1,7 @@
+import {
+    designSystemProvider,
+    FASTDesignSystemProvider,
+} from "@microsoft/fast-components";
+
+@designSystemProvider("msft-design-system-provider")
+export class MSFTDesignSystemProvider extends FASTDesignSystemProvider {}
```
chrisdholt, line 7: @nicholasrice does this map to what you might expect here? I'd expect the default values to change here once we create the FAST system.

packages/web-components/fast-components-msft/src/badge/index.ts
eljefe223: The `fast-components` api would remain the same?

eljefe223:
Smoked to ensure everything is working properly

packages/web-components/fast-components-msft/src/badge/index.ts
chrisdholt: I think this is part of the question...since we can't pull off existing attrs, extending from the Badge would expose all of those values - so yes, the API would stay the same. I think providing some default "named" configurations is helpful here, but wondering what our language/naming should align to. Recipes?

eljefe223: I think the naming seems right to me. It aligns with what we do elsewhere.

eljefe223: The one thing about our react `badge` is that the `filled` version was yellow and not aligned with the accent color. So might want @bheston to chime in here.

chrisdholt: Yeah, I'd like some thoughts on the defaults - we can still keep things extensible. My concern with the yellow is that it won't respond to dark mode unless we leverage the provider which seems heavy-handed for a default value.

Bigger question in my mind is what to do with the color value (if anything)...do we provide it but not include a default value b/c the fill defaults will be recipe associated?

packages/web-components/fast-components-msft/README.md
```diff
@@ -0,0 +1,12 @@
+# FAST Components MSFT
+
+## Development
+
+To start the component development environment, run `yarn start`.
```
janechu, line 5: The contributor documentation should not be in the `README`, same issue with the `fast-components` `README`. This should be for developers looking to install the package and use the exports.

chrisdholt: Sure, this was a complete copy/paste, I'll update prior to merging.

packages/web-components/fast-components-msft/src/badge/index.ts
EisenbergEffect: One option might be to move `color` so that it's defined on `FASTBadge` rather than `Badge`.

packages/web-components/fast-components-msft/src/anchor/index.ts
```diff
@@ -0,0 +1,14 @@
+import { customElement } from "@microsoft/fast-element";
+import { Anchor, AnchorTemplate as template } from "@microsoft/fast-components";
+import { AnchorStyles as styles } from "./anchor.styles";
```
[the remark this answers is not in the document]
chrisdholt: yeah, I wanted to do that work in a separate branch as part of the feature branch work. I can do in this package, but we'll need both packages to be touched prior to the feature branch merging in

packages/web-components/fast-components-msft/src/badge/index.ts
chrisdholt: @EisenbergEffect there would be changes to the template as well I think - those values are set in a nested div as they are setting `style` attributes. I wonder if the deltas here are worth different API's or if we should keep consistent for the time being.

EisenbergEffect: Ah, I see. Well, then it becomes a matter of what is worse, having some properties that don't apply vs. writing a whole new component.

packages/web-components/fast-components-msft/src/design-system-provider/index.ts
nicholasrice: Yep

chrisdholt:
> One thing I've been wondering about...should we use a different element prefix for each design system or not?
> 
> We've got "fast-" and now "msft-" for the element names. This will make it hard for people to switch out design systems. They'll have to re-write all their html. Also, the scenario where somebody builds a single UI and then just dynamically loads different design systems based on platform (ios, vs. material vs. msft) isn't really possible because the element names are different. If the element names were the same, then the design language and implementation could simply be switched at load.
> 
> Also, with different element names, it feels a bit more like different components as apposed to the same components with a different design system, at least to me.
> 
> In any case, I wanted to bring this up as I'm not sure if we've thought through these scenarios.

I think there are different ways to accomplish this. It really depends. The underlying intent here is that these are intended to be different stylesheets for the design systems. There will be different design system values as well between the FAST implementation and the MSFT implementation. This is not simply a configuration change at this point where we're swapping out the design system values - the entire system may be different for certain folks' implementations. Further, design systems apply values (such as corner radius) in different ways and based on different principles. Something like `elevated-corner-radius` isn't actually something that exists currently in the FAST design system. We simply have a corner-radius. Where we may want to make FAST extremely flexible and robust, it's quite possible that the MSFT design system will tighten up a bit more in an attempt to provide greater alignment with something like Fluent. It's possible over time that we will gain insight that will help us better support endless configuration of design systems easily, but we aren't there yet and I don't think it's wise for us to try and force alignment out of the gate.

I'm fine keeping this prefixed with `fast-`, but I'm curious if you think there are concerns with naming collision. Additional concerns are that the design systems and their default values are different, but we now have alignment in name. 

To your last point - I think it's extremely important to be clear - these are different custom elements and not just a different design system configuration. If we swap the default stylesheet to align with the FAST, we cannot create exactly the MSFT styled components. Currently we would need class overrides to ensure that the components look exactly like we currently expect the MSFT components to look. There are certain things which are just going to be different, like the thumb on slider. To accomplish that with the design system, we would need to elevate a value or principle to the design system that currently doesn't really exist. That's just one example. Over time, there may be greater divergence here between systems as well.

packages/web-components/fast-components-msft/package.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "@microsoft/fast-components-msft",
+    "description": "A library of Web Components",
```
janechu, line 3: Should we specify these are Microsoft styles?

chrisdholt: yes

packages/web-components/fast-components-msft/src/design-system-provider/index.ts
```diff
@@ -0,0 +1,650 @@
...
+     * Define design system property attributes
+     */
+    @attr({ attribute: "background-color", ...fromView })
+    @designSystemProperty({
+        cssCustomProperty: "background-color",
+        default: DesignSystemDefaults.backgroundColor,
+    })
+    public backgroundColor: string;
+
+    @attr({ attribute: "accent-base-color", ...fromView })
```
nicholasrice, line 49: No need for the `@attr` or `@observable` anymore - see the `fast-components` version

eljefe223:
Smoked both packages, There seems to be a small amount of visual issues especially around focus. But I think all these issues were already there.

The only big issue I see is the tabbing behavior on the MSFT Text Field. You need to tab twice on each control, once to focus the container and again to be able to start typing

chrisdholt:
> Smoked both packages, There seems to be a small amount of visual issues especially around focus. But I think all these issues were already there.
> 
> The only big issue I see is the tabbing behavior on the MSFT Text Field. You need to tab twice on each control, once to focus the container and again to be able to start typing

Great catch - missing delegate focus. We'll do another run on any remaining visual issues. Well done!

PR #3104, opened 2020-05-11
chore: add design system provider components to fast-website

radium-v:
# Description

Adds the `fast-design-system-provider` and `website-design-system-provider` components. To accommodate this, I also added `mini-css-extract-plugin` to the webpack build and moved the stylesheet to `src/app/css/style.css` for future (related) improvements.

## Motivation & context
We want DS properties to propagate when controls values change. Having two allows us to keep the styles of the page and the styles of the `fast-frame` component separated.

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/components/section-header/section-header.styles.ts
```diff
@@ -9,4 +13,14 @@ export const SectionHeaderStyles = css`
...
-`;
+
+    :host ::slotted(.section-badge) {
+        color: var(--accent-color);
+    }
+`.withBehaviors(
+    cssCustomPropertyBehaviorFactory(
+        "accent-color",
+        d => d.accentBaseColor,
+        DesignSystemProvider.findProvider
```
nicholasrice, line 24: This should be a reference to the `WebsiteDesignSystemProvider` since that is the custom element constructor you're expecting to resolve this recipe.

sites/fast-website/src/app/css/style.css
```diff
@@ -1,68 +1,79 @@
...
-}
-
-.section-heading {
-  font-size: 46px;
-  line-height: 56px;
-  margin: 0;
-}
\ No newline at end of file
+:root {
+  /* FIXME: Is there a better way to set default props to prevent FOUC? */
```
nicholasrice, line 2: Try making the above provider reference change and see if this still repros.

sites/fast-website/src/app/components/section-header/section-header.styles.ts
nicholasrice: actually - I *think* this should just use the `accentForegroundRestBehavior` from `@microsoft/fast-components`

sites/fast-website/src/public/index.html
```diff
@@ -3,11 +3,14 @@
   <head>
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <link rel="stylesheet" href="style.css">
     <title><%= htmlWebpackPlugin.options.title %></title>
   </head>
   <body>
-    <fast-design-system-provider use-defaults>
+    <fast-design-system-provider
```
nicholasrice, line 9: should this be the `website-design-system-provider`?

sites/fast-website/src/public/index.html
```diff
@@ -3,11 +3,14 @@
...
+      use-defaults
+      background-color="#242424"
+      accent-base-color="#F33378"
```
nicholasrice, line 12: You'll probably also want to change the accentPalette property here also - that's a bit tricky because none of this is in JavaScript. You'll likely to need find this element imperatively and set the palette property or set this up as a custom element that sets that property. You can reference`fast-dna/packages/react/fast-components-styles-msft/generate-palettes.js` for how to do that.

This might be fine for now but just FYI.

sites/fast-website/src/public/index.html
radium-v: I wasn't sure if we should use separate buckets or buckets of the same type. Does it make any real difference when they're nested?

nicholasrice: I guess it depends on the requirements - what requirement is driving toward using two different provider types?

sites/fast-website/src/app/css/style.css
radium-v: After making the change above, I'm still getting the same white flash without setting at least this var.

nicholasrice: hmm. Could be due to script loading after the DOM is parsed - I can look into this one

sites/fast-website/src/app/components/section-header/section-header.styles.ts
```diff
@@ -9,4 +9,8 @@ export const SectionHeaderStyles = css`
         box-sizing: border-box;
         text-align: center;
     }
+
+    :host ::slotted(.section-badge) {
```
chrisdholt, line 13: Is the space between :host and ::slotted causing an issue here?

sites/fast-website/src/public/index.html
```diff
@@ -3,11 +3,14 @@
   <head>
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <link rel="stylesheet" href="style.css">
     <title><%= htmlWebpackPlugin.options.title %></title>
   </head>
   <body>
-    <fast-design-system-provider use-defaults>
+    <website-design-system-provider id="body-provider" class="body-provider"
```
janechu, line 9: Just a note that if this is temporary because we are missing something from the `fast-design-system-provider` can you add a comment in here with "TODO" so we can scrape it later and clean it up.

sites/fast-website/src/app/components/design-system-provider/website-design-system-provider.ts
```diff
@@ -0,0 +1,23 @@
...
+export class WebsiteDesignSystemProvider extends FASTDesignSystemProvider {
+    @observable
+    @designSystemProperty({
+        cssCustomProperty: false,
+        default: createColorPalette(parseColor("#F33378") as ColorRGBA64),
+    })
+    public accentPalette: string[];
+
+    accentBaseColorChanged(): void {
+        const parsedColor = parseColor(this.accentBaseColor);
```
nicholasrice, line 20: Can you make this handle the null case more gracefully? If the color can't be parsed then maybe it is reset back to the previous color?

Then you can diff the previous and next values, and only create the accent palette when the two are different and current is a valid color

sites/fast-website/src/app/index.ts
```diff
@@ -1,2 +1,4 @@
+import "./css/style.css";
```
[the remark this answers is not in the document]
radium-v: The `mini-css-extract-plugin` handles this correctly in Webpack. There should be a `dist/main.css` file getting generated.

sites/fast-website/src/app/components/design-system-provider/website-design-system-provider.ts
radium-v: Which null cases do you mean? When the attribute is present but it's empty, when the attribute isn't present, or when the attribute is set to an invalid color value?

janechu:
Not necessarily in this PR, but our convention is to use `dist` for a distributed npm package and `www` for site files, so at some point this needs to be changed.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -129,20 +158,40 @@ export const FastFrameTemplate = html<FastFrame>`
...
                         <fast-button appearance="accent">
-                            Button                            
-                            <svg class="icon-cut" slot="end" width="11" height="15" viewBox="0 0 11 15" fill="none" xmlns="http://www.w3.org/2000/svg">
-                                <path d="M10.4309 13.8721C10.7451 13.872 11 14.1238 11 14.4345C11 14.7452 10.7454 14.9971 10.4312 14.9972L0.56913 15C0.254899 15.0001 0 14.7483 0 14.4376C0 14.127 0.25457 13.875 0.568801 13.8749L10.4309 13.8721ZM5.42279 0.0051353L5.5 0C5.78804 0 6.0261 0.211628 6.06377 0.486201L6.06897 0.562533L6.06821 10.8269L8.89165 8.03648C9.09367 7.83679 9.40976 7.81866 9.63248 7.98209L9.69629 8.03656C9.89827 8.23629 9.9166 8.54881 9.7513 8.76901L9.69621 8.8321L5.90489 12.5798C5.70299 12.7794 5.38711 12.7976 5.16438 12.6344L5.10057 12.58L1.30485 8.83233C1.08251 8.6128 1.08226 8.25662 1.3043 8.03679C1.50616 7.83694 1.82224 7.81857 2.04509 7.98182L2.10894 8.03625L4.93028 10.8216L4.93103 0.562533C4.93103 0.277745 5.14508 0.0423843 5.42279 0.0051353L5.5 0L5.42279 0.0051353Z"/>
+                            Button
+                            <svg
```
eljefe223, line 162: Interesting this formatting is happening in your review. Prettier should have handled this for us.

radium-v: This is how it came in from the base branch for me.

Issue #3076, opened 2020-05-07
Homepage Chore Checklist


radium-v:
This issue is intended to hold chores and minor bugfixes/best practice nudges for the new homepage before we go live.

- Meta:
  - [ ] Favicon (#3388)
  - [ ] title and description
- Background:
  - [x] Background should dim further after scrolling past the hero
- Navbar:
  - [x] **Logo link is redundant - this page should not navigate to itself**
  - [x] text links should be darker at rest
  - [x] Logo text should be SVG (#3355)
  - [x] Github logo should be the accent
  - [x] non-mobile breakpoints should have same alignment style (logo on left, links in center, github link on right)(#3344)
- Section indicator:
  - [x] Items shouldn't be clickable (#3338)
- All Sections:
  - [x] **Add Activ Grotesque - Chris has CDN link and values** (#3368)
  - [x] **Reduce height of each module**
  - [x] `.section-badge` should have: (#3296)
  - [x] more letter spacing (#3296)
  - [x] black background (#3296)
  - [x] rounded corners (#3296)
  - [x] `<site-section-header>`s should constrain width at non-mobile breakpoints (#3347)
  - [x] Increase margin between paragraphs and buttons (#3296)
- Hero Section:
  - [ ] "Get Started" button has no link
  - [x] paragraph text should be larger (#3296)
- Adaptive UI Section:
  - [ ] Frame width causes other sections to overflow at middle breakpoints
  - [ ] "Explore components" button has no link
  - [ ] card bg not updating when dark mode toggled or bg color value updated
  - [ ] Hue and saturation sliders missing track styles
- Frameworks section:
  - [ ] "Learn More" button has no link
  - [x] **Add backplate to match features** (#3366)
  - [x] **Add transition to backplate** (#3366)
- Features section:
  - [x] **Drop the hand on hover as these are not clickable** (#3363)
  - [x] **Add transition to backplate** (#3366)
  - [x] **Grey text using filter should shift to opacity. Grey looks muddy.** (#3366)
  - [ ] "Read documentation"
  - [x] Each card should also use `:focus-within` to set background color for visibility when using  keyboard navigation(#3284)
  - [ ] Every anchor needs a link
- Community section:
  - [x] Each card should also use `:focus-within` to set background color for visibility when using keyboard navigation (#3284)
- Footer:
  - [x] Missing divider line (#3346)
  - [x] text links should be darker at rest (#3346)

khamudom:
responsive

[...]

PR #3105, opened 2020-05-11
feat: create content-placement component for marketing site

SamanthaAO:
# Description

this component is used to generate cards in the frameworks and community section.
Screen Shot 2020-05-11 at 10 03 21 AM
Screen Shot 2020-05-11 at 10 03 38 AM
the template consists of one default slot and 4 named slots  (icon, body, action, and compatibility). the styles define the layouts of the cards and change them based on hover

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/components/content-card/content-card.styles.ts
```diff
@@ -0,0 +1,51 @@
...
+        display: none;
+    }
+
+    fast-card {
+        background: transparent;
+        padding: 0 calc(var(--design-unit) * 5px) calc(var(--design-unit) * 3px);
+        text-align: left;
+    }
+
+    fast-card:hover .contentCard_footer__hoverVisible {
```
EisenbergEffect, line 21: I don't recall seeing BEM anywhere else. I'm not opposed to it, but just want to make sure we're being consistent. /cc @chrisdholt

sites/fast-website/src/app/components/content-card/content-card.template.ts
```diff
@@ -0,0 +1,23 @@
+import { html, when } from "@microsoft/fast-element";
+import { ContentCard } from "./content-card";
+
+export const ContentCardTemplate = html<ContentCard>`<fast-card>
```
chrisdholt, line 4: This is really only "card-like" on hover and I think we'll have a competing "content card" potentially based on the design. Can we find another name that might work here?

SamanthaAO: would framework-and-community-content be more suitable?

SamanthaAO: So for this template FrameworkAndCommunityContentTemplate.

chrisdholt: Let's go with something like `content-block` or `content-placement` or `content-_____` to try and keep it less verbose and less focused on the sections. If we add additional sections that name would not extend to support it

sites/fast-website/src/app/components/content-card/content-card.styles.ts
radium-v: BEM was discussed in an earlier PR but I don't think it's a really high priority. It's being tracked in #3076 to be handled before launch.

eljefe223:
Seeing what others think, but to me it's more intuitive if the default slot is the largest section of text / content. It just seems easier to manage than having the slot assigned to that large bit of text.

I propose ditching the `body` slot and making that the default and then adding a title slot.

chrisdholt:
> Seeing what others think, but to me it's more intuitive if the default slot is the largest section of text / content. It just seems easier to manage than having the slot assigned to that large bit of text.
> 
> I propose ditching the `body` slot and making that the default and then adding a title slot.

I tend to think in terms of what the component or design would *require* rather than the largest amount of content. Would we ever have body without a heading given the current design? I think it's more likely that the heading would *always* be there. Things like an image, action, or body/summary text may not always be required.

eljefe223:
> > Seeing what others think, but to me it's more intuitive if the default slot is the largest section of text / content. It just seems easier to manage than having the slot assigned to that large bit of text.
> > I propose ditching the `body` slot and making that the default and then adding a title slot.
> 
> I tend to think in terms of what the component or design would _require_ rather than the largest amount of content. Would we ever have body without a heading given the current design? I think it's more likely that the heading would _always_ be there. Things like an image, action, or body/summary text may not always be required.

I'm under the impression that they are both required, could be a rule to consider if you have multiple required attributes.

sites/fast-website/src/public/index.html
```diff
@@ -65,7 +65,47 @@
         <!-- frameworks:start -->
         <section class="section">
           <div class="section-component">
-            Framework components
+            <site-content-placement divider=true >
```
nicholasrice, line 68:
```suggestion
            <site-content-placement divider>
```

For Boolean attributes you don't need to assign a value - you can just add the attribute name :)

sites/fast-website/src/app/components/content-placement/content-placement.ts
```diff
@@ -0,0 +1,5 @@
+import { FASTElement, attr } from "@microsoft/fast-element";
+
+export class ContentPlacement extends FASTElement {
+    @attr divider: boolean = false;
```
eljefe223, line 4: you'll want to do:
```
@attr({ mode: "boolean" })
public divider = true;
```
this will keep the attribute from writing to the DOM when its false

sites/fast-website/src/public/index.html
SamanthaAO: Awesome! Thank you.

sites/fast-website/src/app/components/content-placement/content-placement.ts
nicholasrice, line 4:
```suggestion
    @attr({mode: "boolean"}) divider: boolean = false;
```

You'll want to make sure this attribute *converts* to a boolean when the attribute is added. You can do that by setting the "mode" to "boolean".

SamanthaAO: Will do! Thank you.

eljefe223:
This should also be a feat not a chore

sites/fast-website/src/app/components/content-placement/content-placement.ts
SamanthaAO: Will do! Thank you.

khamudom:
Looks like you can get into a state where all cards blurred. By hover over the edge of the section.
image

SamanthaAO:
That is because currently the whole site uses a grid so on the left and right there is a section where the container is a bit larger than the cards because of how the grid is set up. If the overall layout is changed to flex then the container could be the same size as the cards instead of being the size of the column in the grid that it is defined as. I did not want to make a change to the overall layout at this time with this component, but it is definitely something that should be addressed.

PR #3106, opened 2020-05-11
fix(fast-element): html helper should handle literal and directive interpolated together

EisenbergEffect:
# Description

Adds unit tests for the `html` tagged template helper, as well as some type fixes, a bug fix in the `html` helper, and some improvements to npm tasks.

## Motivation & context

In preparation for working on the new "content binding" feature proposal, I wanted to start putting some tests around the areas that might be affected. This PR adds tests around the `html` tagged template helper only, not the full template compiler, as a first step. Here's what's included:

* 20 unit tests on the `html` tagged template helper asserting that strings and interpolated values are properly interwoven across value types and positioning.
* Fixed a bug in the `html` tagged template helper that arises when interpolating both literal and directive values together.
* Fixed some types in the compilation functions to more accurately portray expectations.
* Added API docs for the `html` tagged template helper as well as `ViewTemplate` and associated interfaces. Not all of these APIs were documented though. I intentionally left out a few which I think might go away after the "content binding" feature lands so I can prevent unnecessary work.
* Fixed the test watch tasks.
* Added a new `dev` and `tdd` task for streamlined development. Simply execute `npm run tdd` and the code builds in watch mode and the tests are run in watch mode. You can now alter code and tests and on save code will recompile and tests will rerun.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-element/src/template.ts
```diff
@@ -139,7 +169,7 @@ export class ViewTemplate<TSource = any, TParent = any> extends Directive
 }
 
 // Much thanks to LitHTML for working this out!
-export const lastAttributeNameRegex =
```
janechu, line 142: Wouldn't this be a breaking change since this is being exported from the `index.ts`?

EisenbergEffect: Good catch. Technically, yes. That was never intended to be exported though. I can't imagine anyone using that directly.

PR #3107, opened 2020-05-11
fix: slider labels did not change when parent slider attributes changed in javascript

marjonlynch:
# Description
If a developer wrote javascript to change the orientation, min, or max attribute of the slider parent all children labels would not update accordingly. This is important from an absolute positioning standpoint for the labels. Also this PR fixes RTL support which has been broken for awhile.

Changes include using the notifier/subscriber API to have the slider labels get notification of orientation, max and min attribute changes in the parent slider and making adjustments accordingly.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/slider-label/slider-label.ts
```diff
@@ -25,63 +26,86 @@ export class SliderLabel extends FASTElement {
...
+                        break;
+                    case "min":
+                        this.sliderLabel.min = source.min;
+                        break;
+                    default:
+                        break;
+                }
+                this.sliderLabel.positionStyle = this.sliderLabel.positionAsStyle();
+            },
+        };
```
nicholasrice, line 68: I think this *could* cause a memory leak because you're not maintaining the object reference to unsubscribe when the component disconnects.

To circumvent you could either:
1. store this handler object on the element instance
2. give this element a `handleChange` method and subscribe the instance itself: `notifier.subscribe(this, "min")`

Then unsubscribe on tear-down

EisenbergEffect: Unfortunately, no weak ref magic. I wish, but it's not in the language yet. @nicholasrice is correct in that these handlers will need to be torn down. You can just store the notifier and the handler and then unsubscribe in the disconnectedCallback.

packages/web-components/fast-components/src/slider/fixtures/base.html
```diff
@@ -1,29 +1,30 @@
...
-            </fast-slider-label>            
-            <fast-slider-label
-                position="10"
-            >
-                10&#8451;
-            </fast-slider-label>
-            <fast-slider-label
-                position="90"
+        <div style="margin-bottom: 60px;">
+            <button style="width: 90px;margin: 10px;" onclick="const slider = document.getElementById('switcher'); slider.getAttribute('orientation') === 'horizontal' ? slider.setAttribute('orientation', 'vertical') : slider.setAttribute('orientation', 'horizontal');">Change orientation</button>
```
EisenbergEffect, line 4: That's some serious inline JS there. Any better way to handle this? I know storybook has some limitations, but maybe some code can just run a selector at startup and add a normal even handler?

packages/web-components/fast-components/src/slider/slider.ts
```diff
@@ -206,8 +185,27 @@ export class Slider extends FormAssociated<HTMLInputElement>
         }
     };
 
+    private setThumbPositionForOrientation = (direction: Direction): void => {
```
EisenbergEffect, line 188: I appreciate this being factored out. Thanks!

packages/web-components/fast-components/src/slider/fixtures/base.html
marjonlynch: I tried a few simpler things I thought should work like an inline style block but couldn't seem to get it to resolve my function. I'll try again though. believe me I didn't want to do it that way :)

packages/web-components/fast-components/src/slider-label/slider-label.ts
marjonlynch: I think my recent push does all the recommended things, and bonus, it still works :)

packages/web-components/fast-components/src/slider/slider.ts
marjonlynch: I did that because I had to for direction to work right  but you are welcome sir! :)

packages/web-components/fast-components/src/slider-label/slider-label.ts
```diff
@@ -1,13 +1,20 @@
...
+    attr,
+    FASTElement,
+    Notifier,
+    Observable,
+    observable,
+} from "@microsoft/fast-element";
+import { Direction, Orientation } from "@microsoft/fast-web-utilities";
+import { SliderConfiguration } from "../slider";
 import { convertPixelToPercent } from "../slider/slider-utilities";
+import { FASTSlider } from "../slider";
```
janechu, line 11: you can combine these and just do:

```ts
import {
    FASTSlider,
    SliderConfiguration
} from "../slider";

File: packages/web-components/fast-components/src/slider/slider.template.ts [hunk 5]
```diff
@@ -1,10 +1,13 @@
 import { html } from "@microsoft/fast-element";
 import { ref } from "@microsoft/fast-element";
+import { Orientation } from "@microsoft/fast-web-utilities";
 import { Slider } from "./slider";
 
 export const SliderTemplate = html<Slider>`
     <template
         role="slider"
+        class="${x => (x.readOnly ? "readonly" : "")} 
+        ${x => (x.orientation === Orientation.horizontal ? "horizontal" : "vertical")}"
```
janechu, line 10: It looks like if x.orientation is undefined it will be vertical. Is that what you want?

If not I think you can just do:

```ts
${x => x.orientation || Orientation.horizontal}
```

The `Orientation` enum uses strings so you should just need to use the enum.

marjonlynch: orientation is defaulted true, so if the element doesn't have the attribute specified it will always be horizontal. Good point on the string thing, doh! :)

janechu: Do you mean defaulted to `Orientation.horizontal`?

packages/web-components/fast-components/src/slider-label/slider-label.ts
marjonlynch: done and done

packages/web-components/fast-components/src/slider/slider.template.ts
marjonlynch: doubly done, thanks, looks much better

packages/web-components/fast-components/src/slider/fixtures/base.html
marjonlynch: I think I might need to become more of a storybook pro to get this looking nicer, I will do that but might take some time, do we want to hold this PR up or just remove this code or leave it as is until I get storybook fu upgraded?

EisenbergEffect: We can move ahead. It's not a huge issue.

marjonlynch: Ok, I'm pretty sure there are nice ways to test this in storybook.

PR #3111, opened 2020-05-11
feat: add navigation menu component

khamudom:
# Description

Created a navigation component to use for the navigation page menu. The template is a nav element with a slotted directive. Each navigation item also gets an attribute of `role="listitem"` added. Implemented the control on the page in the header and footer.

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
Approved but had a second though. I think we should prefix all the components built specifically for the site with something other than `fast-`. We could get name conflicts that break things in the future if we use the same "namespace" and then add new components to the actual fast components. /cc @chrisdholt

sites/fast-website/src/public/style.css
```diff
@@ -65,4 +54,21 @@ body {
...
+
+.footer {
+  grid-column: 2;
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 84px;
+}
+
+.footer-navigation-item {
+  margin-right: 50px;
```
marjonlynch, line 73: does this have any rtl issue? do you care about rtl in this site and do you have any ability to get the direction? I'm not sure of any of these, just asking

sites/fast-website/src/app/components/site-navigation/site-navigation.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export class SiteNavigation extends FASTElement {
+    @observable
+    items: HTMLElement[];
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.items.forEach((item: any) => {
+            if (item instanceof HTMLElement) {
+                item.setAttribute("role", "listitem");
```
chrisdholt, line 11: @khamudom can you point to why we're passing a role of listitem to whatever is passed within this?

khamudom: Thanks for bringing this up, I should have made a note asking if adding a role was necessary or not. Since I was not using a ul and li in this nav list, I wanted to make sure each of the items is part of a list group.

chrisdholt: I'd consider using something like `repeat()` to return each slotted item into a list item. We'll want the items here (if `fast-anchor`) to present to AT as anchors. If we want to render the items into a list, I would try and do so in the template.

khamudom: Initially, I thought of using `repeat()`, with an anchor, but decided on slotted directive instead, to give the flexibility of being able to slot other components inside the nav.

chrisdholt: It's a matter of *where*. If you add it to `fast-anchor`, the control will not present to AT with a role of anchor. If you render each item within a list item, you get the list behavior and it doesn't clobber the role of the anchors passed into slots.

sites/fast-website/src/public/style.css
nicholasrice: margin-inline-end will swap this value in RTL if that is desired.

sites/fast-website/src/public/style.css
```diff
@@ -36,21 +40,6 @@ body {
   background-color: #3A3A3A;
```
eljefe223, line 40: This should be using color recipes now that John has the design system provider up

khamudom: I will just delete it. It a temporary background to show a sectioned space.

sites/fast-website/src/app/components/site-navigation/site-navigation.ts
khamudom: sounds like it is not necessary for this scenario. I will remove it

nicholasrice:
I think this should be a feat - not a chore :)

sites/fast-website/src/app/components/navigation/navigation.styles.ts
```diff
@@ -0,0 +1,10 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "@microsoft/fast-components";
+
+export const NavigationStyles = css`
+    ${display("flex")} :host {
+        font-family: var(--body-font);
+        color: var(--neutral-foreground-rest);
```
nicholasrice, line 7: You'll want to make sure you include the `neutralForegroundRestBehavior` as a dependency of this stylesheet

sites/fast-website/src/app/components/navigation/navigation.ts
```diff
@@ -0,0 +1,6 @@
+import { FASTElement, observable } from "@microsoft/fast-element";
+
+export class Navigation extends FASTElement {
+    @observable
+    items: HTMLElement[];
```
chrisdholt, line 5: @khamu are we doing anything with `items` here? If we're just dropping anchors in then you can likely drop the slotted directive from the template and drop this here.

eljefe223:
I think we should add `start` and `end` slots for any supplemental content

sites/fast-website/src/app/components/navigation/navigation.template.ts
```diff
@@ -0,0 +1,10 @@
+import { html, slotted } from "@microsoft/fast-element";
+import { Navigation } from "./navigation";
+
+export const NavigationTemplate = html<Navigation>`
+    <template>
+        <nav>
```
eljefe223, line 6: Add `start` and `end` slots here.

sites/fast-website/src/app/components/navigation/navigation.template.ts
```diff
@@ -0,0 +1,10 @@
...
+            <slot ${slotted("items")}></slot>
```
eljefe223, line 7: add `start` and `end` slots here.

eljefe223: This will allow you place the logo and other items inside the navigation component.

image

sites/fast-website/src/app/components/navigation/navigation.ts
khamudom: not doing anything with the items, so will remove the slotted directive

sites/fast-website/src/app/components/navigation/navigation.styles.ts
khamudom: thanks, added

sites/fast-website/src/app/components/navigation/navigation.template.ts
khamudom: added

PR #3115, opened 2020-05-12
feat: add navigational items on the page, toolbar and menu

### LINKED ISSUE #3099: Create a menu for the tooling examples site
Child item of #3093 

## Requirements
- Create a menu on the left side of the site in `fast-tooling-examples`
- Include in the menu 2 Categories, `Tooling` and `React Tooling`

### LINKED ISSUE #3100: Create a toolbar for the tooling examples site
Child item of #3093

## Requirements
- Create a toolbar link at the top which includes:
  - A right and left aligned section that has a minimum width
  - A link to the location of the site in github on the right side
  - A link back to the index of the site on the left side
- The toolbar should be at the top of the page
- The toolbar should not have any special UX and should remain in its location as the page is scrolled (not sticky)

janechu:
# Description

Adds a menu and toolbar as well as basic elements to scaffold the examples site.

## Motivation & context

Resolves #3099, Resolves #3100 

## Issue type checklist

## Process & policy checklist


sites/fast-tooling-examples/app/style.css
```diff
@@ -1,3 +1,63 @@
...
+    min-height: 100vh;
+}
+
+.toolbar {
+    grid-column-start: 1;
+    grid-column-end: 3;
+    grid-row-start: 1;
+    grid-row-end: 1;
+    display: grid;
+    grid-template-columns: 50% 50%;
```
eljefe223, line 28: I prefer using `fr` units. `1fr 1fr` It holds up better since `grid-gap` gets included in the calculation. I know your not using grid-gap but if we decide to change down the line it would just work.

sites/fast-tooling-examples/app/style.css
```diff
@@ -1,3 +1,63 @@
...
+    border-bottom: 1px solid #e1e1e1;
```
eljefe223, line 29: No design system values?

janechu: A design will be worked out at a later date, this site has no schedule to be live so design details are kept to a minimum.

sites/fast-tooling-examples/app/style.css
```diff
@@ -1,3 +1,67 @@
...
+    grid-column-end: 2;
+    text-align: end;
+}
+
+.menu {
+    grid-column-start: 1;
+    grid-column-end: 1;
+    grid-row-start: 2;
+    grid-row-end: 2;
+    border-right: 1px solid #e1e1e1;
```
bheston, line 48: What about `border-inline-end. Seems like we're getting away from the directionSwitch config, and this should work for RTL.

sites/fast-tooling-examples/app/style.css
```diff
@@ -1,3 +1,67 @@
...
+}
+
+.category-title {
```
bheston, line 51: Per our chat, this is good temporarily. Ideally styling would be applied to all <h*> elements by default, or classes generated through a type ramp helper method, which is part of the headings spec, but not implemented yet.

janechu: Sounds good, I'll put a comment in here to we can come back to it.

PR #3119, opened 2020-05-13
feat: add feature-card and card-section components

freefaller69:
# Description

Create feature-card template component extended from fast-card with a header slot, footer slot, and default slot for content.  Create card-section component for grid placement of feature-cards.  Implement hover behaviors on cards and card sections to match design.

Screen Shot 2020-05-13 at 1 40 35 PM
Screen Shot 2020-05-13 at 1 39 57 PM

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/components/card-section/card-section.template.ts
```diff
@@ -0,0 +1,4 @@
+import { html } from "@microsoft/fast-element";
+import { CardSection } from "./card-section";
+
+export const CardSectionTemplate = html<CardSection>` <slot></slot> `;
```
awentzel, line 4: are the spaces intended before and after slots?

chrisdholt: This is a good catch, I think this is going to actually include spaces within the template, which (I think) is unintended.

sites/fast-website/src/app/components/card-section/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { CardSectionTemplate as template } from "./card-section.template";
+import { CardSectionStyles as styles } from "./card-section.styles";
+import { FASTCard } from "@microsoft/fast-components";
+
+@customElement({
+    name: "card-section",
+    template,
+    styles,
+})
+export class FeatureCard extends FASTElement {}
```
awentzel, line 11: I probably would remove FAST as part of the imported name and just go with Element or Card.  The imported class is already fast, so it just simplifies slightly.  Unless there is extra meaning that's necessary.  This is just my opinion.

chrisdholt: Just for context, FASTElement is the name of the class which is being extended. It does appear that FASTCard is not being used here though - if it's not, can we drop that import?

sites/fast-website/src/app/components/card-section/index.ts
```diff
@@ -0,0 +1,14 @@
+import { customElement, FASTElement } from "@microsoft/fast-element";
```
awentzel, line 1: Your other names start capitalized, but, not "customElement" or "css", "display" etc.  What naming standard are we following?  Can we align these to be consistent, throughout this PR and actually entire project?

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,69 @@
+import { css } from "@microsoft/fast-element";
+import { display, neutralForegroundRestBehavior } from "@microsoft/fast-components";
+
+export const FeatureCardStyles = css`
+    ${display("grid")} :host {
+        contain: layout;
+        grid-template-columns: 1fr 2fr;
+        color: var(--neutral-foreground-rest);
+        font-family: var(--text-font);
```
chrisdholt, line 9: where does this var exist? We may want to simply inherit unless there are alternate fonts being loaded and used.

sites/fast-website/src/public/index.html
```diff
@@ -131,7 +131,160 @@ <h3>AngularJS (1.x) 1.7.9</h3>
         <!-- documentation:start -->
         <section class="section">
           <div class="section-component">
-            Documentation components
+            <card-section>
+							<feature-card>
```
awentzel, line 135: Something off with the major intentation of `<feature-card>`.  Can we pull that back in?  seem to also happen on properties of `<fast-anchor>` also?  If Prettier is doing this we should correct.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+        box-sizing: border-box;
+        padding: 20px;
+        box-shadow: unset;
+        height: 140px;
```
chrisdholt, line 13: If the content ever overflows the card here we may see it cut off. Let's consider how this works in responsive scenarios where the content might vary and increase the size.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+        width: 600px;
```
chrisdholt, line 14: What happens if my screen size is 320px?

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    :host(:hover) ::slotted(fast-anchor) {
+        opacity: 1;
+    }
+
+    :host(:hover)::before {
+        opacity: 0;
+    }
+
+    :host(:hover) ::slotted(h5) {
```
chrisdholt, line 36: Is this always going to be an h5?

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    ::slotted(p) {
+        margin: 0;
+    }
+
+    ::slotted(fast-anchor) {
+        margin-right: 20px;
+        opacity: 0;
+        color: var(--accent-fill-rest);
```
chrisdholt, line 67: Do we need to reset the color on the anchor? This seems like it's possibly covering up a bug.

sites/fast-website/src/public/index.html
```diff
@@ -131,7 +131,160 @@ <h3>AngularJS (1.x) 1.7.9</h3>
...
+									>
+									<fast-anchor
+											slot="footer"
+											class="link"
+											href="#"
+											appearance="lightweight"
+											>Read Documentation</fast-anchor
+									>
+							</feature-card>
+						</card-section>
```
chrisdholt, line 287: Yikes - can we fix this formatting?

sites/fast-website/src/public/index.html
chrisdholt: Prettier does not target html files currently.

sites/fast-website/src/app/components/card-section/index.ts
awentzel: That's a great point.  Let's make sure that there are no unused imports, variables, or even compiler warnings.

sites/fast-website/src/app/components/card-section/index.ts
chrisdholt: FYI, this is the API of `fast-element` and intentional. @EisenbergEffect can provide more detail but we're mapping to standards here.

sites/fast-website/src/app/components/card-section/card-section.template.ts
freefaller69: Good question.  I had that on separate lines so prettier must have moved it to one line.  I'll remove the spaces.

sites/fast-website/src/app/components/card-section/index.ts
freefaller69: Excellent point.  I will update that.  I actually meant to extend FASTCard as it is, so that needed a change anyway.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: I saw it being used in the main css file and assumed it was going to be established in :root.  Inherit works just as well.  I will change it to inherit.

sites/fast-website/src/app/components/card-section/index.ts
freefaller69: customElement is the function export defined in fast-element.  css and display are functions as well, so the casing seems to follow convention.

sites/fast-website/src/public/index.html
freefaller69: Yeah, I noticed that.  It doesn't reflect the file in my local, so I suspect it is a prettier issue.
Screen Shot 2020-05-14 at 11 32 31 AM

sites/fast-website/src/app/components/card-section/index.ts
EisenbergEffect: Currently, decorators are functions, so we apply function casing. `customElement` is one such decorator. `html` and `css` are tag template functions, so we also apply function casing. `FASTElement` is a class, so we apply class casing.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Understandable.  I was going off of the design initially here.  Kham said he was checking on what responsive behaviors design was looking for.  I will remove the height.  It's easy enough to put back in if need be.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Not necessarily I suppose.  I could try an alternative selector to try and grab that element.

sites/fast-website/src/public/index.html
freefaller69: Aaron pointed the same thing out.  Local file looks fine.  If there is something I can change on my end to fix that I'd love to do so, I'm not sure what it would be though.

radium-v: @freefaller69 It looks like these are tabs instead of spaces.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Is there a better option for setting the color on the anchor?  If I set appearance="accent" I do get the right color but it is a large button style appearance instead of the accent colored hypertext link.  I could use the section-anchor class that Sam added to the site stylesheet if that is preferred.

sites/fast-website/src/public/index.html
freefaller69: > Prettier does not target html files currently.

Is it something I can't see in my html file that I could fix?  Or a quirk of GitHub perhaps?

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
chrisdholt: Accent is not an anchor appearance, which is likely why this is not working. Is the goal to target the `lightweight` version?

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: If I remove the width and scale it down, things get pretty tight.  It would look better at that point to reorient the card so the number and header are above the other content which I am happy to do.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Looking closer at the design in Figma it could well be neutral.  Although if I set appearance to neutral or remove appearance entirely it changes the positioning slightly and stops applying the accent color.

sites/fast-website/src/public/index.html
freefaller69: Let me see what I can do.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
chrisdholt: What color are you trying to get?

freefaller69: Screen Shot 2020-05-14 at 12 44 47 PM
The accent pink here.  Which I believe is the accent pink that Jason mentioned revising in discord.

chrisdholt: That's the lightweight appearance.

sites/fast-website/src/public/index.html
chrisdholt: Not a simple way to fix this currently with any kind of script. I'd just drop the formatting all the way back and try to re-indent and see if it's fixed. It could be an issue with your editor settings; hard to tell without digging in.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: OK, then I have the appearance set right.  Would using the class from the main stylesheet for the site be preferred for the color?

chrisdholt: The color should use the accent palette, so if you set "lightweight" that should use the correct recipe to deliver an accent colored lightweight anchor depending on the contrast ratio. I guess I'm not following the issue here. The anchor should be delivering the color. If it is not, something else is going on.

freefaller69: Perhaps that is yet to be set then?  If I remove the setting the text is the same as the base text off-white color.
Screen Shot 2020-05-14 at 1 29 54 PM

sites/fast-website/src/public/index.html
freefaller69: I think I got it.  My editor was setting tabs instead of spaces to indent on return even though I don't have it set to do so as far as I could find.

sites/fast-website/src/public/index.html
freefaller69: I think I got it.  My editor was setting tabs instead of spaces to indent on return even though I don't have it set to do so as far as I could find.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Changed the selector from h5 to :first-child.

sites/fast-website/src/app/components/index.ts
```diff
@@ -3,3 +3,5 @@ export { SiteSectionHeader } from "./section-header";
 export { FastFrame } from "./fast-frame";
 export { SiteNavigation } from "./navigation";
 export { SiteContentPlacement } from "./content-placement";
+export { FeatureCard } from "./feature-card";
+export { CardSection } from "./card-section";
```
SamanthaAO, line 7: Could these have the prefix site added to them to align with the other components?

sites/fast-website/src/app/components/feature-card/index.ts
```diff
@@ -0,0 +1,14 @@
+import { customElement } from "@microsoft/fast-element";
+import { FeatureCardTemplate as template } from "./feature-card.template";
+import { FeatureCardStyles as styles } from "./feature-card.styles";
+import { FASTCard } from "@microsoft/fast-components";
+
+@customElement({
+    name: "feature-card",
```
SamanthaAO, line 7: could site be added to this name? such as site-feature-card

sites/fast-website/src/app/components/card-section/index.ts
```diff
@@ -0,0 +1,14 @@
+import { customElement } from "@microsoft/fast-element";
+import { CardSectionTemplate as template } from "./card-section.template";
+import { CardSectionStyles as styles } from "./card-section.styles";
+import { FASTCard as Card } from "@microsoft/fast-components";
+
+@customElement({
+    name: "card-section",
```
SamanthaAO, line 7: could site be added to this name? such as site-card-section

sites/fast-website/src/app/components/card-section/card-section.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+    }
+
+    ${display("grid")} :host {
+        grid-template-rows: repeat(4, min-content);
+        grid-auto-flow: var(--flow);
+    }
+
+    :host(:hover) ::slotted(feature-card) {
+        filter: blur(1px);
```
SamanthaAO, line 21: I believe the blur in the figma is 2px. I know this is a tiny difference, but I want to make sure they my content and your cards have the same blur as they do in the figma.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,67 @@
+import { css } from "@microsoft/fast-element";
+import { display, neutralForegroundRestBehavior } from "@microsoft/fast-components";
+
+export const FeatureCardStyles = css`
+    ${display("grid")} :host {
+        contain: layout;
+        grid-template-columns: 1fr 2fr;
+        color: inherit;
+        font-family: var(--text-font);
```
SamanthaAO, line 9: A few of the other components use  var(--body-font). I was wondering if there was a benefit in using --text-font instead or if this could be changed to --body-font for consistency since they all have the same font-family in the figma.

sites/fast-website/src/public/index.html
```diff
@@ -131,7 +131,57 @@ <h3>AngularJS (1.x) 1.7.9</h3>
...
+                <fast-anchor slot="footer" class="link" href="#" appearance="lightweight">View Github</fast-anchor>
+                <fast-anchor slot="footer" class="link" href="#" appearance="lightweight">Read Documentation</fast-anchor>
+              </feature-card>
+              <feature-card>
+                <h5 slot="header">08</h5><h4 slot="header">Other</h4>
+                <p>Leverage a toolkit of general utilities such as keyboarding, Right-To-Left (RTL), number, and string manipulation.</p>
+                <fast-anchor slot="footer" class="link" href="#" appearance="lightweight">View Github</fast-anchor>
+                <fast-anchor slot="footer" class="link" href="#" appearance="lightweight">Read Documentation</fast-anchor>
+							</feature-card>
+						</card-section>
```
SamanthaAO, line 184: I noticed that there were other comments about spacing, and thought I would put a note here so you saw it was present here as well.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
freefaller69: Yeah, in the design is says 2px but 2px blur in browser looks WAY more blurry than 2px blur looks in Figma.  I made the tweak to get the look.

sites/fast-website/src/public/index.html
freefaller69: I think I have it fixed now, although I have no idea what actually caused it.  My editor was setting tabs on RETURN while the rest of the document is spaces.

sites/fast-website/src/app/components/index.ts
freefaller69: Has that been settled?  I was following the discussion but don't recall seeing an official decision.  Brian was the last comment I saw on the matter.

chrisdholt: we need to prefix one way or another; we should use whichever prefix we've landed for the site

sites/fast-website/src/app/components/card-section/index.ts
freefaller69: Done

sites/fast-website/src/app/components/feature-card/index.ts
freefaller69: Done

sites/fast-website/src/app/components/index.ts
freefaller69: Done

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Added responsive behavior to the card.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Looks like the recipe is now in place.  Removed the component color styling.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: I just removed it for now until we have a defined one.

sites/fast-website/src/app/components/feature-card/feature-card.template.ts
```diff
@@ -0,0 +1,14 @@
...
+export const FeatureCardTemplate = html<FeatureCard>`
+    <header>
+        <slot name="header"></slot>
+    </header>
+    <main>
+        <slot></slot>
+        <div>
+            <slot name="footer"></slot>
+        </div>
+    </main>
```
chrisdholt, line 13: There should only be one main element per page and I'm not sure how this is going to present to assistive technologies. Similarly, I don't know that this is the appropriate usage of the header html element. I think the content here is going to dictate the markup - IE, passing a h2 or h3 as the header and passing other semantic markup as body text. Let's swap those elements to divs to be safe.

sites/fast-website/src/app/components/feature-card/feature-card.ts
```diff
@@ -0,0 +1 @@
+export class FeatureCard {}
```
chrisdholt, line 1: What's the scenario here? Is this being used at all?

sites/fast-website/src/app/components/feature-card/feature-card.template.ts
```diff
@@ -0,0 +1,14 @@
+import { html } from "@microsoft/fast-element";
+import { FeatureCard } from "./feature-card";
+
+export const FeatureCardTemplate = html<FeatureCard>`
+    <header>
+        <slot name="header"></slot>
```
chrisdholt, line 6: I think the default slot should be what you have as header currently. I don't think volume of content is important here. I tend to think in terms of long term extensibility as the design changes the constant for something like this is going to be some kind of "heading" or "header" area which represents whatever is dropped into the content area.

SamanthaAO:
Not right now of course because today we just want to get everything in, but I think it would be cool to work together to make one container that worked for both of our cards. just to make the formatting a bit more similar. I think that if we added a when with a section of your card and data for the repeat my container might work, but we would have to see if the styling could find a happy middle ground for both of our cards/content-placements. Just something to think about.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -0,0 +1,76 @@
...
+        opacity: 0;
+    }
+
+    :host(:hover) ::slotted(:first-child) {
+        color: var(--accent-fill-rest);
+    }
+
+    @media screen and (max-width: 540px) {
+        ${display("grid")} :host {
+            grid-template-columns: repeat(3, minmax(50px, 1fr));
```
EisenbergEffect, line 40: Nice use of grid!

sites/fast-website/src/app/components/feature-card/feature-card.template.ts
freefaller69: I'll get it switched up!

sites/fast-website/src/app/components/feature-card/feature-card.ts
freefaller69: It is, but this wasn't actually set up right.  I've updated it.

sites/fast-website/src/app/components/feature-card/feature-card.template.ts
freefaller69: OK, I will make that change.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Thank you, sir!  :)

sites/fast-website/src/app/components/card-section/index.ts
awentzel: @EisenbergEffect Can this be added to the documentation if not already?

sites/fast-website/src/public/index.html
```diff
@@ -89,7 +89,57 @@
         <!-- documentation:start -->
         <section class="section">
           <div class="section-component">
-            Documentation components
+            <site-card-section>
+              <site-feature-card>
+                <h5 slot="header">01</h5>
```
awentzel, line 94: I wouldn't expect an <h5> to precede an <h4> for semantic and accessibility?

chrisdholt: Div has no semantics, so you aren't going to get your intended structure. @awentzel is the recommendation here to have this be an h4 as well? We technically can take this as a follow-up, but we should absolutely ensure that the HTML markup is both semantic and valid, which I believe is @awentzel's point here.

sites/fast-website/src/app/components/feature-card/feature-card.template.ts
freefaller69: I changed the header element to a div with a card_heading class and the main to a div with card_body class.  Sound reasonable and sufficiently semantic?

sites/fast-website/src/app/components/feature-card/feature-card.template.ts
freefaller69: I've switched things up here.  Default slot goes to the first div with .card_header class and the second div, with .card_body class, has a slot for body content and a footer slot for the anchors.

packages/web-components/fast-components/src/default-palette.ts
```diff
@@ -96,101 +96,101 @@ export const neutralPalette: string[] = [
...
+    "#002B4D",
+    "#002847",
+    "#002542",
+    "#00223C",
+    "#001F36",
+    "#001B30",
+    "#00182B",
+    "#001525",
+    "#00121F",
+    "#000000"
```
chrisdholt, line 195: @freefaller69 this looks like a bad rebase. The default palette changed and this appears to be reverting it. See if you can work with @radium-v or @eljefe223 to revert this change.

sites/fast-website/src/app/components/card-section/card-section.ts
```diff
@@ -0,0 +1 @@
+export class CardSection {}
```
chrisdholt, line 1: Should this extend anything or include anything? This class doesn't appear to support generic HTML attributes?

freefaller69: It should.  Let me get that squared away.

packages/web-components/fast-components/src/default-palette.ts
freefaller69: fixed

sites/fast-website/src/app/components/card-section/card-section.ts
freefaller69: updated the index to extend FastElement.  Although as I think about it, that may not really be necessary.  Other than applying a filter on hover, this is all structural.  Sam and I are going to be looking at composing a more universal component for laying out the various cards on the page instead of having two different ones.

PR #3122, opened 2020-05-13
feat: add fast-foundation as a new package

chrisdholt:
# Description
This PR adds fast-foundation as a new package. This package includes the templates, classes, and common utilities used to compose fast web components.

## Issue type checklist

## Process & policy checklist


janechu:
I believe this should be marked as a breaking change.

chrisdholt:
> I definitely think this is the right way to go. Thanks @chrisdholt, especially for adding the usage docs. Nice touch.
> 
> One teeny, weeny, very annoying request have to make....can we add the .js file extension to all the imports? I can do it myself when I fix up the builds, but I'd prefer to have only one PR that touches all the files as opposed to two if possible.

We're having an issue with storybook for some reason handling the .js extensions. I'll start by adding it to the foundation package for sure :). We're going to need to work to solve it for the `fast-components` package though.

packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
```diff
@@ -416,7 +416,7 @@ export class DesignSystemProvider extends FASTElement
 /**
  * Defines a design-system-provider custom element
  */
-export function designSystemProvider(name: string) {
+export function designSystemProvider(name: string, styles: ElementStyles | undefined) {
```
chrisdholt, line 419: @nicholasrice and @EisenbergEffect want to bring attention to this API change as part of this. Because we aren't passing styles I needed to add that as part of the decorator here for composing the custom element.

packages/web-components/fast-foundation/README.md
```diff
@@ -0,0 +1,2 @@
+# FAST Foundation
+A suite of templates, classes, and utilities for composing web components.
```
janechu, line 2: Can we make a note that this should be fleshed out like our other README files to show installation and basic use?

packages/web-components/fast-components/package.json
```diff
@@ -92,6 +92,7 @@
   "dependencies": {
     "@microsoft/fast-colors": "^5.0.8",
     "@microsoft/fast-element": "^0.8.1",
+    "@microsoft/fast-foundation": "^0.0.0",
     "@microsoft/fast-web-utilities": "^4.4.5",
     "tabbable": "^4.0.0"
```
janechu, line 97: we should now be able to remove any dependencies that fast-foundation also takes, such as tabbable

packages/web-components/fast-foundation/src/storybook-typings.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.html" {
```
janechu, line 1: I think we can remove the storybook typing if we're not using it. I also saw mention it in some of the "." prefixed files in the root of the package.

packages/web-components/fast-foundation/README.md
chrisdholt: Yeah I wanted to get feedback on what all we want to outline in here.

packages/web-components/fast-components/package.json
chrisdholt: correct - good call. Although, if we reference something like `fast-web-utilities` or `fast-element` directly will we still want those?

packages/web-components/fast-foundation/src/storybook-typings.d.ts
chrisdholt: yes, indeed will remove

chrisdholt: I believe I've pulled all references (including the now-unnecessary tsconfig.build.json file)

eljefe223:
smoked `fast-components` works

packages/web-components/fast-components/package.json
janechu: as dependencies? we shouldn't need to add them in the package json even if we references them.

packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
nicholasrice: Do we need to pass the template here also?

Rob and I discussed potentially opening this up to have the same API as `@customElement` - this might be the right time to make that change.

chrisdholt: In that case, I think we should do that. I'll update.

chrisdholt: Updated!

packages/web-components/fast-components/package.json
janechu: Per our discussion ,if they are directly referenced then explicitly add them.

packages/web-components/fast-foundation/src/badge/README.md
```diff
@@ -0,0 +1,13 @@
...
+import { customElement } from "@microsoft/fast-element";
+import { Badge, BadgeTemplate as template } from "@microsoft/fast-foundation";
+import { BadgeStyles as styles } from "./badge.styles";
+
+@customElement({
+    name: "fast-badge",
+    template,
+    styles,
+})
+export class FASTBadge extends Badge {}
```
janechu, line 12: just a note that we probably need to expand on badge with the inclusion of css variables (not necessarily in this PR but without that the usage is incomplete).

packages/web-components/fast-components/src/index.ts
```diff
@@ -3,14 +3,13 @@ export * from "./badge";
 export * from "./button";
 export * from "./card";
 export * from "./checkbox";
-export * from "./custom-properties";
+export * from "./color";
```
nicholasrice, line 6: you'll need to export the recipe behaviors from "./styles/recipes".

If we don't *need* the color recipes exported right now I would hold off

PR #3125, opened 2020-05-13
fix: set slider thumb offset against its left position in the document

### LINKED ISSUE #3121: thumb position wildly offset in fast-slider
When placing a horizontal `<fast-slider>`, the thumb position doesn't align with the mouse cursor and is offset by margin/grid/content.

Example:

As shown in this image, I added `margin: auto;` to the first div in the storybook fixture, then clicked the very left of the slider. The slider's value is getting set to 3 instead of 0.
image

This can be reproduced in Storybook by adding a `margin-left` to the first div in the fixture and clicking around on the horizontal sliders.

The slider's thumb should be positioned to match the user's selection. Vertical sliders don't seem to have this problem.

* OS & Device: Windows 10 on PC
* Browsers: Tested in Edge 81.0.416.72 and Firefox 76.0.1

radium-v:
# Description

When any margin or offset was present with a slider, that offset wasn't taken into account when calculating the thumb's position.

## Motivation & context

Fixes #3121

## Issue type checklist

## Process & policy checklist

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast-dna/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/en/contributing/standards) for this project.

### DIFF
--- a/packages/web-components/fast-components/src/slider/slider.ts
+++ b/packages/web-components/fast-components/src/slider/slider.ts
@@ -254,9 +254,13 @@ export class Slider extends FormAssociated<HTMLInputElement>
         if (this.readOnly || this.disabled || e.defaultPrevented) {
             return;
         }
+
         // update the value based on current position
         const eventValue: number =
-            this.orientation === Orientation.horizontal ? e.pageX : e.pageY;
+            this.orientation === Orientation.horizontal
+                ? e.pageX - this.getBoundingClientRect().left
+                : e.pageY;
+
         this.value = `${this.calculateNewValue(eventValue)}`;
         this.updateForm();
     };
@@ -303,7 +307,9 @@ export class Slider extends FormAssociated<HTMLInputElement>
             window.addEventListener("mousemove", this.handleMouseMove);
 
             const controlValue: number =
-                this.orientation === Orientation.horizontal ? e.pageX : e.pageY;
+                this.orientation === Orientation.horizontal
+                    ? e.pageX - this.getBoundingClientRect().left
+                    : e.pageY;
             this.value = `${this.calculateNewValue(controlValue)}`;
             this.updateForm();
         }
--- a/packages/web-components/fast-components/src/slider/fixtures/base.html
+++ b/packages/web-components/fast-components/src/slider/fixtures/base.html
@@ -1,5 +1,5 @@
 <fast-design-system-provider use-defaults>
-    <div style="display: flex; flex-direction: column; justify-content: space-evenly; width: 500px; height: 100vh">
+    <div style="display: flex; flex-direction: column; justify-content: space-evenly; width: 500px; height: 100vh;">
         <div style="margin-bottom: 60px;">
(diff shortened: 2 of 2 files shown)


packages/web-components/fast-components/src/slider/slider.ts
```diff
@@ -254,9 +254,13 @@ export class Slider extends FormAssociated<HTMLInputElement>
...
             return;
         }
+
         // update the value based on current position
         const eventValue: number =
-            this.orientation === Orientation.horizontal ? e.pageX : e.pageY;
+            this.orientation === Orientation.horizontal
+                ? e.pageX - this.getBoundingClientRect().left
+                : e.pageY;
+
```
chrisdholt, line 263: Tiny wrench here is #3122 - the entire pathing here is about to change as these are moving into a separate package. Can we hold for #3122?

packages/web-components/fast-components/src/slider/slider.ts
```diff
@@ -254,9 +254,13 @@ export class Slider extends FormAssociated<HTMLInputElement>
         if (this.readOnly || this.disabled || e.defaultPrevented) {
             return;
         }
+
         // update the value based on current position
         const eventValue: number =
-            this.orientation === Orientation.horizontal ? e.pageX : e.pageY;
+            this.orientation === Orientation.horizontal
+                ? e.pageX - this.getBoundingClientRect().left
```
nicholasrice, line 261: I'm a bit concerned about calling `getBoundingClientRect` in a mouseMove event handler since it can fire very rapidly - `getBoundingClientRect` is expensive and forces a layout which can be janky - how does this feel? Is there a way to accomplish this *without* `getBoundingClientRect`?

radium-v: We could reduce the number of calls by caching the value during the `mousedown` for use during the `mousemove`, or by implementing a ResizeObserver (which isn't supported in Firefox Android).

There are other performance concerns I noticed while poking around in this component, and it would probably be good to come back around to optimize things (particularly the `transition: all` part).

packages/web-components/fast-components/src/slider/slider.ts
radium-v: This is currently blocking my progress as it makes it difficult to change the slider when it's positioned on the page. I'm fine with re-doing it for #3122 if the timeline brings it in sooner rather than later - I could also make this PR against that feature branch maybe?

chrisdholt: If we can land a solution for this with @nicholasrice sometime today I'm good having this go in. The changes in #3122 are significant and I've got a good handle on everything there it's likely safer for you to make the change against master prior if we can. If we don't have a solution and 3122 goes in first we can work together to make sure the rebase is all good.

packages/web-components/fast-components/src/slider/slider.ts
nicholasrice: If you think an all-up perf pass on it is necessary lets take that as a follow-up and address this there - I don't want to block progress but could see this being an issue.

radium-v: @nicholasrice https://github.com/microsoft/fast-dna/issues/3126

PR #3130, opened 2020-05-14
fix: replaces childNodes.length check with slotted

eljefe223:
# Description
This fix removes `childNodes.length` check and replaces it with the `slotted` directive. `childNodes.length` is a one-time render. If child nodes changes, `childNodes.length` won't update because it's not observable.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/checkbox/checkbox.template.ts
+++ b/packages/web-components/fast-foundation/src/checkbox/checkbox.template.ts
@@ -1,4 +1,4 @@
-import { html, when, slotted } from "@microsoft/fast-element";
+import { html, slotted } from "@microsoft/fast-element";
 import { Checkbox } from "./checkbox.js";
 
 export const CheckboxTemplate = html<Checkbox>`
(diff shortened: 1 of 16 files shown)


packages/web-components/fast-foundation/src/checkbox/checkbox.template.ts
```diff
@@ -34,9 +34,14 @@ export const CheckboxTemplate = html<Checkbox>`
...
         </div>
-        ${when(
-            x => x.childNodes.length,
-            html` <label part="label" class="label"><slot></slot></label> `
-        )}
+        <label
+            part="label"
+            class="${x =>
+                x.defaultSlottedNodes && x.defaultSlottedNodes.length
+                    ? "label"
```
nicholasrice, line 41: I think this should always have the label class - but additionally a 'hidden' class if it is hidden.

packages/web-components/fast-foundation/src/text-area/text-area.template.ts
```diff
@@ -8,14 +8,12 @@ export const TextAreaTemplate = html<TextArea>`
...
-        ${when(
-            x => x.childNodes.length,
-            html`
-                <label part="label" class="label" for="control">
-                    <slot name="label"></slot>
-                </label>
-            `
-        )}
+        <label part="label" for="control" class="${x =>
+            x.defaultSlottedNodes && x.defaultSlottedNodes.length
```
nicholasrice, line 12: same comment on classes here.

PR #3132, opened 2020-05-14
feat: wrap attr and observable into designSystemProperty decorator

### LINKED ISSUE #3110: designSystemProperty decorator should set up attribute / property
Currently, the `@designSystemProperty` decorator does not set up observable behaviors or HTML element mapping. Because of this, the implementer must also decorate the property with observable or attr.

To reduce boilerplate it would be beneficial to roll that behavior into the `designSystemProperty` decorator, with capability to disable the behavior if desired.

nicholasrice:
# Description
The `@designSystemProperty` decorator originally made an assumption that the property being decorated was *also* an observable property. 

This change removes that assumption and does work to set up the attr / observable itself.

## Motivation & context
closes #3110

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/design-system-provider/design-system-property.ts
+++ b/packages/web-components/fast-foundation/src/design-system-provider/design-system-property.ts
@@ -44,7 +44,7 @@ export function designSystemProperty<T extends DesignSystemProvider>(
             observable(source, prop);
         } else {
             /**
-             * Default to fromView so we don't preform a bunch of DOM writes
+             * Default to fromView so we don't perform un-necessary DOM writes
              */
             if (config.mode === void 0) {
                 config = { ...config, mode: "fromView" };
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-foundation/src/design-system-provider/design-system-property.ts
```diff
@@ -6,26 +11,57 @@ import { DesignSystemProvider } from "./design-system-provider.js";
...
+
         if (!source.designSystemProperties) {
             source.designSystemProperties = {};
         }
 
+        if (attribute === false) {
+            observable(source, prop);
+        } else {
+            /**
+             * Default to fromView so we don't preform a bunch of DOM writes
```
nicholasrice, line 47:
```suggestion
             * Default to fromView so we don't perform un-necessary DOM writes
```

PR #3135, opened 2020-05-14
feat: initial work to update the fast-components design system

chrisdholt:
# Description
This is the initial work to update the fast-components design system. Related to #3088 

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/default-palette.ts
```diff
@@ -96,101 +96,101 @@ export const neutralPalette: string[] = [
     "#151515",
     "#121212",
     "#101010",
-    "#000000"
```
nicholasrice, line 99: Somthing is going on with these trailling commas - I see some PRs removing them and some adding them.

packages/web-components/fast-components/src/dialog/dialog.styles.ts
```diff
@@ -37,7 +37,7 @@ export const DialogStyles = css`
     .root {
         ${elevation} margin-top: auto;
         margin-bottom: auto;
-        border-radius: calc(var(--elevated-corner-radius));
+        border-radius: calc(var(--corner-radius));
```
EisenbergEffect, line 40: Just an interesting note...if you wanted to share styles in the future, variables can fall back to other variables. So, you could do something like this:

`border-radius: var(--elevated-corner-radius, var(--corner-radius));`

If the elevated radius is available it will use that, otherwise it will fall back to normal corner radius.

QQ: Why is `calc` needed here?

chrisdholt: On #1, there's no concept in the FAST of elevated corner radius, so we don't actually want to include that in the design system. On #2, I don't think it's needed, I just dumped the var. I can take a look at this though.

packages/web-components/fast-components/src/default-palette.ts
janechu: I'm guessing it's a prettier mismatch, chris can you try updating?

packages/web-components/fast-components/src/dialog/dialog.styles.ts
chrisdholt: Ah, #2 is because it's supposed to be multiplying by 1px

EisenbergEffect: Make sense. I wondered if it needed the calc to apply a unit.

PR #3139, opened 2020-05-14
chore(fast-foundation): new build/test/docs setup

EisenbergEffect:
# Description

Prepares the build for production release and sets up both Node.js and browser-based testing for `fast-foundation`.

## Motivation & context

We need to fix our build output so that it works with modern browsers by itself as well as within Node.js and with popular bundlers. We also need tests 😄 

There are a few things this PR puts in place.

* All `imports` have been updated to use file extensions. This ensures that the entire library will work natively in the browser with `<script type="module">` lazy loading as well as with all major bundlers and Node.js in ESM mode.
* The build now outputs an ESM build to the `esm` folder, so that it's isolated from other build output.
* The build now outputs all individual `d.ts` files to a `dts` folder, so that they are isolated from other build output. A concatenated `.d.ts` for the library is output in the root of `dist`.
* The build now outputs an `api-report.md` file that will help identify missing documentation and prevent us from unintentional breaking changes.
* The build now outputs a json file that contains all extracted documentation info, suitable as input into our documentation static generator.
* Mocha and Chai have been added for authoring tests.
* Tests run in Node.js using `mocha`, `jsdom`, and `jsdom-global`.
* Tests run in browser using Karma.
* Both Node.js and Browser tests can (and do) have custom test setup code configured with `setup-node` and `setup-browser` respectively.
* Istanbul coverage reporting is in place.
* Several commands have been added to the `package.json` to facilitate both Node and browser testing, including debugging and watching.
* The `package.json` is now set to `type: "module"` so that it runs as esm in Node.js.
* A singe cross-environment test was validated to ensure all build and setup work correctly.
* Improved the output of `npm pack` so that only dist and docs are part of the release (along with required readme, package.json, etc.)

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/package.json
```diff
@@ -15,29 +15,72 @@
...
+    "tsconfig-paths": "^3.9.0",
+    "typescript": "^3.7.5",
+    "webpack": "^4.43.0"
   },
   "dependencies": {
     "@microsoft/fast-element": "^0.8.1",
     "@microsoft/fast-web-utilities": "^4.4.5",
-    "tabbable": "^4.0.0"
+    "tabbable": "^4.0.0",
+    "tslib": "^1.13.0"
```
janechu, line 84: Where do we use tslib?

packages/web-components/fast-foundation/src/checkbox/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from "./checkbox.template";
-export * from "./checkbox";
+export * from "./checkbox.template.js";
```
awentzel, line 1: Is it possible to have TypeScript files instead of JavaScript?

EisenbergEffect: TS doesn't convert the .ts extensions to .js when it does the build, so you have to use the .js extension, but it's actually a TS file.

packages/web-components/fast-foundation/package.json
EisenbergEffect: In the places where we use decorators, TS will generate helper functions to enable that feature. Without `tslib` it will generate the same helper function into every file, bloating the end result. So, we can take a dependency on `tslib` and it will import the helpers wherever they are needed instead.

PR #3147, opened 2020-05-15
feat: update badge API and styles

chrisdholt:
# Description
This PR udpates the badge template to conditionally render the CSS variable map values when one is actually supplied. Additionally, this updates the base styles of Badge in fast-components and adds appearances to `fast-components-msft`.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components-msft/src/badge/index.ts
+++ b/packages/web-components/fast-components-msft/src/badge/index.ts
@@ -2,7 +2,7 @@ import { attr, customElement } from "@microsoft/fast-element";
 import { Badge, BadgeTemplate as template } from "@microsoft/fast-foundation";
 import { BadgeStyles as styles } from "./badge.styles";
 
-export type BadgeFill = "accent" | "lightweight" | "neutral" | "string";
+export type BadgeAppearance = "accent" | "lightweight" | "neutral" | string;
 
 @customElement({
     name: "fast-badge",
@@ -10,9 +10,12 @@ export type BadgeFill = "accent" | "lightweight" | "neutral" | "string";
     styles,
 })
 export class FASTBadge extends Badge {
-    @attr
-    public fill: BadgeFill = "lightweight";
-    private fillChanged(oldValue: BadgeFill, newValue: BadgeFill): void {
+    @attr({ mode: "fromView" })
+    public appearance: BadgeAppearance = "lightweight";
+    private appearanceChanged(
+        oldValue: BadgeAppearance,
+        newValue: BadgeAppearance
+    ): void {
         if (oldValue !== newValue) {
             this.classList.add(newValue);
             this.classList.remove(oldValue);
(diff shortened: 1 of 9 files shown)


packages/web-components/fast-components-msft/src/badge/index.ts
```diff
@@ -2,17 +2,20 @@ import { attr, customElement } from "@microsoft/fast-element";
...
-    @attr
-    public fill: BadgeFill = "lightweight";
-    private fillChanged(oldValue: BadgeFill, newValue: BadgeFill): void {
+    @attr({ mode: "fromView" })
+    public appearance: BadgeAppearance = "lightweight";
+    private appearanceChanged(
+        oldValue: BadgeAppearance,
+        newValue: BadgeAppearance
+    ): void {
         if (oldValue !== newValue) {
```
nicholasrice, line 19: This class should be applied through the template or this can cause a runtime error when this is set during construction

chrisdholt: Are there alternatives? This isn't global across all badges and ideally we'd reuse the template.

chrisdholt: Leveraging DOM.queueUpdate() here

packages/web-components/fast-components-msft/src/badge/index.ts
```diff
@@ -1,21 +1,26 @@
-import { attr, customElement } from "@microsoft/fast-element";
+import { attr, customElement, DOM } from "@microsoft/fast-element";
 import { Badge, BadgeTemplate as template } from "@microsoft/fast-foundation";
 import { BadgeStyles as styles } from "./badge.styles";
 
-export type BadgeFill = "accent" | "lightweight" | "neutral" | "string";
+export type BadgeAppearance = "accent" | "lightweight" | "neutral" | string;
```
EisenbergEffect, line 5: Good catch on "string"

packages/web-components/fast-foundation/src/badge/badge.template.ts
```diff
@@ -5,8 +5,10 @@ export const BadgeTemplate = html<Badge>`
     <template class="${x => (x.circular ? "circular" : "")}">
         <div
             class="badge"
-            style="background-color: var(--badge-fill-${x =>
-                x.fill}); color: var(--badge-color-${x => x.color})"
+            style="${x =>
+                x.fill || x.color
```
EisenbergEffect, line 9: Is it possible that there's a fill but no color or a color but no fill?

chrisdholt: Yeah, this isn't ideal. I think we can update but I wanted to mitigate the issue of this being applied when neither was used. I think it's less egregious to have one undefined variable here that's not in use than both. We should fix this though to have both only written if they exist.

PR #3148, opened 2020-05-15
feat: style tabs to match fast components

eljefe223:
# Description
This updates the default styles for the `fast-tabs` in the `fast-component` package.

image

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/tabs/tabs.styles.ts
```diff
@@ -81,11 +88,17 @@ export const TabsStyles = css`
...
-        width: 3px;
-        height: 20px;
+        width: 5px;
+        height: 100%;
+        margin-inline-end: 10px;
         border-radius: calc(var(--corner-radius) * 1px);
         align-self: center;
         background: var(--accent-fill-rest);
+        margin-top: 0;
+        border-top-left-radius: 0;
```
radium-v, line 98: These can be combined with the shorthand
```suggestion
        border-radius: calc(var(--corner-radius) * 1px) calc(var(--corner-radius) * 1px) 0 0;
```

packages/web-components/fast-components/src/tabs/tabs.styles.ts
```diff
@@ -33,11 +35,15 @@ export const TabsStyles = css`
...
         grid-column: 1;
-        width: 20px;
-        height: 3px;
-        border-radius: calc(var(--corner-radius) * 1px);
+        width: 100%;
+        height: 5px;
         justify-self: center;
         background: var(--accent-fill-rest);
+        margin-top: 10px;
+        border-top-left-radius: calc(var(--corner-radius) * 1px);
```
radium-v, line 43:
```suggestion
        border-radius: calc(var(--corner-radius) * 1px) 0 0 calc(var(--corner-radius) * 1px);
```

eljefe223:
I made some assumptions about what we should do in the text in the tab case here. Looking at it now we may want the text to just always be `--neutral-foreground-rest` instead of `--accent-foreground-rest` Thoughts @bheston @Falkicon ?

packages/web-components/fast-components/src/tabs/tabs.styles.ts
```diff
@@ -20,6 +20,8 @@ export const TabsStyles = css`
         position: relative;
         width: max-content;
         align-self: end;
+        padding: 16px 16px 0 16px;
```
radium-v, line 23: The final value is redundant if it's the same as the second (unless that's something enforced by prettier?)
```suggestion
        padding: 16px 16px 0;
```

eljefe223: Nope :)

packages/web-components/fast-components/src/tabs/tabs.styles.ts
```diff
@@ -81,11 +86,14 @@ export const TabsStyles = css`
     :host(.vertical) .activeIndicator {
         grid-column: 1;
         grid-row: 1;
-        width: 3px;
-        height: 20px;
-        border-radius: calc(var(--corner-radius) * 1px);
+        width: 5px;
+        height: 100%;
+        margin-inline-end: 10px;
```
EisenbergEffect, line 91: Nice.

EisenbergEffect:
No problem with the code. However, regarding the actual design, this might be a bit to bulky for a tab design if we want to use it in our documentation. Our docs will need tabs, so no sure how that plays out here.

chrisdholt:
> No problem with the code. However, regarding the actual design, this might be a bit too bulky for a tab design if we want to use it in our documentation. Our docs will need tabs, so no sure how that plays out here.

Changing the density or height multiplier for a certain section of the docs might be something to leverage to solve for this :).

packages/web-components/fast-components/src/tabs/tab/tab.styles.ts
```diff
@@ -24,25 +34,45 @@ export const TabStyles = css`
...
+        fill: var(--neutral-foreground-hover);
     }
 
     :host(:active) {
         color: var(--neutral-foreground-active);
+        fill: var(--neutral-foreground-active);
+    }
+
+    :host([aria-selected="true"]) {
+        z-index: 2;
```
bheston, line 58: Any way to do this without z-index. Seems like it shouldn't be used for a flat component like this. Something to do with the indicator moving between? Can it stay fixed if we need it? Also, can it be defined only once on the base style, not each state?

packages/web-components/fast-components/src/tabs/tab/tab.styles.ts
```diff
@@ -24,25 +34,45 @@ export const TabStyles = css`
         font-weight: 400;
         line-height: 18px;
         height: calc(${heightNumber} * 1px);
-        padding: 0 calc((6 + (var(--design-unit) * 2 * var(--density))) * 1px);
-        color: var(--neutral-foreground-rest);
+        padding: calc(var(--design-unit) * 5px) calc(var(--design-unit) * 4px);
```
bheston, line 37: Tab has this for design system support, but Tabs doesn't. We might want to keep this consistent, but I think it would be fine to not add support for calculations based on design-unit until we add density support. This will likely change a bit.

packages/web-components/fast-components/src/tabs/tab/tab.styles.ts
eljefe223: appears we can, that was in the original file and not sure why it was that way orginally

packages/web-components/fast-components/src/tabs/tab/tab.styles.ts
eljefe223: I updated tabs to use design-unit.

PR #3156, opened 2020-05-16
feat(web-components): new build/test/docs setup

EisenbergEffect:
# Description

This PR puts in place the new build/test/docs setup for `fast-element`, `fast-foundation`, `fast-components`, and `fast-components-msft`.

## Motivation & context

All four packages listed above now have the following commands:

* `npm run build`- First, this will build the TypeScript source into an `esm` folder, with one JS file per TS file. It will also place the type definitions in a `dts` folder with one d.ts file per TS file. Next, it will output a concatenated single JS file build containing all source of the library and its dependencies, which can be used in a `script` tag of `type="module"`. Between these two outputs we can support Node (with esm module), browsers with script tag, and all major bundlers. Third, the build outputs a concatenated d.ts file in the `dist` folder as well as a `.api.json` file that contains a data representation of all types and documentation comments from the d.ts file. This is used to generate documentation pages (upcoming docs PR). Finally, this command outputs an `api-report.md` to the `docs` folder that documents which APIs are not documented as well as making it easy, via git diff, to see which API changes constitute breaking changes.
* `npm run test*` - A collection of test commands for running tests in Node, Chrome, and FireFox, including watch and debug modes. By default `npm run test` will run in Node for `fast-element` and `fast-foundation` and in Chrome for `fast-components` and `fast-components-msft`. This is due to missing web components features in jsdom. We can change this in the future to either use browser for everything or adopt something like https://github.com/capricorn86/happy-dom instead of jsdom for Node testing of web components.

Other items of note:

* `npm start` - Both `fast-components` and `fast-components-msft` have Storybook configured to run with this command.
* The `esm` package has been installed to enable running our esm code directly in node. This is configured with Mocha for the Node tests and can be used by anyone who wants to execute our esm code in Node.
* Some corrections/additions were made to prettier and eslint ignores around tests, which can be broken by certain source alterations made by these tools.
* The color tests in `fast-components` have been ported to `chai` and now run in both Node and browser. The tests which were previously disabled were left that way. I commented out a few tests that used mocks/spies/fakes and will circle back later by either adding an appropriate library (e.g. sinon) or writing a simple mock or two by hand.

## Next Steps

Shortly, I'll have a PR coming that adds a new docs site to the repo based on Docusaurus 2. The site will pull in markdown docs from the above repos, aggregate them with custom site markdown, and also use each package's `.api.json` file to generate API docs. Once that's in, we can then discuss merging that with the new site home page work to form the unified site for launch.

I validated the `fast-components` script file build and confirmed that one can simply add a script tag and use the components, with no extra setup, code, etc. I'll also be validating that this works correctly with Blazor and ASP.NET. I'll update my Webpack starter kit as well, to ensure that it doesn't have any issues. Each of these setups will be accompanied by docs.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components-msft/.storybook/main.js
```diff
@@ -4,14 +4,15 @@ module.exports = {
     stories: ["../src/**/*.stories.ts"],
     webpackFinal: async config => {
         config.module.rules.push({
-            test: /\.(ts)$/,
+            test: /\.(|ts|tsx)$/,
```
nicholasrice, line 7: is tsx needed?

EisenbergEffect: Ii think we can remove that. It's a vestigial copy paste oops. I take care of that real quick.

EisenbergEffect:
@nicholasrice @chrisdholt Any reason why my tests keep getting re-formatted? The tests are breaking because something is reformatting the string literals, which is rather important when verifying compiler tests. I added the specs to the prettier ignore but, which seems to work when running at the project level, but for some reason it's breaking in the build.

nicholasrice:
@EisenbergEffect Does the change get applied when you commit? In other words, is it getting malformed in by the pre-commit prettier process?

EisenbergEffect:
@nicholasrice I believe that's what happened initially. I think I've got it fixed now as I haven't seen any other commits since a much older one that affect the files under question.

packages/web-components/fast-components-msft/docs/api-report.md
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// @public (undocumented)
+export class FASTCheckbox extends Checkbox {
+}
+
+// @public (undocumented)
+export class FASTDesignSystemProvider extends DesignSystemProvider implements Omit<DesignSystem, "contrast" | "direction" | "fontWeight" | "neutralForegroundDarkIndex" | "neutralForegroundLightIndex"> {
+    // (undocumented)
+    accentBaseColor: string;
```
chrisdholt, line 58: @nicholasrice @EisenbergeEffect should we take a task to add typedoc comments to some of these?

EisenbergEffect: I think so. I plan to add a bunch to fast-element while I'm traveling on Tuesday. One of the benefits of the api report is that it will show us where we're missing docs. It also helps us to see when we make breaking changes to APIs. One thing the api-extractor team recommends is to actually use code-owners on the api-report file so that you can ensure that specific people review PRs that affect that file, since it represents changes to docs or public APIs. Seems like an interesting idea.

PR #3158, opened 2020-05-18
feat: add data grid to base React component set

scomea:
# Description
The data grid component enables authors to create displays of tabular data than can display sizeable data sets with good performance.

## Motivation & context
Fills a gap in our component set - displaying tabular data is a common need.

## Issue type checklist

## Process & policy checklist


packages/react/fast-components-react-base/src/data-grid/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+A simple data set using a "clientId" attribute as a unique identifier  could look like this.
+
+myData = [
+    {clientId: a1, name: Bob, age: 50},
+    {clientId: a2, name: Doug, age: 51},
+    etc...
+]
+
+What data is displayed in each row is controlled by the component's  "columnDefinitions" prop which consists of an array of columnDefinitions - one for each column of data to be displayed.  The columdefinition requires authors to specify a "columnDataKey" which identifies the particular field in the datarow to be displayed in the column, a "title" which describes what goes in the fixed column header cell, and a "columnWidth" which is string that sets the width of the column as for a css grid column (i.e. "50px", "1fr", "20%", etc...). The order in which columns appear is determined by the order of the "columnDefinitions" in the array.
```
janechu, line 16: Instead of quotation marks since these are properties can you use backticks?

packages/react/fast-components-react-base/src/data-grid/README.md
```diff
@@ -0,0 +1,59 @@
...
+    dataRowKey="clientId"
+    columnDefinitions={myColumnDefinitions}
+/>
+`
+
+### Virtualized content
+
+*Data grid* rendering performance is enhanced because it does not render all of the data to the DOM all the time, but rather limits what is rendered to those items that are near the visible area of the grid. 
+
+There are two mechanims that govern this in the component:
```
janechu, line 52: nit: spelling -- mechanisms

packages/react/fast-components-react-base/src/data-grid/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+- limiting how many data rows are converted to React Nodes and passed to the underlying display panel.  Only one "page" worth of data is passed to the display panel at one time and is centered on the data currently in view.  When the user scrolls near the end of a page a new page of data centered on the current scroll position is generated.  How many items get passed to the display panel is controlled by the grid's "pageSize" prop.  A page size greater than the number of items in the dataset ensures all items are passed to the display at all times if that is desired.
```
janechu, line 54: also here, can you use backticks around pageSize.

packages/react/fast-components-react-base/src/data-grid/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+- whether the underlying panel also virtualizes nodes that are not currently in view which is can be controlled by the "virtualizeItems" prop of the *data grid* component.
```
janechu, line 56: also here on virtualizeItems, can you use backticks.

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.schema.2.ts
```diff
@@ -0,0 +1,45 @@
...
+                },
+                columnWidth: {
+                    title: "Column Width",
+                    type: "string",
+                },
+            },
+        },
+        children: {
+            title: "Children",
+            type: "children",
```
janechu, line 40: To comply with tooling v2 you would need to use the export from `@microsoft/fast-tooling`, `linkedDataSchema`, see the `accent-button.schema.2.ts` for an example

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.tsx
```diff
@@ -0,0 +1,176 @@
...
+        } else {
+            return this.defaultCellRenderFunction(
+                this.props,
+                this.generateClassNames(),
+                this.props.columnDefinition.columnDataKey,
+                this.rootElement,
+                this.focusTarget,
+                unhandledProps
+            );
+        }
```
janechu, line 76: I would recommend that this is a single argument object like so (without passing through all the props):

```typescript
const config = {
    rowData: this.props.rowData[props.columnDefinition.columnDataKey],
    columnIndex: this.props.columnIndex,
    classNames: this.generateClassNames(),
    columnDataKey: this.props.columnDefinition.columnDataKey,
    rootElement: this.rootElement,
    focusTarget: this.focusTarget,
    unhandledProps: unhandledProps
};
```

you can also cut down on the duplication:

```typescript
return !isNil(this.props.columnDefinition.cell)
    ? this.props.columnDefinition.cell(config)
    : this.defaultCellRenderFunction(config);
```

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.tsx
```diff
@@ -0,0 +1,176 @@
...
+    private isDesiredFocusCell = (): boolean => {
+        if (this.context === null) {
+            return false;
+        }
+        return this.props.rowData[this.context.dataGridProps.dataRowKey] ===
+            this.context.desiredFocusRowKey &&
+            this.props.columnDefinition.columnDataKey ===
+                this.context.desiredFocusColumnKey
+            ? true
+            : false;
```
janechu, line 135: this does not need to be a ternary, you can just use the conditional and it will be a boolean.

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.tsx
```diff
@@ -0,0 +1,176 @@
...
+    protected generateClassNames(): string {
+        return super.generateClassNames(
+            classNames(this.props.managedClasses.dataGridCell)
+        );
+    }
+
+    /**
+     * The default function that renders an unstyled content display
+     */
+    public defaultCellRenderFunction = (
```
janechu, line 100: could this be `unstyledCellRenderer`? also, is this being public intended? how should it be used if so?

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.tsx
```diff
@@ -0,0 +1,176 @@
...
+     */
+    private isFocusCell = (): boolean => {
+        if (this.context === null) {
+            return false;
+        }
+        return this.props.rowData[this.context.dataGridProps.dataRowKey] ===
+            this.context.focusRowKey &&
+            this.props.columnDefinition.columnDataKey === this.context.focusColumnKey
+            ? true
+            : false;
```
janechu, line 149: dont need this to be a ternary.

packages/react/fast-components-react-base/src/data-grid/data-grid-row.schema.2.ts
```diff
@@ -0,0 +1,26 @@
...
+        gridTemplateColumns: {
+            title: "Grid template columns",
+            type: "string",
+        },
+        rowIndex: {
+            title: "Row index",
+            type: "number",
+        },
+    },
+    required: ["rowData", "gridTemplateColumns", "gridRowTop", "rowIndex"],
```
janechu, line 25: I don't see gridRowTop enumerated in the properties

packages/react/fast-components-react-base/src/data-grid/data-grid-row.schema.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Complies with FAST Tooling 1.0
```
janechu, line 2: This looks the same as 2.0, you can just use one schema you dont need two unless you have to specify react children which is the only breaking change between 1 and 2.

packages/react/fast-components-react-base/src/data-grid/data-grid-row.tsx
```diff
@@ -0,0 +1,120 @@
...
+        const rowKey: React.ReactText = !isNil(
+            this.props.rowData[this.context.dataGridProps.dataRowKey]
+        )
+            ? this.props.rowData[this.context.dataGridProps.dataRowKey]
+            : this.props.rowIndex;
+
+        // Check if there is already a style object being passed as props
+        const styleProps: React.CSSProperties = get(this.props, "style");
+
+        /* tslint:disable:no-string-literal */
```
janechu, line 52: we are an eslint now, you can remove this and any other tslint comments

packages/react/fast-components-react-base/src/data-grid/data-grid.props.ts
```diff
@@ -0,0 +1,122 @@
...
+    columnWidth: string;
+
+    /**
+     *  Custom render function for the header cell of the column
+     */
+    header?: (
+        title: React.ReactNode,
+        key: React.ReactText,
+        columnIndex: number,
+        className: string
```
janechu, line 36: could this be an object for extensibility?

packages/react/fast-components-react-base/src/data-grid/data-grid.props.ts
```diff
@@ -0,0 +1,122 @@
...
+     * set the 'ref' attribute of the internal element that should get focus to this value.
+     * unhandledProps: authors will generally want to write these to their custom implementations
+     */
+    cell?: (
+        props: DataGridCellProps,
+        className: string,
+        cellId: React.ReactText,
+        rootElement: React.RefObject<any>,
+        focusTarget: React.RefObject<any>,
+        unhandledProps: object
```
janechu, line 55: also here, could this be an object?

packages/react/fast-components-react-base/src/data-grid/data-grid.props.ts
```diff
@@ -0,0 +1,122 @@
...
+    itemHeight?: number;
+
+    /**
+     * This callback function overrides the itemHeight prop and will be called for each
+     * row of data when the gridData is updated in props.  Allows for non-uniform row heights.
+     */
+    itemHeightCallback?: (
+        rowData: object,
+        rowIndex: number,
+        defaultItemHeight: number
```
janechu, line 100: also here, could this be an object?

packages/react/fast-components-react-base/src/data-grid/data-grid.schema.2.ts
```diff
@@ -0,0 +1,50 @@
...
+    $schema: "http://json-schema.org/schema#",
+    title: "Data grid",
+    description: "A data grid component's schema definition.",
+    type: "object",
+    id: "@microsoft/fast-components-react-base/data-grid",
+    formPluginId: "@microsoft/fast-components-react-base/data-grid",
+    properties: {
+        gridData: {
+            title: "Array of data (objects by row) to be rendered",
+            type: "array",
```
janechu, line 14: you need to enumerate the items in the array in order to add them

packages/react/fast-components-react-base/src/data-grid/data-grid.schema.2.ts
```diff
@@ -0,0 +1,50 @@
...
+            title: "Virtualize items",
+            type: "boolean",
+        },
+        dataRowKey: {
+            title: "Data row key",
+            type: "string",
+        },
+        columnDefinitions: {
+            title: "Array of columndefinitions",
+            type: "array",
```
janechu, line 34: also here, in order for the UI to add items to the array they must be enumerated

packages/react/fast-components-react-base/src/data-grid/data-grid.schema.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Complies with FAST Tooling 1.0
```
janechu, line 2: looks like there are no react children, you can consolidate these and just use one schema since there are no differences.

packages/react/fast-components-react-base/src/data-grid/index.ts
```diff
@@ -0,0 +1,9 @@
+import DataGrid from "./data-grid";
+// import dataGridCellSchema from "./data-grid-cell.schema";
+// import dataGridRowSchema from "./data-grid-row.schema";
+// import dataGridSchema from "./data-grid.schema";
```
janechu, line 4: Why are these commented out?

packages/react/fast-components-react-base/src/data-grid/data-grid.stories.tsx
```diff
@@ -0,0 +1,277 @@
...
+    rootElement: React.RefObject<any>,
+    focusTarget: React.RefObject<any>,
+    unhandledProps: object
+): React.ReactNode {
+    return (
+        <div
+            {...unhandledProps}
+            data-cellid={cellId}
+            className={className}
+            ref={rootElement}
```
janechu, line 24: it seems like the developer is being asked to do a lot of work to implement this, can you add a helper wrapper that can be used?

packages/react/fast-components-react-base/src/data-grid/data-grid.tsx
```diff
@@ -0,0 +1,1103 @@
...
+                                this.props.managedClasses,
+                                "dataGrid_scrollingPanel_items",
+                                ""
+                            ),
+                            stackPanel__scrollable: get(
+                                this.props.managedClasses,
+                                "dataGrid_scrollingPanel__scrollable",
+                                ""
+                            ),
+                        }}
```
janechu, line 237: Do you need to use `get` here?

packages/react/fast-components-react-base/src/data-grid/data-grid.tsx
```diff
@@ -0,0 +1,1103 @@
...
+                } else {
+                    focusRowIndex = newFocusRowIndex;
+                }
+
+                // ensure data page conforms to new data length
+                currentDataPageStartIndex =
+                    focusRowIndex - Math.floor(this.props.pageSize / 2);
+
+                currentDataPageStartIndex =
+                    currentDataPageStartIndex < 0 ? 0 : currentDataPageStartIndex;
```
janechu, line 350: seems like you can do something like `currentDataPageStartIndex = Math.max(0, focusRowIndex - Math.floor(this.props.pageSize/2);` instead of repeatedly checking to see if the number is less than 0.

packages/react/fast-components-react-base/src/data-grid/data-grid.tsx
```diff
@@ -0,0 +1,1103 @@
...
+
+                currentDataPageEndIndex = currentDataPageStartIndex + this.props.pageSize;
+
+                currentDataPageEndIndex =
+                    currentDataPageEndIndex > this.props.gridData.length - 1
+                        ? this.props.gridData.length - 1
+                        : currentDataPageEndIndex;
```
janechu, line 357: same situation here, seems like you can use `Math.max`

packages/react/fast-components-react-base/src/data-grid/data-grid.tsx
```diff
@@ -0,0 +1,1103 @@
...
+        }
+    }
+
+    /**
+     * Generates class names
+     */
+    protected generateClassNames(): string {
+        const { dataGrid }: DataGridClassNameContract = this.props.managedClasses;
+
+        return super.generateClassNames(classNames(dataGrid));
```
janechu, line 408: don't need to use classNames in this instance since you only ever return dataGrid

packages/react/fast-components-react-base/src/data-grid/README.md
```diff
@@ -0,0 +1,59 @@
+# Data grid
+The *data grid* component enables a display of tabular data in a scrolling grid with fixed column headers. 
+
+## Usage
+
+Authors provide the data the component will display through the "gridData" prop.  This data must take the form of an array of objects, each object in the array corresponds to one row of data in the grid.  Additionally, each object in the array must have a property which uniquely identifies it that is passed to the component through the 'dataRowKey' prop. 
+
+A simple data set using a "clientId" attribute as a unique identifier  could look like this.
+
+myData = [
```
nicholasrice, line 10: This should probably be code-formatted. Same for the reset of the code examples that are not code formatted.

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.props.ts
```diff
@@ -0,0 +1,30 @@
...
+export const CellIdKey: string = "data-cellid";
+
+export type DataGridCellManagedClasses = ManagedClasses<DataGridCellClassNameContract>;
+export type DataGridCellUnhandledProps = React.AllHTMLAttributes<HTMLElement>;
+
+export interface DataGridCellHandledProps extends DataGridCellManagedClasses {
+    /**
+     *  Data for this row
+     */
+    rowData: object;
```
nicholasrice, line 17: `object` isn't a very specific data structure - is there a reason this is so vague?

scomea: This is the actual user provided data object for each row, it could be almost anything .  The column definition specifies the name of field in the object to display in each cell.  So rowData could be {customerId: 1234, name: John Doe, age: 34, etc...} and the column definition would specify which items to display in each cell in the row.

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.tsx
scomea: I like the config approach but I think we should keep on passing the whole row data to the cell because someone could create custom cells that render based on multiple data points, for example "display 'paymentAmount' field and highlight if 'dueDate' field is passed.

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.tsx
scomea: Well it does pass the class names so "unstyled" doesn't seem right (I took out "unstyled from the comment").  There won't be a "styled" version of this is the MSFT branch.  How about "renderCell"?  

As for the public bit - we've had cases where we have had similar customizable render functions (ie. select) which partners have used because they wanted to decorate ui elements for things such as generating telemetry and ended up duplicating the entire default render function to do that.  By having the default renderer public they can create a custom renderer that can optionally call the default one to create a cell node that they can modify.  That way they don't have to copy the base function and pick up changes without needing to recopy.  I'm also changing it to a public static function.

packages/react/fast-components-class-name-contracts-base/src/data-grid/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * The rows of the data grid component
+     */
+    dataGrid_row?: string;
+
+    /**
+     * Applied when focus is in a particular row
+     */
+    dataGrid_row__focusWithin?: string;
```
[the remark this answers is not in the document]
scomea: Yep

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.schema.2.ts
scomea: Only needed one schema since there are no children.

packages/react/fast-components-react-base/src/data-grid/data-grid-row.schema.2.ts
scomea: removed.

packages/react/fast-components-react-base/src/data-grid/data-grid-row.schema.ts
scomea: now with only one schema

scomea:
I've realized there is a behavior missing in data grid that I think we need which is when virtualization is turned off we also fall back to using basic html/css driven layout.  Currently row heights and placement are still controlled by the grid code in this state.  Figuring out best way to do this and debating whether it would be better to have the underlying stackPanel do this or not use stackpanel when the grid virtualization is off.

packages/react/fast-components-class-name-contracts-base/src/data-grid/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * A cell within the data grid component
+     */
+    dataGrid_cell?: string;
+}
+
+export interface DataGridRowClassNameContract {
+    dataGridRow?: string;
+    dataGridRow__focusWithin?: string;
```
janechu, line 53: Seems like this should also be changed to `focusedWithin`

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.props.ts
nicholasrice: Maybe this should be a generic then? This seems really permissive: `T extends {[key: string]: any}`

scomea: Good suggestion.  I'll see about doing something like that.

packages/react/fast-components-react-base/src/data-grid/index.ts
scomea: uncommented

packages/react/fast-components-react-base/src/data-grid/data-grid.stories.tsx
scomea: I reduced the number of required props and the component will autogenerate basic column definitions if not provided

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.props.ts
scomea: Unfortunately was not able to figure out a good way to contstrain the type beyond object.

packages/react/fast-components-react-base/src/data-grid/data-grid.tsx
```diff
@@ -0,0 +1,1375 @@
...
+        }
+
+        let newFocusColumnIndex: number =
+            currentFocusColumnIndex + direction * directionMod;
+
+        if (newFocusColumnIndex < 0) {
+            newFocusColumnIndex = 0;
+        } else if (newFocusColumnIndex >= this.state.columns.length) {
+            newFocusColumnIndex = this.state.columns.length - 1;
+        }
```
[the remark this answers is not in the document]
scomea: yep

packages/react/fast-components-react-base/src/data-grid/data-grid-cell.props.ts
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import {
+    DataGridCellClassNameContract,
+    ManagedClasses,
+} from "@microsoft/fast-components-class-name-contracts-base";
+import { DataGridColumn } from "./data-grid.props";
+
+export const CellIdKey: string = "data-cellid";
```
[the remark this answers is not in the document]
scomea: removed

scomea:
Question - Do people think the default should be virtualization on or off?  I can think of arguments for both.

nicholasrice:
@scomea good question. I can't say I have have a strong opinion on that. Is this component design for any arbitrary grid? Or is it more tailored toward *large* grids?

chrisdholt:
> Question - Do people think the default should be virtualization on or off? I can think of arguments for both.

@scomea my initial bent is to have it off by default. I prefer the opt-in approach to this as it's a specific optimization that is best utilized with large data sets. It's possible that the extra overhead of turning it on when unnecessary could have a slightly negative performance impact. With that possibility I would think opt-in makes the most sense.

scomea:
Ok, went with making virtualization off by default and updated accordingly.

PR #3161, opened 2020-05-19
bug fix: update package.json

radium-v:
# Description

Updates `package.json` to point to the latest versions in the workspace.

## Motivation & context

Fixes package linking for the `fast-website` project.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/package.json
+++ b/sites/fast-website/package.json
@@ -6,7 +6,10 @@
   "private": true,
   "license": "MIT",
   "devDependencies": {
-    "@microsoft/fast-components": "^0.10.1",
+    "@microsoft/fast-element": "0.9.0",
+    "@microsoft/fast-foundation": "1.0.0",
+    "@microsoft/fast-web-utilities": "4.5.0",
+    "@microsoft/fast-components": "1.0.0",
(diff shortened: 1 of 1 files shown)


sites/fast-website/package.json
```diff
@@ -6,7 +6,10 @@
   "private": true,
   "license": "MIT",
   "devDependencies": {
-    "@microsoft/fast-components": "^0.10.1",
+    "@microsoft/fast-element": "0.9.0",
```
chrisdholt, line 9: I think we’ll ultimately want to opt into semver. If the publish again we may have the same issue but we’ll definitely want to let lerna handle this once it’s in master.

chrisdholt: Is there a reason we are moving to dev deps here?

eljefe223: I believe these should just stay as dep

radium-v: @eljefe223 @chrisdholt Installing them as `dependencies` would leave them hanging out in `node_modules` for no real reason on production, since the packages are only needed for building. Once the `dist` directory is built, there's no need for any other packages in order to run the site.

PR #3164, opened 2020-05-19
chore: add responsive navigation to marketing site

khamudom:
# Description
I added media queries to address the responsiveness of the page.

## Motivation & context
The main focus is the navigation in the header and the footer. In addition to fixing responsive, I put the links inside ul and li tags. I also set a breakpoint at 1330px to match the width of the card section, when they are wrapping. Added icons for the FAST brand, and menu.
Remove the start and end template styles, from the component, to have the flexibility to style them at the implementation level.

The collapsed navigation animates down from the top. 
collapse-nav

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/svg/icon-menu.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="35" height="19" viewBox="0 0 35 19" xmlns="http://www.w3.org/2000/svg">
```
radium-v, line 1:
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" class="icon icon-menu" width="35" height="19" viewBox="0 0 35 19">
```

sites/fast-website/src/public/index.html
```diff
@@ -14,20 +14,26 @@
...
+            <fast-anchor class="header-github" slot="end" appearance="lightweight" href="#">
               Github
-              <svg width="19" height="20" viewBox="0 0 19 20" xmlns="http://www.w3.org/2000/svg">
-                <path d="M6.71154 17.0776C2.09615 18.4906 2.09615 14.7226 0.25 14.2517L6.71154 17.0776ZM13.1731 19.9036V16.2581C13.2077 15.8089 13.1482 15.3574 12.9986 14.9335C12.849 14.5096 12.6127 14.123 12.3054 13.7995C15.2038 13.4698 18.25 12.3489 18.25 7.20563C18.2498 5.89046 17.754 4.62572 16.8654 3.67319C17.2862 2.52257 17.2564 1.25074 16.7823 0.121918C16.7823 0.121918 15.6931 -0.207776 13.1731 1.51605C11.0574 0.930913 8.82722 0.930913 6.71154 1.51605C4.19154 -0.207776 3.10231 0.121918 3.10231 0.121918C2.62819 1.25074 2.59844 2.52257 3.01923 3.67319C2.12396 4.63279 1.62771 5.90895 1.63462 7.23389C1.63462 12.3394 4.68077 13.4604 7.57923 13.8278C7.27554 14.148 7.04132 14.5299 6.89182 14.9486C6.74233 15.3674 6.6809 15.8135 6.71154 16.2581V19.9036" />
-              </svg>
+              <%= require("svg/icon-github.svg") %>
             </fast-anchor>
-        </site-navigation>
+          </site-navigation>
+          <div class="header-spacer"/>
```
radium-v, line 32: Instead of a dummy `<div>`: style the container (`.header`) to have the style of the bar, push the navigation out beyond the top edge, and use padding to account for the overlap.

sites/fast-website/src/app/css/style.css
```diff
@@ -70,9 +89,100 @@ body {
...
+  }
+
+  .header-navigation ul {
+    flex-direction: column;
+    margin-top: 20px;
+  }
+
+  .header-navigation__open {
+    transform: translateY(0%);
+    height: 100%;
```
radium-v, line 178: `height` should always be set to 100%. Without it, the menu collapses right before transitioning out of view. Increase the `transition` duration in `.header-navigation` to investigate it more closely.

sites/fast-website/src/app/svg/icon-brand.svg
```diff
@@ -0,0 +1,26 @@
+<svg width="114" height="40" viewBox="0 0 114 40" fill="none" xmlns="http://www.w3.org/2000/svg">
```
radium-v, line 1: We may need to optimize this SVG a little bit due to all of the IDs hanging out. If we're using it in more than one place, it could cause conflicts.

sites/fast-website/src/public/index.html
```diff
@@ -162,23 +168,33 @@ <h4>Other</h4>
...
+              </ul>
             </site-navigation>
+            <fast-anchor class="footer-brand" appearance="lightweight" href="#">
+              <%= require("svg/icon-brand.svg") %>
+            </fast-anchor>
           </div>
         <!-- footer:end -->
       </div>
     </fast-design-system-provider>
+    <script>
```
radium-v, line 193: We should probably make the hamburger button its own fast element-based component to move this function into a module.

sites/fast-website/src/app/css/style.css
khamudom: height set

sites/fast-website/src/public/index.html
khamudom: the button is moved into the component

sites/fast-website/src/app/svg/icon-brand.svg
khamudom: I will check with Jason R, to get me a better-optimized SVG

sites/fast-website/src/public/index.html
khamudom: dummy div removed

sites/fast-website/src/app/components/navigation/navigation.ts
```diff
@@ -1,7 +1,35 @@
...
+        if (!e.matches) {
+            this.opened = false;
+        }
+    };
+
+    constructor() {
+        super();
+
+        this.mediaQueryList = window.matchMedia("screen and (max-width: 800px)");
+        this.mediaQueryList.addListener(this.mqlListener);
```
nicholasrice, line 25: Should the listener cb get fired immediately if the view is loaded and it matches the matchMedia listener? If so, you might want to invoke your cb immediately because I don't think the attached listener will get fired in that case.

radium-v: @nicholasrice Since `debounce` is true by default and the component is instantiated with that value, the class would be persistent until the `MediaQueryListEvent` is triggered. Also I think triggering the nav button's click handler wouldn't be able to happen before the constructor finishes executing, but I'm not 100% on that.

I think the only way to guarantee that a `MediaQueryListEvent` happens after it's created is to force a window resize. But I'm confident that we don't need to do that here.

PR #3167, opened 2020-05-20
fix: bug in positioning when negative numbers were involved

marjonlynch:
# Description
Existing value and positioning calculations would break when negative values were used for positioning, min, max.

* Unified the percentage calc for label positioning and thumb positioning
* Made the calculations work when min or position values are negative numbers

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/slider/fixtures/base.html
+++ b/packages/web-components/fast-components/src/slider/fixtures/base.html
@@ -33,6 +33,50 @@
         <!-- Margin offsets -->
         <fast-slider style="margin-left:100px"></fast-slider>
 
+        <!-- Negative positions -->
+        <fast-slider
+            min="-3"
+            max="3"
+            step="1"
+        >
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-components/src/slider/fixtures/base.html
```diff
@@ -33,6 +33,50 @@
         <!-- Margin offsets -->
         <fast-slider style="margin-left:100px"></fast-slider>
 
+        <!-- Negative positions -->
```
chrisdholt, line 36: It could be done in a separate PR, but I believe in many of the test templates we have headings. Perhaps that would help visually break this up and provide more context than just the comments?

marjonlynch: ok will do, I also think now there are some labels not using the design system so in fast-components they are dark text on default dark mode. Will fix both in a separate PR

PR #3169, opened 2020-05-20
fix: children components not always added to the internal slotted radio control collection

marjonlynch:
# Description
Some child elements in the default slot were getting filtered out of the slottedRadios collection thus breaking the group related functionality, also there was no check to see there was a collection and at least one item in that collection before setting focus to the first element.

* now checking there is at least one element in the radio collection before setting default focus
* replaced HTMLInputElement with any to avoid compilation error and to allow more component types to be in the collection

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/radio-group/radio-group.ts
+++ b/packages/web-components/fast-foundation/src/radio-group/radio-group.ts
@@ -95,7 +95,7 @@ export class RadioGroup extends FASTElement {
             }
         });
 
-        if (this.value === undefined) {
+        if (this.value === undefined && radioButtons && radioButtons.length > 0) {
             radioButtons[0].setAttribute("tabindex", "0");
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-foundation/src/radio-group/radio-group.ts
```diff
@@ -95,7 +95,7 @@ export class RadioGroup extends FASTElement {
             }
         });
 
-        if (this.value === undefined) {
+        if (this.value === undefined && radioButtons && radioButtons.length > 0) {
```
janechu, line 98: since radioButtons will always be at least an empty array, I think you can skip the radioButtons check and just look at the length

marjonlynch: done and done

PR #3175, opened 2020-05-21
chore: update to add actions to deploy sites

awentzel:
# Description
This adds the ability to deploy Component Explorer, Color Explorer, and Creator to Azure using the publish profile strategy, while also updating CDN for any assets they want to use placed into `sites/statics/assets` which also goes for marketing site.  These are image type assets not webpack generated file names, these will be automatically cached by Front Door.

The sites require a node server with node_modules and so additional work was created to support this need from a central location of `/sites/statics/` whereby two files were added, `server.js` and `package.json` in order to install the node_modules.  

Additionally, the CDN requires a `/sites/statics/assets` folder where all infrequently updated assets will live (images, fonts, etc).  This PR include a pipeline for the complete end-to-end process.

I chose this strategy as opposed to adding these packages to each of our different sites to eliminate code duplication in multiple (potentially 6) web sites which would increase technical debt and maintainability effort.

There may be alternative approaches and performance improvements.  Please create proposals for that work.  This is only the initial work to get a full end-to-end release cycle working.  These sites should be tested.

- http://color-west-app-stage.azurewebsites.net
- http://create-west-app-stage.azurewebsites.net
- http://explore-west-app-stage.azurewebsites.net
- http://color-east-app-stage.azurewebsites.net
- http://create-east-app-stage.azurewebsites.net
- http://explore-east-app-stage.azurewebsites.net

After this merges into master, and the sites are approved by their respective owners we will need to swap the staging site into production, update Front Door routes, and update DNS to point to these new production sites.

- http://color.fast.design
- http://create.fast.design
- http://explore.fast.design

## Motivation & context
Wire up deployment to new resilient infrastructure on Azure.

## Issue type checklist

## Process & policy checklist


1
```diff
@@ -0,0 +1,16 @@
+update to create required server.js file and dist it
```
radium-v, line 1: extra file?

sites/statics/server.js
```diff
@@ -0,0 +1,18 @@
...
+var publicDir = path.resolve(__dirname);
+
+// Set static application options
+app.use("/", express.static(publicDir));
+
+// Set fallback application options
+app.use(fallback("index.html", { root: publicDir }));
+
+// Serve up application on specified port
+app.listen(process.env.PORT);
```
radium-v, line 18: From the node docs:
> If port is omitted or is 0, the operating system will assign an arbitrary unused port

we can pass a port with an environment variable or modify this script to set a default:

```
const PORT = process.env.PORT || 3000;
```

awentzel: It doesn't matter once in production what port it runs on.  Rather than hard-coding and/or setting in-app settings, I wonder if we just let the operating system pick.  this eliminates a configuration requirement currently, someone forgetting to set it.  I'm in favor letting OS pick.  What do you think.

sites/statics/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
EisenbergEffect, line 1: Is this intended to be published to npm? If not, you may want to make it a private package.

awentzel: Oops, forgot that step. Definitely.

.github/workflows/cd-explore-stage.yml
```diff
@@ -6,41 +6,47 @@ on:
...
       run: yarn install --frozen-lockfile
     
-    - name: Build Web Application
-      run: | 
-        cd sites/fast-component-explorer
+    - name: Build & Prepare Web Application
+      run: |
+        cd ${{ env.AZURE_WEBAPP_BUILD_PATH }}
         yarn build:app
+        cp ../statics/package.json ../statics/server.js www
```
chrisdholt, line 33: I'm not sure we should have a single package.json for these and a single server.js for all sites. I'm specifically concerned with a single package.json file as the requirements across sites are different. Some require react and others may not.

awentzel: @chrisdholt @janechu The package.json file is required to be distributed so I can install node server packages during the packaging process.  What do you suggest?  Would you rather include, this static approach within each site itself?  I would then include the commands in the scripts of the root package.json to perform install and copy during the build.  @radium-v also as we collaborated quite a bit on different solutions and he's aware of the scenario I'm covering.

sites/statics/server.js
awentzel: @chrisdholt what pros/cons do you see with allowing Node to choose the port in production?  It feels all upside for our production scenarios.

.github/workflows/cd-explore-stage.yml
janechu: I think it's worth asking the question, will this work for a web app that controls its own routing like a React site and will it work for a static site that has `index.html` files in various folders as its routing mechanism. If it works for both then this might not be an issue.

awentzel: @janechu if you can provide more info I can answer your question. What are these sites running, please fill in [React | Static | ...] for each:

- color
- create
- explore
- app
- www
- motion

Thanks,

janechu: I can answer a few:
- create [react]
- explore [react]
- tooling examples [static]
- fast.design [static]

sites/statics/server.js
radium-v: Predictability and reproducibility are the main reasons to set a port. Once each package has its own `server.js` it will also allow us to use it for local testing (and will be really important when we move towards k8s, where it's proper to set the ports explicitly)

.github/workflows/cd-explore-stage.yml
awentzel: @janechu react and static sites can all run on node express server.

sites/statics/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "These packages are required to run node server on Azure\"",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/microsoft/fast-dna.git"
+  },
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "private": true,
+  "dependencies": {
+    "path": "^0.12.7",
```
nicholasrice, line 13: You shouldn't need to enumerate "path" as a dependency. The `path` module referenced in `server.js` is a built-in module: https://nodejs.org/api/path.html

sites/statics/server.js
```diff
@@ -0,0 +1,19 @@
+var path = require("path");
```
janechu, line 1: This file may be better in the `site-utilities` folder which is meant to serve as shared assets between sites.

awentzel: Per discussion with @janechu, moving my statics into `site-utilities`

PR #3180, opened 2020-05-22
chore: update single column and hover styles on feature cards

freefaller69:
# Description

Update single column styling on feature cards.
Update styles to match hover behavior changes in design on feature cards.

Screen Shot 2020-05-22 at 9 52 02 AM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/card-section/card-section.styles.ts
+++ b/sites/fast-website/src/app/components/card-section/card-section.styles.ts
@@ -1,6 +1,7 @@
 import { css } from "@microsoft/fast-element";
 import { display } from "@microsoft/fast-foundation";
 import { siteBreakpoints } from "../../style/constants";
+import { neutralFillHoverBehavior } from "@microsoft/fast-components";
 
 export const CardSectionStyles = css`
     :host {
@@ -29,4 +30,4 @@ export const CardSectionStyles = css`
         background-color: var(--neutral-fill-hover);
         filter: saturate(1);
     }
-`;
+`.withBehaviors(neutralFillHoverBehavior);
--- a/sites/fast-website/src/app/components/card-section/card-section.styles.ts
+++ b/sites/fast-website/src/app/components/card-section/card-section.styles.ts
@@ -1,12 +1,13 @@
 import { css } from "@microsoft/fast-element";
 import { display } from "@microsoft/fast-foundation";
+import { siteBreakpoints } from "../../style/constants";
 
 export const CardSectionStyles = css`
     :host {
         --flow: column;
     }
 
-    @media screen and (max-width: 900px) {
+    @media screen and (max-width: ${siteBreakpoints[2]}) {
         :host {
             --flow: row;
         }
--- a/sites/fast-website/src/app/components/card-section/card-section.styles.ts
+++ b/sites/fast-website/src/app/components/card-section/card-section.styles.ts
@@ -26,7 +26,7 @@ export const CardSectionStyles = css`
     :host ::slotted(site-feature-card:hover) {
         cursor: pointer;
         color: currentColor;
-        background-color: rgba(27, 27, 27, 1);
+        background-color: var(--neutral-fill-hover);
         filter: saturate(1);
     }
 `;
--- a/sites/fast-website/src/app/components/card-section/card-section.styles.ts
+++ b/sites/fast-website/src/app/components/card-section/card-section.styles.ts
@@ -1,6 +1,5 @@
 import { css } from "@microsoft/fast-element";
 import { display } from "@microsoft/fast-foundation";
-import { siteBreakpoints } from "../../style/constants";
 import { neutralFillHoverBehavior } from "@microsoft/fast-components";
 
 export const CardSectionStyles = css`
@@ -8,7 +7,7 @@ export const CardSectionStyles = css`
         --flow: column;
     }
 
-    @media screen and (max-width: ${siteBreakpoints[2]}) {
+    @media screen and (max-width: 900px) {
         :host {
             --flow: row;
         }
--- a/sites/fast-website/src/app/components/card-section/card-section.styles.ts
+++ b/sites/fast-website/src/app/components/card-section/card-section.styles.ts
@@ -6,7 +6,7 @@ export const CardSectionStyles = css`
         --flow: column;
     }
 
-    @media screen and (max-width: 1000px) {
+    @media screen and (max-width: 900px) {
         :host {
             --flow: row;
         }
@@ -15,13 +15,17 @@ export const CardSectionStyles = css`
     ${display("grid")} :host {
         grid-template-rows: repeat(4, min-content);
         grid-auto-flow: var(--flow);
+        justify-content: center;
     }
 
     :host(:hover) ::slotted(site-feature-card) {
-        filter: blur(1px);
+        filter: saturate(0);
     }
 
     :host ::slotted(site-feature-card:hover) {
-        filter: blur(0);
+        cursor: pointer;
+        color: currentColor;
+        background-color: rgba(27, 27, 27, 1);
+        filter: saturate(1);
     }
 `;
--- a/sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
+++ b/sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
@@ -7,10 +7,10 @@ import {
     neutralOutlineRestBehavior,
     neutralFillActiveBehavior,
 } from "@microsoft/fast-components";
-import { siteBreakpoints } from "../../style/constants";
 
 export const ContentPlacementContainerStyles = css`
     ${display("block")}:host {
+        --flow: column;
         --border-width: calc(var(--outline-width) * 1px);
         --border: var(--border-width) solid var(--neutral-outline-rest);
         contain: none;
@@ -41,17 +41,19 @@ export const ContentPlacementContainerStyles = css`
         border-bottom: none;
     }
 
-    @media screen and (min-width: ${siteBreakpoints[1]}) {
+    @media screen and (min-width: 700px) {
         .framework_ContentPlacement {
             border-width: 0 var(--border-width) var(--border-width) 0;
         }
     }
 
-    @media screen and (min-width: ${siteBreakpoints[1]}) and (max-width: ${siteBreakpoints[3]}) {
+    @media screen and (max-width: 900px) {
         :host {
             --flow: row;
         }
+    }
 
+    @media screen and (min-width: 700px) and (max-width: 999px) {
         .framework_container {
             grid-template-columns: repeat(2, 1fr);
         }
@@ -70,11 +72,7 @@ export const ContentPlacementContainerStyles = css`
         }
     }
 
-    @media screen and (min-width: ${siteBreakpoints[4]}) {
-        :host {
-            --flow: column;
-        }
-
+    @media screen and (min-width: 1000px) {
         .framework_container,
         .community_container {
             grid-template-columns: repeat(3, 1fr);
--- a/sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
+++ b/sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
@@ -7,6 +7,7 @@ import {
     neutralOutlineRestBehavior,
     neutralFillActiveBehavior,
 } from "@microsoft/fast-components";
+import { siteBreakpoints } from "../../style/constants";
 
 export const ContentPlacementContainerStyles = css`
     ${display("block")}:host {
@@ -40,17 +41,13 @@ export const ContentPlacementContainerStyles = css`
         border-bottom: none;
     }
 
-    @media screen and (min-width: 700px) {
+    @media screen and (min-width: ${siteBreakpoints[1]}) {
         .framework_ContentPlacement {
             border-width: 0 var(--border-width) var(--border-width) 0;
         }
     }
 
-    @media screen and (min-width: 700px) and (max-width: 999px) {
-        :host {
-            --flow: row;
-        }
-
+    @media screen and (min-width: ${siteBreakpoints[1]}) and (max-width: ${siteBreakpoints[3]}) {
         .framework_container {
             grid-template-columns: repeat(2, 1fr);
         }
@@ -69,11 +66,7 @@ export const ContentPlacementContainerStyles = css`
         }
     }
 
-    @media screen and (min-width: 1000px) {
-        :host {
-            --flow: column;
-        }
-
+    @media screen and (min-width: ${siteBreakpoints[4]}) {
         .framework_container,
         .community_container {
             grid-template-columns: repeat(3, 1fr);
--- a/sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
+++ b/sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
@@ -48,6 +48,10 @@ export const ContentPlacementContainerStyles = css`
     }
 
     @media screen and (min-width: ${siteBreakpoints[1]}) and (max-width: ${siteBreakpoints[3]}) {
+        :host {
+            --flow: row;
+        }
+
         .framework_container {
             grid-template-columns: repeat(2, 1fr);
         }
@@ -67,6 +71,10 @@ export const ContentPlacementContainerStyles = css`
     }
 
     @media screen and (min-width: ${siteBreakpoints[4]}) {
+        :host {
+            --flow: column;
+        }
+
         .framework_container,
         .community_container {
             grid-template-columns: repeat(3, 1fr);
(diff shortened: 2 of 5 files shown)


sites/fast-website/src/app/components/card-section/card-section.styles.ts
```diff
@@ -6,7 +6,7 @@ export const CardSectionStyles = css`
         --flow: column;
     }
 
-    @media screen and (max-width: 1000px) {
+    @media screen and (max-width: 900px) {
```
eljefe223, line 9: It might good to start storing these breakpoints in constants file so we can use them elsewhere. We could follow a similar model that we do with the react version of the `BreakpointTracker` which is store and array of numbers and then reference them throughout the app. This case we can just use `string[]`.

```
export siteBreakpoints: string[] = [0, 900px, 1200px];
```

```
@media screen and (max-width: siteBreakpoints[1]) {}
```

freefaller69: That sounds great.  I'll get on that.  Thanks Jeff.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
```diff
@@ -15,13 +15,17 @@ export const CardSectionStyles = css`
...
     :host(:hover) ::slotted(site-feature-card) {
-        filter: blur(1px);
+        filter: saturate(0);
     }
 
     :host ::slotted(site-feature-card:hover) {
-        filter: blur(0);
+        cursor: pointer;
+        color: currentColor;
+        background-color: rgba(27, 27, 27, 1);
```
SamanthaAO, line 28: should we be using a design-system color here?

freefaller69: I couldn't find a design system option that matched the black in the Figma design, but I am happy to change it if one can be recommended.  The closest I found was more of a charcoal grey which had more of a washed out look than a look of focus.

SamanthaAO: yeah I am on the same page with you. I am just unsure if the priority is using a color in the design system or a color that is more closely aligned with the figma.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
freefaller69: Done

sites/fast-website/src/app/components/card-section/card-section.styles.ts
```diff
@@ -1,12 +1,13 @@
...
 import { display } from "@microsoft/fast-foundation";
+import { siteBreakpoints } from "../../style/constants";
 
 export const CardSectionStyles = css`
     :host {
         --flow: column;
     }
 
-    @media screen and (max-width: 1000px) {
+    @media screen and (max-width: ${siteBreakpoints[2]}) {
```
EisenbergEffect, line 10: I understand the thinking behind having the breakpoints in an array and accessing it. However, I think the readability suffers a bit. I wonder if there's an alternative way to represent them that creates greater clarity at the point where they are used. Not a huge deal, but something worth thinking about.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
EisenbergEffect: Let's definitely follow up on that to find out. Even if it's a unique color not in the core design system, I think it would make sense to extract this into a CSS variable.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
```diff
@@ -26,13 +27,13 @@ export const ContentPlacementContainerStyles = css`
         border-bottom: none;
     }
 
-    @media screen and (min-width: 700px) {
+    @media screen and (min-width: ${siteBreakpoints[1]}) {
```
EisenbergEffect, line 30: Readability: see comments above above.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
```diff
@@ -26,13 +27,13 @@ export const ContentPlacementContainerStyles = css`
...
         .framework_ContentPlacement {
             border-width: 0 var(--border-width) var(--border-width) 0;
         }
     }
 
-    @media screen and (min-width: 700px) and (max-width: 999px) {
+    @media screen and (min-width: ${siteBreakpoints[1]}) and (max-width: ${siteBreakpoints[3]}) {
```
EisenbergEffect, line 36: Readability: see comments above above.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
```diff
@@ -47,7 +48,7 @@ export const ContentPlacementContainerStyles = css`
         }
     }
 
-    @media screen and (min-width: 1000px) {
+    @media screen and (min-width: ${siteBreakpoints[4]}) {
```
EisenbergEffect, line 51: Readability: see comments above above.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -19,15 +20,11 @@ export const FeatureCardStyles = css`
         background-color: currentColor;
         position: fixed;
         height: 1px;
-        width: 93%;
+        width: calc(100% - (var(--design-unit) * 5px));
```
EisenbergEffect, line 23: Nice improvement here!

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
```diff
@@ -58,7 +46,7 @@ export const FeatureCardStyles = css`
     }
 
     ::slotted(h4) {
-        font-size: 20px;
+        font-size: var(--type-ramp-plus-2-font-size);
```
EisenbergEffect, line 49: Cool to see the type ramp being used!

sites/fast-website/src/app/style/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const siteBreakpoints: string[] = [
+    "0",
```
radium-v, line 2: The min breakpoint should probably be `320px` to indicate that it's the smallest size we want to support

sites/fast-website/src/app/style/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const siteBreakpoints: string[] = [
```
radium-v, line 1: The way I've usually done it was to give the sizes some specific names that correspond to the type of device:

```ts
export const tinyDeviceWidth: number = 320;

export const midDeviceWidth: number = 700;
export const midDeviceWidthMax: number = midDeviceWidth - 1; // upper bound for devices smaller than 700px
export const midDeviceWidthMin: number = midDeviceWidth + 1; // lower bound for devices larger than 700px

export const normalDeviceWidth: number = 1000;
export const normalDeviceWidthMax: number = normalDeviceWidth - 1; // upper bound for devices smaller than 1000px
export const normalDeviceWidthMin: number = normalDeviceWidth + 1; // lower bound for devices larger than 1000px

export const bigDeviceWidth: number = 1200;

// we could add extra named variables as needed without worry:
// export const weirdDeviceWidth: number = 672;
// export const hugeDeviceWidth: 1600;
```

sites/fast-website/src/app/components/card-section/card-section.styles.ts
```diff
@@ -1,12 +1,13 @@
 import { css } from "@microsoft/fast-element";
 import { display } from "@microsoft/fast-foundation";
+import { siteBreakpoints } from "../../style/constants";
```
radium-v, line 3: We can add an alias to `webpack.common.js` so we can easily reference this module. In `module.exports.resolve.alias`, add a line similar to the `svg` property. This will allow us to reference `style/constants` directly without having to do manual path finding.

Then, these can be imported with proper their names:

```ts
import {
    normalDeviceWidth, // see my other comment on constants.ts
    bigDeviceWidth
} from "style/constants";
```

sites/fast-website/src/app/components/card-section/card-section.styles.ts
chrisdholt: The question I tend to ask in these situations are, “what are the implications of ____?” In this case, if we set a static color, it will not theme with the rest of the page. That can create accessibility issues depending on how the page changes, etc. Thats just one that can be identified here without digging in, but I think it’s an important one. We should leverage the system here to ensure appropriate contrast. If we need a new design system provider to do so, we can do that. My guess here is that this is likely just a mistake or there is possibly a lack of awareness. When it comes to colors on the site, let’s be sure to leverage the system rather than hardcode colors.

freefaller69: Screen Shot 2020-05-26 at 8 40 21 AM
neutral-fill-hover

Screen Shot 2020-05-26 at 8 41 17 AM
rgba

I initially tried neutral-fill-hover which is the first image, but that does look notably different than the design.  Since the design system is still being implemented I opted to match the design in the interim.  I've changed the rgba to neutral-fill-hover.  If there is a design system variable better suited here, please let me know what that may be.

sites/fast-website/src/app/components/feature-card/feature-card.styles.ts
freefaller69: Thank you, Rob!  :)

sites/fast-website/src/app/components/card-section/card-section.styles.ts
eljefe223: We followed this pattern before in react with `BreakpointTracker`. I'm not saying it's the right way but I suggested doing it this way to follow that pattern we have used in the past.

sites/fast-website/src/app/style/constants.ts
freefaller69: This definitely makes sense and provides better readability.  What do you think of a siteBreakpoint object with each of these as properties vs individual exports and imports?

eljefe223: I think we should discuss the best way forward. In react we have used an array to represent the app breakpoints using `BreakpointTracker`. I think the problem with naming these is that the names are kind of meaningless and can get messy pretty quickly. @freefaller69 maybe post up in Discord to gather opinions on how we want to track breakpoints.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
freefaller69: That sounds like a great idea.  I haven't done much work directly in webpack and that was a while ago, but I'm game to give it a go.

sites/fast-website/src/app/style/constants.ts
freefaller69: Makes sense.  Easy enough to change.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
nicholasrice: You will also need to tell *TypeScript* about this alias. Historically we've avoided this as it seemed to cause more issues than it really solved, but if you think its helpful go for it - just note the above.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.template.ts
```diff
@@ -23,26 +23,28 @@ export const ContentPlacementContainerTemplate = html<ContentPlacementContainer>
...
-                        >
-                        <fast-anchor
-                            slot="footer"
-                            href=${x => x.documentationLink}
-                            appearance="lightweight"
-                            >Read Documentation</fast-anchor
-                        >
-                    </site-feature-card>`
-                )}
+                <site-card-section>
```
SamanthaAO, line 26: site-card-section should not need to be added here. Part of the purpose of moving feature-card here was to be able to use content-placement-container to style it. In my PR that you have rebased with I moved all the current styling of site-card-section into content-placement-container's styles. If there are updates to site-card-section they should be made in content-placement-container.style. It is redundant to apply site-card-section on top because content-placement-container should have the styles of site-card-section.

freefaller69: Without it there, features is a single column.

SamanthaAO: It looks as it should in feature/fast-frame, and feature/fast-frame does not use site-card-section. I am thinking that this is an issue with the rebase.

SamanthaAO: Yeah when looking through your code it looks like you deleted where the flow variables were defined.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
```diff
@@ -69,11 +66,7 @@ export const ContentPlacementContainerStyles = css`
         }
     }
 
-    @media screen and (min-width: 1000px) {
-        :host {
```
SamanthaAO, line 73: I believe deleting this flow's value assignment is what is causing your issue.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
```diff
@@ -69,11 +66,7 @@ export const ContentPlacementContainerStyles = css`
...
-            --flow: column;
```
SamanthaAO, line 74: I believe deleting this variable's value assignment(--flow) is what is causing your issue.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
```diff
@@ -40,17 +41,13 @@ export const ContentPlacementContainerStyles = css`
...
-    @media screen and (min-width: 700px) {
+    @media screen and (min-width: ${siteBreakpoints[1]}) {
         .framework_ContentPlacement {
             border-width: 0 var(--border-width) var(--border-width) 0;
         }
     }
 
-    @media screen and (min-width: 700px) and (max-width: 999px) {
-        :host {
-            --flow: row;
```
SamanthaAO, line 51: The assignment was deleted here as well.

freefaller69: Well that's just weird.  The only thing I actively changed in this file was the breakpoints.  I have no clue how those wound up getting removed.

freefaller69: Reverted

sites/fast-website/src/app/style/constants.ts
chrisdholt: This has been hanging out for awhile and it seems like we don't have full resolution on this. I think @radium-v is going to write a proposal for handling breakpoints in our new reality of W/C's and the modern web and I'd prefer to resolve this in that issue. For now, let's move ahead with *something* for this specific implementation so we aren't blocked on the ideal path forward for FAST in the future. We can add a TODO that links to the issue and that will help us follow-up once we land the "blessed" thing.

My primary concern here is that we land the responsiveness of this so we can move ahead with the site and not wait to land our overall principle for how we handle media queries overall as that convo could take a bit to resolve.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
freefaller69: Reverting this to fixed numbers for now.  See Chris' comment below.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
freefaller69: Reverting this to fixed numbers for now.  Proposal coming forth on this.  See Chris' comment below.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
freefaller69: Reverting this to fixed numbers for now.  Proposal coming forth on this.  See Chris' comment below.

sites/fast-website/src/app/components/content-placement-container/content-placement-container.styles.ts
freefaller69: Reverting this to fixed numbers for now.  Proposal coming forth on this.  See Chris' comment below.

sites/fast-website/src/app/style/constants.ts
freefaller69: Reverting this to fixed numbers for for the sake of readability with John's forthcoming proposal on a uniform reusable approach.

sites/fast-website/src/app/components/card-section/card-section.styles.ts
freefaller69: Reverting this to fixed numbers for now.  Proposal coming forth on this.  See Chris' comment below.

Issue #3039, opened 2020-04-29
[Proposal] Content Composition


EisenbergEffect:
## Description

Today, if you provide a binding in your element's content, you can only return a primitive value from your expression. For example:

```TypeScript
const template = html`
  <div>${x => x.name}</div>
`;
```

Or, if you have an existing template, you can compose it like this:

```TypeScript
const sharedTemplate = html`...`;
const template = html`
  <div>${sharedTemplate}</div>
`;
```

But, what if you want a binding expression to return a template? You can't do that because the text binding doesn't know how to handle anything except for primitives, and the composed template directive doesn't know how to handle expressions 😢 

I propose that we teach the "text binding" how to be a "content binding" so that if a template is returned from an expression, it will create a view, bind it to the data source, and then insert it at that location in the DOM. Furthermore, I propose that we extends the `repeat` and `when` directives so that they can receive not only an inline template, but an expression that returns a template as well. By combining these things, we can greatly expand the compositional capabilities of the template engine.

### Examples

Below are a few examples of what we can enable with the above described capabilities.

#### Dynamic Composition

```TypeScript
const defaultTemplate = html`...`;
const templatesByType = {
  foo: html`...`,
  bar: html`...`,
  baz: html`...`
};

const template = html<MyElement>`
  <div>${x => x.selectTemplate()}</div>
`;

@customElement(...)
export class MyElement extends FASTElement {
  @observable data;

  selectTemplate() {
    return templatesByType[this.data.type] || defaultTemplate;
  }
}
```

#### Custom List Item Override

```TypeScript
const defaultItemTemplate = html`
  <li>${x => x.name}</li>
`;

const template = html`
  <ul>
    ${repeat(x => x.items, x => x.itemTemplate)}
  <ul>
`;

@customElement(...)
export class MyElement extends FASTElement {
  @observable items: any[] = [];
  @observable itemTemplate = defaultItemTemplate;
}
```

#### Per Item List Types

```TypeScript
const defaultTemplate = html`...`;
const templatesByType = {
  foo: html`...`,
  bar: html`...`,
  baz: html`...`
};

const template = html<MyElement>`
  <ul>
    ${repeat(x => x.items, html`
      <li>
        ${(x, c) = c.parent.selectTemplate(x)}
      </li>
    `)}
  <ul>
`;

@customElement(...)
export class MyElement extends FASTElement {
  @observable items: any[] = [];

  selectTemplate(item) {
[...]

PR #3182, opened 2020-05-22
feat(fast-element): add content composition

EisenbergEffect:
# Description

This feature adds the ability for expressions located in arbitrary HTML content to return a template to be rendered into that location.

## Motivation & context

For an overview of the feature, have a read through issue #3039. The technical approach outlined there was exactly what was put into place here. Additionally, some basic cache smarts were added to the content binding mechanism, enabling the entire `when` directive to be deleted and instead simply be *syntax sugar* on top of the low-level composition capabilities. Everything from #3039 is implemented in this PR with the exception of the changes to `repeat`. Those are planned for a separate PR next week.

Unit tests have been added around content composition, `repeat`, and other areas touched by this PR. As part of the unit test writing process, I realized it would be nice to have a helper to await the next dom batch update, so `DOM.nextUpdate(): Promise` was added.

The best place to look for an understanding of the new content composition features is the update to the documentation in this PR.

## Issue type checklist

Technically, this is a breaking change as it removes `WhenDirective` and `WhenBehavior`. However, to my knowledge, no one was using those directly. Instead, they were using the `when` helper in their templates. The `when` helper remains, but its implementation is now changed to leverage the core composition system.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-element/docs/guide/defining-elements.doc.md
+++ b/packages/web-components/fast-element/docs/guide/defining-elements.doc.md
@@ -511,7 +511,7 @@ Some context properties are opt-in because they are more costly to update. So, f
 
 #### Composing Templates
 
-The `ViewTemplate` returned from the `html` tag helper has special handling when it is used inside of another template. The is that you can create templates and compose them into other templates.
+The `ViewTemplate` returned from the `html` tag helper has special handling when it is used inside of another template. This is done so that you can create templates and compose them into other templates.
 
 **Example: Composing Templates**
 
--- a/packages/web-components/fast-element/docs/guide/defining-elements.doc.md
+++ b/packages/web-components/fast-element/docs/guide/defining-elements.doc.md
@@ -406,21 +406,38 @@ const template = html<MyApp>`
 })
 export class MyApp extends FASTElement {
   @observable ready: boolean = false;
+  @observable data: any = null;
 
   connectedCallback() {
     super.connectedCallback();
+    this.loadData();
+  }
 
-    fetch('some/resource').then(response => {
-      // do something with response...
-      this.ready = true;
-    });
+  async loadData() {
+    const response = await fetch('some/resource');
+    const data = await response.json();
+    
+    this.data = data;
+    this.ready = true;
   }
 }
 ```
 
 > **REMINDER:** The `@observable` decorator creates a property that the template system can watch for changes. It is similar to `@attr`, but the property is not surfaced as an HTML attribute on the element itself. While `@attr` can only be used in a `FASTElement`, `@observable` can be used in any class.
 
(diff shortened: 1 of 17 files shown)


awentzel:
Is `ACKNOWLEDGEMENT.md` a special file type?  Curious, why it's in all caps, yet the other files are not.

packages/web-components/fast-element/src/directives/when.ts
```diff
@@ -1,77 +1,26 @@
...
+): CaptureType<T> {
+    let getTemplate: Expression<T, SyntheticViewTemplate>;
 
-    public createBehavior(target: any): WhenBehavior {
-        return new WhenBehavior(target, this.expression, this.template);
+    if (typeof templateOrTemplateExpression === "function") {
+        getTemplate = templateOrTemplateExpression;
+    } else {
+        getTemplate = (): SyntheticViewTemplate => templateOrTemplateExpression;
     }
```
janechu, line 22: nit: i'd suggest:
```typescript
const getTemplate: Expression<T, SyntheticViewTemplate> = typeof templateOrTemplateExpression === "function"
    ? templateOrTemplateExpression
    : (): SyntheticViewTemplate => templateOrTemplateExpression;
```

or:
```typescript
return (source: any, context: ExecutionContext): SyntheticViewTemplate | null =>
    !condition(source, context)
        ? null
        : typeof templateOrTemplateExpression === "function"
        ? templateOrTemplateExpression(source, context)
        : (): SyntheticViewTemplate => templateOrTemplateExpression(source, context);
```

EisenbergEffect: I'll refactor to the first one. The second would be a bit slower due to repeated type checks, where as the first option does it one time for all instances created from the template.

EisenbergEffect:
@awentzel It's not special in the way that GitHub understands it, but it is a special file in the general sense that we clearly acknowledge other libraries, etc. that influence us. I picked up this practice 5ish years ago and have used it in my own open source. There's precedent in other projects as well. See Angular for another example: https://github.com/angular/angular/tree/master/docs

packages/web-components/fast-element/docs/guide/defining-elements.doc.md
```diff
@@ -495,14 +511,12 @@ Some context properties are opt-in because they are more costly to update. So, f
 
 #### Composing Templates
 
-The `HTMLTemplate` returned from the `html` tag helper is also a directive itself. As a result, you can create templates and compose them into other templates.
+The `ViewTemplate` returned from the `html` tag helper has special handling when it is used inside of another template. The is that you can create templates and compose them into other templates.
```
nicholasrice, line 514: > **The** is that you can create templates and compose them into other...


Should "the" be "this"? Phrasing seems off.

EisenbergEffect: Definitely off 🤣  I'll fix it.

nicholasrice:
Looks great!

PR #3184, opened 2020-05-26
feat(fast-foundation): add element test helper

EisenbergEffect:
# Description

Creates a new `fixture` helper function to set up custom element tests. Also a `uniqueElementName` helper was added to facilitate generating unique element names for test scenarios.

## Motivation & context

We need some basic utilities to aid in testing web components, including creating them, binding them, and pushing them through their lifecycle, while being able to make assertions along the way.

See `fixture.spec.ts` for example usage.
This PR is dependent on APIs in this PR: https://github.com/microsoft/fast-dna/pull/3182

## Issue type checklist

## Process & policy checklist

Note: I'll push another commit in a bit that adds code docs.

## Next Steps

After we get some usage of this within our components, I'd like to move this either to its own package, like `@microsoft/fast-testing` where we could also add other test helpers and make this available for the community to use.

### DIFF
--- a/packages/web-components/fast-foundation/src/fixture.ts
+++ b/packages/web-components/fast-foundation/src/fixture.ts
@@ -32,7 +32,7 @@ function findElement(view: HTMLView): HTMLElement {
     return current as any;
 }
 
-export function uniqueName(): string {
+export function uniqueElementName(): string {
     return `fast-unique-${Math.random().toString(36).substring(7)}`;
 }
 
--- a/packages/web-components/fast-foundation/src/fixture.ts
+++ b/packages/web-components/fast-foundation/src/fixture.ts
@@ -5,20 +5,80 @@ import {
     ViewTemplate,
 } from "@microsoft/fast-element";
 
+/**
+ * Options used to customize the creation of the test fixture.
+ */
 export interface FixtureOptions {
+    /**
+     * The document to run the fixture in.
+     * @defaultValue `globalThis.document`
+     */
     document?: Document;
+
+    /**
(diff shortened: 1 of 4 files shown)


packages/web-components/fast-foundation/src/fixture.ts
```diff
@@ -0,0 +1,92 @@
...
+    const context = options.context || defaultExecutionContext;
+    const element = findElement(view) as any;
+    let isConnected = false;
+
+    view.bind(source, context);
+    view.appendTo(parent);
+
+    customElements.upgrade(parent);
+
+    await Promise.resolve();
```
nicholasrice, line 61: Why are you awaiting empty promises and not storing the value? Is this just to queue a new microtask and give the above time to render to the DOM?

EisenbergEffect: Yep! Just building in the micro task await to ensure everything is settled.

EisenbergEffect: I also chose to make everything async for consistency sake around this, and to prevent any future breaking changes if additional async work needs to happen within the helpers.

nicholasrice: Gotcha makes sense. Could you add a comment here addressing that question? I could see this being a bit confusing if folks aren't aware of the inner workings of fast-element or the task queue. Pretty cool though!

EisenbergEffect: Definitely!

PR #3220, opened 2020-06-02
fix: target slotted start and end elements in buttons

radium-v:
# Description
Anchors and buttons were getting empty 16x16 spans. This fix applies the sizing only when there's slotted content.

## Motivation & context
Buttons and anchors were incorrectly getting widened due to the presence of the extra space intended for start/end slots.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components-msft/src/styles/patterns/button.ts
+++ b/packages/web-components/fast-components-msft/src/styles/patterns/button.ts
@@ -51,21 +51,21 @@ export const BaseButtonStyles = css`
         opacity: var(--disabled-opacity);
     }
 
-    .start,
-    .end,
+    ::slotted([slot="start"]),
+    ::slotted([slot="end"]),
     ::slotted(svg) {
         ${
-            /* Glyph size and margin-left is temporary - 
+            /* Glyph size and margin-left is temporary -
             replace when adaptive typography is figured out */ ""
         } width: 16px;
         height: 16px;
     }
 
-    .start {
+    ::slotted([slot="start"]) {
         margin-inline-end: 11px;
     }
 
-    .end {
+    ::slotted([slot="end"]) {
         margin-inline-start: 11px;
     }
 `;
@@ -73,7 +73,7 @@ export const BaseButtonStyles = css`
 export const AccentButtonStyles = css`
     :host(.accent) .control {
         background: var(--accent-fill-rest);
-        color: var(--accent-foreground-cut-rest);  
+        color: var(--accent-foreground-cut-rest);
     }
 
     :host(.accent) .control:hover {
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-components-msft/src/styles/patterns/button.ts
```diff
@@ -51,29 +51,30 @@ export const BaseButtonStyles = css`
         opacity: var(--disabled-opacity);
     }
 
-    .start,
-    .end,
+    ::slotted([slot="start"]),
+    ::slotted([slot="end"]),
```
chrisdholt, line 55: I think we discussed continuing with the class here and pulling the class addition/removal from the template. @EisenbergEffect to confirm though

packages/web-components/fast-components-msft/src/styles/patterns/button.ts
```diff
@@ -51,29 +51,30 @@ export const BaseButtonStyles = css`
...
     ::slotted(svg) {
         ${
-            /* Glyph size and margin-left is temporary - 
+            /* Glyph size and margin-left is temporary -
             replace when adaptive typography is figured out */ ""
         } width: 16px;
+        outline: 1px solid #F0F;
```
chrisdholt, line 61: Can we remove this (assume it was for testing)?

packages/web-components/fast-components-msft/src/styles/patterns/button.ts
```diff
@@ -51,29 +51,30 @@ export const BaseButtonStyles = css`
...
         height: 16px;
     }
 
-    .start {
+    ::slotted([slot="start"]) {
         margin-inline-end: 11px;
     }
 
-    .end {
+    ::slotted([slot="end"]) {
```
chrisdholt, line 69: Reminder to revert back to the class implementation.

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -51,29 +51,29 @@ export const BaseButtonStyles = css`
         opacity: var(--disabled-opacity);
     }
 
-    .start,
-    .end,
+    ::slotted([slot="start"]),
+    ::slotted([slot="end"]),
```
chrisdholt, line 55: Turtles all the way down for this :)

packages/web-components/fast-components-msft/src/styles/patterns/button.ts
radium-v: oops, my bad

packages/web-components/fast-components-msft/src/styles/patterns/button.ts
EisenbergEffect: Yes, let's go with the class removal on the container.

radium-v: The class is still present in the template, we're just applying the `width/height/margin-inline` to the slotted content, when it's present.

PR #3224, opened 2020-06-03
doc: add webpack guide

EisenbergEffect:
# Description

This PR adds a basic step-by-step guide on how to setup a TypeScript+Webpack project for FAST Components from scratch.

## Issue type checklist

- [x] **Documentation**: Adds or improves documentation.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/docs/guide/webpack.doc.md
+++ b/packages/web-components/fast-foundation/docs/guide/webpack.doc.md
@@ -4,3 +4,216 @@ title: Webpack
 sidebar_label: Webpack
 custom_edit_url: https://github.com/microsoft/fast-dna/edit/master/packages/web-components/fast-foundation/docs/guide/webpack.doc.md
 ---
+
+FAST-DNA works great with TypeScript and Webpack, using a fairly standard setup. Let's take a look at how you can set up a TypeScript+Webpack project, starting from scratch.
+
+## Setting up the Package
+
+First, let's make a directory for our new project. From the terminal:
+
+```shell
+mkdir fast-webpack
+```
+
+Next, let's move into that directory, where we'll set up our project:
+
+```shell
+cd fast-webpack
+```
+
+From here, we'll initialize npm:
+
+```shell
+npm init
+```
+
+Follow the prompts from npm, answering each question in turn. You can always accept the defaults at first and then make changes later in the package.json file.
+
+Next, we'll install the FAST-DNA packages, along with supporting libraries. To do that, run this command:
+
+```shell
+npm install --save @microsoft/fast-components @microsoft/fast-element lodash-es tslib
+```
+
+We also need to install the Webpack build tooling:
+
+```shell
+npm install --save-dev rimraf ts-loader typescript webpack webpack-cli webpack-dev-server
+```
+
+## Adding Configuration and Source
+
+Now that we've got our basic package and dependencies setup, let's create some source files and get things configured. Since we're going to be writing a bit of code, now is a great time to involve a code editor in the process. If you're looking for a great editor for TypeScript and front-end in general, we highly recommend [VS Code](https://code.visualstudio.com/).
+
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-foundation/docs/guide/webpack.doc.md
```diff
@@ -3,4 +3,217 @@ id: webpack
...
+```shell
+npm init
+```
+
+Follow the prompts from npm, answering each question in turn. You can always accept the defaults at first and then make changes later in the package.json file.
+
+Next, we'll install the FAST-DNA packages, along with supporting libraries. To do that, run this command:
+
+```shell
+npm install --save @microsoft/fast-components @microsoft/fast-element lodash-es tslib
```
nicholasrice, line 35: I see `@microsoft/fast-components` enumerates `tslib` as a dependency - is this required here?

packages/web-components/fast-foundation/docs/guide/webpack.doc.md
```diff
@@ -3,4 +3,217 @@ id: webpack
...
+}
+```
+
+This setup uses `ts-loader` to process TypeScript. It will also enable both a production mode and a development mode that watches your source, recompiling and refreshing your browser as things change. You can read more about Webpack configuration in [the official Webpack documentation](https://webpack.js.org/).
+
+To enable easy execution of both our production and development builds, let's add some script commands to our `package.json` file. Find the `scripts` section of your `package.json` file and add the following two scripts:
+
+```json
+"scripts": {
+  "build": "rimraf dist && webpack --config webpack.config.js --mode=production",
```
nicholasrice, line 131: weblack has a "clean" plugin - maybe we should reference that instead of `rimraf` to keep the tool-chain consistent?

https://webpack.js.org/guides/output-management/#cleaning-up-the-dist-folder

nicholasrice, line 131: I think by default webpack will just pick up `webpack.config.js` as the config file. I'm fine with this as-is also but if you want to de-clutter the example I think you can omit.

EisenbergEffect: Let me test and confirm that and if so, I'll definitely omit.

EisenbergEffect: Will explore this evening and see if we can go that way. Thanks for the tip!

packages/web-components/fast-foundation/docs/guide/webpack.doc.md
EisenbergEffect: Hmm. Not sure actually. I'll play with removing it. I tried to remove lodash but that was no go.

packages/web-components/fast-foundation/docs/guide/webpack.doc.md
EisenbergEffect: Yep. Works like a charm. Thanks!

packages/web-components/fast-foundation/docs/guide/webpack.doc.md
EisenbergEffect: It works without it. However...it is a direct dep of the build when using the `importHelpers` setting in tsconfig...so I'm a bit nervous about removing it, esp. in the case that we change the dependencies of fast-components. If it's ok, I'll leave this one ine.

packages/web-components/fast-foundation/docs/guide/webpack.doc.md
radium-v: We're using the clean plugin in the `fast-website` project.

`npm install --save-dev clean-webpack-plugin`

then in webpack.config.js: 
```
const { CleanWebpackPlugin } = require("clean-webpack-plugin");

module.exports = {
    plugins: [
        new CleanWebpackPlugin(),
...
```

PR #3237, opened 2020-06-05
fix: overflow issue between 661px and 800px widths

freefaller69:
# Description

Added max-width of 95vw to fast-frame styles host.  Between 661px and 800px screen width there was an overflow issue that this resolves.

Before:
Screen Shot 2020-06-05 at 9 45 36 AM
Screen Shot 2020-06-05 at 9 45 52 AM

After:
Screen Shot 2020-06-05 at 5 05 33 PM
Screen Shot 2020-06-05 at 5 05 50 PM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/fast-frame/fast-frame.styles.ts
+++ b/sites/fast-website/src/app/components/fast-frame/fast-frame.styles.ts
@@ -23,8 +23,8 @@ export const FastFrameStyles = css`
     }
 
     :host {
-				--gutter: 20;
-				max-width: 95vw;
+        --gutter: 20;
+        max-width: 95vw;
     }
 
     .icon {
--- a/sites/fast-website/src/app/components/fast-frame/fast-frame.styles.ts
+++ b/sites/fast-website/src/app/components/fast-frame/fast-frame.styles.ts
@@ -24,7 +24,7 @@ export const FastFrameStyles = css`
 
     :host {
         --gutter: 20;
-        max-width: 95vw;
+        max-width: 94vw;
     }
 
     .icon {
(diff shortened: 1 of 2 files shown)


sites/fast-website/src/app/components/fast-frame/fast-frame.styles.ts
```diff
@@ -20,7 +20,8 @@ export const FastFrameStyles = css`
     }
 
     :host {
-        --gutter: 20;
+				--gutter: 20;
```
nicholasrice, line 23: Is prettier adding all this whitespace?

radium-v: looks like tabs instead of spaces. Is it possible that since this is a template literal, Prettier isn't picking it up?

freefaller69: It must be.  I know I didn't add it intentionally.  Let me see if I can do something to fix it.

freefaller69: It was tabs instead of spaces for some reason, and running prettier did not catch it.  Got it fixed manually so should be all good now.

eljefe223:
Based on the after screenshot it seems like the gutters on the left and right side of the horizontal rules don't match.

eljefe223:
Mike and I worked together to solve the above issue mentioned above. looks like there was regression when the `section` class was added back to the `<section/>` that host the `<fast-frame/>` by simply removing that the page lays out properly.

freefaller69:
> Mike and I worked together to solve the above issue mentioned above. looks like there was regression when the `section` class was added back to the `<section/>` that host the `<fast-frame/>` by simply removing that the page lays out properly.

Thanks to Jeff's help we got this dialed in and the fix has been pushed up.  Looks great at all screen widths now.

PR #3252, opened 2020-06-08
docs(fast-element): add public API docs

EisenbergEffect:
# Description

Adds API docs to all public APIs in `fast-element`.

## Issue type checklist

- [x] **Documentation**: Adds or updates documentation content.

This involves several minor breaking changes where types were fixed up. It should not affect anyone.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-element/src/attributes.ts
+++ b/packages/web-components/fast-element/src/attributes.ts
@@ -142,7 +142,7 @@ export class AttributeDefinition implements Accessor {
     public readonly converter?: ValueConverter;
 
     /**
-     *
+     * Creates an instance of AttributeDefinition.
      * @param Owner - The class constructor that owns this attribute.
      * @param name - The name of the property associated with the attribute.
      * @param attribute - The name of the attribute in HTML.
--- a/packages/web-components/fast-element/src/attributes.ts
+++ b/packages/web-components/fast-element/src/attributes.ts
@@ -16,7 +16,7 @@ export interface ValueConverter {
 
     /**
      * Converts a DOM string to a typed value.
-     * @param value The DOM string to convert to a typed value.
+     * @param value - The DOM string to convert to a typed value.
      */
     fromView(value: string): any;
 }
--- a/packages/web-components/fast-element/src/attributes.ts
+++ b/packages/web-components/fast-element/src/attributes.ts
@@ -172,9 +172,9 @@ export class AttributeDefinition implements Accessor {
     }
 
     /**
-     * Sets the value of the property on the source object.
-     * @param source - The source object to access.
-     * @param value - The value to set the property to.
+     * Sets the value of the attribute/property on the source element.
+     * @param source - The source element to access.
+     * @param value - The value to set the attribute/property to.
      */
     public setValue(source: HTMLElement, newValue: any): void {
         const oldValue = source[this.fieldName];
@@ -198,8 +198,8 @@ export class AttributeDefinition implements Accessor {
     }
 
     /**
-     * Gets the value of the property on the source object.
-     * @param source - The source object to access.
+     * Gets the value of the attribute/property on the source element.
+     * @param source - The source element to access.
      */
     public getValue(source: HTMLElement): any {
         Observable.track(source, this.name);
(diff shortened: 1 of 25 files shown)


packages/web-components/fast-element/src/attributes.ts
```diff
@@ -78,6 +171,11 @@ export class AttributeDefinition implements Accessor {
         }
     }
 
+    /**
+     * Sets the value of the property on the source object.
+     * @param source - The source object to access.
```
janechu, line 176: source object or HTML Element?

packages/web-components/fast-element/src/attributes.ts
```diff
@@ -99,11 +197,16 @@ export class AttributeDefinition implements Accessor {
         }
     }
 
+    /**
+     * Gets the value of the property on the source object.
+     * @param source - The source object to access.
```
janechu, line 202: Also here, should we specify HTML Element?

packages/web-components/fast-element/src/attributes.ts
```diff
@@ -56,19 +103,65 @@ export const nullableNumberConverter: ValueConverter = {
...
+    public readonly mode: AttributeMode;
+
+    /**
+     * A {@link ValueConverter} that integrates with the property getter/setter
+     * to convert values to and from a DOM string.
+     */
+    public readonly converter?: ValueConverter;
+
+    /**
+     *
```
janechu, line 145: Should we put something like "Creates an instance of AttributeDefinition"?

packages/web-components/fast-element/src/directives/node-observation.ts
```diff
@@ -2,22 +2,54 @@ import { Accessor, Observable } from "../observation/observable";
...
+ * A base class for node observation.
+ * @internal
+ */
 export abstract class NodeObservationBehavior<T extends NodeBehaviorBehaviorOptions>
     implements Behavior {
     private source: any = null;
     private shouldUpdate!: boolean;
 
+    /**
+     *
```
janechu, line 26: "Creates an instance of NodeObservationBehavior"?

packages/web-components/fast-element/src/directives/repeat.ts
```diff
@@ -52,7 +63,14 @@ export class RepeatBehavior implements Behavior, Subscriber {
     private childContext: ExecutionContext | undefined = void 0;
     private bindView: typeof bindWithoutPositioning = bindWithoutPositioning;
 
-    constructor(
+    /**
+     * Creates an instance of RepeatBehavior.
+     * @param location - The location in the DOM to render the repeat.
+     * @param binding - The array to render.
+     * @param template - The template to render for each i9tem.
```
janechu, line 70: nit: item (sp)

packages/web-components/fast-element/src/attributes.ts
EisenbergEffect: They are one and the same in this particular case. I'll change it to `source element` to clarify.

packages/web-components/fast-element/src/attributes.ts
EisenbergEffect: Oops ;) On it.

packages/web-components/fast-element/src/directives/node-observation.ts
EisenbergEffect: On it.

packages/web-components/fast-element/src/attributes.ts
```diff
@@ -56,19 +103,65 @@ export const nullableNumberConverter: ValueConverter = {
...
+     */
+    public readonly name: string;
+
+    /**
+     * The name of the attribute in HTML.
+     */
+    public readonly attribute: string;
+
+    /**
+     * The {@link AttributeMode} that describes the behavior of this attribute.
```
nicholasrice, line 134: Do we have to `@link` all of the non-builtin types? Is this not resolved by TypeScript?

EisenbergEffect: This is used by the api-extractor. If you put the `{@link name}` it will turn that into an `<a>` with a link to the docs page for that API.

EisenbergEffect: You only need to use it if you want to make a link from within your description or remarks. The types in the API should be linked automatically.

PR #3257, opened 2020-06-09
fix: make the web component definition for values more permissable

janechu:
# Description

This makes the list of values more permissable when defining a web component definition.

## Motivation & context

The type string use case did not map to the `DataType` enum, this brings both into alignment so that a `DataType.number` for example can map to a value with type `number`.

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tooling/src/data-utilities/web-component.ts
```diff
@@ -4,7 +4,7 @@ export interface WebComponentAttributeValues {
     /**
      * The value
      */
-    name: string;
+    name: string | number | boolean | null;
```
nicholasrice, line 7: What is the expected behavior for boolean and null values? I can understand an implicit conversion from number -> string but the other two aren't as clear to me. Do these *all* convert directly to strings? `attr="true"` and `attr="null"`?

janechu: Good point, I'll take those out since they can be confusing.

PR #3261, opened 2020-06-09
fix: corrects behavior of style removal

nicholasrice:
# Description
I noticed when using `Controller.removeStyles(ElementStyles)` that the method was removing all styles *except* the styles I was trying to remove. This corrects that behavior.

The removal method invokes the `Array.prototype.filter` method, where a callback returning `true` keeps the item and returning `false` removes the item. `AdoptedStyleSheetsStyles.removeStylesFrom` was returning `false` for all stylesheets that were *not* associated to the `AdoptedStyleSheetsStyles` instance, and `true` for all that were. This is the opposite behavior that we want.

## Issue type checklist

## Process & policy checklist


nicholasrice:
Seeing if I can add one for this real quickly ;)

EisenbergEffect:
@nicholasrice If you could add a single unit test for this case, I can come in tomorrow and fill in the rest of the cases for the `Controller`. Would be good to capture this as part of the bug fix though.

PR #3296, opened 2020-06-12
chore: update badge style and section spacing

SamanthaAO:
# Description
- update badge styling to match figma/chore checklist
- update section spacing to match figma/Jason's email

Screen Shot 2020-06-12 at 2 41 59 PM

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/css/style.css
```diff
@@ -187,20 +187,39 @@ fast-divider::after {
...
 .section-badge {
-  color: var(--accent-fill-rest);
+  color: var(--accent-foreground-rest);
   font-weight: bold;
+  background: var(--neutral-fill-focus);
+  width: max-content;
+  align-self: center;
+  letter-spacing: 0.12em;
+  border-radius: calc(var(--corner-radius) * 1px);
+  padding: calc(var(--design-unit) * 1px) calc(var(--design-unit) * 4px);
```
chrisdholt, line 197: The badge supports a fill by default - I would expect us to use the API there to make this change. There are examples of how to create a map for the badge in the fixtures.

SamanthaAO: @chrisdholt Thank you for the highlighting those attributes. They were very straightforward to assign, but the issue that I am having is that the fill and color are applied to the div inside of the template with the class badge and not to the fast-badge itself. So to apply the padding and the other stylings, I need to be able to assign them to that internal div with the badge class as well from the main css file, and this is proving quite challenging. In the past for buttons I have been able to use part to access internal elements within a template, but since this template does not include a part, is there another way to do it in this case?

sites/fast-website/src/app/css/style.css
```diff
@@ -187,20 +187,39 @@ fast-divider::after {
 }
 
 .section-badge {
```
chrisdholt, line 189: @SamanthaAO a PR recently went in to support a CSS part so you can target the badge there.

To be clear on my previous comment, the badge supports attributes to map colors to it.

```css
fast-badge {
    --badge-fill-foo: var(--background-color);
    --badge-color-foo: var(--accent-foreground-rest);
}
```

```html
<fast-badge fill="foo" color="foo">Badge</fast-badge>
```

SamanthaAO: @chrisdholt yes, that was my understanding for your previous comment. However, once the fill is applied to the internal div through the variable assigned to fast-badge it means that I must also apply the padding in the same place that the fill is assigned. I will be able to do this now with part thanks to the other PR. Thank you for letting me know.

PR #3297, opened 2020-06-12
fix: screen reader for anchors and buttons and tab focus on fast-frame component

khamudom:
# Description

Using screen readers like VoiceOver, Narrator, and NVDA, tabbing onto anchors and buttons, that only have icons were not descriptive, the screen reader only reads the link. I went through and added `aria-label` on any anchors and buttons that only displays an icon.
When tabbing through the `fast-frame` component, while in full screen, I was unable to tab into the controls in the last column, because they were set to tabIndex="-1". I added an `isResponsive` observable that will toggle the tabIndex between the full and mobile size window.
Also set tabIndex="0" on the color swatch item.

Known issue that I will tackle separately as to not hold up these changes. They could be changes that have to be made within the components.
- Running NVDA, the aria-label does not get read on the anchor and button components.
- NVDA and Narrator do not read out the `fast-text-field` component when aria-label is added. The only time it will read the aria-label is when a value is set to it.

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -53,6 +53,7 @@ export const FastFrameTemplate = html<FastFrame>`
                         <div class="content-control-container" >
                             <label for="background-color-pickers">Background color</label>
                             <fast-radio-group
+                                tabindex="0"
```
chrisdholt, line 56: @marjonlynch can you advise as to why this might be necessary?

khamudom: Mark and I synced on this and the correct approach here is adding the tabIndex on the item and not the group.

marjonlynch: I created an issue #3300 to fix this in radio-group but maybe Kham should go ahead with what he has here until that's fixed.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -318,22 +320,22 @@ export const FastFrameTemplate = html<FastFrame>`
                             <span class="sample-control-icon"></span>
                             <span class="sample-control-text">Label</span>
                             <div class="sample-control-actions">
-                                <fast-button appearance="stealth"
+                                <fast-button appearance="stealth" aria-label="stealth"
```
chrisdholt, line 323: What's the purpose of this aria-label here? I don't think this is descriptive of the intent of the button. Should this say the appearance name, or should this say what this is doing?

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -318,22 +320,22 @@ export const FastFrameTemplate = html<FastFrame>`
...
                                     >${ShareIcon}</fast-button
                                 >
-                                <fast-button appearance="stealth"
+                                <fast-button appearance="stealth" aria-label="stealth"
```
chrisdholt, line 326: What's the purpose of this aria-label here? I don't think this is descriptive of the intent of the button. Should this say the appearance name, or should this say what this is doing?

sites/fast-website/src/public/index.html
```diff
@@ -110,7 +110,7 @@ <h2 class="section-heading">Join the community</h2>
                 <li><p>&copy; 2020 Microsoft</p></li>
               </ul>
             </site-navigation>
-            <fast-anchor class="logo" appearance="lightweight" href="/">
+            <fast-anchor class="logo" appearance="lightweight" href="/" aria-label="Fast logo">
```
chrisdholt, line 113: Does this direct users to the logo or does it direct them to the homepage? The label here should be descriptive of the link and not the visuals.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
khamudom: I agree it should say what the button is doing. In this case, the button here is only used to show what a button would look like when we change properties inside the fast frame control. My thought was to inform the user that its a stealth button instead of just a button. I could make it say "stealth button example" or "button example". But if we think just saying "button" is ok for this scenario, I can remove the aria-label.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
khamudom: Similar to the other stealth button scenario. But for this, it could say "Share button example"

sites/fast-website/src/public/index.html
khamudom: Good point. Will change to "homepage".

khamudom:
Closing to redo PR against master branch

PR #3323, opened 2020-06-16
feat(fast-element): enable children and slotted filters

EisenbergEffect:
# Description

This PR adds a long requested feature to `slotted` and `children` directives to be able to supply a filter for what nodes you want pushed into your property. For convenience, an `elements()` filter was added to enable easy filtering of only elements or elements with a particular tag name.

As part of this work, unit tests were added for both `slotted` and `children` and a couple of bugs in both typings and functionality were discovered as a result.

## Issue type checklist

## Process & policy checklist

The documentation update will come as part of my upcoming documentation PR, which is currently in draft form. I'll add it there in order to avoid some conflicts.


packages/web-components/fast-element/src/directives/node-observation.ts
```diff
@@ -55,7 +88,7 @@ export abstract class NodeObservationBehavior<T extends NodeBehaviorBehaviorOpti
             (x: Accessor) => x.name === name
         );
         this.source = source;
-        this.updateTarget(this.getNodes());
+        this.updateTarget(this.getNodes().filter(this.options.filter!));
```
nicholasrice, line 91: How often do you anticipate custom filters to not be provide? Would it make sense to skip the filter operation if no filter is provided?

EisenbergEffect: I'm honestly not sure. It might still be good to make the optimization though. Let me adjust that.

PR #3334, opened 2020-06-17
fix: screen reader for anchors and buttons and tab focus inside fast-frame component 

khamudom:
# Description

While using screen readers like VoiceOver, Narrator, and NVDA, and tabbing onto anchors and buttons, that only have icons, the screen readers only read out 'link'. I went through and added aria-label on any anchors and buttons that only displays an icon.
When tabbing through the fast-frame component, while in full screen, I was unable to tab into the controls in the last column, because they were set to tabIndex="-1". I added an isResponsive observable that will toggle the tabIndex between the full and mobile size window.
Also set tabIndex="0" on the color swatch item.

Known issue that I will tackle separately as to not hold up these changes. They could be changes that have to be made within the components.

Running NVDA, the aria-label does not get read on the anchor and button components.
NVDA and Narrator do not read out the fast-text-field component when aria-label is added. The only time it will read the aria-label is when a value is set to it.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
+++ b/sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
@@ -3,10 +3,18 @@ import { html } from "@microsoft/fast-element";
 const featureTemplate = html`<site-feature-card>
     <h4>${x => x.header}</h4>
     <p slot="body">${x => x.body}</p>
-    <fast-anchor slot="footer" href=${x => x.githubLink} appearance="lightweight"
+    <fast-anchor
+        slot="footer"
+        href=${x => x.githubLink}
+        appearance="lightweight"
+        aria-label=${x => x.header}
         >View Github</fast-anchor
     >
-    <fast-anchor slot="footer" href=${x => x.documentationLink} appearance="lightweight"
+    <fast-anchor
+        slot="footer"
+        href=${x => x.documentationLink}
+        appearance="lightweight"
+        aria-label=${x => x.header}
         >Read Documentation</fast-anchor
     >
 </site-feature-card> `;
--- a/sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
+++ b/sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
@@ -3,18 +3,10 @@ import { html } from "@microsoft/fast-element";
 const featureTemplate = html`<site-feature-card>
     <h4>${x => x.header}</h4>
     <p slot="body">${x => x.body}</p>
-    <fast-anchor
-        slot="footer"
-        href=${x => x.githubLink}
-        appearance="lightweight"
-        aria-label=${x => x.header}
+    <fast-anchor slot="footer" href=${x => x.githubLink} appearance="lightweight"
         >View Github</fast-anchor
     >
-    <fast-anchor
-        slot="footer"
-        href=${x => x.documentationLink}
-        appearance="lightweight"
-        aria-label=${x => x.header}
+    <fast-anchor slot="footer" href=${x => x.documentationLink} appearance="lightweight"
         >Read Documentation</fast-anchor
     >
 </site-feature-card> `;
--- a/sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
+++ b/sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
@@ -63,6 +63,7 @@ export const FastFrameTemplate = html<FastFrame>`
 
                                     html<string>`
                                         <site-color-swatch
+                                            tabindex="0"
                                             value="${x => x}"
                                             background-color="${x => x}"
                                             checked="${(x, c) =>
@@ -84,6 +85,7 @@ export const FastFrameTemplate = html<FastFrame>`
 
                                     html<string>`
                                         <site-color-swatch
+                                            tabindex="0"
                                             value="${x => x}"
                                             background-color="${x => x}"
                                             checked="${(x, c) =>
@@ -255,7 +257,7 @@ export const FastFrameTemplate = html<FastFrame>`
                                     position="0"
                                 >
                                     0
-                                </fast-slider-label>            
+                                </fast-slider-label>
                                 <fast-slider-label
                                     hide-mark
                                     position="6"
@@ -329,11 +331,11 @@ export const FastFrameTemplate = html<FastFrame>`
                     </div>
                 </fast-card>
                 <div
-                    aria-hidden="${x => !x.expanded}"
                     class="preview-controls"
                 >
                     <fast-progress></fast-progress>
-                    <fast-menu tabIndex="${x => (x.expanded ? "0" : "-1")}">
+                    <fast-menu tabIndex="${x =>
+                        !x.expanded && x.isResponsive ? "-1" : "0"}">
                         <fast-menu-item role="menuitem">Menu item 1</fast-menu-item>
                         <fast-menu-item role="menuitem">Menu item 2</fast-menu-item>
                         <fast-menu-item role="menuitem">Menu item 3</fast-menu-item>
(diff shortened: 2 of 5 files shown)


sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
```diff
@@ -3,10 +3,18 @@ import { html } from "@microsoft/fast-element";
 const featureTemplate = html`<site-feature-card>
     <h4>${x => x.header}</h4>
     <p slot="body">${x => x.body}</p>
-    <fast-anchor slot="footer" href=${x => x.githubLink} appearance="lightweight"
+    <fast-anchor
+        slot="footer"
+        href=${x => x.githubLink}
+        appearance="lightweight"
+        aria-label=${x => x.header}
```
chrisdholt, line 10: @khamudom does this need an aria-label? Doesn't this already have a visual label?

sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
```diff
@@ -3,10 +3,18 @@ import { html } from "@microsoft/fast-element";
...
         >View Github</fast-anchor
     >
-    <fast-anchor slot="footer" href=${x => x.documentationLink} appearance="lightweight"
+    <fast-anchor
+        slot="footer"
+        href=${x => x.documentationLink}
+        appearance="lightweight"
+        aria-label=${x => x.header}
```
chrisdholt, line 17: @khamudom does this need an aria-label? Doesn't this already have a visual label?

sites/fast-website/src/app/components/side-navigation/side-navigation.template.ts
```diff
@@ -14,6 +14,7 @@ export const SideNavigationTemplate = html<SideNavigation>`
                                 href=${x => x.actionLink}
                                 appearance="lightweight"
                                 :innerHTML=${x => x.icon}
+                                aria-label=${x => x.header}
```
chrisdholt, line 17: I think you'll want a specific property for this unless "Header" is specifically set to show "Medium", "Github", etc...

sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
khamudom: will remove this, for now, to see if we get any feedback from CELA

sites/fast-website/src/app/components/content-placement-container/templates/feature.template.ts
khamudom: also will remove this, for now, to see if we get any feedback from CELA

sites/fast-website/src/app/components/side-navigation/side-navigation.template.ts
khamudom: switching to use `actionText` instead of `header`

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -343,39 +345,44 @@ export const FastFrameTemplate = html<FastFrame>`
...
                         <fast-slider tabIndex="${x =>
-                            x.expanded ? "0" : "-1"}"></fast-slider>
-                        <fast-flipper aria-hidden="false" tabIndex="${x =>
-                            x.expanded ? "0" : "-1"}"></fast-flipper>
-                        <fast-flipper disabled tabIndex="${x =>
-                            x.expanded ? "0" : "-1"}"></fast-flipper>
+                            !x.expanded && x.isResponsive ? "-1" : "0"}"></fast-slider>
+                        <fast-flipper aria-label="Example 'flipper' button" tabIndex="${x =>
+                            !x.expanded && x.isResponsive ? "-1" : "0"}"></fast-flipper>
+                        <fast-flipper disabled aria-label="Example disabled 'flipper' button"></fast-flipper>
```
chrisdholt, line 376: Just an FYI that normally these are hidden from screen readers, so they don't end up being exposed and are out of the tabindex. We should consider if we want to expose these just for the sake of exposing them. I don't know that there is any value as they are not actionable.

I'm curious as to why we might want any of these non-actionable but interactive elements exposed here.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -343,39 +345,44 @@ export const FastFrameTemplate = html<FastFrame>`
...
                     </div>
                     <div class="control-container">
-                        <fast-button appearance="accent" tabIndex="${x =>
-                            x.expanded ? "0" : "-1"}">
+                        <fast-button appearance="accent" aria-label="Example 'download' button" tabIndex="${x =>
```
chrisdholt, line 379: @khamudom - `tabindex` -> should be lowercase all the way through here.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
khamudom: good point. they don't do anything actionable but are only there to show a sited user a visual change. I will remove the floppers from the tab index. With that said, and out of curiosity, should we also remove the tab index from all of the example controls?

chrisdholt: I’d just drop the tab index from the flipper (and aria-label) for now...let’s get feedback from our external review on this all up.

awentzel:
I'd like to see in the title and description of what changes were made rather than describing the problem.  Often changelogs are used to see what features or fixes were implemented, and this will better align us for the future when we release prettier changelogs.

khamudom:
> I'd like to see in the title and description of what changes were made rather than describing the problem. Often changelogs are used to see what features or fixes were implemented, and this will better align us for the future when we release prettier changelogs.

I agree. I tried to make the title as descriptive as I can, at a glance, without going into full detail. Sometimes what I think is descriptive, might not be for the majority, probably it's because I know what the changes are. If you have any suggestions please add.

sites/fast-website/src/app/data/community.data.ts
```diff
@@ -38,6 +38,7 @@ export const communityContentPlacementData: CommunityContentPlacementData[] = [
     },
     {
         actionLink: "https://medium.com/fast-dna",
+        actionText: "Follow us on Medium",
```
SamanthaAO, line 41: nit: Could we maybe change this to something like "Read our blog on Medium"? That is the action text on the current website. Follow us sounds more like Twitter than Medium to me.

sites/fast-website/src/app/components/fast-frame/fast-frame.ts
```diff
@@ -170,8 +173,8 @@ export class FastFrame extends FASTElement {
         this.backgroundColor = this.previewBackgroundPalette[this.lastSelectedIndex];
     };
 
-    private resetExpanded = (): void => {
-        this.expanded = false;
+    private resetExpandedResponsive = (): void => {
+        (this.expanded = false), (this.isResponsive = !this.isResponsive);
```
janechu, line 177: If the drawer breakpoint is hit, then `isResponsive` is inverted? Can you explain this logic? Seems like you get the reverse depending on if you are starting with a width of greater than the drawer breakpoint vs starting with less than the drawer breakpoint.

khamudom: I might be calling the observable, `isResponsive` by a wrong name, and that might make this confusing. But the main goal I am trying to achieve here is to make sure `tabindex` on the controls is always 0 for when its in full-width, beyond the mobile breakpoint. Then when it hits the mobile breakpoint, the tabindex can go to -1. This is what `isResponsive` is doing. Then at the mobile breakpoint, `expanded` will take over the tabindex.
Currently, we only have the observable, `expanded`, toggling the `tabindex` to 0 and -1, but only when it is at the mobile breakpoint.

janechu: I think the sticking point is "when the mobile breakpoint is hit" which this could either start off as being in the mobile breakpoint, or it could start larger than the mobile breakpoint and the user could resize the window and hit it. These seem like the produce inverted values because all this does is reverse the boolean. Shouldn't this just be setting it to `false` or whatever the case may need to be? Why reverse whatever it was initially.

khamudom: Didn't think about it starting off on mobile. Did some restructuring, created a property, called `mql`, set that as the windows.matchMedia breakpoint, and added the event listener to that.
In addition, renamed `isResponsive` to `isMobile` and created a `setTabIndex` property.
Now when you start off in mobile or full-width, the tabindex is set to match the window size.
Thanks to John for helping me out with this.

PR #3338, opened 2020-06-18
chore: update side-nav scroll styles and remove scroll indicator click behavior

SamanthaAO:
# Description

- update scroll indicator and icon styles (side-nav) to more closely align with figma
- remove click smooth scroll from scroll indicators
- add changes to improve the side navs alignments with each other compared to the original version of this PR which was intended for feature/fast-frame

scroll-indicator

## Issue type checklist

## Process & policy checklist


sites/fast-website/src/app/components/side-navigation/side-navigation.styles.ts
```diff
@@ -2,56 +2,54 @@ import { css } from "@microsoft/fast-element";
...
+    accentForegroundCutRestBehavior,
     accentForegroundHoverBehavior,
-    neutralForegroundRestBehavior,
 } from "@microsoft/fast-components";
 
 export const SideNavigationStyles = css`
-    ${display("flex")}:host {
-        --engaged-height: calc(var(--design-unit) * 9px);
+    ${display("flex")} :host {
+        --active-height: calc(var(--design-unit) * 13px);
```
chrisdholt, line 11: The references to --design-unit here seem somewhat arbitrary in terms of what we are multiplying by. Do we want this to change size in coordination with the design unit, or should we just hardcode the value?

sites/fast-website/src/app/components/side-navigation/side-navigation.styles.ts
```diff
@@ -2,56 +2,54 @@ import { css } from "@microsoft/fast-element";
...
+
+    .list-item:hover .icon {
         fill: var(--accent-fill-hover);
     }
 
     .scroll-indicator {
-        height: var(--design-unit-large);
-        width: calc(var(--design-unit) * 1px);
-        background-color: ${neutralForegroundRestBehavior.var};
+        height: calc(var(--design-unit) * 9px);
```
chrisdholt, line 38: Similar question to above - does this change with the design unit, or should it just be hardcoded?

sites/fast-website/src/app/css/style.css
```diff
@@ -31,34 +31,23 @@ body {
...
+  width: 100%;
+  height: min-content;
   position: fixed;
-  top: 40%;
-  right: 0%;
+  top: 50%;
+  transform: translateY(-50%);
+  display: flex;
+  justify-content: space-between;
+  z-index: -1;
```
chrisdholt, line 42: Do we need z-index here? Why do we need to bury this container?

sites/fast-website/src/app/components/side-navigation/side-navigation.styles.ts
SamanthaAO: @chrisdholt My thinking was that if I used design unit as part of the height that then the scroll indicators would adjust with the rest of the design system as we adjusted the page's density. If this is not the case I will hard code them.

chrisdholt: @SamanthaAO I'm primarily wondering if we've considered or seen how vast that difference can be. Could this end up obstructing the UI, etc. Technically, I'm not sure changing the density multiplier would directly affect this. Changing the design unit certainly would. A small change to the design unit could introduce significant changes here. I think we could certainly look at making these adaptive, I'm just not sure that we have enough insight into how we should go about that now.

sites/fast-website/src/app/css/style.css
radium-v: Without the `z-index`, the container sits on top of the content and prevents interacting with any content in the center of the page.

sites/fast-website/src/app/components/side-navigation/side-navigation.styles.ts
SamanthaAO: That makes sense. I think that the safest decision now would be to hard code these values. I will update it.

PR #3345, opened 2020-06-19
feat(fast-element): enable CSSStyleSheet and string to be used as styles

EisenbergEffect:
# Description

Based on feedback from MS Graph, who is using sass, this PR makes it easier to pass a string directly to the element definition styles. Additionally, support for passing a CSSStyleSheet instance directly to the definition, as well as composing CSSStyleSheet instances with css has been enabled. This makes FAST forward compatible with the upcoming CSS Modules platform feature which allows importing from a ".css" file where the import object is transformed by the platform into an instance of CSSStyleSheet.

This change loosens the types and is non-breaking.

## Process & policy checklist


packages/web-components/fast-element/docs/guide/leveraging-css.md
```diff
@@ -124,6 +124,10 @@ const styles = css`
 
 Rather than simply concatenating CSS strings, the `css` helper understands that `normalize` is `ElementStyles` and is able to re-use the same Constructable StyleSheet instance as any other component that uses `normalize`. 
 
+:::note
+You can also pass a CSS string or a CSSStyleSheet instance directly to the element definition, without needing to use the `css` helper and it will automatically be converted into `ElementStyles`. The advantage of using the `css` helper is that it enables the rich composition and reuse of styles described above, with automatic runtime caching for memory efficiency and performance.
```
nicholasrice, line 128:
```suggestion
You can also pass a CSS string or a [CSSStyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet) instance directly to the element definition, without needing to use the `css` helper and it will automatically be converted into `ElementStyles`. The advantage of using the `css` helper is that it enables the rich composition and reuse of styles described above, with automatic runtime caching for memory efficiency and performance.
```

PR #3350, opened 2020-06-22
fix: selected tab becoming undefined when tab slot pass anything except icon or text node

### LINKED ISSUE #3349: selected tab resets to tab one in Tabs when a tab slot is passed anything except a text node or icon
In Tabs, if a tab slot is passed anything except and icon or text node for content, when that tab content is clicked Tabs always switches to the first tab.

1. Pass a `div` or `p` tag (almost any element) with content to at least one _tab slot_ in Tabs, and have at least two tabs.
2. Click on the element of second tab. (note: click on the tab just outside of the content/element works)
3. Notice that the tab switches back to the first tab when clicking on the content of the tab.

The active tab should be the selected tab.

Screen-Recording-2020-06-22-at-9

SethDonohue:
# Description

-Updated selectedTab to be the `event.currentTarget` to ensure the selected tab is always an existing tab so the activeid does not become undefined.
-Added fixture example to check for this

## Motivation & context

Solves the issue of any element except a text node or icon being passed to the tabs causes the tabs to switch to  the first tab even when selecting the other tabs.

Before:
Screen-Recording-2020-06-22-at-9

After:
Screen-Recording-2020-06-22-at-1

closes #3349 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/tabs/fixtures/base.html
+++ b/packages/web-components/fast-components/src/tabs/fixtures/base.html
@@ -36,6 +36,21 @@
             Tab three content. This is for testing.
         </fast-tab-panel>
     </fast-tabs>
+    <h4>Div and Paragraph Elements in Tab</h4>
+    <fast-tabs>
+        <fast-tab id="TabOne">Tab one</fast-tab>
+        <fast-tab id="TabTwo"><p class="element-tab">div tab</p></fast-tab>
+        <fast-tab id="TabThree"><div class="element-tab">paragraph tab</div></fast-tab>
+        <fast-tab-panel>
+            Tab one content. This is for testing.
+        </fast-tab-panel>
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-components/src/tabs/fixtures/base.html
```diff
@@ -91,7 +106,7 @@ <h4>Icons - Vertical</h4>
             Tab three content. This is for testing.
         </fast-tab-panel>
     </fast-tabs>
-    <h4>Suplemental content</h4>
+    <h4>Supplemental content</h4>
```
eljefe223, line 109: This should be fixed in fast-components-msft also

SethDonohue: ahh good catch

PR #3377, opened 2020-06-25
fix: removed the underline from the logo

khamudom:
# Description

Removed the underline in both the header and footer logo.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/css/style.css
+++ b/sites/fast-website/src/app/css/style.css
@@ -131,6 +131,10 @@ site-navigation ::part(content) {
   align-items: flex-end;
 }
 
+.logo::part(content)::before {
+  background: transparent;
+}
+
 .header .github-link::part(control) {
   justify-content: flex-end;
 }


sites/fast-website/src/app/css/style.css
```diff
@@ -122,6 +122,10 @@ site-navigation ::part(content) {
   align-items: flex-end;
 }
 
+.logo::part(content)::before {
+  background: transparent;
+}
+
```
chrisdholt, line 128: Another way you could manage this would be to pass the logo to the start or end slot. This would require no CSS changes as that slot doesn't get underlined. Hovering the GH link will illustrate.

khamudom: I could put it in the start slot. but it looks like the log and FAST text are both SVGs. I could do this,
`<span slot="start">
              <%= require("svg/icon-brand.svg") %>
              <%= require("svg/icon-brand-fast.svg") %>
            </span>`
but I will still need to add another class to fix the alignment.

chrisdholt: I'm fine with either implementation. The current one seems to be less intrusive as it currently is, so let's roll with it.

PR #3400, opened 2020-06-26
fix: update standard luminance values for fast-components

chrisdholt:
## Issue type checklist

## Process & policy checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5934,18 +5934,10 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*", "@types/react@^16.3.0", "@types/react@^16.4.18", "@types/react@^16.7.17", "@types/react@^16.8.0", "@types/react@^16.8.19":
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -5934,22 +5934,14 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*", "@types/react@^16.3.0", "@types/react@^16.4.18", "@types/react@^16.7.17", "@types/react@^16.8.0", "@types/react@^16.8.19":
+"@types/react@*", "@types/react@^16.3.0", "@types/react@^16.4.18", "@types/react@^16.7.17", "@types/react@^16.8.0", "@types/react@^16.8.19", "@types/react@^16.9.17":
```
EisenbergEffect, line 5937: Just want to make sure this is expected.

nicholasrice: It does seem strange that yarn is doing this. I think this means that only 1 version of `@types/react` is installed though, so I think it's what we want.

PR #3402, opened 2020-06-26
feat(fast-element): support bindings and observables with volatile dependencies

EisenbergEffect:
# Description

Updates the observation system, binding system, and directives to handle scenarios where a binding has "volatile" dependencies. For example, when successive runs of the binding might yield a different set of dependent observables. This is the case when the binding contains conditional logic of some sort.

## Motivation & context

We discovered an issue where a binding expression like the following was not properly re-evaluating:

```ts
x => x.count < 1 ? 'No Results' : x.resultCountMessage
```

If the first time the binding evaluated, the condition was true, then only the `count` property would be accessed and the observation system would think this binding was only dependent on `count`. If the condition changed to false, then `resultCountMessage` would be accessed but not tracked. So, if `resultCountMessage` ever changed, the binding would not refresh.

This PR introduces the idea of *volatility* to binding expressions and computed properties. For expressions, we can check whether it's volatile with a simple regex over the function body, to determine whether or not it contains conditional indicators. If so, we mark the binding as volatile and then each time it is evaluated, we refresh the dependencies. For computed properties, we introduce the `@volatile` decorator, which can be used on computed properties with conditional logic. There's also an imperative API for volatility so that decorators are not required.

## Issue type checklist

Some low-level APIs have been changed. This is unlikely to affect customers at this time.

## Process & policy checklist


packages/web-components/fast-element/docs/guide/observables-and-state.md
```diff
@@ -22,22 +22,93 @@ The `@attr` decorator can only be used in a `FASTElement` but the `@observable`
...
+}
+```
+
+Here's how you would do it without a decorator:
+
+```ts
+import { Observable } from '@microsoft/fast-element';
+
+export class MyClass {
+  // elided
```
janechu, line 94: elided?

packages/web-components/fast-element/src/observation/observable.ts
```diff
@@ -1,6 +1,7 @@
 import { DOM } from "../dom";
 import { Notifier, PropertyChangeNotifier, SubscriberSet, Subscriber } from "./notifier";
 
+const volatileRegex = /(\?|\&\&|if)/;
```
nicholasrice, line 4: should this be testing for `x.foo || x.bar` instead of `x.foo && x.bar`?

packages/web-components/fast-element/docs/guide/observables-and-state.md
EisenbergEffect: I just didn't want to repeat all the property definitions again from the example above. I can say "properties elided" if that's clearer.

packages/web-components/fast-element/src/observation/observable.ts
EisenbergEffect: Hmm. Probably need to test both actually. Because if the first term of an `&&` is false, it won't access the second but if the first term of an `||` is true, it also wouldn't access the second. I'll fix that.

nicholasrice: ahh yea you're right - `&&` won't always access both.

nicholasrice: oh hmm - you probably also want to catch switch cases huh?

EisenbergEffect:
I think I can switch out `?` for `:` and cover both ternary and switch/case, and then I can add `||` as well. I think that will be good enough. Most people won't inline complex logic into the expression. Instead they'll put it in a computed property and mark it as `volatile`. So, I think covering a few telltale symbols should be pretty good.

packages/web-components/fast-element/docs/guide/observables-and-state.md
EisenbergEffect: I went ahead and expanded the code out to help clarify the scenario.

packages/web-components/fast-element/src/observation/observable.ts
EisenbergEffect: Fixed and added tests for `||` and `switch/case`.

Issue #3404, opened 2020-06-26
Update the logic for the website side navigation indicators.


chrisdholt:
The issue is that when the scroll indicator was created each of the sections had the height of 100vh, so the intersection observer was written to compare the area of the sections intersecting the screen and if a section took up more than half of the screen then it was assigned as the currentSection. Since the sections are now all different heights there are issues with smaller sections such as hero not ever being able to take up half of the screen to become currentSection/active. There are also issues with larger screens not registering the correct section because as the screen size gets larger and the section sizes stay the same, sections become unable to take up half of the screen and then do not become the currentSection.

chrisdholt:
Closed with #3404

PR #3407, opened 2020-06-26
fix: for scroll indicator #3404

SamanthaAO:
# Description

- fix for #3404
- update intersection observer function to fix missed sections on large screens and sections with reduced heights

**note: Please smoke these changes! I do not have a large monitor, so have had Kham run them for me to test that they work, but if we could have further large monitor testing it would be greatly appreciated. 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/side-navigation/side-navigation.ts
+++ b/sites/fast-website/src/app/components/side-navigation/side-navigation.ts
@@ -4,6 +4,11 @@ import {
     communityContentPlacementData,
 } from "../../data/community.data";
 
+export interface sectionData {
+    section: string;
+    value: number;
+}
+
 export class SideNavigation extends FASTElement {
     @attr
     public category: string;
@@ -14,6 +19,9 @@ export class SideNavigation extends FASTElement {
     @observable
     public sectionArray: HTMLElement[] = [];
 
+    @observable
+    public objArray: sectionData[];
+
     public socialData: CommunityContentPlacementData[] = communityContentPlacementData.filter(
         x => x.header !== "Github"
     );
@@ -21,14 +29,20 @@ export class SideNavigation extends FASTElement {
     loadScrollers() {
         this.sectionArray = Array.from(document.querySelectorAll("section[id]"));
 
+        this.objArray = this.sectionArray.map(x => {
+            return { section: x.id, value: 0 };
+        });
+
         let observer = new IntersectionObserver(
             entries => {
                 entries.forEach(entry => {
-                    const areaOnScreen =
-                        entry.intersectionRatio * entry.boundingClientRect.height;
-                    if (areaOnScreen > 0.5 * window.innerHeight) {
-                        this.currentSection = entry.target.id;
-                    }
+                    const intersectingSection: sectionData = this.objArray.find(
+                        ({ section }) => section === entry.target.id
+                    ) as sectionData;
+                    intersectingSection.value = entry.intersectionRatio;
+                    this.currentSection = this.objArray.reduce((prev, current) =>
+                        prev.value > current.value ? prev : current
+                    ).section;
                 });
             },
             { threshold: [0, 0.2, 0.4, 0.6, 0.8] }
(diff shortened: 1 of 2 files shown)


sites/fast-website/src/app/components/side-navigation/side-navigation.ts
```diff
@@ -4,6 +4,11 @@ import {
     communityContentPlacementData,
 } from "../../data/community.data";
 
+export interface sectionData {
```
scomea, line 7: More descriptive svp.  Maybe value could be "intersectionRatio" to describe what it is? sectionData.value could be anything...

sites/fast-website/src/app/components/side-navigation/side-navigation.ts
```diff
@@ -14,21 +19,30 @@ export class SideNavigation extends FASTElement {
...
-                    const areaOnScreen =
-                        entry.intersectionRatio * entry.boundingClientRect.height;
-                    if (areaOnScreen > 0.5 * window.innerHeight) {
-                        this.currentSection = entry.target.id;
-                    }
+                    const intersectingSection: sectionData = this.objArray.find(
+                        ({ section }) => section === entry.target.id
+                    ) as sectionData;
+                    intersectingSection.value = entry.intersectionRatio;
+                    this.currentSection = this.objArray.reduce((prev, current) =>
```
scomea, line 43: I think you can do this once after running through all the entries rather than after each entry update.

sites/fast-website/src/app/components/side-navigation/side-navigation.ts
```diff
@@ -14,21 +19,30 @@ export class SideNavigation extends FASTElement {
...
+                        prev.value > current.value ? prev : current
```
scomea, line 44: Does this do what we want? I think this will set the "current section" to be the one at the highest index that is the most visible.  ie. if all sections are visible the last one is the current one.

sites/fast-website/src/app/components/side-navigation/side-navigation.ts
SamanthaAO: Thank you very much for the suggestion. I have changes value to intersectionRatio and it is much clearer now.

sites/fast-website/src/app/components/side-navigation/side-navigation.ts
SamanthaAO: I have moved the location of this so that it is run after all the entries instead of being run ever time.

sites/fast-website/src/app/components/side-navigation/side-navigation.ts
SamanthaAO: I have added a function that checks if the page is scrolling up or down. If the page is scrolling down it runs through the array in the standard order meaning that if all sections are visible the one lowest on the page is the current one. If the page is scrolling up the array is reversed meaning that if all sections are visible the one highest on the page one is the current one.

sites/fast-website/src/app/components/side-navigation/side-navigation.ts
```diff
@@ -14,25 +19,52 @@ export class SideNavigation extends FASTElement {
...
     public socialData: CommunityContentPlacementData[] = communityContentPlacementData.filter(
         x => x.header !== "Github"
     );
 
+    reduceFunc = (prev: sectionData, current: sectionData) =>
+        prev.intersectionRatio > current.intersectionRatio ? prev : current;
+
+    scrollDirectionFunc = () => {
+        const currentY: number = window.pageYOffset;
+        const scrollArray: sectionData[] = this.sectionDataArray;
```
nicholasrice, line 37: Just want to make sure you're intending to reverse `this.sectionDataArray` and not a just a local copy of the array

SamanthaAO: I had been intending to just reverse the local copy. Thank you so much for explaining to me the current issues with that on discord. I have added .concat() to when I create the scrollArray variable so that it is a new array instance and does not update this.sectionDataArray.

PR #3410, opened 2020-06-27
fix: react autosuggest arrow keys properly navigate flyout menu

scomea:
# Description
Not sure when this stopped working, but arrow keys were not always navigating up/down the flyout menu as expected.  Fix was to be explicit about forcing focus to the menu on arrow key presses if it is already open.

## Motivation & context
Broken functionality.

## Issue type checklist
- [ x ] **Chore**: A change that does not impact distributed packages.

## Process & policy checklist


packages/react/fast-components-react-base/src/auto-suggest/auto-suggest.tsx
```diff
@@ -546,6 +544,21 @@ class AutoSuggest extends Foundation<
...
+        }
+
+        if (validOption !== null) {
+            const element: HTMLElement = document.getElementById(
+                (validOption as React.ReactElement<ListboxItem>).props["id"]
+            );
+            if (element !== null) {
+                element.focus();
+            }
+        }
```
janechu, line 561: since `this.updateFocusedItem` sets state, it seems like all of this is conceivably happening afterwards, can you reorganize so that all logic is performed and then `setState` is hit once?

scomea: What is the issue?  React batches state updates so I I don't see the need for us to accumlate to one call?

janechu: Readability and maintainability is my primary concern.

I don't believe we should rely on batch state and we should reduce the number of times `setState` is called. Otherwise the code will become very hard to decipher as we have to go through multiple avenues of branching logic to determine all the things a single method might do to manipulate state.

scomea: I disagree.  I very much doubt that there is a perf gain for calling setstate twice to set different variables vs us setting up all our function calls to pass around accumulated state changes for a single call at the end the call stack, and React's mechanism for doing this should be stable and well tested.  Also, I don't see the readability gain with that approach and find that having a single function to toggle something like menu state easy to understand and debug.  I'd agree that perhaps setting the same state var in multiple different places could be difficult to track, as could setting multiple closely related state vars separately.  In any case I think this would be well beyond the scope of this bug fix.

janechu: I didn't mention perf and I'm not advocating for some kind of huge re-write only that we address the logic so that `setState` is called a single time to aide in the branching logic paths issue.

scomea: Yes, and I'm saying that I don't think it should be a requirement that any time a component responds to an event we should worry about it batching all state changes to one call to setState because I don't think it necessarily helps make the logic easier to understand.  I can see it being a good call when setting a bunch of closely related state like viewport positioner's clutch of positioning vars, or certainly a single function should certainly avoid calling setState more than once, but I don't see any gain in conflating state calls that aren't closely intertwined as in this case with tracking the currently focused item and whether the menu is open.  Actually, I see the added complexity of batching results and passing through the call stack to be added complexity we don't need for setting simple state like this when React does it for free.

lilliankravitz:
@scomea thanks for making this fix! Anything I can do to help get this PR reviewed?

PR #3414, opened 2020-06-30
fix: slider thumb does not respond to touch events, vertical sliders track parameters incorrect after scrolling page

marjonlynch:
# Description
* added touch related events for dragging the slider thumb on mobile devices
* issue fixed when the vertical slider was further down the page scrolling could cause the track min and max to be incorrect.

Screen Shot 2020-06-30 at 3 08 59 PM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/slider/slider.ts
+++ b/packages/web-components/fast-foundation/src/slider/slider.ts
@@ -317,6 +317,7 @@ export class Slider extends FormAssociated<HTMLInputElement>
         this.addEventListener("keydown", this.keypressHandler);
         this.addEventListener("mousedown", this.clickHandler);
         this.thumb.addEventListener("mousedown", this.handleThumbMouseDown);
+        this.thumb.addEventListener("touchstart", this.handleThumbMouseDown);
     };
 
     private setupDefaultValue = (): void => {
@@ -341,6 +342,8 @@ export class Slider extends FormAssociated<HTMLInputElement>
         (event.target as HTMLElement).focus();
         window.addEventListener("mouseup", this.handleWindowMouseUp);
         window.addEventListener("mousemove", this.handleMouseMove);
+        window.addEventListener("touchmove", this.handleMouseMove);
+        window.addEventListener("touchend", this.handleWindowMouseUp);
         this.isDragging = true;
     };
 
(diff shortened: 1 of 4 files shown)


packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -408,6 +411,12 @@ export class Slider extends FormAssociated<HTMLInputElement>
                     ? e.pageX - this.getBoundingClientRect().left
                     : e.pageY;
 
+            if (this.orientation === Orientation.vertical) {
```
nicholasrice, line 414: Will this same problem exist in horizontal-scrolling scenarios?

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -408,6 +411,12 @@ export class Slider extends FormAssociated<HTMLInputElement>
...
+                const clientRect: DOMRect = this.track.getBoundingClientRect();
```
scomea, line 415: getBoundingClientRect is an expensive call, you can't just get the clientHeight from track element directly?  I see we also do it on mousemove which is concerning because it can get called a lot.

marjonlynch: I believe we had something else that wasn't getBoundingClientRect before and it hosed up some other situation. I'll give clientHeight a shot and see if it'll work in at least the fixtures we have now though

scomea: Resize observer might be another option.  I'd consider calling getBoundingClientRect on a high frequency event like mousemove a perf red flag.

packages/web-components/fast-foundation/src/slider/slider.ts
marjonlynch: I'm only calling it in mouseDown handler now and I'm doing it for both orientations

packages/web-components/fast-foundation/src/slider/slider.ts
marjonlynch: Thanks to a suggestion from Jane, I'm only doing the getClientBoundingRect in mouseDown and holding on to that for the duration of mousemove, so I believe this should alleviate the perf concern, also fixes the bug.

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -216,7 +222,7 @@ export class Slider extends FormAssociated<HTMLInputElement>
      */
     public disconnectedCallback(): void {
         this.removeEventListener("keydown", this.keypressHandler);
-        this.removeEventListener("mousedown", this.clickHandler);
+        this.removeEventListener("mousedown", this.handleMouseDown);
         this.thumb.removeEventListener("mousedown", this.handleThumbMouseDown);
```
janechu, line 226: should we also remove the touchstart event listener?

marjonlynch: doh! removing

PR #3420, opened 2020-06-30
feat: move appearance attributes of anchor and button out of fast-foundation

chrisdholt:
# Description
Currently, the appearance attribute isn't a standardized attribute across all anchors and buttons. Even if the attribute existed, we don't see semantic alignment across what those appearances are. This change moves the appearance attribute into the implementation for a given design system - FAST and MSFT.

This opens up the ability to extend the foundation anchor or button without that additional attribute. For instance, if I have a navigation item with a fixed appearance, it's less-than-ideal to have an appearance attribute on that component where it's not needed.

## Issue type checklist

For those using fast-foundation and leveraging the appearance attribute, this breaks and would require the addition of the appearance attr along with an `appearanceChanged` implementation to fix. For anyone using the components, this change will be transparent.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components-msft/src/anchor/index.ts
+++ b/packages/web-components/fast-components-msft/src/anchor/index.ts
@@ -1,6 +1,13 @@
-import { customElement } from "@microsoft/fast-element";
+import { attr, customElement } from "@microsoft/fast-element";
 import { Anchor, AnchorTemplate as template } from "@microsoft/fast-foundation";
 import { AnchorStyles as styles } from "./anchor.styles";
+import { ButtonAppearance } from "../button";
+
+/**
+ * Types of anchor appearance.
+ * @public
+ */
+export type AnchorAppearance = ButtonAppearance | "hypertext";
 
 /**
  * The FAST Anchor Element. Implements {@link @microsoft/fast-foundation#Anchor},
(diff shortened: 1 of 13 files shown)


packages/web-components/fast-components-msft/src/anchor/index.ts
```diff
@@ -21,4 +28,23 @@ import { AnchorStyles as styles } from "./anchor.styles";
...
-export class FASTAnchor extends Anchor {}
+export class FASTAnchor extends Anchor {
+    /**
+     * The appearance the anchor should have.
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: appearance
+     */
+    @attr
```
EisenbergEffect, line 39: Have considered doing this without actually adding the attribute? This could be done with a host+attribute selector and would be "cheaper" at runtime.

chrisdholt: This is somewhat step 1 of 2 - the next step is going to be moving away from the styles being associated to a class and attaching the appearance stylesheet based on the appearance "behavior". I could be wrong, but it seems like that would throw a wrinkle in the above.

packages/web-components/fast-foundation/src/button/button.template.ts
```diff
@@ -6,7 +6,7 @@ import { Button } from "./button";
  * @public
  */
 export const ButtonTemplate = html<Button>`
-    <template class="${x => x.appearance}">
+    <template>
```
nicholasrice, line 9: You don't need the `<template>` element if you're not adding any attributes to it.

packages/web-components/fast-foundation/src/anchor/anchor.template.ts
```diff
@@ -7,7 +7,7 @@ import { Anchor } from "./anchor";
  * @public
  */
 export const AnchorTemplate = html<Anchor>`
-    <template class="${x => x.appearance}">
+    <template>
```
nicholasrice, line 10: You don't need the <template> element if you're not adding any attributes to it.

PR #3422, opened 2020-06-30
feat: update typescript version and remove utility types dependencies for react packages

chrisdholt:
# Description
This PR updates typescript and utility-types dependencies to leverage the latest major version across the repo.

## Issue type checklist

## Process & policy checklist


packages/react/fast-components-react-base/package.json
```diff
@@ -112,7 +112,6 @@
     "ts-jest": "^25.4.0",
     "ts-loader": "^4.0.1",
     "typescript": "3.3.3",
-    "utility-types": "^2.1.0",
```
nicholasrice, line 115: I don't think utility-types is needed now that we've updated TypeScript. I think we can simply remove it.

chrisdholt:
> Approved pending the actual need for `utility-types`. I looked over that package and see a number of types that have been added directly to TypeScript. So, if we are using only those types that have been added, I would recommend removing the dependency. Less dependencies === better 😄

We decided to update Typescript and do what we can to drop :)

yarn.lock
```diff
@@ -24000,27 +24000,12 @@ typescript@3.2.x:
...
-  version "3.8.3"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.8.3.tgz#409eb8544ea0335711205869ec458ab109ee1061"
-  integrity sha512-MYlEfn5VrLNsgudQTVJeNaQFUAI7DkhnOjdpAp4T+ku1TfQClewlbSuTVHiA+8skNBgaf02TL/kLOvig4y3G8w==
+typescript@^3.9.0:
+  version "3.9.6"
+  resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.9.6.tgz#8f3e0198a34c3ae17091b35571d3afd31999365a"
+  integrity sha512-Pspx3oKAPJtjNwE92YS05HQoY7z2SFyOpHo9MqJor3BXAGNaPUs83CuVp9VISFkSjyRfiTpmKuAYGJB7S7hOxw==
 
-typescript@^3.7.5, typescript@~3.7.2:
+typescript@~3.7.2:
```
chrisdholt, line 24008: Almost had it.

PR #3433, opened 2020-07-02
feat: anchored region sizing options & other fixes

scomea:
# Description

While building the tooltip component I made some fixes/additions to anchored region.  These include:

- the "scaling" bool has been replaced with a scaling scaling mode enum which allows the region to size itseft according to the content, the anchor element or the available space.
- renamed vars and functions to replace "positioner" with "region" for consistency
- template now has a root "template" element (needed to fix an issue getting ref that had a parent node in dom)

## Motivation & context
- needed changes for tooltip development, first time anchored region excersized in real world scenario

## Issue type checklist

*note: since we're not yet exporting the anchored region in the package I'm not considering the api change around scaling a breaking change

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
@@ -2,7 +2,10 @@ import { html, ref, when } from "@microsoft/fast-element";
 import { AnchoredRegion } from "./anchored-region";
 
 export const AnchoredRegionTemplate = html<AnchoredRegion>`
-    <div class="region" part="region" ${ref("region")} style=${x => x.regionStyle}>
-        ${when(x => x.initialLayoutComplete, html`<slot></slot>`)}
-    </div>
+    <template ${ref("root")}>
+        <div class="region" part="region" ${ref("region")} style=${x => x.regionStyle}>
+            ${when(x => x.initialLayoutComplete, html<AnchoredRegion>`<slot></slot>`)}
+        </div>
+        <template> </template
+    ></template>
 `;
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
@@ -6,6 +6,5 @@ export const AnchoredRegionTemplate = html<AnchoredRegion>`
         <div class="region" part="region" ${ref("region")} style=${x => x.regionStyle}>
             ${when(x => x.initialLayoutComplete, html<AnchoredRegion>`<slot></slot>`)}
         </div>
-        <template> </template
-    ></template>
+    </template>
 `;
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
@@ -190,7 +190,7 @@ export class AnchoredRegion extends FASTElement {
     private resizeDetector: ResizeObserverClassDefinition;
 
     private viewportRect: ClientRect | DOMRect | null;
-    private positionerDimension: Dimension;
+    private regionDimension: Dimension;
 
     private anchorTop: number;
     private anchorRight: number;
@@ -281,7 +281,7 @@ export class AnchoredRegion extends FASTElement {
         this.yTransformOrigin = Location.top;
 
         this.viewportRect = null;
-        this.positionerDimension = { height: 0, width: 0 };
+        this.regionDimension = { height: 0, width: 0 };
 
         this.anchorTop = 0;
         this.anchorRight = 0;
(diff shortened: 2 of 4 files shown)


packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
```diff
@@ -2,7 +2,10 @@ import { html, ref, when } from "@microsoft/fast-element";
...
 
 export const AnchoredRegionTemplate = html<AnchoredRegion>`
-    <div class="region" part="region" ${ref("region")} style=${x => x.regionStyle}>
-        ${when(x => x.initialLayoutComplete, html`<slot></slot>`)}
-    </div>
+    <template ${ref("root")}>
+        <div class="region" part="region" ${ref("region")} style=${x => x.regionStyle}>
+            ${when(x => x.initialLayoutComplete, html<AnchoredRegion>`<slot></slot>`)}
+        </div>
+        <template> </template
```
chrisdholt, line 9: What's with the nested template here?

chrisdholt, line 9: If we made calls to getBoundingClientRect would we need all this markup? I'm concerned that in our attempt to avoid calls to gBCR we're forcing people to work with DOM they would otherwise not include. It seems like that might solve some of the issues we are running up against.

Full transparency, I would take calls to gBCR versus the prescriptive DOM we have here. I'm curious how much it's creating the complexity that's being added below. While I understand that this does complex things, I think we're exacerbating it by doing everything in our power to avoid gBCR.

scomea: removed.

scomea: I don't think using gBCR would change the reasoning behind the markup, it could perhaps simplify some of the component's internal logic but I"m not sure we could avoid calling it too much.

chrisdholt: I may not understand the reasoning for the markup, perhaps you can elaborate on it? 

I think gBCR would end up providing support for Fixed position items where we currently don't have that. This would end up being helpful in scenarios where you have a fixed or sticky header and need to flyout an item from it.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
```diff
@@ -2,7 +2,9 @@ import { html, ref, when } from "@microsoft/fast-element";
 import { AnchoredRegion } from "./anchored-region";
 
 export const AnchoredRegionTemplate = html<AnchoredRegion>`
-    <div class="region" part="region" ${ref("region")} style=${x => x.regionStyle}>
-        ${when(x => x.initialLayoutComplete, html`<slot></slot>`)}
-    </div>
+    <template ${ref("root")}>
```
nicholasrice, line 5: I don't think that you need a ref on the `<template>` - you can just use the `this` keyword to refer to the element in any method.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -18,6 +18,7 @@ declare global {
 }
 
 export type AxisPositioningMode = "uncontrolled" | "locktodefault" | "dynamic";
+export type AxisScalingMode = "anchor" | "availablespace" | "content";
```
nicholasrice, line 21: Thoughts on changing "availablespace" to be "grow"?

scomea: Hmmm....not sure if that's any clearer.  But I think "availablespace" could be improved on.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
scomea: Oh yeah.  Sweet.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
nicholasrice: My concern wasn't so much clarity as verbosity. Both seem equally clear to me but "grow" is much less verbose. Totally open to other ideas though.

scomea: "scale"?

nicholasrice: "fill" could work also maybe

scomea: "fill" is the best yet I think

scomea: Going with "fill" unless someone has a better suggestion

packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
scomea: Primarily, the template wrapper around the region provides a known coordinate space to move the region around in, if we did not have this we'd have to suss out the parent's coordinate space and the region's position within it.

scomea:
Closing as I have more extensive changes in the works.

PR #3437, opened 2020-07-02
feat: allow the autocompletion of a linked data value

janechu:
# Description

This change allows the autocompletion on the input of the linked data control in the `<Form />`.
Now, when `Tab` is pressed the value will autocomplete if there is a single matching item in the linked data list. If `Enter` is pressed the current value will be evaluated against lazily (lowercase values) or strictly (exact value) and if a single match can be found using either of these methods then the value will be added.

An example of where this helps:

Example 1: The list of supplied linked data includes:
```json
[
    "Foo",
    "foo"
]
```

When `"foo"` is typed, it will match `"Foo"` and `"foo"` in the dropdown, however if enter is pressed then it will match `"foo"` exactly and add this value.

Example 2: the list of supplied linked data includes:
```json
[
    "foo-Bar",
    "foo",
]
```

When `"foo-bar"` is typed it will match `"foo-Bar"` lazily and when enter is pressed it will add the linked data.
Likewise if `"foo-"` is typed and then `Tab` is pressed, it will autocomplete to `"foo-Bar"` and the next enter is pressed it will add the linked data.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
+++ b/packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
@@ -288,28 +288,28 @@ class LinkedDataControl extends React.Component<
      * Change handler for editing the search term filter
      */
     private handleSearchTermUpdate = (e: React.ChangeEvent<HTMLInputElement>): void => {
-        const normalizedValue: string = e.currentTarget.value.toLowerCase();
+        const normalizedValue: string = e.target.value.toLowerCase();
         const hasSingleMatchedValue = this.matchExactValueWithASingleSchema(
-            e.currentTarget.value
+            e.target.value
         );
 
         // If an exact match is available, the linked data will be added
         if (hasSingleMatchedValue) {
-            this.addLinkedData(normalizedValue, e.currentTarget.value);
+            this.addLinkedData(normalizedValue, e.target.value);
         }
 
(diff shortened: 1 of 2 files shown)


packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
```diff
@@ -212,27 +212,71 @@ class LinkedDataControl extends React.Component<
...
+                    this.addLinkedData(normalizedValue, e.currentTarget.value);
+
+                    /**
+                     * Adding items to the linked data causes the items to
+                     * move the input down while the datalist remains in the same location,
+                     * to prevent the datalist from overlapping the input
+                     * the datalist is dismissed by defocusing and refocusing the input
+                     */
+                    (e.target as HTMLElement).blur();
+                    (e.target as HTMLElement).focus();
```
scomea, line 235: I worry the blur/focus trick could cause strangeness with accessibility.  Won't tools like Narrator announce the focus event and would that be confusing?

janechu: I've checked other accessibility voice over tools and found no strangeness. I think there's definitely room for improvement, but this single trick doesn't seem to make it worse.

PR #3441, opened 2020-07-06
fix: removed extra padding around fast card component

### LINKED ISSUE #3440: Border padding being added in fast-card
Trying to use fast-card to create a new card type.
Seeing a small amount of padding being added around the border of the card.

For my design requirements I need my image content to extend all the way to the border edge.
Can this default styling be changed or overwritten?

image
image

khamudom:
# Description

There was a transparent border set around the card that is only visible in high contrast.
High contrast work will be done in a separate PR

## Motivation & context

closes #3440 

before
image
after
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components-msft/src/card/card.styles.ts
+++ b/packages/web-components/fast-components-msft/src/card/card.styles.ts
@@ -1,8 +1,7 @@
 import { css } from "@microsoft/fast-element";
-import { display, forcedColorsStylesheetBehavior } from "@microsoft/fast-foundation";
+import { display } from "@microsoft/fast-foundation";
 import { elevation } from "../styles";
 import { neutralLayerCardBehavior } from "../styles/index";
-import { SystemColors } from "@microsoft/fast-web-utilities";
 export const CardStyles = css`
     ${display("block")} :host {
         --elevation: 4;
--- a/packages/web-components/fast-components-msft/src/card/card.styles.ts
+++ b/packages/web-components/fast-components-msft/src/card/card.styles.ts
@@ -15,15 +15,4 @@ export const CardStyles = css`
         border-radius: calc(var(--elevated-corner-radius) * 1px);
         ${elevation};
     }
-`.withBehaviors(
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-components-msft/src/card/card.styles.ts
```diff
@@ -13,6 +14,16 @@ export const CardStyles = css`
...
+    neutralLayerCardBehavior,
+    forcedColorsStylesheetBehavior(
+        css`
+            :host {
+                forced-color-adjust: none;
+                background: ${SystemColors.Canvas};
+                box-shadow: 0 0 0 1px ${SystemColors.CanvasText};
+            }
+        `
+    )
```
chrisdholt, line 28: This looks like a feature - should we call out that you're including high contrast styles here?

chrisdholt: Should these be seperate PR's? Once appears to be a fix, the other appears to be a feature.

khamudom: I put them in one as to not regress the high contrast work when fixing the padding bug. But I am all for making them separate PRs.

packages/web-components/fast-components-msft/src/card/card.styles.ts
```diff
@@ -1,7 +1,8 @@
 import { css } from "@microsoft/fast-element";
-import { display } from "@microsoft/fast-foundation";
+import { display, forcedColorsStylesheetBehavior } from "@microsoft/fast-foundation";
 import { elevation } from "../styles";
 import { neutralLayerCardBehavior } from "../styles/index";
+import { SystemColors } from "@microsoft/fast-web-utilities";
```
chrisdholt, line 5: @khamudom I know these are coming in with the other PR, but they should be removed from here as they are currently referenced.

khamudom: shoot. thank you!

PR #3444, opened 2020-07-07
fix: removed accessibility attributes and added type to input for text field component

### LINKED ISSUE #3436: Narrator and NVDA is not reading out the fast-text-field component
While running Narrator or NVDA, as you tab onto the fast-text-field component, it is silent.

Example:

1. Run Narrator or NVDA
2. Go to https://explore.fast.design/components/fast-text-field'
3. tab onto the text-field. Observe that the screen reader not saying anything.

Would expect the screen reader to say "edit"

Windows 10
Microsft Edge browser

* Browser [e.g. Edge, Chrome, Safari, FireFox]
* Version [e.g. 1.8.0]

khamudom:
# Description

Since we are using `delegatesFocus ` option, the focus is being set inside the custom element and not on the host element, I removed the aria and role attributes from the template and added the `type` attribute to the input.

## Motivation & context
closes #3436 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/text-field/text-field.template.ts
+++ b/packages/web-components/fast-foundation/src/text-field/text-field.template.ts
@@ -8,11 +8,7 @@ import { TextField } from "./text-field";
  */
 export const TextFieldTemplate = html<TextField>`
     <template
-        role="textbox"
         appearance="${x => x.appearance}"
-        aria-required="${x => x.required}"
-        aria-disabled="${x => x.disabled}"
-        aria-readonly="${x => x.readOnly}"
         tabindex="${x => (x.disabled ? null : 0)}"
         class="
             ${x => x.appearance}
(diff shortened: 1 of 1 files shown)


khamudom:
@EisenbergEffect  No, this one is on the text field component. The other issue was NVDA not recognizing aria-label on host element. The issue for this is for screen readers like Narrator or NVDA is not recognizing the input of the text field. So it doesn’t say anything.

chrisdholt:
> @khamudom Is this related to the screen reader bug we discussed last week? For my own information, is this the expected behavior according to spec that when focus is delegated the aria info will be picked up from the control that is delegated to? Also, did we test this on browsers that don't yet support delegated focus, to make sure this change works correctly in those cases?

@eisenbergeffect, I think it likely is related and there may be some things we'll need to figure out given this nuance. @nicholasrice linked out to [this explainer](https://github.com/TakayoshiKochi/tabindex-focus-navigation-explainer/blob/master/TabindexFocusNavigationExplainer.md) which detail delegatesFocus but I don't see what we are dealing with explicitly called out. That being said, this _seems_ to be intuitive behavior...if we delegate focus within the host, AT sees that as the focusable element as well and announces that element. Given that, I think this may explain the behavior of both `fast-anchor` and `fast-button` which also delegate focus.

PR #3452, opened 2020-07-08
feat: simplify rollup configs and compress tagged template literals

radium-v:
# Description
This PR modifies the rollup configs for `fast-element`, `fast-foundation`, `fast-components`, and `fast-components-msft`.

- the main dist file and the minified version are run in parallel for each package
- CSS-tagged template literals (``css`...``) are compressed with [rollup-plugin-minify-tagged-css-template](https://github.com/notlmn/rollup-plugin-minify-tagged-css-template)
- HTML-tagged template literals (``html`...``) are compressed with [rollup-plugin-transform-tagged-template](https://github.com/notlmn/rollup-plugin-transform-tagged-template)
- Quotes are now present on attribute values in templates, though this is more for consistency (compression should still work fine without them)

## Motivation & context
Our bundles weren't outputting the correct file sizes when processed with rollup, and the remaining files (namely `.min.js` bundles) had a lot of newlines and spaces in them.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components-msft/src/styles/patterns/button.ts
```diff
@@ -21,7 +21,7 @@ import {
     neutralOutlineActiveBehavior,
     neutralOutlineHoverBehavior,
     neutralOutlineRestBehavior,
-} from "../index";
+} from "../behaviors";
```
EisenbergEffect, line 24: Good catch!

radium-v:
> How hard do you think it would be to get something similar working with Webpack as well?

I haven't found any existing plugins that work in the same way as these (with template literal chunks rather than trying to handle the entire string), so I suppose a new webpack loader/plugin with the same AST-based method that can handle the template literal chunks individually would be a good solution to take on next.

PR #3468, opened 2020-07-08
fix: reflect aria-hidden attribute in the DOM for the flipper component

### LINKED ISSUE #3464: [fast-flipper] should only have a role of button when *not* hidden from AT
Fast flipper is essentially a presentational component when hidden from AT. Having role of button implies to AT that the element should be focusable but isn't which elevates this to an error. We should only apply role="button" when this is exposed to AT. When hidden, it's simply a convenience feature for sighted users.

khamudom:
# Description

Removed `mode: "fromView"` to reflect the value to the DOM.

## Motivation & context

closes #3464 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/flipper/flipper.ts
+++ b/packages/web-components/fast-foundation/src/flipper/flipper.ts
@@ -35,6 +35,11 @@ export class Flipper extends FASTElement {
      */
     @attr({ attribute: "aria-hidden", mode: "fromView", converter: booleanConverter })
     public hiddenFromAT: boolean = true;
+    private hiddenFromATChanged(): void {
+        if (!this.hiddenFromAT) {
+            this.setAttribute("role", "button");
+        }
+    }
 
     /**
      * The direction that the flipper implies navigating.
--- a/packages/web-components/fast-foundation/src/flipper/flipper.ts
+++ b/packages/web-components/fast-foundation/src/flipper/flipper.ts
@@ -33,13 +33,8 @@ export class Flipper extends FASTElement {
      * @remarks
      * HTML Attribute: aria-hidden
      */
-    @attr({ attribute: "aria-hidden", mode: "fromView", converter: booleanConverter })
+    @attr({ attribute: "aria-hidden", converter: booleanConverter })
     public hiddenFromAT: boolean = true;
-    private hiddenFromATChanged(): void {
-        if (!this.hiddenFromAT) {
-            this.setAttribute("role", "button");
-        }
-    }
 
     /**
      * The direction that the flipper implies navigating.
--- a/packages/web-components/fast-foundation/src/flipper/flipper.template.ts
+++ b/packages/web-components/fast-foundation/src/flipper/flipper.template.ts
@@ -7,6 +7,7 @@ import { Flipper, FlipperDirection } from "./flipper";
  */
 export const FlipperTemplate = html<Flipper>`
     <template
+        role="button"
         aria-disabled="${x => (x.disabled ? true : void 0)}"
         tabindex="${x => (x.hiddenFromAT ? -1 : 0)}"
         class="${x => x.direction} ${x => (x.disabled ? "disabled" : "")}"
--- a/packages/web-components/fast-foundation/src/flipper/flipper.template.ts
+++ b/packages/web-components/fast-foundation/src/flipper/flipper.template.ts
@@ -7,7 +7,6 @@ import { Flipper, FlipperDirection } from "./flipper";
  */
 export const FlipperTemplate = html<Flipper>`
     <template
-        role="button"
         aria-disabled="${x => (x.disabled ? true : void 0)}"
         tabindex="${x => (x.hiddenFromAT ? -1 : 0)}"
         class="${x => x.direction} ${x => (x.disabled ? "disabled" : "")}"


packages/web-components/fast-foundation/src/flipper/flipper.ts
```diff
@@ -35,6 +35,11 @@ export class Flipper extends FASTElement {
      */
     @attr({ attribute: "aria-hidden", mode: "fromView", converter: booleanConverter })
     public hiddenFromAT: boolean = true;
+    private hiddenFromATChanged(): void {
+        if (!this.hiddenFromAT) {
+            this.setAttribute("role", "button");
```
nicholasrice, line 40: It is dangerous to call `setAttribute` as a reaction to an observable change because it can call `setAttribute` before the element constructs, which throws a runtime error.

you'll want to import `DOM` from `fast-element` and use `DOM.queueUpdate()` to queue an callback that adds the attribute

chrisdholt: @khamudom could this live in the template conditionally based on if hiddenFromAt exists? 

```
role=${x => (x.hiddenFromAT ? "button" : void 0)}"
```

chrisdholt:
> My other comment should be addressed before this goes in.
> 
> I am curious to the _why_ of this change as well. Is there a bug this is solving? The linked issue states that the flipper shouldn't have a role when hidden - is true? Presumably an element could have a role of `button` and be `aria-hidden` as well - the two aren't mutually exclusive. @chrisdholt can you elaborate on the _why_ of the linked issue?

The current implementation defaults to hiding this control from AT as it is most often a form of duplicate navigation. More on this can be found in the spec or the PR for the spec (I believe we discussed the purpose here at length). The method of exposing is to pass aria-hidden="false":

```ts
    /**
     * Indicates the flipper should be hidden from assistive technology. Because flippers are often supplementary navigation, they are often hidden from assistive technology.
     *
     * @public
     * @defaultValue - true
     * @remarks
     * HTML Attribute: aria-hidden
     */
    @attr({ attribute: "aria-hidden", mode: "fromView", converter: booleanConverter })
    public hiddenFromAT: boolean = true;
```


We current do not reflect this attribute, which is likely the reason we have the bug. There is no aria-hidden value so AT does not treat it as hidden and instead throws an error.

It does occur to me that we could likely solve this by simple reflecting the value to the DOM. I think in most cases, we don't want to reflect unused attributes to the DOM, but in this case it makes sense. A change does need to be made here, and I think the most elegant change is likely to just remove the mode of fromView and allow this to default to "reflect".

```ts
    /**
     * Indicates the flipper should be hidden from assistive technology. Because flippers are often supplementary navigation, they are often hidden from assistive technology.
     *
     * @public
     * @defaultValue - true
     * @remarks
     * HTML Attribute: aria-hidden
     */
    @attr({ attribute: "aria-hidden", converter: booleanConverter })
    public hiddenFromAT: boolean = true;
```

packages/web-components/fast-foundation/src/flipper/flipper.ts
khamudom: yeah, ideally I was thinking of the making the change in the template but wasn't sure of the approach, and I went with something familiar. Thanks @chrisdholt 
@nicholasrice I am going undo and remove the function I added and just make the change in the template. 

Thanks!

chrisdholt: @khamudom, take a look at my comment below - I think we can skip the template change if we reflect the attribute.

EisenbergEffect: On a related note, I think we should make an improvement to these callbacks somehow, so that making this mistake isn't possible. The first pass on the provider work inside fast-element will adjust some timings. As part of that, I can look at delaying these callbacks until just before the initial bind in the connected callback, for example, so that they don't run during construct. Or we could prevent them from running at all if the value is changed in the constructor, since that would be an initialization, rather than a change, and we could add a beforeBind lifecycle hook for special cases. I'd love to hear people's thoughts on those ideas. I'll try to put together an issue with various options.

packages/web-components/fast-foundation/src/flipper/flipper.ts
```diff
@@ -35,6 +35,11 @@ export class Flipper extends FASTElement {
      */
     @attr({ attribute: "aria-hidden", mode: "fromView", converter: booleanConverter })
```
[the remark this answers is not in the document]
khamudom: dropped the mode and reverted the removal of the role attribute

packages/web-components/fast-foundation/src/flipper/flipper.ts
nicholasrice: Conceptually I don't think of initialization as a property *change* so to me it would make sense not to run the code during initialization / construction. That said, I think there many cases where an author (ourselves included) would want the same code to run during initialization *and* change. 

I think a `beforeBind` could work so long as it is easy to associate the same code to `beforeBind` as it is to the change handler. I think a config option on the decorator could work, controlling if it fires during initialization. I also like the idea of simply deferring until after construction. If an author *needs* code to run during construction that work can be done in the constructor - I don't think any of the cases we've run into would break if the change handler is deferred.

EisenbergEffect: Maybe don't run change handlers during the constructor, but you can opt in to run them once automatically right before bind. Then also add a beforeBind hook for custom logic that runs after the change handlers. Then after that you bind the view....hmm. I can also see splitting the view bind into two phases so that we could run an early phase to set up all the refs (and maybe the attr delegate if we add that), then let the change callbacks fire and the beforeBind hook, and then finally bind the rest of the template once that setup is done.

I've got a todo to write up an official proposal for this. I'll keep thinking about it.

PR #3470, opened 2020-07-09
feat: adds mixin to support global aria-* attributes for components delegating focus

### LINKED ISSUE #3466: Non-enumerated ARIA attributes are not handled when delegating focus
There are a few sets of components which leverage delegateFocus. When focus is delegated away from the host element, assistive technology ignores the host element and looks for any aria attributes, roles, etc on the element to which focus has been delegated. Currently, this causes an issue where HTML attributes are not enumerated. For instance, given the below example of anchor, the aria-label will not be read by assistive technology and the element will not have an accessible name.

```html
<fast-anchor href="/foo" aria-label="Foo"></fast-anchor>
```

To accommodate this, all HTML attributes need to be enumerated and handled with change events to ensure that they respond when the host element is updated.

chrisdholt:
# Description
This PR does initial work to add global aria-* attributes to support passing those attributes to components which delegate focus to an internal element. There may be additional attributes added in the future, but this solves the immediate need. Additionally, there may be other attributes added to text field or textarea, but additional investigation is needed into that.

## Motivation & context
closes #3466 

## Issue type checklist

## Process & policy checklist


chrisdholt:
Converting to draft temporarily to add documentation and update the definition and configuration files.

nicholasrice:
@chrisdholt I'm wondering if we should go a bit further with this and re-organize the mixin classes to align closer to the ARIA widget structure. The attributes that should be forwarded are directly related to the role of the element that focus is being delegated to and that relationship isn't captured with a single "DelegatesFocus" mixin.

What are your thoughts on renamaing `DelegatesFocus` to `ARIARoleType` - this aligns directly to https://www.w3.org/TR/wai-aria-1.1/#roletype. A `DelegatesARIALink` and `DelegatesARIAButton` can extend that type, where button adds `aria-expanded` and `aria-pressed` and link ands only `aria-expanded`. The main idea here would be to scaffold the type along the actual ARIA role / inherited properties / supported properties taxonomy. 

Also not set on any of the above names - there are likely better ones but I think they illustrate the idea.

chrisdholt:
> @chrisdholt I'm wondering if we should go a bit further with this and re-organize the mixin classes to align closer to the ARIA widget structure. The attributes that should be forwarded are directly related to the role of the element that focus is being delegated to and that relationship isn't captured with a single "DelegatesFocus" mixin.
> 
> What are your thoughts on renamaing `DelegatesFocus` to `ARIARoleType` - this aligns directly to https://www.w3.org/TR/wai-aria-1.1/#roletype. A `DelegatesARIALink` and `DelegatesARIAButton` can extend that type, where button adds `aria-expanded` and `aria-pressed` and link ands only `aria-expanded`. The main idea here would be to scaffold the type along the actual ARIA role / inherited properties / supported properties taxonomy.
> 
> Also not set on any of the above names - there are likely better ones but I think they illustrate the idea.

@nicholasrice I think we can make an organizational change, but I wonder if it would create confusion as to why only certain elements require `ARIARoleType` and other elements don't leverage something so generic. What about something like `DelegatesARIA`? That name makes more intuitive sense to what is going on and the intent of having the class. Then we could have `DelegatesARIAButton`, `DelegatesARIALink`? Alternatively, the class currently is `GlobalARIA` - the `DelegatesARIALink`, etc could extend `GlobalARIA`. Either of those seem like they would preserve the  intent you propose, but the semantics might be more intuitive. I'm partial to something like `DelegatesARIA` because it's quite literal to what is going on and why. Thoughts on either of those directions?

nicholasrice:
My primary goal is to align our exported structures with the actual names and hierarchy used by ARIA. I'm not a fan of `GlobalARIA` because AFAIK there is no such thing as "global" ARIA attributes - the attributes you currently have enumerated are provided from the [roletype](https://www.w3.org/TR/wai-aria-1.1/#roletype.) abstract ARIA role.

I'm fine with prepending `Delegates` if that clarifies the use case.

packages/web-components/fast-foundation/docs/api-report.md
```diff
@@ -200,6 +203,29 @@ export interface DecoratorDesignSystemPropertyConfiguration extends Omit<Decorat
     default: any;
 }
 
+// @public
+export class DelegatesFocus {
+    ariaAtomic: "true" | "false";
```
nicholasrice, line 208: Need to re-build docs

chrisdholt: aha

chrisdholt: This should be good and everything should be up to date

Issue #3305, opened 2020-06-15
Deprecate React Packages


nicholasrice:
In the interests of platform alignment and prioritization of Web Components, FAST collaborators are proposing the deprecation of the following packages:

- `@microsoft/fast-components-class-name-contracts-base`
- `@microsoft/fast-components-class-name-contracts-msft`
- `@microsoft/fast-foundation-react`
- `@microsoft/fast-components-react-base`
- `@microsoft/fast-components-react-msft`
- `@microsoft/fast-components-styles-msft`
- `@microsoft/fast-jss-manager`
- `@microsoft/fast-jss-manager-react`
- `@microsoft/fast-jss-utilities`
- `@microsoft/fast-layouts-react`
- `@microsoft/fast-react-utilities`
- `@microsoft/fast-storybook-design-system-addon`
- `@microsoft/fast-storybook-presets`
- `@microsoft/fast-tslint-rules` *this isn't specific to React, just needs to be deprecated*

**Steps for deprecation:**

1. In the interests of preserving the repository history, a fork of the FAST repository will be made under the name **microsoft/fast-dna-react.** The forked repository will be immediately archived using the [github archive feature]([https://help.github.com/en/github/creating-cloning-and-archiving-repositories/archiving-repositories](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/archiving-repositories)). The archived repository will exist under the Microsoft domain.
2. The existing FAST repository will add language to the repository README.md that the above packages have been deprecated and can be viewed / accessed from the archived repository. A link to the archived repository will be provided.
3. All of the deprecated packages will have a deprecation notice added to their individual README.md files. The deprecated packages will then be published a final time so the deprecation notice is visible from the NPM registry.

**What about React support?**

FAST intends to support React through an interop layer written as a React component. A good summary of the challenges [can be found here](https://custom-elements-everywhere.com/libraries/react/results/results.html), and our own [investigation summarized here](https://github.com/microsoft/fast/issues/3389).

Shivam60:
i think a link is missing in `A good summary of the challenges can be found here.`

nicholasrice:
It sure was! Thanks @Shivam60 - updated.

chrisdholt:
@nicholasrice, before we deprecate we'll likely need to remove the dependency on `fast-components-styles-msft` from `fast-components-msft`. Do we have an issue tracking that?

nicholasrice:
[...]

PR #3472, opened 2020-07-09
fix: fast-components type ramp should use design system defaults

CuddleBunny:
# Description

Type ramp values in `fast-design-system-provider` are hard coded rather than using values set in `fastDesignSystemDefaults`.

## Motivation & context

Fixes unexpected behavior of design system property values not changing when defaults are changed.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/design-system-provider/index.ts
+++ b/packages/web-components/fast-components/src/design-system-provider/index.ts
@@ -164,109 +164,109 @@ export class FASTDesignSystemProvider extends DesignSystemProvider
 
     @designSystemProperty({
         attribute: "type-ramp-minus-2-font-size",
-        default: "10px",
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@nicholasrice @chrisdholt I see that we also have these values hard-coded in the -msft design system provider, but I didn't see an equivalent "defaults" object to derive the values from. Should we update the -msft version to use the same pattern?

nicholasrice:
Hmm yea - we have a disconnect because the `-msft` design system mostly inherits from `@microsoft/fast-components-styles-msft`. I think we could look at divorcing from that package as part of https://github.com/microsoft/fast/issues/3305

EisenbergEffect:
Ah, yes. Thanks @nicholasrice !

PR #3478, opened 2020-07-09
fix: removed aria-required from switch

### LINKED ISSUE #3458: [fast-switch] Switch implements aria-required which is not allowed
The switch component implements `aria-required` which is not allowed per https://www.w3.org/TR/wai-aria-1.1/#switch. The aria-required attribute should be removed from the foundational component.

SethDonohue:
# Description
Fixes #3458 

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/switch/switch.template.ts
```diff
@@ -9,7 +9,6 @@ export const SwitchTemplate = html<Switch>`
     <template
         role="switch"
         aria-checked="${x => x.checked}"
-        aria-required="${x => x.required}"
```
nicholasrice, line 12: hmmm - I'd expect that this component shouldn't support a `required` prop / attribute either then. It inherits that property from `FormAssociated` though - I'll write up a quick issue to track

PR #3482, opened 2020-07-09
fix: add aria label to the components in fast frame section on the website

### LINKED ISSUE #3465: [fast-website] Form components are missing labels
On the website, many of the form components are missing labels. This is creating accessibility issues as users do not have context for those elements.

All form elements should have a corresponding label or an aria label to give it an accessible name.

To preserve the design integrity, we should likely add aria-labels to the form elements and any other elements missing accessible names in the "fast-frame" portion of the website.

khamudom:
# Description

Since all of these components in fast-frame are considered examples, to be consistent, they all prefixed with "Example"

Changed to read "dark mode" as that is more specific than "mode"
image
 
Notes:
- I would like to update the slider component to have something like an "aria-value-now-unit" attribute. To be able to have the screen reader also read out the type of unit, like pixels, inches, etc. Currently, as you change the numbers on the slider, for example, border-radius, visualizing "3 pixels" is better than "3".
image
- There is know bug that NVDA will not read out aria-label on host element components, like an anchor, button, and text field.

## Motivation & context

Fixes #3465 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
+++ b/sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
@@ -5,8 +5,6 @@ import ContrastIcon from "svg/icon-contrast.svg";
 import DownloadIcon from "svg/icon-download.svg";
 import PaletteIcon from "svg/icon-palette.svg";
 import PlayIcon from "svg/icon-play.svg";
-import ScreenIcon from "svg/icon-screen.svg";
-import ShareIcon from "svg/icon-share.svg";
 import SwatchesIcon from "svg/icon-swatches.svg";
 import { FastFrame } from "./fast-frame";
 import { ColorHSL, hslToRGB } from "@microsoft/fast-colors";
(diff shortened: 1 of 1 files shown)


sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -62,6 +60,7 @@ export const FastFrameTemplate = html<FastFrame>`
                                     html<string>`
                                         <site-color-swatch
                                             tabindex="0"
+                                            aria-label="background color"
```
nicholasrice, line 63: Should these read the hexadecimal color as part of the aria-label so that information is accessible?

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -5,8 +5,6 @@ import ContrastIcon from "svg/icon-contrast.svg";
 import DownloadIcon from "svg/icon-download.svg";
 import PaletteIcon from "svg/icon-palette.svg";
 import PlayIcon from "svg/icon-play.svg";
-import ScreenIcon from "svg/icon-screen.svg";
```
radium-v, line 8: Should these files be removed as well? Doesn't look like they're being used anywhere else in the project.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
khamudom: yep, thanks for bringing that up @nicholasrice . I should have made a note that it was an option I also had. Will make the change.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
khamudom: I removed the ones that are not being used

PR #3484, opened 2020-07-09
feat: adds menu and menu item styling to MSFT web component package

eljefe223:
# Description
Adds menu and menu item styling to MSFT web component package.

image

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components-msft/src/menu-item/menu-item.styles.ts
```diff
@@ -0,0 +1,134 @@
...
+    neutralFillStealthActiveBehavior,
+    neutralFillStealthHoverBehavior,
+    neutralFillStealthRestBehavior,
+    neutralFocusBehavior,
+    neutralForegroundRestBehavior,
+    forcedColorsStylesheetBehavior(
+        css`
+            :host {
+                border-color: transparent;
+                forced-color-adjust: none;
```
khamudom, line 101: To also make sure the colors are correct at rest in high contrast, on `:host`, could you also add,
```
color: ${SystemColors.ButtonText};
fill: ${SystemColors.ButtonText};
```

eljefe223: Thanks Kham!

packages/web-components/fast-components-msft/src/menu/menu.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+    neutralLayerFloatingBehavior,
+} from "../styles/index";
+
+export const MenuStyles = css`
+    ${display("block")} :host {
+        --elevation: 11;
+        background: ${neutralLayerFloatingBehavior.var};
+        border: calc(var(--outline-width) * 1px) solid transparent;
+        border-radius: var(--elevated-corner-radius);
+        ${elevation} margin: 0;
```
EisenbergEffect, line 15: Is `elevation` correct here?

eljefe223: That does looks wrong, I copied this from `fast-components` and made the necessary color changes. so its wrong there to.

eljefe223: needs a semicolon and a return

EisenbergEffect:
@eljefe223 Sorry to send you on a wild goose chase! It looks like `elevation` includes the semicolon already. Here's the source:

```ts
export const elevation = `box-shadow: ${ambientShadow}, ${directionalShadow};`;
```

So, I think we need to remove the additional semicolon now 😢  But, let's keep the newline that you added because I think that makes it clearer that there's a full property being set there. Without the newline, it looked like an error (even though it technically was now).

Apologies!

eljefe223:
> @eljefe223 Sorry to send you on a wild goose chase! It looks like `elevation` includes the semicolon already. Here's the source:
> 
> ```ts
> export const elevation = `box-shadow: ${ambientShadow}, ${directionalShadow};`;
> ```
> 
> So, I think we need to remove the additional semicolon now 😢 But, let's keep the newline that you added because I think that makes it clearer that there's a full property being set there. Without the newline, it looked like an error (even though it technically was now).
> 
> Apologies!

That's interesting @EisenbergEffect it appears we add the semicolon elsewhere we call `${elevation}`. I'll grab those as part of this.

EisenbergEffect:
Not for this PR, but I think we may want to think about establishing some naming conventions for these constants. Just looking at something, it's hard to tell if it's a var, a literal, or a full property and value. So, it's easy to make a mistake in writing CSS and also difficult when reading it.

packages/web-components/fast-components-msft/src/menu-item/menu-item.styles.ts
```diff
@@ -0,0 +1,135 @@
...
+        grid-column-start: 2;
+        justify-self: start;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+    .start,
+    .end,
+    ::slotted(svg) {
+        ${
```
nicholasrice, line 76: @radium-v I think I saw you removing the interpolation of strings in css template literals in your minification PR - is this still okay to do?

radium-v: It'll lead to an expression like `${""}` since the transformer handles the string chunks (`.start,.end,::slotted(svg){`) and the regular JS compression removes the comment in the expression.

Unfortunately, I don't think the transformers in place right now are able to remove the comment if it were part of the CSS, since the string chunks can't be processed as real CSS or HTML (we end up with parts like `;\n    color: ` and `"\n    attribute="` which aren't really parsable as anything on their own).

From a technically-focused perspective, the `${ /* comment */ ""}` is the one that collapses to the fewest bytes, but it still can invoke more cycles depending on how many there are if the engine has to stop and fill in all of those gaps at runtime.

PR #3488, opened 2020-07-10
chore: adds React integration guide to documentation

### LINKED ISSUE #3389: FAST React Wrapper
## Goals
1. Provide an interop layer to use Web Components in React and Preact applications.
2. Define general design-patterns for React/Preact components so that a component implementation can be either React or wrapped Web Component, and changed between the two without changing application code.

## Resources:

- [https://www.sitepen.com/blog/wrapping-web-components-with-react/](https://www.sitepen.com/blog/wrapping-web-components-with-react/)
- [https://github.com/facebook/react/issues/11347](https://github.com/facebook/react/issues/11347)

## Events:

React uses a synthetic event system that is incompatible with the custom event triggering used by WC. It is possible to partially circumvent this with a wrapping component by exposing an `on{Event}` prop and attaching an event listener imperatively. An important distinction in this approach is that this approach *does not support bubbling or capturing;* It simply proxies a listener on the wrapping component. **This means that a listener for a custom-element event *must* be set on the custom-element wrapper itself**; a listener cannot be added higher in the React tree. See a test case of this [un-supported scenario](https://codesandbox.io/s/react-synthetic-event-bubbling-xfq2u?file=/src/App.js).

> **Note** The non-bubbling / capturing behavior is due to React's synthetic event system. This means that event listeners in *Preact* apps will not behave congruously with those in *React* apps. See [this example](https://codesandbox.io/s/react-synthetic-event-bubbling-1wt0y) of a *Preact* app catching a bubbling custom event from a WC.

## Property vs Attribute binding:

A heuristic similar to Preact will need to be implemented to circumvent React's simplistic model of passing all data to attributes, which does not work for all data types. For reference, [Preact's heuristic lives here](https://github.com/preactjs/preact/blob/c317a96f80b93cc31a4e1b81c722221493165491/src/diff/props.js#L59). [`reactify-wc`](https://www.npmjs.com/package/reactify-wc) implements a slightly less complicated heuristic which may also be appropriate.

### `className`

It is very easy for React to clobber any classes applied by the web-component code as is illustrated [https://codesandbox.io/s/react-class-name-clobbering-o9jwl](https://codesandbox.io/s/react-class-name-clobbering-o9jwl)

**Potential solutions:**

- Avoid adding classes from the WC implementation, favoring stylesheet composition instead, or applying styles based on attribute selectors. eg `:host([appearance=lightweight])` instead of `:host(.lightweight)`
- Track class list in state and manually manage the property, ensuring not to clobber any classes added by the component itself
- ?

### Indirect attribute -> property names
It is somewhat common for the IDL attribute name to not match the content attribute controlling the IDL attribute. We will need a mechanism to establish relationships between IDL attributes and content attributes when names are not identical. 

### Coerce application method
We'll likely encounter scenarios where we'll need to force an incoming prop to be applied as a IDL or content attribute.

### Unhandled props

Unhandled props exists to pass arbitrary attributes and properties to the root element or component. This is generally done to accomplish accessibility requirements or to attach telemetry data. There shouldn't be any significant change due to this behavior - the wrapper will implement the prop heuristic to apply the attribute or property to the element.

### Virtual DOM synchronization

Components that set attributes on external elements (not in Shadow DOM) create opportunity for React's virtual DOM to fall out of sync with the real DOM, creating a variety of issues (`classNames` for instance).

In general, the wrapper will shield React from the element implementation and will serve as the point of severance from React's virtual DOM. To React, the element will appear as a custom element with a single ref prop applied. This will help avoid issues caused by real DOM → virtual DOM mismatches.

The exception to this will be during SSR. React will not capture the SSR string correctly if attributes do not exist in the virtual DOM when being rendered. See section on Server Side Rendering (SSR).

## Children, slots, and render props.
The `children` React property will not have any special behavior - `children` will be rendered inside of a Web Component automatically by React.

Web Components and React both provide a mechanism to provide keyed children to a component. . Web Components support [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) content where React *generally* implements a [render prop](https://reactjs.org/docs/render-props.html) model, where a function is provided that *returns* an element - that function is invoked where appropriate in a component's markup code. For example:

```tsx
// React
function Button(props) {
  return (
    <button>
      {props.icon() /*render prop */} 
      {props.children}
    </button>
  )
}
```
```html
<!-- Web Component Template -->
<button>
  <slot name="icon"></slot>
  <slot></slot>
</button>
```

This presents several complications:
1. The wrapper will need to be able to map a render prop to a slot name, rendering the product of the render prop with `[slot={name}]`
2. Render props are often invoked with arguments. A slotted element is just an element - no control logic or parameterized input. *In FAST - the paramater is often a class-name so that the slotted element can be styled correctly for the control. In WC this requirement is handled by the `::slotted` CSS selector. There are likely other cases but class-names is the most common.*

**Potential Solutions:** 
1. Use a sigil or naming convention to map render props to slots. (eg `props.beforeSlot` -> `element[slot="before"]`). **This would also require not expecting arguments to the render prop.**
    - A related approach is to expect React elements as children instead of function types - this would reduce ambiguity around expecting arguments for render props.
2. Components expect slots to be applied by React code to children. This would work as expected in WC but we would need to update any React components to surface the same API - extracting child elements from `props.children` and putting them in the appropriate place.

## Refs
Refs are typically used for several purposes:
1. Invoking imperative methods
2. Accessing component properties (not common in React but common in WC)
3. Managing document focus on `HTMLElements`

We'll need to determine what the most intuitive model for ref resolution is for wrapped components. Both options below presume we desire ref support for wrapped Web Components.

> **Note.** There is already opportunity for mis-alignment between React/Preact implementations and wrapped WC with the above presumption. If a React/Preact component implementation is a function component, then it cannot accept a ref, where a wrapped WC could. This should largely be a non-issue but something to be aware of.

### Resolve refs to wrapper
If no action is taken, a ref applied to a React/Preact Component will resolve to the class instance. For wrapped components, the class instance will be the wrapper, which is generally *not* the object the app author is trying to retrieve. This would likely always be accompanied by finding the DOM node of the wrapping component, resolving to the custom element instance.

### Forward refs to custom element
We can choose to forward any refs to the wrapped custom element instance. This is *likely* the object an app-author needs when using refs, so this would generally save a `ReactDOM.findDOMNode` call. This *does* open an inconsistency between wrapper and React implementations though. When the React component is used, the ref will resolve to a React / Preact component. When a wrapped WC is used, it will resolve to a HTMLElement. I don't anticipate this will be a significant issue but worth being aware of.

## Server Side Rendering (SSR)
Sever-side rendering  (SSR) poses a particular challenge to wrapped components because attributes must exist in Reacts virtual-DOM to be output from SSR tools like `ReactDOMServer.renderToString()`. This means that attributes for wrapped WC must be applied during Reacts render and not deferred to `componentDidUpdate` - a technique used by [reactify-wc](https://www.npmjs.com/package/reactify-wc). 

To circumvent, the wrapper could attempt to detect SSR renders using [exenv-es6]([https://www.npmjs.com/package/exenv-es6](https://www.npmjs.com/package/exenv-es6)); diverging behavior to apply attributes during render. There are still cases where there will be un-tracked attributes on the element. For instance, setting `[appearance=accent]` causes `fast-button` to attach a class to the custom element, which is leveraged by CSS. Simply applying the attribute during render in SSR scenarios, React will not capture the class applied to the DOM element. Other common cases are hoisting ARIA attributes up to the WC host after connection. I don't believe this will be an issue client-side because the underlying WC will update itself during the upgrade process but this is something to be aware of.

### [https://www.npmjs.com/package/reactify-wc](https://www.npmjs.com/package/reactify-wc)

`reactify-wc` Looks pretty promising with the following notable exceptions:

#### Does not support class-name diffing:

[This example](https://codesandbox.io/s/reactify-wc-class-name-test-c2ej4?file=/src/App.js) illustrates that the implementation does not correctly reconcile classes applied by React and classes applied by the component itself. See note above on `className`

#### Each update of the React component does quite it a bit of work:

This could make it more difficult to reach performance goals, but is worth doing some perf testing to validate. Each render does the following:

1. remove all event listeners
2. add new event listeners
3. add all properties
4. add all attributes

Additionally, when the prop is a string, number, or boolean - *both* the property and attribute are set. In the case of FAST components this is un-necessary because attributes and properties are kept in sync with `@observable`.

#### Element re-use causes certain code patterns to fail:

There are certain cases where React will re-use element instances. Because `reactify-wc` uses only an additive prop model for properties and attributes, element re-use can cause issues when the props applied to the re-used element are different. [See this example](https://codesandbox.io/s/reactify-wc-element-reuse-with-different-attributes-du4rj) where the button is always an "accent" button even though only one instance applies `[appearance=accent]`. What's happening is one render path adds the attribute, and the next render path re-uses the element but the attribute is not removed.

#### Does not seem to add attributes during SSR
See [this example](https://codesandbox.io/s/reactify-wc-ssr-q3gnp?file=/src/App.js) - note that `use-defaults` and `appearance="accent"` are not in the SRR string.

### [https://github.com/skatejs/val](https://github.com/skatejs/val) 
skatejs/val works similarly to reactify-wc except that it is designed to be UI library agnostic. It accomplishes this at compile time by leveraging the JSX Pragma, which instructs the compiler to use a wrapped `createElement` function constructed by the app author. This pragma must be included in each file implementing a wrapped WC:

```
// my-wrapper.ts
import { createElement } from "react";
import val from "@skatejs/val";

export default val(createElement);
```
```
// my-component.tsx
/** @jsx h */

import h from "my-wrapper";
```

[This example](https://codesandbox.io/s/fast-comoponents-with-skatejs-val-l5mti) shows FAST components working with @skate/val. There is a nice ergonomics here that allows using the Web Component *as* a React / Preact element. Unfortunately, this approach [does not play well with TypeScript](https://codesandbox.io/s/compassionate-brook-emroo?file=/src/App.tsx), so we would need to figure out a way to create, maintain, and associate different typings should we wish to use that approach. It also supports simply using the element tag name instead of the element constructor, which may be easier to ship types for.

### [https://github.com/webcomponents/react-integration](https://github.com/webcomponents/react-integration) (Looks stale)

nicholasrice:
# Description
This change parrots the existing integration guides, adapting it for a React application. It extends to format at the end to discuss some of the hurdles and edge cases that can be encountered using WC and React.

closes #3389

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/docs/integrations/react.md
```diff
@@ -0,0 +1,157 @@
+---
+id: react
+title: React
+sidebar_label: React
+custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/docs/integrations/react.md
+---
+
+FAST can be used in React applications. Let's take a look at how you can set up an Project project, starting from scratch.
```
EisenbergEffect, line 8:
```suggestion
FAST can be used in React applications. Let's take a look at how you can set up a project, starting from scratch.
```

packages/web-components/fast-foundation/docs/integrations/react.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Setting up the React project
+
+First, you'll need to make sure that you have Node.js >= 8.2 and npm >= 5.6 are installed. You can learn more and download that [on the official site](https://nodejs.org/).
```
EisenbergEffect, line 12:
```suggestion
First, you'll need to make sure that you have Node.js >= 8.2 and npm >= 5.6 installed. You can learn more and download that [on the official site](https://nodejs.org/).
```

packages/web-components/fast-foundation/docs/integrations/react.md
```diff
@@ -0,0 +1,157 @@
...
+## Configuring packages
+
+Next, we'll install the FAST packages, along with supporting libraries. To do that, run this command from your new project folder:
+
+```shell
+npm install --save @microsoft/fast-components @microsoft/fast-element lodash-es
+```
+
+## Configure create-react-app
+[create-react-app](https://reactjs.org/docs/create-a-new-react-app.html#create-react-app) ships with a [eslint](https://eslint.org/) rule that makes working with FAST components difficult. There are two changes that will need to be made in the `package.json`:
```
EisenbergEffect, line 29:
```suggestion
[create-react-app](https://reactjs.org/docs/create-a-new-react-app.html#create-react-app) ships with an [eslint](https://eslint.org/) rule that makes working with FAST components difficult. There are two changes that will need to be made in the `package.json`:
```

nicholasrice:
@EisenbergEffect fixed! Thanks

packages/web-components/fast-foundation/docs/integrations/react.md
```diff
@@ -0,0 +1,157 @@
...
+```
+See [configuring eslint](https://create-react-app.dev/docs/setting-up-your-editor#experimental-extending-the-eslint-config) for more information.
+
+## Using the components
+
+With all the basic pieces in place, let's run our app in dev mode with `npm start`. Right now, it displays the React logo and some editing instructions, since we haven't added any code or interesting HTML. Let's change that.
+
+First, open your `src/app.js` file and add the following code:
+
+```ts
```
janechu, line 65: should this be js?

nicholasrice, line 65:
```suggestion
```js
```

nicholasrice: yep

PR #3494, opened 2020-07-10
fix: move setAttribute calls for proxy elements to connectedCallback

### LINKED ISSUE #3456: Runtime error in <fast-button> and <fast-anchor>

There is a runtime error:
```console
Uncaught DOMException: Failed to construct 'CustomElement': The result must not have attributes
```

Example:

1. Add <fast-button> to the page using `createElement` or `createElementNS`

This appears to be because there is a modification to an attribute in the constructor as indicated by the top answer in this post: https://stackoverflow.com/questions/43836886/failed-to-construct-customelement-error-when-javascript-file-is-placed-in-head

We should not therefore be setting any DOM in the constructor and should wait for the `connectedCallback`.

chrisdholt:
# Description
This PR fixes an issue where attributes were defined too early for the custom element, specifically as it relates to form associated elements. This work moves the work to set proxy attributes to the connected callback.

Closes #3456 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components-msft/docs/api-report.md
+++ b/packages/web-components/fast-components-msft/docs/api-report.md
@@ -29,6 +29,8 @@ import { TabPanel } from '@microsoft/fast-foundation';
 import { Tabs } from '@microsoft/fast-foundation';
 import { TextArea } from '@microsoft/fast-foundation';
 import { TextField } from '@microsoft/fast-foundation';
+import { TreeItem } from '@microsoft/fast-foundation';
+import { TreeView } from '@microsoft/fast-foundation';
 
 // Warning: (ae-internal-missing-underscore) The name "AccentButtonStyles" should be prefixed with an underscore because the declaration is marked as @internal
 //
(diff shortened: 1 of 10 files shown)


janechu:
I have created a branch with `createElement` and applied your changes - `users/janechu/custom-element-test`, still getting an error.

janechu:
By setting `appearance` as undefined in the button this appears to fix the problem, I think this indicates these `attr` default settings are happening during the ctor phase, causing an issue. See https://github.com/microsoft/fast/commit/726db0cfb7730b4f4ec34d7616168be6adabfd03

chrisdholt:
> By setting `appearance` as undefined in the button this appears to fix the problem, I think this indicates these `attr` default settings are happening during the ctor phase, causing an issue. See [726db0c](https://github.com/microsoft/fast/commit/726db0cfb7730b4f4ec34d7616168be6adabfd03)

Yes, as I was building your branch I realized that I had started working on that, but it was on your branch. I jumped back to apply some changes and spaced. I think we'll need to set the default appearance differently.

chrisdholt:
@janechu pushed an update. I don't get an error thrown here when testing on your branch. @EisenbergEffect can you take one more look at this when available? One slight change made since you signed off.

Issue #3481, opened 2020-07-09
Menu has erratic keyboard behavior, announced by AT as collapsed, can't check radio items


patrickhlauke:
Trying the menu in https://fast.design/docs/components/menu/, the first time around it seems possible, using `TAB`, to set focus initially to "Menu item 1", then "Menu item 4". Next time around though, only "Menu item 1" takes focus. If tabbing to "Menu item 4", it also then seems to break cursor up/down to move through the menu.

Additionally, at least just checking briefly with Chrome/Firefox and NVDA, menu items are announced as "collapsed", and trying to use `SPACE` to check/select radio menu item doesn't work/just scrolls the page

**What are the steps to reproduce the issue?**

Example:

1. Go to https://fast.design/docs/components/menu/
2. Navigate to the menu using `TAB` (avoiding the keyboard trap that is the live editor)
3. On first `TAB`, focus goes to "Menu item 1"
4. Verify that cursor up/down moves the focus
5. cursor back to "Menu item 1"
6. `TAB` again
7. focus lands on "Menu item 4"
8. try using cursor up/down - note that focus disappears
9. `TAB` out of the menu.
10. reverse `SHIFT`+`TAB` back to the menu
11. note that "Menu item 2" takes focus?

fire up NVDA

12. Set focus to one of the menu items
13. note it's announced as "collapsed"
14. set focus to one of the radio menu items
15. try using `SPACE`
16. note that instead of checking the item, the page scrolls

**What behavior did you expect?**

Consistent tabbing/cursor behavior, menu items not announced as collapsed unless they have a submenu, radio items to be checkable with `SPACE` (and on this last point, the radio menu items are confusing, as they don't seem to visually do anything even for mouse users?)

**In what environment did you see the issue?**

* OS & Device: Windows
* Browser: Chrome, Firefox
* AT: NVDA
* Version: latest stable

chrisdholt:
Thanks @patrickhlauke, I thought we had a task tracking the menuitemradio and menuitemcheckbox designs for the both of these, but it looks like that's missing. I'll add that and link here.

The keyboard behavior detailed in the bug report certainly seems odd and is definitely a bug somewhere. For the NVDA issue, the announcement seems like a bug on the side of NVDA as we appear to be following the roles model and the appropriate structure. We definitely need to get the menuitemcheckbox and menuitemradio behaviors locked and a design implemented and should likely do so as part of the fix for the keyboarding behavior.

chrisdholt:
[...]

PR #3496, opened 2020-07-10
fix: wc menu keyboarding and aria-expanded

scomea:
## Description
A timing issue was causing the menu to check the role attribute of children to determine focusability before those children had a chance to update their roles.  Waiting to check on the value change callback resolved that. Also, we were always setting an "aria-expanded" attribute on menu items which would cause menu items to announce as not expanded even when they had no expanded state to begin with.

## Motivation & context
partially fixes https://github.com/microsoft/fast/issues/3481

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -34,10 +34,10 @@ export class MenuItem extends FASTElement {
      *
      * @public
      * @remarks
-     * HTML Attribute: aria-expanded
+     * HTML Attribute: expanded
      */
-    @attr({ attribute: "aria-expanded", mode: "reflect", converter: booleanConverter })
-    public expanded: boolean = false;
+    @attr({ attribute: "expanded" })
+    public expanded: boolean;
```
chrisdholt, line 40: This needs to be `aria-expanded` or AT will not recognize it as expanded. The attr should stay the same here. I think we you'll want is to keep the boolean attribute converter but change "reflect" to "fromView".

scomea: We write to the "aria-expanded" attribute in the template, just like we do for disabled:
aria-expanded="${x => x.expanded}"

scomea: If it's undefined we don't render a value, which fixes the bug

chrisdholt: gotcha, thanks for the additional context.

PR #3502, opened 2020-07-11
chore: update the configuration and definition files for tree view

janechu:
# Description

This PR solves the following:
- Removes `<fast-tree-item>` from the component explorer
- Corrects a definition file and config file for `<fast-tree-view>`

## Issue type checklist

## Process & policy checklist


sites/fast-component-explorer/app/fast-components/configs/fast-tree-item.ts
```diff
@@ -1,30 +0,0 @@
-import {
```
chrisdholt, line 1: @janechu are we not showing the items for menu or tree item in the explorer currently? I noticed that menu item doesn't show up either. Is that intentional?

janechu: Yes, only whole components are shown not parts of them. This is intended for clarity of use. While menu item is technically a component, its intended use as a part of menu makes it so that menu has to give it a context for it to make sense.

janechu:
@chrisdholt we can also just eliminate the default slot for `<fast-tree-view>`. If it's only children are items and the concern is met to have the same nesting pattern as `<fast-tree-item>`, that still fits the bill. I don't think divider was ever considered as a possibly child but perhaps someone can correct me there. I also don't necessarily think all nesting patterns need to mirror each other, this is different from menu in that menu is a flat list of menu items and creating a new menu from a menu item as a root is possible, but that's not the same structure as a tree view which can only have a single tree view.

EisenbergEffect:
I like this change. It definitely feels more consistent. Also note that we can make `fast-tree-item` automatically project to the item slot, if we make this change. We can do that by simply doing this in its template.

```html
<template slot="item">
...
</template>
```

The templating engine will then always set the item slot on the host when it binds. We couldn't do this before because adding the slot would mess things up when placed inside of the tree-view directly.

So, the end user could have this markup:

```html
<fast-tree-view>
    Foo <!-- this text is treated as content for the default slot -->
    <fast-tree-item> <!--automatically gets an item slot -->
        <fast-tree-item></fast-tree-item> <!--automatically gets an item slot -->
    </fast-tree-item>
</fast-tree-view>
```

packages/web-components/fast-foundation/src/tree-view/tree-view.template.ts
```diff
@@ -14,6 +14,6 @@ export const TreeViewTemplate = html<TreeView>`
         @focus="${(x, c) => x.handleFocus(c.event as FocusEvent)}"
         @blur="${(x, c) => x.handleBlur(c.event as FocusEvent)}"
     >
-        <slot ${slotted("slottedTreeItems")}></slot>
```
chrisdholt, line 17: I think we still need a default slot for things which are not items. I brought up the divider example primarily because I know it's an existing use case for a few folks. I don't think it's a requirement for tree items, but it is one for tree view. Dividers are passed to break up the tree view visually but it's still part of the same keyboarding/interaction model. Example:

image

chrisdholt: This shows a single instance, but there may be multiples and their placement is fairly arbitrary. I'm not sure how we would slot this aside from adding a slot of "item", which doesn't seem semantically correct as it's not a tree item - but that seems to be the only way to get it to slot in order given the API change here. I'm good with the item portion of this, but unsure how it would continue to support this type of behavior. Perhaps @EisenbergEffect has a suggestion on how we could accommodate this outside of using a default slot for these or adding `slot="item"` to this?

janechu: Perhaps I could suggest multiple tree views at that point, so:
```html
<fast-tree-view>
    <fast-tree-item>Foo</fast-tree-item>
</fast-tree-view>
<fast-divider />
<fast-tree-view>
    <fast-tree-item>Bar</fast-tree-item>
</fast-tree-view>
```

chrisdholt: I don't think that would meet the requirements of having it navigate within the same tree construct - I'd need to move from one to the other via tab which isn't the intended behavior and may be confusing to users.

janechu:
@chrisdholt so, I changed this to use a reflected attribute for the slot, but it's not working. It renders in the DOM correctly but doesn't render visibly render items with the slot="item". Not sure what's going on there.

janechu:
@chrisdholt @EisenbergEffect well, using `setAttribute` for the slot didn't work either. Not sure what's going on there.

janechu:
I've backtracked on this to just be a fix for the component explorer and am leaving the API for tree-view and tree-item alone since I've run into a few roadblocks. See https://github.com/microsoft/fast/issues/3511.

Issue #3504, opened 2020-07-13
Add aria-valueunit attribute to the slider component


khamudom:
**What feature are you requesting?**
Allow the developer to add a readout, for screen readers, of the unit, set on the slider label. 

**What would be the benefit of adding this feature?**
Having the screen reader tell users the unit of the value, for example,
instead of just hearing the number, "90", the screen reader can say, 90 degrees celsius"
![image](https://user-images.githubusercontent.com/37851220/87342894-dc9a6800-c500-11ea-9717-b1385e4a0358.png)
other units could be "pixels" or "inches" etc.

**What solution would be ideal for you?**
In the slider component, create an attribute called, `aria-valueunit`
```
@attr({ attribute: "aria-valueunit" })
    public ariaValueUnit: string;
```

In the template, we would use `aria-valuetext` so the screen reader will read out a string.
https://www.w3.org/TR/wai-aria-1.1/#aria-valuetext

Adding `ariaValueUnit`
```
aria-valuetext="${x => x.value} ${x => x.ariaValueUnit ? x.ariaValueUnit : ""}"
```

When implementing,
```
<fast-slider 
        id="switcher" 
        orientation="vertical" 
        min="0" 
        max="100" 
        step="10" 
        value="70" 
        aria-valueunit="&#8451;"
>
    <fast-slider-label position="0">
        0&#8451;
    </fast-slider-label>
    <fast-slider-label position="10">
        10&#8451;
    </fast-slider-label>
    <fast-slider-label position="90">
         90&#8451;
     </fast-slider-label>
     <fast-slider-label position="100">
         100&#8451;
    </fast-slider-label>
</fast-slider>
```

**What alternatives have you considered?**
An alternative to doing the ternary option in the template is maybe creating a `nullableStringConverter` for empty strings.
This to help when `aria-valueunit` is not defined.

**Is there any additional context for your request?**

**Attach screenshots or images add detail to your request:**

marjonlynch:
This issue was resolved with the introduction of the attribute valueTextFormatter

PR #3508, opened 2020-07-14
feat: create an observable function to support aria-valuetext on slider component

khamudom:
# Description

Added `valueTextFormatter` to use on `aria-valuetext`.

To smoke branch. 
Run fast-component(or fast-component-msft)
Enable screen reader
If there is a value unit on the slider example, the screen reader will read it out.

## Motivation & context

feature-request #3504 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/slider/slider.template.ts
+++ b/packages/web-components/fast-foundation/src/slider/slider.template.ts
@@ -13,7 +13,7 @@ export const SliderTemplate = html<Slider>`
         class="${x => (x.readOnly ? "readonly" : "")} 
         ${x => x.orientation || Orientation.horizontal}"
         tabindex="${x => (x.disabled ? null : 0)}"
-        aria-valuetext="${x => x.valueText}"
+        aria-valuetext="${x => x.valueTextFormatter(x.value)}"
         aria-valuenow="${x => x.value}"
         aria-valuemin="${x => x.min}"
         aria-valuemax="${x => x.max}"
--- a/packages/web-components/fast-foundation/src/slider/slider.template.ts
+++ b/packages/web-components/fast-foundation/src/slider/slider.template.ts
@@ -13,6 +13,7 @@ export const SliderTemplate = html<Slider>`
         class="${x => (x.readOnly ? "readonly" : "")} 
         ${x => x.orientation || Orientation.horizontal}"
         tabindex="${x => (x.disabled ? null : 0)}"
+        aria-valuetext="${x => x.value} ${x => (x.ariaValueUnit ? x.ariaValueUnit : "")}"
         aria-valuenow="${x => x.value}"
         aria-valuemin="${x => x.min}"
         aria-valuemax="${x => x.max}"
--- a/packages/web-components/fast-foundation/src/slider/slider.template.ts
+++ b/packages/web-components/fast-foundation/src/slider/slider.template.ts
@@ -13,7 +13,7 @@ export const SliderTemplate = html<Slider>`
         class="${x => (x.readOnly ? "readonly" : "")} 
         ${x => x.orientation || Orientation.horizontal}"
         tabindex="${x => (x.disabled ? null : 0)}"
-        aria-valuetext="${x => x.value} ${x => (x.ariaValueUnit ? x.ariaValueUnit : "")}"
+        aria-valuetext="${x => x.valueText}"
         aria-valuenow="${x => x.value}"
         aria-valuemin="${x => x.min}"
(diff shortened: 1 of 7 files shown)


nicholasrice:
@khamudom I don't see this attribute in https://www.w3.org/TR/wai-aria-1.1/#slider - is there somewhere I can read about this attribute?

khamudom:
> @khamudom I don't see this attribute in https://www.w3.org/TR/wai-aria-1.1/#slider - is there somewhere I can read about this attribute?

It is a custom aria attribute that I created for the component, using it on `aria-valuetext`. Maybe giving it a different name could help it stand out as it being a custom attribute?

nicholasrice:
I see. Do we actually need to add code for this? I think with your implementation the same product could be accomplished by the author adding the `aria-valuetext` attribute themselves.

scomea:
I don't think we should label custom attributes with the "aria-" prefix.  That should be reserved for actual aria attributes IMO.

scomea:
Our react slider has a "valuetextStringFormatter" callback function that allows developers to format value text as the slider value changes.  I don't think just a string will work as the final aria-valuetext string may have variable formatting based on the value, the locale, etc...

janechu:
To add to this, only `dataset` attributes are allowed to be customized, such as `data-foo`.

scomea:
I think the notion of passing a "unit" is a red herring.  One of the W3C examples is converting a 1-7 slider to "Days of the week".  So when the user selects 1 valuetext is "Monday", 2 is "Tuesday", etc...  So what we really need is the custom function so developers can format as they need.  There may be cases where devs want to append a unit, but we'd be better off not worrying about that and giving them a function to do whatever.

EisenbergEffect:
@scomea I'd agree with that. All we would need is an observable for the function with a default function that just returns the value itself.

scomea:
@EisenbergEffect Yes.  I'm not sure we should even populate valuetext if it's the same as valuenow tho.

EisenbergEffect:
We could have the default function return null, which would remove the attribute.

khamudom:
Thank you to those that have added feedback. PR is updated to remove the custom attr and add an observable function.

packages/web-components/fast-foundation/src/slider/slider.template.ts
```diff
@@ -13,6 +13,7 @@ export const SliderTemplate = html<Slider>`
         class="${x => (x.readOnly ? "readonly" : "")} 
         ${x => x.orientation || Orientation.horizontal}"
         tabindex="${x => (x.disabled ? null : 0)}"
+        aria-valuetext="${x => x.valueText}"
```
EisenbergEffect, line 16: Was there a reason that this couldn't be implemented without the code below like this?

```js
aria-valuetext="${x => x.valueTextFormatter(x.value)}"
```

With this code, the `area-valuetext` attr would update any time the formatter or the value changed. You would just need to set the formatter to a default function that returns null, like so:

```ts
@observable valueTextFormatter: (value: string) => string | null = () => null;
```

khamudom: Updated to go with Rob's suggestions. Thanks!

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -113,6 +113,11 @@ export class Slider extends FormAssociated<HTMLInputElement>
     @observable
     public trackMinHeight: number = 0;
 
+    /**
+     * @internal
```
nicholasrice, line 117: I think we want this `@public` because the expectation is that can override this right? Also please add doc comments to describe the API and what purpose the property serves.

khamudom: thanks! updated to public and added doc comments

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -113,6 +113,13 @@ export class Slider extends FormAssociated<HTMLInputElement>
     @observable
     public trackMinHeight: number = 0;
 
+    /**
+     * Custom function that generates a string for the component's "aria-valuetext" attribute based on the currrent value.
+     *
+     * @public
+     */
+    @observable valueTextFormatter: (value: string) => string | null = () => null;
```
chrisdholt, line 121: nit, can you denote this as `public` in TS as well?

```suggestion
@observable
public valueTextFormatter: (value: string) => string | null = () => null;
```

khamudom: updated

PR #3510, opened 2020-07-14
fix: switch to webpack for homepage building

radium-v:
# Description
Switch from yarn process spawning to webpack API to build the `fast-website` package directly in the `website` process.

## Motivation & context
The spawn process method doesn't work on Windows without changing some flags, and this method is safer.

## Issue type checklist

The only caveat is that we lose the progress flag, but there may be ways to get stats back if we really miss it.

## Process & policy checklist

I've tested it on Ubuntu (WSL2) and Windows (node and yarn directly installed) and both build and run successfully.

### DIFF
--- a/sites/website/src/build-homepage.js
+++ b/sites/website/src/build-homepage.js
@@ -3,52 +3,43 @@
 /* eslint-env node */
 /* eslint-disable @typescript-eslint/typedef */
 /* eslint-disable @typescript-eslint/no-var-requires */
-/* eslint-disable @typescript-eslint/explicit-function-return-type */
-const { spawn } = require("child_process");
 const path = require("path");
 const copyfiles = require("copyfiles");
+const webpack = require("webpack");
+const fastWebsiteConfig = require("fast-website/webpack.prod.js");
 
-const projectPath = path.resolve(__dirname, "../../fast-website");
 const staticPath = path.resolve(__dirname, "../static");
-const indexPath = path.resolve(projectPath, "dist/index.html");
 const utilitiesPath = path.resolve(__dirname, "../../site-utilities");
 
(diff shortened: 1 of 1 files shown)


sites/website/src/build-homepage.js
```diff
@@ -3,52 +3,30 @@
...
-yarnBuild.on("error", error => console.log(`error: ${error.message}`));
-
-yarnBuild.on("close", exitCode => {
-    console.log(`child process exited with code ${exitCode}`);
-    if (exitCode !== 0) {
-        process.exit(exitCode);
+console.log("Building fast-website with webpack...");
+webpack(fastWebsiteConfig, (err, stats) => {
+    if (err || stats.hasErrors()) {
+        console.log(err);
```
nicholasrice, line 20: I'd kinda expect this to throw if there are errors or exit non-zero - is there a reason not to do that?

radium-v: I updated the error checking so it will exit on stack errors and stat errors.

nicholasrice: I think you want to exit `1`, sorry for the confusion: https://nodejs.org/api/process.html#process_process_exit_code

sites/website/src/build-homepage.js
```diff
@@ -3,52 +3,43 @@
...
-    if (exitCode !== 0) {
-        process.exit(exitCode);
+    if (stats.hasWarnings()) {
+        info.warnings.forEach(w => console.warn(w));
     }
 
-    copyBundle();
+    console.info("Built fast-website to static");
+
+    copyfiles(
```
EisenbergEffect, line 40: Is this call blocking? Does it matter? Just curious.

PR #3512, opened 2020-07-14
feat: basic menu item check state toggling

scomea:
# Description
Adding a basic checked state toggling to radio and checkbox type menu items.  Currently both have the same behavior but radio type items are expected to have radio group capabilities added at a later date.

## Motivation & context
https://github.com/microsoft/fast/issues/3481

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/menu-item/menu-item.ts
+++ b/packages/web-components/fast-foundation/src/menu-item/menu-item.ts
@@ -87,8 +87,8 @@ export class MenuItem extends FASTElement {
         }
 
         switch (this.role) {
-            case "menuitemcheckbox":
-            case "menuitemradio":
+            case MenuItemRole.menuitemcheckbox:
+            case MenuItemRole.menuitemradio:
                 this.checked = !this.checked;
                 break;
         }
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -72,10 +78,21 @@ export class MenuItem extends FASTElement {
...
     };
 
-    private change = (): void => {
+    private invoke = (): void => {
+        if (this.disabled) {
+            return;
+        }
+
+        switch (this.role) {
+            case "menuitemcheckbox":
```
janechu, line 90: Could you use the enum for these cases?

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -63,7 +64,12 @@ export class MenuItem extends FASTElement {
...
     public handleMenuItemKeyDown = (e: KeyboardEvent): boolean => {
-        this.change();
+        switch (e.keyCode) {
+            case keyCodeEnter:
+            case keyCodeSpace:
+                this.invoke();
+                return false;
+        }
 
         return true;
```
janechu, line 74: unrelated to your change, but I am wondering why this is returning a boolean.

scomea: I guess I could return void instead of false because both result in the event being default prevented, but in order to not preventDefault on the event we need to return true.   I found this surprising.

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
scomea: Yes, done.

PR #3513, opened 2020-07-15
fix: export all style utilities

bheston:
# Description

Update export statement to export all style utility functions in fast-components.

## Motivation & context

Noticed that elevation wasn't exported, and found that only recipes were being exported. Looks unintentional.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/index.ts
+++ b/packages/web-components/fast-components/src/index.ts
@@ -17,7 +17,7 @@ export * from "./radio/index";
 export * from "./radio-group/index";
 export * from "./slider/index";
 export * from "./slider-label/index";
-export * from "./styles/recipes";
+export * from "./styles/index";
 export * from "./switch/index";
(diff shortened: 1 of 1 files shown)


bheston:
@EisenbergEffect I considered that. There was a brief chat about it and we weren't sure, but I'm good with skipping this if we think we'll change this with the modular work, at least initially for elevation. Ping @nicholasrice to make a call. :)

chrisdholt:
> @EisenbergEffect I considered that. There was a brief chat about it and we weren't sure, but I'm good with skipping this if we think we'll change this with the modular work, at least initially for elevation. Ping @nicholasrice to make a call. :)

@bheston I would prefer to land the modular elevation story just to ensure we aren't going to be creating a breaking change here before we decide to export this. If someone wants to use this, they can still drill into the package.

PR #3514, opened 2020-07-15
fix: mimic the events on fast-text-field to match the standard HTML input

### LINKED ISSUE #3453: Text field change event not fired
`change` event is not emitted on an `fast-text-field` element when the value is cleared or the last character is deleted.

call `addEventListener('change', ()=>console.log('change'))` on an element and observe console

'change' event is emitted when value is cleared

* Browser Chrome
* Version 83.0.4103.116

This is important for integration with Aurelia e.g., where binding engine listens for events to update bindings

khamudom:
# Description

Updated `handleTextInput` to emit `input` and added more events for `@keydown`, `@blur` and `@focus`. This is to mimic the behavior on the platform's `<input />` element.
The change will now happen on key enter and blur when a value has changed.

## Motivation & context

Fixes #3453 

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/text-field/text-field.ts
```diff
@@ -251,7 +252,27 @@ export class TextField extends FormAssociated<HTMLInputElement> {
         if (this.control && this.control.value) {
```
EisenbergEffect, line 252:
```suggestion
        if (this.control) {
```

The reason for this is that `value` could be the empty string after backspacing, which is a legitimate change. So, that needs to be reported.

packages/web-components/fast-foundation/src/text-field/text-field.ts
```diff
@@ -218,15 +219,15 @@ export class TextField extends FormAssociated<HTMLInputElement> {
         if (this.proxy instanceof HTMLElement) {
             this.proxy.value = this.value;
         }
```
EisenbergEffect, line 221:
```suggestion
        }
       if (this.control) {
            this.control.value = this.value;
        }
```

Because we've got both a proxy input and a control input, we need to update both when the value changes. Otherwise, things can get out of sync. I experience this bug while preparing my demos for day of learning and had to hack this fix locally to get things working correctly.

nicholasrice:
> Also, I wanted to learn from @nicholasrice why we have both a proxy input and a control input. I know we talked about the form association fallback a while ago but I want to make sure I understand correctly in this context. Is the idea that the control is always present but the proxy is needed when form association is not supported? If so, it seems like the proxy is always being created in the constructor, regardless of browser support for form association.

Your understanding is correct. The element is *created* in the constructor and handed off to `FormAssociated`. `FormAssociated` only appends the element to the host when form-association is not supported.

The proxy functions by being in the same document context as the parent `<form>`. The `[part=control]` in this case is in the `TextField`'s shadow-dom, so it won't be associated to the parent `<form>` (does not exist in `form.elements` or in `formdata` events). If we *did* move the `[part=control]` to the light-dom then for form associated browsers there would be *two* form elements for every text-field which is also un-desirable.

nicholasrice:
Did a bit more digging on this, and I think there is a more simple solution:

1. The `input` event is actually already emitted from the `[part=control]` and that can be caught by an event listener on the host. Nothing to do here.
2. The *change* event from the `[part=control]` can't be listened for from the host. *I'm actually not sure why this is the case* - we might want to do some more investigation here. However, we *can* successfully listen for that event *from* the host, meaning we can simply re-emit the event whenever the `[part=control]` emits the event.

packages/web-components/fast-foundation/src/text-field/text-field.ts
khamudom: added this suggestion

packages/web-components/fast-foundation/src/text-field/text-field.ts
khamudom: added this suggestion

khamudom:
> Did a bit more digging on this, and I think there is a more simple solution:
> 
> 1. The `input` event is actually already emitted from the `[part=control]` and that can be caught by an event listener on the host. Nothing to do here.
> 2. The _change_ event from the `[part=control]` can't be listened for from the host. _I'm actually not sure why this is the case_ - we might want to do some more investigation here. However, we _can_ successfully listen for that event _from_ the host, meaning we can simply re-emit the event whenever the `[part=control]` emits the event.

reverted to remove the focus, keydown, and blur events. Still unsure on the bubbling issue.

packages/web-components/fast-foundation/src/text-field/text-field.ts
```diff
@@ -222,9 +224,10 @@ export class TextField extends FormAssociated<HTMLInputElement> {
      * @internal
      */
     public handleTextInput(): void {
-        if (this.control && this.control.value) {
+        if (this.control) {
             this.value = this.control.value;
         }
+        this.$emit("input", this.value);
```
nicholasrice, line 230: You don't need to emit an input event - it will bubble from the shadowed input.

nicholasrice:
I figured out why `input` events bubble out of the control but `change` events do not. `input` event's have the `composed` flag and bubble through shadow DOM boundaries but `change` event's do not.

@khamudom I'm doing a round of bug fixes surrounding form components. If you don't mind, I'll merge this work in with that work and get *all* the components working as expected.

nicholasrice:
Closing because this work will be is being done in #3581

PR #3527, opened 2020-07-17
feat: anchored region revision

scomea:
# Description
This change is a major revision to the anchored region component which does the following:
- now only a single div
- more control over region scaling (can size based on content, anchor, or fill available space)
- tracks layout parent resizing (ie. the offet parent or document root)
- supports fixed placement (useful for breaking out of parent divs)
- we no longer track viewport scrolling, but developpers can update the offset themselves
- falls back to using getBoundingClientRect in scenarios where intersection observer does not work 
- enables authors to always/never use gbcr if the default behavior isn't ideal

## Motivation & context
- addressing feedback on the previous iteration

## Issue type checklist

## Process & policy checklist


scomea:
And yes, I'd love to come up with a better attribute name than "use-gbcr"...

packages/web-components/fast-components/src/anchored-region/resize-observer.ts
```diff
@@ -1,17 +1,17 @@
-import { ResizeObserverEntry } from "./resize-observer-entry";
+// import { ResizeObserverEntry } from "./resize-observer-entry";
 
-export type ConstructibleResizeObserver = new (
-    callback: ResizeObserverCallback
-) => ResizeObserverClassDefinition;
+// export type ConstructibleResizeObserver = new (
```
nicholasrice, line 3: Can this file be removed?

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -56,124 +65,250 @@ enum Location {
...
+    private useGbcrChanged(): void {
+        if ((this as any).$fastController.isConnected && this.initialLayoutComplete) {
+            this.initialize();
+        }
+    }
 
+    /**
+     * The HTML element being used as the anchor
+     *
+     * @ public
```
nicholasrice, line 274:
```suggestion
     * @public
```

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -56,124 +65,250 @@ enum Location {
...
-            this.initialLayoutComplete = false;
-            this.reset();
-        }
+        this.reset();
     }
 
+    /**
+     * The HTML element being used as the viewport
+     *
+     * @ public
```
nicholasrice, line 285:
```suggestion
     * @public
```

packages/web-components/fast-components/src/anchored-region/resize-observer.ts
scomea: yes, there was a redundant copy of those, gone.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -376,41 +604,120 @@ export class AnchoredRegion extends FASTElement {
...
+            return null;
         }
-        this.requestLayoutUpdate();
+
+        this.regionDimension = {
+            height: regionRect.height,
+            width: regionRect.width,
+        };
+
+        // todo: scomea, try to avoid getting viewportRect when it is document root
```
nicholasrice, line 686: Is this still actionable?

scomea: Yes, I think there is potential perf optimization there, but deferring on it until later.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -734,22 +1024,26 @@ export class AnchoredRegion extends FASTElement {
         desiredHorizontalPosition: AnchoredRegionHorizontalPositionLabel,
         nextPositionerDimension: Dimension
     ): void => {
+        let layoutParentWidth = 0;
+        if (this.offsetParent !== null) {
+            layoutParentWidth = this.offsetParent.clientWidth;
+        } else {
+            layoutParentWidth = document.body.clientWidth;
+        }
```
janechu, line 1032:
```suggestion
        const layoutParentWidth = this.offsetParent !== null
            ? this.offsetParent.clientWidth
            : document.body.clientWidth;
```

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -779,142 +1084,116 @@ export class AnchoredRegion extends FASTElement {
         desiredVerticalPosition: AnchoredRegionVerticalPositionLabel,
         nextPositionerDimension: Dimension
     ): void => {
+        let layoutParentHeight = 0;
+        if (this.offsetParent !== null) {
+            layoutParentHeight = this.offsetParent.clientHeight;
+        } else {
+            layoutParentHeight = document.body.clientHeight;
+        }
```
janechu, line 1092:
```suggestion
        const layoutParentHeight = this.offsetParent !== null
            ? this.offsetParent.clientHeight
            : document.body.clientHeight;
        let layoutParentHeight = 0;
```

packages/web-components/fast-foundation/src/anchored-region/anchored-region.spec.md
```diff
@@ -191,6 +155,11 @@ A region that renders below the anchor until that space is less than 100px.
 </div>
 ```
 
+The anchored region can be configured to scale to either the size of the content in it's slot, to the size of the element it is anchored to, or the available space between the anchor and the edge of the viewport element by setting the vertical/horizontal scaling attribute:
+- "anchor" - size to match anchor
+- "fill" - size to match space between the anchor and the viewport edge
+- "content" - the default, matches the size of the content in the region's slot.
```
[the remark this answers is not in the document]
scomea: possesive "region's" is correct

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -56,124 +65,250 @@ enum Location {
...
+     * @public
+     * @remarks
+     * HTML Attribute: use-gbcr
      */
-    @observable
-    public initialLayoutComplete: boolean = false;
+    @attr({ attribute: "use-gbcr" })
+    public useGbcr: GbcrUsage = "default";
+    private useGbcrChanged(): void {
+        if ((this as any).$fastController.isConnected && this.initialLayoutComplete) {
```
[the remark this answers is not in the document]
scomea: Yes, that works.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.spec.md
```diff
@@ -233,34 +200,41 @@ NOTE: this component api will not be exposed outside of the fast-components pack
...
 ```
 
 ## Implementation
 
 ### States
 Layout update checks in the component happen when:
-- the viewport scrolls
 - intersection observer reports a collision with the viewport
-- resize observer reports a resize event on the anchor, the viewport or the component itself.
+- resize observer reports a resize event on the anchor, the viewport, the component's offsetParent or the component itself.
```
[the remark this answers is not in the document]
scomea: added the ticks, but "component's" is correct for possessive

packages/web-components/fast-foundation/src/anchored-region/anchored-region.spec.md
```diff
@@ -4,59 +4,23 @@
...
-
-A primary goal of the component was to enable authors to create responsive layouts without resorting to expensive [getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) calls. It instead depends on the more performant [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) and [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) interfaces.
-
-For a more in-depth understanding of how this component works under the covers please refer to the [intersection observer api](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API). 
-
 ### Use Cases
 
-It is envisioned that this component would be used as a building block for other components in this library (select, flyout, tooltip, etc) as well as being available for standalone use in responsive layouts.
+It is envisioned that this component would be used as a building block for other components in this library who need to position elements relative to another HTMLElement(select, flyout, tooltip, etc) as well as being available for standalone use in responsive layouts.
```
awentzel, line 9: Would "that" be more grammatically correct than "who" need to position elements....

PR #3528, opened 2020-07-17
fix: address nested styling issues and provide a more intuitive API for nesting

### LINKED ISSUE #3516: Tree view indentation is visually broken
The tree view is not visually aligned as expected in both web component packages. When a tree view has a root node with child nodes, the expand/collapse button is included and shifts the text. When there are other tree items on the same level, the text should align to the text of the tree item with the expand/collapse button. Currently there is no such functionality to apply this spacing. The preferred visual styling can be seen by inspecting the React package (so long as it's still in the repo :) ). This specific problem is *only* an issue for root tree items and not nested tree items.

**Expected:**
image

**Actual:**
image

While investigating this, I also noticed that it appears that the indenting of the tree items is also off when you have tree items nested within a tree item. 

image

marjonlynch:
## Description
Tree items at the root level were not styled correctly if they did not have children, indentation was incorrect. Also this fix allow developers to add arbitrary elements into the tree view structure.

closes #3516

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components-msft/src/tree-item/tree-item.styles.ts
+++ b/packages/web-components/fast-components-msft/src/tree-item/tree-item.styles.ts
@@ -86,7 +86,7 @@ export const TreeItemStyles = css`
         ${
             /* Font size should be based off calc(1em + (design-unit + glyph-size-number) * 1px) - 
             update when density story is figured out */ ""
-        } font-size: calc(1em + var(--type-ramp-base-font-size));
+        } font-size: calc(1em + (var(--design-unit) + 16) * 1px);
     }
 
     .expand-collapse-button {
@@ -181,11 +181,6 @@ export const TreeItemStyles = css`
         position: absolute;
         ${/* value needs to be localized  */ ""}
         left: var(--expand-collapse-button-nested-width, calc(${heightNumber} * -1px));
-    }
-
-    :host(.leaf) .content-region {
-        position: relative;
-        margin-inline-start: calc((${heightNumber} + (var(--design-unit) * 2) + 2) * 1px);
     }
 
     ::slotted(fast-tree-item) {
--- a/packages/web-components/fast-components-msft/src/tree-item/tree-item.styles.ts
+++ b/packages/web-components/fast-components-msft/src/tree-item/tree-item.styles.ts
@@ -183,6 +183,11 @@ export const TreeItemStyles = css`
         left: var(--expand-collapse-button-nested-width, calc(${heightNumber} * -1px));
     }
 
+    :host(.leaf) .content-region {
+        position: relative;
+        margin-inline-start: calc((${heightNumber} + (var(--design-unit) * 2) + 2) * 1px);
+    }
+
     ::slotted(fast-tree-item) {
         --tree-item-nested-width: 1em;
         --expand-collapse-button-nested-width: calc(${heightNumber} * -1px);
--- a/packages/web-components/fast-components-msft/src/tree-view/fixtures/tree-view.html
+++ b/packages/web-components/fast-components-msft/src/tree-view/fixtures/tree-view.html
@@ -3,25 +3,30 @@
     <fast-tree-view render-collapsed-nodes="false">
         <fast-tree-item>
             Root item 1
-            <fast-tree-item slot="item">
+            <fast-divider slot="item"></fast-divider>
+            <fast-tree-item>
                 Flowers
-                <fast-tree-item slot="item" disabled>Daisy</fast-tree-item>
-                <fast-tree-item slot="item">Sunflower</fast-tree-item>
-                <fast-tree-item slot="item">Rose</fast-tree-item>
+                <fast-tree-item disabled>Daisy</fast-tree-item>
+                <fast-tree-item>Sunflower</fast-tree-item>
+                <fast-tree-item>Rose</fast-tree-item>
             </fast-tree-item>
-            <fast-tree-item slot="item">Nested item 2</fast-tree-item>
-            <fast-tree-item slot="item">Nested item 3</fast-tree-item>
+            <fast-tree-item>Nested item 2</fast-tree-item>
+            <fast-tree-item>Nested item 3</fast-tree-item>
         </fast-tree-item>
         <fast-tree-item>
             Root item 2
-            <fast-tree-item slot="item">
+            <fast-divider slot="item"></fast-divider>
+            <fast-tree-item>
                 Flowers
-                <fast-tree-item slot="item" disabled>Daisy</fast-tree-item>
-                <fast-tree-item slot="item">Sunflower</fast-tree-item>
-                <fast-tree-item slot="item">Rose</fast-tree-item>
+                <fast-tree-item disabled>Daisy</fast-tree-item>
+                <fast-tree-item>Sunflower</fast-tree-item>
+                <fast-tree-item>Rose</fast-tree-item>
             </fast-tree-item>
-            <fast-tree-item slot="item">Nested item 2</fast-tree-item>
-            <fast-tree-item slot="item">Nested item 3</fast-tree-item>
+            <fast-tree-item>Nested item 2</fast-tree-item>
+            <fast-tree-item>Nested item 3</fast-tree-item>
+        </fast-tree-item>
+        <fast-tree-item>
+            Root item 3 - Leaf Erikson
         </fast-tree-item>
     </fast-tree-view>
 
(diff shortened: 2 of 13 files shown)


packages/web-components/fast-components-msft/src/tree-item/tree-item.styles.ts
```diff
@@ -183,6 +183,11 @@ export const TreeItemStyles = css`
         left: var(--expand-collapse-button-nested-width, calc(${heightNumber} * -1px));
     }
 
+    :host(.leaf) .content-region {
```
nicholasrice, line 186: FYI in general we should avoid setting classes on the `:host` - the reason for this is interop with UI frameworks where classNames may be data-bound - see https://github.com/microsoft/fast/issues/3389 for an example.

Not a huge issue here because this component *already* adds classes, but good to be aware of going forward

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -198,6 +198,15 @@ export class TreeItem extends FASTElement {
         }
     };
 
+    public childItemLength(): number {
```
nicholasrice, line 201: Are you using this somewhere?

packages/web-components/fast-foundation/src/tree-view/tree-view.template.ts
```diff
@@ -14,6 +14,7 @@ export const TreeViewTemplate = html<TreeView>`
         @focus="${(x, c) => x.handleFocus(c.event as FocusEvent)}"
         @blur="${(x, c) => x.handleBlur(c.event as FocusEvent)}"
     >
-        <slot ${slotted("slottedTreeItems")}></slot>
+        <slot></slot>
+        <slot name="item" ${slotted("slottedTreeItems")}></slot>
```
janechu, line 18: It seems like we would have all default items appearing before the slotted named items, what scenario are we trying to allow by doing this?

packages/web-components/fast-foundation/src/tree-view/README.md
```diff
@@ -6,15 +6,22 @@ custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-comp
...
 
 ```html live
     <fast-tree-view>
         Root
-        <fast-tree-item slot="item">Item 1</fast-tree-item>
-        <fast-tree-item slot="item">Item 2</fast-tree-item>
+        <fast-tree-item>
+            Item 1
+            <fast-divider slot="item"></fast-divider>
```
janechu, line 19: I thought things like `<fast-divider>` would only occur as children of the `<fast-tree-view>` and not `<fast-tree-item>`?

packages/web-components/fast-components-msft/src/tree-view/fixtures/tree-view.html
```diff
@@ -2,33 +2,49 @@
     <h4>With items (don't render collapsed nodes)</h4>
     <fast-tree-view render-collapsed-nodes="false">
         <fast-tree-item>
-            Root item
-            <fast-tree-item slot="item">
+            Root item 1
+            <fast-divider slot="item"></fast-divider>
```
janechu, line 6: I thought we were going to try to avoid declaring things that are not items with the slot "item"?

chrisdholt:
@EisenbergEffect @nicholasrice @janechu can you take a look? I've pushed some changes to fix the nesting and allow for nesting without setting a slot of item.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
chrisdholt: we are now!

packages/web-components/fast-foundation/src/tree-view/README.md
```diff
@@ -6,15 +6,22 @@ custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-comp
 ---
 
 An implementation of a [tree](https://w3c.github.io/aria/#tree) as a web-component.
+The `fast-tree-view` supports other elements being included in the tree structure with the introduction of `slot="item"`.
```
chrisdholt, line 9: Note to self - remove

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -23,16 +23,36 @@ export class TreeView extends FASTElement {
...
         if (this.$fastController.isConnected) {
             // filter the tree items until that's done for us in the framework
             this.resetItems();
             this.treeItems = this.getVisibleNodes();
             this.setItems();
+
+            // check if any tree items have nested items
+            // if they do, apply the nested attribute
+            if (this.checkForNestedItems()) {
+                console.log("yes nested items!");
```
janechu, line 40: still needed?

chrisdholt, line 40: DOH

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -23,16 +23,35 @@ export class TreeView extends FASTElement {
...
+                        node.setAttribute("nested", "true");
+                    }
+                });
+            }
         }
     }
 
+    private checkForNestedItems(): boolean {
+        return this.slottedTreeItems.some((node: HTMLElement) => {
+            return isTreeItemElement(node) && node.querySelector("[role='treeitem'");
```
nicholasrice, line 51:
```suggestion
            return isTreeItemElement(node) && node.querySelector("[role='treeitem']");
```

packages/web-components/fast-components-msft/src/tree-item/tree-item.styles.ts
```diff
@@ -86,7 +86,7 @@ export const TreeItemStyles = css`
         ${
             /* Font size should be based off calc(1em + (design-unit + glyph-size-number) * 1px) - 
             update when density story is figured out */ ""
-        } font-size: calc(1em + var(--type-ramp-base-font-size));
+        } font-size: calc(1em + (var(--design-unit) + 16) * 1px);
```
nicholasrice, line 89: Why isn't this on the typeramp?

chrisdholt: It's not related to the typeramp, it's related to a var we don't have included yet. Imaging the type size changing on the page and the nesting changing. That seems broken to me. It appears as though we were using the type ramp value to get the number 16. Rather, the original intent of 16 was related to horizontal spacing.

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
chrisdholt: gah - committed

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -198,6 +199,21 @@ export class TreeItem extends FASTElement {
...
+        const treeChildrens: HTMLElement[] = this.childItems.filter(
+            (item: HTMLElement) => {
+                return isTreeItemElement(item);
+            }
+        );
+        return treeChildrens ? treeChildrens.length : 0;
```
nicholasrice, line 208: lol I like this variable name.

chrisdholt: not mine, but I'm keeping it lol

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -75,15 +75,15 @@ export class TreeItem extends FASTElement {
...
-                if (node instanceof TreeItem) {
+                if (isTreeItemElement(node)) {
                     // TODO: maybe not require it to be a TreeItem?
                     (node as TreeItem).nested = true;
                 }
             });
         }
     }
 
-    @observable
```
nicholasrice, line 86: `nested` should have tsdoc comments

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -75,15 +75,15 @@ export class TreeItem extends FASTElement {
...
+    @attr
```
nicholasrice, line 86: also should this be a boolean mode?

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -23,16 +23,35 @@ export class TreeView extends FASTElement {
...
             this.resetItems();
             this.treeItems = this.getVisibleNodes();
             this.setItems();
+
+            // check if any tree items have nested items
+            // if they do, apply the nested attribute
+            if (this.checkForNestedItems()) {
+                this.slottedTreeItems.forEach(node => {
+                    if (isTreeItemElement(node)) {
+                        node.setAttribute("nested", "true");
```
nicholasrice, line 42: assuming `nested` is a boolean attribute (see above comment) should this just set empty string?

chrisdholt: yeah, likely

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -75,15 +75,21 @@ export class TreeItem extends FASTElement {
...
 
-    @observable
+    /**
+     * Tracks the nested state of the item
+     * @public
+     * @remarks
+     * HTML Attribute: nested
+     */
+    @attr({ mode: "boolean" })
     public nested: boolean;
```
EisenbergEffect, line 93: If this isn't intended to be set in HTML by a consumer, then probably better to make this into an `@observable`.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -198,6 +205,21 @@ export class TreeItem extends FASTElement {
...
+    public childItemLength(): number {
+        const treeChildren: HTMLElement[] = this.childItems.filter(
+            (item: HTMLElement) => {
+                return isTreeItemElement(item);
+            }
+        );
+        return treeChildren ? treeChildren.length : 0;
+    }
+
+    public isNestedItem(): null | boolean {
```
EisenbergEffect, line 217: Might make sense for this to be a readonly property that always returns a boolean.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
chrisdholt: I like these improvements - I’ll take a follow up here. Wanted to land a fix ASAP, sorry we missed you the first round!

Issue #3246, opened 2020-06-08
How are LTR and RTL specific CSS rules applied?


nicholasrice:
We currently don't have guidance or reference implementations for handling RTL style rules that cannot be achieved directly through platform features (`grid`, `flex`, `margin-inline-start`, etc). The solution should support the following:

1. property switching ("left" -> "right")
2. value inversion (20px -> -20px)
3. value replacement (0deg -> -90deg)

### Complication
LTR and RTL specific style rules is sometimes accomplished through `[dir="ltr"]` and `[dir="rtl"]` selectors. This approach will not work with style encapsulation without setting the dir attribute on every custom element w/ encapsulated styles.

#### Behavior stylesheets
We could write a Behavior to attach/detach based on document direction. While likely the most robust solution, this would relies on JavaScript execution for every component on initialization. Additionally, accurately [determining document direction](https://www.w3.org/TR/html401/struct/dirlang.html) is not the most straight forward operation.

#### CSS Custom Properties + calc
Value inversion is easily accomplished through CSS [calc](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) and [CSS Custom Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*):

```css
html {
   /* LTR */
   --dir-inverter: 1;

   /* RTL */
   --dir-inverter: -1;
}
/* ... */
:host {
    left: calc(var(--dir-inverter) * 20px);
}
```

Additionally, *certain* property replacements can be accomplished w/ CSS Custom Properties and calc. Take the following where `right` and `left` should switch:

```css
html {
    /* LTR */
    --ltr: 1;
    --rtl: dskfjadsklj; /* gibberish that invalidates a calc statement */

    /* RTL */
    --ltr: dskfjadsklj; /* gibberish that invalidates a calc statement */
    --rtl: 1;
}


:host {
   left: calc(var(--ltr) * 20px);
   right: calc(var(--rtl) * 20px);
}
```

I don't believe that 2. is possible w/ CSS Calc + CSS Custom Properties.

#### Others?
I'm not certain what our other options are - please chime in with ideas if you have them.

nicholasrice:
It was brought to my attention that there exist more logical properties than margin/padding-inline-*:
https://css-tricks.com/css-logical-properties/

The properties include border and positioning, though positioning seems to be behind flags in edge/chrome

nicholasrice:
For robustness, my recommendation is to add a `direction` property to the design system provider:
[...]

PR #3535, opened 2020-07-21
feat: adds direction property to design system

nicholasrice:
# Description
This change adds a `direction` property to the design systems for FAST and MSFT to track the primary document direction. It also adds two behaviors to resolve CSS custom properties to direction-based values based on the design-systems direction. Ideally this wouldn't be necessary, but there are still significant gaps in browser support for https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties.

Related to #3246. This addresses most of the issues in that thread but does not add the behavior to conditionally apply `ComposableStyles` based on direction. That piece is not blocking so taking as a follow-up.

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
@nicholasrice Do you happen to have a practical example of how this would be used in a component? That might be good to add to the PR.

nicholasrice:
@EisenbergEffect I'll add examples to the API docs

PR #3539, opened 2020-07-22
fix: update nested to be observable and set isNestedItem method to readonly

chrisdholt:
# Description
This PR makes a couple of small updates to the tree view and tree item components based on feedback provided after merging #3528.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -212,7 +212,7 @@ export class TreeItem extends FASTElement {
     }
 
     public readonly isNestedItem = (): boolean => {
-        return !!isTreeItemElement(this.parentElement as Element);
+        return isTreeItemElement(this.parentElement as Element);
     };
 
     private handleArrowLeft(): void {
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -84,12 +84,9 @@ export class TreeItem extends FASTElement {
     }
 
     /**
-     * Tracks the nested state of the item
-     * @public
-     * @remarks
-     * HTML Attribute: nested
+     * @internal
      */
-    @attr({ mode: "boolean" })
+    @observable
     public nested: boolean;
 
     @observable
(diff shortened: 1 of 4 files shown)


packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -214,11 +208,11 @@ export class TreeItem extends FASTElement {
         return treeChildren ? treeChildren.length : 0;
     }
 
-    public isNestedItem(): null | boolean {
-        return (
+    public readonly isNestedItem = (): boolean => {
+        return !!(
```
radium-v, line 212: Can this be made a little clearer? Would this work?

```suggestion
        return isTreeItemElement(this.parentElement as HTMLElement);
```

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -83,13 +83,7 @@ export class TreeItem extends FASTElement {
...
     }
 
-    /**
-     * Tracks the nested state of the item
-     * @public
-     * @remarks
-     * HTML Attribute: nested
-     */
-    @attr({ mode: "boolean" })
+    @observable
```
EisenbergEffect, line 86: Let's mark this as `@internal` as well I think.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
chrisdholt: For the most part - may still need to ensure boolean, but updating to use isTreeEl would be a good use of existing code. I'll update

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -39,7 +39,7 @@ export class TreeView extends FASTElement {
             if (this.checkForNestedItems()) {
                 this.slottedTreeItems.forEach(node => {
                     if (isTreeItemElement(node)) {
```
nicholasrice, line 41: I'm noticing `isTreeItemElement` is a type guard for `HTMLElement` - should it be a type guard for `TreeItem`?

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -214,11 +211,9 @@ export class TreeItem extends FASTElement {
         return treeChildren ? treeChildren.length : 0;
     }
 
-    public isNestedItem(): null | boolean {
-        return (
-            this.parentElement && this.parentElement.getAttribute("role") === "treeitem"
-        );
-    }
+    public readonly isNestedItem = (): boolean => {
+        return !!isTreeItemElement(this.parentElement as Element);
```
nicholasrice, line 215: `isTreeItemElement` always returns a bool so you don't need to `!!`

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
chrisdholt: Technically we support anything with a role of tree item, I'm not sure we want to type guard it for our class here.

PR #3540, opened 2020-07-22
fix: move text field and text area appearances from foundation to components

chrisdholt:
# Description
This PR moves the final "appearances" residing in foundation to the components layer. Appearances and the semantics surrounding them are specific to component instances and design systems. Moving this opinionated layer to the components layer.

## Issue type checklist

Users leveraging the components will experience no disruption. Users leveraging this functionality from `fast-foundation` will need to include the appearance in their class. Users leveraging the enum will need to update to leverage the type.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/text-field/index.ts
+++ b/packages/web-components/fast-components/src/text-field/index.ts
@@ -37,6 +37,10 @@ export class FASTTextField extends TextField {
      */
     @attr
     public appearance: TextFieldAppearance;
+
+    /**
+     * @internal
+     */
     public appearanceChanged(
         oldValue: TextFieldAppearance,
         newValue: TextFieldAppearance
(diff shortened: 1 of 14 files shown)


packages/web-components/fast-components/src/text-field/index.ts
```diff
@@ -21,4 +27,34 @@ import { TextFieldStyles as styles } from "./text-field.styles";
...
+    /**
+     * The appearance of the element.
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: appearance
+     */
+    @attr
+    public appearance: TextFieldAppearance;
+    public appearanceChanged(
```
nicholasrice, line 40: the `appearanceChanged()` should be marked as `@internal`

chrisdholt: done - I'll create an issue to run the repo as there are others which don't have this pattern setup either 👍

chrisdholt:
> I think we can go with this for now but we'll want to rework it a bit as part of the provider work or maybe as part of the React integration story.

Agreed - I anticipate how this is applied to be different with the provider work, but I think we need to move it out of foundation ASAP. As for the classes, that's simply to retain parity. We should do another pass to change our styling to map to attributes.

PR #3549, opened 2020-07-24
feat: add tooltip component

scomea:
# Description

This pull request adds a tooltip component.

noteworthy bits:
- it uses anchored region for placement
- tooltips render with fixed placement so they break out of parent containers and can easily appear on top of everything with the right z-index.  Down side they don't scroll with their parent container.  This seems desirable as a default behavior to me. Is it? Do we (initially) need other options like being able to associate tooltips with a different viewport so they could move with scrolled content, be clipped by parent container, etc...?  
- current default placement logic is above or below the anchor depending on available space.  Authors can specify top/right/bottom/left. Is this adequate for the initial version?  Anchored region supports much finer positioning control so at a minimum we should ensure that we don't need to make breaking changes to take advantage later.
- tooltips are super thin until they are shown, before then they just render the outer template to the dom and run a minimal amount of code.  This should be excellent for loading perf on pages that have many of these, but that does mean items in the tooltip slot are not rendered to the dom before then.  Does this need to be controllable?
 

## Motivation & context
We want a tooltip component

## Issue type checklist

**Adding or modifying component(s) in `@microsoft/fast-components` checklist*

## Process & policy checklist


EisenbergEffect:
My thoughts on the questions:

* I don't think tooltips need to scroll. I did a quick test and from what I can tell the native tooltips don't scroll.
* Current placement behavior and options seem good to me.
* I don't think we need to render the tooltip slot content when it's now showing. I wouldn't make this controllable in our first version. If we get significant customer request for that, we could add it later. Not sure why we would really need that though.

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
```diff
@@ -0,0 +1,352 @@
...
+                );
+                this.anchorElement.addEventListener(
+                    "mouseout",
+                    this.handleAnchorMouseOut,
+                    { passive: true }
+                );
+
+                const anchorId: string = this.anchorElement.id;
+
+                document.querySelectorAll(":hover").forEach(element => {
```
EisenbergEffect, line 68: Can't rely on being able to find it in the document. Might need to use `this.getRootNode()`.

scomea: Even when wrapped in a "$fastController.isConnected" check?

EisenbergEffect: No, because if your component is inside of another component's shadow dom, a selector executed against the document will not be able to fin it. You need to get the root node (which might be the document...but could be another shadow root) and then run the query selector on that.

scomea: So whenever we want to call something on document from code in a web component we should recursively call getRootNode until the return type is HTMLDocument?

EisenbergEffect: You don't need to recursively call it. Calling it once will get you the nearest shadow root that your element is in, or the document if it's not in a shadow DOM.

EisenbergEffect: Maybe I misunderstood the code...are you trying to find all :hover in the same "scope" as the current element? If so, then you would use getRootNode()

scomea: No, I'm actually trying to figure out if the anchor is being hovered at that point in time. It could be anywhere in the document.

EisenbergEffect: In the document or in the DOM scope that the tooltip is being used within?

EisenbergEffect: There can be N doms in a web page. The document is one but so is each shadow DOM. The tooltip could exist directly in the document or also inside of a shadow dom. My assumption is that the tooltip is used in the same dom as its anchor, in which case you would need to use getRootNode(). If the anchor can be anywhere...then we probably need to change the design of this component or at least introduce the restriction that the tooltip and the anchor need to exist in the same DOM.

scomea: Ok, so I can query the anchor element itself to determine hover, which is actually way better than querying the document.  But there are other places where we query the document for elements by id to get the anchor and viewport.  I think that's ok as that's the most basic use case and we present more advanced authors who may be working in a shadow dom with the ability to pass in those elements directly via JS.

EisenbergEffect: I think that sounds fine.

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+        overflow: visible;
+        height: 0;
+        width: 0;
+    }
+
+    .tooltip {
+        box-sizing: border-box;
+        border-radius: calc(var(--corner-radius) * 1px);
+        border: calc(var(--outline-width) * 1px) solid ${accentFillActiveBehavior.var};
+        border-color: ${neutralFocusBehavior.var};
```
chrisdholt, line 29: Can we drop the accent fill behavior from line 28, assuming line 29 is overriding it.

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+        background: ${neutralFillRestBehavior.var};
+        color: ${neutralForegroundRestBehavior.var};
+        padding: 4px;
+        height: fit-content;
+        width: fit-content;
+        font: inherit;
+        font-size: var(--type-ramp-base-font-size);
+        line-height: var(--type-ramp-base-line-height);
+        white-space: nowrap;
+        z-index: 10000;
```
chrisdholt, line 40: I see the 10000, I wonder if we should create a semantic ramp for values that we can invoke. This also may be beneficial as a var with a default value. That would make it easy to manage as the owner of an app - especially if CSS Parts are not supported in some of my primary browsers.

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+        margin-top: 4px;
+    }
+
+    :host(.left) .tooltip {
+        margin-right: 4px;
+    }
+
+    :host(.right) .tooltip {
+        margin-left: 4px;
+    }
```
chrisdholt, line 74: Will these need to localize? Perhaps instead of left/right we should use language similar to the CSS logical properties and map to those? Curious on thoughts from @EisenbergEffect, @nicholasrice, others?

```
:host(.inline-start) {
    margin-inline-start: 4px;
}
```

packages/web-components/fast-foundation/src/tooltip/README.md
```diff
@@ -0,0 +1,8 @@
+---
+id: fast-tooltip
+title: fast-tooltip
+sidebar_label: fast-tooltip
+custom_edit_url: https://github.com/microsoft/fast-dna/edit/master/packages/web-components/fast-foundation/src/tooltip/README.md
+---
+
+## Applying Custom Styles
```
chrisdholt, line 8: Should this be filled out?

packages/web-components/fast-foundation/src/tooltip/tooltip.spec.md
```diff
@@ -30,7 +30,7 @@ Tooltip widgets do not receive focus. A hover that contains focusable elements c
 *Attributes:*
 - `anchor` - The html id of the HTMLElement which the tooltip is attached to.
 - `delay` - time in milliseconds to wait before showing and hiding the tooltip. Defaults to 300.
-- `visible` - the visiblity of the tooltip
+- `hidden` - boolean value to toggle the visibility of the tooltip (defaults to undefined).
```
chrisdholt, line 33: Thanks! 😄

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
EisenbergEffect: That makes good sense to me. If that's not enough we also have the new RTL behaviors that Nick wrote. Those are going to be more heavy weight than a standard CSS property though. Just mentioning it for completeness.

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
EisenbergEffect: I think we should have a central way to manage the zindex for modals, tooltips, select popups, etc. For example, what if we have a modal with a select in it and there's a tooltip on something in the select popup? Do we alway treat tooltips as highest, then popups, then modals? Would something simple like that work? Can we just encode that in a simple lookup in foundation that can be used by various components and design systems?

chrisdholt: Yeah, there are a few ways that we've done this before - I'm happy to open an issue and propose something.

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
chrisdholt: For sure - I like the logical aspect of the naming; we did something similar with the start/end slots.

scomea: The current configuration works in rtl.  The host "left"/"right" classes are applied based on the actual position of the region relative to the anchor.  If the region is left we want a margin on the right and vice versa.

scomea: Anchored region supports "start/end" as a choice for horizontal position when authors want position to change based on direction.  I shall add that as a position choice as well as "left/right".

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
chrisdholt: I think it's fine for now - we don't have to block on this. Could you add a note w/ a TODO so I can find it :)

packages/web-components/fast-components/src/tooltip/fixtures/base.html
```diff
@@ -0,0 +1,151 @@
+<fast-design-system-provider use-defaults>
+    <h1>Tooltip</h1>
+
+    <h4>Default</h4>
```
awentzel, line 4: Our headings should be semantic for those using screen readers.  They seem to be used stylistically when they should be used contextually and semantically.  So H4 really should be H2.  Additionally, should we be attributing with styles?  As documentation, should we be using our own design system just like a production scenario would?

packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
```diff
@@ -0,0 +1,39 @@
+import { STORY_RENDERED } from "@storybook/core-events";
+import addons from "@storybook/addons";
+import { FASTDesignSystemProvider } from "../design-system-provider";
+import TooltipTemplate from "./fixtures/base.html";
+import { FASTTooltip } from "./";
+
+// Prevent tree-shaking
```
awentzel, line 7: I'd prefer to see an explanation of why we are preventing tree-shaking in the comment.

packages/web-components/fast-components/src/tooltip/fixtures/base.html
chrisdholt: Can we create an issue to do a larger pass here? This is an issue across all controls and not specific to this one, outside that I agree - this is net new and should be correct.

packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
chrisdholt: I think this comment is clear enough for the stories page. We prevent tree shaking here because without the Class defined or used in this it will be tree shaken out.

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+        background: ${neutralFillRestBehavior.var};
+        color: ${neutralForegroundRestBehavior.var};
+        padding: 4px;
+        height: fit-content;
+        width: fit-content;
+        font-family: var(--body-font);
+        font-size: var(--type-ramp-base-font-size);
+        line-height: var(--type-ramp-base-line-height);
+        white-space: nowrap;
+        // TODO: a mechanism to manage z-index across components
```
awentzel, line 39: Can you link the corresponding `todo` with an GH issue.

EisenbergEffect:
@chrisdholt Did you want to try to get this in for Thursday's release? Anything else not assigned to a release that you want to try to get in?

chrisdholt:
> @chrisdholt Did you want to try to get this in for Thursday's release? Anything else not assigned to a release that you want to try to get in?

Would love this to make it into this release, but *only* if it's ready. Let's not hold for it, but would be great to have it in!

scomea:
Let's get the review moving forward without  rushing.

Get Outlook for Android<https://aka.ms/ghei36>

marjonlynch:
I think you want this showing up in fast-component-explorer and in the Docs, I'm not seeing it in either. I'm pretty sure at least you will need to update sidebar.js for docs among other things

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
```diff
@@ -0,0 +1,462 @@
...
+        if ((this as FASTElement).$fastController.isConnected) {
+            this.updateTooltipVisibility();
+            this.updateLayout();
+        }
+    }
+
+    /**
+     * The id of the element the tooltip is anchored to
+     *
+     * @defaultValue - undefined
```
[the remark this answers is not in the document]
scomea: I think leaving it as undefined is fine.  Not sure about how we should handle the @defaultValue when that is the case tho.

packages/web-components/fast-components/src/tooltip/fixtures/base.html
scomea: https://github.com/microsoft/fast/issues/3812

packages/web-components/fast-components/src/tooltip/tooltip.styles.ts
scomea: https://github.com/microsoft/fast/issues/3813

scomea:
Should show up in the component explorer now

marjonlynch (approved):
Approving with 2 suggestions
1. Update the component explorer example to have a component with the tooltip anchored instead of just a tooltip
2. Write a spec based on Chris's PR that just went in adding all the specs

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
```diff
@@ -0,0 +1,469 @@
...
+     * The timer that tracks delay time before the tooltip is shown on hover
+     */
+    private delayTimer: number | null = null;
+
+    /**
+     * Indicates whether the anchor is currently being hovered
+     */
+    private isAnchorHovered: boolean = false;
+
+    constructor() {
```
nicholasrice, line 231: You don't need the constructor if you're only calling `super()`

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
```diff
@@ -0,0 +1,469 @@
...
+        super.disconnectedCallback();
+    }
+
+    /**
+     * invoked when the anchored region's position relative to the anchor changes
+     *
+     * @internal
+     */
+    public handlePositionChange = (ev: Event): void => {
+        this.classList.toggle("top", this.region.verticalPosition === "top");
```
nicholasrice, line 255: Is there a reason this work isn't being done in the template where it is batched?

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
```diff
@@ -0,0 +1,469 @@
...
+        this.region.viewportElement = this.viewportElement;
+        this.region.anchorElement = this.anchorElement;
+        (this.region as any).addEventListener("change", this.handlePositionChange);
+    };
+
+    /**
+     *  gets the current direction
+     */
+    private getDirection = (): Direction => {
+        const closest: Element | null = this.closest(
```
nicholasrice, line 460: There is a utility in fast-foundation for this now; `getDirection`

nicholasrice (approved):
Looks good! Left a few notes but signed off

packages/web-components/fast-foundation/src/tooltip/README.md
chrisdholt: @scomea - *did this get filled out*?

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
scomea: switched to that, thx!

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
scomea: There used to be some code there, removedd.

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
scomea: I like that this approach for core functionality does not depend on the template side implementation which can be swapped out.  Should revisit if we suspect a perf cost though.

chrisdholt (approved):

EisenbergEffect:
@scomea Can you do a follow on PR to make sure the readme matches the other components and then add this component into the list of components in `sidebars.js` in the site? Also, if we want to ship anchored region for external use we'd need to do the same with that as well as add it from the exclude in the `prepare-docs.js` script.

PR #3552, opened 2020-07-24
docs(spec): add breadcrumb spec

khamudom:
# Description

The file has specs for `fast-breadcrumb` and `fast-breadcrumb-item`

image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/specs/breadcrumb.md
+++ b/specs/breadcrumb.md
@@ -37,7 +37,7 @@ He clicks on `Kitchen and Dining` to go back to the category of kitchen items, s
 ```
 
 *Slot Names*
-- default slot for items
+- default slot for items.
 
 ---
 
@@ -46,10 +46,10 @@ He clicks on `Kitchen and Dining` to go back to the category of kitchen items, s
 ```html
 <fast-breadcrumb>
     <fast-breadcrumb-item>
-        <fast-anchor appearance="lightweight">Home</fast-anchor>
+        <fast-anchor appearance="lightweight" href="#">Home</fast-anchor>
     </fast-breadcrumb-item>
     <fast-breadcrumb-item>
-        <fast-anchor appearance="lightweight">Templates</fast-anchor>
+        <fast-anchor appearance="lightweight" href="#">Templates</fast-anchor>
     </fast-breadcrumb-item>
     <fast-breadcrumb-item aria-current="page">
         <fast-anchor appearance="lightweight">Template Prices</fast-anchor>
@@ -82,12 +82,16 @@ The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It 
 *Component Name*
 - `fast-breadcrumb-item`
 
+ *Attributes:*
+- `aria-current` - aria attribute set on the last item of breadcrumb.
+- `separator` - a divider between two breadcrumb items. The default of this attribute is `/`.
+
 *Slots*
-- default slot for element
+- default slot for element.
 - `separator` - used as part of the visual presentation that signifies the breadcrumb trail. The separator is defaulted to a `/`, but authors can override to use an `svg`.
 
 *CSS Parts*
-- `separator` - there is a style to adjust margin and another to hide the separator when a breadrumb item is set to `aria-current="page"`
+- `separator` - there is a style to adjust margin and another to hide the separator when a breadrumb item is set to `aria-current="page"`.
 
 ### Anatomy and Appearance
 
@@ -96,7 +100,6 @@ The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It 
     role="listitem"
     class="content"
     part="content"
-    tabindex="0"
     aria-current="${x => x.ariaCurrent}"
 >
     <slot></slot>
@@ -114,15 +117,15 @@ The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It 
 
 ```html
 <fast-breadcrumb-item>
-    <fast-anchor appearance="lightweight">Home</fast-anchor>
+    <fast-anchor appearance="lightweight" href="#">Home</fast-anchor>
 </fast-breadcrumb-item>
 ```
 
-*Overriding separator*
+*Overriding default separator*
 
 ```html
 <fast-breadcrumb-item>
-    <fast-anchor appearance="lightweight">Home</fast-anchor>
+    <fast-anchor appearance="lightweight" href="#">Home</fast-anchor>
     <span slot="separator">
         <svg>
             <path .../>
--- a/specs/breadcrumb.md
+++ b/specs/breadcrumb.md
@@ -31,7 +31,7 @@ He clicks on `Kitchen and Dining` to go back to the category of kitchen items, s
 
 ### Anatomy and Appearance
 
-```ts
+```html
 <template
     role="navigation"
     aria-label="breadcrumb"
@@ -51,21 +51,16 @@ He clicks on `Kitchen and Dining` to go back to the category of kitchen items, s
 
 ```html
 <fast-breadcrumb>
-    <fast-breadcrumb-item>
-        <fast-anchor appearance="lightweight" href="#">Home</fast-anchor>
-    </fast-breadcrumb-item>
-    <fast-breadcrumb-item>
-        <fast-anchor appearance="lightweight" href="#">Templates</fast-anchor>
-    </fast-breadcrumb-item>
-    <fast-breadcrumb-item>
-        <fast-anchor appearance="lightweight">Template Prices</fast-anchor>
-    </fast-breadcrumb-item>
+    <fast-breadcrumb-item name="Home" href="#"></fast-breadcrumb-item>
+    <fast-breadcrumb-item name="Template" href="#"></fast-breadcrumb-item>
+    <fast-breadcrumb-item name="Template Prices" href="#"></fast-breadcrumb-item>
 </fast-breadcrumb>
 ```
 
 ### Accessibility
 
 The breadcrumb should align to the interaction model provided by the [W3C](https://www.w3.org/TR/wai-aria-practices/#breadcrumb)
+- `aria-current` - W3 specs says when using `aria-current="page"` the element should be a link. `fast-breadcrumb` will set `aria-current="page"` to the last element if there is an href attribute.
 
 ### Globalization
 
@@ -81,7 +76,7 @@ While testing is still TBD for our web components, I would expect this to align 
 
 ## Overview
 
-The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It provides a default slot to implement any element. There is also a slot for a separator that defaults using a `/`.
+The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It provides a slot with a default anchor. The component also provides a slot for a separator that defaults using a `/`.
 
 ### API
 
@@ -92,7 +87,7 @@ The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It 
 - `showSeparator` - is a boolean to show and hide the separator.
 
 *Slots*
-- default slot for element.
+- a slot that has an `<a></a>` as a default control.
 - `separator` - used as part of the visual presentation that signifies the breadcrumb trail. The separator is defaulted to a `/`. To override the default separator, add an element with an attribute of `slot="separator"`.
 
 *CSS Parts*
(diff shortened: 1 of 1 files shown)


specs/breadcrumb.md
```diff
@@ -0,0 +1,132 @@
...
+<fast-breadcrumb>
+    <fast-breadcrumb-item>
+        <fast-anchor appearance="lightweight">Home</fast-anchor>
+    </fast-breadcrumb-item>
+    <fast-breadcrumb-item>
+        <fast-anchor appearance="lightweight">Templates</fast-anchor>
+    </fast-breadcrumb-item>
+    <fast-breadcrumb-item aria-current="page">
+        <fast-anchor appearance="lightweight">Template Prices</fast-anchor>
+    </fast-breadcrumb-item>
```
chrisdholt, line 56: Is there a reason to have `fast-breadcrumb-item` if it's going to include fast anchors?

specs/breadcrumb.md
```diff
@@ -0,0 +1,132 @@
...
+The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It provides a default slot to implement any element, and also contains the separator.
+
+### API
+
+*Component Name*
+- `fast-breadcrumb-item`
+
+*Slots*
+- default slot for element
+- `separator` - used as part of the visual presentation that signifies the breadcrumb trail. The separator is defaulted to a `/`, but authors can override to use an `svg`.
```
chrisdholt, line 87: Should this be in the item or in `fast-breadcrumb`?

specs/breadcrumb.md
```diff
@@ -0,0 +1,132 @@
...
+### Anatomy and Appearance
+
+```ts
+<div
+    role="listitem"
+    class="content"
+    part="content"
+    tabindex="0"
+    aria-current="${x => x.ariaCurrent}"
+>
```
chrisdholt, line 101: I'm not sure we want to focus on the list item if the focusable element is within this. Additionally I think we'd want aria-current on the `link` that's slotted in here. This seems extraneous and potentially unnecessary if the primary goal is to serve as a wrapper for the link and a visual separator.

Is there a way to simplify here? Do we need this layer?

EisenbergEffect:
@khamudom and I worked through a lot of options and there was definitely a difficulty in dealing with getting to right structure and incorporating the separator. It seems like the only way to get the separator is to have a wrapper element to render it because there's no way to wrap individual slotted elements.

As I was reading the comments I did come up with a new idea though. Caveat: it's dependent on the new work in fast-element and the proposed new FoundationElement base class. With the new base class, all of our components would have a `template` property to allow for dynamically swapping the template of an element. With that in place, we could do something really interesting...we could have the breadcrumb component grab the template of each of its children, and then create a new template that wraps the current template so that it can apply the configured separator. It can then assign the new composed template back to the anchor instances. What do you all think of that idea?

chrisdholt:
@EisenbergEffect I was rolling around something like that in my head but wasn’t sure how to reconcile with composability. I think that would be preferred here and a welcome addition. It’s worth waiting for that work in my opinion.

EisenbergEffect:
There are a few more tricks to it since the child anchors would not be connected when the breadcrumbs connected callback fires. So, the child templates wouldn't be in place yet to wrap. I've got a couple of solutions in mind to it and will keep thinking about it a bit more. It might be worth introducing "template decorators" as an official concept for things like this and have some standard way of doing it. Need to think about it a bit more.

radium-v:
Would [css counter styles](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters) work for the separators?

EisenbergEffect:
@radium-v I'm not sure. It probably depends on how flexible we want to make the separator. What if someone has a custom SVG, for example, do we want to support that or just a basic character?

radium-v:
@EisenbergEffect my bad, I was thinking about [`list-style` properties](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style) which do allow for images. Interestingly, the [breadcrumb example linked in the README](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html) uses empty `::before` pseudo-elements to avoid excessive verbosity with screen readers.

EisenbergEffect:
Is it possible to apply a list style to something that isn't actually an li? If we give it an aria `listitem` role, would that styling work? I'd certainly prefer that type of solution if possible.

specs/breadcrumb.md
```diff
@@ -0,0 +1,135 @@
...
+The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It provides a default slot to implement any element, and also contains the separator.
+
+### API
+
+*Component Name*
+- `fast-breadcrumb-item`
+
+ *Attributes:*
+- `aria-current` - aria attribute set on the last item of breadcrumb.
+- `separator` - a divider between two breadcrumb items. The default of this attribute is `/`.
```
nicholasrice, line 87: Thoughts on allowing this to be a `html` template instance? That would provide authors far more options than just a content character

EisenbergEffect: In this case do you mean a vanilla template (not a FAST template with bindings)?

specs/breadcrumb.md
```diff
@@ -0,0 +1,135 @@
...
+
+The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It provides a default slot to implement any element, and also contains the separator.
+
+### API
+
+*Component Name*
+- `fast-breadcrumb-item`
+
+ *Attributes:*
+- `aria-current` - aria attribute set on the last item of breadcrumb.
```
janechu, line 86: Should it be noted here that this is always `aria-current="page"`?

specs/breadcrumb.md
```diff
@@ -0,0 +1,135 @@
...
+```ts
+<div
+    role="listitem"
+    class="content"
+    part="content"
+    aria-current="${x => x.ariaCurrent}"
+>
+    <slot></slot>
+    <span
+        class="separator ${x => (x.ariaCurrent ? "hide" : "")}
```
janechu, line 107:
```suggestion
        class="separator ${x => (x.ariaCurrent ? "hide" : "")}"
```

khamudom: Thanks

specs/breadcrumb.md
khamudom: Will not that.

specs/breadcrumb.md
khamudom: removed the `separator` attribute. This was to set a default separator. Will used the slot instead to set a default.

specs/breadcrumb.md
khamudom: `fast-anchor` is just a component you can use as a breadcrumb item. breadcrumb-item has a default slot to implement any control and also has the separator.

specs/breadcrumb.md
khamudom: updated to set aria-current on the control inside the breadcrumb item.

specs/breadcrumb.md
chrisdholt: In the current spec that doesn't seem to be the case; I need to slot fast-anchor into the fast-breadcrumb-item, which seems like an unnecessary abstraction here.

specs/breadcrumb.md
chrisdholt: We should call out that aria-current is *only* necessary if the element is a link. If this doesn't have an href it'll be read incorrectly by AT.

specs/breadcrumb.md
khamudom: I agree, does feel unnecessary. The issue I am currently running into is a way to inject the separator between each item and also have separator be any element. That is why we have breadcrumb-item. 
I am going to keep investigating this so we don't need a breadcrumb-item.

specs/breadcrumb.md
khamudom: I'll update the accessibility section explaining this. Thanks

khamudom:
Updated the fast-breadcrumb-item. instead of having an empty slot, added am anchor as a default control. This cleans up the html, it avoids having to add another control inside fast-breadcrumb-item.

specs/breadcrumb.md
```diff
@@ -0,0 +1,139 @@
...
+- slotted directive, with a property called `slottedBreadcrumbItems`.
+
+---
+
+## Implementation
+
+```html
+<fast-breadcrumb>
+    <fast-breadcrumb-item name="Home" href="#"></fast-breadcrumb-item>
+    <fast-breadcrumb-item name="Template" href="#"></fast-breadcrumb-item>
```
nicholasrice, line 55: I like this a lot better - not needing to add an an anchor inside. nice!

specs/breadcrumb.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+### Accessibility
+
+The breadcrumb should align to the interaction model provided by the [W3C](https://www.w3.org/TR/wai-aria-practices/#breadcrumb)
+- `aria-current` - W3 specs says when using `aria-current="page"` the element should be a link. `fast-breadcrumb` will set `aria-current="page"` to the last element if there is an href attribute.
+
+### Globalization
+
+The first and last element inside the breadcrumb will be reversed when switching in LTR/RTL.
```
chrisdholt, line 67: Nit: I think this would be the overall order is reversed. If there are only three items, this is true - if there are six, not as much :).

specs/breadcrumb.md
```diff
@@ -0,0 +1,139 @@
...
+- `listitem-container` - class style to align the default slot content and the separator.
+- `separator` - class style to adjust margin layout
+
+### Anatomy and Appearance
+
+```html
+<div
+    role="listitem"
+    class="listitem-container"
+    part="listitem-container"
```
chrisdholt, line 103: Just an FYI to catch this prior to implementation...I think this can just be `listitem`; it doesn't contain the list item, it's serving as the list item.

specs/breadcrumb.md
```diff
@@ -0,0 +1,139 @@
...
+        <a class="control" part="control" href="${x => x.href}">
+            <span class="content" part="content">
+                ${x => x.name}
+            </span>
+        </a>
+    </slot>
+    ${when(
+        x => x.showSeparator,
+        html<BreadcrumbItem>`
+            <span class="separator" aria-hidden="true">
```
chrisdholt, line 115: You'll want the part here where you have the class.

specs/breadcrumb.md
khamudom: Good point. Thanks

specs/breadcrumb.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `fast-breadcrumb-item` is placed inside the `fast-breadcrumb` component. It provides a slot with a default anchor. The component also provides a slot for a separator that defaults using a `/`.
+
+### API
+
+*Component Name*
+- `fast-breadcrumb-item`
+
+*Properties*
+- `showSeparator` - is a boolean to show and hide the separator.
```
chrisdholt, line 87: If I'm using fast-breadcrumb with these items, do I need to set this on every item if I want to change from the default (assuming the default is true, I want to hide it and set false)? Or is the intent with this being a property that we have a method of hiding all nested fast-breadcrumb-items separators if we want?

specs/breadcrumb.md
```diff
@@ -0,0 +1,139 @@
...
+
+*Attribute*
+- aria-current - attribute set on the last breadcrumb item, if it has an `href` set. It will always be `aria-current="page"`.
+
+### Anatomy and Appearance
+
+```html
+<template
+    role="navigation"
+    aria-label="breadcrumb"
```
chrisdholt, line 37: Is this aria-label part of the implementation perhaps? I'm assuming you don't intend to hardcode this as it needs to localize. Additionally, I believe that someone could also use aria-labelledby in place of the aria-label. How do we anticipate supporting both? Should it be guidance or should we have first class attributes for these? One _is_ required: https://www.w3.org/TR/wai-aria-practices-1.2/#wai-aria-roles-states-and-properties-2

specs/breadcrumb.md
khamudom: Yes, `showSeparator` is true by default. We can create a method that will hide all separator, currently, my thought was to only hide the separator on the last item. In Breadcrumb, I will use the `instanceof` BreadcrumbItem, and on the last item, set `showSeparator` to false. 
You also brought up a point on changing separator on each item from default. I will have to come up with a way to only change the separator once and not each item.

specs/breadcrumb.md
khamudom: I will create attributes for both `aria-label` and `aria-labelledby`

specs/breadcrumb.md
khamudom: Yes, `showSeparator` is defaulted true. In Breadcrumb, I would use `instanceof` BreadcrumbItem and set `showSeparator` on the last item to false. Since the separator visually represents a breadcrumb trail, I did not think to hide all of them.
To change the default of the separator, I will have to come up with a way to change it once, and not have to change it on all the items.

PR #3555, opened 2020-07-25
docs(spec): add avatar spec

dlesaca:
# Description

## Issue type checklist

## Process & policy checklist


specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+<test-avatar 
+  altText="Jenny's profile image"
+  link="">
+  <fast-badge slot="badge" 
+    fill="primary"
+    circular=>C</fast-badge>
+</test-avatar>
+```
+
+### Accessibility
```
EisenbergEffect, line 85: Are there any aria attributes that we should have for this, in particular because it's potentially acting as a link to a profile, for example?

specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+An avatar is a graphical represention of a user or object.
+
+### Use Cases
+
+A common use case would be to display an image or text (usually initials) of a user or an object.
+
+### Features
+- Image slot: Able to slot in an avatar image
+- Badge slot: Able to slot in a badge component
+- `size`, a numeric value that determines the size of the avatar
```
EisenbergEffect, line 14: I think this is the first time we have an instance of `size` in one of our components. Is this something we want to do broader thinking about? For example, some systems have `size` for buttons as well. If we want to enable that, I think we'd want to do it in a consistent way.

specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+- `fast-avatar`
+
+#### Attributes
+|   Name    | Description                                  | Type                |
+|-----------|----------------------------------------------|---------------------|
+| `name`    | Accepts name string for coin display         | `string`            |
+| `shape`   | Determines the avatar coin shape.            | `circle | square`   |
+| `fill`    | Sets the background fill of the avatar coin. | `string: hex color` |
+| `color`   | Sets the color of the avatar coin text.      | `string: hex color` |
+| `size`    | Sets the size of avatar coin.                | `number`            |
```
EisenbergEffect, line 40: Seems like image source and alt text may be missing here. Also, I think I saw some of it above, but might be worth some further clarity about which attrs are required or which combinations work.

specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Use Cases
+
+A common use case would be to display an image or text (usually initials) of a user or an object.
+
+### Features
+- Image slot: Able to slot in an avatar image
+- Badge slot: Able to slot in a badge component
+- `size`, a numeric value that determines the size of the avatar
+- `shape`, a circle or square shape can be chosen
```
EisenbergEffect, line 15: In the case of `square` would the corners be rounded based on the design system properties?

specs/avatar.md
EisenbergEffect: Also, is there a minimum size? For example, I noticed in some examples that when the size is small enough the image/coin is hidden entirely and only the badge is displayed.

specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+---
+
+## Implementation
+
+```
+<test-avatar 
+  imgSrc=""
+  altText="Jenny's profile image"
```
janechu, line 69: HTML does not have capital letters, these should be `img-src` and `alt-text`. I would also think the name of this would be `fast-avatar`. You also are specifying above that there is an image slot, yet here you create an image and apply some attributes to it that have been passed. Which are you intending to do?

specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+
+It is important to ensure that when the contrast of text in the coin meets [1.4.3 Contrast (Minimum)](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html).
+
+### Globalization
+
+If a badge is used it should appear on the appropriate side of the coin.
+
+### Documentation
+
+*What additions or changes are needed for user documentation and demos? Are there any architectural/engineering docs we should create as well, perhaps due to some interesting technical challenge or design decisions related to this component?*
```
janechu, line 95: These comments are meant to be removed and the section either filled in or removed.

specs/avatar.md
```diff
@@ -0,0 +1,105 @@
...
+
+---
+
+## Resources
+
+- [WCAG - 1.4.3 Contrast (Minimum)](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html)
+
+## Next Steps
+
+I believe some feature validation will be needed to determine whether some features should be natively added to this component or if they should be added by the developer party for their particular use case.
```
janechu, line 105: Are there any features in particular that come to mind?

specs/avatar.md
dlesaca: Thanks for the feedback. This is great. I'll think through this a little bit more and submit something.

specs/avatar.md
dlesaca: Ah yes. An oversight on my part. I'll revise this.

specs/avatar.md
dlesaca: Yes, that is my approach since radius can vary.

specs/avatar.md
dlesaca: Thanks for your feedback! I'll revise the attribute names. The intent is to have an image attribute. It seems I left an old idea in there. Will revise and remove.

specs/avatar.md
dlesaca: Will address. Thank you.

specs/avatar.md
dlesaca: I addressed this by making by conditionally adding a `role="aria-link` when a `link` is provided.

specs/avatar.md
dlesaca: I've tabled the addition of a `size` attribute until a broader sizing convention decision can be reached.

specs/avatar.md
dlesaca: This has been revised.

specs/avatar.md
dlesaca: I added a sentence to clarify this.

specs/avatar.md
dlesaca: This has been addressed.

specs/avatar.md
dlesaca: Reworked to be more specific.

specs/avatar.md
```diff
@@ -0,0 +1,122 @@
...
+    <div class="avatar-coin">
+      <a class="avatar-link" href="${ x => x.link}">
+        ${when(x => x.imgSrc, html`
+          <img 
+            class="avatar-image" 
+            alt="${ x => x.alt }" 
+            src="${ x => x.imgSrc }" />
+        `)}
+        <span class="avatar-name">${x => x.name}</span>
+      </a>
```
chrisdholt, line 60: Late to the party here, and this is 100% dependent on implementation -

In this case, I would expect we may need to manage tabindex as well as click and keypress events to mimic the native anchor element which is nested within here. An alternative to this could be to delegate focus to the anchor within. In that case, I would not expect the `aria-link` role as the host role will be ignored and focus will defer to the internal anchor. If we went that route, we could also consider using a single template and just not passing href if it's not provided. While that would still have an anchor in the shadow, it wouldn't receive focus and should be treated as text by AT.

Not an imperative that we build to the above, but something to consider which may simplify the control and implementation and allow us to offload native behavior to the native elements.

dlesaca: Thanks for the feedback!

This is good. VoiceOver does focus on the `a` tag itself when navigating to the component. The `alt` attribute is also read. I'll play around with some of the approaches you mentioned, my goal is to make the implementation as accessible as possible without over complicating the template.

dlesaca: I pushed some changes up that I think might be better. This will have `delegatesFocus` on. A little more of a simplified approach. The only thing is if no link is provided then there will be an empty `<a>` wrapping the `img` and `span`, which I've read is [okay](https://www.w3.org/TR/2011/WD-html5-20110525/text-level-semantics.html#the-a-element). Any thoughts, @EisenbergEffect, @chrisdholt?

chrisdholt: Sweet - yeah, that should be ok. AT should treat an anchor without an href like its contents. If it's text, such as below, it would be like text:
```
<a>This is text</a>
<a href="#">This is a link</a>
```

PR #3559, opened 2020-07-27
feat: adding directional stylesheet behavior

### LINKED ISSUE #3246: How are LTR and RTL specific CSS rules applied?

nicholasrice:
# Description
This change implements the directional stylesheet behavior described in #3246. This will allow component authors to supply a stylesheet to be applied for both LTR and RTL specific component instances.

This change adds the behavior to `@microsoft/fast-foundation` but **does** require use with a Design System Provider that implements a `direction` property set to either `"ltr"` or `"rtl"`.

closes #3246

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/docs/design/localization.md
```diff
@@ -0,0 +1,53 @@
...
+
+const styles = css`
+    :host {
+        float: ${inlineStartBehavior.var}
+    }
+`.withBehaviors(inlineStartBehavior)
+```
+
+### DirectionalStyleSheetBehavior
+[`DirectionalStyleSheetBehavior`](/docs/api/fast-foundation.directionalstylesheetbehavior/) can be used to apply arbitrary LTR and RTL stylesheets, depending on the nearest [`FASTDesignSystemProvider's direction`](/docs/api/fast-components-msft.fastdesignsystemprovider.direction/)` property.
```
janechu, line 29:
```suggestion
[`DirectionalStyleSheetBehavior`](/docs/api/fast-foundation.directionalstylesheetbehavior/) can be used to apply arbitrary LTR and RTL stylesheets, depending on the nearest [`FASTDesignSystemProvider`s direction](/docs/api/fast-components-msft.fastdesignsystemprovider.direction/) property.
```

packages/web-components/fast-components/docs/design/localization.md
```diff
@@ -0,0 +1,53 @@
...
+
+### DesignSystemProvider
+The [`FASTDesignSystemProvider`](/docs/api/fast-components-msft.fastdesignsystemprovider/) exposes a `direction` property. This should be set to the primary document direction and defaults to `ltr`. This value will be used to inform the stylesheet behaviors described below.
+
+### `inlineStartBehavior` and `inlineEndBehavior`
+[inlineStartBehavior](/docs/api/fast-components-msft.inlinestartbehavior/) and [inlineEndBehavior](/docs/api/fast-components-msft.inlineendbehavior/) can be used to apply localized [float](https://developer.mozilla.org/en-US/docs/Web/CSS/float) properties. These are drop-in replacements for the browsers `inline-start` and `inline-end` float values and should be used when the native values are not supported. If your browser-matrix supports `inline-start` and `inline-end` float values, please use the native values.
```
EisenbergEffect, line 14: Looks like you are linking from components-msft but importing from components. Should this be moved to foundation as well? Do we need to clarify this more?

nicholasrice: oops I'll correct

PR #3597, opened 2020-07-30
fix: update fast-card and component sampler color handling

bheston:
# Description

Updated fast-card to use a relative color recipe.
Updated the fast.design site component sampler to better respond to changes of the neutral color and light or dark mode,

image

## Motivation & context

A card has two applicable color recipes for it's background color, but the most flexible one is `neutralFillCardRest` because it will adjust the card background for whatever container it's in. `neutralLayerCard` is a fixed value intended for an optimized and backwards compatible case over `neutralLayerCardContainer` only. This was causing the fast.design component sampler to not reflect the color change correctly.

Further, the sampler is updated to adjust both the layer luminance and the neutral color used for the background. The latter shows off more of the color handling with the addition of non-grey swatches. The former more correctly applies light and dark mode, which is done setting the `baseLayerLuminance` and using a `layer` recipe for the background color instead of a fixed shade of grey that doesn't come from the neutral ramp.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/fast-frame/fast-frame.ts
+++ b/sites/fast-website/src/app/components/fast-frame/fast-frame.ts
@@ -184,8 +184,15 @@ export class FastFrame extends FASTElement {
         this.saturation = accentColorHSL.s;
         this.lightness = accentColorHSL.l;
 
-        this.updateBackgroundColor();
-
         this.mql.addListener(this.resetExpandedResponsive);
+    }
+
+    /**
+     * @internal
+     */
+    public connectedCallback() {
+        super.connectedCallback();
+
+        this.updateBackgroundColor();
     }
 }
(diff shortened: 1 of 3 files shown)


sites/fast-website/src/app/components/fast-frame/fast-frame.ts
```diff
@@ -141,20 +148,25 @@ export class FastFrame extends FASTElement {
         this.accentPalette = createColorPalette(accentRGB);
     }
 
-    public themeChange = (e: CustomEvent): void => {
+    private updateBackgroundColor(): void {
```
chrisdholt, line 151: @bheston I might be missing it, but with the default value gone, should this be initially invoked during connected callback?

bheston: I wasn't entirely sure how the lifecycle worked, but I put the call in the constructor and it seemed to come up fine. I'm not sure how that works if used as an attribute though, or if it matters if it's defaulted or not.

chrisdholt: @bheston can we move to connectedCallback() to be safe here?

PR #3611, opened 2020-07-31
fix: add aria-label to item button to fix screen reader from reading out the Unicode

khamudom:
# Description

The screen reader was reading out the Unicode, so I added an aria-label attribute to the to-do item button. By default it reads out "Remove item".

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/examples/todo-app-tutorial/src/todo-app.ts
+++ b/examples/todo-app-tutorial/src/todo-app.ts
@@ -34,7 +34,7 @@ const template = html<TodoApp>`
                     </span>
                     <fast-button
                         @click=${(x, c) => c.parent.removeTodo(x)}
-                        aria-label="${x => x.ariaLabel}"
+                        aria-label="Remove item"
                     >
                         &times;
                     </fast-button>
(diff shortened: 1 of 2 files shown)


examples/todo-app-tutorial/src/todo-app.ts
```diff
@@ -32,7 +32,10 @@ const template = html<TodoApp>`
                     <span class="list-text ${x => (x.done ? "done" : "")}">
                         ${x => x.description}
                     </span>
-                    <fast-button @click=${(x, c) => c.parent.removeTodo(x)}>
+                    <fast-button
+                        @click=${(x, c) => c.parent.removeTodo(x)}
+                        aria-label="${x => x.ariaLabel}"
```
EisenbergEffect, line 37: Since the label isn't going to change, I'd be inclined to remove the attr and just hard code the value in the template. Thoughts?

khamudom: Yeah, if it is not going to change, then I am ok with just hard coding the value.

PR #3619, opened 2020-08-04
feat: add select spec

eljefe223:
# Description
Adds spec for `<fast-select/>` and `<fast-option/>`

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
@eljefe223 Doesn't look like I'll get a chance to review today but I'll try to take a look first thing in the morning. Real quick, I wanted to mention that there's a lot of work on select happening in OpenUI. If you haven't checked that out, take a look since we want to try to align with them as much as possible and we may then also have some feedback into that process.

chrisdholt:
> @eljefe223 Doesn't look like I'll get a chance to review today but I'll try to take a look first thing in the morning. Real quick, I wanted to mention that there's a lot of work on select happening in OpenUI. If you haven't checked that out, take a look since we want to try to align with them as much as possible and we may then also have some feedback into that process.

As FYI @eisenbergeffect this is based off the OpenUI work. FYI to @gregwhitworth that we're officially starting work in FAST on select as part of our partnership 😄

EisenbergEffect:
Awesome. I hadn't had a chance to review yet but just wanted to double check that at the start.

janechu:
You may want to move this to `/specs` since that's typically where they start before implementation.

eljefe223:
> You may want to move this to `/specs` since that's typically where they start before implementation.

@janechu Yeah, I sort of did this out of defiance, Is there a good reason we do it this way? I'm happy to move it, but I it seems weird that we place it in specs only temporarily and then ultimately move it into its own folder.

janechu:
@eljefe223 Well, speaking broadly, I think we wanted to keep it so that specs are find-able for things (not just components) that were being worked on conceptually but did not yet have any code associated with them. When they do have code associated then we would move them to tie them to the code as documentation. If, for example, you wrote a spec and it stayed in some nested folder and it was forgotten, someone might be going through the `/specs` folder, identifying things to contribute and have no idea that in some nested folder somewhere was a spec that could be used for contribution.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+
+### API
+*Component Name*:
+- `fast-option`
+
+*Attributes:*
+- `defaultSelected` - Sets the default of the `<fast-option>` to selected.
+- `disabled` - disables the control
+- `selected` - If the `<fast-option>` is selected or not.
+- `value` - string value of the number field, can be an empty string
```
EisenbergEffect, line 130: Looks like a copy/paste here "number field".

Also, I think it would be valuable for `value` to allowed to be any type when accessed through the property. Attributes have to be strings, but the property doesn't. This would allow templating over lists of complex objects, and associating the complex object as the value if desired.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+*Slots*
+- content - the content of the button
+- start - often times a glyph, icon, or button precedes the content
+- end - often times a glyph, icon, or button follows the content
+
+### Anatomy and Appearance
+**Structure:**
+
+```html
+  <template>
```
EisenbergEffect, line 141: How will hover and selected states be handled in the item? Will that be done with background color, for example? Just wondering how best to handle that if I provide a complex set of markup for the option.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+
+---
+
+### API
+
+*Component Name*:
+- `fast-select`
+
+*Attributes:*
+- `autocomplete` - Allows the developer to provide a hint on how to search the content within the `<fast-option>`(s)
```
EisenbergEffect, line 40: Would be great to see a more through explanation of how this is intended to work.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+- `autofocus` - automatically focuses the control
+- `disabled` - disables the control
+- `length` - The number of `<fast-option>` elements in the `<fast-select>` element
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - the name of the control
+- `options` - Returns a `HTMLOptionsCollection` of the `<fast-option>` elements contained by the `<fast-select>` element
+- `required` - boolean value that sets the field as required
+- `selectedIndex` - The index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
+- `selectedOptions` - An `HTMLCollection` of the selected `<fast-option>` elements
```
EisenbergEffect, line 49: I think we should stick with an array.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+- `fast-select`
+
+*Attributes:*
+- `autocomplete` - Allows the developer to provide a hint on how to search the content within the `<fast-option>`(s)
+- `autofocus` - automatically focuses the control
+- `disabled` - disables the control
+- `length` - The number of `<fast-option>` elements in the `<fast-select>` element
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - the name of the control
+- `options` - Returns a `HTMLOptionsCollection` of the `<fast-option>` elements contained by the `<fast-select>` element
```
EisenbergEffect, line 46: I think we should stick with an array.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+- `required` - boolean value that sets the field as required
+- `selectedIndex` - The index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
+- `selectedOptions` - An `HTMLCollection` of the selected `<fast-option>` elements
+- `size` - If the `<fast-select>` is shown with a scrollbar, this represents how many `<fast-option>`s are visible.
+- `value` - string value of the number field, can be an empty string
```
EisenbergEffect, line 51: Would be nice if this wasn't forced to a string.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+- *A customer using the component on a web page.*
+On a web page a customer is shopping for a new shirt in size medium. The customer opens the size `select` and sees a list of `<option>`s and selects size medium.
+
+### Features
+- form association
+- focus delegation
+
+### Features
+- Select one or multiple `<fast-option/>`(s) using the `multiple` attribute
+- When Dropdown does not have enough screen real estate to open below `<fast-select>` it will open above.
```
scomea, line 22: Do we want more positioning choices (ie. always open above)? Leaving that to later version works too.

scomea: Have you thought about how the flyout menu is placed in the dom, breaks out of parent containers or not, positioning mode?

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+
+**required**: An `<fast-option>` from the `<fast-select>` must be selected when the `required` attribute is set to `true`
+
+**valid**: The `<fast-select>` meets all its validation constraints, and is therefore considered to be valid.
+
+**invalid**: The `<fast-select>` does not meet its validation constraints, and is therefore considered to be invalid.
+
+### Accessibility
+
+*Select* are RTL compliant and support the following aria best practices for listbox [W3C aria-practices](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox)
```
scomea, line 110: What are we using for a "listbox"?  We had that functionality wrapped in a component in the React version, should we do that again?  Listbox behavior shows up in other components like "auto  suggest" so we should be making it sharable somehow.

EisenbergEffect: This is a good point...in fact there is plenty of scenarios where list box is needed on its own and there's a lot in common between the components.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+- `autofocus` - automatically focuses the control
+- `disabled` - disables the control
+- `length` - The number of `<fast-option>` elements in the `<fast-select>` element
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - the name of the control
+- `options` - Returns a `HTMLOptionsCollection` of the `<fast-option>` elements contained by the `<fast-select>` element
+- `required` - boolean value that sets the field as required
+- `selectedIndex` - The index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
+- `selectedOptions` - An `HTMLCollection` of the selected `<fast-option>` elements
+- `size` - If the `<fast-select>` is shown with a scrollbar, this represents how many `<fast-option>`s are visible.
```
scomea, line 50: Not clear on what 'size' does.  Does it control the height of the menu?

gregwhitworth:
> FYI to @gregwhitworth that we're officially starting work in FAST on select as part of our partnership

Can you make sure to file or provide feedback against the Open UI spec as you find gaps and if you have a proposal put it into the issue or file a PR?

specs/select.spec.md
gregwhitworth: It controls how many options are visible within the listbox which does indirectly control the height of the listbox. https://html.spec.whatwg.org/multipage/form-elements.html#attr-select-size

specs/select.spec.md
gregwhitworth: this is currently how `value` works (in the platform) and raises an interesting question with regards to Open UI conformance. If you intend to have value take <n> types rather than following `name=foo[]` it's worth discussing how to handle this divergence.

EisenbergEffect: It's something that I've seen as broken with form controls for a long time. In real apps, we often have to abstract over value just to maintain the type, for example. So, if I've got a list of Person objects and I want to put them in a select, I want the selected value to be the person instance. Folks will often get around this by using the id of an object but then they have to write extra code in their models to do lookups and convert back and forth. Beyond that, not every model has an id of course.

specs/select.spec.md
```diff
@@ -0,0 +1,156 @@
...
+- `value` - string value of the number field, can be an empty string
+
+*Events*
+- `change` - raise the change event for external parties to be informed of the `selectedIndex`'s value change.
```
scomea, line 54: Do we need events around menu open/close?

specs/select.spec.md
EisenbergEffect: For text input, I expect value to be string. For a number input, I expect a number. For a select or listbox, I expect the value to be any object type. For a date picker, I expect a date object, etc.

specs/select.spec.md
eljefe223: D'oh was following an example of a form associated component a bit too closely, ;)

specs/select.spec.md
eljefe223: I'm not sure but open to suggestions here, the change event can be called when the menu opens and when the selection changes, (which would close the menu in single select mode) if we add an attribute `menu-open` change will return the context of the entire component knowing that the menu is open or close and what the selected index is.

specs/select.spec.md
eljefe223: The was the w3c has it in their spec is a `<ul>` with a role of `listbox` which is shown in the template.

image

specs/select.spec.md
eljefe223: Background changes with follow the stealth pattern

specs/select.spec.md
scomea: Also, should the open/close state of the menu be controllable by a developper via an attribute?  You will probably want an easy way to force the menu open during development in any case.

eljefe223: Thats what I suggested above ;)

specs/select.spec.md
eljefe223: I'm open here, but thinking it would just use the intersection observer work you have done.

specs/select.spec.md
```diff
@@ -0,0 +1,158 @@
+# Select + Select-Option
```
awentzel, line 1: I think there should be some documentation around proper usage of a Select inside or outside of a `<form>`

specs/select.spec.md
eljefe223: I was following the Open UI spec, I believe in the past we filtered the potential results, but thinking about this I'm a bit confused on why select would have this. This seems like more of feature of a dropdown on an autosuggest of textbox. @gregwhitworth can you clarify this feature?

specs/select.spec.md
```diff
@@ -0,0 +1,158 @@
...
+- `autofocus` - Automatically focuses the control
+- `disabled` - Disables the control
+- `length` - Number of `<fast-option>` elements in the `<fast-select>` element
+- `menu-open` -If the menu of the `<fast-option>` is open or not
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - Name of the control
+- `options` - Returns an array of the `<fast-option>` elements contained by the `<fast-select>` element
+- `required` - Boolean value that sets the field as required
+- `selectedIndex` - index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
+- `selectedOptions` - An array of the selected `<fast-option>` elements
```
janechu, line 50:
```suggestion
- `selected-options` - An array of the selected `<fast-option>` elements
```

specs/select.spec.md
```diff
@@ -0,0 +1,158 @@
...
+- `autocomplete` - Allows the developer to provide a hint on how to search the content within the `<fast-option>`(s)
+- `autofocus` - Automatically focuses the control
+- `disabled` - Disables the control
+- `length` - Number of `<fast-option>` elements in the `<fast-select>` element
+- `menu-open` -If the menu of the `<fast-option>` is open or not
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - Name of the control
+- `options` - Returns an array of the `<fast-option>` elements contained by the `<fast-select>` element
+- `required` - Boolean value that sets the field as required
+- `selectedIndex` - index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
```
janechu, line 49:
```suggestion
- `selected-index` - index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
```

specs/select.spec.md
```diff
@@ -0,0 +1,158 @@
...
+
+*Attributes:*
+- `autocomplete` - Allows the developer to provide a hint on how to search the content within the `<fast-option>`(s)
+- `autofocus` - Automatically focuses the control
+- `disabled` - Disables the control
+- `length` - Number of `<fast-option>` elements in the `<fast-select>` element
+- `menu-open` -If the menu of the `<fast-option>` is open or not
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - Name of the control
+- `options` - Returns an array of the `<fast-option>` elements contained by the `<fast-select>` element
```
janechu, line 47: Are we expecting this list of options to be updated at any point? If you are the use of index for `selected-index` and `selected-options` may be problematic.

specs/select.spec.md
```diff
@@ -0,0 +1,158 @@
...
+    <fast-option>Option 4</fast-option>
+    <fast-option>Option 5</fast-option>
+</fast-select>
+```
+
+### States
+
+**open**: This state is applied to the `<fast-select>` when the listbox is visible to the user.
+
+**required**: An `<fast-option>` from the `<fast-select>` must be selected when the `required` attribute is set to `true`
```
janechu, line 104: I would add `**disabled**`

specs/select.spec.md
```diff
@@ -0,0 +1,158 @@
...
+## Overview
+
+The `option` component is an option that is meant to be used with `<fast-select>`. `option` renders inside the `<fast-select>` listbox dropdown.
+
+### API
+*Component Name*:
+- `fast-option`
+
+*Attributes:*
+- `defaultSelected` - Sets the default of the `<fast-option>` to selected.
```
janechu, line 129:
```suggestion
- `default-selected` - Sets the default of the `<fast-option>` to selected.
```

eljefe223: Thanks for these suggestions @janechu I agree with the kebab casing, we are aligning with https://open-ui.org/components/select/ We should file a bug on them to update to kebab casing. @gregwhitworth for visibility

janechu: I think it's fine if they have camelCase, we will be using it for our properties, it's just that these are under "Attributes".

eljefe223: They have there's listed under "Attributes" in `code` format, I think theirs should be updated also.

specs/select.spec.md
chrisdholt: We can also hold on adding autocomplete in the initial implementation.

chrisdholt: Actually this may be a good thing to move to "next steps" - investigate autocomplete, etc...

specs/select.spec.md
eljefe223: We use this type of language elsewhere like `active-id` or `selected-index` the are attributes that can be updated in the DOM when a selection change is made. It can be initially set by the app author. Tabs is a goos example of this:

Screen Shot 2020-08-12 at 9 02 00 AM
Screen Shot 2020-08-12 at 9 01 55 AM

specs/select.spec.md
eljefe223: @gregwhitworth and I were having a discussion offline and wanted to include those discussions here. 

We will move this to next steps but wondering if select should have a text input or not, This seems more like a combobox or search box to me. I could not find any documentation online that "searching" options should be part of a select. 

My issue with using the terminology of "search" is that when you search the text you input is not always in the suggested dropdown, which makes it not a select. Because you are not selecting some predetermined value

specs/select.spec.md
scomea: Is this necessary if folks can use css to control listbox dimensions?

specs/select.spec.md
janechu: Sure, the difference is I would expect tabs to come with the tab and tabpanel already included. I might expect a call to a server to populate a list of select options.

chrisdholt:
This will close #3751

EisenbergEffect:
@eljefe223  What do we need to help move this through? stale bot marked it a few days ago.

specs/select.spec.md
eljefe223: We probably should just stick with `slectedOptions` then. @gregwhitworth, thoughts here?

specs/select.spec.md
```diff
@@ -0,0 +1,161 @@
...
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - Name of the control
+- `options` - Returns an array of the `<fast-option>` elements contained by the `<fast-select>` element
+- `required` - Boolean value that sets the field as required
+- `selected-index` - index of the first or last selected `<fast-option>` element, depending on the value of `multiple`
+- `selected-options` - An array of the selected `<fast-option>` elements
+- `value` - The value of select is typed as `any`
+
+*Events*
+- `change` - raise the change event for external parties to be informed of the `selectedIndex`'s value change.
```
scomea, line 52: Do we need an event for menu open/close?

eljefe223: We have an attribute let can open and close, what would the vent give us?

specs/select.spec.md
scomea: Positioning choices (left/right/above/below/responsive/etc...) can be deferred until later.  Breaking out of parent containers is a different and more difficult problem.  Only approach I've come up with would use fixed positioning to position the menu, but the side effect of that is that the open menu no longer tracks the button if the page scrolls.  I suppose we could add that or something like it as a feature later as well.

eljefe223:
There is some discussion around whether this should be a button or text input, talking with @gregwhitworth and @chrisdholt a few weeks ago it was decided to leave out the search functionality to the initial version.

I have a few issues with making select an input:
Does a select component convey that it is able to be used it to search a list of options?
If so, to me select implies that you need to select an option from the list, where search merely makes suggestions but you can "select" or "input" a value that is not in the suggested option dropdown.

scomea (approved):

specs/select.spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+### API
+
+*Component Name*:
+- `fast-select`
+
+*Attributes:*
+- `auto-focus` - Automatically focuses the control
+- `disabled` - Disables the control
+- `menu-open` -If the menu of the `<fast-option>` is open or not
```
EisenbergEffect, line 42: Is this needed as an attribute or is it more a matter of state?

specs/select.spec.md
```diff
@@ -0,0 +1,160 @@
...
+- `multiple` - Allows user to select more than one `<fast-option>`
+- `name` - Name of the control
+- `options` - Returns an array of the `<fast-option>` elements contained by the `<fast-select>` element
+- `required` - Boolean value that sets the field as required
+- `selected-options` - An array of the selected `<fast-option>` elements
```
EisenbergEffect, line 47: Same question here.

specs/select.spec.md
```diff
@@ -0,0 +1,160 @@
...
+- `value` - The value of select is typed as `any`
+
+*Events*
+- `change` - raise the change event for external parties to be informed of the `selected-option`'s value change.
+
+*Slots*
```
EisenbergEffect, line 53: Is this list of slots up to date per the latest open UI spec? I recall there was a listbox slot, for example.

specs/select.spec.md
```diff
@@ -0,0 +1,160 @@
...
+- default - the list of option(s)
+- content - the content of the button
+- start - often times a glyph, icon, or button precedes the content
+- end - often times a glyph, icon, or button follows the content
+- expand-glyph - glyph that indicates the select is expandable, often represented by a downward chevron.
+
+### Anatomy and Appearance
+**Structure:**
```
EisenbergEffect, line 61: Same question on structure as above on slots.

chrisdholt (commented):
@eljefe223 @radium-v is this up to date with the current control in PR?

specs/select.spec.md
chrisdholt: DIdn't see this previously. As select is a form element I think guidance san simply be that we recommend only using it within a form.

radium-v:
Closing and moving to #4194.

Issue #3361, opened 2020-06-24
[Proposal] Provider DSL


EisenbergEffect:
## Problem

Having gone through the process of implementing a significant number of components for our library, writing documentation, and working with a 3rd party to create a tutorial for FAST, a number of "rough edges" with some of our approaches have popped up:

### Using multiple design systems on the same page is challenging

Today, if you import from both `fast-components` and `fast-components-msft` you will experience a runtime error. This is due to the fact that both libraries register web components under the same element names. We did this to enable easily swapping the design system, under the goal of having a single set of components with multiple design systems. However, under the current implementation, the template and styles for a component are still statically linked to the custom element definition (and name) itself, preventing us from having both design systems present.

### Element registration is a side-effect of importing

The underlying call to the `customElements.define` API in the platform is not predictable. It happens as a result of importing a component because the decorator on our components calls the define API during module evaluation. While this is a desirable convenience when building site or app-specific components, it's a problem for customers who want to consume a 3rd party library of components (e.g. fast-components).

### There's no way to disambiguate element definitions

Due to the fact that element registration is a side-effect of importing, there's no opportunity to change the registered element names. This means there's no way to disambiguate our elements from another library that might happen to use the same names. Furthermore, there's no way to disambiguate between multiple design systems built on our same foundation, as described above in the issue on "Using multiple design systems on the same page is challenging".

### We have an awkward need for seemingly inert expressions just to prevent tree shaking

Also due to the fact that element registration is a side-effect of importing, bundlers get confused about what the actual dependencies are and will tree-shake away all our components unless they are referenced directly in code. As a result, we're forced to do things like this:

```ts
import { FASTAnchor } from '@microsoft/fast-components';

FASTAnchor; // very strange!
```

### Our design systems add an additional layer of inheritance just to associate styles

As mentioned above, we currently statically associate styles and templates with the element definition of an element. We're using the `customElement` decorator to do this, which requires a unique class. However, FASTElement has an imperative API that would allow us to make the association without needing the extra layer of inheritance. As a result of our current approach, we also end up with a number of classes in our API docs that have no API. They are just there to inherit from a base class so that styles can be associated.

### There's no way to dynamically change an element's template or styles (or shadow dom mode)

Our API allows us to add or remove additional styles and behaviors, but not change out the base styles dynamically. Templates and styles are statically bound to the component and unchangeable. This prevents several interesting scenarios such as:

* Implementing layouts/master-pages for a router via changes to the router's template.
[...]

PR #3621, opened 2020-08-04
feat(fast-element): move template instance timing and enable dynamic templates and styles

EisenbergEffect:
# Description

This PR moves the timing of the FASTElement template instantiation out of the constructor and into the connected callback. Additionally, new hooks for dynamic template and style selection have been added as well as the capability to hot swap templates and styles at any point for any FASTElement.

## Motivation & context

This PR covers the work in the fast-element phase of #3361 by moving the template creation timing to the connected callback and introducing the `resolveTemplate()` and `resolveStyles()` hooks. Additionally, `template` and `styles` properties have been added to `Controller` so that the template and base styles for any FASTElement can be changed at any time. This enables the implementation of `FoundationElement` as called out in #3361 as well as scenarios around a future router web component, and more dynamic or conditional element rendering. Furthermore, the timing change helps ensure that elements are properly defined time-wise in more complex scenarios, addressing some bugs our partners have been seeing.

While working on this change, some additional refactoring was done around element definitions, in the attempt to consolidate logic and create a better API. A few new APIs were made public or improved on templates and views, as needed for the functionality of dynamic templates and styles.

In addition to these changes, Controller tests have been added to validate its correct function during construction, connection, and operation. Documentation has been updated and extended in several places to help bring further clarity around the lifecycle and new capabilities.

> **Reviewer Request:** I would very much appreciate someone reviewing the fast-components via Storybook to ensure that no unexpected behavior has been introduced as a result of this change.

## Issue type checklist

While this doesn't appear to cause current functionality to break, it's possible that there are some unknown side-effects of the timing change. Additionally, some low-level APIs in fast-element have changes during the refactor. This is not likely to affect anyone though.

## Process & policy checklist


packages/web-components/fast-element/docs/guide/leveraging-css.md
```diff
@@ -164,7 +164,15 @@ const styles = css`
...
 :::
 
-## Hiding undefined elements
+## Styles and the element lifecycle
+
+It is during the `connectedCallback` phase of the Custom Element lifecycle that `FASTElement` adds the element's styles. The styles are only added the first time the element is connected.
+
+In most cases, the styles that `FASTElement` renders are determined by the `styles` property of the Custom Element's configuration. However, you can also implement a method on your Custom Element class named `resolveStyles()` that returns an `ElementStyles` instance. If this method is present, it will be called during `connectedCallback` to obtain the styles to use. This allows the element author to dynamically select completely different styles based on the state of the element at the time of connection.
+
+In addition to dynamic style selection during the `connectedCallback`, the `$fastController` property of `FASTElement` enables dynamically changing the styles at any time through setting the controller's `styles` property to any valid styles.
```
chrisdholt, line 173: This is a great addition

packages/web-components/fast-element/docs/guide/declaring-templates.md
```diff
@@ -190,4 +190,16 @@ Besides rendering content, attributes, and properties, you'll often want to add
...
+
+## Templating and the element lifecycle
+
+It is during the `connectedCallback` phase of the Custom Element lifecycle that `FASTElement` creates templates and binds the resulting view. The creation of the template only occurs the first time the element is connected, while binding happens every time the element is connected (with unbinding happening during the `disconnectedCallback` for symmetry).
+
+:::note
+In the future we're planning new optimizations that will enable us to safely determine when we do not need to unbind/rebind certain views.
+:::
+
+In most cases, the template that `FASTElement` renders is determined by the `template` property of the Custom Element's configuration. However, you can also implement a method on your Custom Element class named `resolveTemplate()` that returns a template instance. If this method is present, it will be called during `connectedCallback` to obtain the template to use. This allows the element author to dynamically select completely different templates based on the state of the element at the time of connection.
```
nicholasrice, line 203: Is it possible to *conditionally* override the `template` property from this method? For instance, does the controller defer to the `template` property template if the returned value from this is null?

packages/web-components/fast-element/src/controller.ts
```diff
@@ -292,6 +320,68 @@ export class Controller extends PropertyChangeNotifier {
...
+
+        if (this.view !== null) {
+            this.view.dispose();
+            (this as Mutable<this>).view = null;
+        } else if (!this.needsInitialization) {
+            host.innerHTML = "";
+        }
+
+        if (template) {
+            // we have a new template to render to the Light
```
nicholasrice, line 380: heh - the Light!

EisenbergEffect: Oops. 🤣  Old comment left around after I refactored. I pushed a commit that adds some better internal comments on what's going on thoughout.

packages/web-components/fast-element/docs/guide/declaring-templates.md
EisenbergEffect: If you return `null` from `resolveTemplate()` then it won't render any template. But you can come back later and set `$fastController.template` and that will trigger a render. Due to the timing of the template location now, you could have your `resolveTemplate()` implementation check the state of its attrs and observables and use that to choose one or another template, because that should all be setup by the time the hook gets called.

Does that answer your question?

packages/web-components/fast-element/src/controller.ts
chrisdholt: I mean, I've seen, written, and found worse comments @EisenbergEffect 😆

packages/web-components/fast-element/docs/guide/declaring-templates.md
nicholasrice: Yep! I am pondering if that capability could be useful but can't think of a solid use-case at the moment

EisenbergEffect: I'm fearful of the internal logic getting a bit complicated and hard to reason about. So, I kept it to a straight priority. First check if the property has been set, then check the custom hook, then check the static definition. I think any advanced scenario can be handled by having code set the property as needed. This also allows the property to be surfaced on FoundationElement as an observable and be bound on a per instance basis as well, which would enable per element template or style overrides with very little effort.

packages/web-components/fast-element/src/controller.ts
```diff
@@ -292,6 +320,83 @@ export class Controller extends PropertyChangeNotifier {
...
+        // we're doing a Light DOM render to the element's direct children.
+        const host = getShadowRoot(element) || element;
+
+        if (this.view !== null) {
+            // If there's already a view, we need to unbind and remove through dispose.
+            this.view.dispose();
+            (this as Mutable<this>).view = null;
+        } else if (!this.needsInitialization) {
+            // If there was previous custom rendering, we need to clear out the host.
+            host.innerHTML = "";
```
janechu, line 391: Might be better perf (see https://jsperf.com/innerhtml-vs-removechild/418):
```suggestion
            for (let child = host.firstChild; child !== null; child = host.firstChild) {
                host.removeChild(child);
            }
```

EisenbergEffect: Added a utility to do this and then refactored. Thanks!

PR #3627, opened 2020-08-05
feat: add fast-breadcrumb and fast-breadcrumb-item web components

khamudom:
# Description
Breadcrumb examples:
image

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,40 @@
...
+        <slot>
+            <a
+                class="control"
+                part="control"
+                href="${x => x.href}"
+                aria-current="${x => (x.isCurrent ? "page" : void 0)}"
+            >
+                <span class="content" part="content">
+                    ${x => x.name}
+                </span>
```
chrisdholt, line 22: What's the purpose of having this span here?

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    .control .content::before {
+        content: "";
+        display: block;
+        height: calc(var(--outline-width) * 1px);
+        position: absolute;
+        top: calc(1em + 4px);
+        width: 100%;
+    }
```
chrisdholt, line 60: What is this doing and what's the purpose of the absolute positioning with the before pseudo selector on the span inside the anchor? Does that span give us something?

khamudom: This is creating the underline when hovering over the anchor.
image

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
khamudom: In this case, to style the underline of the anchor. Similar to how our fast-anchor is structured.

chrisdholt:
@khamudom I'll see if I can pull this down and find what's causing the issues you are running into.

packages/web-components/fast-components/src/breadcrumb-item/README.md
```diff
@@ -0,0 +1,4 @@
+# fast-breadcrumb-item
+*BreadcrumbItem*
```
chrisdholt, line 2: Should this be here?

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.stories.ts
```diff
@@ -0,0 +1,13 @@
+import { FASTDesignSystemProvider } from "../design-system-provider";
+import BreadcrumbItemTemplate from "./fixtures/base.html";
+import { FASTBreadcrumbItem } from ".";
+
+// Prevent tree-shaking
+FASTBreadcrumbItem;
+FASTDesignSystemProvider;
+
+export default {
+    title: "BreadcrumbItem",
```
chrisdholt, line 10: Breadcrumb Item

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,54 @@
...
+                        href="${x => x.href}"
+                        aria-current="${x => (x.isCurrent ? "page" : void 0)}"
+                    >
+                        <span class="content" part="content">
+                            ${x => x.name}
+                        </span>
+                    </a>
+                `
+            )}
+        </slot>
```
chrisdholt, line 38: Do we need both of these? It seems like we could essentially just have a single default slot with default content inside it. When you pass other content, it'll replace the default content (control) with the slotted content.

I might be missing something, but I don't see why we couldn't drop lines 13-21 and remove the name of the slot on line 22.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,54 @@
...
+        ${startTemplate}
+        <slot
+            ${slotted({
+                property: "defaultSlottedNodes",
+                filter: value =>
+                    (value.nodeType === 3 && value.textContent!.trim().length !== 0) ||
+                    value.nodeType === 1 ||
+                    false,
+            })}
+        ></slot>
```
chrisdholt, line 21: Might be helpful to have some kind of detail as to why we need this. It seems related to the default slotted content below, but almost as though it's a bit of trickery. I noticed that when I removed this and just went down to a single slot, the control "broke" when passing start/end content. I assume this is why this is here. If that is the case, I think it would be helpful to add a bit of a note in the API docs or somewhere as to what is happening here. Without it, I think it'll cause confusion.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,54 @@
...
+        <slot name="control">
+            ${when(
+                x => x.defaultSlottedNodes.length === 0,
+                html<BreadcrumbItem>`
+                    <a
+                        class="control"
+                        part="control"
+                        href="${x => x.href}"
+                        aria-current="${x => (x.isCurrent ? "page" : void 0)}"
+                    >
```
chrisdholt, line 31: What happens if I want to pass additional attributes here? Should they pass to the host or to this? Should we delegate focus here?

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
```diff
@@ -0,0 +1,114 @@
...
+        white-space: nowrap;
+    }
+
+    .control:hover {
+        color: ${accentForegroundHoverBehavior.var};
+    }
+
+    .control:active {
+        color: ${accentForegroundActiveBehavior.var};
+    }
```
chrisdholt, line 54: I'm noticing a couple of things:

1. Anchors without an href are being styled as anchors - we aren't going to want that. That will be an issue overall.
2. We likely want the item with `aria-current` to be styled differently than the other items. 
> The aria-current attribute is used when an element within a set of related elements is visually styled to indicate it is the current item in the set. For example:
>
>A page token used to indicate a link within a set of pagination links, where the link is visually styled to represent the currently-displayed page.

It's possible that the styles for no href and aria-current are similar; though I would expect no href to not have hover styles (it should be treated as text), and the aria-current would include hover styles matching it's visual style.

packages/web-components/fast-components/src/breadcrumb-item/README.md
khamudom: Updated with a description.
"The `fast-breadcrumb-item` component is meant to be used with [breadcrumb](../breadcrumb/README.md)."

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
khamudom: yeah, delegate focus should be here. Will add that.

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
khamudom: I will check with design on this, but here is an update where the last item has no href and no aria-current. It is style to be a text.
image

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
khamudom: Yeah, the `defaultSlottedNodes` slot, is to use as the default slot. The slot below used to be the default slot, but in order to get both the name attribute and SVG to render, I had to give it a slot name and create another slot used as a default.

Just to give some background context.
The structure I had before this was,
```ts
<slot>
            <a
                class="control"
                part="control"
                href="${x => x.href}"
                aria-current="${x => (x.isCurrent ? "page" : void 0)}"
            >
                ${startTemplate}
                <span class="content" part="content">
                    ${x => x.name}
                </span>
                ${endTemplate}
            </a>
</slot>
```
For some reason, every time I add a `start` or an `end slot, the `name` attribute disappears. I found that if the HTML is formatted this way,
```html
<fast-breadcrumb-item name="Home" href="#">
   <svg slot="start">
       <path ... />
   </svg>
</fast-breadcrumb-item>
```
Nothing gets rendered, but if I format the HTML this way,
```html
<fast-breadcrumb-item name="Home" href="#"><svg slot="start">
    <path ... />
</svg></fast-breadcrumb-item>
```
then I see the icon and the name attribute. Noticed that both the component tag and the SVG tag need to be on the same line.
I am still figuring out why putting both tags on the same line is working. One theory is that in the breadcrumb item template, the anchor is inside of the default <slot>. That means that any nodes, including spaces, replace the content.
Hoping if anyone has seen this before and why having the tags on the same line is working.

khamudom: added `defaultSlottedNodes ` to the spec

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
khamudom: added `delegatesFocus` and aria attributes to the anchor

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
khamudom: worked with @bheston on design.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
chrisdholt: I think we'll just want to document this in code as well. It's not apparent until you try to remove it why it's there.

khamudom: After more investigation and understanding why this is needed and how we can avoid having to do this, I have updated the template to use the `when` directive. If the breadcrumb-item has an href it will be anchor.

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
```diff
@@ -0,0 +1,125 @@
...
+    accentForegroundHoverBehavior,
+    accentForegroundRestBehavior,
+    neutralForegroundRestBehavior,
+    heightNumber,
+} from "../styles/index";
+import { SystemColors } from "@microsoft/fast-web-utilities";
+
+export const BreadcrumbItemStyles = css`
+    ${display("inline-flex")} :host {
+        background: transparent;
```
bheston, line 18: Is this not transparent by default?

packages/web-components/fast-components/src/breadcrumb/breadcrumbs.styles.ts
```diff
@@ -0,0 +1,11 @@
+import { css } from "@microsoft/fast-element";
+import { display } from "@microsoft/fast-foundation";
+
+export const BreadcrumbStyles = css`
+    ${display("inline-block")} :host {
+        box-sizing: border-box;
+        font-family: var(--body-font);
+        font-size: var(--type-ramp-minus-1-font-size);
```
bheston, line 8: I don't think this needs to be smaller. I suggest using the base size here.

packages/web-components/fast-foundation/src/breadcrumb/README.md
```diff
@@ -0,0 +1,35 @@
+---
+id: breadcrumb
+title: fast-breadcrumb
+sidebar_label: breadcrumb
+custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/src/breadcrumb/README.md
+---
+
+A `fast-breadcrumb` widget is used as a navigational aid, allowing users to maintain awareness of their locations within a program, app, or a website.
```
bheston, line 8: Could we avoid "widget"? Prefer "component" if necessary, but since the whole package is components we can often write the sentence without mentioning it.

packages/web-components/fast-components/src/breadcrumb/breadcrumbs.stories.ts
```diff
@@ -0,0 +1,15 @@
+import { FASTDesignSystemProvider } from "../design-system-provider";
```
chrisdholt, line 1: Should this file be named `breadcrumb.stories.ts` or `breadcrumbs.stories.ts`? Same for the others which are inconsistent.

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.template.ts
```diff
@@ -0,0 +1,20 @@
...
+ * The template for the {@link @microsoft/fast-foundation#Breadcrumb} component.
+ * @public
+ */
+export const BreadcrumbTemplate = html<Breadcrumb>`
+    <template
+        role="navigation"
+        aria-label="${x => x.ariaLabel}"
+        aria-labelledby="${x => x.ariaLabelledby}"
+    >
+        <div role="list">
```
chrisdholt, line 14: This is missing a part and class. If you place a class on this and set the value to `display: flex` that will fix your problem when one (but not all) items has start or end content.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
```diff
@@ -0,0 +1,65 @@
...
+                    aria-owns="${x => x.ariaOwns}"
+                    aria-relevant="${x => x.ariaRelevant}"
+                    aria-roledescription="${x => x.ariaRoledescription}"
+                >
+                    ${startTemplate}
+                    <span class="content" part="content">
+                        <slot></slot>
+                    </span>
+                    ${endTemplate}
+                </a>
```
chrisdholt, line 45: @eisenbergeffect is there a way we can reference the `fast-anchor` template here instead of duplicating it?

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,57 @@
...
+export class DelegatesARIABreadcrumbItem extends ARIAGlobalStatesAndProperties {
+    /**
+     * See {@link https://www.w3.org/WAI/PF/aria/roles#link} for more information
+     * @public
+     * @remarks
+     * HTML Attribute: aria-expanded
+     */
+    @attr({ attribute: "aria-expanded", mode: "fromView" })
+    public ariaExpanded: "true" | "false" | undefined;
+}
```
chrisdholt, line 47: Isn't this just a link? Should we just use the existing link definition? There's not an "ARIA breadcrumb item", I think this should just use the anchor class.

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
```diff
@@ -0,0 +1,124 @@
...
+        display: flex;
+        align-items: center;
+    }
+
+    .separator {
+        margin: 0 6px;
+    }
+
+    .control {
+        align-items: center;
```
chrisdholt, line 38: Perhaps it's the SVG's, but when I pulled this down (with the start/end fix mentioned here for "list" in breadcrumb) these don't look centered. Can you validate that start/end are actually being truly centered? Again, perhaps it's just visual with the SVG's.

chrisdholt (changes requested):
A couple more questions and a few things to address, including a fix for the issue with only one item having start/end content.

packages/web-components/fast-components/src/breadcrumb/breadcrumbs.stories.ts
khamudom: I think breadcrumbs.

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.template.ts
khamudom: thank you!

packages/web-components/fast-components/src/breadcrumb/breadcrumbs.stories.ts
chrisdholt: Let’s make sure all files in the folder are named consistently.

chrisdholt: FWIW, the ARIA spec refers to “Breadcrumb”.

chrisdholt: We should strive to align in the direction that Open-UI is most likely to go. I’ll reach out to a couple folks.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
khamudom: If we can use the `fast-anchor`, I would be happy with that.

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
khamudom: Yeah, it's the SVG, especially the "Home" SVG. I did some testing in `fast-anchor` and adding the same text and SVG, and it looks the same, which is slightly off-center. I think what is happening, is the SVG is aligning to the full height of the text and the underline. When the anchor is not being hovered, it looks off, but when you hovering on the anchor and the underline appears, it looks centered.
no_underline
with_underline

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.ts
khamudom: yeah, I didn't think this was needed. will remove.

packages/web-components/fast-components/src/breadcrumb/breadcrumbs.stories.ts
chrisdholt: 7/9 on OpenUI seem to be named "Breadcrumb" are there others which are named "Breacrumbs" besides the two outliers?

chrisdholt: And to be clear - talking about the file names; the control appears to be named "Breadcrumb"

khamudom: I have only seen those two

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,48 @@
...
+    @observable
+    public isCurrent: boolean = false;
+}
+
+/**
+ * Includes ARIA states and properties relating to the ARIA link role
+ *
+ * @public
+ */
+export class DelegatesARIABreadcrumbItem extends ARIAGlobalStatesAndProperties {}
```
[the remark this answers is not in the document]
khamudom: updated to extend `DelegatesARIALink`

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.template.ts
EisenbergEffect: If it's identical, you should be ale to compose it by just doing this

```
${AnchorTemplate}
```

khamudom: It looks like we will not be able to use `fast-anchor` there are a few attributes inside `fast-anchor` that is different in `breadcrumb-item` anchor. For example, breadcrumb has aria-current set to an observable created for `breadcrumb-item`
`aria-current="${x => (x.isCurrent ? "page" : void 0)}"`

chrisdholt: @khamudom - why not map to the aria property? The following is part of aria-globals which gets mixed in:
```
    /**
     * Indicates the element that represents the current item within a container or set of related elements.
     *
     * {@link https://www.w3.org/TR/wai-aria-1.1/#aria-current}
     * @public
     * @remarks
     * HTML Attribute: aria-current
     */
    @attr({ attribute: "aria-current", mode: "fromView" })
    public ariaCurrent:
        | "page"
        | "step"
        | "location"
        | "date"
        | "time"
        | "true"
        | "false"
        | string;

```

chrisdholt: It seems like the template should be fine unless we're trying to get fancy :)

khamudom: yeah, everything works great. just need to figure out how to change `aria-current`. Will give your suggestion a go.

chrisdholt: You shouldn't need to do anything but remove `isCurrent` from the class and from the template.

nicholasrice:
Converting this to a draft because it is a draft PR.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.ts
```diff
@@ -0,0 +1,27 @@
...
+ * A Breadcrumb Item Custom HTML Element.
+ *
+ * @public
+ */
+export class BreadcrumbItem extends Anchor {
+    /**
+     * @internal
+     */
+    @observable
+    public showSeparator: boolean = true;
```
chrisdholt, line 16: @khamudom a bit of a nit, but in preference of brevity here, could `separator` work just as well here? Technically, this renders the separator - so it's less about conditionally showing this and more about whether it is present (still true/false).

khamudom: updated to `separator`

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+
+        await disconnect();
+    });
+
+    it("should NOT set the `target` attribute on the internal anchor if `href` is not passed", async () => {
+        const { element, connect, disconnect } = await setup();
+        const targetExample = "_blank";
+
+        element.target = targetExample;
+
```
marjonlynch, line 107: I don't understand the need for this test, you already have a test for no internal anchor if href is not passed, in this case there shouldn't be an internal anchor (just like the previous test) let alone a target on that non-existent anchor. The reason I would suggest culling tests that don't prove anything new is around maintenance cost. perhaps this is a nit, but my suggestion is to take a look at the rest of these and see if some of them are testing stuff that you've already tested or you could tweak another test to accommodate all of one testable concept.

packages/web-components/fast-foundation/src/breadcrumb/README.md
```diff
@@ -0,0 +1,35 @@
+---
+id: breadcrumb
+title: fast-breadcrumb
+sidebar_label: breadcrumb
+custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/src/breadcrumb/README.md
+---
+
+`fast-breadcrumb` is used as a navigational aid, allowing users to maintain awareness of their locations within a program, app, or a website.
```
marjonlynch, line 8: route or path instead of locations? or maybe "navigated path"

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Use Cases
+- James is browsing an ecommerce website and is looking for a few items to add to his new kitchen. He needs a knife set, and an espresso machine. He checks out `Home and Kitchen` page, then selects `Kitchen and Dining`, first category he picks is `Cutlery`, he finds the sets of knives. There is now a breadcrumb trail of the pages James has clicked through. `Home and Kitchen /  Kitchen and Dining / Cutlery / Knife Sets`
+He clicks on `Kitchen and Dining` to go back to the category of kitchen items, selects `Coffee and Espresso` finds the best one that suits his needs. If James wants to continue looking for another item for his kitchen, or other things for his home, he just needs to click on any of the links in the breadcrumb trail to continue his search.
+
+### Prior Art/Examples
+- [Ant Design](https://ant.design/components/breadcrumb/)
+- [Atlaskit](https://atlaskit.atlassian.com/packages/core/breadcrumbs)
+- [Carbon Design](https://www.carbondesignsystem.com/components/breadcrumb/code/)
+- [FAST Breadcrumb (React)](https://www.npmjs.com/package/@microsoft/fast-components-react-msft)
```
marjonlynch, line 14: I'm not sure this is the right path anymore, this link takes you to an npmjs page that says the package has been deprecated and the documentation link on that page doesn't work. I'm not sure what the right link would be though.

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+async function setup() {
+    const { element, connect, disconnect } = await fixture<FASTBreadcrumb>(
+        "fast-breadcrumb"
+    );
+
+    return { element, connect, disconnect };
+}
+
+describe("Breadcrumb", () => {
```
marjonlynch, line 20: Do you have any tests with regard to breadcrumb's relationship with breadcrumbitems? like stuff around the slot for the items and what if they put something incorrectly in that slot and so on?

packages/web-components/fast-components/src/breadcrumb/fixtures/base.html
```diff
@@ -0,0 +1,178 @@
+<fast-design-system-provider use-defaults>
+    <h1>Breadcrumb</h1>
```
marjonlynch, line 2: Do you have a scenarios folder and index.html file? I think this changed just yesterday, here's an example for radio-group:
Screen Shot 2020-10-16 at 10 12 21 AM

packages/web-components/fast-components/src/breadcrumb/README.md
```diff
@@ -0,0 +1,4 @@
+# fast-breadcrumb
+*Breadcrumb* component is used as a navigational aid, allowing users to maintain awareness of their locations within a program, an app, or a website.
```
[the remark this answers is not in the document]
khamudom: updating to `fast-breadcrumb`

packages/web-components/fast-components/src/breadcrumb/breadcrumb.open-ui.definition.json
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "Breadcrumb",
```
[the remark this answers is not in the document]
khamudom: we are not using FAST prefix for any of the other components in the open-ui.definition file

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.spec.ts
```diff
@@ -0,0 +1,585 @@
...
+    });
+
+    it("should set the `href` attribute on the internal anchor equal to the value provided", async () => {
+        const { element, connect, disconnect } = await setup();
+        const hrefExample: string = "https://fast.design";
+
+        element.href = hrefExample;
+
+        await connect();
+
```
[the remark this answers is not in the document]
chrisdholt: I'd agree with this so long as it pertains to a given test. We should be careful to not duplicate tests which already exist. Similarly, we should be careful to have tests which span across different scenarios.

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
```diff
@@ -0,0 +1,124 @@
...
+        background: none;
+    }
+
+    .start,
+    .end,
+    ::slotted(svg) {
+        ${
+            /* Glyph size and margin-left is temporary - 
+            replace when adaptive typography is figured out */ ""
+        } width: 16px;
```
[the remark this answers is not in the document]
khamudom: it looks like we don't. all components are patterned this way for `::slotted(svg)`

packages/web-components/fast-components/src/breadcrumb-item/README.md
```diff
@@ -0,0 +1,4 @@
+# fast-breadcrumb-item
+The `fast-breadcrumb-item` component is meant to be used with [breadcrumb](../breadcrumb/README.md).
```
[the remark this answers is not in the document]
khamudom: updated to `fast-breadcrumb`

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.open-ui.definition.json
```diff
@@ -0,0 +1,4 @@
+{
+    "name": "Breadcrumb-item",
```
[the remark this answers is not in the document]
khamudom: no FAST prefix is used in this file

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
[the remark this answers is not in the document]
chrisdholt: There's an entire v-team doing density work and there are new typography stories underway by @bheston that will likely affect this. Add in that I think we want to drop this requirement to support 20x20 icons and I think we need an issue to track any changes to this pattern across the repo.

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.md
chrisdholt: I mean, it is deprecated and technically correct in terms of pointing to this. If you want a new link to the code: https://github.com/microsoft/fast-react/tree/master/packages/react/fast-components-react-msft/src/breadcrumb

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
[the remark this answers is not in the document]
khamudom: I will sync with Brian and create an issue for updating

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.md
khamudom: thanks. I will use the new link

packages/web-components/fast-components/src/breadcrumb/fixtures/base.html
khamudom: thank you. I added a scenario file.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.spec.ts
khamudom: I understand. I am keeping the test to set `target` on the anchor if href is set.

packages/web-components/fast-foundation/src/breadcrumb-item/breadcrumb-item.spec.ts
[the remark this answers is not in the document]
khamudom: The test here sets an href inside the anchor. The test up on line 35, renders and anchor, only when an href is passed.

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.ts
khamudom: added one for the separator

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+            <fast-breadcrumb>
+                <fast-breadcrumb-item>Item 1</fast-breadcrumb-item>
+                <fast-breadcrumb-item>Item 2</fast-breadcrumb-item>
+                <fast-breadcrumb-item>Item 3</fast-breadcrumb-item>
+            </fast-breadcrumb>
+        `);
+
+        await connect();
+
+        // expect(element?.shadowRoot?.querySelector("")).to.not.equal(null);
```
marjonlynch, line 54: Remove the comment?

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export class Breadcrumb extends FASTElement {
+    /**
+     * @internal
+     */
+    @observable
+    public slottedBreadcrumbItems: HTMLElement[];
+    public slottedBreadcrumbItemsChanged() {
+        if (this.$fastController.isConnected) {
+            const lastNode: HTMLElement = this.slottedBreadcrumbItems[
```
marjonlynch, line 17: can slottedBreadcrumItems be empty? might not want to do some of this code if so

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,49 @@
...
+            if (lastNode.childElementCount > 0) {
+                lastNode.childNodes.forEach((item: HTMLElement) => {
+                    if (item instanceof HTMLElement && item.hasAttribute("href")) {
+                        childNodeWithHref = item;
+                    }
+                });
+            }
+
+            /**
+             *  If child node with href is found then apply aria-current to child node otherwise apply aria-current to the host element, with an href
```
marjonlynch, line 36: I would do this whole chunk around setting aria-current in a function and call it here, maybe even break the part about finding a child node with an href into its own function as well

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,49 @@
+import { FASTElement, observable, attr } from "@microsoft/fast-element";
```
marjonlynch, line 1: is `attr` still used?

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.spec.ts
khamudom: missed that. thanks!

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
khamudom: removed

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
khamudom: created a checked to see slottedBreadcrumbItems is undefined

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
khamudom: updated to break up the function

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @internal
+     * Finds href on childnodes in the light DOM or shadow DOM.
+     * We look in the shadow DOM because we insert an anchor when breadcrumb-item has an href.
+     */
+    private findChildWithHref(node: HTMLElement): HTMLElement | null {
+        let findNodeWithHref: HTMLElement | null = null;
+
+        if (node.childElementCount > 0) {
```
marjonlynch, line 47: I'm not normally a fan of multiple return statements but in this case, since this is intended to be a small simple function, maybe just return the values inline from the querySelector calls and get rid of the variable findNodeWithHref.

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -0,0 +1,71 @@
...
+            findNodeWithHref = node.querySelector("a[href]");
+        } else if (node.shadowRoot?.childElementCount) {
+            findNodeWithHref = node.shadowRoot?.querySelector("a[href]");
+        }
+        return findNodeWithHref;
+    }
+
+    /**
+     *  If child node with href is found then apply aria-current to child node otherwise apply aria-current to the host element, with an href
```
marjonlynch, line 56: It may be obvious but I would add child node with anchor tag that has its href attribute in the comment here

marjonlynch (approved):

EisenbergEffect (approved):

chrisdholt (approved):
Can you add a check mark to the spec readme detailing the component as done as well?

EisenbergEffect:
Woohoo!

Issue #3634, opened 2020-08-05
Width is not being respected in Fast-Slider


barahonajm:
**Describe the bug; what happened?**

When you set the `<fast-slider>` width to 0% it is not being respected since it already has a min-width set to:
```css
:host(.horizontal) {
    min-width: calc(var(--design-unit) * 60px);
}
```

**What are the steps to reproduce the issue?**

Use the following code:

```html
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
  <title>Repro</title>
</head>

<body>
  <fast-design-system-provider use-defaults>
    <fast-slider min="0" max="100" value="1" style="width:0%"></fast-slider>
  </fast-design-system-provider>

  
  <script type="module" src="https://unpkg.com/@microsoft/fast-components"></script>
</body>

</html>

```

**What behavior did you expect?**

The width must be respected when set to 0%,

barahonajm:
I think the problem here is the set default value:
```javascript
private setupDefaultValue = (): void => {
    if (this.value === "") {
        this.value = `${this.convertToConstrainedValue((this.max + this.min) / 2)}`;
    }
};
```

This is setting the `dirtyValue = true` so when we try to assign the a new value to `value` it gets ignored, so if you have this:

```html

  <fast-slider id="slider" min="0" max="100" style="width:0%"></fast-slider>

  <script>
      window.onload = function() {
        let slider = document.getElementById("slider");
        slider.setAttribute("value", "90");
      };
  </script>

```

the `slider.setAttribute("value", "90")` will not work at all unless you have defined the `value` attribute to any value in your `fast-slide`.

I think the `setupDefaultValue` should set `dirtyvalue = false` after finishing

barahonajm:
This issue was already fixed for both fluent and fast, further discussion will continue on #3644

PR #3640, opened 2020-08-06
fix: update min-width of fast-slider to design-unit * 1px

barahonajm:
# Description

As explained on #3634, the min-width default value (design-unit * 60px) is not a number that give enough flexibility for a default instance of the slider, preventing us to change the width from 100% to 0% unless override the default of min-width

## Motivation & context

#3634 Is the cause of this pull request, and as discussed on Discord the value is being changed to `design-unit * 1px`, but I would like to have more discussion around it.

## Findings

While doing the change I have discovered the same issue will happen with a `fast-slider` in a vertical mode, and I believe the fix here will not be as simple as setting it to `design-unit * 1px` because doing that the slider will normally appear with a height of `4px` in case of fast design system provider, on Firefox when you don't have set the height manually it appears as 160px but the `min-height` remains as auto 

I have been playing with the slider and discovered a few things that can help to take a decision:

- For the normal `input range` both in vertical and horizontal the min-width and min-height are set to `auto` on Firefox, and the min-width to `0` on Edge 
- The width and height are who have a computed value which is dynamically set based on the context
  - Width is set to an arbitrary value (129px on my case on edge) when the input element is added into a div or directly into the body
  - Width is set to 100% when the parent is `display: grid`

Possibly the most straightforward solution will be to leave everything as it is right now and document that min-width and min-height are set to `design-unit * 60px`, any thoughts? 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/slider/slider.styles.ts
+++ b/packages/web-components/fast-components/src/slider/slider.styles.ts
@@ -82,7 +82,7 @@ export const SliderStyles = css`
         transform: translateY(calc(var(--thumb-translate) * 1px));
     }
     :host(.horizontal) {
-        min-width: calc(var(--design-unit) * 1px);
+        min-width: calc(var(--thumb-size) * 1px);
     }
     :host(.horizontal) .track {
         right: calc(var(--track-overhang) * 1px);
--- a/packages/web-components/fast-components/src/slider/slider.styles.ts
+++ b/packages/web-components/fast-components/src/slider/slider.styles.ts
@@ -82,7 +82,7 @@ export const SliderStyles = css`
         transform: translateY(calc(var(--thumb-translate) * 1px));
     }
     :host(.horizontal) {
-        min-width: calc(var(--design-unit) * 60px);
+        min-width: calc(var(--design-unit) * 1px);
     }
     :host(.horizontal) .track {
         right: calc(var(--track-overhang) * 1px);


barahonajm:
> Thanks! Would you want to make a similar fix to the FluentUI repo where our fluent components live? No pressure 😁!

😁 Sure, but there are some comments in the description you might want to check before that, related with the vertical orientation for the slider which happen to have the same issue

packages/web-components/fast-components/src/slider/slider.styles.ts
```diff
@@ -82,7 +82,7 @@ export const SliderStyles = css`
         transform: translateY(calc(var(--thumb-translate) * 1px));
     }
     :host(.horizontal) {
-        min-width: calc(var(--design-unit) * 60px);
+        min-width: calc(var(--design-unit) * 1px);
```
marjonlynch, line 85: This works, or rather doesn't break any of our fixtures, but I'm wondering now does it make sense to set the min-width to something smaller than what the thumb is going to be set to? which is calc(var(--thumb-size) * 1px), typically I think this is 16px, for DesignSystemDefaults at least I think.

barahonajm: I am totally agreed so the code for `min-width` will be:

```suggestion
        min-width: calc(var(--thumb-size) * 1px);
```

right? 

Also, as I mentioned in the **findings** section of the pull request what can we do with the vertical one?

marjonlynch: I read your findings there about the vertical part, if you could I think it would be good to create an issue for that and maybe link back to this one, and we can look at it again.

barahonajm: Good idea, I am goint to do that thanks.

marjonlynch: that looks good to me

chrisdholt:
Thanks @barahonajm for this fix! Feel free to submit the same over in the Fluent repo if you like at your convenience 😄

PR #3649, opened 2020-08-07
fix: use higher bits for DOM marker

### LINKED ISSUE #3648: Marker length in fast-element
The way the marker is currently generated, it may contain less random characters than intended. It is possible (although somewhat unlikely) for the marker to contain 0 random characters.

This code will find a marker with only two random characters. It finds one instantly on my machine. Markers with one random character are also near instant.

```js
let marker;
do { 
  marker = Math.random().toString(36).substring(7);
} while (marker.length > 2);
console.log(marker);
```

The marker should contain sufficient random characters to avoid collision.

ruphin:
# Description

Ensure the random marker (almost) always has the same amount of random characters. This fixes a bug where in edge cases, the marker contains less characters than expected.

Fixes #3648

## Motivation & context

The `Number.prototype.toString()` function does not have a fixed length and does not emit a `0` character at the end of decimal strings. That means that `.subString(7)` will emit less characters if the random number happens to end with a lot of `0` bits. Because of this it is possible to end up with an empty marker. This is an example that will quickly emit a 2 character random string:

```js
let marker;
do { 
  marker = Math.random().toString(36).substring(7);
} while (marker.length > 2);
console.log(marker);
```

By using the higher order characters from the random decimal string instead, the chance of characters being dropped is significantly smaller, and the random string will nearly always be 6 characters. 

A simple visual of which characters are used to create the marker string before and after this change:

```
old:        "3y7tzi"
     "0.2ouzh3y7tzi"
new:   "2ouzh3"
```

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
Nice catch @ruphin ! Thank you for the contribution. We'll get this merged soon.

EisenbergEffect:
@awentzel FYI Looks like we're still having some issues with CI on community PRs.

EisenbergEffect:
Thanks @ruphin ! I'm very happy to have someone else digging into the internals of fast-element. I'd love to see what other improvements you come up with. Glad to have your contribution 🥇

PR #3687, opened 2020-08-11
fix: hitting enter key when in the text-field did not submit form

### LINKED ISSUE #3661: bug: fast-text-field does not submit form on enter
When using a `fast-text-field` inside of a `form` element, pressing the `enter` key should cause the form's `submit` event to fire, just as with a normal text `input` element. Presently, nothing happens when the enter key is pressed.

marjonlynch:
# Description
* enable form submit on pressing enter key in text-field, other events input and change works as before
closes #3661 

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/text-field/fixtures/text-field.html
```diff
@@ -97,4 +97,10 @@ <h4>Visual vs audio label</h4>
     <!-- With aria-label -->
     <h4>With aria-label</h4>
     <fast-text-field aria-label="Text field with aria-label"></fast-text-field>
+
+    <form onsubmit="console.log('form submitted!');">
```
marjonlynch, line 101: not sure if I should leave this in the fixture or not

chrisdholt: @marjonlynch - if we need a method of testing, I'd recommend using an `alert()` rather than console.log in this case so there is a visible indication.

marjonlynch: made it an alert instead of console.log

PR #3691, opened 2020-08-11
feat: add flyout wc spec

SethDonohue:
# Description

Adds Flyout Web Component Spec

## Motivation & context

FAST needs an Popover component as the Dialog does not currently cover cases where positioning is needed of a focusable floating container to display UI related to what the user is doing.

## Issue type checklist

## Process & policy checklist


specs/flyout.md
```diff
@@ -0,0 +1,152 @@
...
+![Actionable More Info](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/images/flyout-example2.png)
+
+---
+
+### API
+*Component Name*
+- `fast-flyout`
+
+*Attributes*
+- `anchor` - string, the html id of the HTMLElement that the Flyout is attached to
```
scomea, line 43: How do you select the viewport the flyout is positioned relative to?  Document root, the parent of the overlay?

specs/flyout.md
```diff
@@ -0,0 +1,152 @@
...
+- `overlay` - boolean, specifies whether an overlay is drawn to dim the out of scope UI until the Flyout is dismissed
+- `light-dismiss` - boolean, specifies whether or not the flyout can be dismissed by clicking elsewhere or pressing `esc` key, defaults to true
+- `visible` - boolean, whether or not the Flyout is visible, defaults to undefined, as this is controlled by the author
+- `responsive` - boolean, whether or not the positioning is responsive based on available space, defaults to true, uses Anchored Region logic
```
scomea, line 47: How does "responsive" interact with "position"?  Does it affect both axis?  Is a single bool enough to cover the "responsive" behavior we want?  For example "I want my flyout to be above or below the anchor" and "I want my flyout to be left or right of the anchor" seem like reasonable choices.  Not sure how much complexity do we want here but we probably need more than a single bool?  Note that in tooltip we currently default to a "above or below based on available space" default placement if no fixed position is provided.  We could do something similar here, skip the responsive prop and avoid having to define a that part of the api now and just pick the default behavior.

specs/flyout.md
```diff
@@ -0,0 +1,152 @@
...
+    - left-centered
+    - top-left-corner
+    - top-right-corner
+    - bottom-right-corner
+    - bottom-left-corner
+    - top-left-aligned
+    - right-top-aligned
+    - bottom-left-aligned
+    - left-top-aligned
+        - **ASK (1)**
```
scomea, line 61: top-right-aligned, bottom-right-aligned and others may be missing or I may not be clear on what "aligned" means?  I'm thinking it controls which edge the flyout abuts to and the direction it grows?

specs/flyout.md
```diff
@@ -0,0 +1,152 @@
...
+
+*Properties*
+- `anchorElement` - the reference to the HTMLElement used as an anchor
+- `visible` - boolean, whether or not the Flyout is visible
+*Methods*
+- `isNonFocusableContent()` - check if the slotted content is a focusable element or not, could use a filter to check the node type.
```
scomea, line 67: What's the use case for this?  Flyout could have complex content.

specs/flyout.md
scomea: Looks like the parent of the flyout.

specs/flyout.md
```diff
@@ -0,0 +1,152 @@
...
+- `content` - the region where the content is rendered, has `role=dialog`
+
+
+*Template:*
+**ASK (2)**
+```html
+${when(
+    x => x.visible,
+    html<Flyout>`
+        <div>
```
scomea, line 88: You might not need this wrapper div?

specs/flyout.md
SethDonohue: I see what you're saying. I was going for a more general position change. `responsive` set to true would allow the flyout to change it's location if the location the author specified doesn't have available space. Having `responsive` set to false would NOT allow the flyout to move if there is not enough space in the position the author has specified.
I could see some authors NOT wanting the flyout to move at all based on available space and want to give them the option to specify if it is "fixed" vs "responsive" with regards to available space. Does that make sense with Anchored Region? I was under the impression that we could setup anchored region so that it does not change its location based on available space.

specs/flyout.md
SethDonohue: Yeah I was going to go with the "default" viewport that Anchored Region would get, which is the parent of the Flyout, correct?

specs/flyout.md
SethDonohue: I should make a diagram for this...
I was thinking `top` meant the general location of the flyout is *above* its anchor element, and `right-aligned` meant the *left* edge of the flyout would be aligned with the *right* edge of the anchor. That does seem confusing as I write it out here. I'll try to clarify this, and think of better wording.

specs/flyout.md
SethDonohue: Yes, but if Flyout does NOT have complex/focusable content we need to ensure that the flyout is still accessible by making the flyout itself focusable.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+### API
+*Component Name*
+- `fast-flyout`
+
+*Attributes*
+- `anchor` - string, the html id of the HTMLElement that the Flyout is attached to, (viewport for anchored region defaults to the parent of the anchor)
+- `overlay` - boolean, specifies whether an overlay is drawn to dim the out of scope UI until the Flyout is dismissed
+- `light-dismiss` - boolean, specifies whether or not the flyout can be dismissed by clicking elsewhere or pressing `esc` key, defaults to true
```
nicholasrice, line 45: I sometimes see this called "soft dismiss" - did you happen to look at other libraries to see what is more common?

chrisdholt: I think there's been some discussion or usage of terminology over in Open UI. Might be good to dig into issues and PR's to see how they've referred to it. In Dialog, we emit an event regardless in the case of light dismiss rather than having at attribute here. Would that be preferred here as well? I think the benefit is that if you want to do something on dismiss you listen for the event; if you don't want to - you don't. Thoughts on creating alignment between these two?

specs/flyout/flyout.md
```diff
@@ -0,0 +1,168 @@
...
+- As a popup showing additional info for an item on the page.
+- To display a larger image of a thumbnail image in a photo gallery.
+- To display an action related to a page item
+
+### Features
+**Placement Logic:** - This component will use *Anchored Region* to be positioned relative to the Flyouts anchor element. This way it can be dynamic and adapt to the available space if `responsive` is true for the Flyout.
+
+**Light Dismiss (aka Soft Dismiss):** - A way to quickly close the Flyout by clicking outside it, or pressing the `esc` key.
+
+**Overlay:** - A control of whether or not the Flyout has an overlay effect that dims the surrounding UI when the Flyout is active.
```
chrisdholt, line 21: How common is this with other instances of flyout-like controls across libraries? This seems more common with a dialog pattern, whereas things like flyout, callout, coachmarks seem to rarely have overlays. If this is common, can you link to some examples here?

chrisdholt:
This will close #3752

specs/flyout/flyout.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Prior Art/Examples
+##### Context Menu
+*Note:* the "Manage, Apply, Show more options" is the Flyout for this example.
+![Context Menu](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/images/flyout-nested.png)
```
chrisdholt, line 26: I actually don't think this would be a flyout here - I think this would be menu/menu item which already exists. Perhaps it might be better to list these examples by Design system rather than "type". I would expect that the base flyout is primarily going to be similar to dialog or where it is a container with semantics and behaviors. It would likely be helpful to have additional examples from other design systems as currently it seems like we only have UWP examples. An example from Fluent might be their "callout" or "teaching bubble". Other design systems may call this "popover". I think having other examples will help here.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,168 @@
...
+![Actionable More Info](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/images/flyout-example2.png)
+
+---
+
+### API
+*Component Name*
+- `fast-flyout`
+
+*Attributes*
+- `anchor` - string, the html id of the HTMLElement that the Flyout is attached to, (viewport for anchored region defaults to the parent of the anchor)
```
chrisdholt, line 43: I know that in "anchored region" the attribute is "anchor", but I wonder if that is too specific to a *type* of element. These are often triggered by elements with a role of button. I'm not sure I would expect "button" to be a replacement, but something more broad such as "trigger" seems more reasonable. What do other design systems do for this?

specs/flyout/flyout.md
```diff
@@ -0,0 +1,168 @@
...
+
+##### Actionable
+*Note:* the button is not part of the Flyout, but shown for context as an anchor element.
+![Actionable More Info](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/images/flyout-example2.png)
+
+---
+
+### API
+*Component Name*
+- `fast-flyout`
```
chrisdholt, line 40: There seem to be quite a few names related to this type of control; while I've referred to it as "flyout" it seems like a more common term is "popover". If we're ultimately hoping to be aligned to open ui, we should probably look for a name that has the best path forward for alignment.

specs/flyout.md
scomea: Why not do this with the position attribute?  If an author sets the position to "top" it goes on top. For the first release if position is "undefined" we do "above or below based on space"?  That leaves us free to evolve a more complex placement system later.

specs/flyout.md
scomea: Is there an easy/reliable way to detect that with complex flyout content?  What if it changes during the lifetime of the flyout?  Wondering if this is a decision we could let the author make.

chrisdholt: This should be the same behavior as Dialog I believe; do we need a method for this? If we need to set and trap focus for modal instances we'll be querying already for this. If there are no focusable elements I'd assume we'd set focus here. Not sure we need a method for authors to call here.

scomea: Could we not let authors decide this somehow rather than trying to detect it?

chrisdholt: This should likely have a role of "dialog" so I assume we'll want to handle this in the same way - I'm not sure why we would treat it differently.

specs/flyout/flyout.md
amis92: Just browsing around for new features, but I have an example I find quite common - it'd be for tutorial/intro/demo interactions where a user is guided to perform actions, sometimes not allowed to proceed without completing required action.

SethDonohue (commented):
I'll change the name of the spec file to popover once it is complete so that we don't lose track of the comments.

specs/README.md
```diff
@@ -19,6 +19,7 @@ Here you'll find specifications for custom elements and other library features.
 | Combobox | -- | -- |
 | [Dialog](../packages/web-components/fast-foundation/src/dialog/dialog.spec.md) | :white_check_mark: | :white_check_mark: |
 | [Disclosure](../packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md) | :white_check_mark: | -- |
+| [Flyout](./flyout.md) | :white_check_mark: | -- |
```
SethDonohue, line 22: I'll change the name of the file before merge to not outdate all the comments due to a file name change.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,168 @@
...
+
+*Attributes*
+- `anchor` - string, the html id of the HTMLElement that the Flyout is attached to, (viewport for anchored region defaults to the parent of the anchor)
+- `overlay` - boolean, specifies whether an overlay is drawn to dim the out of scope UI until the Flyout is dismissed
+- `light-dismiss` - boolean, specifies whether or not the flyout can be dismissed by clicking elsewhere or pressing `esc` key, defaults to true
+- `visible` - boolean, whether or not the Flyout is visible, defaults to undefined, as this is controlled by the author
+- `responsive` - boolean, whether or not the positioning is responsive based on available space, defaults to true, uses Anchored Region logic
```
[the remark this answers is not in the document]
scomea: Works for me.  We should try to have similar "position selection" options for this, tooltip and other components that have this kind of behavior.

specs/flyout/flyout.md
SethDonohue: Seems like target or trigger is used in others. I think target is good, but this will mean that we won't be using common names across our own components as tooltip uses _*anchor*_

SethDonohue: @chrisdholt I also think it may be beneficial to keep it as anchor as we may be able to reuse the tooltip control and only change a few things to get a popover to work. The content and trigger method will be different, as well as the role, but the control will be very similar.

SethDonohue:
@chrisdholt @scomea Do we want to have this just be a slot and have no default layout of a popover, or would having areas defined like a header, footer, and close button most likely be too prescriptive? It may be best to be like the Dialog where we don't supply any layout _inside_ of the popover.

scomea:
@SethDonohue Personally I'd rather see the base component should avoid defining internals as much as possible.  Folks with specific needs can derive specialized versions from that.

SethDonohue:
> @SethDonohue Personally I'd rather see the base component avoid defining internals as much as possible. Folks with specific needs can derive specialized versions from that.

I agree with this direction and assume it'd be the best, but figured it be good to clarify it.

SethDonohue:
> @chrisdholt @scomea Do we want to have this just be a slot and have no default layout of a popover, or would having areas defined like a header, footer, and close button most likely be too prescriptive? It may be best to be like the Dialog where we don't supply any layout _inside_ of the popover.

Related to the above; Focus management seems to be fairly consistent with Dialog across most popovers, but not all use it the same or require it.
Since we are setting the **role=dialog** I believe we should be following the same focus management as Dialog so that it follows the accessibility pattern expected for a Dialog.
This means we will have focus entering the Popover once it is opened and focus will be trapped inside the Popover until the Dialog is closed with:
- escape key
- by clicking an author included close/action button
- a click outside of the Popover.

Forcing focus inside a popover when we don't know of the content that will be in there, since we may not prescribe a close button ourselves, could be a problem.
**What do we focus on if there is *no focusable content (buttons, actions, etc)* inside the Popover?**

SethDonohue:
I just added some updates to focus management and a salesforce example.
The name of this component is still up in the air between flyout and popover depending on feedback here, and what further research finds.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
+# Popover (aka Flyout)
+## Overview
+Popover is a focusable floating container that shows over a pages content and displays UI related to what the user is doing and is typically triggered by a clickable element. Similar to a Dialog, but with positioning controls, no overlay option, and no specific action is required, a Popover can be used to reveal a secondary control or show more detail about an item.
```
litteredwitherrors, line 3:
```suggestion
Popover is a focusable floating container that shows over a page's content and displays UI related to what the user is doing and is typically triggered by a clickable element. Similar to a Dialog, but with positioning controls, no overlay option, and no specific action is required. A Popover can be used to reveal a secondary control or show more detail about an item.
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Background
+A Popover can be less intrusive to the user experience than a Dialog and should be used instead of a Dialog when information or input is needed to be displayed, but not in the center of the screen, such as in the case of collection of additional info before an action, displaying information that is only relevant some of the time, or just displaying more information.
```
litteredwitherrors, line 6:
```suggestion
A Popover can be less intrusive to the user experience than a Dialog and should be used instead of a Dialog when information or input is needed to be displayed, but not in the center of the screen, such as in the case of the collection of additional info before an action, displaying information that is only relevant some of the time or for just displaying more information.
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Use Cases
+- As a popup showing additional info for an item on the page.
+- To display a larger image of a thumbnail image in a photo gallery.
+- To display an action related to a page item.
+
+### Features
+**Placement Logic:** - This component will use *Anchored Region* to be positioned relative to the Popover anchor element. This way it can be dynamic and adapt to the available space.
+
+**Soft Dismiss (aka Light Dismiss):** - A way to quickly close the Popover by clicking outside it, or pressing the `esc` key. An event would be emitted anytime the Popover is dismissed. The Popover would always be soft-dismissable. Detecting the users next click of a different element can trigger this when clicking outside the Popover.
```
litteredwitherrors, line 16:
```suggestion
**Soft Dismiss (aka Light Dismiss):** - A way to quickly close the Popover by clicking outside it, or pressing the `esc` key. An event would be emitted anytime the Popover is dismissed. The Popover would always be soft-dismissable. Detecting the user's next click of a different element can trigger this when clicking outside the Popover.
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+
+**Focus Management:** - Should follow the same process for focus management as a Dialog primarily as the Popover's role will be dialog. This includes a trap focus method, which can be optional. For accessibility the focus should return to the target that triggers the popover after the popover is closed.
```
litteredwitherrors, line 18:
```suggestion
**Focus Management:** - Should follow the same process for focus management as a Dialog primarily as the Popover's role will be dialog. This includes a trap focus method, which can be optional. For accessibility, the focus should return to the target that triggers the popover after the popover is closed.
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+### Prior Art/Examples
+
+##### Popover from Salesforce
+- https://www.lightningdesignsystem.com/components/popovers/
+
+![Popover](./images/salesforce-popover.png)
+##### Popover from Evergreen
+- triggered by a button
```
litteredwitherrors, line 28:
```suggestion
- Triggered by a button
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+- https://evergreen.segment.com/components/popover/
+
+![Popover](./images/evergreen-popover.png)
+
+##### Microsoft Windows Developer Flyout, Text Only
+*Note:* the button is not part of the Popover, but shown for context as an anchor element.
```
litteredwitherrors, line 34:
```suggestion
*Note:* The button is not part of the Popover, but shown for context as an anchor element.
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+![Text Only](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/images/flyout-wrapping-text.png)
+
+##### Microsoft Windows Developer Flyout, Actionable
+*Note:* the button is not part of the Popover, but shown for context as an anchor element.
```
litteredwitherrors, line 38:
```suggestion
*Note:* The button is not part of the Popover, but shown for context as an anchor element.
```

litteredwitherrors (changes requested):
Overall, I'd suggest going through and adding some consistency to where you capitalize after bullet-points and hyphens.

SethDonohue (commented):
> Overall, I'd suggest going through and adding some consistency to where you capitalize after bullet-points and hyphens.

I added your suggestions as a commit. Good call on the consistency

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+    <fast-button>Action</fast-button>
+</fast-popover>
+```
+
+### States
+`visible` - Whether or not Popover is hidden. This state is managed solely by the app author via the visible attribute.
+
+### Accessibility
+*Keyboard Navigation and Focus*
+Keyboard and navigation will follow the same rules as Dialog per the [W3C Specification](https://w3c.github.io/aria-practices/#dialog_modal), expect when there is no focusable content then the Popover itself will receive focus.
```
litteredwitherrors, line 163:
```suggestion
Keyboard and navigation will follow the same rules as Dialog per the [W3C Specification](https://w3c.github.io/aria-practices/#dialog_modal), except when there is no focusable content then the Popover itself will receive focus.
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Globalization
+The component visuals should change when in RTL scenarios as the component is positioned relative to it's anchor.
```
litteredwitherrors, line 166:
```suggestion
The component visuals should change when in RTL scenarios as the component is positioned relative to its anchor.
```

litteredwitherrors (approved):
Just a couple of grammatical things, otherwise, looks good.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+**Placement Logic:** - This component will use *Anchored Region* to be positioned relative to the Popover target element. This way it can be dynamic and adapt to the available space.
+
+**Soft Dismiss (aka Light Dismiss):** - A way to quickly close the Popover by clicking outside it, or pressing the `esc` key. An event would be emitted anytime the Popover is dismissed. The Popover would always be soft-dismissable. Detecting the users next click of a different element can trigger this when clicking outside the Popover.
+
+**Focus Management:** - Should follow the same process for focus management as a Dialog, primarily as the Popover's role will be 'dialog'.
+
+This includes:
+- Trap focus - optional, defaults to enabled - Focus is contained within the popover while it is visible.
+- First element focus - Focusing on the first focusable element when trap focus is enabled, but not overriding an element that is set to be focused by the author.
+- Returning focus to the target - optional, defaults to enabled, after popover is closed the focus would return to the target element.
```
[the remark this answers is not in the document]
nicholasrice: Does the flyout close automatically on scroll?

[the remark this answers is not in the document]
SethDonohue: No it does not. I never saw that behavior when looking at other flyout and popover components.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
+# Popover (aka Flyout)
+## Overview
+Popover is a focusable floating container that shows over a pages content and displays UI related to what the user is doing. It is typically triggered by a clickable element. Similar to a Dialog, but with positioning controls, no overlay option, and no specific action being required, a Popover can be used to reveal a secondary control or show more detail about an item.
```
radium-v, line 3: > with positioning controls

Are the positioning controls part of the UI or are they part of the API?

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+### API
+*Component Name*
+- `fast-popover`
+
+*Attributes*
+- `visible` - boolean, whether or not the Popover is visible, defaults to undefined, as this is controlled by the author
+- `trap-focus` - boolean, whether or not to keep focus contained inside the popover, defaults to true
+- `target` - string, the html id of the HTMLElement that the Popover is attached to, (viewport for anchored region defaults to the parent of the anchor) and triggered by
+- `focus-target-on-close` - boolean, whether or not to focus back on to the _target element_ when the popover is closed, defaults to true
+- `position` - enum, where the tooltip should appear relative to its target, uses Anchored Region logic=
```
radium-v, line 65: This should describe all valid values for `position`, or link to the Anchored Region documentation for `position` (or both).

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+- `delay` - number, the time delay before the popover is shown once triggered
```
radium-v, line 66: What unit of time is the delay value?

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+- `aria-labelledby` - optional based on implementation**
+- `aria-describedby` - optional based on implementation**
+- `aria-label` - optional, based on implementation**
+
+    ** See the [W3C Specification](https://w3c.github.io/aria-practices/#dialog_roles_states_props) for requirements and details.
+
+*Properties*
+- `targetElement` - the reference to the HTMLElement used as an target
```
radium-v, line 74:
```suggestion
- `targetElement` - the reference to the HTMLElement used as a target
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+- `viewportElement` - the reference to the HTMLElement used as the viewport, required ot setup anchored region properly
```
radium-v, line 75:
```suggestion
- `viewportElement` - the reference to the HTMLElement used as the viewport, required to set up anchored region properly
```

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+- `popoverVisible` - boolean, whether or not the Popover is visible
```
radium-v, line 76: How about `isVisible` or `open`?

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+
+*Slots*
+- default for content
+
+*Events*
+- `dismiss` - event fired when the Popover is dismissed
+
+*Functions*
+- `createPopoverTemplate(string: prefix)` - generates a `ViewTemplate` for the popover based on the given provided prefix string. This is required as popover uses an `anchored-region` internally and the create function generates a template using the appropriate `anchored-region` tag (ie "fast-anchored-region", "fluent-anchored-region"). Note that the appropriate `anchored-region` component must also be declared.
```
radium-v, line 85: `createTemplate` would make it easier to rename this component if we decide to go with something other than `popover`.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+*Template:*
+**ASK (2)**
+```html
+${when(
+    x => x.visible,
+    html<Popover>`
+        <div>
+            <fast-anchored-region
```
radium-v, line 130: I ran into this issue when building `fast-select` - I initially used other fast components in the template, which creates a weird circular dependency where we need fast-components to build fast-foundation, which is needed to build fast-components. This also means that the `fast-` tag "namespace" will need to be permanent, which could create conflicts for authors who want to compose their own components off of `fast-foundation`.

specs/flyout/flyout.md
SethDonohue: They are part of the API

specs/README.md
SethDonohue: Flyout may make this component more distinguishable from future popup/pop related components, so may still be the best option to make it distinct.

specs/flyout/flyout.md
chrisdholt: +1 - Should we support the same positioning values as tooltip? Less? More?

specs/flyout/flyout.md
SethDonohue: Ahh yes, this will be resolved with a recent change similar to Tooltip where a prefix is given to createPopoverTemplate, but i failed to include that here in this example. Updated now.

specs/flyout/flyout.md
SethDonohue: I have updated with a note to see the _Position Options_ section of this spec, as it has several options available.

specs/flyout/flyout.md
SethDonohue: In keeping with how Tooltip has this, I think staying with popoverVisible helps clearly separate it from the `visible` attr.

specs/flyout/flyout.md
SethDonohue: We currently have 2 other create***Template functions that are public and exported so I think we need to have Popover in there for clarification.

specs/flyout/flyout.md
SethDonohue: updated to flyoutVisible

specs/flyout/flyout.md
SethDonohue: updated

specs/flyout/flyout.md
SethDonohue: updated

specs/flyout/flyout.md
chrisdholt: Is there a visible attr? If so, there is a visible JavaScript property. What would the difference be between a "visible" attribute and this property?

specs/flyout/flyout.md
SethDonohue: ms, updated

specs/flyout/flyout.md
```diff
@@ -0,0 +1,181 @@
...
+
+### API
+*Component Name*
+- `fast-flyout`
+
+*Attributes*
+- `visible` - boolean, whether or not the flyout is visible, defaults to undefined, as this is controlled by the author
+- `trap-focus` - boolean, whether or not to keep focus contained inside the Flyout, defaults to true
+- `target` - string, the html id of the HTMLElement that the Flyout is attached to, (viewport for anchored region defaults to the parent of the anchor) and triggered by
+- `focus-target-on-close` - boolean, whether or not to focus back on to the _target element_ when the Flyout is closed, defaults to true
```
SethDonohue, line 64: This is definitely open to other naming ideas or approaches to solving this problem, which is described in the focus management section comments

specs/flyout/flyout.md
```diff
@@ -0,0 +1,181 @@
...
+**Placement Logic:** - This component will use *Anchored Region* to be positioned relative to the Flyout target element. This way it can be dynamic and adapt to the available space.
+
+**Soft Dismiss (aka Light Dismiss):** - A way to quickly close the Flyout by clicking outside it, or pressing the `esc` key. An event would be emitted anytime the Flyout is dismissed. The Flyout would always be soft-dismissable. Detecting the users next click of a different element can trigger this when clicking outside the Flyout.
+
+**Focus Management:** - Should follow the same process for focus management as a Dialog, primarily as the Flyout's role will be 'dialog'.
+
+This includes:
+- Trap focus - optional, defaults to enabled - Focus is contained within the Flyout while it is visible.
+- First element focus - Focusing on the first focusable element when trap focus is enabled, but not overriding an element that is set to be focused by the author.
+- Returning focus to the target - optional, defaults to enabled, after Flyout is closed the focus would return to the target element.
```
SethDonohue, line 23: recapturing original comments:
https://github.com/microsoft/fast/pull/3691#discussion_r554124652

specs/flyout/flyout.md
```diff
@@ -0,0 +1,181 @@
...
+### API
+*Component Name*
+- `fast-flyout`
+
+*Attributes*
+- `visible` - boolean, whether or not the flyout is visible, defaults to undefined, as this is controlled by the author
+- `trap-focus` - boolean, whether or not to keep focus contained inside the Flyout, defaults to true
+- `target` - string, the html id of the HTMLElement that the Flyout is attached to, (viewport for anchored region defaults to the parent of the anchor) and triggered by
+- `focus-target-on-close` - boolean, whether or not to focus back on to the _target element_ when the Flyout is closed, defaults to true
+- `position` - enum, where the tooltip should appear relative to its target, uses Anchored Region logic, see *Position Options*
```
SethDonohue, line 65: recapturing original comment: 
https://github.com/microsoft/fast/pull/3691#discussion_r561230159

specs/flyout/flyout.md
SethDonohue: After re-reading the spec, it is expected that the target that triggers a dialog is focused on once the dialog is closed and since we have set the role of flyout to dialog we should follow that. It is currently defaulted to do so, but should we allow something like `focus-taget-on-close` or not even make that optional?

radium-v: Maybe intercepting the `dismiss` event to redirect focus somewhere else (or prevent the default focusing behavior) would be the best bet, instead of having an attribute for the behavior?

chrisdholt: Yeah, I'm not sure that this is valuable here - are there other places where we handle this? Lots of other controls require focus to be placed back onto the invoking element, but I can't think of a place where we are currently handling that with the primitives themselves. Unless we have a more universal story or principle around handling this, I would probably drop this.

SethDonohue: Yeah, even though it is disruptive to refocus after scrolling, this is still a dialog so it should follow that design pattern.

specs/flyout/flyout.md
SethDonohue: This was changed to target.

awentzel:
@SethDonohue what is the status for this PR?  Is there still work to do here? If so, could we outline that to help facilitate pushing this to completion.  Thanks

SethDonohue:
> @SethDonohue what is the status for this PR? Is there still work to do here? If so, could we outline that to help facilitate pushing this to completion. Thanks

I have asked for plenty of final reviews on this, but I am not sure where the priority is at this point for this component.

@chrisdholt Are we still pushing forward with a flyout component? If so we can get this spec merged with reviews, and then I can update the actual component's PR with the most recent changes to anchored region and work with Stephane if I have any difficulties.

chrisdholt:
> > @SethDonohue what is the status for this PR? Is there still work to do here? If so, could we outline that to help facilitate pushing this to completion. Thanks
> 
> I have asked for plenty of final reviews on this, but I am not sure where the priority is at this point for this component.
> 
> @chrisdholt Are we still pushing forward with a flyout component? I assume so.
> I can get this spec merged once I get some final reviews, and then I can update the actual component's PR with the most recent changes to anchored region and work with Stephane if I have any difficulties.

As with everything else, once this spec is approved by @microsoft/fast-collaborators it can be merged in.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,180 @@
+# Flyout (aka Popover)
```
chrisdholt, line 1: We should drop "popover" for now as I believe we've moved away from that name.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Features
+**Placement Logic:** - This component will use *Anchored Region* to be positioned relative to the Flyout target element. This way it can be dynamic and adapt to the available space.
+
+**Soft Dismiss (aka Light Dismiss):** - A way to quickly close the Flyout by clicking outside it, or pressing the `esc` key. An event would be emitted anytime the Flyout is dismissed. The Flyout would always be soft-dismissable. Detecting the users next click of a different element can trigger this when clicking outside the Flyout.
+
+**Focus Management:** - Should follow the same process for focus management as a Dialog, primarily as the Flyout's role will be 'dialog'.
+
+This includes:
+- Trap focus - optional, defaults to enabled - Focus is contained within the Flyout while it is visible.
```
chrisdholt, line 21: We should add this to the risks section - we're currently re-evaluating how we do this, but trapping focus w/r/t the Shadow DOM can be quite problematic: https://nolanlawson.com/2021/02/13/managing-focus-in-the-shadow-dom/

SethDonohue: Added!

specs/flyout/flyout.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+---
+
+### API
+*Component Name*
+- `fast-flyout`
+
+*Attributes*
+- `visible` - boolean, toggle controlled by the author of whether or not the flyout is visible, defaults to undefined
```
nicholasrice, line 65: Did we consider logically flipping this attribute and using the platform's 'hidden' attribute instead?

nicholasrice: I see that Dialog uses 'hidden' and not 'visible', I'd recommend using 'hidden' here both for platform and FAST component alignment.

nicholasrice (approved):
Looks good, just one change I think we should make. Thanks @SethDonohue!

specs/flyout/flyout.md
SethDonohue: Godo call

specs/flyout/flyout.md
```diff
@@ -0,0 +1,184 @@
...
+- `delay` - number, milliseconds (ms) the time delay before the Flyout is shown once triggered
+- `aria-labelledby` - optional based on implementation**
+- `aria-describedby` - optional based on implementation**
+- `aria-label` - optional, based on implementation**
+
+    ** See the [W3C Specification](https://w3c.github.io/aria-practices/#dialog_roles_states_props) for requirements and details.
+
+*Properties*
+- `targetElement` - the reference to the HTMLElement used as target
+- `viewportElement` - the reference to the HTMLElement used as the viewport, required to set up anchored region properly
```
scomea, line 78: This defaults to the document if unset in anchored region, so viewport can be left unset.  "target" is required.   You could probably have an attribute for viewport as well as target.

specs/flyout/flyout.md
```diff
@@ -0,0 +1,184 @@
...
+*Slots*
+- default for content
+
+*Events*
+- `dismiss` - event fired when the Flyout is dismissed
+
+*Functions*
+- `createFlyoutTemplate(string: prefix)` - generates a `ViewTemplate` for the Flyout based on the given provided prefix string. This is required as Flyout uses an `anchored-region` internally and the create function generates a template using the appropriate `anchored-region` tag (ie "fast-anchored-region", "fluent-anchored-region"). Note that the appropriate `anchored-region` component must also be declared.
+
+*Position Options*
```
scomea, line 90: I'm not clear on the api for positioning.  What choices does the author have for this?   I'd also suggest that authors have the option to either force a position (locktodefault) or be based on available space.

scomea (changes requested):

specs/flyout/flyout.md
scomea: Anchored region can do "absolute" or "fixed" positioning.  I think we can elaborate on whether we want to present that choice in the api after we get an initial version in, but what do you see the default being?  "fixed" breaks out of parent containers more easily but absolute doesn't detach from the anchor as often.

scomea:
Anchored region can do "absolute" or "fixed" positioning. I think we can elaborate on whether we want to present that choice in the api after we get an initial version in, but what do you see the default being? "fixed" breaks out of parent containers more easily but absolute doesn't detach from the anchor as often.

specs/flyout/flyout.md
SethDonohue: Ok I'll swap the required statements.

specs/flyout/flyout.md
SethDonohue: I updated the names, does this make more sense? These would be the options available.

I had a `responsive` attribute early on in this spec that would determine if the flyout was forced or would move based on available space, but removed it per some other comments.

I think having something like `responsive` (default to true) which means that the flyout would move along the same axis based upon available space is a great idea. If the flyout `position` option was "right" and `responsive` was true by default then when there was not enough space on the right of the target then the flyout would render on the left of the target. Does that make sense?

SethDonohue:
> Anchored region can do "absolute" or "fixed" positioning. I think we can elaborate on whether we want to present that choice in the api after we get an initial version in, but what do you see the default being? "fixed" breaks out of parent containers more easily but absolute doesn't detach from the anchor as often.

I am still not clear on the difference here. What do you think would be better to start with?
I fee like absolute is a good start because a flyout generally is expected to stay next to it's target. Like a tooltip in that way. Thoughts?

SethDonohue:
> > Anchored region can do "absolute" or "fixed" positioning. I think we can elaborate on whether we want to present that choice in the api after we get an initial version in, but what do you see the default being? "fixed" breaks out of parent containers more easily but absolute doesn't detach from the anchor as often.
> 
> I am still not totally clear on the difference here. What do you think would be better to start with?
> I fee like absolute is a good start because a flyout generally is expected to stay next to it's target. Like a tooltip in that way. Thoughts?

After our discussion elsewhere let's go with fixed for now, since it will break out of its parent containers more easily and act more like a document level popup/flyout.

specs/flyout/flyout.md
SethDonohue: I rebased which outdated these comments :/ ... I have updated though so let me know what you think!

specs/flyout/flyout.md
scomea: I'm thinking we should simplify the number of positioning options available.   What if the choices were simply 'top', 'bottom', 'left' and right.  If an author chooses responsive their choice simply flips to the other side based on their choice (ie. left goes to right, top goes to bottom, etc...)?  My thinking is that our base components hava a simple positioning api, and if there is demand we can think about having a "pro" interface where advanced authors could pass a full "anchored region config" object and do all the things.

SethDonohue:
> > > Anchored region can do "absolute" or "fixed" positioning. I think we can elaborate on whether we want to present that choice in the api after we get an initial version in, but what do you see the default being? "fixed" breaks out of parent containers more easily but absolute doesn't detach from the anchor as often.
> > 
> > 
> > I am still not totally clear on the difference here. What do you think would be better to start with?
> > I fee like absolute is a good start because a flyout generally is expected to stay next to it's target. Like a tooltip in that way. Thoughts?
> 
> After our discussion elsewhere let's go with fixed for now, since it will break out of its parent containers more easily and act more like a document level popup/flyout.

@scomea does this satisfy your question and hold on this PR?

scomea:
@SethDonohue Yeah, the only thing I'm not clear about is how a user might configure the component to position itself based on available space (ie. below, but if if doesn't fit above)

SethDonohue:
> @SethDonohue Yeah, the only thing I'm not clear about is how a user might configure the component to position itself based on available space (ie. below, but if if doesn't fit above)

@scomea I actually forgot I already added a `responsive` attribute back in to the *Attributes* section:
`- 'responsive' - boolean, whether or not the positioning is responsive based on available space, defaults to true`

To me this says that unless they set `responsive` to false, the flyout will update it's position based on available space, but that does not mean the user can chose where it goes based on available space, i.e. to the bottom only instead of left or right as well. Is that something we can do with anchored region?

Currently if I set `position` to top, and have `responsive` to true (default) then if there is not enough space on top it will move down, but it could *also* move left or right depending on available space.

EisenbergEffect:
This PR is very old now. What is our plan? We should close, move to draft, or get this in soon.

EisenbergEffect:
One year old this week. /cc @chrisdholt

chrisdholt:
> One year old this week. /cc @chrisdholt

This should be updated to align and leverage the Popover specification from Open-UI: https://github.com/openui/open-ui/pull/355

Lots of research has now been gathered there and there is an editors draft available. We should update our spec accordingly from a naming standpoint as well as look to align the API to the current guidance there. At that point I think we can move ahead with implementation. Ideally, we should ensure we spell out in the spec that our preference is to use `Popover` when available in browser and fall back to our code implementation when not. 

@SethDonohue do you want to rework this based on the above or close and we can look to kick off a new one given the deltas?

SethDonohue:
Let's have a new one kicked off when ready!

PR #3694, opened 2020-08-12
doc(webpack): add viewport meta and babelrc config

ramonck:
- Added the view-port to make it mobile friendly in the tutorial.
- Added additional instructions regarding the decorator tag error on webpack compilation.

# Description

## Motivation & context
I tested the tutorial on Windows 10 with WSL2 and Visual Studio Code.

Help other people to start off.

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
@ramonck Can you provide a little more information around the babelrc file addition? I've been working with this setup for some time and haven't noticed any errors related to this. Just trying to understand a bit better and see if I missed something somewhere. Thanks!

ramonck:
> @ramonck Can you provide a little more information around the babelrc file addition? I've been working with this setup for some time and haven't noticed any errors related to this. Just trying to understand a bit better and see if I missed something somewhere. Thanks!

@EisenbergEffect is basically when you do a free start in visual studio without anything on it really, no reactjs or babelrc setup if someone just followed the instructions without doing those pre-setups they enconter this babelrc issue when using the notations.
You can replicate this yourself on a Windows 10 setup with VSCode and you'll see the issue.
image

packages/web-components/fast-foundation/docs/integrations/webpack.md
```diff
@@ -217,6 +218,13 @@ This code imports the `<fast-design-system-provider>` component as well as the `
 </body>
 ```
 
+Create a `.babelrc` in your project root and put the following content to mitigate decorator tag error on compilation within webpack.
+```json
```
nicholasrice, line 222: The preceding instructions don't include installing babel or the plugin - should those dependencies be added to the install list?

ramonck: > The preceding instructions don't include installing babel or the plugin - should those dependencies be added to the install list?

@nicholasrice this is something that I putted in the comments that currently is not included in the package.json do you also want this there due to the annotations or should this be optional? Babel is probably already embedded inside fast packaging the issue is with the annotation plugin required when using annotations is annotations always optional in fast? Then there should be manual instructions for installing it if not if there's always some level of annotation then it should be looked into including it within the project's package.json file.

EisenbergEffect: I admit that I'm still not following entirely what causes this issue. This isn't a tutorial for setting up Babel for FAST, but rather TypeScript. We aren't using Babel at all. Is this something that's somehow triggered only on Windows in some strange way? I see no issue at all on Mac with this basic setup, for example. I can't reproduce it.

ramonck: > I admit that I'm still not following entirely what causes this issue. This isn't a tutorial for setting up Babel for FAST, but rather TypeScript. We aren't using Babel at all. Is this something that's somehow triggered only on Windows in some strange way? I see no issue at all on Mac with this basic setup, for example. I can't reproduce it.

@EisenbergEffect have you tried using this type of code in your demo setup without issues in compile or run?
**@customElement('name-tag')**
https://www.fast.design/docs/fast-element/defining-elements
```
import { FASTElement, customElement } from '@microsoft/fast-element';

@customElement('name-tag')
export class NameTag extends FASTElement {

}
````

EisenbergEffect: I'm not seeing it with this example: https://github.com/microsoft/fast/tree/master/examples/todo-app-tutorial

ramonck: > I'm not seeing it with this example: https://github.com/microsoft/fast/tree/master/examples/todo-app-tutorial

@EisenbergEffect I'm going to try that in a while in my other PC.
In the mean while the comment about Fast not using babel when we think about ES6 and webpack it's part of the heart of it to have babel so if you look into the fast packages you'll find them, see below.
https://github.com/microsoft/fast/tree/master/packages/utilities/fast-web-utilities
image

https://github.com/microsoft/fast/blob/master/packages/utilities/fast-web-utilities/package.json
image

https://github.com/microsoft/fast/blob/master/package.json
image

EisenbergEffect: I don't think those are deps of the web components packages though. /cc @janechu @chrisdholt 
Thanks for your patience @ramonck

chrisdholt: Correct, those are for the overall repo

ramonck: > I don't think those are deps of the web components packages though. /cc @janechu @chrisdholt
> Thanks for your patience @ramonck

@EisenbergEffect it's not a bug fix or anything so there's no rush on this the idea is just to help others really so not a problem at all, the more we socialize on how to help others the better it will be in the end so it's good that we're doing this.

I tried to run the example package as informed: https://github.com/microsoft/fast/tree/master/examples/todo-app-tutorial
The steps: npm install and then npm run dev
I'm getting an error as follows.
image
image

The screen:
image

Here's the npm run dev output:
image

What I find it weird is that you're forcing into the bundle in the main.ts and it still didn't recognize it.
image

chrisdholt: OK, this is *really good* context ^^^ :)! @eisenbergeffect I'm getting the above issues from other folks asking a similar question (FASTElement definition is missing error).

EisenbergEffect: This smells of having two versions of the library loaded somehow....

ramonck: > This smells of having two versions of the library loaded somehow....

@EisenbergEffect and @chrisdholt I found out it's because of the tag: todo-form and design-property-panel they are the tags giving the errors.
image

image

EisenbergEffect: Those components are the locally implemented ones vs. the ones that are exported from the fluent package. So, I'm thinking there's maybe a version mismatch between what fluentui is using vs. what this project is using. I was just talking to @chrisdholt about possibly using a peerDependency somewhere. He can link some relevant issues maybe.

chrisdholt: Yeah, I think this will need a bit more investigation as to where we would use the peer dependency - what is our guidance. Should libraries such as fast-components or fluentui/web-components expect a peer dependency, or ??? Either way, I think a fix here will help: https://github.com/microsoft/fluentui/pull/14573

ramonck: > Those components are the locally implemented ones vs. the ones that are exported from the fluent package. So, I'm thinking there's maybe a version mismatch between what fluentui is using vs. what this project is using. I was just talking to @chrisdholt about possibly using a peerDependency somewhere. He can link some relevant issues maybe.

@EisenbergEffect yes you're correct.

ramonck: > Yeah, I think this will need a bit more investigation as to where we would use the peer dependency - what is our guidance. Should libraries such as fast-components or fluentui/web-components expect a peer dependency, or ??? Either way, I think a fix here will help: [microsoft/fluentui#14573](https://github.com/microsoft/fluentui/pull/14573)

Ideally Fluent UI should just work separatly no matter what from a theory perspective and thinking about web components as well but in practice we know how things go.

EisenbergEffect: @ramonck Now that we've fixed the core build issue with the Fluent components, would you mind checking this out again?

ramonck: @EisenbergEffect same issue on my end.
My steps are the following:
1) Removed the previous fast folder and cleaned the browser caching (Edge).
2) git clone https://github.com/microsoft/fast.git
3) cd into the "examples/todo-app-tutorial"
4) npm install
5) npm run dev (even tried a npm run build and then a run dev still same thing)
image

EisenbergEffect: Apologies @ramonck This is the same issue as before. We had another release on Thursday of last week, but the Fluent packages didn't get updated. By the time you got around to testing the original bug we fixed had been re-introduced. @chirsdholt Can we update and release the Fluent Web Components today?

ramonck: @EisenbergEffect no problem, I don't know how you guys are working the continuos integration there for this project and the Fluent UI but if I may suggest if you're using Azure DevOps you can create a test scenario for the examples and see if you're coming out properly before any future release of either and if it's not ok either by visual comparison or by logs you can prevent this from happening in any future build you wil have.
If you're using any other tool you can also do the same on the Continuous Integration by focusing the test cases and making sure the examples are working properly if not then there's something wrong that needs to be fixed before the next release, I would say it can be your quality assurance metric there.

EisenbergEffect: 100% agree.

ramonck: Hi @EisenbergEffect any updates? Was wondering if I could test that example here with the fluentui going so we can see it all ok.

EisenbergEffect: @ramonck We've got a new FAST release coming this week. I think we should wait until that is out and ensure the Fluent components are updated based on that, then circle back around to this.

ramonck: Hi @EisenbergEffect sounds good, ignite week.

EisenbergEffect: @chrisdholt Have we released the fluent components since we released the fast components? We keep getting stuck in resolving this issue because an unrelated bug appears when trying to resolve this issue if the fluent components haven't been updated to the latest FAST libraries. I wanted to verify that everything is in order before asking @ramonck to try this out again.

chrisdholt: I don't think we're bumping every instance of FAST in the Fluent library - if I recall, we should be trying to leverage semver with reasonable minor version instances. We've released the Fluent components several times as they are on a nightly release. v0.5.0 released last night.

chrisdholt: @eisenbergeffect looking deeper into the collapsed detail above - I believe this issue is related to multiple instances of `fast-element` being loaded? Typically we see it manifest in `cannot resolve FASTElement definition`. That's been a common problem we've had with yarn generating and referencing multiple versions even in cases where they both should resolve to the same version. I've seen the issue resolved with a quick removal of the dupe in the yarn.lock file. I agree this bug is annoying and frustrating and I've been considering opening an issue to discuss how to best handle it. 

I don't know that the best path forward is to update to exact versions of `fast` packages at every publish. For teams internally that leverage this elsewhere (including mine) that would create a lot of overhead and churn across 3-4 repositories.

EisenbergEffect: I think it comes from the fact that fast-element is on a minor version with no major. We could go to a 1.0 release with that. Additionally, I could make the definition lookup storage exist in the global scope so that multiple versions could be loaded...I'm just not sure yet if we shouldn't error if that happens. It's not something you would normally want.

@ramonck Go ahead and give things another try. Seems like the libraries should be up to date.

EisenbergEffect:
I think we're going to close this out since we haven't been able to reproduce the issue.

PR #3703, opened 2020-08-12
fix: wrapping set focus in dom.queueupdate

### LINKED ISSUE #3685: issue with autofocus in text-field

When I use a `<fast-text-field>` with `autofocus` (`<fast-text-field autofocus>`) I get issues with the rest of the application. My use case is an Aurelia 2 project where using the `autofocus` interferes with the routing. I had a similar issue when using `autofocus` inside a `<fast-dialog>` where the autofocus would not work.

In both scenario, if I do the focus manually myself after a 50ms timeout (with `myElement.focus()`) then it works as it should.

[I suspect that the `focus()` is called a little too early here. ](https://github.com/microsoft/fast/blob/2879571796f210da49be8df90a74faee429bd34b/packages/web-components/fast-foundation/src/text-field/text-field.ts#L208)

* Maybe set up a page with a `<fast-text-field autofocus>` inside a `<fast-dialog hidden>`
* And then show the fast dialog and check if everything works fine, including the focus of the text field

* autofocus should work
* autofocus should not cause any issues with the rest of the app

* MacOS
* Chrome
* Aurelia 2 project

marjonlynch:
## Description
Autofocus seems to work but in the context of other wrapping frameworks it might be happening too soon.
potentially closes #3685 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/text-field/text-field.ts
+++ b/packages/web-components/fast-foundation/src/text-field/text-field.ts
@@ -204,11 +204,11 @@ export class TextField extends FormAssociated<HTMLInputElement> {
 
         this.proxy.setAttribute("type", this.type);
 
-        DOM.queueUpdate(() => {
-            if (this.autofocus) {
+        if (this.autofocus) {
+            DOM.queueUpdate(() => {
                 this.focus();
-            }
-        });
+            });
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-foundation/src/text-field/text-field.ts
```diff
@@ -204,9 +204,11 @@ export class TextField extends FormAssociated<HTMLInputElement> {
 
         this.proxy.setAttribute("type", this.type);
 
-        if (this.autofocus) {
-            this.focus();
-        }
+        DOM.queueUpdate(() => {
```
scomea, line 207: Can you check autofocus before queuing an update?  I don't think we want all text fields queing something.

chrisdholt: +1 - would suggest we only do this in the case autofocus is true.

marjonlynch: this makes sense, will change

marjonlynch: I'm assuming nothing can change with regard to this.autofocus between checking and queuing the update, so only 1 check before the queue call

PR #3705, opened 2020-08-12
feat: add initial version of fast-avatar

litteredwitherrors:
# Description

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+  ${display("flex")} 
+  :host {
+    max-width: calc(${heightNumber} * 2.5px);
+    height: calc(${heightNumber} * 2.5px);
+    position: relative;
+  }
+
+  .link {
+    text-decoration: none;
+    color: var(--neutral-foreground-res);
```
chrisdholt, line 19: We'll want to interpolate the color variables behaviors here and point to them such as [this example](https://github.com/microsoft/fast/blob/1cdf9ef352d6a0f1c385badc8d3255e9c0a78cc6/packages/web-components/fast-components/src/divider/divider.styles.ts#L12).

packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    align-items: center;
+    min-width: 100%;
+  }
+
+  .square {
+    border-radius: 5px;
```
chrisdholt, line 28: For square, assuming per the spec that this would map to the corner-radius design system value?

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
...
+ * The template for {@link @microsoft/fast-foundation#Avatar} component.
+ * @public
+ */
+export const AvatarTemplate = html<Avatar>`
+  <template>
+    <div 
+      class="coin ${x => (x.shape === "square" ? "square" : "circle")}"
+      style="${x =>
+        x.fill || x.color
+            ? `background-color: var(--badge-fill-${x.fill}); color: var(--badge-color-${x.color})`
```
chrisdholt, line 14: Should this be `badge-fill-____` and `badge-color-____` or should we go with a more semantic name for mapping here?

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export class Avatar extends FASTElement {
+  /**
+   * Indicates the Avatar should have a name.
+   * 
+   * @public
+   * @remarks
+   * HTML Attribute: name
+   */
+  @attr({ attribute: "name" })
```
chrisdholt, line 16: For the attrs, you can leave the object off here unless you need it. An example where you might use this would be:

```ts
@attr({ attribute: "background-color" })
public backgroundColor: string;
```

chrisdholt: I think you can drop all uses below :)

packages/web-components/fast-components/src/avatar/avatar.styles.ts
dlesaca: Yes, that is correct.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
dlesaca: That, my friend, is a copy-pasta mistake. Will fix, thanks.

EisenbergEffect:
@litteredwitherrors What do we need to help move this through? stale bot marked it a few days ago.

packages/web-components/fast-foundation/src/avatar/avatar.spec.md
```diff
@@ -45,21 +45,28 @@ A common use case would be to display an image or text (usually initials) of a u
...
-          alt="${x => x.alt}"
-          tabindex="${x => (!x.link ? "0" : void 0)}"
-        />
-      `)}
-      <span class="avatar-name">${x => x.name}</span>
-    </a>
-  </div>
-  <slot class="avatar-badge" name="badge"></slot>
-</template> 
+  <template>
```
chrisdholt, line 48: Looks like we may not need the template here if we aren't using it

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
+import { html, when } from "@microsoft/fast-element";
+import { Avatar } from "./avatar";
+
+/**
+ * The template for {@link @microsoft/fast-foundation#Avatar} component.
+ * @public
+ */
+export const AvatarTemplate = html<Avatar>`
+  <template>
```
chrisdholt, line 9: We can drop this if we aren't using it.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
...
+    <div 
+      class="coin ${x => (x.shape === "square" ? "square" : "circle")}"
```
chrisdholt, line 11: We may want to use the name of "control" here (vs the :host) - looks like we need a CSS part name here as well.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
...
+      style="${x =>
+        x.fill || x.color
+            ? `background-color: var(--avatar-fill-${x.fill}); color: var(--avatar-color-${x.color});`
+            : void 0}"
+    >
+      <a class="link" href="${x => (x.link ? x.link : void 0)}">
```
chrisdholt, line 17: CSS part name here as well

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
...
+        ${when(x => x.imgSrc, html`
+          <img 
+            src="${x => x.imgSrc}"
+            alt="${x => x.alt}"
+            tabindex="${x => (!x.link ? "0" : void 0)}"
+            class="image"
```
chrisdholt, line 23: CSS part

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
...
+          />
+        `)}
+        <span class="name">${x => x.initials}</span>
```
chrisdholt, line 26: CSS part

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,97 @@
...
+  @attr public link: string;
+
+  /**
+   * Indicates the Avatar shape should be. By default it will be set to "circle".
+   * 
+   * @public
+   * @remarks
+   * HTML Attribute: shape
+   */
+  @attr public shape: string = 'circle';
```
chrisdholt, line 91: Perhaps we can have this be a type defined above?

```ts
type AvatarShape = "circle" | "square";

public shape: AvatarShape;
```

Last thing - we'll want to set this default in connectedCallback rather than here. In certain scenarios this will throw an exception.

packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+  accentFillRestBehavior,
+  heightNumber,	
+  neutralForegroundRestBehavior
+} from "../styles/index";
+
+export const AvatarStyles = css`
+  ${display("flex")} 
+  :host {
+    max-width: calc(${heightNumber} * 2.5px);
+    height: calc(${heightNumber} * 2.5px);
```
chrisdholt, line 13: Curious if we might want to add a a variable here that would allow us to control the size if we wanted. We could default the variable to this value?

packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  .name {
+    font-size: var(--type-ramp-plus-5-font-size);
+    line-height: var(--type-ramp-plus-5-line-height);
+    display: block;
+  }
+
+  ::slotted(fast-badge) {
+    --badge-fill-primary: #00FF00;
```
chrisdholt, line 57: Let's go with `accent` here if we're going to add a default slotted value - I know we've used "primary" in docs, but we use accent in practice for our appearances. We likely should do so here as well.

chrisdholt: Do we have a set map here as well? I noticed this is a fairly bright green - just curious of the use case.

chrisdholt (changes requested):
This is looking good! A couple questions and a few things I noticed that we'll want to address prior to merging.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
litteredwitherrors: > We may want to use the name of "control" here (vs the :host)

Could elaborate on this for me?

> \- looks like we need a CSS part name here as well

I'm not 100% sure what you mean by "CSS part name". Could you define that?

litteredwitherrors: Ah, I think found the definition. You're saying each element needs a corresponding `part` attributed. Is that correct?

chrisdholt: Yeah, we need a CSS Part to provide the opportunity to reach into the shadow DOM when supported by the browser.

EisenbergEffect:
For `shape`, `fill`, `size`, `text-ratio`, and `color` should we have some defaults configured in the design system?

chrisdholt:
> For `shape`, `fill`, `size`, `text-ratio`, and `color` should we have some defaults configured in the design system?

Just want to note that I see a default for shape (circle), and we've removed "size" as we don't have enough research yet to really enable something that scales. I do think there's ways with this implementation for anyone to support size.

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,99 @@
...
+  @attr public initials: string;
+
+  /**
+   * Invoked when the `name` property changes.
+   * 
+   * @public
+   * @remarks
+   */
+  protected nameChanged(): void {
+    const result = this.name?.match(/\b\w/g) || [];
```
scomea, line 36: more readable if propertyChanged() function immediately follows the property it is associated with

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,92 @@
...
+   * 
+   * @public
+   * @remarks
+   * HTML Attribute: shape
+   */
+  @attr public shape: AvatarShape;
+
+  public connectedCallback(): void {
+    super.connectedCallback();
+    this.shape = "circle";
```
marjonlynch, line 90: This is over writing anything set on the element, discussed offline

marjonlynch: Also, talked offline about adding the definitions and config for fast-component-explorer

packages/web-components/fast-foundation/src/avatar/avatar.ts
litteredwitherrors: @chrisdholt I set the default like this, but it isn't working properly:

```js
    super.connectedCallback();
    this.shape = "circle";
  }
```
The above is taking precedence over any value passed into it (i.e. Square doesn't work). Do you know what the issue might be?

chrisdholt: Have you removed the default from line 91?

```suggestion
  @attr public shape: string;
```

chrisdholt: Otherwise shape is always circle in connected callback

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,35 @@
...
+                    : void 0}"
+        >
+            <a class="link" part="link" href="${x => (x.link ? x.link : void 0)}">
+                ${when(
+                    x => x.imgSrc,
+                    html`
+                        <img
+                            src="${x => x.imgSrc}"
+                            alt="${x => x.alt}"
+                            tabindex="${x => (!x.link ? "0" : void 0)}"
```
chrisdholt, line 25: @litteredwitherrors can you elaborate on this being focusable when there is no link?

litteredwitherrors: Currently, adding a link adds an `href` to the anchor tag. The anchor is still present when `link` is omitted, and focus is still delegated to it. I am noticing that there is no focus indication, though, on any of the components. I'm going to have to fix this.

litteredwitherrors: @chrisdholt Apparently, `delegatesFocus` was never added. I added it and that fixed all the weird focus issues. I'll push the fix soon.

chrisdholt: @litteredwitherrors still seeing this here - I think if the anchor is a link it should receive focus now. If it's not, I'd expect it to not focusable. aShould be able to have this just removed.

litteredwitherrors: @chrisdholt Strange. As I'm testing it right now, focus does not fall on a component that does not have a `link` attribute.

chrisdholt: Is this the most recent instance of the code? You mentioned pulling this, but GH doesn't seem to recognize that.

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Internal
+     */
+    public connectedCallback(): void {
+        super.connectedCallback();
+        if (!this.shape) {
+            this.shape = "circle";
+        }
+        this.shape;
```
marjonlynch, line 96: is this line necessary?

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,98 @@
...
+     * Indicates the Avatar should have a name.
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: name
+     */
+    @attr public name: string;
+    protected nameChanged(): void {
+        const result = this.name?.match(/\b\w/g) || [];
+        this.initials = ((result.shift() || "") + (result.pop() || "")).toUpperCase();
```
marjonlynch, line 21: What happens if the initials derived here are different than what is set in the initials attribute, if they set both but the initials are something different than in the name they'll be overwritten? Is that acceptable? It might be I'm just asking the question.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,35 @@
+import { html, when } from "@microsoft/fast-element";
+import { Avatar } from "./avatar";
+
+/**
+ * The template for {@link @microsoft/fast-foundation#Avatar} component.
+ * @public
+ */
+export const AvatarTemplate = html<Avatar>`
+    <template>
```
marjonlynch, line 9: should the alt text or some aria stuff be on the template/host element here for the screen reader?

packages/web-components/fast-foundation/src/avatar/avatar.ts
litteredwitherrors: Removed

packages/web-components/fast-foundation/src/avatar/avatar.ts
litteredwitherrors: Added a check to see if an initial is given. If there is no initial given it defaults to the initialize name.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
litteredwitherrors: All this should be delegated to the image and anchor elements.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,33 @@
+import { html, when } from "@microsoft/fast-element";
+import { Avatar } from "./avatar";
+
+/**
+ * The template for {@link @microsoft/fast-foundation#Avatar} component.
+ * @public
+ */
+export const AvatarTemplate = html<Avatar>`
+    <div
+        class="coin ${x => (x.shape === "square" ? "square" : "circle")}"
```
chrisdholt, line 10: nit - and I could be missing something, but could this be:

```suggestion
    class="coin ${x => x.shape === "square" ? "square" : "circle")}"
```

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,33 @@
...
+                : void 0}"
+    >
+        <a class="link" part="link" href="${x => (x.link ? x.link : void 0)}">
+            ${when(
+                x => x.imgSrc,
+                html`
+                    <img
+                        src="${x => x.imgSrc}"
+                        alt="${x => x.alt}"
+                        tabindex="${x => (!x.link ? "0" : void 0)}"
```
chrisdholt, line 24: I'm not sure we want to bring focus here in the event that there isn't a link - I don't think anything is actionable>

```suggestion
```

EisenbergEffect (approved):
Pending Chris' feedback.

packages/web-components/fast-components/src/avatar/scenarios/index.html
```diff
@@ -0,0 +1,27 @@
...
+
+            fast-badge {
+                right: 0px;
+                bottom: 0px;
+            }
+        </style>
+        <fast-avatar
+            alt="Annie's profile image"
+            link="https://microsoft.com"
+            name="Carlos Riviera"
```
marjonlynch, line 21: You need 2 more things for the scenarios/component explorer, a config file and to export that config file from the index.ts in that same directory. pasting an example of the anchor version
Screen Shot 2020-12-15 at 1 33 23 PM

SethDonohue:
Should the HC colors be tweaked a bit so the badge shows up?
@khamudom  Thoughts?
Screen Shot 2020-12-15 at 3 28 49 PM

SethDonohue (approved):

khamudom:
> Should the HC colors be tweaked a bit so the badge shows up?
> @khamudom Thoughts? Maybe double check since my HC colors aren't always correct....
> Screen Shot 2020-12-15 at 3 28 49 PM

Since the badge is only a color, this is to be expected for high contrast. I have a feeling the badge will be a case by case scenario, depending on what the badge is (svg, image, color).

marjonlynch (approved):
This is looking very good, the only nit I have now is what ends up in fast-component-explorer, from the scenarios, you might want to tidy up just a bit in terms of making the two example names match, the color attribute is coming out as "foo" and the link should maybe be "#" so if the user clicks on it in the preview window they don't actually leave the page

packages/web-components/fast-components/src/avatar/avatar.definition.ts
```diff
@@ -0,0 +1,62 @@
...
+                    name: "link",
+                    description: "The link attribute",
+                    type: DataType.string,
+                    default: undefined,
+                    required: false,
+                },
+                {
+                    name: "shape",
+                    description: "The shape attribute",
+                    type: DataType.string,
```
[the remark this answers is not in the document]
litteredwitherrors: I believe it's supposed to reference the name of the attribute.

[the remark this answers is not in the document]
robarbms: I thought it might refer to the HTML element vs the internal attribute.

packages/web-components/fast-components/src/avatar/avatar.definition.ts
```diff
@@ -0,0 +1,62 @@
...
+export const fastAvatarDefinition: WebComponentDefinition = {
+    version: 1,
+    tags: [
+        {
+            name: "fast-avatar",
+            description: "The FAST avatar element",
+            attributes: [
+                {
+                    name: "initials",
+                    description: "The initials attribute",
```
bheston, line 13: For all the descriptions, can we use something more informative? Something like "Person's initials to show in place of a profile image".

packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const AvatarStyles = css`
+    ${display("flex")} :host {
+        max-width: var(--avatar-size, var(--avatar-size-default));
+        height: var(--avatar-size, var(--avatar-size-default));
+        position: relative;
+        --avatar-size-default: calc(
+            (
+                    (var(--base-height-multiplier) + var(--density)) * var(--design-unit) +
+                        (var(--design-unit) * 10)
```
bheston, line 13: This seems very large. I'm working on adding the designs for this now, but until we land a direction where this should be much larger can we make this the standard height (which will end up 32px with default values)?

packages/web-components/fast-components/src/avatar/fixtures/base.html
```diff
@@ -0,0 +1,108 @@
+<fast-design-system-provider use-defaults>
+    <style>
+        fast-avatar {
+            --avatar-fill-accent-primary: #cf4073;
```
bheston, line 4: Is this a different case than other token APIs because they involve custom names? With other components we've avoided prefixing them with the component name.

packages/web-components/fast-components/src/avatar/avatar.definition.ts
```diff
@@ -0,0 +1,62 @@
...
+                },
+                {
+                    name: "link",
+                    description: "The link attribute",
+                    type: DataType.string,
+                    default: undefined,
+                    required: false,
+                },
+                {
+                    name: "shape",
```
bheston, line 47: Can we add supported values here?

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,31 @@
+import { html, when } from "@microsoft/fast-element";
+import { Avatar } from "./avatar";
+
+/**
+ * The template for {@link @microsoft/fast-foundation#Avatar} component.
+ * @public
+ */
+export const AvatarTemplate = html<Avatar>`
+    <div
+        class="coin ${x => x.shape}"
```
bheston, line 10: Could we use the part name `coin` to target styles and remove the css class?

packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,68 @@
...
+        position: absolute;
+        display: block;
+    }
+
+    .name {
+        font-size: calc(
+            (var(--avatar-text-size) + var(--avatar-size, var(--avatar-size-default))) /
+                var(--avatar-text-ratio)
+        );
+        line-height: var(--type-ramp-plus-5-line-height);
```
bheston, line 58: I'm not entirely sure what algorithm we should use to scale this, but I don't think we can get away with a fixed value like this. For text that's contained like this, where we don't support wrapping and want it centered, we could consider setting the line height to the control height.

packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    .image {
+        width: 100%;
+        position: absolute;
+        display: block;
+    }
+
+    .name {
+        font-size: calc(
```
bheston, line 54: I think this is the only place this has come up, but I'm concerned designers won't want calculated font sizes that don't land on the ramp. Thinking through the algorithm, I think it would be something like "use the next smallest type ramp size where the line height is less than or equal to half the avatar height". I realize there's not a good way to handle that in css though. Wanted to propose it to see what solutions we might have.

EisenbergEffect:
How are we doing on this one? The PR has been around for about 10 months.

chrisdholt:
> How are we doing on this one? The PR has been around for about 10 months.

@EisenbergEffect @litteredwitherrors just did a pass on this last week to ensure that updates made to a couple design systems (Fluent for example) can be accommodated with the current implementation. If that's the case, we're good to move this in.

litteredwitherrors:
@EisenbergEffect @chrisdholt Here's an overview of the comparison between FAST Avatar vs Fluent (React) Avatar.

# Avatar Component Feature Parity

| Feature           | FAST | FLUENT (React) |
|-------------------|------|--------|
| Badge slot        | X    | X      |
| Label slot        |      | X      |
| Image slot        | X    | X      |
| Activity indicator|      | X      |
| Activity display  |      | X      |
| Square attr       | X    | X      |
| Size attr         |      | X      |
| Color attr        |      | X      |
| Actionable        | X    |        |
| Anchor based      | X    |        |
| Span based        |      | X      |

## Feature Analysis
### Slots
Both FAST and Fluent have slots for a badge and image. Only the Fluent React has a label slot.

### Props
The main props that aren't supported by FAST are the "activity" properties.

### Shape
Both elements can be set to a square shape.

### Actionable 
The FAST element is actionable via delegation to an optional anchor tag that is applied when a link is passed through the `link` attribute. The fluent React component is a static `span` element.

packages/web-components/fast-components/src/avatar/avatar.styles.ts
chrisdholt: @bheston do you have additional thoughts or ideas on this?

packages/web-components/fast-components/src/avatar/avatar.styles.ts
chrisdholt: @bheston do you have additional thoughts or ideas on this?

packages/web-components/fast-components/src/avatar/avatar.styles.ts
litteredwitherrors: > This seems very large. I'm working on adding the designs for this now, but until we land a direction where this should be much larger can we make this the standard height (which will end up 32px with default values)?

Could you clarify this? Are you suggesting that the default size be set to 32px?

chrisdholt:
@litteredwitherrors is going to move this branch to work against `features/design-system-vNext`. We're going to take advantage of a couple of the new API's and since an update is needed anyway, that makes the most since. Closing this in favor of that incoming PR.

PR #3728, opened 2020-08-15
chore: combine and add validation against Open UI definitions

janechu:
# Description

This change:
- Combines Web Component definitions and Open UI definitions
- Adds a validation step for Open UI definitions

This will allow the Open UI project to consume the definitions and use them in their project rather than maintaining their own version, which is already out of date.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/flipper/flipper.definition.ts
```diff
@@ -29,13 +28,13 @@ export const fastFlipperDefinition: WebComponentDefinition = {
                     description: "The direction attribute",
                     values: [
                         {
-                            name: FlipperDirection.previous,
+                            name: "previous",
```
janechu, line 31: For some reason I can't resolve enums coming from `fast-foundation`. If someone has a clue as to why that might be happening let me know, as of right now these all had to change to strings.

packages/web-components/fast-components/build/generate-open-ui-definition.js
```diff
@@ -0,0 +1,211 @@
...
+                    ...tag,
+                    attributes: Array.isArray(tag.attributes)
+                        ? tag.attributes.map((attribute) => {
+                            return {
+                                ...attribute,
+                                default: typeof attribute.default === "string"
+                                    ? attribute.default
+                                    : attribute.default === void 0
+                                    ? void 0
+                                    : `${attribute.default}`
```
janechu, line 48: Web component definitions must always have strings as attribute values, right now we can specify number and booleans and this aids when converting to a JSON schema that the tooling can use. Here for our exported Open UI schema, we change it back so that it will validate as a string.

packages/tooling/fast-tooling/src/schemas/web-component.schema.ts
```diff
@@ -5,8 +5,7 @@
  */
 export default {
     $schema: "http://json-schema.org/schema#",
-    $id: "custom-element-definition",
-    id: "custom-element-definition",
+    $id: "https://fast.design/docs/custom-element-definition",
```
janechu, line 8: Although this does not currently exist at this location, this must be a valid URI.

packages/web-components/fast-components/src/flipper/flipper.definition.ts
EisenbergEffect: It's possible these are const enums, which could make TS remove them by default. There's a compiler setting to have it generate code for const enums if that's the case.

janechu: Should they be changed to `type` as was done for a few other components recently? So `type FlipperDirection = "previous" | "next";`? Tagging @chrisdholt as I believe he made the change to other components. Also if we decide to do that I can leave these as strings, if we export them as enums I worry we will face a breaking change in the future if we want to backtrack and change the `enum` to `type` so maybe we should confirm that now.

nicholasrice: Maybe you just needed to re-build? I don't get an error importing and using this enum; it is not a const enum and is exported.

packages/web-components/fast-components/package.json
```diff
@@ -33,7 +33,8 @@
     "start": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "generate-default-palettes": "node -r esm build/generate-default-palettes.js",
-    "test": "yarn test-chrome:verbose && yarn doc:ci",
+    "generate-open-ui-definition": "tsc -p ./tsconfig.definitions.json && node -r esm build/generate-open-ui-definition",
+    "test": "yarn run build && yarn test-chrome:verbose && yarn doc:ci",
```
nicholasrice, line 37: This is going to cause ci to run build twice

janechu: Ah, I can remove, the goal was to generate the open ui definitions and run the validator against them

packages/web-components/fast-components/src/flipper/flipper.definition.ts
janechu: You were right, just needed a rebuild, thanks!

PR #3742, opened 2020-08-18
WIP: fix: move component proxy creation to connectedCallback

### LINKED ISSUE #3741: bug: input proxy causes empy label to display

When a proxy input is needed it is created in the slot. This prevents the `label__hidden` class from being added.

Edge:
```
<fast-text-field>
  #shadow-root
    <label class="label label__hidden" part="label" for="control">
      <slot></slot>
    </label>
    ...
</fast-text-field>
```

Firefox:
```
<fast-text-field>
  #shadow-root
    <label class="label" part="label" for="control">
      <slot><input /></slot>
    </label>
    ...
</fast-text-field>
```

Label should be hidden in all browsers.

* Browser Mozilla FireFox
* Version 1.8.0

CuddleBunny:
# Description

Moved creation of the proxy element to connectedCallback.

## Motivation & context

Resolves #3741.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/button/README.md
+++ b/packages/web-components/fast-components/src/button/README.md
@@ -1,4 +1,4 @@
 # fast-button
-`fast-button` is a web component implementation of an [HTML button element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button). The component supports several visual apperances (accent, lightweight, neutral, outline, stealth).
+`fast-button` is a web component implementation of an [HTML button element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button). The component supports several visual appearances (accent, lightweight, neutral, outline, stealth).
 
 For more information view the [component specification](../../../fast-foundation/src/button/button.spec.md).
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.spec.md
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.spec.md
@@ -10,7 +10,7 @@ It is envisioned that this component would be used as a building block for other
 
 ### Features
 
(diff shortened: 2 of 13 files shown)


CuddleBunny:
@chrisdholt - Sure thing, I will take a look at them.

CuddleBunny:
@chrisdholt - done.

EisenbergEffect:
Definitely want a review on this particular PR from @nicholasrice when he gets the chance.

nicholasrice:
Thanks for looking into this @CuddleBunny. I'm unsure why this fixes the issue; The proxy is appended during the `connectedCallback` of `FormAssociated` and I wouldn't expect changing *when* the proxy is created to affect the `slotted` binding. The proxy is still appended in the default slot when it is appended isn't it? If so, still expect the `slotted` binding to update the observable property and add the hidden class. Is there something I'm missing?

nicholasrice:
A thought I had was that `FormAssociated` could append a `<slot name="proxy"></slot>` to the shadow root and then append the proxy to proxy slot. That would prevent this sort of issue in the future as well. @EisenbergEffect do you have thoughts on that? Would that screw up anything in the templating engine?

EisenbergEffect:
@nicholasrice That's shouldn't break anything. The template engine just does and appendChild on the shadow root. It doesn't remove any nodes that may have been placed there outside of the templating engine's code. The one exception to that is if the element is doing custom rendering, meaning that it doesn't have a template, and then someone dynamically configures a template later, the engine doesn't know any other way to handle this than to clear out the entire shadow root and then append the new template. I don't think that's something that needs to be considered in this case though. Just including that for completeness.

CuddleBunny:
Closing this to create a new PR which takes a new approach.

PR #3746, opened 2020-08-19
fix: consistent with min-width, min-height is at least the thumb size

### LINKED ISSUE #3644: Height is not being respected in fast-slider

When you set the height to 0% of `<fast-slider orientation="vertical">`  it is not being respected since it already has a min-height set to:

```css
    :host(.vertical) {
        height: 100%;
        min-height: calc(var(--design-unit) * 60px);
        min-width: calc(var(--design-unit) * 20px);
    }
```

Example:
Run the following code:

```html
<!DOCTYPE html>
<html>

  <head>
    <meta charset="UTF-8" />
    <script type="module" src="https://unpkg.com/@fluentui/web-components"></script>
  </head>

  <body>

      <fluent-design-system-provider use-defaults>
          <fluent-slider max="100" value="20" style="height: 0%;"  orientation="vertical"></fluent-slider>  
      </fluent-design-system-provider>
    
  </body>

</html>
```

I would like the height to be adjusted at least at a minimum size of the thumb

This issue is related to #3634, which has already been fixed, but for the height the solution will not be too straightforward at least from my point of view

From what I saw I believe the fix here will not be as simple as setting it to `thumb-size * 1px` because doing that the slider will normally appear with a height of `16px`. The normal `<input type="range">` will appear as 160px height when you did not set the value manually and the `min-height` remains as auto

marjonlynch:
# Description
We now have a default height, --fast-slider-height,  of (--thumb-size * 10px), which should translate to 160px for our design defaults. There is also a min-height based on --thumb-size

Some rationale behind this decision, height:100% works different than width:100% in that if you don't have anything specifying an actual height then height:100% does nothing, whereas width:100% will have the control span all the width available. Because pages are intended to scroll vertically, the height is arbitrary
but pages are intended to span the full width. So that the slider doesn't appear broken at first we want to specify a default height that can easily be overridden by inline style.

closes #3644

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/slider/fixtures/base.html
+++ b/packages/web-components/fast-components/src/slider/fixtures/base.html
@@ -8,7 +8,7 @@
         "
     >
         <h4>Toggle orientation</h4>
-        <div style="margin-bottom: 60px;">
+        <div style="margin-bottom: 60px; height: 200px;">
             <button
                 style="width: 90px; margin: 10px;"
                 onclick="const slider = document.getElementById('switcher'); slider.getAttribute('orientation') === 'horizontal' ? slider.setAttribute('orientation', 'vertical') : slider.setAttribute('orientation', 'horizontal');"
@@ -46,7 +46,11 @@
 
         <!-- Margin offsets vertical -->
         <h4>Vertical with top offset</h4>
-        <fast-slider orientation="vertical" , style="margin-left: 100px;"></fast-slider>
+        <fast-slider
+            orientation="vertical"
+            ,
+            style="margin-left: 100px; height: 120px;"
+        ></fast-slider>
 
         <!-- Negative positions -->
         <h4>Negative positions</h4>
--- a/packages/web-components/fast-components/src/slider/fixtures/base.html
+++ b/packages/web-components/fast-components/src/slider/fixtures/base.html
@@ -48,7 +48,6 @@
         <h4>Vertical with top offset</h4>
         <fast-slider
             orientation="vertical"
-            ,
             style="margin-left: 100px; height: 120px;"
         ></fast-slider>
 
--- a/packages/web-components/fast-components/src/slider/slider.styles.ts
+++ b/packages/web-components/fast-components/src/slider/slider.styles.ts
@@ -105,7 +105,7 @@ export const SliderStyles = css`
     }
     :host(.vertical) {
         height: 100%;
-        min-height: calc(var(--design-unit) * 60px);
+        min-height: calc(var(--thumb-size) * 1px);
         min-width: calc(var(--design-unit) * 20px);
     }
     :host(.disabled) .label,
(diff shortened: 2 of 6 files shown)


chrisdholt:
@marjonlynch can you elaborate on what you mean by, "This requires developers to set a height on either the fast-slider when it's vertical or it's container." Does this mean that every instantiation of a vertical slider will require an explicit height to be set? That seems way less than ideal. Can you share what the tradeoffs are there and how you decided on this being the best course of action?

packages/web-components/fast-components/src/slider/fixtures/base.html
```diff
@@ -46,7 +46,11 @@ <h4>horizontal with left offset</h4>
 
         <!-- Margin offsets vertical -->
         <h4>Vertical with top offset</h4>
-        <fast-slider orientation="vertical" , style="margin-left: 100px;"></fast-slider>
+        <fast-slider
+            orientation="vertical"
+            ,
```
chrisdholt, line 51: Noticed this while running this just now, I don't think this comma is meant to be an attribute :).

chrisdholt:
@marjonlynch I think a default height here would be helpful, especially if it needs to be set regardless. This way, if I put it on the page *without* an explicit height, I can at least see the entire slider on the screen and it doesn't feel broken. Assigning a specific CSS variable for here may be a good way to manage this as the variable would be assigned with a default value. Changing that variable would change the height.

packages/web-components/fast-components/src/slider/slider.styles.ts
```diff
@@ -104,8 +105,8 @@ export const SliderStyles = css`
         border-radius: calc(var(--corner-radius) * 1px);
     }
     :host(.vertical) {
-        height: 100%;
-        min-height: calc(var(--design-unit) * 60px);
+        height: calc(var(--fast-slider-height) * 1px);
```
EisenbergEffect, line 108: I like the approach of providing a variable specifically for this purpose and then having that fallback to a calculation based on thumb size. This seems pretty reasonable as a default.

packages/web-components/fast-foundation/src/slider/slider.spec.md
```diff
@@ -47,7 +47,7 @@ A designer can override any internal styling applied to the slotted labels, the
 - `step` - limits the values of the `slider` to increments of the `step` value added to the minimum value of the 
 `slider`'s total range.  The default value is 1. The minimum and maximum values of a `slider`'s range are always valid results regardless of the `step` prop. The `step` prop is used as the value for incrementing the thumb by pressing the arrow keys.
  - `value` - Allows authors to specify the initial selected range of the `slider`.  It defaults to a (step constrained) value at the midpoint on the `slider`'s total range. In `range` mode `value` is expected to be a 2 element array where the lower value is the first value and upper value is the second.
- - `orientation` - horizontal or vertical values allowed.
+ - `orientation` - horizontal or vertical values allowed. Note that if the slider is in vertical orientation by default the component will get a height using the css var `--fast-slider-height`, by default that equates to `(10px * var(--thumb-size))` or 160px. Inline style will override that height.
```
EisenbergEffect, line 50: I think we may want to at least add this as a note in the readme for the slider as most folks aren't likely to jump to the spec and are probably going to be looking at the docs on the site instead.

marjonlynch: ok will do

PR #3759, opened 2020-08-20
feat(examples): add new example showing a list-box component

EisenbergEffect:
## Description

This is a PR intended primarily as an example. It demonstrates building a list-box component.

> Note: The component is not production-quality. For example, no keyboard navigation was implemented.

The approach is to create both the list and item components and design the list-box so that its items/rows exist in light dom and can be manually created. At the same time, the component supports setting data properties and templates which enable it to render the items on behalf of the developer. There are several demos which show these different usages.

By allowing the items to exist in the light dom and not requiring the component to be driven by data, it enables 3rd party frameworks to generate the items. Without that, we would have to force devs into using FAST's templating directly. With this approach, they can decide for themselves how they want to work with the component.

Of special note is how the repeat directive is used internally to render to the light dom only when an array of items is provided. A VirtualRepeatDirective could be constructed in a similarly reusable form for a data-grid type scenario.

Screen Shot 2020-08-20 at 5 37 42 PM

## Instructions

To run the example, first `npm install` in the specific example folder and then `npm run dev` to see it in action.

### DIFF
--- a/examples/list-box-sample/package.json
+++ b/examples/list-box-sample/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "fast-todo-app",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "rimraf dist && webpack --config webpack.config.js --mode=production",
+    "dev": "webpack-dev-server",
+    "serve": "http-server -c-1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@microsoft/fast-element": "latest",
+    "tslib": "^1.11.1"
+  },
+  "devDependencies": {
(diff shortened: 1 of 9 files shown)


examples/list-box-sample/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "fast-todo-app",
```
SethDonohue, line 2: Probably want to change this name?

SethDonohue:
Thanks for this demo! `:itemTemplate=${personTemplate}` is neat :)

examples/list-box-sample/package.json
EisenbergEffect: Ooops. Copy/paste error. Will fix it up.

PR #3770, opened 2020-08-25
docs(spec): add skeleton spec

litteredwitherrors:
# Description

Initial draft for Skeleton spec

## Issue type checklist

## Process & policy checklist


specs/skeleton.md
```diff
@@ -0,0 +1,99 @@
...
+- [React Loading Skeleton](https://github.com/dvtng/react-loading-skeleton#readme)
+- [Semantic UI (Placeholder)](https://semantic-ui.com/elements/placeholder.html)
+- [Telus](https://tds.telus.com/community/index.html#skeleton)
+- [Thumbprint](https://thumbprint.design/guide/product/loaders/#example-2.-skeleton-loader)
+- [Vercel](https://vercel.com/design/skeleton)
+
+### API
+
+#### Component Name
+- `fast-avatar`
```
nicholasrice, line 43:
```suggestion
- `fast-skeleton`
```

specs/skeleton.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### Attributes
+|   Name    | Description                                                 | Type                                |
+|-----------|-------------------------------------------------------------|-------------------------------------|
+| `shape`   | Determines the avatar coin shape. Rectangle by default.     | `string: rect (default) | circle`   |
```
bheston, line 48: A couple more "avatar" here.

specs/skeleton.md
```diff
@@ -0,0 +1,99 @@
...
+| `fill`    | Sets the background fill of the avatar coin.                | `string: hex color`                 |
```
bheston, line 49: It seems like the fill would be a complicated value, presumably a gradient and presumably based on the design system so the color coordinates with the rest of the app design system.

specs/skeleton.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Non-goals
+
+#### Specfic Skeletons
+
+Should not be made to reflect a specific element (e.g. `skeleton-button`, `skeleton-text`, `skeleton-heading`, etc...). The skeleton should be as generic of a container as possible as to represent various elements of different shapes and sizes.
+  
+### Features
+
+- `Shape`: Circle or rect shapes can be selected. Border radius of rectagled shape will be determined by the design system.
```
bheston, line 23: It would be possible to create most skeleton layouts with just circles and rects, though layout could be a considerable amount of effort and similarity to existing component templates. It would be cool if there were a way to apply this effect to components so they could go into that state easily.

Applying the border radius is great, and applying more aspects from the design system would help reduce this manual construction. I realize the other skeleton implementations don't follow that pattern, but they don't have an adaptive design system that could change so drastically between apps or even people. For instance, you'd represent a title with a rect, but you'd have to manually size it if there was no way to indicate it represented a page heading. Same with a button, whose size we also know based on the design system.

Maybe this is heading into v2 territory, but I'd like to see the adaptive design system play a role here and substantially reduce the markup required to create a skeleton interface for a full component or page.

chrisdholt: I think the feedback on how adaptive UI plays into this is good stuff to consider, but we should be hesitant to build our foundation templates and classes Specifically around adaptive UI. We are building the template for those systems as well as ours. Fast Frame and adaptive UI is just one of many design systems being supported here. If need be, we can create a more simple template and swap it out if that’s best. Foundation (And this spec) should be built to support as broad an application as we can. I’d expect for if the adaptive UI conversation to drive our implementation of this, not the foundation of it.

litteredwitherrors:
Hey everyone, 

After thinking through this spec a little more, I've come up with a concept that I'd like feedback on.

Thanks,
Dexter

# SVG Templates Concept

## Description
This feature allows a `fast-skeleton` component to accept a source for a skeleton template file through a `template` attribute or inline via component slot. The skeleton template would provide a transparent visual shape and layout of the component it is representing.

When used in this way the skeleton component would essentially act as a container for the provided SVG mask which can then be applied anywhere in an application.

This feature allows the `fast-skeleton` component to provide maximum design flexibility to represent many shapes and layouts.

The SVG would have CSS mask properties applied to it allowing underlying colors and effects (i.e. [shimmer](https://developer.microsoft.com/en-us/fluentui#/controls/web/shimmer)) to be visible through the transparent template.




SVG Template passed via `template` attribute:
```
<fast-skeleton template="https://example.com/template.svg"></fast-skeleton>
```

SVG Template passed via slot:
```
<fast-skeleton>
    <svg></svg>
</fast-skeleton>
```

Potential accepted formats:
- .svg (best)
- .png (suboptimal)




## Considerations

### Design
A design asset is required to use this feature. SVG code quality will have to be enforced to ensure proper rendering of the template masks.

### Development
If a developer does not have design support, they may use skeleton components as a stand-alone elements in an existing component. Developers also have the option to manually code SVG's to be used as templates. Again, code quality is a must.

### Accessibility
To maximize the information presented to assistive technologies inline SVG's are the preferred method for applying a template.

## Similar Concepts
The **GitLab** Design System utilizes and SVG code in the implementation of their "[Skeleton loader](https://design.gitlab.com/components/skeleton-loader/#presentation)".

chrisdholt:
@litteredwitherrors I love the above idea and I think it provides numerous methods of use and flexibility depending on how you'd use. Technically you could also build DOM with a background color and transparent shapes and slot that into place and as well. I really like the flexibility being proposed here as it opens up a lot of opportunity for implementation and provides a good foundation to scale from.

EisenbergEffect:
I'm wondering if we can have a different attribute name than `template`. My primary reason for that is that we were talking about introducing `FoundationElement` on top of `FASTElement` as a new way to integrate our components with the design system, and that would have a `template` property by default, allowing consumers to swap the shadow dom template out on a per instance basis (or maybe look up named templates in the design system). 

What about something like `pattern`?

chrisdholt:
> I'm wondering if we can have a different attribute name than `template`. My primary reason for that is that we were talking about introducing `FoundationElement` on top of `FASTElement` as a new way to integrate our components with the design system, and that would have a `template` property by default, allowing consumers to swap the shadow dom template out on a per instance basis (or maybe look up named templates in the design system).
> 
> What about something like `pattern`?

I like something like pattern; I think it makes sense and would remove any confusion with the above as well as existing "template" references.

EisenbergEffect:
On that note...we could even allow patterns to be looked up by name in the design system, once we've got the new provider model in place. We don't need to block for that but it's an enhancement we could do in the future.

`<fast-skeleton pattern="my-pattern"></fast-skeleton>`

EisenbergEffect:
I like the approach for this component, btw. I think it aligns nicely with our principles around composition and extensibility.

specs/skeleton.md
litteredwitherrors: Valid. I might have to explore this more.

specs/skeleton.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Background
+
+- [Github Issue #3755](https://github.com/microsoft/fast/issues/3755)
+
+### Use Cases
+
+Used to give users a sense of loading progress activity rather than merely presenting a blank page or container.
+
+### Non-goals
```
chrisdholt, line 15: We can remove this and any other unused sections :thumbsup:

chrisdholt (approved):

EisenbergEffect (approved):

PR #3776, opened 2020-08-25
fix: remove headings tag on fast-frame template

khamudom:
# Description

This was a pass to ensure we don't have multiple h1 on the page.
Fix the heading on the fast-frame template to be h4, and add content h4 style selector.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
+++ b/sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
@@ -20,11 +20,11 @@ export const FastFrameTemplate = html<FastFrame>`
                     class="${x => (x.expanded ? "tab-panel-expanded" : "")}"
                 >
                     <div class="content">
-                        <p class="heading">
-                            <span class="content-heading-highlight">Fast Frame</span>
-                            Dark Mode
-                        </p>
-                        <p class="sub-heading">Pre-built for both light and dark modes</p>
+                        <fast-badge class="content-badge">
+                            Fast Frame
+                            <span class="content-badge-highlight">Dark Mode</span>
+                        </fast-badge>
+                        <h3 class="heading">Pre-built for both light and dark modes</h3>
                         <p>
                             Switching between dark and light mode is as easy as changing
                             the background color of the design system.
@@ -48,11 +48,11 @@ export const FastFrameTemplate = html<FastFrame>`
                     class="${x => (x.expanded ? "tab-panel-expanded" : "")}"
                 >
                     <div class="content">
-                        <p class="heading">
-                            <span class="content-heading-highlight">Fast Frame</span>
-                            Colors
-                        </p>
-                        <p class="sub-heading">Pre-existing color you can customize</p>
+                        <fast-badge class="content-badge">
+                            Fast Frame
+                            <span class="content-badge-highlight">Colors</span>
+                        </fast-badge>
+                        <h3 class="heading">Pre-existing color you can customize</h3>
                         <p>
                             Color is applied by using color recipes which require two
                             color palettes, neutral and accent, applied to the design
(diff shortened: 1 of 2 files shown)


sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -114,22 +128,34 @@ export const FastFrameTemplate = html<FastFrame>`
...
+                    id="style-tab-panel"
+                    class="${x => (x.expanded ? "tab-panel-expanded" : "")}"
+                >
                     <div class="content">
-                        <h1><span class="content-heading-highlight">Fast Frame</span> Styles</h4>
+                        <h4>
+                            <span class="content-heading-highlight">Fast Frame</span>
+                            Styles
+                        </h4>
                         <h2>Adjust style settings on the fly</h2>
```
EisenbergEffect, line 155: Is this `h2` correct given the above change from `h1` to `h4`?

chrisdholt: This would continue to be an issue, let’s rake a full look across the entire page and ensure proper DOM hierarchy. There should be online tools that will transfer your markup to a more readable outline.

khamudom: @chrisdholt are there any specific tools you recommend? I found some chrome extensions that deal with headings, but not sure if they are doing the right job.

chrisdholt: I’d have to dig in here, and likely won’t have time this morning. I’d try to find a tool that outlines the document.

awentzel: Headings also will differ depending on where they are used.  It may be worthwhile investigating programmatic solutions such that inheritance determines they heading values used in the controls.  Also, it's confusing if you jump from H2 to H4 in a screen reader.  Headings should be contextual based on content, not styles.  In fact, the easiest way to implement correct heading structure is to turn off stylesheets together. Then come back through and style as needed once the semantic headings structure is correctly represented.

chrisdholt: Agree with the above in general, but not sure for the homepage a systematic approach is required or needed right now. I’d be happy with the immediate fixes for the correct semantics and structure. @khamudom, @awentzel details some helpful stuff in the above comment on landing and testing accurate structure.

khamudom: What I've noticed is when I am running a heading check extension, it doesn't find any other headings. For example, the headings in the `fast-frame` component are not getting checked as part of the website. But while taking a look at the `fast-frame` template, I think we can update the <h4> and <h2> tags to be a <p> tag. I did some testing and this seems to work when I also update the style selector.

sites/fast-website/src/app/components/fast-frame/fast-frame.template.ts
```diff
@@ -15,15 +15,21 @@ export const FastFrameTemplate = html<FastFrame>`
...
-                <fast-tab-panel id="contrast-tab-panel" class="${x =>
-                    x.expanded ? "tab-panel-expanded" : ""}">
+                <fast-tab-panel
+                    id="contrast-tab-panel"
+                    class="${x => (x.expanded ? "tab-panel-expanded" : "")}"
+                >
                     <div class="content">
-                        <h1><span class="content-heading-highlight">Fast Frame</span> Dark Mode</h4>
-                        <h2>Pre-built for both light and dark modes</h2>
+                        <p class="heading">
```
chrisdholt, line 23: Can you elaborate on how this is more semantic than using the correct heading tag here? This seems to bypass the issue of incorrect headings by removing any hierarchy altogether.

khamudom: I just thought we already have a section header and the headers in fast-frame could just be `p` elements but you are right, that it would not be semantically right, within the component itself. 
After further investigation, this looks similar to our sections, with a badge and a heading. The `h1` can be a badge, as an arbitrary element. The `h2` will be an `h3`.

`<fast-badge class="content-badge">
    Fast Frame
    <span class="content-badge-highlight">Dark Mode</span>
</fast-badge>
<h3 class="heading">Pre-built for both light and dark modes</h3>
`

PR #3779, opened 2020-08-25
fix: proxy element causes empty check on slots to fail

### LINKED ISSUE #3741: bug: input proxy causes empy label to display

nicholasrice:
# Description
The `FormAssociated` proxy element is appended to the host element when `ElementInternals` is not supported. This was causing checks for an empty default slot to have inconsistent behavior that depended on `ElementInternals` support.

This changes the way that the proxy is appended, creating a dedicated slot element for the proxy element and appending the slot element imperatively. This means the proxy will not be in the default slot and won't interfere with component implementations.

closes #3741

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/form-associated/form-associated.ts
```diff
@@ -424,6 +430,8 @@ export abstract class FormAssociated<
      * Attach the proxy element to the DOM
      */
     protected attachProxy() {
+        const proxySlotName = "form-associated-proxy";
```
EisenbergEffect, line 433: This can probably be lifted to the module scope.

packages/web-components/fast-foundation/src/form-associated/form-associated.ts
```diff
@@ -444,8 +452,14 @@ export abstract class FormAssociated<
             if (typeof this.value === "string") {
                 this.proxy.value = this.value;
             }
+
+            this.proxy.setAttribute("slot", proxySlotName);
         }
 
+        this.proxySlot = document.createElement("slot");
+        this.proxySlot.setAttribute("name", proxySlotName);
```
EisenbergEffect, line 460: Should these two lines be inside the `if` block as well? I was looking at the detachProxy code below and wondering if there was a need to re-create the slot or not.

Issue #3773, opened 2020-08-25
Create a custom CSS control to replace a textarea


janechu:
# Abstract

Representations of CSS currently exist as strings. This is too unwieldy as the default control in the `@microsoft/fast-tooling-react` export `Form` will be a `<textarea>`. To overcome this a better UI that ultimately exports a string should be used. There are some similarities with inline CSS and CSS inside a `<style>` element, both of these should be explored.

A style element (internal CSS) may contain a string that looks like:

```CSS
:host {
    color: green;
}
div {
    background: red;
    outline: none;
}
```

While an inline CSS string may need to be:
```CSS
background: red; outline: none;
```

### Abstract class `CSSControl`
- Create an abstract class CSSControl in the `src/form/custom-controls` directory following the naming structure `control.css.tsx`
- Create render methods for each "type" of CSS item and list them out, for example "color", "width", etc. which can be used singularly (eg. `background-color: red`) or combined (eg. `border: 1px solid red`)
- Create a local list of the CSS properties mapped to the type of control (see above)
- Create a method `convertObjectWithSInglePropertyToCSS` which will take an object with a single property:

```typescript
{
    background: "red"
}
```

And convert it to:

```typescript
"background: red;"
```

- Create a method `convertCSSPropertyToObject`

Which will break down the following string:

```typescript
"background: red;"
```

And convert it to:

```typescript
{
    background: "red"
}
```

### class `CSSInlineControl`
- Create a class CSSInlineControl in the `src/form/custom-controls` directory following the naming structure `control.css-inline.tsx`
- Create a method `convertObjectToCSS` which will break down the following object:

```typescript
{
    background: "red",
    color: "green"
}
```

And convert it to:

```typescript
"background: red; color: green; "
```

- Use the abstract class render methods to construct the UI

### class `CSSInternalControl`
- Create a class CSSInlineControl in the `src/form/custom-controls` directory following the naming structure `control.css-internal.tsx`
- Create a method `convertObjectToCSS` which will break down the following object:

```typescript
{
    background: "red",
    color: "green"
}
```

And convert it to:

```typescript
"{
      background: red;
      color: green;
}"
```

- Create a hierarchy of added property names to apply the style to, such as `:host`. This should behave similarly to the dictionary control behavior, allowing the addition of any number of keys.
- Use the abstract class render methods to construct the UI

### Testing

Add the `CSSInlineControl` and `CSSInternalControl` to the `app` custom controls example.

janechu:
[...]

PR #3780, opened 2020-08-25
chore: add spec for css custom control

janechu:
# Description

This change adds a spec for the CSS custom control for both inline CSS and internal CSS. This should provide additional functionality towards editing CSS.

## Motivation & context

Contributes a spec for #3773

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
```diff
@@ -0,0 +1,122 @@
...
+    {
+        "border": { // The property name should be pascalCase
+            name: "border", // The name should be spinal-case
+            default: "none",
+            value: [
+                {
+                    type: "width",
+                    default: "1px"
+                },
+                {
```
nicholasrice, line 24: Trying to follow this example - should this be "style" instead of "border-style" since none of the other border properties prefix with "border"?

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
```diff
@@ -0,0 +1,122 @@
...
+                    type: "border-style",
+                    default: "solid"
```
nicholasrice, line 26: Did you consider mapping more closely to the CSS properties themselves rather than the shorthand? "border" vs "border-style", "border-color", "border-width" or even "border-left-color", "border-left-style", "border-left-width".

Thinking out loud here - mostly just asking if this is going to provide all the flexibility needed because CSS shorthand gets a bit complicated, with argument count and argument order mattering.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: Good point, I'll change it, I think I left it in there accidentally.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: The generated UI should reflect the shorthand, but we could do it that way where we just have "border-width", "border-style", "border-color" in a sort of logical "grouping". I considered allowing everything and then having some method of configuration so that "border-left-style" would be omitted for the sake of brevity.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: So I added a `mapToCSSProperty` to indicate the CSS property the value maps to, that is what I was trying to also capture.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: I've updated to make `shorthand` CSS properties the default, if a user wants to set it so that they get all CSS properties they can set `shorthandOnly` to `false`. What we're trying to capture here is something designer friendly, so for example `border` should be a single line such as a `<fieldset>` with the relevant CSS properties it contains as number input (for width), select dropdown (for style) and autocomplete (for color). It's a similar UI for other design tools like figma.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
```diff
@@ -0,0 +1,140 @@
...
+Some utilities that may be required:
+- `convertObjectToCSS` - convert an object + selectors to a CSS string, if no selectors are available convert it to the inline equivalent.
+- `convertCSSToObject` - convert a string to an object + selectors (if available)
+
+## Overrides
+
+To make this flexible, the abstract class should allow a way to override the default creation of any property. The implementation of this in the CSS inline or CSS internal controls may look like:
+
+```tsx
+<CSSInline
```
EisenbergEffect, line 85: I wonder if we could start introducing web components for any new controls...

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
```diff
@@ -0,0 +1,140 @@
...
+### Render methods
+
+Render methods should be created for the following inputs:
+- Text field
+    - Used as a default value if the property name is not recognized
+- Autocomplete dropdown
+    - Used by the property name selector, once this has been chosen, the value should reflect different inputs
+    - Used by color (full list of available colors can be autocompleted)
+    - Used by border-style
+    - Used by anything which could use a CSS function such as var()
```
EisenbergEffect, line 67: Can you expand more on how things like `var` and `calc` would be handled?

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
```diff
@@ -0,0 +1,140 @@
+# Abstract
+
+The CSS controls consist of an abstract control and two extended controls, inline CSS and internal CSS. These controls will provide a way to interpret CSS strings for inline and internal CSS into form UI and back to strings.
```
EisenbergEffect, line 3: By "internal" CSS, does that refer to editing the adopted style sheets of a component? I wasn't sure what that referred to.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: I thought about that, but I think introducing a wrapper, among other things, plus the additional testing, is going to be more of a headache than just having two sets and eventually deprecating the React version. I'm keeping the React components as web component-like as possible as an intermediary step.

janechu: Open for discussion on that though, I'm just not sure the payoff is worth it.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: Now that I've thought about it, CSS functions need their own supplied input, I'll re-work this section.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
janechu: Oh, so, CSS can exist in 3 locations as defined by [mdn](https://developer.mozilla.org/en-US/docs/Learn/CSS/First_steps/How_CSS_is_structured) and reference to a stylesheet in the `<head>` of the DOM for example is "external", in a `<style>` element as "internal" and as the style property on an element as "inline". I've used the word "internal" to mean in the `<style>` element, I know this isn't super common nomenclature but there doesn't seem to be a better description.

janechu: Since "external" and "internal" would functionally be the same syntax, I could change this to "styles" and "stylesheet".

EisenbergEffect: I like that distinction as it will also cover adopted stylesheets.

packages/tooling/fast-tooling-react/src/form/custom-controls/control.css.md
EisenbergEffect: Deferring to your judgement on when the right time is for that.

EisenbergEffect (approved):
One thing that would be helper for the render methods section is to see some quick back of the napkin type wireframes for what these might look like in the UI. The CSS variables/calc one is still a bit foggy in my mind. That said, I think we can work that out while implementing it most likely.

awentzel (approved):

chrisdholt (approved):

PR #3781, opened 2020-08-26
chore: add spec for shortcuts utility

### LINKED ISSUE #3774: Create a shortcut utility spec
# Abstract

A utility must be created to capture and respond to keyboard shortcuts that interact with the `MessageSystem`.

## Requirements
- Create a spec for a utility that will register itself with the `MessageSystem` and provide key capture events that can affect the `DataDictionary`, navigation and history.
- The spec should reside in the `@microsoft/fast-tooling` package in the `data-utilities` folder.

## Additional information
- Devise a method for injecting individual shortcut behaviors into the class
- Devise a method for attaching the listener behavior, this can be approached similarly to the validation, where controls that are registered to the `MessageSystem` can apply listeners based on custom events.

janechu:
# Description

This creates the spec for a shortcuts utility that registers various shortcut actions and provides enough hooks for a context menu.

## Motivation & context

Resolves #3774

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
+++ b/packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
@@ -20,7 +20,7 @@ Shortcut actions can either be custom or imported from a set of actions availabl
 
 ```typescript
 import { DuplicateShortcutAction, ShortcutAction, Shortcuts } from "@microsoft/fast-tooling"
-import { keyCodeAlt, keyCodeAltName , keyCodeTab, keyCodeTabName } from "@microsoft/fast-web-utilities"
+import { keyCodeTab, keyCodeTabName } from "@microsoft/fast-web-utilities"
 
 new Shortcuts({
     messageSystem: fastMessageSystem,
@@ -30,8 +30,7 @@ new Shortcuts({
             name: "Random action",
             keyCodes: [
                 {
-                    name: keyCodeAltName, // "Alt"
-                    value: keyCodeAlt, // 18
+                    altKey: true
                 },
                 {
                     name: keyCodeTabName, // "Tab"
--- a/packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
+++ b/packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
@@ -1,0 +1,126 @@
+# Abstract
+
+The shortcuts utility accepts an instance of the `MessageSystem` and through this sends information about available shortcuts and events.
+
+## Instantiating
+
+All the `Shortcuts` class needs to instantiate is an instance of the `MessageSystem`. It should then fire a custom event notifying all other items registered to the message system that shortcuts are available, with a listener that will capture keypress and identify which shortcut action to execute.
+
+```typescript
+import { Shortcut } from "@microsoft/fast-tooling"
+
+new Shortcuts({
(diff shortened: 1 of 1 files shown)


packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
```diff
@@ -0,0 +1,126 @@
...
+import { keyCodeAlt, keyCodeAltName , keyCodeTab, keyCodeTabName } from "@microsoft/fast-web-utilities"
+
+new Shortcuts({
+    messageSystem: fastMessageSystem,
+    actions: [
+        // Custom shortcut
+        new ShortcutAction({
+            name: "Random action",
+            keyCodes: [
+                {
```
nicholasrice, line 32: Is the expectation here that the user is pressing `alt` and `tab`?  I wonder if it would make sense to map this more closely to the platform event object itself, where `altKey`, `ctrolKey`, and `shiftKey` are properties on the event object:

```js
new shortcutAction({
  keyCodes: [{name: keyCodeTabName, value: keyCodeTab, altKey: true}]
})
```

janechu: In hindsight, that was a bad example, that's a good point I'll update.

packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
```diff
@@ -0,0 +1,119 @@
...
+Shortcut actions can either be custom or imported from a set of actions available.
+
+```typescript
+import { DuplicateShortcutAction, ShortcutAction, Shortcuts } from "@microsoft/fast-tooling"
+
+new Shortcuts({
+    messageSystem: fastMessageSystem,
+    actions: [
+        // Custom shortcut
+        new ShortcutAction({
```
EisenbergEffect, line 28: It would be interesting to consider adding a `matches` method to the `ShortcutAction` implementation that would take an `event` object as input and check whether its keys match the event. In that way, various consumers of these shortcuts wouldn't need to duplicate that kind of code or necessarily know the details of how things are triggered. They simply wire the indicated event and then check matches on the event object.

packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
```diff
@@ -0,0 +1,119 @@
...
+            name: "Random action",
+            keys: [
+                {
+                    altKey: true
+                },
+                {
+                    value: "Tab"
+                }
+            ],
+            action: () => {
```
EisenbergEffect, line 38: Is there a way to get an action by name and invoke it manually?

packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
janechu: That's a good point, I did something similar on the custom components that go into the `<Form />` react component.

packages/tooling/fast-tooling/src/data-utilities/shortcuts.spec.md
janechu: I've added an `id` to the supplied config that can then be used to identify and run an action that has been supplied.

janechu:
I've merged this as the PR for this functionality has also already been approved and merged - see #3814

Issue #3584, opened 2020-07-28
compliance: Securing scripts and links used on third-party CDNs for Subresource Integrity


awentzel:
All products and services that provide web-accessible content must source all JavaScript and stylesheets (CSS) from one of the following locations:  This applies to all of our web sites. www.fast.design explore.fast.design color.fast.design and should be the model to use as we reintroduce the other websites over time.

1. The product/service itself (i.e. locally-hosted)
2. A Microsoft-operated service or Microsoft-operated CDN
3. A third-party CDN used with Subresource Integrity (SRI)

Fonts and images cannot be used with SRI, and so must be sourced from one of the following locations:

1. The product/service itself (i.e. locally-hosted)
2. A Microsoft-operated service or Microsoft-operated CDN

All web-accessible resources sourced from a location outside of the product/service itself should be accompanied by fallback code to a locally-hosted resource, or the application should be tested to ensure it functions appropriately if those resource are unavailable.

Ref: https://microsoft.visualstudio.com/Edge/_workitems/edit/26058078/ 

https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#:~:text=Subresource%20Integrity%20%28SRI%29%20is%20a%20security%20feature%20that,cryptographic%20hash%20that%20a%20fetched%20resource%20must%20match.

https://www.w3.org/TR/SRI/

### Known Issues to resolve

- The URL 'https://use.typekit.net/spx2dgm.css' is referenced from a link/script tags to a site that is not in whitelist, and subresource integrity is not used.
- The URL 'https://badge.fury.io/js/%40microsoft%2Ffast-components-msft.svg' is referencing a site that is not in whitelist.
- The URL 'https://badge.fury.io/js/%40microsoft%2Ffast-foundation.svg' is referencing a site that is not in whitelist.
- The URL 'https://badge.fury.io/js/%40microsoft%2Ffast-components.svg' is referencing a site that is not in whitelist.
- The URL 'https://badge.fury.io/js/%40microsoft%2Ffast-element.svg' is referencing a site that is not in whitelist.
- The URL 'https://img.shields.io/badge/License-MIT-yellow.svg' is referencing a site that is not in whitelist.
- The URL 'https://unpkg.com/@microsoft/fast-components' is referenced from a link/script tags to a site that is not in whitelist, and subresource integrity is not used.
- The URL 'https://placehold.it/300x150' is referencing a site that is not in whitelist.
[...]

PR #3782, opened 2020-08-26
chore: replace badges with local static versions on docs site

radium-v:
# Description
Handle security concerns described in #3584:
- Generate and host static badge icons for the documentation site
  - add a `generate:badges` step to `@microsoft/site-utilities` to generate the badges
    - static badges are generated based on a config
    - dynamic badges (ex. "npm package: v.v.v") are generated by reading versions in `package.json` files for each relevant package
  - remove `placehold.it` image from `fast-foundation`'s card documentation

## Motivation & context
Subresource Integrity isn't a problem if we aren't loading files from external services.

## Issue type checklist

## Process & policy checklist


radium-v:
@awentzel The `site-utilities` package is where the badge generator script lives but the build-static script in `website` copies them to `static/`. The `website` project has to collect these files and ensure that they exist in `static/` so that Docusaurus can find them when it builds. Since `static/` is gitignored, the static files have to be generated in, or copied to, `static/`.

Here's an overview - in `sites/website`, the `yarn build` command:
1. runs `yarn workspace @microsoft/site-utilities generate:badges` to generate the badges
2. runs `yarn build-static`, which:
    - mutates the `fast-website` webpack config to output to the `static/` directory, then builds for production
    - **copies files from `site-utilities`, (favicon, docs site logo, and the previously-generated badges) to `static/`**
    - copies `fast-components.iife.min.js` from the `@microsoft/fast-components` to `static/` for docusaurus
3. runs `yarn prepare-docs`, which:
    - copies and modifies all of the markdown files from the main packages as well as the api-extractor
    - mutates URLs pointing to `badge.fury.io` and `img.shields.io` to point to files in the `static/` directory
4. runs `docusaurus build`
    - outputs its files to `build/`
    - copies everything present in `static/` to `build/`

(Side note: most path traversal between packages is handled by `require.resolve` to avoid having a bunch of `../` everywhere.)

radium-v:
@awentzel I realize now that I went on a semi-unrelated tangent with my previous response - good morning!

I'm not familiar with how we currently handle files on the CDN layer for the website. Updating the PR to point to them on `static.fast.design` instead of hosting them with the bundle may only require:
- make `site-utilities` aware of the icons it's already generating, so they can be passed on to `static.fast.design`
- change the regex that modifies the badge URLs in the markdown
- eliminate the `copy` step

I don't know if this is the right move for all of the files since the badges for the "npm package: v.v.v" have to be rebuilt for each version-bumping release, and should be considered generated dynamic files.

awentzel:
@radium-v good morning. The extra tangent was helpful so I appreciate that.  I'll approve this, then we can determine at later time implications for moving to use CDN.  Essentially, all websites should request all assets from the CDN at `https://static.fast.design`.  The bundles are not necessary, because Azure Front Door will act as a load balancer with caching of its own built-in.  Thus while the entire site does not run on a CDN, all resources are cached across different services.

sites/website/src/prepare-docs.js
```diff
@@ -48,77 +61,53 @@ function identifyPackage(path) {
...
+        const dest = path.join(__dirname, "../docs", docsFolderDest, filename);
 
         await safeCopy(source, dest);
     }
 }
 
 async function copyArticleMarkdown() {
     await moveMarkdownFiles(
-        "../../packages/web-components/fast-foundation/docs/integrations",
+        path.resolve(fastFoundation, "docs/integrations"),
```
EisenbergEffect, line 92: Yeah...seems like I forgot about `path.resolve` when originally whipping some of this up.... 🤣

Issue #3659, opened 2020-08-10
add https support for json-schema validation


awentzel:
**What feature are you requesting?**
The AJV Validation code should be upgraded to support https instead of http moving forward.  Http was converted everywhere to use https in the source code for consistency and make sure not external linked to http (less secure) is being done.  Somewhere on ajv-validation.spec.ts support is failing when changing from http to https and this should pass testing.

**What would be the benefit of adding this feature?**
Better security support for HTTPS to bring into compliance https://webscout/#/scans/757563/issues?issueId=3692645 

**What solution would be ideal for you?**
convert all usage of http to https

jackrmomo:
package.json

awentzel:
I tried converting all schemas to use `https` and there was tooling issue that was not compatible.  I believe, if this is needed for schema's, @janechu has built the tool and perhaps can provide further context around if this is needed for schemas and what areas of the project are impacted.

robarbms:
AJV currently does not support schemas using https.

https://github.com/ajv-validator/ajv/issues/1104

janechu:
@robarbms yes, thanks for verifying, @awentzel this is why we cannot change that string.

janechu:
Closing this in favor of #3795

PR #3789, opened 2020-08-26
chore: add https support for json-schema validation

robarbms:
chore: add https support for json-schema validation

# Description
AJV package does not support https. Created a wrapper function that will resolve https schemas as http if they fail. This will allow us to use https schemas in the interim until the AJV packages is updated to support https. Updated all json shcemas to use https.

## Motivation & context

Better security support for HTTPS to bring into compliance https://webscout/#/scans/757563/issues?issueId=3692645

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/tooling/fast-tooling-react/README.md
+++ b/packages/tooling/fast-tooling-react/README.md
@@ -311,7 +311,7 @@ Form validation uses the [ajv](https://github.com/epoberezkin/ajv) package. The 
 
 ### JSON schema metadata
 
-The schema form generator can interpret most [JSON schemas](http://json-schema.org/), however there are some things to note when writing JSON schemas that make for a better UI.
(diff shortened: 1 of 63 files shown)


janechu:
Unfortunately I think there is some miscommunication as the issue this is referencing (#3659) is attempting to fix something that isn't broken. The code scanner is mistakenly picking up this valid URI and assuming it should be https, but as it's used as an identifier and is part of the JSON schema spec itself, all of these http links are correct. I believe we should hold on this as even draft 7 of the JSON schema spec uses http - http://json-schema.org/draft-07/schema#. However, there may be a change coming with the next draft as it *does* use https - https://json-schema.org/draft/2019-09/schema but more work needs to be done before we can pin to that version.

I've filed an issue to look into it #3795

PR #3792, opened 2020-08-26
fix: move focus ring to button host to allow padding

### LINKED ISSUE #3790: Button has odd focus behavior when adding padding

The above SS has padding applied to the `fast-button` host. This causes the focus rect to draw inside the box of the host because focus is *actually* on an internal button element. We should move the focus rect to the host element and use `:focus-within` to account for this.

CuddleBunny:
Moved the focus ring to the button/anchor host to allow for padding. Resolves #3790 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/styles/patterns/button.ts
+++ b/packages/web-components/fast-components/src/styles/patterns/button.ts
@@ -39,6 +39,7 @@ export const BaseButtonStyles = css`
         min-width: calc(${heightNumber} * 1px);
         background-color: ${neutralFillRestBehavior.var};
         color: ${neutralForegroundRestBehavior.var};
+        border: calc(var(--outline-width) * 1px) solid transparent;
         border-radius: calc(var(--corner-radius) * 1px);
         fill: currentColor;
         cursor: pointer;
@@ -56,8 +57,8 @@ export const BaseButtonStyles = css`
         white-space: nowrap;
         outline: none;
         text-decoration: none;
-        border: calc(var(--outline-width) * 1px) solid transparent;
         color: inherit;
+        border: none;
         border-radius: inherit;
         fill: inherit;
         cursor: inherit;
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -71,11 +72,10 @@ export const BaseButtonStyles = css`
         background-color: ${neutralFillActiveBehavior.var};
     }
 
-    .control:${focusVisible} {
```
nicholasrice, line 74: Doh - I just realized this is going to break `:focus-visible` when that rolls in. :focus-visible is not out from behind a flag yet so I *think* this is okay for now but we should get other folk's opinion on that. @EisenbergEffect @chrisdholt

EisenbergEffect: Hmm. Assuming we had focus visible, what would be the alternative fix? If we can plan for forward compat, that's definitely preferred.

CuddleBunny: Propose `:focus-visible-within` be added to the spec? 🙃 I can't think of a pure CSS way to keep `:focus-visible`. Does fast have a mechanism for forwarding styles to child elements?

nicholasrice: [It's been discussed!]( https://github.com/WICG/focus-visible/issues/151)

I can't either... You can forward styles to shadowed element has a [part attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/part), which button does.

chrisdholt: @nicholasrice @CuddleBunny `:focus-visible` is shipping in Chrome 86 which is currently in Beta: https://twitter.com/ChromiumDev/status/1301262894617640960

This will cover both Chrome and I'm checking on if it will also be in "Edgium" (h/t @CuddleBunny) for 86...

CuddleBunny: Perhaps this should just be closed as a non-issue and documented that button padding can be controlled by targeting the control part? At this point it is down to developer experience vs accessibility. The latter is probably more important.

nicholasrice: @CuddleBunny Unfortunately I think that is the best course for now. Thanks for looking at this though!

PR #3814, opened 2020-08-28
feat: add a shortcut utility

janechu:
# Description

Creates a shortcut utility and shortcut action utility for describing shortcuts and posting custom events through the `MessageSystem` to allow these shortcut events to be attached to DOM nodes.

For full spec see #3781 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
+++ b/packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
@@ -1,0 +1,107 @@
+export interface MetaKey {
+    metaKey: true;
+}
+
+export interface ShiftKey {
+    shiftKey: true;
+}
+
+export interface CtrlKey {
+    ctrlKey: true;
+}
+
+export interface AltKey {
+    altKey: true;
+}
+
+export type ModifierKey = AltKey | CtrlKey | ShiftKey | MetaKey;
+
+export interface SpecificKey {
+    /**
+     * The key string value
+     */
+    value: string;
+}
+
+export type KeyConfig = SpecificKey | ModifierKey;
+
+export interface ShortcutActionConfig {
+    /**
+     * The display name of the shortcut action
+     */
+    name: string;
+
+    /**
+     * The keys needed to execute the action
+     */
+    keys: KeyConfig[];
+
+    /**
+     * The action to take when the keycodes have been pressed
+     */
+    action: () => void;
+}
+
+export class ShortcutAction {
+    private action: () => void;
+    public keys: KeyConfig[];
+    public name: string;
+
+    constructor(config: ShortcutActionConfig) {
+        this.action = config.action;
+        this.keys = config.keys;
+        this.name = config.name;
+    }
+
+    public runAction = (keys: KeyConfig[]): void => {
+        if (this.shouldRunAction(keys)) {
+            this.action();
+        }
+    };
+
+    private shouldRunAction = (keys: KeyConfig[]): boolean => {
+        let matchesKeys = true;
+        const keysLength = this.keys.length;
+
+        if (keys.length === keysLength) {
+            for (let i = 0; i < keysLength; i++) {
+                if (
+                    keys.findIndex((suppliedKey: KeyConfig) => {
+                        // determine if this matches a specified keycode
+                        if (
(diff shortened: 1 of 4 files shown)


packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.name = config.name;
+    }
+
+    public runAction = (keys: KeyConfig[]): void => {
+        if (this.shouldRunAction(keys)) {
+            this.action();
+        }
+    };
+
+    private shouldRunAction = (keys: KeyConfig[]): boolean => {
```
EisenbergEffect, line 62: Practically, this will almost always be run in response to a keyboard event. Per my comment on the spec, I'd expect an API that took that event object as input. If I'm reading this right, with this API, we'd have to have an additional translation layer to translate the keyboard event into an array of KeyConfig objects before this API could be called.

As a consumer, I think I'd like to have something like this:

```js
if (action.matches(event)) {
  action.invoke();
}
```

packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
```diff
@@ -0,0 +1,107 @@
...
+                                (this.keys[i] as SpecificKey).value
+                        ) {
+                            return true;
+                            // determine if this matches a modifier keycode
+                        } else if (
+                            typeof (suppliedKey as SpecificKey).value === "undefined" &&
+                            typeof (this.keys[i] as SpecificKey).value === "undefined"
+                        ) {
+                            return (
+                                ((this.keys[i] as MetaKey).metaKey &&
```
nicholasrice, line 84: I don't think this will correctly handle a `KeyConfig` that expects *two* or more modifiers, will it? Is that a case we want to handle? For instance ctrl+shift+some-key

packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
```diff
@@ -0,0 +1,107 @@
...
+export type ModifierKey = AltKey | CtrlKey | ShiftKey | MetaKey;
+
+export interface SpecificKey {
+    /**
+     * The key string value
+     */
+    value: string;
+}
+
+export type KeyConfig = SpecificKey | ModifierKey;
```
nicholasrice, line 26: With this typing, a `KeyConfig` with only a modifier key is valid - is that accurate?

```
const foo: KeyConfig = { // Totally cool
    shiftKey: true
}
```

packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * The keys needed to execute the action
+     */
+    keys: KeyConfig[];
+
+    /**
+     * The action to take when the keycodes have been pressed
+     */
+    action: () => void;
```
nicholasrice, line 42: You might consider supplying the action with the config. Doing so would allow authors more flexibility in how they provide callbacks to the system because they can perform delegation in their own code if they prefer to not supply unique functions to each action config.

packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
janechu: Yes, that's correct.

packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
janechu: It is a case we want to handle, and it will, all keys must match, this should loop through all of them to check.

packages/tooling/fast-tooling/src/data-utilities/shortcut-action.ts
janechu: Whoops, so it seems like I misinterpreted what `|` does, I am re-doing the typescript to use exclusionary interfaces so that they can't be combined.

nicholasrice (approved):

EisenbergEffect (approved):
Nice!

PR #3837, opened 2020-09-01
fix: update the padding to start and end slot on tabs to add align self

khamudom:
# Description

This change is an update to the previous work, which was to align the start and end slot to the tab list. With further discussion on finding the best approach, it was decided that we should not change the padding on the slot but to just align itself to the center.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/tabs/tabs.styles.ts
+++ b/packages/web-components/fast-components/src/tabs/tabs.styles.ts
@@ -29,13 +29,8 @@ export const TabsStyles = css`
         box-sizing: border-box;
     }
 
-    .start {
-        padding: 2px;
-        align-self: center;
-    }
-
+    .start,
     .end {
-        padding: 2px;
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-components/src/tabs/tabs.styles.ts
```diff
@@ -30,11 +30,13 @@ export const TabsStyles = css`
     }
 
     .start {
```
nicholasrice, line 32: You can combine these selectors to reduce this:

```
.start, .end {
   padding: 2px;
   align-self: center;
}
```

khamudom: you are correct. thanks!!

khamudom: combined and also removed the padding

PR #3861, opened 2020-09-08
fix: implicit form-associated submission

### LINKED ISSUE #3860: implicit form submission not matching native implementation
Despite this comment `// Match native behavior` (1) on the form associated implementation, the form implicit submission is far from the spec (2).

The form is implicitely submited when the enter key is pressed, no mater if the form is valid, and that's just one of the problem,.
We should find the first [type=submit] and trigger a click event (3). That may not be the exact behavior of all browsers but that whould be way better then the current behavior.

(1) https://github.com/microsoft/fast/blob/78787c35631edb2b86f985dbfbc68036797f60d0/packages/web-components/fast-foundation/src/form-associated/form-associated.ts#L492
(2) https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-algorithm
(3) https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission

dcharbonnier:
close #3860

### DIFF
--- a/packages/web-components/fast-foundation/src/form-associated/form-associated.ts
+++ b/packages/web-components/fast-foundation/src/form-associated/form-associated.ts
@@ -491,7 +491,7 @@ export abstract class FormAssociated<
             case keyCodeEnter:
                 if (this.form instanceof HTMLFormElement) {
                     // Implicit submission
-                    const defaultButton = this.form.querySelector("[type=submit]") as HTMLElement;
+                    const defaultButton = this.form.querySelector("[type=submit]") as (HTMLElement | null);
                     defaultButton?.click();
                 }
                 break;
--- a/packages/web-components/fast-foundation/src/form-associated/form-associated.ts
+++ b/packages/web-components/fast-foundation/src/form-associated/form-associated.ts
@@ -491,7 +491,9 @@ export abstract class FormAssociated<
             case keyCodeEnter:
                 if (this.form instanceof HTMLFormElement) {
                     // Implicit submission
-                    const defaultButton = this.form.querySelector("[type=submit]") as (HTMLElement | null);
+                    const defaultButton = this.form.querySelector(
+                        "[type=submit]"
+                    ) as HTMLElement | null;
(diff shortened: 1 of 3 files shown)


EisenbergEffect (approved):

packages/web-components/fast-foundation/src/form-associated/form-associated.ts
```diff
@@ -489,10 +490,10 @@ export abstract class FormAssociated<
         switch (e.keyCode) {
             case keyCodeEnter:
                 if (this.form instanceof HTMLFormElement) {
-                    // Match native behavior
-                    this.form.submit();
+                    // Implicit submission
+                    const defaultButton = this.form.querySelector("[type=submit]") as HTMLElement;
```
nicholasrice, line 494:
```suggestion
                    const defaultButton = this.form.querySelector("[type=submit]") as (HTMLElement | null);
```

nicholasrice: This should fix the build error.

nicholasrice (approved):

packages/web-components/fast-foundation/src/form-associated/form-associated.ts
```diff
@@ -489,10 +490,10 @@ export abstract class FormAssociated<
         switch (e.keyCode) {
             case keyCodeEnter:
                 if (this.form instanceof HTMLFormElement) {
-                    // Match native behavior
-                    this.form.submit();
+                    // Implicit submission
+                    const defaultButton = this.form.querySelector("[type=submit]") as (HTMLElement | null);
```
nicholasrice, line 494:
```suggestion
                    const defaultButton = this.form.querySelector(
                        "[type=submit]"
                    ) as HTMLElement | null;
```

nicholasrice: Prettier didn't like this being on one line.

awentzel (approved):

PR #3866, opened 2020-09-08
refactor: extract an abstraction of the registered action and utilities

janechu:
# Description

Due to the increase in utilities that will register with the message system, an abstraction is needed to keep the API consistent and also consolidate common logic.

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tooling/src/data-utilities/registered-action.ts
```diff
@@ -0,0 +1,44 @@
...
+    private action: (config: C | RegisteredActionCallbackConfig) => void;
+    public id: string;
+
+    constructor(config: RegisteredActionConfig) {
+        this.id = config.id;
+        this.action = config.action;
+    }
+
+    /**
+     * Invokes the action
```
nicholasrice, line 29: Is this comment accurate?

awentzel (approved):

packages/tooling/fast-tooling/src/data-utilities/registered-action.ts
janechu: Whoops, nope, that was a leftover.

packages/tooling/fast-tooling/src/data-utilities/registered-action.ts
```diff
@@ -0,0 +1,44 @@
...
+     */
+    action: () => void;
+}
+
+/**
+ * This abstract class is for actions intended to be part
+ * of a registered class
+ */
+export abstract class RegisteredAction<C, M> {
+    private action: (config: C | RegisteredActionCallbackConfig) => void;
```
nicholasrice, line 20: Is `C` truly generic (could be any type, `null` for instance) or should it always be of at *least* type `RegisteredActionCallbackConfig`?

janechu: `C` is not truly generic, it really should be any additional interface that can be combined with `RegisteredActionCalbackConfig`, so `null` would be erroneous. What solution would be best?

nicholasrice: I think you want somthing like the following then:

```
export abstract class RegisteredAction<C extends RegisteredActionCallbackConfig, M> {
  private action: (config: C) => void;
}
```
And maybe something similar for `M`? I'm not sure...

packages/tooling/fast-tooling/src/data-utilities/registered-action.ts
```diff
@@ -0,0 +1,44 @@
+export interface RegisteredActionCallbackConfig {
+    /**
+     * The unique identifier
+     */
+    id: string;
+}
+
+export interface RegisteredActionConfig extends RegisteredActionCallbackConfig {
```
nicholasrice, line 8: Is there a name other than "Registered" that would work here? "Registered" seems more like the state of a thing relative to another thing than the thing itself.

packages/tooling/fast-tooling/src/data-utilities/registered-utility.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    /**
+     * The utility should always handle incoming messages
+     * from the message system
+     */
+    abstract handleMessageSystem(e: MessageEvent): void;
+
+    /**
```
nicholasrice, line 73: Is this comment accurate? It doesn't seem to run the action

nicholasrice (approved):

packages/tooling/fast-tooling/src/data-utilities/registered-action.ts
janechu: Hmm, going to have to take a further look at this... That doesn't quite work because I'm using an exclusionary version to execute on the child class.

janechu: I'm using the same method react uses and indicating that these generics should be objects with <C = {}, M = {}>, there wasn't a good way to extrapolate this out since it's not used as an extension of another interface.

packages/tooling/fast-tooling/src/data-utilities/message-system-utility-action.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * This abstract class is for actions intended to be part
+ * of a registered class
+ *
+ * It takes two generics, C, an object which is added to the
+ * MessageSystemUtilityActionCallbackConfig,
+ * and M, a generic used for matching
+ * the action to some specific parameters
+ */
+export abstract class MessageSystemUtilityAction<C = {}, M = {}> {
```
EisenbergEffect, line 25: I haven't been totally consistent about this in the past, but looking at this I'm reminded of how I've sometimes named generic parameters to make them clearer. I'll sometimes use a T prefix and then provide a meaningful name. So in this case, you might have `TConfig` and `TMatchable` or something like that.

packages/tooling/fast-tooling/src/data-utilities/type.utilities.ts
```diff
@@ -0,0 +1,4 @@
+type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
+export type XOR<T, U> = T | U extends object
```
EisenbergEffect, line 2: Clever 😉

EisenbergEffect (approved):

PR #3876, opened 2020-09-10
feat:  Anchored region reliability/perf improvements

scomea:
# Description
The current version of anchored region falls back to call to expensive getBoundingClientRect calls in circumstances when the provided viewport/anchor pair doesn't "work" with intersection observer as can happen with intervening "position: fixed" elements in the dom hierarchy.  I recently discovered that intersection observer seems to always work at returning an initial event with valid entries for any element when the viewport is null, so I modified anchored region to take advantage of that for reliability and performance.

The cost of this change is that anchored region no longer attempts to watch for intersection events to update its position, but instead with only update itself when prompted by a call to its "update()" function, when an attribute or property is changed, or when the anchor element or the component itself is resized.  

This hopefully gets us out of having to explain why the component behaves differently based on the local DOM structure  (many devs are not very familiar with the intricacies of intersection observer api).

## Motivation & context
Anchored region performance and reliability.

## Issue type checklist

## Process & policy checklist

<!-

### DIFF
--- a/packages/web-components/fast-components/src/anchored-region/anchored-region.stories.ts
+++ b/packages/web-components/fast-components/src/anchored-region/anchored-region.stories.ts
@@ -27,8 +27,8 @@ addons.getChannel().addListener(STORY_RENDERED, (name: string) => {
         const scalingViewportOffset: HTMLElement | null = document.getElementById(
             "viewport-scaling-offset"
         );
-        if (scalingViewportUpdate !== null) {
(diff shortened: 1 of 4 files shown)


EisenbergEffect (approved):

marjonlynch (approved):
Smoked in Edge and Firefox, assuming the known issue on loading in Firefox is understood then looks good to me.

PR #3877, opened 2020-09-10
feat: add skeleton component

litteredwitherrors:
# Description

## Motivation & context
Skeleton component is used as a placeholder for components while in a loading state.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/skeleton/index.ts
```diff
@@ -0,0 +1,12 @@
+import { customElement } from "@microsoft/fast-element";
+import { Skeleton, SkeletonTemplate as template } from "@microsoft/fast-foundation";
+import { SkeletonStyles as styles } from "./skeleton.styles";
+
+@customElement({
+  name: "fluent-skeleton",
```
scomea, line 6: Shouldn't it be "fast-skeleton" in this directory?

packages/web-components/fast-foundation/src/skeleton/skeleton.ts
```diff
@@ -0,0 +1,39 @@
...
+  /**
+   * Indicates what the shape of the Skeleton should be.
+   * 
+   * @public
+   * @remarks
+   * HTML Attribute: shape
+   */
+  @attr public shape: skeletonShape = "rect";
+
+  @attr public ariaBusy: boolean;
```
scomea, line 29: public attributes and properties need remarks

packages/web-components/fast-foundation/src/skeleton/skeleton.template.ts
```diff
@@ -0,0 +1,13 @@
+import { html } from "@microsoft/fast-element";
+import { Skeleton } from "./skeleton";
+
+export const SkeletonTemplate = html<Skeleton>`
+  <template
+    aria-busy="${x => x.ariaBusy}"
+    shape="${x => x.shape}"
+    class="${x => (x.shape === "circle" ? "circle" : "rect")}"
```
scomea, line 8: Rather than set the class on the template you could also manipulate the classlist directly in skeleton.ts.  For example this.classlist.toggle("myClass", this.showMyClass === true)

chrisdholt: @nicholasrice  are we still looking to do this in the template when possible?

packages/web-components/fast-components/src/skeleton/index.ts
chrisdholt: Yeah, this will get pulled and moved to Fluent once we've had a solid review of it. I had @litteredwitherrors move this out of Draft because it wasn't getting any looks 😄

packages/web-components/fast-foundation/src/skeleton/skeleton.ts
```diff
@@ -0,0 +1,37 @@
+import { attr, FASTElement } from "@microsoft/fast-element";
+
+export type skeletonShape = 'rect' | 'circle';
```
EisenbergEffect, line 3: Let's make this `SkeletonShape` for consistency with other custom types.

packages/web-components/fast-foundation/src/skeleton/skeleton.template.ts
```diff
@@ -0,0 +1,18 @@
+import { html } from "@microsoft/fast-element";
+import { Skeleton } from "./skeleton";
+
+export const SkeletonTemplate = html<Skeleton>`
+  <template
+    shape="${x => x.shape}"
```
EisenbergEffect, line 6: You should be able to remove the `shape` binding since `template` is just a stand-in for the element host itself in this case. Since the element has the shape attr, it doesn't need the binding.

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+    position: relative;
+    background-color: ${neutralFillRestBehavior.var};
+    --skeleton-animation-gradient-default: linear-gradient(270deg, ${neutralFillRestBehavior.var} 0%, #F3F2F1 51.13%, ${neutralFillRestBehavior.var} 100%);
+  }
+
+  :host(.rect) {
+    border-radius: calc(var(--corner-radius) * 1px);
+  }
+
+  :host(.circle) {
```
EisenbergEffect, line 20: @nicholasrice Should we use attribute selectors for this? `:host[shape=circle]` instead of making it class based? Will that help out React scenarios?

packages/web-components/fast-components/src/skeleton/fixtures/base.html
```diff
@@ -0,0 +1,140 @@
...
+          <fluent-skeleton class="text-bone" shape="rect"></fluent-skeleton>
+          <fluent-skeleton class="button-bone" shape="rect"></fluent-skeleton>
+        </div>
+  
+      </div>
+
+      <h3>Used as element blocks with shimmer effect element</h3>
+      <div class="card-placeholder">
+        <fluent-skeleton class="avatar-bone" shape="circle">
+          <span class="shimmer"></span>
```
EisenbergEffect, line 94: I wonder if this should be done as a boolean attr `shimmer` instead of a slotted child. Based on the boolean attr we could conditionally render the element in shadow dom, for example. You would have an API like this:

```html
 <fluent-skeleton shape="circle" shimmer> </fluent-skeleton>
```

EisenbergEffect (changes requested):

packages/web-components/fast-components/src/skeleton/fixtures/base.html
litteredwitherrors: I like this, it makes sense. Will implement.

packages/web-components/fast-components/src/skeleton/fixtures/base.html
```diff
@@ -0,0 +1,169 @@
...
+            shape="rect"
+        ></fluent-skeleton>
+    </div>
+
+    <h4>Used as element blocks with shimmer effect element</h4>
+    <div class="card-placeholder">
+        <fluent-skeleton
+            style="
+                --neutral-fill-rest: #e1dfdd;
+                border-radius: 4px;
```
chrisdholt, line 69: should this be part of the control itself as a default opted into the design system?

packages/web-components/fast-components/src/skeleton/fixtures/base.html
```diff
@@ -0,0 +1,169 @@
...
+            "
+            shape="rect"
+        ></fluent-skeleton>
+    </div>
+
+    <h4>Used as element blocks with shimmer effect element</h4>
+    <div class="card-placeholder">
+        <fluent-skeleton
+            style="
+                --neutral-fill-rest: #e1dfdd;
```
chrisdholt, line 68: Is this contextual? I see it repeated throughout here. If it's the common pattern, perhaps it should be the default?

packages/web-components/fast-components/src/skeleton/fixtures/base.html
litteredwitherrors: Agreed, it should be. Will revise.

packages/web-components/fast-components/src/skeleton/fixtures/base.html
litteredwitherrors: You're right, it looks like it should be a default. I'll revise.

packages/web-components/fast-components/src/skeleton/fixtures/base.html
litteredwitherrors: Upon further investigation, the border-radius is controlled by the `rect` class. This is just an inline-style override.

packages/web-components/fast-components/src/skeleton/fixtures/base.html
```diff
@@ -0,0 +1,169 @@
...
+    <h4>Using SVG via Pattern attribute</h4>
+    <fluent-skeleton
+        style="
+            --neutral-fill-rest: #e1dfdd;
+            border-radius: 4px;
+            width: 500px;
+            height: 250px;
+        "
+        shape="rect"
+        pattern="https://static.fast.design/assets/skeleton-test-pattern.svg"
```
radium-v, line 82: Referencing this soon-to-be-CDN-hosted SVG means the fixture relies on an internet connection to properly execute. Can we include the file next to this fixture file and use a relative URL here?

packages/web-components/fast-components/src/skeleton/skeleton.definitions.ts
```diff
@@ -0,0 +1,48 @@
...
+                },
+                {
+                    name: "shape",
+                    description: "The shape attribute",
+                    type: DataType.string,
+                    default: "circle",
+                    required: false,
+                },
+                {
+                    name: "pattern",
```
radium-v, line 26: The `pattern` attribute already exists on some HTML elements for regular expression pattern matching (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern).

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
```diff
@@ -0,0 +1,80 @@
...
+        animation-timing-function: var(
+            --skeleton-animation-timing,
+            var(--skeleton-timing-default)
+        );
+        animation-direction: normal;
+        z-index: 1;
+    }
+
+    ::slotted(svg) {
+        z-index: 9999;
```
radium-v, line 64: z-index too high

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
```diff
@@ -0,0 +1,80 @@
...
+    :host(.circle) {
+        border-radius: 100%;
+        overflow: hidden;
+    }
+
+    object {
+        position: absolute;
+        width: 100%;
+        height: auto;
+        z-index: 1000;
```
radium-v, line 34: z-index too high

radium-v (changes requested):

packages/web-components/fast-components/src/skeleton/skeleton.definitions.ts
chrisdholt: Would there be a collision here? It's a valid attribute but I'd be curious of @EisenbergEffect's thoughts on if we should mind existing attributes. If this were the text field or input, I think we for sure would have an issue. I'm unsure if this is an issue. Confusion could be a concern here, but I think we're going to end up with less-than-ideal names if we avoid any attrribute any HTML element has already used. Just some thoughts.

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
chrisdholt: Yeah, because this is within the context here, we can use the lowest number possible IMO.

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
chrisdholt: +1 - next highest number please (possibly 2)?

packages/web-components/fast-components/src/skeleton/skeleton.definitions.ts
EisenbergEffect: I don't think it's an issue but we could choose another word if we felt that would help avoid confusion. A few synonyms based on quick search: design · decoration · motif · marking · ornament · ornamentation · device · figure

packages/web-components/fast-components/src/skeleton/fixtures/base.html
litteredwitherrors: Since I'm using the `<object>` element I can set the fallback image source to a local file, but including the file next to the fixture doesn't work. @chrisdholt thoughts?

chrisdholt: @radium-v would you be open to us taking this as a follow-up issue? Agree on it being an issue for long-term dev and the ability to do so without an internet connection. I'd prefer to not block on this and figure out a good method for testing with images. While I'm hesitant to do so, I think we're going to need to and I'd like to have a solid solution in place.

What do you think?

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
litteredwitherrors: I'm going to resolve this for now. We can update again later once a decision is made.

packages/web-components/fast-components/src/skeleton/fixtures/base.html
radium-v: I'll create a ticket to follow up on this.

radium-v (approved):

nicholasrice:
I'm converting this to a draft PR because your PR description designates this PR as a draft. If that is stale feel free to convert back to a real(?) PR and update the description. Thanks!

packages/web-components/fast-foundation/src/skeleton/README.md
```diff
@@ -0,0 +1,94 @@
+---
+id: skeleton
+title: fluent-skeleton
+sidebar_label: skeleton
```
EisenbergEffect, line 4:
```suggestion
sidebar_label: skeleton
custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/src/skeleton/README.md
```

packages/web-components/fast-foundation/src/skeleton/README.md
```diff
@@ -0,0 +1,94 @@
+---
+id: skeleton
+title: fluent-skeleton
```
EisenbergEffect, line 3:
```suggestion
title: fast-skeleton
```

packages/web-components/fast-foundation/src/skeleton/README.md
```diff
@@ -0,0 +1,94 @@
...
+sidebar_label: skeleton
+---
+
+The `fluent-skeleton` component is used as a placeholder for another component that is in a loading state.
```
EisenbergEffect, line 7:
```suggestion
The `fast-skeleton` component is used as a placeholder for another component that is in a loading state.
```

packages/web-components/fast-components/src/skeleton/skeleton.definitions.ts
```diff
@@ -0,0 +1,48 @@
+import { WebComponentDefinition } from "@microsoft/fast-tooling/dist/data-utilities/web-component";
+import { DataType } from "@microsoft/fast-tooling";
+
+export const fastSkeletonDefinitions: WebComponentDefinition = {
+    version: 1,
+    tags: [
+        {
+            name: "fluent-skeleton",
```
EisenbergEffect, line 8:
```suggestion
            name: "fast-skeleton",
```

packages/web-components/fast-components/src/skeleton/index.ts
```diff
@@ -0,0 +1,12 @@
+import { customElement } from "@microsoft/fast-element";
+import { Skeleton, SkeletonTemplate as template } from "@microsoft/fast-foundation";
+import { SkeletonStyles as styles } from "./skeleton.styles";
+
+@customElement({
+    name: "fluent-skeleton",
+    template,
+    styles,
+})
+export class FluentSkeleton extends Skeleton {}
```
EisenbergEffect, line 10:
```suggestion
export class FASTSkeleton extends Skeleton {}
```

EisenbergEffect (changes requested):
Need to change all references from "fluent" to "fast". Also, we need to add this to the sidebars.json file in the website so that it shows up in the docs.

EisenbergEffect:
Actually, I am confused. Are we adding this to FAST components or not? There's an implementation in the fast-components package and there are docs for FAST, but everything is named Fluent...

EisenbergEffect:
Also, I thought this being merged was one of the reasons why we held the release until today...

litteredwitherrors:
@nicholasrice Description is now updated to no longer reflect a draft.

litteredwitherrors:
@EisenbergEffect Everything is now updated to reflect FAST.

EisenbergEffect (approved):

PR #3890, opened 2020-09-11
fix: tree-view right and left arrow keyboarding should behave as wai-aria spec suggests

### LINKED ISSUE #3886: bug: nested tree views has odd arrow key behavior
Using arrow keys to navigate nested tree view has odd behaviors.
- sometimes pressing an arrow key does nothing, then pressing it again moves focus
- if you navigate to an item by pressing the right arrow key you should be able to go back by pressing left arrow key

Go to https://explore.fast.design/components/fast-tree-view
click on tree item 1 to focus on it
press right arrow key, item opens
press right arrow key, focus to nested child
press rigth arrow key, nothing happens (?)
press right arrow key, focus to tree item 2
press left arrow key, nothing happens

Consistent and predictable arrow key navigation

marjonlynch:
# Description
* fixed a bug where hitting right arrow key on leaf node would mess up the next left arrow keydown
* right arrow key expands a node if not expanded, or moves to first child, or does nothing

closes #3886 

## Motivation & context
Expected keyboarding behavior for arrow keys and an example in action
https://w3c.github.io/aria-practices/#TreeView
https://w3c.github.io/aria-practices/examples/treeview/treeview-2/treeview-2a.html

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -235,10 +235,17 @@ export class TreeItem extends FASTElement {
             return;
         }
 
-        if (!this.expanded) {
+        if (!this.expanded && this.childItemLength() > 0) {
             this.setExpanded(true);
         } else {
-            this.focusNextNode(1);
+            if (this.childItems.length > 0) {
+                const childTreeItems: HTMLElement[] = this.childItems.filter(
+                    (item: HTMLElement) => {
+                        return isTreeItemElement(item);
+                    }
+                );
+                childTreeItems[0].focus();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -235,10 +247,12 @@ export class TreeItem extends FASTElement {
             return;
         }
 
-        if (!this.expanded) {
+        if (!this.expanded && this.childItemLength() > 0) {
             this.setExpanded(true);
         } else {
-            this.focusNextNode(1);
+            if (this.childItemLength() > 0) {
+                this.enabledChildTreeItems()[0].focus();
```
EisenbergEffect, line 254: Is it possible that childItemLengty is greater than 0 but that the enabled child tree items is 0?

chrisdholt: I think this is technically possible and a good case to guard against. I think there was an email where I called this out as a noop ;). We should likely account for this. Good catch @eisenbergeffect!

marjonlynch: doh! will fix

nicholasrice (approved):

janechu (approved):

PR #3905, opened 2020-09-15
fix: ensure badge always applies inline-style when attributes exist

chrisdholt:
# Description
The badge component had odd behavior when passing inline styles and there was a scenario you could end up in where nothing would be applied if fill was not passed. This change creates a function to ensure attributes are handled correctly here.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/badge/badge.template.ts
+++ b/packages/web-components/fast-foundation/src/badge/badge.template.ts
@@ -7,14 +7,7 @@ import { Badge } from "./badge";
  */
 export const BadgeTemplate = html<Badge>`
     <template class="${x => (x.circular ? "circular" : "")}">
-        <div
-            class="control"
-            part="control"
-            style="${x =>
-                x.fill || x.color
-                    ? `background-color: var(--badge-fill-${x.fill}); color: var(--badge-color-${x.color})`
-                    : void 0}"
-        >
+        <div class="control" part="control" style="${x => x.generateBadgeStyle()}">
             <slot></slot>
         </div>
     </template>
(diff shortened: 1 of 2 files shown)


scomea (approved):

packages/web-components/fast-foundation/src/badge/badge.ts
```diff
@@ -36,4 +36,21 @@ export class Badge extends FASTElement {
      */
     @attr({ mode: "boolean" })
     public circular: boolean;
+
+    public generateBadgeStyle = () => {
+        const fill: string = `background-color: var(--badge-fill-${this.fill});`;
+        const color: string = `color: var(--badge-color-${this.color});`;
+
+        if (!this.fill && !this.color) {
```
nicholasrice, line 44: You should move this early-exit condition to the top so we don't allocate variables if they aren't needed.

chrisdholt: moved 👍 - thanks!

packages/web-components/fast-foundation/src/badge/badge.template.ts
```diff
@@ -7,14 +7,7 @@ import { Badge } from "./badge";
...
     <template class="${x => (x.circular ? "circular" : "")}">
-        <div
-            class="control"
-            part="control"
-            style="${x =>
-                x.fill || x.color
-                    ? `background-color: var(--badge-fill-${x.fill}); color: var(--badge-color-${x.color})`
-                    : void 0}"
-        >
+        <div class="control" part="control" style="${x => x.generateBadgeStyle()}">
```
nicholasrice, line 10: Does this update if fill or color change?

chrisdholt: I tested by changing color in the DOM and it updated with the expected colors.

EisenbergEffect (approved):

nicholasrice (approved):

PR #3920, opened 2020-09-17
chore: add disclosure spec

hardikpthv:
# Description

This adds a spec for the disclosure web component.
Find the draft version at https://github.com/microsoft/fast/pull/3921

## Motivation & context

It looks like as follow:
image

## Issue type checklist

## Process & policy checklist

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.


packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,98 @@
...
+    -   `toggle()`: to toggle the content
+-   _Events_
+    -   `expanded-changed: CustomEvent` - No custom data.
+
+### Anatomy and Appearance
+
+```html
+<host>
+    <slot name="invoker"></slot>
+    <slot name="content"></slot>
```
EisenbergEffect, line 54: I think we should have the content go into the default slot so we don't get two named slots and no default slot.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,98 @@
...
+-   [Bootstrap](https://getbootstrap.com/docs/4.3/components/collapse/)
+-   [Material UI](https://material-ui.com/api/collapse/)
+
+---
+
+## Design
+
+The disclosure component can easily be extended for customization, for example adding animation to reveal the content or get the height of extra content slot.
+
+-   `showAnimation({ contentNode })` and `hideAnimation({ contentNode })` `async` methods are `protected` and designed to override in the derived class to add animations.
```
EisenbergEffect, line 34: I'd like to learn a bit more about the animation thinking. Could this be done with css instead for example?

EisenbergEffect: For example, if we gave the content a part name and added additional part names dynamically in different states, could a css animation be targeted at the part to transition between open and close?

EisenbergEffect (changes requested):

EisenbergEffect:
Very excited about this community contribution! Thanks @hardikpthv !

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,98 @@
+# Disclosure
+
+## Overview
+
+A disclosure component is the combination of a button and section of content, a button that toggles the visibility of the section. As defined by the W3C:
```
chrisdholt, line 5: Could this instead potentially use the platform elements of `details` and `summary`? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: Yes, I think that's a good idea.

```html
<host>
    <slot name="invoker"></slot>
    <slot></slot>
</host>
```

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: I tried CSS earlier e.g. set `max-height: 100vh`  and apply the desired transition or animation effect for the extra content toggling, however, `100vh` would consider complete viewport height and extra content size is always going to be varied (could be `100vh` or `10px` too). Therefore, while animating the consideration of `100vh` is better to be avoided and calculate the actual height of the content.

In addition, when the transition is in the process there can be a control to disable the button or wait for the first transition to be finished, etc than queueing.

The `fast-foundation` component would just do toggling of the extra content and the `fast-component` or any custom component that extends the foundation component would take care of the animation.

Because of the height calculation being done dynamically, I am not sure if we could target the CSS animation at the part to transition between open and close.🤔

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: That's a good idea too. It would require style changes to override the native look and feel. I thought of `details` and `summary`, however, the support for IE is missing.

- https://caniuse.com/details
- https://caniuse.com/summary

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: Done :)

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
chrisdholt: IE isn't in our matrix and truthfully to leverage the components and styling themselves the existing polyfills won't work. Leaving aside IE, are most of the styling issues something that can be overcome? For near all of our web components, we're fully composing new styles.

I know that one consideration could be the details "arrow". Webkit browsers should support styling of this using the `-webkit-details-marker`. I'm not sure if there's support for Firefox to style/hide that - so that's one thing to consider.

The only other question would be a11y support across browsers. Are you open to doing a quick investigation on the styling bit and whether or not the keyboard interaction matches that proposed by the W3C across major browsers? I'd prefer to use the platform here if there isn't a good reason to recreate it; but if there isn't support in these two areas, I think we can move forward from the idea for now.

hardikpthv: Okay, I think it should be fine if we don't consider IE.

Yes, I will investigate on Styling and A11Y support across all the browsers.

hardikpthv: Looks like for styling and A11Y:
- **`summary` should act as a button:** custom implementation
- **`enter` and `space` buttons to toggle:** natively supported in `summary`
- **`list-style-type: none; ` should hide marker in FF:** custom implementation
-  **`details`'s height should be dynamic:** custom implementation

If we go with the `details` and `summary` approach, I think we don't need an invoker slot. `title` attribute should be added for invoker title i.e.` summary` text. @EisenbergEffect, we can also handle animation at CSS level I believe.

We can two variants for `summary` style, 1) as button 2) as link/hypertext

Could you confirm the above specs? After that, I will update both the PRs.

EisenbergEffect: We could add a `title` slot that falls back to a `title` attribute. That would make setting a basic title very simple but allow flexibility to slot in custom HTML for the title as well.

```html
<slot name="title">${x => x.title}</slot>
```

Great to hear we could use CSS for the animation.
@chrisdholt may have thoughts on the other items.

chrisdholt: It sounds like details/summary may be a good path forward here based on the research. Can you update the spec with the above considered and I'll take another pass at it! I think this will move us in a great direction. Thanks!

hardikpthv: I have updated the specs. Please go through it once and do let me know for further changes if any.

hardikpthv: I have updated the specs, @chrisdholt. Please go through it once and do let me know for further changes if any.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,119 @@
...
```
chrisdholt, line 5: Nit on this intro - with the updated spec to follow the disclosure pattern, does this still make sense?

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,119 @@
...
+
+-   _Component Name:_ `fast-disclosure`
+-   _Props/Attrs:_
+    -   `expanded: boolean` - Current state of the disclosure component
+-   _Methods_:
+    -   `show()`: to show the content
+    -   `hide()`: to hide the content
+    -   `toggle()`: to toggle the content
+-   _Events_
+    -   `toggle: CustomEvent` - No custom data.
```
chrisdholt, line 47: Would toggle be the event here, or does a generic change event make more sense?

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,119 @@
...
+The disclosure component can easily be extended for customization, for example adding animation to reveal the content or get the height of extra content slot.
+
+-   Basic component which toggles the extra content and to add animation extra styles can be applied to derived/extended component.
+-   To have further control to the extra content, there are also `contentNode` and `contentHeight` available as getters.
+
+### API
+
+-   _Component Name:_ `fast-disclosure`
+-   _Props/Attrs:_
+    -   `expanded: boolean` - Current state of the disclosure component
```
chrisdholt, line 41: You call out `contentNode` and `contentHeight` above - are those potential properties that should be listed here?

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,119 @@
...
+    <details>
+        <summary>
+            <slot name="title"></slot>
+        </summary>
+        <div>
+            <slot></slot>
+        </div>
+    </details>
+</host>
+```
```
chrisdholt, line 62: Would it make sense to include start/end content in this control?

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
```diff
@@ -0,0 +1,119 @@
...
+        Green Arrow is a fictional superhero who appears in comic books published by DC
+        Comics. Created by Mort Weisinger and designed by George Papp, he first appeared
+        in More Fun Comics #73 in November 1941. His real name is Oliver Jonas Queen, a
+        wealthy businessman and owner of Queen Industries who is also a well-known
+        celebrity in Star City.
+    </div>
+</fast-disclosure>
+```
+
+Invoker as hypertext
```
chrisdholt, line 87: Can you elaborate on the purpose of this or the intent here? This seems purely visual. I would expect that we would have a single implementation due to functionality. Visually, you could have different appearances - is that the intent?

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: I think the `start` slot can be useful for example adding an icon or something. I am not sure what `end` will be used for though.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: Either one is fine I believe. However, the native details/summary element has a `toggle` event.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: For the extended component, I think it would be nice to know about these properties because it will be used to find out the dynamic height after the component is expanded.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: Indeed, it required some changes. Updated. :)

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: Yes, that's the idea. For example, using it for a blog or description where "Read more" would like a link or button.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
chrisdholt: Can you elaborate on "extended" component? If it's not being added or needed here, we can simply drop from the spec and it would be added in a component which extends. Alternatively, if you think it's valuable for folks as an API, let's consider adding methods which returns these values below.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
chrisdholt: If the native event is `toggle` let's stay with that.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
chrisdholt: If we add one, let's add both. I could see someone wanting to add a glyph in the start slot and then have a glyph for toggle state of the control in the end slot.

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
chrisdholt: Gotcha, in this case I don't think we need to call it out in the functional spec as part of the implementation. I think for the FAST implementation, if we wanted to support both appearances, we would do that in the `fast-components` implementation rather than `fast-foundation`. If you take a look at anchor, button, badge - those all have different appearances being applied "outside" the foundational control. The idea is that the semantics used for appearance, as well as the variations may vary by design system. Some may have many, some may have none. In your example above, you call out "as-button" or "as-hypertext". We may use appearances which align with existing semantics such as those in button (stealth, justified, etc).

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: 👍

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: Sounds cool. I will add "neutral" | "hypertext" appearances then. 😅

chrisdholt: @hardikpthv FWIW, I don't know that we need to call that out in the spec explicitly. The only place I might is under "non-goals". Essentially something like:

"Many times disclosures may take different appearances, such as a button or anchor. Inclusion of such attributes should be applied at the design system / component implementation level".

hardikpthv: Ah, yes, I got you. I will add it under "Non-goals".

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.md
hardikpthv: I think `contentNode` or `contentHeight` would not be completely accurate naming if we want to keep them. In the component implementation, what we need is the `<details>` itself and as we extend it from the foundation component we already have the access to it. Thus, it's not needed after all. 😅

chrisdholt (approved):
This looks great, thanks @hardikpthv! @EisenbergEffect to take another pass on this when he has a moment :)

chrisdholt:
@hardikpthv - one thing I noticed that's missing. Can you add this alphabetically to the README.md in the specs/ file and put a check mark in the spec column? That should be the last thing needed here :)

hardikpthv:
Thanks, @chrisdholt. Added! I will start working on the changes based on spec on my existing PR. :)

EisenbergEffect (approved):
Looks great 🎊 One addition I would recommend in terms of ergonomics. Give the disclosure a `title` attribute and render the title into the title slot as fallback content for the slot. This allows setting a simple attribute for the title but if the consumer wants custom html then they can always provide the slotted content.

hardikpthv:
Thanks for the review, @EisenbergEffect. Added `title` attribute information as well. 😄

EisenbergEffect:
@chrisdholt Shall we merge? 😄

PR #3921, opened 2020-09-17
feat: add disclosure component

### LINKED ISSUE #3731: Add collapsible component
collapsible component

This could be used to show or hide extra information on button click. It's similar to accordion but independent of the header where the header is button itself an invoker.

I can create this component and submit the MR,

None yet.

**Attach screenshots or images add detail to your request:**
To give an idea:
image

hardikpthv:
# Description

Adds disclosure component based on the specification from https://github.com/microsoft/fast/pull/3920

I will create separate PR for the explorer component changes. If it has to be a part of this PR, I can add it here as well.

## Motivation & context

Also closes #3731 
It looks like as follow:
image

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/disclosure/disclosure.styles.ts
```diff
@@ -0,0 +1,19 @@
+import { css } from "@microsoft/fast-element";
+
+export const DisclosureStyles = css`
+    :host {
+        display: block;
+    }
+
+    :host ::slotted([slot="invoker"]) {
+        margin: 1rem auto;
```
janechu, line 9: @chrisdholt are we using `rem`?

packages/web-components/fast-foundation/src/disclosure/disclosure.template.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Disclosure } from "./disclosure";
+
+/**
+ * The template for the {@link @microsoft/fast-foundation#Disclosure} component.
+ * @public
+ */
+export const DisclosureTemplate = html<Disclosure>`
+    <template>
+        <slot name="invoker"></slot>
+        <slot name="content"></slot>
```
janechu, line 11: Hmm, why not use the default slot so this can just nest? `<slot></slot>`

janechu (commented):
Thanks for the contribution @hardikpthv! This is a lot of work 💯

hardikpthv:
The pleasure is all mine @janechu. Looking forward to contributing more. 😄

packages/web-components/fast-foundation/src/disclosure/disclosure.template.ts
hardikpthv: Yes, I am going to change it https://github.com/microsoft/fast/pull/3920#discussion_r490782451

hardikpthv: Done :)

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * @public
+ */
+export class Disclosure extends FASTElement {
+    /**
+     * Determines if the element should show the extra content or not.
+     *
+     */
+    @attr({ mode: "boolean" })
+    public expanded: boolean;
```
scomea, line 14: Do you need an expandedChanged handler to respond to attribute changes?  I think you'd need to call update(), for example.

chrisdholt (changes requested):
Thanks for this @hardikpthv! I'm going to hold on this primarily to ensure we do our due diligence in the spec first. I'd hate to churn things over here without having a clear specification we are building to.

hardikpthv:
@chrisdholt, I have adjusted the disclosure as per the latest specs. Could you already take a look? I will still have to fix the test case. 😅

packages/web-components/fast-components/src/disclosure/fixtures/disclosure.html
```diff
@@ -0,0 +1,98 @@
+<fast-design-system-provider use-defaults>
+    <h1>Disclosure</h1>
+    <fast-card
+        class="class-override"
+        style="--card-width: 400px; --card-height: auto; --elevation: 24; padding: 12px;"
+    >
```
chrisdholt, line 6: Is there any value in having this render in a card, or is it carryover from the card file?

packages/web-components/fast-components/src/disclosure/index.ts
```diff
@@ -0,0 +1,97 @@
+import { attr, customElement } from "@microsoft/fast-element";
+import { Disclosure, DisclosureTemplate as template } from "@microsoft/fast-foundation";
+import { DisclosureStyles as styles } from "./disclosure.styles";
+/**
+ * Types of anchor appearance.
+ * @public
+ */
+export type DisclosureAppearance = "accent" | "hypertext";
```
chrisdholt, line 8: I'm not sure we'll want this to render inline with text - we should likely update all references of `hypertext` to lightweight. The intent of hypertext is to style and have it rendered inline with text. Lightweight is more intended as a block-level element with a text-only appearance.

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.details.removeEventListener("toggle", this.onToggle);
+    }
+
+    constructor() {
+        super();
+        this.ariaExpanded = false;
+        this.ariaControls = `disclosure-content-${this.uuid()}`;
```
chrisdholt, line 69: I'd prefer to either tie this to an `id` passed by the user or a fixed id value. If this is in the shadow, this won't conflict and the ID should be unique to the Shadow DOM. No need to generate a UUID here IMO.

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,117 @@
...
+    protected onToggle() {
+        this.expanded = this.details.open;
+        this.ariaExpanded = this.expanded;
+        this.$emit("toggle");
+    }
+
+    /**
+     * Generate random UUID
+     */
+    private uuid = (): string => Math.random().toString(36).substr(2, 10);
```
chrisdholt, line 116: Let's remove this in favor of above. As an FYI, for things like this in the future you can leverage lodash-es where acceptable.

packages/web-components/fast-foundation/src/disclosure/disclosure.template.ts
```diff
@@ -0,0 +1,24 @@
+import { html, ref } from "@microsoft/fast-element";
+import { Disclosure } from "./disclosure";
+
+/**
+ * The template for the {@link @microsoft/fast-foundation#Disclosure} component.
+ * @public
+ */
+export const DisclosureTemplate = html<Disclosure>`
+    <template>
```
chrisdholt, line 9: Do we need this template element if we aren't using it?

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.details.addEventListener("toggle", this.onToggle);
+        if (this.expanded) {
+            this.show();
+        }
+    }
+
+    /**
+     * Update the aria attr and fire `toggle` event
+     */
+    protected onToggle() {
```
chrisdholt, line 107:
```suggestion
    protected toggle = () => {
```

packages/web-components/fast-components/src/disclosure/index.ts
```diff
@@ -0,0 +1,97 @@
...
+            this.hide();
+            this.height = getCurrentHeight();
+            this.show();
+        } else {
+            this.height = getCurrentHeight();
+            this.show();
+            this.totalHeight = getCurrentHeight();
+            this.hide();
+        }
+        this.details.style.setProperty("transition", "height 0.35s");
```
chrisdholt, line 85: Can you elaborate on the goals here - seems costly to setup. What's the value being added by running this as part of the component API?

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export class Disclosure extends FASTElement {
+    /**
+     * See {@link https://www.w3.org/WAI/PF/aria/roles#button} for more information
+     * @public
+     * @remarks
+     * HTML Attribute: aria-expanded
+     */
+    @attr({ attribute: "aria-expanded" })
+    public ariaExpanded: true | false;
```
chrisdholt, line 16: Do we need this if it's internal? Couldn't this just be related directly to the `expanded` property?

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Identifies the element (or elements) whose contents or presence are controlled by the current element.
+     *
+     * {@link https://www.w3.org/TR/wai-aria-1.1/#aria-controls}
+     * @public
+     * @remarks
+     * HTML Attribute: aria-controls
+     */
+    @attr({ attribute: "aria-controls" })
+    public ariaControls: string;
```
chrisdholt, line 27: Does this need to be publicly available or can it not just be set to the value of an `id` provided by the user? Alternatively, the Shadow DOM id should be scoped so you could technically bake this in...

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,117 @@
...
+            this.show();
+        }
+    }
+
+    /**
+     * Update the aria attr and fire `toggle` event
+     */
+    protected onToggle() {
+        this.expanded = this.details.open;
+        this.ariaExpanded = this.expanded;
```
chrisdholt, line 109: As noted - do we need `ariaExpanded` if the template just binds to `expanded`?

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
chrisdholt, line 107: should this be private?

packages/web-components/fast-components/src/disclosure/fixtures/disclosure.html
hardikpthv: Just to make a separation between all components in the preview.

packages/web-components/fast-components/src/disclosure/index.ts
hardikpthv: Yes, I think that would be more sensible to change it to `lightweight`. Will do 😄

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
hardikpthv: It's in shadow DOM so will replace it with a fixed id.

packages/web-components/fast-foundation/src/disclosure/disclosure.template.ts
hardikpthv: Not anymore.

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
hardikpthv: Not entirely sure from the A11Y aspects. Indeed `expanded` attribute is toggling in the component, and, the `aria-expanded` is given internally to the `<summary>` as well. Shall I remove it?

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
hardikpthv: Remove it.

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
hardikpthv: It's used in the extended component for animation purposes.

packages/web-components/fast-components/src/disclosure/index.ts
hardikpthv: It can be a part of the style file itself, however, I noticed that it would by default animate on the first render as well irrespective of the expanded or collapsed state.

packages/web-components/fast-components/src/disclosure/fixtures/disclosure.html
chrisdholt: Gotcha - this may end up causing issues as cards are soon going to extend design system provider which would cause the visual appearance to push based on the card background. We might want to simply leverage the heading OR add a slight bit of margin to each element using a style block:

```html
<style>
    fast-disclosure {
        margin-top: 12px;
    }
</style>
```

packages/web-components/fast-components/src/disclosure/index.ts
chrisdholt: That sounds like something that we can work through without needing to call GBCR a few times, hiding, then showing the control. Animating height is difficult regardless. Can you articulate the end goal you are going for and we can see if we can help?

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
chrisdholt: `aria-expanded` is needed in the template, but it should work fine to set it's value to `expanded`. We shouldn't need both. The attribute of aria-expanded isn't really needed here as it's just storing the expanded value.

```
expanded => set aria-expanded attribute => set aria-expanded in template
```

The workflow could simply be:
```
expanded => set aria-expanded in template
```

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
```diff
@@ -0,0 +1,100 @@
...
+     * @internal
+     */
+    public disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.details.removeEventListener("toggle", this.onToggle);
+    }
+
+    constructor() {
+        super();
+        this.ariaExpanded = false;
```
chrisdholt, line 57: This can likely be removed. If we want to default expanded, we can do that in connectedCallback():
```
if (!this.expanded) {
    this.expanded = false;
}
```

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
hardikpthv: Yes, thanks for the explanation. Will remove it.

packages/web-components/fast-foundation/src/disclosure/disclosure.ts
hardikpthv: Not needed anymore 😅

packages/web-components/fast-components/src/disclosure/index.ts
hardikpthv: The idea is to calculate the heights dynamically when the disclosure is expanded and collapsed. These heights are calculated once on `connectedCallback` and stored for later use. However, I was able to remove line no. 85 and add a transition in style file itself after removing the card from the demo.

hardikpthv: @chrisdholt, the changes are made. Could you review one more time? Do let me know in case more changes expected. 😄

packages/web-components/fast-components/src/disclosure/index.ts
```diff
@@ -0,0 +1,93 @@
...
+        super.setup();
+        if (!this.appearance) {
+            this.appearance = "accent";
+        }
+
+        const getCurrentHeight = () => this.details.getBoundingClientRect().height;
+        this.show();
+        this.totalHeight = getCurrentHeight();
+        this.hide();
+        this.height = getCurrentHeight();
```
chrisdholt, line 77: I'm still unsure we want to do all this work on initialization, but we can test it and see how it goes.

chrisdholt (approved):

packages/web-components/fast-components/src/disclosure/disclosure.styles.ts
```diff
@@ -0,0 +1,84 @@
...
+        font-size: var(--type-ramp-base-font-size);
+        border-radius: calc(var(--corner-radius) * 1px);
+        outline: none;
+        cursor: pointer;
+        margin: 16px 0;
+        padding: 12px;
+        max-width: max-content;
+    }
+
+    :host(.accent) .invoker:active {
```
EisenbergEffect, line 38: Can we move these host selectors to be attribute based instead of class based?

packages/web-components/fast-components/src/disclosure/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * The appearance the anchor should have.
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: appearance
+     */
+    @attr
+    public appearance: DisclosureAppearance;
+    public appearanceChanged(
```
EisenbergEffect, line 44: This change handler can be replaced by a binding on the host element.

EisenbergEffect (changes requested):

EisenbergEffect:
Just wanted to check in with you @hardikpthv I think we're really close on this. Just a couple more adjustments needed. Are you still interested in working on it? Let us know and thanks for the great contribution so far!

hardikpthv:
> Just wanted to check in with you @hardikpthv I think we're really close on this. Just a couple more adjustments needed. Are you still interested in working on it? Let us know and thanks for the great contribution so far!

Sorry for the delay @EisenbergEffect. I will make the adjustments sooner. :)

EisenbergEffect:
No need to apologize! I just wanted to check in 😄 Ping me when you are ready for another review and we'll get this moving along. Very excited!

packages/web-components/fast-components/src/disclosure/disclosure.styles.ts
hardikpthv: Done :)

hardikpthv:
@EisenbergEffect, @chrisdholt I added final changes. Could you take a look at them? Thanks.

EisenbergEffect (approved):
Awesome! @chrisdholt Ready to merge?

chrisdholt:
> Awesome! @chrisdholt Ready to merge?

@EisenbergEffect Almost, looks like API report needs to be run as tests fail that check with untracked files. @hardikpthv can you run `yarn` from the root of the repo and commit the files generated automatically?

chrisdholt:
> > Awesome! @chrisdholt Ready to merge?
> 
> @EisenbergEffect Almost, looks like API report needs to be run as tests fail that check with untracked files. @hardikpthv can you run `yarn` from the root of the repo and commit the files generated automatically?

@hardikpthv You'll definitely need to locally run `yarn` from the root of the repo to generate and commit the API report. Alternatively, I did this locally but am unable to push to your fork. If you can check the box allowing edits by maintainers I can manage this and get the PR in! [Allowing edits by maintainers](https://stackoverflow.com/questions/63341296/github-pull-request-allow-edits-by-maintainers)

hardikpthv:
@chrisdholt, Apparently I see that it’s checked already to allow to edit by maintainers. Could you check once?

chrisdholt:
Thanks @hardikpthv for this significant contribution!

hardikpthv:
> Thanks @hardikpthv for this significant contribution!

It's my pleasure, @chrisdholt! Looking forward to contributing more and more. :)

EisenbergEffect:
Thanks @hardikpthv !

Issue #3914, opened 2020-09-16
Tree item must have click handler on root element


scomea:
Fast tree-views don't work in narrator's scan mode because there is no click handler on the root element of Fast tree item.  This is required in order to support assistive technologies as per - https://www.w3.org/TR/core-aam-1.2/#mapping_actions.   

An additional concern is that in scan mode it is not possible to open/close a node via keyboard (space selects, but that's normally how we expand/collapse)

Example:

1. Navigate to a https://explore.fast.design/components/fast-tree-view
2.  open the debugger and hook up a breakpoint on the tree view item's handleSelect function (or find a tree view with a visible select action)
3. Turn on narrator, turn on scan mode - narrator key ("caps lock" for me) + space to toggle
4. with scan on try to select tree item with space bar

**What behavior did you expect?**
Support accessibility standards

PR #3951, opened 2020-09-23
fix: tree view item should have a click handler on root element

scomea:
# Description
In order to comply with https://www.w3.org/TR/core-aam-1.2/#mapping_actions. and enable some accessibility scenarios tree view items must be able to react to click events sent to the root of the component.  

Additionally this fix marks some handled key events with prevent default so that they don't trigger additional actions (I noticed space bar was triggering scrolls when I tested)

## Motivation & context
Fixes https://github.com/microsoft/fast/issues/3914

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -162,25 +162,27 @@ export class TreeItem extends FASTElement {
 
         switch (e.keyCode) {
             case keyCodeArrowLeft:
+                e.preventDefault();
                 this.handleArrowLeft();
                 break;
             case keyCodeArrowRight:
+                e.preventDefault();
                 this.handleArrowRight();
                 break;
             case keyCodeArrowDown:
-                // preventDefault to ensure we don't scroll the page
                 e.preventDefault();
                 this.focusNextNode(1);
                 break;
             case keyCodeArrowUp:
-                // preventDefault to ensure we don't scroll the page
                 e.preventDefault();
                 this.focusNextNode(-1);
                 break;
             case keyCodeEnter:
+                e.preventDefault();
                 this.handleSelected(e);
                 break;
             case keyCodeSpace:
+                e.preventDefault();
                 this.handleSpaceBar();
                 break;
         }
@@ -188,21 +190,20 @@ export class TreeItem extends FASTElement {
         return true;
     };
 
-    public handleExpandCollapseButtonClick = (): void => {
+    public handleExpandCollapseButtonClick = (e: MouseEvent): void => {
         if (!this.disabled) {
+            e.preventDefault();
             this.setExpanded(!this.expanded);
         }
     };
 
-    public handleContainerClick = (e: MouseEvent): void => {
-        const expandButton: HTMLElement | null = this.expandCollapseButton;
-        const isButtonAnHTMLElement: boolean = isHTMLElement(expandButton);
+    public handleClick = (e: MouseEvent): void => {
         if (
-            (!isButtonAnHTMLElement ||
-                (isButtonAnHTMLElement && expandButton !== e.target)) &&
-            !this.disabled
+            !this.disabled &&
+            !e.defaultPrevented
         ) {
             this.handleSelected(e);
+            e.preventDefault();
         }
     };
 
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -162,12 +162,10 @@ export class TreeItem extends FASTElement {
 
         switch (e.keyCode) {
             case keyCodeArrowLeft:
-                e.preventDefault();
-                this.handleArrowLeft();
+                this.handleArrowLeft(e);
                 break;
             case keyCodeArrowRight:
-                e.preventDefault();
-                this.handleArrowRight();
+                this.handleArrowRight(e);
                 break;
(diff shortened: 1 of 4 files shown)


marjonlynch (approved):
smoked, reviewed, looks good

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -162,47 +162,48 @@ export class TreeItem extends FASTElement {
...
             case keyCodeArrowLeft:
+                e.preventDefault();
                 this.handleArrowLeft();
                 break;
             case keyCodeArrowRight:
+                e.preventDefault();
                 this.handleArrowRight();
                 break;
             case keyCodeArrowDown:
-                // preventDefault to ensure we don't scroll the page
```
chrisdholt, line 171: Is there a reason we're dropping these comments?

scomea: Because all the "preventDefault" calls do some variation of that.  Seems redundant.

chrisdholt (changes requested):
I'm not sure we should prevent default on spacebar - there is no expectation from the spec that spacebar be handled by single select vertical trees: https://w3c.github.io/aria-practices/#TreeView

Our tree view does not currently manage multi-select, so I see no reason to handle an event with spacebar. To prevent the click handler from being blocked, we should likely take any questions on space to an issue.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
chrisdholt: Certain others are not necessary, such as spacebar - https://github.com/microsoft/fast/pull/3951#pullrequestreview-494981192

scomea: (and even arrow key events could cause side effects other than page scrolling as they bubble, we just picked that one because we saw it)

scomea:
We should mark space bar if it takes an action, I see that we need to add some logic to not mark it when it does not expand/collapse a node tho.

chrisdholt:
> We should mark space bare events if it takes an action, I see that we need to add some logic to not mark it when it does not expand/collapse a node tho.

Per the specification - what is the behavior of spacebar in a vertically oriented tree? I don't see an interaction defined, in which case I would expect it to scroll the page.

scomea:
I'm further modifying the space bar handler to only mark the event with default prevented if the expanded state is actually changed so we don't preventDefault unless there was a ui change.
```
    private handleSpaceBar(e: KeyboardEvent): void {
        if (typeof this.expanded !== "boolean") {
            return;
        }
        e.preventDefault();
        this.setExpanded(!this.expanded);
    }
```

chrisdholt:
> I'm further modifying the space bar handler to only mark the event with default prevented if the expanded state is actually changed so we don't preventDefault unless there was a ui change.
> 
> ```
>     private handleSpaceBar(e: KeyboardEvent): void {
>         if (typeof this.expanded !== "boolean") {
>             return;
>         }
>         e.preventDefault();
>         this.setExpanded(!this.expanded);
>     }
> ```

I think the issue here is that this functionality is incorrect. Even in multi-select trees spacebar does a different thing. Digging in the spec, spacebar is a no-op for tree view keyboard interaction from what I'm seeing. Addressing the spacebar issue would mean removing the behavior and allowing it to scroll the screen. I'm just not seeing spacebar defined in the spec the way we handle it in the control.

scomea:
That would be a different issue.  I'm going on the notion that any time a keyboard input causes a ui change it should be marked with preventDefault so upstream consumers of the bubbled event don't unknowingly trigger a second action.  If we chage the component so that space bar does not do anything we would remove the preventDefault at that time.

chrisdholt:
> That would be a different issue. I'm going on the notion that any time a keyboard input causes a ui change it should be marked with preventDefault so upstream consumers of the bubbled event don't unknowingly trigger a second action. If we chage the component so that space bar does not do anything we would remove the preventDefault at that time.

I think we should start the prevent default as an issue rather than an assumption being introduced into a single component here - that would allow us to address this cohesively across the system. I agree this is a separate issue which is why I would ask that we pull it from the PR focus this on the valid issue of UIA markup (where the click event happens). I agree the removal of space is a different PR as well.

scomea:
Prevent default issue.  https://github.com/microsoft/fast/issues/3953

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -188,20 +188,15 @@ export class TreeItem extends FASTElement {
         return true;
     };
 
-    public handleExpandCollapseButtonClick = (): void => {
+    public handleExpandCollapseButtonClick = (e: MouseEvent): void => {
         if (!this.disabled) {
+            e.preventDefault();
```
chrisdholt, line 193: Hypothetical question on this - lets say that a team passes our upcoming telemetry directive to a tree item and expects to measure whether an item was expanded or selected. Would this suppress the event bubbling to that capture?

scomea: preventDefault does not change bubbling behavior, it just enables other event consumers to tell some other code has taken some action based on that event. That's why I think preventDefault is a better way to address this than stopPropagation.

chrisdholt: doh

chrisdholt (approved):

PR #3957, opened 2020-09-24
refactor(fast-element): clean up compilation context concept in compiler

EisenbergEffect:
# Description

This is an internal refactoring of the template compiler to make the compilation context more explicit, reduce some duplication, and enable future nested calls to the compile function without messing up global state.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-element/src/template-compiler.ts
+++ b/packages/web-components/fast-element/src/template-compiler.ts
@@ -7,14 +7,46 @@ import { ExecutionContext, Binding } from "./observation/observable";
 type InlineDirective = Directive & {
     targetName?: string;
     binding: Binding;
-    targetAtContent();
+    targetAtContent(): void;
 };
 
-const compilationContext = { locatedDirectives: 0, targetIndex: -1 };
+class CompilationContext {
+    public targetIndex!: number;
+    public behaviorFactories!: BehaviorFactory[];
+    public directives: ReadonlyArray<Directive>;
+
+    public addFactory(factory: BehaviorFactory) {
+        factory.targetIndex = this.targetIndex;
+        this.behaviorFactories.push(factory);
+    }
+
+    public captureContentBinding(directive: BindingDirective): void {
+        directive.targetAtContent();
+        this.addFactory(directive);
+    }
+
+    public reset() {
+        this.behaviorFactories = [];
+        this.targetIndex = -1;
+    }
+
+    public release() {
+        sharedContext = this;
+    }
+
+    public static borrow(directives: ReadonlyArray<Directive>) {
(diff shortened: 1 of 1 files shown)


awentzel (approved):

packages/web-components/fast-element/src/template-compiler.ts
```diff
@@ -7,14 +7,46 @@ import { ExecutionContext, Binding } from "./observation/observable";
...
+    public targetIndex!: number;
+    public behaviorFactories!: BehaviorFactory[];
+    public directives: ReadonlyArray<Directive>;
+
+    public addFactory(factory: BehaviorFactory) {
+        factory.targetIndex = this.targetIndex;
+        this.behaviorFactories.push(factory);
+    }
+
+    public captureContentBinding(directive: BindingDirective): void {
```
radium-v, line 23: This method has an explicit `void` return type, though the other methods in the class don't. Is this intentional?

EisenbergEffect: Nope 🤣 Will fix.

packages/web-components/fast-element/src/template-compiler.ts
```diff
@@ -7,14 +7,46 @@ import { ExecutionContext, Binding } from "./observation/observable";
...
+        this.behaviorFactories = [];
+        this.targetIndex = -1;
+    }
+
+    public release(): void {
+        sharedContext = this;
+    }
+
+    public static borrow(directives: ReadonlyArray<Directive>): CompilationContext {
+        const shareable = sharedContext || new CompilationContext();
```
janechu, line 38: So shareable can either be the `CompilationContext` or an instance of `CompilationContenxt`?

radium-v (approved):

janechu (approved):

EisenbergEffect:
@janechu  I started to respond to your comment but couldn't post because I think you removed it. But here's what I was writing for further info:

Basically, the borrow/release mechanism is just a way to reuse the context object across compilations so that when needing to compile a bunch of templates rapidly at startup we don't need to keep creating context instances and immediately throwing them away. So, a perf optimization. An important detail is that we don't have a strict singleton for this object because I want to enable nested runs of the compiler, in which case the context could not be shared. So, this code basically reuses the instance when it's not being used otherwise creates an instance on the fly.

At present we don't have scenarios where compile is recursively run in any way, but I've got some PoCs for new stuff that might require that. So, I addressed the inability to do that along with bit of refactor with this PR.

(This probably would have been good content for the PR description. Sorry!)

marjonlynch (approved):
Smoked, looks good in firefox/edge for storybook and explorer

PR #3976, opened 2020-09-25
fix: allow child radio buttons to be changed multiple times from javascript

### LINKED ISSUE #3903: Externally changing checked attribute in fast-components/fast-radio causes checked state to stop updating
When changing the checked attribute via `<fast-radio ... ?checked=${x => x.shouldBeChecked}>...</fast-radio>` between multiple options after a single round the later options will no longer accept the checked state.

Using the provided example:
1. Click **With Selected Three**
2. Option **Three** should be selected
3. Click **With Selected One**
4. Option **One** should be selected
5. Click **With Selected Three**
6. Option **Three** is not selected, but should be

Changing the `?selected` value should cause the radio to be selected

Initial State
Option Three Selected When Button Clicked
Option One Selected When Button Clicked
Option Three not Selected When Button Clicked

* MacOS
* Firefox 80.0.1 and Chrome 85.0.4183.102

Reproduction:
[fast-radio-issue.zip](https://github.com/microsoft/fast/files/5227322/fast-radio-issue.zip)

marjonlynch:
# Description
If a developer tried to change a particular radio button in a radio group multiple times it would fail to change after the first change.

For input type='radio' there is a bit of a disconnect between the property checked and the attribute checked which is different than most of the other components (though Checkbox has this same behavior). 
* initially when radio group loads we need to look at the last radio with checked attribute and set the property to checked on that radio and set the checked property(NOT the attribute) to false on all the others.
* once that checked property is set, it can only be changed through the property and there is a disconnect between the checked attribute and the checked property.

Also fixed tabindexes when values are changed via code, they were not updated appropriately, only when using mouse interaction.

* still need to add specs

closes #3903 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/radio-group/radio-group.spec.ts
+++ b/packages/web-components/fast-foundation/src/radio-group/radio-group.spec.ts
@@ -266,6 +266,42 @@ describe("Radio Group", () => {
         await disconnect();
     });
 
+    it("should mark the last radio defaulted to checked as checked, the rest should not be checked", async () => {
+        const { element, connect, disconnect } = await fixture(html<FASTRadioGroup>`
+            <fast-radio-group>
+                <fast-radio value="foo">Foo</fast-radio>
+                <fast-radio value="bar" checked>Bar</fast-radio>
+                <fast-radio value="baz" checked>Baz</fast-radio>
+            </fast-radio-group>
+        `);
+
+        await connect();
+        await DOM.nextUpdate();
+
+        const radios: NodeList = element.querySelectorAll("fast-radio");
+        expect((radios[2] as HTMLInputElement).checked).to.equal(true);
+        expect((radios[1] as HTMLInputElement).checked).to.equal(false);
+    });
+
+    it("should mark radio matching value on radio-group over any checked attributes", async () => {
+        const { element, connect, disconnect } = await fixture(html<FASTRadioGroup>`
+            <fast-radio-group value="bar">
+                <fast-radio value="foo">Foo</fast-radio>
+                <fast-radio value="bar" checked>Bar</fast-radio>
+                <fast-radio value="baz" checked>Baz</fast-radio>
+            </fast-radio-group>
+        `);
+
+        await connect();
+        await DOM.nextUpdate();
+
+        const radios: NodeList = element.querySelectorAll("fast-radio");
+        expect((radios[2] as HTMLInputElement).checked).to.equal(false);
+        expect((radios[1] as HTMLInputElement).checked).to.equal(true);
+        // radio-group explicitly sets non-matching radio's checked to false if a value match was found
+        expect((radios[2] as HTMLInputElement).hasAttribute("checked")).to.equal(false);
+    });
+
     it("should NOT set a child radio to `checked` if its value does not match the radiogroup `value`", async () => {
         const { element, connect, disconnect } = await fixture(html<FASTRadioGroup>`
             <fast-radio-group value="baz">
--- a/packages/web-components/fast-foundation/src/radio-group/radio-group.spec.ts
+++ b/packages/web-components/fast-foundation/src/radio-group/radio-group.spec.ts
@@ -267,7 +267,7 @@ describe("Radio Group", () => {
     });
 
     it("should mark the last radio defaulted to checked as checked, the rest should not be checked", async () => {
-        const { element, connect, disconnect } = await fixture(html<FASTRadioGroup>`
+        const { element, connect, disconnect } = await fixture(html`
             <fast-radio-group>
                 <fast-radio value="foo">Foo</fast-radio>
                 <fast-radio value="bar" checked>Bar</fast-radio>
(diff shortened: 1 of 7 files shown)


packages/web-components/fast-foundation/src/radio/radio.ts
```diff
@@ -101,17 +101,20 @@ export class Radio extends FormAssociated<HTMLInputElement> implements RadioCont
...
     public checked: boolean = this.defaultChecked;
     private checkedChanged(): void {
-        if (!this.dirtyChecked) {
-            this.dirtyChecked = true;
-        }
+        // TODO: temporarily removed as it's causing issues with
+        // changing the value via code and from radio-group
+        // if (!this.dirtyChecked) {
+        //     this.dirtyChecked = true;
+        // }
```
chrisdholt, line 108: Is this still temporarily removed? If so, can we have something tracking it?

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
```diff
@@ -87,6 +87,16 @@ export class RadioGroup extends FASTElement {
      */
     @attr
     public value: string;
+    protected valueChanged(): void {
```
chrisdholt, line 90: Can you add tests for setting the value of the group and the correct radio being selected?

marjonlynch: yes, will do, plan to add a few tests for these changes

packages/web-components/fast-foundation/src/radio/radio.ts
marjonlynch: nope, added it back in now

packages/web-components/fast-foundation/src/radio-group/radio-group.spec.ts
```diff
@@ -266,6 +266,42 @@ describe("Radio Group", () => {
         await disconnect();
     });
 
+    it("should mark the last radio defaulted to checked as checked, the rest should not be checked", async () => {
+        const { element, connect, disconnect } = await fixture(html<FASTRadioGroup>`
```
EisenbergEffect, line 270: You don't need the type on the html function in this case. That would only be needed if you had a special view model that you wanted to test *binding* the radio group to.

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
```diff
@@ -139,7 +149,31 @@ export class RadioGroup extends FASTElement {
             this.parentToolbar !== undefined && this.parentToolbar !== null;
     }
 
+    public disconnectedCallback(): void {
+        this.removeEventListener("keydown", this.keydownHandler);
```
EisenbergEffect, line 153: If these event handlers were added directly to the template itself, you wouldn't need to have manual code to remove them.

EisenbergEffect (approved):
A few minor things that could be adjusted. Nothing major so good to go. Would be good to get a review from Nick since he's been looking into various approaches with form controls in general.

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
marjonlynch: tests added, also for the multiple checked child radio situation

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
marjonlynch: moved handlers to template, needed to return true on focusOutHandler but all working again now

packages/web-components/fast-foundation/src/radio-group/radio-group.spec.ts
marjonlynch: I removed it for my specs, as I don't need to cast them specifically to FASTRadioGroup.

packages/web-components/fast-foundation/src/radio-group/radio-group.template.ts
```diff
@@ -11,6 +11,10 @@ export const RadioGroupTemplate = html<RadioGroup>`
         role="radiogroup"
         aria-disabled="${x => x.disabled}"
         aria-readonly="${x => x.readOnly}"
+        @keypress="${(x, c) => x.keypressHandler(c.event as KeyboardEvent)}"
+        @click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
+        @keydown="${(x, c) => x.keydownHandler(c.event as KeyboardEvent)}"
```
scomea, line 16: We can't just use one of keydown/keypress?  Why both?

marjonlynch: well that seems like a valid question, looking into it

marjonlynch: Well it was never getting called anyway as far as I can tell, so just removed it, also fixed the tabbing bug that was caused by me moving event handlers to template. That gets me every time, wondering if that's the default we really want.

scomea (approved):

EisenbergEffect:
@marjonlynch Looks like this is good to go 😄

PR #3986, opened 2020-09-29
fix: tabs directional arrow key bug

eljefe223:
# Description
We were overthinking direction in tabs ;). No need for direction check.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tabs/tabs.ts
+++ b/packages/web-components/fast-foundation/src/tabs/tabs.ts
@@ -239,11 +239,11 @@ export class Tabs extends FASTElement {
             switch (keyCode) {
                 case keyCodeArrowLeft:
                     event.preventDefault();
-                    this.adjust(1);
+                    this.adjust(-1);
                     break;
                 case keyCodeArrowRight:
                     event.preventDefault();
-                    this.adjust(-1);
+                    this.adjust(1);
                     break;
             }
         } else {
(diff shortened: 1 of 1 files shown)


awentzel (approved):

packages/web-components/fast-foundation/src/tabs/tabs.ts
```diff
@@ -242,11 +239,11 @@ export class Tabs extends FASTElement {
             switch (keyCode) {
                 case keyCodeArrowLeft:
                     event.preventDefault();
-                    this.adjust(this.direction === Direction.rtl ? 1 : -1);
+                    this.adjust(1);
```
SethDonohue, line 242: This is backwards I believe. Arrow Key Left should adjust -1.

packages/web-components/fast-foundation/src/tabs/tabs.ts
```diff
@@ -242,11 +239,11 @@ export class Tabs extends FASTElement {
...
                     break;
                 case keyCodeArrowRight:
                     event.preventDefault();
-                    this.adjust(this.direction === Direction.rtl ? -1 : 1);
+                    this.adjust(-1);
```
SethDonohue, line 246: Similar comment as above, but arrow key right should +1.

SethDonohue (commented):
Am I missing something here? I tested this PR and arrow key left moves the tabs to the right, opposite is true for the arrow key right...

Any why do we not need the Direction check and switch?

packages/web-components/fast-foundation/src/tabs/tabs.ts
eljefe223: your right! ;)

nicholasrice (approved):

PR #3988, opened 2020-09-29
fix: tooltip does not position correctly in documentation sample

scomea:
# Description
I noticed that tooltip was not positioning itself properly in our documentation sample at https://www.fast.design/docs/components/tooltip.

After some investigation I found that the component's internal attibute bindings were getting munged into a big string with no spaces:

tooltipbinding

After much poking around I was able to resolve the issue for tooltip by modifying the template to wrap all attribute value bindings in quotes:

```ts
vertical-positioning-mode="${x => x.verticalPositioningMode}"
```

instead of 

```ts
vertical-positioning-mode=${x => x.verticalPositioningMode}
```
I will open a tracking bug to investigate the issue further as technically we don't require the quotes and it isn't clear what is causing the tooltip template to break without them in our document environment.

## Motivation & context
Tooltip should position properly.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tooltip/tooltip.template.ts
+++ b/packages/web-components/fast-foundation/src/tooltip/tooltip.template.ts
@@ -10,17 +10,17 @@ export const TooltipTemplate = html<Tooltip>`
         x => x.tooltipVisible,
         html<Tooltip>`
             <fast-anchored-region
-                ${ref("region")}
-                dir=${x => x.currentDirection}
-                vertical-positioning-mode=${x => x.verticalPositioningMode}
-                vertical-default-position=${x => x.verticalDefaultPosition}
-                vertical-inset=${x => x.verticalInset}
-                vertical-scaling=${x => x.verticalScaling}
-                horizontal-positioning-mode=${x => x.horizontalPositioningMode}
-                horizontal-default-position=${x => x.horizontalDefaultPosition}
-                horizontal-scaling=${x => x.horizontalScaling}
-                horizontal-inset=${x => x.horizontalInset}
                 fixed-placement="true"
+                vertical-positioning-mode="${x => x.verticalPositioningMode}"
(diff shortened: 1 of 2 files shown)


EisenbergEffect (approved):

packages/web-components/fast-foundation/src/tooltip/tooltip.template.ts
```diff
@@ -10,17 +10,17 @@ export const TooltipTemplate = html<Tooltip>`
...
                 fixed-placement="true"
+                vertical-positioning-mode="${x => x.verticalPositioningMode}"
+                vertical-default-position="${x => x.verticalDefaultPosition}"
+                vertical-inset="${x => x.verticalInset}"
+                vertical-scaling="${x => x.verticalScaling}"
+                horizontal-positioning-mode="${x => x.horizontalPositioningMode}"
+                horizontal-default-position="${x => x.horizontalDefaultPosition}"
+                horizontal-scaling="${x => x.horizontalScaling}"
+                horizontal-inset="${x => x.horizontalInset}"
+                dir="${x => x.currentDirection}"
```
chrisdholt, line 22: Should we have tests for these as they're technically part of the tooltip API? That might help catch a regression here in the future.

chrisdholt (approved):
Approving, but perhaps we should add tests for the below?

packages/web-components/fast-foundation/src/tooltip/tooltip.template.ts
scomea: Yes, I will be boosting tooltip and anchored region test coverage as it is currently bare bones, although I don't think we can write a test to catch this particular issue as the only repro so far is in the doc environment. If it turns out that component attribute bindings should be wrapped in quotes generally then maybe that's something we can test for, but I don't think we've decided to make that a requirement yet.

PR #3993, opened 2020-09-30
feat(fast-router): add a web-components-based router to FAST

EisenbergEffect:
# Description

This is a PR for a web-components-based router introduced through a new package named `@microsoft/fast-router`. Most of the initial feature-set is implemented. A few gaps will be filled during the alpha period.

## Motivation & context

So far, we've focused FAST around individual component and design system scenarios. As a result, building larger experiences and apps typically requires adding an additional front-end framework to the mix. While we are committed to continuing to support using FAST with or without any front-end framework, we also want to have a 100% web-components path for those that need or want to go that route (pun intended 🤣). This PR implements a router based entirely on web components, which is the biggest missing piece required to enable expanding FAST from components to apps.

### Features

- [x] The `fast-router` custom element, that forms a placeholder for where the current route will be rendered
- [x] No dependencies other than `fast-element` 
- [x] A router configuration model that allows configuring...
  - [x] static routes
  - [x] parameterized routes
  - [x] splat routes
  - [x] optional parameters in routes
  - [x] redirect routes
  - [x] ignored routes
  - [x] fallback routes
  - [x] user settings for routes
  - [x] custom layouts for routes
  - [x] custom transitions for routes
  - [x] async component resolution (enables dynamically loaded/selected route components)
  - [x] navigate to an element or a complex template
  - [x] history event queue behavior
  - [x] link behavior
  - [x] default layout 
  - [x] default transition/animation
  - [x] all core services (e.g. route recognition, title building) 
- [x] global navigation contributors (guards and other lifecycle contributors, i.e. auth hooks)
- [x] A basic navigation lifecycle with navigation rollback on deny
- [x] Pass route parameters to components
- [x] An API for push, replace, and trigger of navigation
- [x] Simple child route configuration
- [x] Typed route params
  - [x] typed param route syntax
  - [x] built-in basic type converters (string, int, float, bool, Date)
  - [x] custom type converters api
- [x] Deliver typed route parameters to navigation participants
- [x] Infinite nested routers with connected/transactional lifecycles
- [x] Document title manipulation
- [x] Enable DI integration without DI dependency
- [x] Router events
- [x] Parent/child configuration inheritance for nested routers
- [x] Friendly names for routes
- [x] Generate routes from route names or path pattern and params
- [x] Support routes with query strings and query string parse/generate 
- [x] Navigation by friendly name

Of course, in addition to the missing features, we need **a new section in the documentation on routing** as well as **a full suite of tests**. There are over 10k unit tests, but those tests cover only the route recognition engine at this point. More tests will be added covering the rest of the codebase during the alpha period.

## Example

To help understand how the router will work, here's an example application route setup that shows configuring various routes with their components, as well as custom settings that are leveraged by an auth contributor.

```ts
import { RouterConfiguration, Route } from '@microsoft/fast-router';
import { HomeScreen } from './home';
import { AccountLogin } from './account/login';
import { AccountSignup } from './account/signup';
import { ConfirmEmail } from './account/confirm-email';
import { Session } from './account/session';
import { AccountSettings } from './account/settings';
import { StoreArea } from './store/store-area';
import { NotFound } from './not-found';
import { pageLayout } from './layouts/page-layout';
import { anonymousLayout } from './layouts/anonymous-layout';

type RouteSettings = {
  public?: boolean
};

export class AppRouterConfiguration extends RouterConfiguration<RouteSettings> {
  public configure() {
    this.title = 'My App';
    this.defaultLayout = pageLayout;
    this.routes.map(
      { path: '', redirect: 'home' },
      { path: 'home', title: 'Home', element: HomeScreen },
      { 
        path: 'account', 
        layout: anonymousLayout, 
        title: 'Account',
        settings: { public: true }, 
        children: [
          { path: "login", title: 'Login', element: AccountLogin, name: 'login' },
          { path: 'signup' title: 'Signup', element: AccountSignup },
          { path: 'confirm/{confirmation:Confirmation}', title: 'Confirm', element: ConfirmEmail },
          { path: 'settings', title: 'Settings', element: AccountSettings, layout: pageLayout, settings: { public: false } },
        ] 
      },
      { path: 'store', title: 'Store', element: StoreArea, childRouters: true },
      { path: 'not-found', title: 'Not Found', element: NotFound }
    );

    this.routes.fallback(
      () => Session.isLoggedIn
        ? { redirect: 'not-found' }
        : { redirect: 'login' }
    );

    this.routes.converter("Confirmation", async (confirmationId) => {
      // ...fetch confirmation from web service...
      return confirmation;
    });

    this.contributors.push({
      navigate(phase) {
        const settings = phase.route.settings;
  
        if (settings && settings.public) {
          return;
        }
  
        if (Session.loggedIn) {
          return;
        }
  
        phase.cancel(() => {
          Session.returnUrl = Route.path.current;
          Route.name.replace(phase.router, 'login');
        });
      }
    });
  }
}
```

And here's what your main application component would look like that connects this config to the router:

```ts
import { FASTElement, customElement, html, css } from '@microsoft/fast-element';
import { FASTDesignSystemProvider } from '@microsoft/fast-components';
import { FASTRouter } from '@microsoft/fast-router';
import { AppRouterConfiguration } from './routes';

FASTDesignSystemProvider;
FASTRouter;

const template = html<MainApplication>`
    <fast-router :config=${x => x.routerConfiguration}></fast-router>
  </fast-design-system-provider>
`;

const styles = css`
  :host {
    contain: content;
  }

  :host, fast-design-system-provider, fast-router {  
    display: block;
    width: 100%;
    height: 100%;
  }
`;

@customElement({
  name: 'main-application',
  template,
  styles
})
export class MainApplication extends FASTElement {
  routerConfiguration = new AppRouterConfiguration();
}
```

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-router/src/commands.ts
```diff
@@ -0,0 +1,198 @@
...
+
+export class Ignore implements NavigationCommand {
+    public async execute(transaction: NavigationTransaction): Promise<void> {}
+}
+
+export class Redirect implements NavigationCommand {
+    constructor(private redirect: string) {}
+
+    public async execute(transaction: NavigationTransaction): Promise<void> {
+        // TODO: generate new route based on redirect and route params
```
janechu, line 35: What kind of example new route would be generated based on route params? Like if you specified a pattern like: `/foo/:id/:name` and your route params included something like `{ id: "bar", name: "bob" }` you'd want to generate `foo/bar/bob`?

packages/web-components/fast-router/src/commands.ts
```diff
@@ -0,0 +1,198 @@
...
+            return;
+        }
+
+        const router = transaction.owner;
+        let currentView = router.view;
+        let createView = this.createView;
+
+        const viewModel = {}; // TODO: build view model from route params
+
+        if (router.command === this) {
```
janechu, line 76: So, if I understand correctly if the cope of the router.command is the same as the current scope (ie we are on the same route), then createView returns the currentView? Otherwise a new view will be created? What gets returned if the currentView is returned though, won't it still pass the if statement below?

packages/web-components/fast-router/src/commands.ts
```diff
@@ -0,0 +1,198 @@
...
+                if (typeof template === "function") {
+                    template = await template();
+                }
+
+                return template.create();
+            };
+        } else {
+            createView = async () => {
+                let element = definition.element;
+                let template: ViewTemplate = (definition as any).template;
```
janechu, line 141: if the above `if` check fails, won't this always be undefined?

packages/web-components/fast-router/src/enlistment.ts
```diff
@@ -0,0 +1,105 @@
...
+    constructor(
+        private readonly parentRouter: Router,
+        private readonly childRouter: Router
+    ) {
+        console.log("child enlistment", this);
+        parentRouter.addNavigationParticipant(childRouter);
+    }
+}
+
+export type NavigationEnlistment = MainEnlistment | ChildEnlistment;
```
janechu, line 66: Apparently `|` means union, is that what you're intending? Or should MainEnlistment and ChildEnlistment be considered separate for all types of NavigationEnlistment? That's why I created an `XOR` type in tooling see: https://github.com/microsoft/fast/blob/master/packages/tooling/fast-tooling/src/data-utilities/type.utilities.ts#L2

packages/web-components/fast-router/src/links.ts
```diff
@@ -0,0 +1,88 @@
...
+            target.hasAttribute("download") ||
+            target.hasAttribute("router-ignore") ||
+            target.hasAttribute("data-router-ignore")
+        ) {
+            return info;
+        }
+
+        if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
+            return info;
+        }
```
janechu, line 55: seems like these could be combined

packages/web-components/fast-router/src/links.ts
```diff
@@ -0,0 +1,88 @@
...
+    connect(): void;
+    disconnect(): void;
+}
+
+export class DefaultLinkHandler implements LinkHandler {
+    private handler = (e: MouseEvent) => {
+        let { shouldHandleEvent, href } = this.getEventInfo(e);
+
+        if (shouldHandleEvent) {
+            e.preventDefault();
```
janechu, line 19: Should we explicitly check for navigation events before preventing default? As such, is it possible that we accidentally catch a button submit event and preventDefault on that?

packages/web-components/fast-router/src/links.ts
```diff
@@ -0,0 +1,88 @@
...
+            target.hasAttribute("data-router-ignore")
+        ) {
+            return info;
+        }
+
+        if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) {
+            return info;
+        }
+
+        const href = target.getAttribute("href");
```
janechu, line 57: Maybe we should add a check for href above to prevent my previous comment about this potentially being a button submit?

packages/web-components/fast-router/src/links.ts
```diff
@@ -0,0 +1,88 @@
...
+        info.anchor = target;
+        info.href = href;
+
+        const leftButtonClicked = event.which === 1;
+        const isRelative = href && !(href.charAt(0) === "#" || /^[a-z]+:/i.test(href));
+
+        info.shouldHandleEvent = leftButtonClicked && !!isRelative;
```
janechu, line 64: should we handle relative links? It might make sense if we're going prev/next for history.

janechu:
Disclaimer on my comments, I haven't gotten to the recognizer yet.

EisenbergEffect:
@janechu FYI on the recognizer. Read the acknowledgements first. That piece of code has been around for a long time and tested in many thousands of apps 😄 Actually, the link handler has as well. I need to update the ack to reflect that.

EisenbergEffect:
Also, I should mention that quite a bit of this is going to evolve pretty significantly over the next 2-3 months. So, while I may not respond to comments directly, I'll be taking that all into account as I work through the process of moving this from prototype to production ready.

packages/web-components/fast-router/ACKNOWLEDGEMENTS.md
```diff
@@ -0,0 +1,3 @@
+# Acknowledgements
+
+A huge thank-you to Ember for the original implementation of the route recognizer so many years ago. That implementation was evolved by Aurelia 1 for its purposes and then later by Aurelia 2 for its purposes. FAST has continued that progression by leveraging the improvements made along the way and then customizing them for its own needs, with much appreciation for the innovation and hard work of these mentioned forerunners.
```
awentzel, line 3: Linking to Ember in a new window may help our indexing rank for SEO as it helps to establish credibility.

awentzel (approved):

packages/web-components/fast-router/ACKNOWLEDGEMENTS.md
EisenbergEffect: Adding some links now.

packages/web-components/fast-router/api-extractor.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+
```
chrisdholt, line 3: unlikely that it causes issues but these spaces seem suspect

chrisdholt (approved):
This looks good! I'm excited to get this out!

packages/web-components/fast-router/api-extractor.json
EisenbergEffect: Shouldn't be a technical issue but it is interesting that the space is there.

PR #4001, opened 2020-10-03
feat: Add tagged template vscode extension

radium-v:
# Description
This PR adds the source for a Visual Studio Code extension. This extension enables syntax highlighting for our tagged template literals (`html` and `css`).

## Motivation & context
Extensions are available on the VSCode Marketplace though there are compatibility issues in some cases. The goal of this project is to provide an extension that works with both of our template formats, and doesn't break when used for other similar web component projects.

### What this solves:
- syntax highlighting for `html` and `css` template literals

### What this doesn't solve:
- Theme support for all scopes. Missing scopes are still missing and should be added regardless in the theme extension. For example, the theme I use, [Boxy Tomorrow](https://marketplace.visualstudio.com/items?itemName=trongthanh.theme-boxythemekit), has pretty reliable scope support. The default themes for VSC... don't.
- Custom tagged template literals. Right now, CSS and HTML are the only available functions. Of course, we can add on to the extension if another tagged template literal format comes along.
- Error/linting support. This should be handled with `eslint`. It won't fix your errors, but it should make your errors easier to spot.

### What's still missing:
- Proper test integration. See [Testing Extensions](https://code.visualstudio.com/api/working-with-extensions/testing-extension) for documentation on how we should be able to implement a VSC-approved test workflow.
- Extension publishing. The official channels should be used to make sure the extension is published to the Marketplace with the proper administrative overhead. Here are the docs for [Publishing extensions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension) and setting up [Continuous Integration](https://code.visualstudio.com/api/working-with-extensions/continuous-integration).
- A good name. I went with `fast-tagged-templates` but the fully-qualified package name will be a username + a unique package name.

### What else:
- This package doesn't and shouldn't need to be published to NPM. An Azure pipeline should be created to handle publishing to the VSC Marketplace.
- The future pipeline should be a push-button solution for FAST Collaborators, which can have releases independent of our regular cycle. It should probably be handled the same way [other Microsoft-published extensions are handled](https://marketplace.visualstudio.com/publishers/Microsoft).
- Once the extension is published to the Marketplace, we can add it to the `.vscode/extensions.json` under the `recommendations` field.

### Installing locally
Since we don't have a pipeline or account set up to publish this, you can copy the `fast-tagged-templates` to your extensions folder (its location varies by platform - I included a link to the docs in the README).

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tagged-templates/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+## [0.1.0] - 2020-10-01
+
+
+### Added
+- Initial source
```
chrisdholt, line 10: This will automatically version *again* when we go to publish to 0.2.0 and the changelog will automatically be created. Do we want to rev to 0.2.0?

packages/tooling/fast-tagged-templates/package.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name": "fast-tagged-templates",
```
chrisdholt, line 2: I like the name, I think this was the primary outstanding question I had. 

I will say that for the package name, I don't think these have to match in this case. I'd prefer an additional layer to note this as either an extension or plugin. Without that, this could mean quite a bit in the tooling repo. While I think in most all cases it makes sense for our packages to share a folder name and `package.json` name, there are reasons for deltas. One similar delta is the eslint plugin. Eslint looks for a specific pattern of name so we have a mismatch there. Similarly, since this is a vscode extension and vscode uses this for the *name* of the extension, I don't know that we have to have the package name be the same as the name in the `package.json`. 

TLDR;

Can we rename the package to be something a bit more verbose and specific to this being an extension? I think clarity will help in the context of the repo. I think the name here is great for the official name of the extension itself and would be clear in the context of vscode as an extension.

scomea (approved):

packages/tooling/fast-tagged-templates/package.json
chrisdholt: @radium-v If we can get the name updated here I think this is ready to publish beyond just a draft.

packages/tooling/fast-tagged-templates/CHANGELOG.md
radium-v: That would only happen if this extension is treated as a regular node package. This extension should be excluded from the normal publishing workflow, and handled independently.

packages/tooling/fast-tagged-templates/package.json
EisenbergEffect: We've got a meeting coming up to discuss some productivity features for a FAST VS Code plugin. I'm wondering if we should wrap all of this up in the same plugin. In other words, should we rename this to basically be "the" FAST plugin, and then just place all features together so customers have a single plugin to install for the full experience?

chrisdholt (approved):

awentzel (approved):

EisenbergEffect:
@radium-v Can you create another issue to track work on getting this published, and perhaps setting up a pipeline for the publish process? /cc @awentzel Not sure if this will fit in with your new job but it would definitely benefit Bobby's team if their using FAST in VS Code. No idea if it's even possible to have a GH action handle this.

PR #4003, opened 2020-10-05
fix: corrects keyboard behavior of tree view and tree view item.

### LINKED ISSUE #3882: tabbing stops (or loops back) after second tree-view in storybook fixture in Firefox
Tabbing through tree-view components in the storybook fixture starts looping back to the beginning tree-view after the second tree-view in Firefox (80.0.1 on MacOs). This works fine in Edge.

Example:

1. Run storybook in fast-components
2. Click on tree-view
3. Start tabbing until you see it tabbing through tree-views on the right
4. Observe that after the second tree-view it loops through to the first tree-view on the next tab press

Expect it tab to move to the third, fourth and so on tree-view components until it reaches the bottom

Happens in Firefox (80.0.1) on MacOS

nicholasrice:
# Description
There are 3 issues this PR fixes:
1. focus getting looped in Firefox
2. Multiple tab stops per Tree View
3. Shift + tab taking the user back to the wrong item

This change removes tabindex completely from the TreeView for two reasons:
1. TreeView only accepted focus to redirect focus to the item that *should* be focused. We can instead just keep the item in the tab sequence
2. Having the TreeView focusable was causing a strange focus looping behavior which effectively trapped keyboard focus

It also changes the TreeItems to only have a `[tabindex]` attribute when they are focusable; if the item is not focusable, `[tabindex]` is removed completely (opposed to setting it to `-1`). The reason for this is that a slotted custom element with `[tabindex=0]` is not in the sequential tab order if it is slotted into a custom element with `[tabindex=-1]`. See https://webcomponents.dev/edit/ylwdCUY3qLoP3fpkc1J8 for an example.

closes #3882 

## Issue type checklist

## Process & policy checklist


marjonlynch (approved):

marjonlynch:
Smoked in Edge and Firefox, works great

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -140,19 +150,19 @@ export class TreeItem extends FASTElement {
...
-            return;
-        }
+    /**
+     * @deprecated - no longer needed.
+     * @param e - Event object
+     */
+    /* eslint-disable-next-line */
+    public handleFocus = (e: Event): void => {};
 
-        this.focusable = false;
```
SethDonohue, line 154: We are using this kind of focusing process on other components, or we have at least. Should we consider staying away from this sort of pattern in Web components now?

I recall using it on a react grid component, so just curious as to your thoughts on using this type of pattern elsewhere in web component world.

SethDonohue:
Smoked! Works well 👍

SethDonohue (approved):

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
nicholasrice: I don't know that it is a universal rule, the problem only exists for for custom elements that are nested in other custom elements, where the parent has `[tabindex=-1]` and the child has `[tabindex=0]`. Non-nested scenarios work as normal

chrisdholt (approved):

Issue #3930, opened 2020-09-18
Streamline process for adding scenarios to Component explorer site


janechu:
# Abstract

The creation of component scenarios for the component explorer site is unintuitive and should be made easier for contributors.

## Requirements

Ideally instead of writing a `DataDictionary` which is inherently difficult, the scenarios should instead be individual `.html` files located in a scenarios folder. This way a node script can be run to scrape them for content and inject them into the site.

- Create a Node.js script to read contents from a `scenarios` folder in `@microsoft/fast-components` package for each component
- Derive the scenario name from the file name
- Run this Node.js script during build of the component explorer
- Convert the internals of the Node.js script to a `DataDictionary`

PR #4005, opened 2020-10-05
chore: convert configuration files to html

janechu:
# Description

This converts the config files into `HTML` format, each components will have a folder called "scenarios".

The structure requires the following:
- Each scenario will be denoted its place in the select dropdown by the number in the file name (eg. `scenario.1.html` comes before `scenario.2.html`).
- The name of the option is recorded as a string between inline comments `<!-- scenario-title:start -->` and `<!-- scenario-title:end -->`.
- The contents of the scenario, which will choose a root tree item from the first item it comes across, will be selected between the inline comments `<!-- scenario:start -->` and `<!-- scenario:end -->`.

## Motivation & context

Begins to solve #3930

## Issue type checklist

## Process & policy checklist


nicholasrice:
@janchu what is the reason that you chose to use HTML comments to define the title and scenario start / end? It strikes me that there may be some easier ways to parse this information out that doesn't rely on functional comments (full transparency, I'm not a big fan of code comments that do things):
1. Use the title attribute on the scenario element
2. Use a `<template>` element to define the scenario

for example:

```html
// my-scenario.1.html
<template title="My Scenario">
  <fast-anchor>I'm an anchor!</fast-anchor>
</template>
```

nicholasrice:
Another thought on the above, if we went with the  template element approach you probably wouldn't need multiple HTML files for each component.

janechu:
@nicholasrice sure, I initially had this as a single file but I thought it would be easier to visually parse changes with one scenario per file. I can change that to a single file. I can make the other change to use `<template>` as well, I was avoiding having anything functional other than the actual HTML for the scenario (hence the use of comments) in these HTML files that would be used for parsing but I can see that the `<template>` could make sense here, and less prone to parsing issues.

nicholasrice:
Cool yea, single-file was just a thought I had that might be easier with `<template>` but I'm also happy with multiple files; whatever makes sense.

nicholasrice (approved):

scomea (approved):

EisenbergEffect (approved):

EisenbergEffect:
I like how this turned out in the end. Very nice 😄

PR #4007, opened 2020-10-06
fix: ensure tree item selection is only managed in tree view

### LINKED ISSUE #4004: Selected item deselects on click in TreeItem.ts
The selected treeitem deselects itself when the user clicks on it.

With a TreeView I was able to repro this
1. select an unselected treeitem. It should be selected
2. select the treeitem again. The treeitem wil be unselected

I expected the treeitem to still be selected.

Observed in Windows, Google Chrome

chrisdholt:
# Description
closes #4004 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -287,7 +287,6 @@ export class TreeItem extends FASTElement {
     }
 
     private handleSelected(e?: Event): void {
-        this.selected = !this.selected;
         this.$emit("selected-change", e);
     }
 
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
@@ -341,7 +341,7 @@ describe("TreeItem", () => {
             await disconnect();
         });
 
-        it("should toggle the selected state when the component is clicked", async () => {
+        it("should NOT toggle the selected state when the component is clicked", async () => {
             const { element, connect, disconnect } = await setup();
             const nestedItem = document.createElement("fast-tree-item");
 
@@ -354,15 +354,7 @@ describe("TreeItem", () => {
 
             await DOM.nextUpdate();
 
-            expect(element.selected).to.equal(true);
-            expect(element.getAttribute("aria-selected")).to.equal("true");
-
-            element.click();
-
-            await DOM.nextUpdate();
-
-            expect(element.selected).to.equal(false);
-            expect(element.getAttribute("aria-selected")).to.equal("false");
+            expect(element.selected).to.equal(undefined);
 
             await disconnect();
         });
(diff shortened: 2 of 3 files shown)


packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -287,7 +287,6 @@ export class TreeItem extends FASTElement {
     }
 
     private handleSelected(e?: Event): void {
-        this.selected = !this.selected;
         this.$emit("selected-change", e);
```
nicholasrice, line 290: It seems the item will still emit a "selected-change" event even though the selected state isn't actually changing, is that accurate? If so, that seems wrong to me. It should only emit a change event if it is actually changing

chrisdholt: I think that makes sense - I think clarity of where we want to handle selection will help. The selected state here was somewhat duplicative to the logic which intended to prevent change in the event the same tree item was selected. Changing here requires us to know about the previous selected state in comparison to other items. Not firing a change event here means that the click and keydown events would move to the tree. I'm fine with that but would love confirmation that folks are good with that shift before I rework.

nicholasrice: I totally agree, focus navigation is split between the tree and the items as well.

In the future I'm starting to think that the Tree should handle all focus navigation *and* selection because:
1. Event delegation makes consolidating event handlers to a single element more efficient
2. Reduce coordinating logic between items and elements
3.  Replacing items with custom item implementations becomes easier because business logic is consolidated.

If that is something we want to aim for, I think we could move this piece now by removing the click handler from the item and having the tree manage selection. *setting* selected would emit a `selected-change' event but the tree-view would disregard the event because it already handled selection; the event would just be for authors.

scomea: Maybe rename the event to something like "item-invoked" would make it clear that the item is not controlling it's own selected state?

nicholasrice: I think I would consider *renaming* an event a breaking change. To me (and this is up for debate), the emitted events could be relied upon but authors and therefore part of the public "API"

scomea: Yes, I'd agree that it would be a breaking change.

chrisdholt: Revisiting this - do we have a desired path for the change here? A couple of proposals; to be honest, I'm open.

chrisdholt: @nicholasrice pinging again - I agree we don't want to necessarily rename here as part of this. Curious if we still feel like the above proposed solution is the right way to address this.

nicholasrice: I still believe so, yes.

EisenbergEffect:
Pinging this one as it's pretty stale. Do we need further discussions to resolve or just need someone to pick up the work?

PR #4023, opened 2020-10-12
fix: tabbing disabled for when linkedDataControl used in form

### LINKED ISSUE #3885: Trapped focus in tooling React Form component add linked data control

When tabbing through a `<Form />` which includes a default control for linked data, tabbing traps focus and the control does not allow tab to move to the next focusable item.

1. Go to explore.fast.design
2. Tab though the `<Form />` on the right
3. Cannot re-focus after the default slot control has been reached

Tab should not be trapped and should move to the next focusable item.

marjonlynch:
# Description

Removed preventDefault() call in keydownHandler in linkedDataControl when key is tab, this prevented normal tabbing from occurring.
closes #3885 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
+++ b/packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
@@ -238,7 +238,7 @@ class LinkedDataControl extends React.Component<
                         searchTerm: "",
                     });
                 }
-                // Tab performs an autocompete if there is a single schema it can match to
+                // Tab performs an auto-complete if there is a single schema it can match to
             } else if (e.keyCode === keyCodeTab) {
                 const normalizedValue = e.currentTarget.value.toLowerCase();
                 const matchedSchema = this.lazyMatchValueWithASingleSchema(
@@ -246,6 +246,8 @@ class LinkedDataControl extends React.Component<
                 );
 
                 if (typeof matchedSchema === "string") {
+                    // prevent normal tabbing when single schema matched
+                    e.preventDefault();
                     this.setState({
                         searchTerm: this.props.schemaDictionary[matchedSchema].title,
                     });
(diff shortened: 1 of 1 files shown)


packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
```diff
@@ -240,8 +240,6 @@ class LinkedDataControl extends React.Component<
                 }
                 // Tab performs an autocompete if there is a single schema it can match to
             } else if (e.keyCode === keyCodeTab) {
```
nicholasrice, line 242: I wonder if this should conditionally `preventDefault()`. Not 100% sure but it seems that in cases when autocomplete is successful, moving focus may be un-intuitive.

janechu: I agree, maybe you could simply move the `preventDefault()` into the `if` that checks for a matchedSchema below.

marjonlynch: Done and done, left a comment in the code. This now requires the user to hit enter to continue tab navigation though. I'm assuming that is the behavior we want in combination with tab completion functionality

packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
```diff
@@ -238,16 +238,16 @@ class LinkedDataControl extends React.Component<
...
-                e.preventDefault();
-
                 const normalizedValue = e.currentTarget.value.toLowerCase();
                 const matchedSchema = this.lazyMatchValueWithASingleSchema(
                     normalizedValue
                 );
 
                 if (typeof matchedSchema === "string") {
+                    // prevent navigating away by tab when single schema matched
+                    e.preventDefault();
```
janechu, line 250:
```suggestion
                    e.preventDefault();

```

janechu (approved):

packages/tooling/fast-tooling-react/src/form/controls/control.linked-data.tsx
marjonlynch: I rebased and just made the change at the same time. so done and done

nicholasrice (approved):

PR #3922, opened 2020-09-17
feat: add data-grid spec

scomea:
# Description

This change adds a spec for a data grid web component. 

The working/prototype branch for the data grid component is here:
https://github.com/microsoft/fast/tree/users/scomea/wc-datagrid

## Motivation & context
We need a data grid component.

https://github.com/microsoft/fast/issues/3750

## Issue type checklist
- [ ] **Chore**: A change that does not impact distributed packages.
- [ ] **Bug fix**: A change that fixes an issue, link to the issue above.
- [x] **New feature**: A change that adds functionality.

**Is this a breaking change?**
- [ ] This change causes current functionality to break.

**Adding or modifying component(s) in `@microsoft/fast-components` checklist**
- [ ] I have added a new component
- [ ] I have modified an existing component
- [ ] I have updated the [definition file](https://github.com/Microsoft/fast/blob/master/packages/web-components/fast-components/CONTRIBUTING.md#definition)
- [ ] I have updated the [configuration file](https://github.com/Microsoft/fast/blob/master/packages/web-components/fast-components/CONTRIBUTING.md#configuration)

## Process & policy checklist
- [ ] I have added tests for my changes.
- [ ] I have tested my changes.
- [x] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

PR #4029, opened 2020-10-13
feat: add a data grid web component

scomea:
# Description
Adds a data grid component to our collection.

## Motivation & context
Need a FAST data grid component.
https://github.com/microsoft/fast/pull/3922

## Issue type checklist
- [ x] **New feature**: A change that adds functionality.

## Process & policy checklist


packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
```diff
@@ -0,0 +1,310 @@
...
+`;
+
+const defaultHeaderCellContentsTemplate: ViewTemplate = html<DataGridCell>`
+    <template>
+        ${x =>
+            x.columnDefinition === null
+                ? null
+                : x.columnDefinition.title === undefined
+                ? x.columnDefinition.columnDataKey
+                : x.columnDefinition.title}
```
janechu, line 34: @EisenbergEffect here's a thought, would we conceivably want a kind of logging system during development to determine if data structures are missing for a given component? It seems like that would be useful here.

packages/web-components/fast-foundation/src/data-grid/data-grid-row.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    private updateItemTemplate(): void {
+        this.activeCellItemTemplate =
+            this.rowType === DataGridRowTypes.default
+                ? this.cellItemTemplate !== undefined
+                    ? this.cellItemTemplate
+                    : this.defaultCellItemTemplate
+                : this.headerCellItemTemplate !== undefined
+                ? this.headerCellItemTemplate
+                : this.defaultHeaderCellItemTemplate;
```
janechu, line 287: I'm a bit torn here, because while I think nested ternarys are very difficult to parse, I also think you end up doing additional checks. I prefer the clarity of removing the nested ternarys though from a maintenance perspective.

```suggestion
            this.rowType === DataGridRowTypes.default && this.cellItemTemplate !== undefined
                ? this.cellItemTemplate
                : this.rowType === DataGridRowTypes.default && this.cellItemTemplate === undefined
                ? this.defaultCellItemTemplate
                : this.headerCellItemTemplate !== undefined
                ? this.headerCellItemTemplate
                : this.defaultHeaderCellItemTemplate;
```

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -0,0 +1,543 @@
...
+    public static generateColumns = (row: object): ColumnDefinition[] => {
+        const definitions: ColumnDefinition[] = [];
+        const properties: string[] = Object.getOwnPropertyNames(row);
+        properties.forEach((property: string, index: number) => {
+            definitions.push({
+                columnDataKey: property,
+                gridColumn: `${index}`,
+            });
+        });
+        return definitions;
```
janechu, line 116:
```suggestion
        return Object.getOwnPropertyNames(row).map((property: string, index: number) => {
            return {
                columnDataKey: property,
                gridColumn: `${index}`,
            };
        });
```

packages/web-components/fast-components/src/data-grid/scenarios/index.html
```diff
@@ -0,0 +1,5 @@
+<template title="Default">
+    <div>
+        <fast-data-grid></fast-data-grid>
```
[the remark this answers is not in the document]
scomea: Sweet, I was wondering how to do that.

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
EisenbergEffect: Yes, that would be nice. We'd need to somehow address the issue of debug builds first I think. Otherwise, we could very quickly blow out our library size targets and hurt perf in critical areas.

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.template.ts
```diff
@@ -0,0 +1,16 @@
...
+ * The template for the {@link @microsoft/fast-foundation#DataGridCell} component.
+ * @public
+ */
+export const DataGridCellTemplate = html<DataGridCell>`
+    <template
+        tabindex="-1"
+        role="${x => (x.cellType === "columnheader" ? "columnheader" : "gridcell")}"
+        class="${x => (x.cellType === "columnheader" ? "column-header" : "")}"
+    >
+        <slot part="cellSlot"></slot>
```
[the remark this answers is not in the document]
nicholasrice: I believe we landed spinal case for HTML attributes and values.

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
```diff
@@ -0,0 +1,310 @@
...
+     * The type of cell
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: cell-type
+     */
+    @attr({ attribute: "cell-type" })
+    public cellType: DataGridCellTypes;
+    private cellTypeChanged(): void {
+        if ((this as FASTElement).$fastController.isConnected) {
```
nicholasrice, line 65: You shouldn't need to `this as FASTElement` here - I think something in your environment is messed up.

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    public handleFocusin(e: FocusEvent): void {
+        if (this.isActiveCell || this.columnDefinition === null) {
+            return;
+        }
+
+        this.isActiveCell = true;
+
+        switch (this.cellType) {
+            case undefined:
```
nicholasrice, line 160: Is there a case where we *don't* want to handle focusin? If not, should this just use the `default` case?

packages/web-components/fast-foundation/src/data-grid/data-grid-row.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/**
+ * A Data Grid Row Custom HTML Element.
+ *
+ * @public
+ */
+export class DataGridRow extends FASTElement {
+    private static cellQueryString =
+        '[role="cell"], [role="gridcell"], [role="columnheader"]';
```
nicholasrice, line 37: is whitespace needed?

packages/web-components/fast-foundation/src/data-grid/data-grid-row.ts
```diff
@@ -0,0 +1,293 @@
...
+     * String that gets applied to the the css gridTemplateColumns attribute for the row
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: grid-template-columns
+     */
+    @attr({ attribute: "grid-template-columns" })
+    public gridTemplateColumns: string;
+    private gridTemplateColumnsChanged(): void {
+        if ((this as FASTElement).$fastController.isConnected) {
```
nicholasrice, line 49: same for all these `this as FASTElement` - doing the casting should not be necessary.

packages/web-components/fast-foundation/src/data-grid/data-grid-row.ts
```diff
@@ -0,0 +1,293 @@
...
+    /**
+     * The index of the row in the parent grid.
+     * This is typically set programmatically by the parent grid.
+     *
+     * @public
+     */
+    @observable
+    public rowIndex: number;
+    private rowIndexChanged(): void {
+        if ((this as FASTElement).$fastController.isConnected) {
```
nicholasrice, line 124: you can probably remove this method?

packages/web-components/fast-foundation/src/data-grid/data-grid-row.ts
scomea: In the role names?  These are the actual names - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Gridcell_Role

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.template.ts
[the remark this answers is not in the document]
scomea: So "cellRegion"?

packages/web-components/fast-foundation/src/data-grid/data-grid-row.ts
scomea: Oh, sorry, read that completely backwards.  yes I can remove the spaces.

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
scomea: yeah, I see that, my VS does the right thing though.

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
scomea: yep

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.template.ts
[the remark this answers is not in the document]
SethDonohue: I have been seeing spinal case for part names like `expanded-icon`, `collapsed-icon`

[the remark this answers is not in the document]
janechu: @scomea I think "cell-region" might work here.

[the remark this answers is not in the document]
scomea: Actually, if this is the component's default slot does it need a part name?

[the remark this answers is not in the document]
janechu: @chrisdholt @nicholasrice can someone answer this? I've no idea.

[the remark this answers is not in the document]
EisenbergEffect: No name is needed if it's intended to be the default slot. Also, we'd want to always use the default slot before adding named slots.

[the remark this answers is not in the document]
EisenbergEffect: Oops. Sorry, I misread. I don't know if a `part` name directly on a slot will have any effect. I'm not sure if you can style that slot directly. If it is possible, and it's the default slot then perhaps we should call the part itself "default" as well?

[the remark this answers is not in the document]
chrisdholt: The part should go on the same level as the class. The intent here is that we style with the class, someone can reach in and change that with the part. There are plenty of methods to style that slotted content from a user perspective. The part itself should be on the DOM inside the component rather than the slot itself.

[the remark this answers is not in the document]
scomea: Ok, so since the slot the single child of the template root no need for a partname

packages/web-components/fast-foundation/src/data-grid/data-grid-cell.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * @internal
+     */
+    public connectedCallback(): void {
+        super.connectedCallback();
+
+        this.addEventListener("focusin", this.handleFocusin);
+        this.addEventListener("focusout", this.handleFocusout);
+        this.addEventListener("keydown", this.handleKeydown);
```
janechu, line 122: As a follow up, consider referencing the `@microsoft/fast-web-utilities` package, I have just created a PR that exports these event type strings https://github.com/microsoft/fast/pull/4161

scomea: Done.

packages/web-components/fast-foundation/src/data-grid/README.md
```diff
@@ -0,0 +1,64 @@
...
+custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/src/data-grid/README.md
+---
+The `fast-data-grid` component is used to display tabular data.  The `fast-data-grid-row` and `fast-data-grid-cell` components are typically created programmatically by the parent grid but some authors may find it useful to create them manually.
+
+## Usage
+
+```html live
+<fast-design-system-provider use-defaults>
+    <fast-data-grid>
+    </fast-data-grid>
```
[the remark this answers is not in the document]
scomea: updated with simple examples

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -0,0 +1,608 @@
...
+    private observer: MutationObserver;
+
+    private rowindexUpdateQueued: boolean = false;
+    private columnDefinitionsStale: boolean = true;
+
+    private generatedGridTemplateColumns: string = "";
+
+    constructor() {
+        super();
+    }
```
janechu, line 292: do you need the constructor if there's no code in it?

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -0,0 +1,608 @@
...
+        this.rowsPlaceholder = document.createComment("");
+        this.appendChild(this.rowsPlaceholder);
+
+        this.toggleGeneratedHeader();
+
+        this.rowsRepeatBehavior = new RepeatDirective(
+            x => x.rowsData,
+            x => x.rowItemTemplate,
+            { positioning: true }
+        ).createBehavior(this.rowsPlaceholder);
```
janechu, line 313: Would this be the same as:

```suggestion
        this.rowsRepeatBehavior = repeat(
            x => x.rowsData,
            x => x.rowItemTemplate,
            { positioning: true }
        ).createBehavior(this.rowsPlaceholder);
```

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -0,0 +1,608 @@
...
+     * @internal
+     */
+    public connectedCallback(): void {
+        super.connectedCallback();
+
+        if (this.rowItemTemplate === undefined) {
+            this.rowItemTemplate = this.defaultRowItemTemplate;
+        }
+
+        this.rowsPlaceholder = document.createComment("");
```
janechu, line 304: Could you explain why this is a comment?

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -0,0 +1,608 @@
...
+
+        this.rowsRepeatBehavior = new RepeatDirective(
+            x => x.rowsData,
+            x => x.rowItemTemplate,
+            { positioning: true }
+        ).createBehavior(this.rowsPlaceholder);
+
+        this.$fastController.addBehaviors([this.rowsRepeatBehavior!]);
+
+        this.addEventListener("row-focused", this.handleRowFocus);
```
janechu, line 317: Can we change this to "rowfocused", I think most events are just lowercase without delineation see: https://developer.mozilla.org/en-US/docs/Web/Events

janechu (commented):
Just a few questions for clarity and some nits.

chrisdholt (approved):
Thanks @scomea, this is a ton of work - excited to get it out into the hands of the community and get some feedback!

janechu (approved):

PR #4031, opened 2020-10-14
fix: leverage FASTController.addStyles to attach css custom property target sheet

nicholasrice:
# Description
This change adjusts the way the DesignSystemProvider's CSSCustomPropetyTarget is determined, expanding the usage of constructible stylesheets and unifying behavior between open and closed shadow roots.

The existing approach of directly manipulating the `adoptedStyleSheets` of the `shadowRoot` to append a CSSCustomProperty target relies on an open shadow root. When the shadowRoot is closed, CSS custom properties are applied via the `HTMLElement.style.setProperty()` API. We can improve this by utilizing the Controller's `addStyles()` API to perform the style attachment for us so that we don't require access to the shadowRoot. The DSP will still use the `HTMLElement.style.setProperty()` API in cases where constructible styles are not supported.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
+++ b/packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
@@ -325,9 +325,9 @@ export class DesignSystemProvider extends FASTElement
 
         if (supportsAdoptedStylesheets) {
             const sheet = new CSSStyleSheet();
-            sheet.insertRule(":host{}");
+            const index = sheet.insertRule(":host{}");
             this.$fastController.addStyles(ElementStyles.create([sheet]));
-            this.customPropertyTarget = (sheet.rules[0] as CSSStyleRule).style;
+            this.customPropertyTarget = (sheet.rules[index] as CSSStyleRule).style;
         } else {
             this.customPropertyTarget = this.style;
         }
--- a/packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
+++ b/packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
@@ -2,6 +2,7 @@ import {
     attr,
     Behavior,
     customElement,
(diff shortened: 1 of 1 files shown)


packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
```diff
@@ -322,15 +323,11 @@ export class DesignSystemProvider extends FASTElement
...
             const sheet = new CSSStyleSheet();
-            sheet.insertRule(":host{}");
-            (this.shadowRoot as any).adoptedStyleSheets = [
-                ...(this.shadowRoot as any).adoptedStyleSheets,
-                sheet,
-            ];
-
-            this.customPropertyTarget = (sheet.rules[0] as CSSStyleRule).style;
+            const index = sheet.insertRule(":host{}");
+            this.$fastController.addStyles(ElementStyles.create([sheet]));
```
EisenbergEffect, line 329: You can remove the custom style sheet creation here and just do this:

```ts
this.$fastController.addStyles(ElementStyles.create([":host{}"]));
```

ElementStyles will create the sheet for you. Not a big deal but would reduce the code a bit.

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
nicholasrice: I need to store a reference to the sheet so I can write properties, can I get that from ElementStyles?

EisenbergEffect: Ah, good point. The sheet itself isn't exposed since it's an abstraction over sheets. Probably keep the code you've got in this case 😄

chrisdholt (approved):

Issue #4020, opened 2020-10-12
fast-slider sets value to NaN if developer tools open


MarcSkovMadsen:
I'm trying to use the `fast-slider` and I see several bugs. In order to understand these bugs I use Chromes developer tools. But if I drag the slider while the developer tools are shown

- I cannot drag the slider. It's fixed.
- The aria-valuenow is set to NaN

![image](https://user-images.githubusercontent.com/42288570/95706582-86e86f80-0c57-11eb-9d70-3e4ce8da7854.png)

```html
<html>
    <head>
        <script type="module" src="https://unpkg.com/@microsoft/fast-components"></script>
    </head>
<body style="margin:0px">
    <fast-design-system-provider use-defaults style="height:100%;">
        <fast-slider style="width:200px" min="0" max="100" step="1" value="10"></fast-slider>
    </fast-design-system-provider>
</body></html>
```

![fast-slider-not-working-developer-tools](https://user-images.githubusercontent.com/42288570/95706547-71734580-0c57-11eb-9821-096ffa72f546.gif)

PR #4033, opened 2020-10-14
fix: fast-slider sets value to NaN if developer tools open

robarbms:
# Description

When setting the browser in emulation of a touch device, the slider would error out and return NaN for the value.
Added an error check that resolves NaN values to the minimum value.
Added TouchEvent support for touch devices.

## Motivation & context

Added touch support and error checking to slider values.
[Open issue](https://github.com/microsoft/fast/issues/4020)

Repro steps. Open developer tools in chrome. Select device emulation and drag the slider. It's easiest to test if you select iPad Pro but any touch device will repro. It will jump towards the beginning of the slider and can't be dragged.

Broken example:
[FAST slider](https://explore.fast.design/components/fast-slider)

Working example: 
[Fixed FAST slider](http://robdor:6006/?path=/story/slider--slider)

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -354,16 +354,24 @@ export class Slider extends FormAssociated<HTMLInputElement>
...
-        // update the value based on current position
-        const eventValue: number =
-            this.orientation === Orientation.horizontal
-                ? e.pageX - this.trackLeft
-                : e.pageY;
+        let eventValue: number = 0;
+
+        if (e instanceof MouseEvent) {
+            eventValue =
+                this.orientation === Orientation.horizontal
```
marjonlynch, line 366: Seems like you could set an object to be the either e or e.touches[0] and then use the same logic to get the calculated value, would be a little more compact and only 1 place for the calc

robarbms: I considered doing that as a source event or possibly a getPosition method but I wasn't sure how to type it. I'll play around with it.

robarbms: Created a source event that points to e.touches when it's a touch event, otherwise it uses the event.

marjonlynch: Yes! love it, just smoking again and will approve

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -354,16 +354,17 @@ export class Slider extends FormAssociated<HTMLInputElement>
...
      *  Handle mouse moves during a thumb drag operation
      */
-    private handleMouseMove = (e: MouseEvent): void => {
+    private handleMouseMove = (e: MouseEvent | TouchEvent): void => {
         if (this.readOnly || this.disabled || e.defaultPrevented) {
             return;
         }
 
         // update the value based on current position
+        const sourceEvent = e instanceof TouchEvent ? e.touches[0] : e;
```
nicholasrice, line 363: I *think* this is going to throw a ReferenceError in Safari because `TouchEvent` is not supported. Can you test this in Safari to confirm?

marjonlynch: I'm not seeing any exceptions/errors in safari
Screen Shot 2020-10-15 at 11 53 24 AM

marjonlynch (approved):

packages/web-components/fast-foundation/src/slider/slider.ts
marjonlynch: I'm using Safari 14 though.

robarbms: It looks like Safari supports touch events in this way according to their documentation. I need to put my build somewhere public so that I can test it on my iPad though.

nicholasrice: I wonder if babel is aliasing that expression to `window.TouchEvent` or something: https://caniuse.com/touch claims Safari 14 does not support `TouchEvent`.

If you pop `"foo" instanceof TouchEvent` into the console of Safari you get a ReferenceError which is why I bring this up. I *do* see it referenced in [webkitjs](https://developer.apple.com/documentation/webkitjs/touchevent) documentation though so I'm a little confused. I'll build this and launch it in safari to see what the compiled source-code is

robarbms: Thanks. I dug up my old iPad mini, charged and updated it to the latest iOS. I tried setting up a public server on Azure but had login issues but realized I was over complicating it. So I set up IIS on my laptop and tried moving the storybook over but neither the one in fast-components or the one in the root directory seem to work. Is there actual files deployed to the disk or is storybook running only from memory? I tried adding an output directory to the build-storybook script in the package.json but that didn't seem to do anything at all.

nicholasrice: `start-storybook` is all in-memory I believe, though I could be mistaken. I also tried to find the relevant code in the generated files from `build-storybook` but couldn't, so there may be something wonky going on there.

I just confirmed it *does* throw a ReferenceError though, @marjonlynch you should see it when you start to drag the slider with the mouse.

nicholasrice: Assuming that this works behaviorally, you can avoid the runtime error by checking if TouchEvent exists in the window prior to checking the if the event is an instance of the TouchEvent type:

```ts
window.TouchEvent && e instanceof TouchEvent ? x : y;
```

robarbms: TypeScript won't let me. I had this same error when I tried to make it more robust with a conditional of (instanceof TouchEvent && TouchEvent.touches && TouchEvent.touches.length > 0). It treats it as if it might be an TouchEvent but one of the other cases is false. Is there a way around this in TypeScript? The only way I can think of doing it is to breaking them out into two statements like I had before and doing a check there. It's more verbose though.

nicholasrice: huh.... I would expect TypeScript to still restrict the type, not sure why it isn't.

I'd say in this case casting the type is okay:
```ts
const sourceEvent = window.TouchEvent && e instanceof TouchEvent ? e.touches[0] : e as MouseEvent;
```

Alternatively, I think you could test for *MouseEvent* instead:

```ts
const sourceEvent = e instanceof MouseEvent ? e : e.touches[0];
```

Both of those appease TypeScript

robarbms: That worked. I like the explicitness of the first way. Thanks.

chrisdholt (approved):

Issue #4035, opened 2020-10-14
bug:  Fast button component not added to tab queue in fast dialog by default


scomea:
Fast button elements that are children of a fast dialog don't appear in the dialog's tab queue because fast buttons are not detected as [tabbable](https://github.com/focus-trap/tabbable) because the button tag is not discovered because it exists in the shadow dom.

A possible solution is to mark fast-button elements with a tabindex of "0" by default so tabbable discovers it.  This may affect other web components that hide focusable elements in the shadow DOM and aren't marked wiht a positive tab index.

Example:

1. Create a modal fast dialog with several fast-button children
2. Try to tab between them, note that it doesn't work
3. Add tabindex="0" attributes to the buttons
4. Try tabbing between them, note that it does work

Fast buttons should show up in dialog tab queues by default.

scomea:
Related issue to be aware of - https://github.com/microsoft/fast/pull/4036

SethDonohue:
For now are we sticking with the author needing to add a non-negative tabindex to the fast-buttons for it to show in the dialog tab queue, or are we thinking of changing fast-button to have a tabindex="0" by default? Just curious if there has been any movement on this.

chrisdholt:
> For now are we sticking with the author needing to add a non-negative tabindex to the fast-buttons for it to show in the dialog tab queue, or are we thinking of changing fast-button to have a tabindex="0" by default? Just curious if there has been any movement on this.

I think when @EisenbergEffect lands the Provider DSL we'll be able to plug into that and leverage some of the benefits for our own trap focus implementation. The problem here is that we need to maintain a list of all HTML focusable elements as well as a list of our focusable custom elements, while also providing a method for users to register their own. Add to this the fact that someone may register our custom elements with a namespace other than "fast". I think the provider DSL will provide a hook for us to handle all of that fairly elegantly. For now, because that's not landed, we'll keep the code as is using `tabbable`, which requires tabindex="0" for custom elements.

m4thieulavoie:
@chrisdholt are you suggesting people manually add `tabindex="0"` in their own app for now, or that FAST handles that internally (by adding `tabindex="0"` if none was specified by the user)? If the latter, I'm happy to contribute if you want 👍

SethDonohue:
For now, we need to add tabindex="0" manually. The work he mentioned above with the Provider DSL will help with this issue once it is ready.

chrisdholt:
[...]

PR #4036, opened 2020-10-14
fix: dialog delays for a frame before finding tabbable children

scomea:
# Description
The dialog component uses [tabbable](https://github.com/focus-trap/tabbable) to build a local tab queue.  Currently tabbable is called on the dialog's connected() callback which is before child elements have had a chance to fully load.  As a result web components that should be in the tab queue have not yet updated their tabIndex values to "0" so they don't get added to the queue.  This change delays the call to tabbable until after a DOM update.

This does not resolve the issue that the button component isn't discovered by tabbable because the "button" tag is hidden in the shadow dom and the component itself does not have a non-negative tab index which is refererred to in https://github.com/microsoft/fast/issues/4035

## Motivation & context
Tabbable web components should show up in our dialog's tab queue.

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):

EisenbergEffect:
I think this approach works for the current scenarios but I wonder if it will run into other issues later. For example, what if the dialog has conditional rendering of its children and at some point during dialog interaction, new DOM gets conditionally rendered which has focusable controls. Will tabbable prevent those from being focused? I don't really know much about tabbable (other than I'd like to get rid of the 3rd party dependency if we can). If this scenario is a problem we may need to switch to using the `children` decorator on the template with a setting of `subtree: true` so that we can be notified when any child is added/removed and re-run tabbable.

scomea:
@EisenbergEffect Yeah, there are some holes.  Even the approach you suggest would not catch a case where a child had its tab index change from negative to positive after load or child change (our components do that).  We may need to run it on each keystroke.

EisenbergEffect:
The underlying mutation observer could be configured to callback on attribute changes as well. I've never actually used it in that mode and don't know the perf implications... Maybe it's not too bad since you don't have a ton of dialogs at once and they usually aren't too complicated. Perhaps we configure that and if the change is a dom structure change or a tab index change, then we re-run tabblable, otherwise ignore?

scomea:
I opened a tracking bug - https://github.com/microsoft/fast/issues/4038

nicholasrice (approved):

PR #4043, opened 2020-10-16
feat: support HTMLStyleElement from Controller.addStyles and Controller.removeStyles

### LINKED ISSUE #4042: Should Controller.addStyles and Controller.removeStyles support HTMLStyleElementInstances?
The FASTDesignSystemProvider requires programmatically setting CSS properties on a CSSStyleRule that selects the element host. In cases where constructable styles are supported, the provider constructs a new CSSStyleSheet instance and supplies that instance to `$fastController.addStyles()` during construction. In cases where constructible styles are *not* supported, the provider instead uses the `HTMLElement.style` API to apply CSS properties. While this works, there is a significant down-side to this approach because these properties can easily be overridden by an app author if the author sets the `style` attribute of the element instance.

It would be useful for this scenario if there was a way to attach styles to the element instance where a reference to the sheet could be held by the component instance (so that properties could be set).

One approach would be to support a  `HTMLStyleElement` argument in `Controller.addStyles` and `Controller.removeStyles` , where those APIs would prepend the element to the shadowRoot.
```ts
const element = document.createElement('style');
element.innerHTML = ":host { background: var(--my-custom-property)}";
this.$fastController.addStyles(element); // synchronously prepend element to shadowRoot so that HTMLStyleElement.sheet is not null
element.sheet.rules[0].setProperty("--my-custom-property", "red"); // Make element background red
```
Another approach would be to enable ElementStyles itself to support `HTMLStyleElement`. I don't like this approach as much though, because:
1. ElementStyles accepts "composable" styles (strings and CSSStyleSheet instances). HTMLStyleElement's aren't by nature composable because the element can only be in one place in the DOM.
2. ElementStyles *could* clone the node to make using HTMLStyleElements more composable, but I fear that isn't an intuitive behavior and would not solve the use-case above because the element reference would not be maintained.

nicholasrice:
# Description
Adds the capability to add and remove a HTMLStyleElement to a custom element instance via the `Controller.addStyles()` and `Controller.removeStyles()` API. This provides an opportunity to hold a reference to a `CSSStyleSheet` that is attached to the custom element's shadowRoot, even when the shadow root is closed.

closes #4042.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-element/src/controller.ts
+++ b/packages/web-components/fast-element/src/controller.ts
@@ -146,16 +146,24 @@ export class Controller extends PropertyChangeNotifier {
      * Adds styles to this element.
      * @param styles - The styles to add.
      */
-    public addStyles(styles: ElementStyles): void {
-        const sourceBehaviors = styles.behaviors;
+    public addStyles(styles: ElementStyles | HTMLStyleElement): void {
         const target =
             getShadowRoot(this.element) ||
             ((this.element.getRootNode() as any) as StyleTarget);
 
-        styles.addStylesTo(target);
+        if (styles instanceof HTMLStyleElement) {
+            target.prepend(styles);
+        } else {
+            const sourceBehaviors = styles.behaviors;
+            const target =
+                getShadowRoot(this.element) ||
+                ((this.element.getRootNode() as any) as StyleTarget);
 
-        if (sourceBehaviors !== null) {
-            this.addBehaviors(sourceBehaviors);
+            styles.addStylesTo(target);
+
+            if (sourceBehaviors !== null) {
+                this.addBehaviors(sourceBehaviors);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


scomea (approved):

packages/web-components/fast-element/src/controller.ts
```diff
@@ -146,33 +146,46 @@ export class Controller extends PropertyChangeNotifier {
...
         const target =
             getShadowRoot(this.element) ||
             ((this.element.getRootNode() as any) as StyleTarget);
 
-        styles.addStylesTo(target);
+        if (styles instanceof HTMLStyleElement) {
+            target.prepend(styles);
+        } else {
+            const sourceBehaviors = styles.behaviors;
+            const target =
```
EisenbergEffect, line 158: I think this can be removed since you have it above.

EisenbergEffect (approved):
Looks like a bit of accidental duplication that can be removed. Other than that, the approach seems fine to me.

PR #4047, opened 2020-10-19
fix: handle initially set value in FormAssociated

### LINKED ISSUE #4009: When binding :value in fast-text-field the initial value of the bound field is not set
When using `:value` to bind the value from a variable the initial value is not set to the variable's value.

Using the provided repro
1. On initial load the value of **Using :value => one** shows the placeholder value, and not the current value of `one`
2. Switching to another source and back to the initial source will correctly update the value

The initial value of the `fast-text-field` should be set to the field value if the value is not `undefined`/`null`/`empty`

#### Initial state
fast-text-field :value initial

#### Changed State
fast-text-field :value changed

* macOS
* Firefox and Chrome
* `fast-components`: 1.11.0, and `fast-element`: 0.19.0

This shares a repro with #4008 
Attached repro:
[fast-text-field-value-repro.zip](https://github.com/microsoft/fast/files/5335038/fast-text-field-value-repro.zip)

nicholasrice:
# Description
This PR adjusts the `connectedCallback` of the `FormAssociated` component to only assign the `initialValue` to the `value` property when the `value` itself is not already set. Prior to this change, if the `value` property is assigned before the `connectedCallback` is invoked, the `connectedCallback` will write-over the assigned value with the `initialValue` (often empty string).

It also adjusts the Slider slightly to accommodate the above change, whereas before, the Slider was relying on the `value` prop being the empty string assigned during `connectedCallback`.
fixes #4009

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):

EisenbergEffect:
Do we feel that we have enough tests on this? Should we have more tests for the individual form elements for this scenario?

nicholasrice:
Individual form elements extend the FormAssociated so I'd expect coverage of FormAssociated to cover the individual elements in most cases, though it is possible individual controls are doing things that break the behavior (such as the fix needed for Slider in this PR). In general I'd advocate for adding more tests.

Is that something that should happen here or in a follow up?

chrisdholt (approved):

EisenbergEffect:
@nicholasrice If you think it can be done in an hour or so, then I'd go for it and add to this PR. Otherwise, we can do it as a follow up at a later time.

PR #4050, opened 2020-10-20
fix: getDirection utility should not throw

scomea:
# Description
The getDirection utility function would call .closest() on the parent node of the element node passed in.  This would throw an exception when .parent is null which happens when the node is a child of a shadow root (there may be other cases too).  Since I'm not sure why we call this on the .parent rather than the node itself I changed the function to use the param directly.

I also removed the local version of getDirection from the AnchoredRegion element and pointed at the utility code instead.  

## Motivation & context
The function should not throw, and should work correctly even if the element passed to the function is at the root of a shadow dom.  

## Issue type checklist

## Process & policy checklist


SethDonohue (approved):

EisenbergEffect (approved):

nicholasrice (approved):
Not sure why it was parentElement as well, I would expect it to respect the elements own dir attribute.

scomea:
My theory is that is was copied from an implementation that called the function only after it had already checked the root object.

PR #4063, opened 2020-10-23
chore: fix broken cdn address

BrunoBlanes:
# Description
Removed an extra '@' from the CDN link.

## Motivation & context
Fixes the CDN address.

## Issue type checklist

## Process & policy checklist


BrunoBlanes:
I realize now that the reason for the second `@` is to escape the first character on Razor pages, however the page instructs the user to place the script on `index.html` and not on `_Layout.cshtml`. Therefore, a decision need to be made on where the script should be put and then update the link accordingly.

BrunoBlanes:
I think if we want to cover both scenarios, it should not contain the second character, because that gives a false idea of the original address. Escaping should be up to the developer.

EisenbergEffect:
Good points. This makes sense. Perhaps we make this change and then also add a `:::note` section that indicates that the extra `@` is needed when in a Razor view. Thoughts?

EisenbergEffect (approved):

chrisdholt (approved):
Agree on the note aspect - I think we've seen this change back and forth a few times. Adding a note here would be helpful :)

EisenbergEffect:
@BrunoBlanes Would you be willing to add another commit to this PR with the note? Here's what the syntax is:

```
:::note
Your content here
:::
```

BrunoBlanes:
How about this?
```
:::note
If you are setting up FAST on a Blazor Server project, you will need to escape the '@' character by repeating it in the source link. For more information check out the [Razor Pages syntax documentation](https://docs.microsoft.com/en-us/aspnet/core/mvc/views/razor#razor-syntax).
:::
```

nicholasrice (approved):

EisenbergEffect:
@BrunoBlanes That's perfect. Go ahead and add that if you don't mind. Ping me and I'll get this merged.
Thanks for contributing!

EisenbergEffect (approved):

PR #4065, opened 2020-10-24
feat: enable shared CSSStyleSheets in DesignSystemProvider

nicholasrice:
# Description
This change introduces a configurable CustomPropertyManager to the DesignSystemProvider which is responsible for:
1. Applying the CSS Stylesheet to the client (DSP) instance
2. Removing the CSS Stylesheet from the client (DPS) instance
3. Writing the individual custom properties to the CSSStyleDeclaration 
4. Registering the custom property dependencies

Refactoring the DSP to use this new CustomPropertyManager has a distinct advantage because it allows *sharing* a manager between DSP instances. This is the mechanism through which DSP instances can share constructible StyleSheets and CSS custom property registries.  

To leverage this for Card scenarios, the card will do something like the following in their constructor or connectedCallback:

```ts
constructor() {
    super();
    
     if (adoptedStyleSheetsSupported) {
        /* would re-use where design system object is identical */
        this.customPropertyManager = getReusableCustomPropertyManager(this.provider, this.backgroundColor);
    
        // Then listen for changes to the inputs of the caching function and update manager on changes 
        const notifier = Observable.getNotifier(this);
        notifier.subscribe(this, "provider");
        notifier.subscribe(this, "backgroundColor");
    }
}

handleChange() {
    this.customPropertyManager = getReusableCustomPropertyManager(this.provider, this.backgroundColor);
}
```

There are two kinds of `CustomPropertyManagers` added - one that uses constructable stylesheets and the other that uses HTMLStyleElements. Note that re-use like the above is *not* supported for the `StyleElementCustomPropertyManager` because the style element that powers it can only be in one place at a time.

This is a large PR but a good portion of it is tests. The meat of the change is in the added file `manager.ts`, and then wiring that into the `design-system-provider.ts`.

It would be good to get a good smoke on this PR as it is a pretty foundational change.

## Issue type checklist

## Process & policy checklist


janechu:
In the description example could we consolidate the subscribe to:

```ts
notifier.subscribe(this, ["provider", "backgroundColor"]);
```

nicholasrice:
@janechu I don't think the Observable API supports that right now.

packages/web-components/fast-foundation/src/custom-properties/manager.ts
```diff
@@ -0,0 +1,282 @@
...
+    private subscribers = new Set();
+
+    constructor(sheet: CSSStyleSheet) {
+        super();
+
+        this.sheet = sheet;
+
+        this.styles = ElementStyles.create([sheet]);
+
+        this.customPropertyTarget = (sheet.rules[
```
janechu, line 203: Should we use `cssRules` since `rules` is considered legacy? https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/rules

chrisdholt (approved):
Thanks for all the tests, super helpful to walk through those.

packages/web-components/fast-foundation/src/design-system-provider/design-system-provider.ts
```diff
@@ -255,42 +284,39 @@ export class DesignSystemProvider extends FASTElement
...
-                if (typeof property === "string") {
-                    this.deleteCustomProperty(property);
-                }
 
-                this.writeCustomProperties();
+                if (manager) {
+                    if (typeof property === "string") {
+                        manager.set({
+                            name: property,
+                            value: "",
```
janechu, line 314: Empty string as a value?

nicholasrice: The intent is to remove it, but I should probably just add a public `delete` method.

packages/web-components/fast-foundation/src/custom-properties/manager.ts
nicholasrice: Good catch, updated.

janechu (approved):

EisenbergEffect (approved):

EisenbergEffect:
Looks pretty good. @nicholasrice Is there some documentation we can write around this?

nicholasrice:
Oops good call - I'll update docs.

PR #4074, opened 2020-10-26
feat: add select component

radium-v:
# Description

This PR adds the `<fast-option>`, `<fast-listbox>`, and `<fast-select>` components.

open select component

## Motivation & context
The `fast-option` is a fast version of the HTML `option` element. The component uses the `option` ARIA role. Clicking a `fast-option` will set its `selected` state to `true`.

The `fast-listbox` is a fast component that uses the `listbox` ARIA role. It can hold `fast-option` elements, and will be able to hold container elements (similar to `optgroup`) in the future. The arrow keys allow users to cycle through its children to change which `fast-option` is selected.

The `fast-select` is a fast version of the HTML `select` element. It contains a button, a slottable `fast-listbox`, and slotted `fast-option`s. The button is displayed by default. Clicking the button will toggle the listbox to display the available options. Clicking an option will close the menu and change the button's display to reflect the selected option.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/listbox/listbox.spec.ts
```diff
@@ -0,0 +1,153 @@
+import { expect } from "chai";
+import { customElement, DOM, html } from "@microsoft/fast-element";
+import { OptionTemplate as itemTemplate, Option } from "../option";
+import { fixture } from "../fixture";
+import { Listbox, ListboxTemplate as template } from "./index";
+
+@customElement({
+    name: "fast-list-box",
```
chrisdholt, line 8: I think we'll want this as a single word, `listbox`.

radium-v: big oof

packages/web-components/fast-foundation/src/select/select.template.ts
```diff
@@ -0,0 +1,55 @@
...
+                            <path
+                                d="M11.85.65c.2.2.2.5 0 .7L6.4 6.84a.55.55 0 01-.78 0L.14 1.35a.5.5 0 11.71-.7L6 5.8 11.15.65c.2-.2.5-.2.7 0z"
+                            />
+                        </svg>
+                    </slot>
+                </span>
+            </slot>
+            ${endTemplate}
+        </button>
+        <fast-listbox
```
nicholasrice, line 39: I'm not super excited by this implicit relationship to the `@microsoft/fast-components` package here, though I'm not 100% sure what to suggest.

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * The value displayed on the button
+     * @public
+     */
+    @observable
+    public displayValue: string;
+
+    /**
+     * The button element that displays the selected value and triggers the visibility of the listbox.
+     * @public
```
nicholasrice, line 72: What is the purpose of this `button` property being expressed in documentation?

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+     */
+    @observable
+    public button: HTMLElement;
+
+    /**
+     * The slot that holds the contents of the button
+     * @public
```
nicholasrice, line 79: Same question here: does this `slottedButtonContainer` need to be in documentation?

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    protected valueChanged(previous: string, next: string): void {
+        super.valueChanged(previous, next);
+        DOM.queueUpdate(() => {
+            const selectedOption = this.listbox.selectedOption;
+            this.displayValue = selectedOption.textContent || selectedOption.value;
+        });
+    }
+
+    private updateForm(): void {
```
nicholasrice, line 101: I'm not sure this `updateForm` method is necessary. FormAssociated takes care of communicating `value` property changes to the parent form.

This *may* have been started prior to that work which is why it was here? I don't think it's needed anymore though...

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+
+        if (keyCode === "Escape") {
+            this.open = false;
+            DOM.queueUpdate(() => this.button.focus());
+            return;
+        }
+
+        if ([" ", "Enter"].includes(`${keyCode}`)) {
+            e.preventDefault();
+            e.stopPropagation();
```
nicholasrice, line 119: Discussed offline, but we should avoid `stopPropagation()` because it prevents other code from receiving the event, which is often un-intuitive. 

My general rule-of-thumb when the urge to use `stopPropagation()` arises is to make event handlers better at determining if they should act.

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    public focusoutHandler = (e: FocusEvent): void => {
+        const focusTarget = e.relatedTarget as HTMLElement;
+
+        if (!this.open) {
+            return;
+        }
+
+        if (this.isSameNode(focusTarget)) {
+            e.stopPropagation();
```
nicholasrice, line 151: same as above re `stopPropagation()`

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+            this.selectedValue.textContent = value;
+        }
+    }
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+
+        this.options = this.listbox.listboxItems;
+
+        this.button.setAttribute("aria-haspopup", "listbox");
```
nicholasrice, line 201: Is there a reason these are added imperatively in the connectedCallback and not declaratively in the template?

packages/web-components/fast-foundation/tsconfig.json
```diff
@@ -14,7 +14,8 @@
         ],
         "lib": [
             "DOM",
-            "ES2015"
+            "ES2015",
+            "ES2016.Array.Include"
```
nicholasrice, line 18: Good call

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+    public autofocus: boolean = false;
+
+    /**
+     * The open attribute
+     */
+    @attr({ attribute: "open", mode: "boolean" })
+    public open: boolean;
+    private openChanged(): void {
+        if (this.$fastController.isConnected) {
+            this.button.setAttribute("aria-expanded", `${this.open}`);
```
nicholasrice, line 30: Any reason you're not binding to this attribute in the template?

packages/web-components/fast-foundation/src/option/option.ts
```diff
@@ -0,0 +1,99 @@
...
+    @observable
+    public selected: boolean = this.defaultSelected;
+    private selectedChanged(oldValue, newValue): void {
+        if (this.$fastController.isConnected) {
+            if (!this.dirtySelected) {
+                this.dirtySelected = true;
+            }
+
+            this.classList.toggle("selected", oldValue !== newValue ? newValue : false);
+
```
nicholasrice, line 64: I wouldn't expect the *option* to emit a change event because the *option* isn't associated to the form. It would also mean that you would get two `change` events emitted whenever selecting one option results in de-selecting another.

packages/web-components/fast-foundation/src/option/option.options.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Option role.
+ * @public
+ */
+export enum OptionRole {
```
nicholasrice, line 5: I don't see this used (though I could be missing it) - can it be removed?

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -0,0 +1,338 @@
...
+            // `label` property could be used here.
+            // Chromium/Firefox's native <select>s ignore whitespace at the
+            // beginning/end of the visible text when matching, so trim()
+            // the search text to align with that behavior.
+            const element: any = option;
+            const matches = element.innerText.trim().match(re);
+
+            if (matches) {
+                // this.setFocusOnOption(element);
+                console.log(element);
```
nicholasrice, line 331: does this area still need work?

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -0,0 +1,338 @@
...
+            captured = this.getFocusedOption();
+        }
+
+        if (captured) {
+            if (captured.selected) {
+                captured.setAttribute("tabindex", "0");
+                return;
+            }
+            captured.removeAttribute("tabindex");
+            // captured.setAttribute("tabindex", captured.selected ? "0" : "-1");
```
nicholasrice, line 150:
```suggestion
```

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -0,0 +1,338 @@
...
+                n.getAttribute("role") === "option" &&
+                !n.disabled
+        );
+
+        this.listboxItemsCount = this.listboxItems.length;
+        this.selectedOptionIndex = this.listboxItems.findIndex(el => el.selected) || 0;
+
+        DOM.queueUpdate(() => this.setupOptions());
+
+        // this.addEventListener("keypress", this.keypressHandler);
```
nicholasrice, line 122:
```suggestion
```

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -0,0 +1,338 @@
...
+     * HTML Attribute: disabled
+     */
+    @attr({ attribute: "disabled", mode: "boolean" })
+    public disabled: boolean;
+    private disabledChanged(): void {
+        if (!this.$fastController.isConnected) {
+            return;
+        }
+
+        if (this.disabled) {
```
nicholasrice, line 25: aria-disabled is bound in the template, I'm not sure you need to have this code.

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -0,0 +1,338 @@
...
+            this.setAttribute("aria-disabled", "true");
+        }
+
+        if (this.listboxItems.length) {
+            this.listboxItems.forEach((option: Option) => {
+                if (this.disabled) {
```
nicholasrice, line 31: nit: can be reduced to `option.disabled = this.disabled`

packages/web-components/fast-foundation/src/listbox/listbox.ts
radium-v: This entire `handleTypeAhead` function is left mostly as-is from the original draft implementation. It needs to be fixed and cleaned up.

packages/web-components/fast-foundation/src/select/select.ts
radium-v: I'm not 100% familiar with our documentation's specifics so I decided to err on the side of document all the things

packages/web-components/fast-foundation/src/select/select.ts
radium-v: The convention in the early draft added these imperatively, stating that they'd need to be added if the default slotted content was replaced. I wasn't sure so I left them as is, but I prefer to have them in the template instead.

packages/web-components/fast-foundation/src/select/select.ts
nicholasrice: To me, the trigger button is an implementation detail of Select. At this point I can't think of a reason to expose the the button as part of the public API so unless there is an explicit reason to expose it I'd recommend marking it `@internal`

packages/web-components/fast-foundation/src/select/select.ts
nicholasrice: It doesn't look like it's in a slot anymore so I think it is probably safe to move to the template.

khamudom (changes requested):

packages/web-components/fast-foundation/src/select/select.template.ts
scomea: There is also a similar problem in data grid and presumably any component that creates other components at runtime.

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            return;
+        }
+
+        // Pass the event to the listbox
+        if (!e.defaultPrevented) {
+            DOM.queueUpdate(() => {
+                e.preventDefault();
+                this.listbox.dispatchEvent(e);
```
scomea, line 138: I'm concerned re-dispatching a keyboard event to a child could lead to weirdness.  Does it get bubble to select parents before listbox handles it, what if a parent calls preventDefault on it?   Why not just set a "selected item" prop on the listbox or call a setter function directly?

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,224 @@
...
+        });
+    }
+
+    private updateForm(): void {
+        const value = this.value ? this.value : null;
+        this.setFormValue(value, value);
+    }
+
+    public keydownHandler = (e: KeyboardEvent): void => {
+        super.keypressHandler(e);
```
scomea, line 107: Invoking keypress on keydown seems odd

packages/web-components/fast-foundation/src/select/select.template.ts
scomea: I'm thinking that when we declare a web component we should be able to set some properties as well, like what tags to use for generated options or grid rows.

packages/web-components/fast-foundation/src/listbox/listbox.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+import { fixture } from "../fixture";
+import { Listbox, ListboxTemplate as template } from "./index";
+
+@customElement({
+    name: "fast-listbox",
+    template,
+})
+class FASTListbox extends Listbox {}
+
+// TODO: Need to add tests for keyboard handling & focus management
```
nicholasrice, line 13: Can we get an issue tracking this TODO?

packages/web-components/fast-components/docs/api-report.md
```diff
@@ -20,10 +20,13 @@ import { Dialog } from '@microsoft/fast-foundation';
 import { Direction } from '@microsoft/fast-web-utilities';
 import { Divider } from '@microsoft/fast-foundation';
 import { Flipper } from '@microsoft/fast-foundation';
+import { Listbox } from '@microsoft/fast-foundation';
 import { Menu } from '@microsoft/fast-foundation';
 import { MenuItem } from '@microsoft/fast-foundation';
+import { Option as Option_2 } from '@microsoft/fast-foundation';
```
awentzel, line 26: `Option-2` seems out of place here.  How is this used?  Two without one or three?  Maybe "SubOption" or "MenuOption" would be more intuitive and align with how you've named other variables.

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -0,0 +1,329 @@
...
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: disabled
+     */
+    @attr({ mode: "boolean" })
+    public disabled: boolean;
+
+    /**
+     * The index of the selected option
```
nicholasrice, line 24:
```suggestion
     * The index of the selected option. -1 when no option is selected.
```

packages/web-components/fast-components/src/select/fixtures/base.html
```diff
@@ -0,0 +1,104 @@
+<h1>Select</h1>
+<select>
+    <option value="" disabled selected>Choose an option</option>
+    <option>Option Two</option>
+    <option>Option Three</option>
+    <option>Option Four</option>
+</select>
+<h2>Default</h2>
```
awentzel, line 8: nit: in other places you have a line break before <h2>

packages/web-components/fast-foundation/docs/api-report.md
```diff
@@ -634,6 +718,54 @@ export const RadioGroupTemplate: import("@microsoft/fast-element").ViewTemplate<
...
+//
+// @public
+export class Select extends Listbox {
+    // @internal (undocumented)
+    clickHandler(e: MouseEvent): boolean | void;
+    // (undocumented)
+    connectedCallback(): void;
+    get displayValue(): string;
+    // (undocumented)
+    focusoutHandler(e: FocusEvent): boolean | void;
```
awentzel, line 732: should it be "focusOutHandler" and "keyDownHandler"? what's the pattern you're following?

packages/web-components/fast-foundation/src/option/option.options.ts
awentzel: Do options have roles, if so, maybe it's just missing the example/unit test?

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,225 @@
...
+class FormAssociatedSelect extends FormAssociated<HTMLSelectElement> {
+    protected proxy: HTMLSelectElement = document.createElement("select");
+
+    protected updateForm(previous, next): void {
+        const value = this.value;
+        if (this.proxy instanceof HTMLElement) {
+            this.proxy.value = value;
+        }
+    }
+}
```
nicholasrice, line 192: I'm a little specious that using a mixin approach with FormAssociated won't work here because the mixin code copies prototype methods and in the process will likely override relevant methods on the FormAssociated (constructor, connectedCallback, disconnectedCallback, etc).

Can you confirm that those methods are called when the Select attaches?

packages/web-components/fast-components/docs/api-report.md
radium-v: yeah this seems wrong.

packages/web-components/fast-foundation/src/select/select.ts
nicholasrice: There is another mixin model I just learned about that could resolve that issue potentially if the FormAssociated lifecycle hooks aren't called.

awentzel (approved):
A few suggestions made, but otherwise, looks very clean and intuitive.

janechu:
Scenarios seem to be missing from this component, it won't show up on the component explorer without them.

packages/web-components/fast-components/docs/api-report.md
radium-v: I think this is happening because DOM already has an `Option()` constructor. Looking at the internals, this would likely happen with `Audio/<audio>` and `Image/` as well.

packages/web-components/fast-foundation/src/option/option.options.ts
chrisdholt: I'd actually prefer we stay away from an enum here personally for option. I see this being used for checking the value in certain places, but enums can be costly as they grow over time and an option *only ever* should have a role of option. If we can remove this I think it would be better.

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -0,0 +1,225 @@
...
+    public keydownHandler(e: KeyboardEvent): boolean | void {
+        super.keydownHandler(e);
+
+        const keyCode: string = e.key || `${e.key.charCodeAt(0)}`;
+
+        if (keyCode === " " || keyCode === "Enter") {
+            this.open = !this.open;
+        }
+
+        if (keyCode === "Escape") {
```
janechu, line 155: Just an additional note, these should be in `@microsoft/fast-web-utilities`.

packages/web-components/fast-foundation/docs/api-report.md
radium-v: I'm following the string values for the events, which are "focusout" and "keydown".

packages/web-components/fast-components/src/select/fixtures/base.html
```diff
@@ -0,0 +1,104 @@
+<h1>Select</h1>
+<select>
+    <option value="" disabled selected>Choose an option</option>
+    <option>Option Two</option>
+    <option>Option Three</option>
+    <option>Option Four</option>
+</select>
```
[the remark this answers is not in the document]
radium-v: I was using it for comparison while building ours. I'll make sure to remove it.

EisenbergEffect:
@radium-v Please ping me when you're finished addressing the feedback above and I'll help elevate visibility for final review.

eljefe223:
Smoked, functionality seems right, small visual elements need to be fixed. 

When menu is open the focus visual corner radius is squared off
image

eljefe223:
Do we want to reserve space for icons on the options when there is no start content on any of the options? Visually seems broken to me.

chrisdholt:
> Do we want to reserve space for icons on the options when there is no start content on any of the options? Visually seems broken to me.
> 
> image

I'd actually prefer we don't, though I'm fine if we do it as a fast follow...

eljefe223:
> > Do we want to reserve space for icons on the options when there is no start content on any of the options? Visually seems broken to me.
> > image
> 
> I'd actually prefer we don't, though I'm fine if we do it as a fast follow...

I'm fine with that too.

packages/web-components/fast-components/src/listbox-option/fixtures/base.html
```diff
@@ -0,0 +1,18 @@
...
+    Text content is the value when the value attribute is absent.
+</fast-option>
+
+<fast-option value="Value content loses">
+    When the value attribute and text are both present, this text should be displayed.
+</fast-option>
+
+<h2>Disabled</h2>
+<fast-option disabled>
+    When the value attribute and text are both present, this text should be displayed.
```
chrisdholt, line 14: nit and doesn't *really* matter, but this is just copy/paste :).

chrisdholt (approved):
Pulled this down and it looks good! One thing that is missing, though I'm okay not holding on it, would be start/end slots for the option. I know that the menu item component doesn't have that, though I think that's a bit of a miss that should be addressed and divergence here would be welcome. Menu items have additional requirements for nested menus and icons to denote that; it adds complexity to the slot model that needs to be addressed. Options however, don't have that same requirement for nesting and visual indicators, so the story for start/end is much simpler there.

I don't really want to hold much longer on this - if it's quick, let's add it visually. If need be, we can add it to the template and styles as a follow up.

marjonlynch:
There are a couple issues around scenarios for fast-component-explorer and keyboarding that John is fixing now I think, once those are fixed, I've smoked other things in firefox and edge and looks good to me

khamudom (approved):

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -0,0 +1,144 @@
...
+    :host(:${focusVisible}) {
+        box-shadow: 0 0 0 calc(var(--focus-outline-width) * 1px) inset ${
+            neutralFocusInnerAccentBehavior.var
+        };
+        border-color: ${neutralFocusBehavior.var};
+        background: ${accentFillHoverBehavior.var};
+        color: ${accentForegroundCutRestBehavior.var};
+    }
+
+    :host(.selected) {
```
nicholasrice, line 54: There was a previous PR that moved many of these selectors away from using a class on the host, to using the attr selector. That is because classes on the host can get clobbered in some frameworks. Can this use an attribute selector?

packages/web-components/fast-components/src/select/select.styles.ts
```diff
@@ -0,0 +1,234 @@
...
+        border-color: ${accentFillActiveBehavior.var};
+    }
+
+    :host(.open.above) .listbox,
+    :host(.open.below) .control {
+        border-bottom-left-radius: 0;
+        border-bottom-right-radius: 0;
+    }
+
+    :host(.open.above) .control,
```
nicholasrice, line 127: same question here regarding attribute selectors.

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
```diff
@@ -0,0 +1,172 @@
+import { attr, FASTElement, observable } from "@microsoft/fast-element";
+import { StartEnd } from "../patterns/start-end";
+import { applyMixins } from "../utilities/apply-mixins";
+
+/**
+ * An Option Custom HTML Element.
+ * Implements {@link https://www.w3.org/TR/wai-aria-1.1/#option | ARIA menuitem }.
+ *
+ * @public
+ */
```
nicholasrice, line 10: Does this need proxy logic? I don't think it is form-associated is it?

marjonlynch (approved):
Smoked in explorer and storybook in Edge and Firefox and all expected changes looked good, keyboard navigation, scenario is there now (inserted a native html element in an option with great success)

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
radium-v: For `option`s, he `selected` attribute represents the `defaultSelected` property. The `selected`-ness of the option is only represented by its `selected` property, which is only expressed in the DOM via the `selected` class and the `aria-selected` attribute.

It looks like `checkbox` is using `[aria-checked="true"]` for styling, so I'll go that route and use `[aria-selected="true"]`.

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
radium-v: The listbox-option isn't form-associated but it does hold its own `option` element to make option manipulation with the form-associated select significantly easier.

packages/web-components/fast-components/src/select/select.styles.ts
radium-v: Updated to use attributes for `[open]`, and restructured to use `[position="above"]` and `[position="below"]`.

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
nicholasrice: gotcha okay

nicholasrice (approved):

PR #4078, opened 2020-10-27
feat: attribute selectors instead of class selectors

### LINKED ISSUE #4076: Update all FAST components to use attributes for styling rather than classes
This issue is meant to track an improvement to leverage attributes for styling rather than classes. Hoisting classes out of web components can have unintended side effects with author applied classes and can be clobbered when integrated with certain frameworks.

This task should be considered "complete" when all components have been updated to leverage attributes instead of classes.

vertonghenb:
# Description
Uses attribute selectors instead of the class selectors since it's not always easy to integrate.
To make the library more open for integration (e.g. React and Blazor) since we don't have to bother with the css classes.

## Motivation & context

- Fixes #4076

## Issue type checklist

css classes can no longer be used.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/checkbox/checkbox.styles.ts
+++ b/packages/web-components/fast-components/src/checkbox/checkbox.styles.ts
@@ -103,22 +103,22 @@ export const CheckboxStyles = css`
         border-color: ${neutralFocusBehavior.var};
     }
 
-    :host([checked]) .control {
+    :host([aria-checked="true"]) .control {
         background: ${accentFillRestBehavior.var};
         border: calc(var(--outline-width) * 1px) solid ${accentFillRestBehavior.var};
     }
 
-    :host([checked]:not([disabled])) .control:hover {
+    :host([aria-checked="true"]:not([disabled])) .control:hover {
         background: ${accentFillHoverBehavior.var};
         border: calc(var(--outline-width) * 1px) solid ${accentFillHoverBehavior.var};
     }
 
-    :host([checked]:not([disabled])) .control:active {
+    :host([aria-checked="true"]:not([disabled])) .control:active {
         background: ${accentFillActiveBehavior.var};
         border: calc(var(--outline-width) * 1px) solid ${accentFillActiveBehavior.var};
     }
 
-    :host([checked]:${focusVisible}:not([disabled])) .control {
+    :host([aria-checked="true"]:${focusVisible}:not([disabled])) .control {
         box-shadow: 0 0 0 2px var(--background-color), 0 0 0 4px ${
             neutralFocusBehavior.var
         };
@@ -133,7 +133,7 @@ export const CheckboxStyles = css`
         cursor: ${disabledCursor};
     }
 
-    :host([checked]:not(.indeterminate)) .checked-indicator,
+    :host([aria-checked="true"]:not(.indeterminate)) .checked-indicator,
     :host(.indeterminate) .indeterminate-indicator {
         opacity: 1;
     }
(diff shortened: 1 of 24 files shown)


EisenbergEffect:
WOW!!!
@chrisdholt Check this out!

packages/web-components/fast-components/src/checkbox/checkbox.styles.ts
```diff
@@ -103,42 +103,42 @@ export const CheckboxStyles = css`
...
-    :host(.disabled) .control {
+    :host([disabled]) .label,
+    :host([readonly]) .label,
+    :host([readonly]) .control,
+    :host([disabled]) .control {
         cursor: ${disabledCursor};
     }
 
-    :host(.checked:not(.indeterminate)) .checked-indicator,
+    :host([checked]:not(.indeterminate)) .checked-indicator,
```
chrisdholt, line 136: should `indeterminate` be an attribute selector?

packages/web-components/fast-components/src/switch/switch.styles.ts
```diff
@@ -41,14 +41,14 @@ export const SwitchStyles = css`
...
 
-    :host(.disabled) {
+    :host([disabled]) {
         opacity: var(--disabled-opacity);
     }
 
-    :host(.disabled) .label,
+    :host([disabled]) .label,
     :host(.readonly) .label,
     :host(.readonly) .switch,
```
chrisdholt, line 50: `readonly` as an attribute selector?

chrisdholt (commented):
Thanks @vertonghenb! I think we'll need to remove the classes from the `fast-foundation` package which isn't reflected here. Thanks for the quick update!

vertonghenb:
Still have to test some stuff out for example the checkbox is a bit buggy could use some help on that one though.

vertonghenb:
Note to self:
Bugged: 
- Checkbox (aria-checked?)
- Switch (aria-checked?)
- Flipper (wrong selector?)

All other components seem to be working correctly.

chrisdholt:
> Note to self:
> Bugged:
> 
> * `Checkbox` (aria-checked?)
> * `Switch` (aria-checked?)
> * `Flipper` (wrong selector?)
> 
> Investigate `indeterminate`
> 
> All other components seem to be working correctly.

I can pull this down and check Flipper if you like. I do think you'll want to use the aria-checked values for styling checkbox and flipper rather than `checked`. Does that help?

vertonghenb:
> > Note to self:
> 
> > Bugged:
> 
> > 
> 
> > * `Checkbox` (aria-checked?)
> 
> > * `Switch` (aria-checked?)
> 
> > * `Flipper` (wrong selector?)
> 
> > 
> 
> > Investigate `indeterminate`
> 
> > 
> 
> > All other components seem to be working correctly.
> 
> 
> 
> I can pull this down and check Flipper if you like. I do think you'll want to use the aria-checked values for styling checkbox and flipper rather than `checked`. Does that help?

@chrisdholt thanks for the help and example already, if you want to track down the flipper that'd be nice! I'll do the last tests tomorrow and check the switch and checkbox with the aria-label, thanks for the tip.

packages/web-components/fast-components/src/checkbox/checkbox.styles.ts
nicholasrice: indeterminate is only a property at this point

chrisdholt: Tada! Thanks @nicholasrice

vertonghenb: Pardon my ignorance but what does this mean exactly? Should be do something about this in this PR or have a follow-up? I'm thinking how to integrate this with for example Blazor and this could cause some trouble for other integrations. Should we use an attribute for this?

packages/web-components/fast-components/src/switch/switch.styles.ts
vertonghenb: fixed.

vertonghenb:
@chrisdholt
@EisenbergEffect 
 I think we're finished here apart from the `Flipper`. Might want to pay attention to the `Tabs` commit, comments are in the commit.
As a follow-up: Remove classes in the Foundation package.

packages/web-components/fast-components/src/flipper/flipper.styles.ts
```diff
@@ -48,8 +48,8 @@ export const FlipperStyles = css`
         transition: all 0.1s ease-in-out;
     }
 
-    .next,
-    .previous {
+    [direction="next"],
+    [direction="previous"] {
```
chrisdholt, line 52: @vertonghenb These classes are fine here (`next`, `previous`). They are within the shadow DOM  rather than on the host so they should be fine. Adding these back in will solve the issue.

vertonghenb: makes sense, the same as the ones for `tabs` thanks!

chrisdholt: Yeah, and thanks for the note on vertical/horizontal in tabs. I think we can follow up on that one if you're seeing issues there as well.

vertonghenb:
Done.

chrisdholt (approved):
We can take the removal of the classes in a separate issue. This resolves the issue of using attribute selectors where appropriate.

EisenbergEffect (approved):
Thanks for this fantastic contribution!

packages/web-components/fast-components/src/checkbox/checkbox.styles.ts
nicholasrice: Sorry, I should be more clear :)

`indeterminate` is a JavaScript property (AKA an "IDL attribute") only and cannot be represented in HTML as an HTML attribute (AKA a "content attribute"), so there is no attribute selector for you to use to style the indeterminate state - you must (at this point) use the class to style this state. The reason for this is historical; the `Checkbox` web-component behaviorally mirrors the `<input type="checkbox />` component, which does not expose an HTML attribute that sets the `indeterminate` property.

Whether we *should* expose `indeterminate` as an attribute could certainly be up for debate, but I think that would need enough investigation that I wouldn't want to resolve that for this PR. In an ideal world, custom element states would allow us to match the `:indeterminate` pseduo-class, but we don't quite have that capability today.

I don't think any action is necessary here, but this control will still exhibit the behavior that this PR aims to fix. I'll open up an issue to track discussion about what we should do about that.

nicholasrice (approved):

packages/web-components/fast-components/src/checkbox/checkbox.styles.ts
nicholasrice: #4084

chrisdholt:
Thanks @vertonghenb!

PR #4080, opened 2020-10-27
chore: add the converted MDN data as an export

janechu:
# Description

This change adds a build script to convert the MDN CSS data to a file that exports it from `@microsoft/fast-tooling`. It does this by creating a new `tsconfig` and exporting `@microsoft/fast-tooling` as `commonjs` to a temp directory which can then be read by the node script to write the file.

## Motivation & context

Foundational work for #3773

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tooling/src/css-data.ts
```diff
@@ -0,0 +1,18730 @@
+/**
+ * This file is generated from build/generate-mdn-data-files.js
+ * any modifications will be overwritten.
+ */
```
nicholasrice, line 4: Do we want this built file checked in?

scomea (approved):

packages/tooling/fast-tooling/src/css-data.ts
janechu: We do, however I've changed this around to be a js file so that I can import it and provide a check to see whether it should be updated. This provides a level of stability should MDN change their formatting which would break our current utilities expectations.

nicholasrice (approved):

PR #4081, opened 2020-10-27
fix: update start and end slot to remove fixed width and height

### LINKED ISSUE #3904: mis-aligned slots in fast-menu-item (and fast-button)

When using the `fast-menu-item` component and placing content in the `start` slot it fails to be nicely align (vertically) with the menu item text.

**EDIT:** This mis-alignment in fact also happen in `<fast-button>` and probably comes from something not totally figured out, see my comment below.

1. Open the [fast explorer](https://explore.fast.design/components/fast-menu) in the `fast-menu` section
2. Add a `fast-button` in the start slot
3. The result should look like this:

Capture d’écran 2020-09-15 à 22 19 47

A nice vertical alignement between each part of the `fast-menu-item` template.

khamudom:
# Description

On buttons and menu-item, the start and end slots have fixed width and height. This will cause elements larger than 16px to clip.
This PR removes the fixed dimension on the start and end slot but keeps the 16px on the SVG.
I also updated the `grid-template-columns` property on the menu-item that will also cause clipping with elements that are larger than 42px.
This should not break any current implementation when using a `button`, `anchor`, or the `menu-item`. This will now allow authors to add larger elements to the start and end slots if the design calls for it.

fixes #3904 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
+++ b/packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
@@ -22,7 +22,7 @@ export const MenuItemStyles = css`
         outline: none;
         box-sizing: border-box;
         height: calc(${heightNumber} * 1px);
-        grid-template-columns: 42px auto 42px;
+        grid-template-columns: minmax(42px, auto) 1fr minmax(42px, auto);
         grid-template-rows: auto;
         justify-items: center;
         align-items: center;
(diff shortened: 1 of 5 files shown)


packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
```diff
@@ -83,7 +83,10 @@ export const MenuItemStyles = css`
     }
 
     .start,
-    .end,
+    .end {
+        display: flex;
+    }
```
chrisdholt, line 88: Are these two the only places we've baked in the start/end to 16px?

khamudom: After another check, I missed text-field, tree-item, and breadcrumb

nicholasrice (approved):

bheston (approved):

PR #4088, opened 2020-11-02
fix: assign all react components an originatorId to determine re-rendering of form elements

### LINKED ISSUE #3628: Form repositions cursor to the end of textarea during typing

When the `<Form />` from `@microsoft/fast-tooling-react` encounters piece of text, during typing the cursor will move to the end of the textarea.

The cursor should remain where it is so that a user can type in the middle of sentences/paragraphs.

janechu:
# Description

This change adds an `originatorId` to the message system for the exports of `@microsoft/fast-tooling-react` and the exported services from `@microsoft/fast-tooling`. This helps identify where messages are being sent from so that, as an additional change in this PR, if a message comes from the `Form`, it will be ignored as a value on the rendered `textarea`, thus bypassing an issue with React on controlled inputs where the cursor will reposition.

This does not completely solve the issue as a check for focus must also be made or the cursor will reposition but solves it repositioning after the initial keypress.

## Motivation & context

Begins to address #3628 

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):
Makes sense. I've had to do things like this in message systems before as well. Rock on 🥁

nicholasrice (approved):

PR #4103, opened 2020-11-04
chore: add operations scripts for maintaining infrastructure

awentzel:
# Description
This merges CLI work for managing the Azure infrastructure.  These scripts perform operations for creating new services, destroying old services, and updating existing services.  This is beta and mostly usable with some features yet to be implemented which are documented in comments and with GH issues.

This begins a FAST CLI extending the Azure CLI simplified maintainability with consistency, speed, and reliability.

## Motivation & context
To provide an equal ability for all collaborators to perform common service needs.  For example, perform release deployments of websites.

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):
I can't comment much on the scripts themselves. I haven't written bash scripts anywhere that complexity ;) Glad to see the documentation and get this in. Just want to also confirm that there's nothing in here that could be a security concern in terms of content, config, etc.

nicholasrice:
@awentzel please add the appropriate feat, fix, or chore flag to the title of this PR.

build/operations/README.md
```diff
@@ -0,0 +1,171 @@
...
+**Resolution**
+-We managed to configure custom domain on both the WebApp and FrontDoor, so they both listen on same hostname. Such Hostname or FQDN, would CNAME to FrontDoor's domain.
+-This custom domain was now used on the URL string within the AAD redirect so would work as expected. Nevertheless, since such AAD module responds with 302 to every unauthenticated user-agent, the FrontDoor probes would mark backend as unhealthy:
+
+https://docs.microsoft.com/en-us/azure/frontdoor/front-door-health-probes
+
+-When having multiple backends, it would round-robin requests, making it unusable.
+-Usually a dummy path is setup in application for the 200 OK probes, but verified with AppServices it cannot be made with the pre-built AAD authentication module. To customize AAD, it's then suggested to implement it a code.
+
+-This scenario doesn't happen with AppGW, since it can just re-write the URL parameter and also allow non-200 OK health responses.
```
janechu, line 128: Looks like formatting was a little off here.

```suggestion
- Issue happens because of URL string within one of the parameters of the HTTP redirect pointing to AAD, which then uses such parameter for redirecting the user accordingly.
- Such URL was the one configured on the WebApp.

**Resolution**
- We managed to configure custom domain on both the WebApp and FrontDoor, so they both listen on same hostname. Such Hostname or FQDN, would CNAME to FrontDoor's domain.
- This custom domain was now used on the URL string within the AAD redirect so would work as expected. Nevertheless, since such AAD module responds with 302 to every unauthenticated user-agent, the FrontDoor probes would mark backend as unhealthy:

https://docs.microsoft.com/en-us/azure/frontdoor/front-door-health-probes

- When having multiple backends, it would round-robin requests, making it unusable.
- Usually a dummy path is setup in application for the 200 OK probes, but verified with AppServices it cannot be made with the pre-built AAD authentication module. To customize AAD, it's then suggested to implement it a code.

- This scenario doesn't happen with AppGW, since it can just re-write the URL parameter and also allow non-200 OK health responses.
```

build/operations/services/create-appi.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
janechu, line 1: should we rename the file to "create-app-insights" for more clarity when browsing the file structure?

build/operations/services/create-asp.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
janechu, line 1: same here, should we avoid the abbreviation and name this file "create-app-service-plan"?

janechu (approved):

build/operations/services/create-appi.sh
awentzel: The names would get crazy long for different services.  So each file uses the recommended short abbr, except for a couple services that don't have one.  This is one reason I clearly add the name and description of the service in the top of each file.  Additionally, these shouldn't be called directly only through a recipe.

build/operations/services/create-asp.sh
awentzel: Normally I would agree not to use abbreviations. In the context of Azure these are well known https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/naming-and-tagging and in addition to my other comment felt like the right choice for now.

marjonlynch (approved):
I'm assuming this is a multi-stage approach, I was able to get the scripts running and after selecting westus got a permissions error which was expected as I don't apparently have permissions for that.

Issue #2279, opened 2019-09-30
Allow all english strings to be provided as translations to the form


janechu:
The form currently has embedded strings that are english only. These should be configurable so that other english strings can be substituted or translations provided.

A dictionary of strings should be provided as a prop, such that:

```tsx
<Form
    strings={{
       addArrayItem: "选择以添加项目" // replacement for "Select to add item"
    }}
/>
```

Most strings are found in the `@microsoft/fast-tooling-react` package in the `src/form/controls/*.tsx` files.

PR #4104, opened 2020-11-06
feat: adding localization for hard coded strings

williamw2:
# Description

Removing hard coded strings from the Form components and placing them into a dictionary that can be overridden as an attribute on the Form component.

## Motivation & context

Allows for dynamic localization of hard coded strings.

https://github.com/microsoft/fast/issues/2279

## Issue type checklist

## Process & policy checklist

- [x ] I have added tests for my changes.
- [ x] I have tested my changes.
- [x ] I have updated the project documentation to reflect my changes.
- [ x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.


packages/tooling/fast-tooling-react/src/form/templates/const-value.props.ts
```diff
@@ -1,3 +1,5 @@
+import { FormStrings } from "@microsoft/fast-tooling-react/src/form/form.props";
```
janechu, line 1: This import should reference the relative filepath, eg. `"../form.props";`

packages/tooling/fast-tooling-react/src/form/form.tsx
```diff
@@ -99,6 +102,10 @@ class Form extends React.Component<
             props.messageSystem.add(this.messageSystemConfig);
         }
 
+        if (props.strings !== undefined && props.strings !== null) {
```
janechu, line 105: in javascript you can force it to be a bool and then just do a falsey check:
```suggestion
        if (!!!props.strings) {
```

packages/tooling/fast-tooling-react/src/form/controls/utilities/control-switch.props.ts
```diff
@@ -1,3 +1,4 @@
+import { FormStrings } from "@microsoft/fast-tooling-react/src/form/form.props";
```
janechu, line 1: Can this be a relative path (another one I spotted was below).

packages/tooling/fast-tooling-react/src/form/controls/control.section.tsx
```diff
@@ -28,6 +28,7 @@ import {
     SchemaSetValidationMessageRequest,
     TreeNavigationItem,
 } from "@microsoft/fast-tooling";
+import { FormStrings } from "@microsoft/fast-tooling-react/src/form/form.props";
```
janechu, line 31: Same issue with the import pathing here, should be a relative path.

packages/tooling/fast-tooling-react/src/form/controls/control.section-link.tsx
```diff
@@ -44,7 +44,7 @@ class SectionLinkControl extends React.Component<
                 )}
                 onClick={this.handleUpdateSection}
             >
-                Edit: {this.props.label}
+                {this.props.strings.sectionLinkEditLabel}: {this.props.label}
```
janechu, line 47: This is an interesting problem with the `:`, I think we can leave it as-is and wait for feedback but I'm wondering how we can deal with a separator in other languages.

packages/tooling/fast-tooling-react/src/form/controls/control.section.props.ts
```diff
@@ -1,4 +1,5 @@
 import { CombiningKeyword } from "@microsoft/fast-tooling";
+import { FormStrings } from "@microsoft/fast-tooling-react/src/form/form.props";
```
janechu, line 2: Same here, relative pathing would be good, we ignore the `src` directory when we publish so this import would break.

packages/tooling/fast-tooling-react/src/form/templates/const-value.props.ts
williamw2: I have replaced all instances with a relative path.

packages/tooling/fast-tooling-react/src/form/form.tsx
williamw2: Done

packages/tooling/fast-tooling-react/src/form/controls/utilities/control-switch.props.ts
williamw2: Done

packages/tooling/fast-tooling-react/src/form/controls/control.section.tsx
williamw2: Done

packages/tooling/fast-tooling-react/src/form/controls/control.section.props.ts
williamw2: Done

packages/tooling/fast-tooling-react/src/form/controls/control.section-link.tsx
nicholasrice: This would generally be handled by a string formatter, where the code would look like: `format(this.props.strings.sectionLinkEditLabel, this.props.label)` and the injected string might look like `"Edit: {0}"`, where `{0}` is replaced by the second argument to the formatter.

nicholasrice: We currently have a `format` function in `@microsoft/fast-web-utilitities` that works for this purpose.

williamw2: Done

packages/tooling/fast-tooling-react/src/form/custom-controls/control.text-align.tsx
```diff
@@ -5,6 +5,12 @@ import styles, { TextAlignControlClassNameContract } from "./control.text-align.
 import { TextAlignControlProps } from "./control.text-align.props";
 import { classNames } from "@microsoft/fast-web-utilities";
 
+enum Direction {
+    Left = "left",
```
nicholasrice, line 9: should these be localized? Does this get rendered as text?

packages/tooling/fast-tooling-react/src/form/templates/template.control.utilities.tsx
```diff
@@ -89,6 +90,7 @@ abstract class ControlTemplateUtilities<P, S> extends React.Component<
                     className={className}
                     disabled={this.props.disabled}
                     onChange={this.handleSetDefaultValue}
```
nicholasrice, line 92: More of an FYI (unless you feel like integrating it): [React Context](https://reactjs.org/docs/context.html) is a really good fit for this problem. It would allow us to not pass strings directly from component to component, and instead just broadcast them from the root component, where any interested component could opt-in without wiring props to it.

nicholasrice (approved):

packages/tooling/fast-tooling-react/src/form/custom-controls/control.text-align.tsx
janechu: doesn't look like it does, he has `this.props.strings.textAlignLeftLabel` for the translation.

williamw2: The enum values are not displayed and are used for comparison against values passed as a prop to the Text Align control. I added the enum to improve logic within the control but left it as a string comparison to the prop in order to keep the change internal to reduce risk of regression.

packages/tooling/fast-tooling-react/src/form/templates/template.control.utilities.tsx
williamw2: Good to know. Maybe something to think about as part of a larger refactor. I think the property dependency chain could be simplified quite a bit.

janechu (approved):

PR #4108, opened 2020-11-09
fix: revert radio to using class selector instead of attr, update slider-label to use internal styles for orientation

marjonlynch:
# Description
This change fixes visual bugs in **radio**, **radio-group** and **slider-label**
A recent update changed all style selectors to be attribute based, this was problematic for radio as the checked attribute is on the internal proxy input element and checked styling was broken. Slider-label also suffered from the style attribute selector change, that has been updated to use internal but protected (so fluent-ui components can utilize them) styles, **horizontalStyles** and **verticalStyles**.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/radio/radio.styles.ts
+++ b/packages/web-components/fast-components/src/radio/radio.styles.ts
@@ -103,22 +103,22 @@ export const RadioStyles = css`
         border-color: ${neutralFocusBehavior.var};
     }
 
-    :host(.checked) .control {
+    :host([aria-checked="true"]) .control {
         background: ${accentFillRestBehavior.var};
         border: calc(var(--outline-width) * 1px) solid ${accentFillRestBehavior.var};
     }
 
-    :host(.checked:not([disabled])) .control:hover {
+    :host([aria-checked="true"]:not([disabled])) .control:hover {
         background: ${accentFillHoverBehavior.var};
         border: calc(var(--outline-width) * 1px) solid ${accentFillHoverBehavior.var};
     }
 
-    :host(.checked:not([disabled])) .control:active {
+    :host([aria-checked="true"]:not([disabled])) .control:active {
         background: ${accentFillActiveBehavior.var};
         border: calc(var(--outline-width) * 1px) solid ${accentFillActiveBehavior.var};
     }
 
-    :host(.checked:${focusVisible}:not([disabled])) .control {
+    :host([aria-checked="true"]:${focusVisible}:not([disabled])) .control {
         box-shadow: 0 0 0 2px var(--background-color), 0 0 0 4px ${
             neutralFocusBehavior.var
         };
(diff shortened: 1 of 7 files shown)


packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -132,7 +132,7 @@ export const RadioStyles = css`
         cursor: ${disabledCursor};
     }
 
-    :host([checked]) .checked-indicator {
+    :host(.checked) .checked-indicator {
```
chrisdholt, line 135: could these use [aria-checked="true"] - just curious?

packages/web-components/fast-foundation/src/slider-label/slider-label.ts
```diff
@@ -17,12 +18,52 @@ const defaultConfig: SliderConfiguration = {
...
+        }
+        .mark {
+            transform: rotate(90deg);
+            align-self: center;
+        }
+        .label {
+            margin-left: calc((var(--design-unit) / 2) * 2px);
+            align-self: center;
+        }
+    `;
```
chrisdholt, line 66: These should all be added in the `fast-components` instance and not `fast-foundation`.

packages/web-components/fast-components/src/radio/radio.styles.ts
marjonlynch: yep prolly, I will try that

chrisdholt (changes requested):

packages/web-components/fast-foundation/src/slider-label/slider-label.ts
marjonlynch: moving it, thanks!

packages/web-components/fast-foundation/src/radio/radio.ts
```diff
@@ -82,14 +82,14 @@ export class Radio extends FormAssociated<HTMLInputElement> implements RadioCont
      * @public
      */
     @observable
-    public defaultChecked: boolean;
+    public defaultChecked: boolean | undefined = undefined;
```
nicholasrice, line 85: You don't really need an assignment here. Declaring the property but not assigning it implicitly sets the value to `undefined`

nicholasrice (approved):

packages/web-components/fast-foundation/src/radio/radio.ts
marjonlynch: I was just trying to get rid of the red in vscode

packages/web-components/fast-foundation/src/slider-label/slider-label.ts
```diff
@@ -17,6 +17,9 @@ const defaultConfig: SliderConfiguration = {
     disabled: false,
 };
 
+const heightNumber =
```
chrisdholt, line 20: Why is this in fast-foundation?

marjonlynch: doh! left behind after the move of styles, will fix

marjonlynch: removed

chrisdholt (approved):

PR #4112, opened 2020-11-10
fix: ensure StyleElementCustomPropertyManager works with disconnected elements

### LINKED ISSUE #4109: Programtically creating a DesignSystemProvider element throws in FF
Repro in FF:
`document.createElement('fast-design-system-provider`)`

New `StyleElementCustomPropertyManager` code relies on accessing the `rules` property of an `HTMLStyleElement.sheet` instance, however when an element is not attached, `HTMLStyleElement.sheet` will be null. We need to defer accessing `HTMLSTyleElement.sheet.rules` until after the element is attached so that `sheet` is not null.

nicholasrice:
# Description
This change adds behaviors that allows using the `StyleElementCustomPropertyManager` with a client element that is not connected.

fixes #4109

## Motivation & context
Prior to this change, calling `document.createElement('fast-design-system-provider')` in FireFox would cause the `StyleElementCustomPropertyManager` to throw because it would attempt to access a property of `null`. The property is `null` when the `<style>` element provided to the manager is not connected to the DOM, which happens during programmatic creation of the `fast-design-system-provider` because the DSP *itself* is disconnected from the DOM.

This change checks if the problematic property is null, and if so, it attaches a subscriber to the `isConnected` property of the element's controller. When the subscriber invokes, only *then* is the customPropertyTarget accessed and assigned.

## Issue type checklist

## Process & policy checklist


scomea (approved):

packages/web-components/fast-foundation/src/custom-properties/manager.ts
```diff
@@ -262,28 +277,64 @@ export class ConstructableStylesCustomPropertyManager extends CustomPropertyMana
...
-        ] as CSSStyleRule).style;
+        controller.addStyles(style);
 
+        this.styles = style;
         this._owner = client;
 
+        // If the element isn't connected when the manager is created, the sheet can be null.
+        // In those cases, set up notifier for when the element is connected and set up the customPropertyTarget
+        // then.
+        controller.isConnected
```
EisenbergEffect, line 314: I'm not sure if this will ever evaluate to true in the constructor...

EisenbergEffect: Maybe in this case you want to check the direct element's `isConnected` and if not connected then subscribe to the controller's `isConnected`?

EisenbergEffect (changes requested):
Just one thing I'm not quite sure about. If we can resolve that, I can approve.

marjonlynch (approved):
This definitely fixes the issue in firefox, smoked

packages/web-components/fast-foundation/src/custom-properties/manager.ts
nicholasrice: Good point, I think that should work

nicholasrice: on second thought, *should* `isConnected` be true in the controller ctor if the element is connected? Seems like we might want to sync that up?

nicholasrice: Changed to check to element's `isConnected`, let me know if I should do a follow-up PR regarding the above.

janechu (approved):

packages/web-components/fast-foundation/src/custom-properties/manager.ts
EisenbergEffect: The Controller's isConnected indicates whether it has run through its own connection logic. So, it's a bit different than the element. It will indicated whether the template and styles have been hooked up in the first render, for example.

EisenbergEffect (approved):

PR #4114, opened 2020-11-12
fix: ensure tree items with children have a value for aria-expanded

### LINKED ISSUE #4113: No default expanded state for Tree Items with children
When you run Narrator on Tree View, for Tree Items with children, there is not default expanded state. As a result, Narrator doesn't state whether the Tree Item is expanded or not.

1. Go to https://explore.fast.design/components/fast-tree-view
2. Turn on Narrator
3. Click on "Tree Item 1"
4. Notice that Narrator doesn't state whether the item is expanded or collapsed

In the example above, would expect Narrator to state that the Tree Item is collapsed initially.

Able to see this in Edge Chromium Stable and Canary.

chrisdholt:
# Description
closes #4113 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/docs/api-report.md
+++ b/packages/web-components/fast-components/docs/api-report.md
@@ -12,9 +12,9 @@ import { BaseProgress } from '@microsoft/fast-foundation';
 import { Breadcrumb } from '@microsoft/fast-foundation';
 import { BreadcrumbItem } from '@microsoft/fast-foundation';
 import { Button } from '@microsoft/fast-foundation';
-import { Card } from '@microsoft/fast-foundation';
 import { Checkbox } from '@microsoft/fast-foundation';
 import { ColorRGBA64 } from '@microsoft/fast-colors';
+import { DesignSystem } from '@microsoft/fast-components-styles-msft';
 import { DesignSystemProvider } from '@microsoft/fast-foundation';
 import { Dialog } from '@microsoft/fast-foundation';
 import { Direction } from '@microsoft/fast-web-utilities';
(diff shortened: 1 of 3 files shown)


scomea (approved):

EisenbergEffect (approved):

packages/web-components/fast-components/docs/api-report.md
```diff
@@ -12,9 +12,9 @@ import { BaseProgress } from '@microsoft/fast-foundation';
 import { Breadcrumb } from '@microsoft/fast-foundation';
 import { BreadcrumbItem } from '@microsoft/fast-foundation';
 import { Button } from '@microsoft/fast-foundation';
-import { Card } from '@microsoft/fast-foundation';
 import { Checkbox } from '@microsoft/fast-foundation';
 import { ColorRGBA64 } from '@microsoft/fast-colors';
+import { DesignSystem } from '@microsoft/fast-components-styles-msft';
```
janechu, line 17: Is this right? It seems to be the legacy design system?

janechu: Looks like it's coming from a previous commit to this file: https://github.com/microsoft/fast/blob/master/packages/web-components/fast-components/src/card/index.ts

chrisdholt: Yeah it’s unrelated here. If we want to fix I’d suggest a PR to address the actual code rather than the API report. If we want to wait on this for that, I’m fine with that - this just needs to go into the release.

marjonlynch (approved):
smoked in Edge and Firefox

janechu (approved):

packages/web-components/fast-components/docs/api-report.md
janechu: Yup, just happened to catch it in the api-report, but it should be resolved before we publish, otherwise that change will break us since we dont take a dependency on that package in `@microsoft/fast-components`.

chrisdholt: @eljefe223 is going to fix this and will re-run the API report.

PR #4115, opened 2020-11-12
feat: convert FormAssociated to a constructable function

radium-v:
# Description
Converts `FormAssociated` to be a "constructable" function call that returns a mixin class.

## Motivation & context
Allows for easier control when working with mixin classes and `FormAssociated`.

## Issue type checklist

This modifies all `FormAssociated`-based components:
- Button
- Checkbox
- Radio
- Slider
- Text Area
- Text Field

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/utilities/constructable.ts
+++ b/packages/web-components/fast-foundation/src/utilities/constructable.ts
@@ -1,6 +1,0 @@
-/**
- * @internal
- */
-export type Constructable<T = {}> = {
-    new (...args: any[]): T;
-};
--- a/packages/web-components/fast-foundation/src/utilities/constructable.ts
+++ b/packages/web-components/fast-foundation/src/utilities/constructable.ts
@@ -1,0 +1,6 @@
+/**
+ * @internal
+ */
+export type Constructable<T = {}> = {
+    new (...args: any[]): T;
+};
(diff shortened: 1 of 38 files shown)


radium-v:
@nicholasrice there are still two tests that are breaking for me which were added recently, can you take a look and see if there's a quick fix?

packages/web-components/fast-foundation/src/utilities/constructable.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * @internal
+ */
+export type Constructable<T = {}> = {
```
EisenbergEffect, line 4: I think we could move this to fast-element and make it a public export. We've got multiple uses for this and I think it's something that fits in at the lowest level.

EisenbergEffect (changes requested):
I like this change. I made one request to move an interface. If we can make that change and get the tests all passing, I think we're good.

nicholasrice:
I *think* the `FormAssociated` test assertion can be changed to expect empty string. At this point I don't recall why that test is there, but if you check the `value` property of an `<input>` prior to connection it is empty string, so I think we should match that behavior.

nicholasrice (approved):
Thanks @radium-v - this is awesome.

chrisdholt (approved):
That's a lot of updates :)

Thanks @radium-v!

EisenbergEffect (approved):

PR #4124, opened 2020-11-16
fix: number field interactions

### LINKED ISSUE #3806: Can't delete first digit in attribute text field in fast-component-explorer
When entering numbers in an attribute text entry field I can delete all of the numbers except the first one, reset/undo button works fine but I expect that I could delete the last number and the text-field/attribute resets

Example:

1. Go to fast-component-explorer
2. select progress component
3. enter some numbers in the min attribute text field in the right panel
4. try to delete all the numbers with the delete key

Screen Shot 2020-08-27 at 1 23 40 PM

Mac OS, Edge (release)

williamw2:
# Description

Allow the number field to be empty while typing. Original logic would prevent the deletion of all numbers in the field since the onchange event handler would ignore the change if the current value was NaN. This changes allows an empty string to be a valid condition. This also fixes a similar issue if a default value is specified for the number field. 

## Motivation & context

Fixes #3806 

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):
Should we add a new test or two to assert the string deletion scenario stays fixed?

scomea (approved):

janechu (approved):

williamw2:
> Should we add a new test or two to assert the string deletion scenario stays fixed?

Added a test for the empty string scenario.

nicholasrice (approved):

PR #4128, opened 2020-11-18
doc(fast-element): add section on array observation

EisenbergEffect:
# Description

In response to some community questions, I've added a section to the documentation that covers array observation.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-element/docs/guide/observables-and-state.md
+++ b/packages/web-components/fast-element/docs/guide/observables-and-state.md
@@ -167,6 +167,7 @@ notifier.subscribe(handler);
 
 There are a couple of important details to note with array observation:
 
+* The `fast-element` library's ability to observe arrays is opt-in, in order that the functionality remain tree-shakeable. If you use a `repeat` directive anywhere in your code, you will be automatically opted in. However, if you wish to use the above APIs and are not using `repeat`, you will need to enable array observation by importing and calling the `enableArrayObservation()` function.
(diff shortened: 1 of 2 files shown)


scomea (approved):

packages/web-components/fast-element/docs/guide/observables-and-state.md
```diff
@@ -143,9 +143,37 @@ notifier.subscribe(handler, 'firstName')
...
+
+So far, we've only seen how to observe properties on objects, but it's also possible to observe arrays for changes. Given an instance of an array, it can be observed like this:
+
+**Example: Observing an Array**
+
+```ts
+const arr = [];
+const notifier = Observable.getNotifier(arr);
+const handler = {
+  handleChange(source: any, splices: Splice[]) {
```
nicholasrice, line 156: It doesn't appear that `Splice` is in the fast-element API docs because it isn't exported from the main `index.ts`. It's marked as public though, so it probably should be. Can we export that as part of this change as well?

EisenbergEffect: Good catch! I noticed another api that needed to be exported too and then that prompted me to think through adding some more docs. I'll get those handled and push the update shortly.

EisenbergEffect: Ok, got it taken care of.

nicholasrice (approved):

PR #4142, opened 2020-11-24
feat: add nested menu support

scomea:
# Description
This change adds support for nested menus to the menu and menu-item components.  

```html
    <fast-menu>
        <fast-menu-item>
            Menu item 1
            <fast-menu slot="submenu">
                <fast-menu-item>Nested Menu item 1.1</fast-menu-item>
                <fast-menu-item>Nested Menu item 1.2</fast-menu-item>
                <fast-menu-item>Nested Menu item 1.3</fast-menu-item>
            </fast-menu>
        </fast-menu-item>
        <fast-menu-item>
            Menu item 2
        </fast-menu-item>
        <fast-menu-item>
            Menu item 3
            <fast-menu slot="submenu">
                <fast-menu-item>
                    Nested Menu item 3.1
                    <fast-menu slot="submenu">
                        <fast-menu-item>Nested Menu item 3.1.1</fast-menu-item>
                        <fast-menu-item>Nested Menu item 3.1.2</fast-menu-item>
                        <fast-menu-item>Nested Menu item 3.1.3</fast-menu-item>
                    </fast-menu>
                </fast-menu-item>
                <fast-menu-item>
                    Nested Menu item 3.2
                    <fast-menu slot="submenu">
                        <fast-menu-item>Nested Menu item 3.2.1</fast-menu-item>
                        <fast-menu-item>Nested Menu item 3.2.2</fast-menu-item>
                        <fast-menu-item>Nested Menu item 3.2.3</fast-menu-item>
                    </fast-menu>
                </fast-menu-item>
                <fast-menu-item>
                    Nested Menu item 3.3
                    <fast-menu slot="submenu">
                        <fast-menu-item>Nested Menu item 3.3.1</fast-menu-item>
                        <fast-menu-item>Nested Menu item 3.3.2</fast-menu-item>
                        <fast-menu-item>Nested Menu item 3.3.3</fast-menu-item>
                    </fast-menu>
                </fast-menu-item>
            </fast-menu>
        </fast-menu-item>
    </fast-menu>
``` 

image

## Motivation & context
A requested feature.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
+++ b/packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
@@ -39,20 +39,20 @@
 <h3>checked</h3>
 <fast-menu-item checked="true" role="menuitemcheckbox">Menu item 1</fast-menu-item>
 
-    <h4>nested sub-menu</h4>
-    <fast-menu-item style="max-width: 240px;" aria-haspopup="true">
-        Menu item 1
-        <fast-menu>
-            <fast-menu-item>Nested Menu item 1</fast-menu-item>
-            <fast-menu-item>Nested Menu item 2</fast-menu-item>
-            <fast-menu-item>Nested Menu item 3</fast-menu-item>
-        </fast-menu>
-    </fast-menu-item>
+<h4>nested sub-menu</h4>
+<fast-menu-item style="max-width: 240px;" aria-haspopup="true">
+    Menu item 1
+    <fast-menu>
+        <fast-menu-item>Nested Menu item 1</fast-menu-item>
+        <fast-menu-item>Nested Menu item 2</fast-menu-item>
+        <fast-menu-item>Nested Menu item 3</fast-menu-item>
+    </fast-menu>
+</fast-menu-item>
 
-    <h4>menuitemradio</h4>
-    <fast-menu-item role="menuitemradio">
-        Menu item 1
-    </fast-menu-item>
+<h4>menuitemradio</h4>
+<fast-menu-item role="menuitemradio">
+    Menu item 1
+</fast-menu-item>
 
 <h3>checked</h3>
 <fast-menu-item checked="true" role="menuitemradio">Menu item 1</fast-menu-item>
--- a/packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
+++ b/packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
@@ -42,9 +42,11 @@
     <h4>nested sub-menu</h4>
     <fast-menu-item style="max-width: 240px;" aria-haspopup="true">
         Menu item 1
-        <fast-menu-item slot="nested-menu">Nested Menu item 1</fast-menu-item>
-        <fast-menu-item slot="nested-menu">Nested Menu item 2</fast-menu-item>
-        <fast-menu-item slot="nested-menu">Nested Menu item 3</fast-menu-item>
+        <fast-menu>
+            <fast-menu-item>Nested Menu item 1</fast-menu-item>
+            <fast-menu-item>Nested Menu item 2</fast-menu-item>
+            <fast-menu-item>Nested Menu item 3</fast-menu-item>
+        </fast-menu>
     </fast-menu-item>
 
     <h4>menuitemradio</h4>
--- a/packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
+++ b/packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
@@ -39,8 +39,18 @@
 <h3>checked</h3>
 <fast-menu-item checked="true" role="menuitemcheckbox">Menu item 1</fast-menu-item>
 
-<h2>menuitemradio</h2>
-<fast-menu-item role="menuitemradio">Menu item 1</fast-menu-item>
+    <h4>nested sub-menu</h4>
+    <fast-menu-item style="max-width: 240px;" aria-haspopup="true">
+        Menu item 1
+        <fast-menu-item slot="nested-menu">Nested Menu item 1</fast-menu-item>
+        <fast-menu-item slot="nested-menu">Nested Menu item 2</fast-menu-item>
+        <fast-menu-item slot="nested-menu">Nested Menu item 3</fast-menu-item>
+    </fast-menu-item>
+
+    <h4>menuitemradio</h4>
+    <fast-menu-item role="menuitemradio">
+        Menu item 1
+    </fast-menu-item>
 
 <h3>checked</h3>
 <fast-menu-item checked="true" role="menuitemradio">Menu item 1</fast-menu-item>
--- a/packages/web-components/fast-components/src/menu-item/menu-item.definition.ts
+++ b/packages/web-components/fast-components/src/menu-item/menu-item.definition.ts
@@ -69,6 +69,7 @@ export const fastMenuItemDefinition: WebComponentDefinition = {
                 },
                 {
                     name: "submenu",
+                    title: "Sub menu slot",
                     description: "The submenu slot",
                 },
             ],
--- a/packages/web-components/fast-components/src/menu-item/menu-item.definition.ts
+++ b/packages/web-components/fast-components/src/menu-item/menu-item.definition.ts
@@ -67,6 +67,10 @@ export const fastMenuItemDefinition: WebComponentDefinition = {
                     description:
                         "Contents of the end slot are positioned after the item content",
                 },
+                {
+                    name: "submenu",
+                    description: "The submenu slot",
+                },
             ],
         },
     ],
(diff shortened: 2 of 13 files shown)


EisenbergEffect:
Is there anyway to remove the need for ` submenu="true"`? For example, could a menu check its parent and if its parent is a menu item, call some method on the parent to let it know it has a submenu?

scomea:
> Is there anyway to remove the need for ` submenu="true"`? For example, could a menu check its parent and if its parent is a menu item, call some method on the parent to let it know it has a submenu?

@EisenbergEffect  I was thinking about that earlier.  Thought maybe the menu-item could check to see if it had any children assigned to the sub-menu slot.

chrisdholt:
> Is there anyway to remove the need for ` submenu="true"`? For example, could a menu check its parent and if its parent is a menu item, call some method on the parent to let it know it has a submenu?

Perhaps we can look at how Tree Item checks if it’s nested and applies the correct slot and move towards a solution like that?

scomea:
We now detect if an item has a submenu based on whether there are any elements in the submenu slot.

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
```diff
@@ -9,18 +9,42 @@ import { MenuItem, MenuItemRole } from "./menu-item";
...
+            x => x.expanded,
+            html<MenuItem>`
+                <fast-anchored-region
+                    :anchorElement="${x => x}"
+                    vertical-positioning-mode="dynamic"
+                    vertical-default-position="bottom"
+                    vertical-inset="true"
+                    horizontal-positioning-mode="dynamic"
+                    horizontal-default-position="end"
+                    class="submenu-region"
```
chrisdholt, line 41: This will likely need a part as well I presume?

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
```diff
@@ -9,18 +9,42 @@ import { MenuItem, MenuItemRole } from "./menu-item";
...
+                    dir="${x => x.currentDirection}"
+                    ${ref("subMenuRegion")}
+                >
```
chrisdholt, line 44: I think we're going to need to fundamentally change the DOM here for the sake of submenu items. A simple example of what submenu DOM structure needs to look like to be semantically accurate and provide the right hierarchy for Assistive Technology.

```html
<button 
    role="menuitem"
    aria-haspopup="true"
    aria-expanded="false"
>
    The menu item content
</button>
<ul 
    role="menu" 
    aria-label="Facts"
>
    <li role="none">
        <a role="menuitem" href="#">
        History
        </a>
    </li>
    <li role="none">
        <a role="menuitem" href="#">
        Current Statistics
        </a>
    </li>
    <li role="none">
        <a role="menuitem" href="#">
        Awards
        </a>
    </li>
</ul>
```

chrisdholt (changes requested):
I think we have some work to do on the DOM front to ensure we have the right semantic and accessible markup. While this seems to include the base keyboard pattern, the current DOM structure doesn't consider the story for conveying this to assistive technology. We essentially need to render a menu button inside these menu items and associate the slotted menu. You can test any of the submenu's here: https://w3c.github.io/aria-practices/#menu

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
scomea: So the example structure here isn't valid?  https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus

Which is roughly:

```html
<ul role="menubar">
  <li role="menuitem">
    Font
    <ul role="menu">
      <li role="menuitemradio">Sans-serif</li>
      <li role="menuitemradio">Serif</li>
    </ul>
  </li> r
  <li role="menuitem">
    Style
    <ul role="menu">
      <li role="menuitemradio">Sans-serif</li>
      <li role="menuitemradio">Serif</li> 
    </ul>
</ul>
```
Which is essentially the structure we create at runtime:
image

packages/web-components/fast-foundation/src/menu/menu.spec.md
```diff
@@ -109,15 +83,14 @@ The third item in the menu has a glyph, which would inset the text from the othe
...
+- default - slot for the content (the default slot for the item)
 - after - slot which comes after content
-- menu - the slot used to generate a submenu
+- submenu - the slot used to generate a submenu
 
 *CSS Parts*
 - before
 - content
 - after
-- menu
```
eljefe223, line 94: Shouldn't there be a `submenu` part?

packages/web-components/fast-foundation/src/menu/menu.spec.md
```diff
@@ -150,12 +123,12 @@ Nested menu:
     <fast-menu-item>Menu item two</fast-menu-item>
     <fast-menu-item>
         Menu item three
-        <fast-menu slot="menu">
+        <fast-menu slot="submenu">
```
chrisdholt, line 126: @scomea could we do what we're doing in tree-item template and automatically set this slot if we are nested within a `fast-menu-item`? Adding the slot would work, but the nesting would also "just work" and map to existing patterns of handling this with our own controls.

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
scomea: added part="submenu-region"

packages/web-components/fast-foundation/src/menu/menu.spec.md
scomea: added "submenu-region" for the layout area.

packages/web-components/fast-foundation/src/menu/menu.spec.md
scomea: There were some timing issues that made this trickier than initially expected, but yes, this works now.

scomea:
Note - occasionally menus in the example will render shifted over horizontally.  This is addressed by this fix to a race condition in anchored region - https://github.com/microsoft/fast/pull/4286

packages/web-components/fast-components/src/menu-item/fixtures/menu-item.html
```diff
@@ -40,8 +40,18 @@ <h2>menuitemcheckbox</h2>
     <h3>checked</h3>
     <fast-menu-item checked="true" role="menuitemcheckbox">Menu item 1</fast-menu-item>
 
-    <h2>menuitemradio</h2>
-    <fast-menu-item role="menuitemradio">Menu item 1</fast-menu-item>
+    <h4>nested sub-menu</h4>
+    <fast-menu-item style="max-width: 240px;" aria-haspopup="true">
+        Menu item 1
+        <fast-menu-item slot="nested-menu">Nested Menu item 1</fast-menu-item>
```
m4thieulavoie, line 46: Does `MenuItem` support this syntax _and_ the following for nested menu?
```html
 <fast-menu>
                <fast-menu-item>Nested Menu item 1.1</fast-menu-item>
                <fast-menu-item>Nested Menu item 1.2</fast-menu-item>
                <fast-menu-item>Nested Menu item 1.3</fast-menu-item>
            </fast-menu>
```

scomea: No, the direct nesting of menu items without a child menu is a bug in the example, fixing now.  The child submenu is required.  Fixing now.

chrisdholt (approved):
@scomea are there instances where hover to display menu items may be appropriate? IE, when I hover over the menu item it displays the nested menu? I'd still expect keyboarding to remain the same, but I wonder if we may want to support this hover/expose behavior.

packages/web-components/fast-components/src/menu-item/menu-item.definition.ts
```diff
@@ -56,6 +56,10 @@ export const fastMenuItemDefinition: WebComponentDefinition = {
                     name: "end",
                     description: "The end slot",
                 },
+                {
+                    name: "submenu",
+                    description: "The submenu slot",
+                },
```
chrisdholt, line 62: I think when you update/rebase the branch - this file will need to include title as well as description

scomea:
I added a hover behavior.  Do we need to be able to turn it off?  (IMO we don't and wait and see if it's wanted)

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
```diff
@@ -22,5 +25,25 @@ export const MenuItemTemplate = html<MenuItem>`
             <slot></slot>
         </span>
         ${endTemplate}
```
eljefe223, line 27: Do we want some way to include an icon if a menu item has a sub menu? I think this would be different then `endTemplate`

eljefe223 (approved):
Have one comment about including an additional slot when a menu item has a sub menu

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
scomea: Sounds good, but I'll do that in another pr.  Starting on that now.

PR #4143, opened 2020-11-25
fix: ensure DSP subscribes to upstream DSP properties

nicholasrice:
# Description
This PR fixes an issue where the `FASTCard` component would not re-evaluate registered recipes when upstream accent palette / base color would change.

When a DSP connects to an upstream provider, the downstream provider iterates over all the properties in the upstream DSP and subscribes to them. This causes the downstream DSP to bring down any upstream values into it's on design system object. But while the properties themselves were successfully being reflected upstream -> downstream, the downstream did not begin observing the *local* property for changes. In effect, changes to the property were transparent and not causing the DSP to re-evaluate recipes registered on it when those properties changed.

This change causes the DSP to check to see if the local DSP property is being observed yet when it connects to an upstream provider. If there is no local property for an upstream property, it means the DSP isn't observing it yet, so it defines the property as an observable and attaches a subscriber to re-evaluate recipes when it changes.

## Issue type checklist

## Process & policy checklist


amis92:
Thanks! I've raised the issue myself, repro is at https://codepen.io/amis92/pen/PozMrVj

@nicholasrice thank you for the PR! :)

scomea (approved):

EisenbergEffect (approved):

janechu (approved):

PR #4148, opened 2020-11-25
add default slot change method

litteredwitherrors:
# Description

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/button/index.ts
+++ b/packages/web-components/fast-components/src/button/index.ts
@@ -62,6 +62,8 @@ export class FASTButton extends Button {
         );
         if (slottedElements.length === 1 && slottedElements[0] instanceof SVGElement) {
             this.root.classList.add("icon-only");
+        } else {
+            this.root.classList.remove("icon-only");
         }
     }
 }
--- a/packages/web-components/fast-components/src/button/index.ts
+++ b/packages/web-components/fast-components/src/button/index.ts
@@ -50,6 +50,12 @@ export class FASTButton extends Button {
             this.appearance = "neutral";
         }
     }
+
+    private defaultSlottedContentChanged(oldValue, newValue): void {
+        if (this.defaultSlottedContent.length === 1) {
+            this.classList.toggle("has-svg");
+        }
+    }
 }
 
 /**
--- a/packages/web-components/fast-components/src/button/index.ts
+++ b/packages/web-components/fast-components/src/button/index.ts
@@ -45,15 +45,23 @@ export class FASTButton extends Button {
 
     public connectedCallback() {
         super.connectedCallback();
-
         if (!this.appearance) {
             this.appearance = "neutral";
         }
     }
 
+    /**
+     * Applies 'icon-only' class when there is only an SVG in the default slot
+     *
+     * @public
+     * @remarks
+     */
     public defaultSlottedContentChanged(oldValue, newValue): void {
-        if (this.defaultSlottedContent.length === 1) {
-            this.classList.add("is-svg");
+        const slottedElements = this.defaultSlottedContent.filter(
+            x => x.nodeType === Node.ELEMENT_NODE
+        );
+        if (slottedElements.length === 1 && slottedElements[0] instanceof SVGElement) {
+            this.root.classList.add("icon-only");
         }
     }
 }
--- a/packages/web-components/fast-components/src/button/index.ts
+++ b/packages/web-components/fast-components/src/button/index.ts
@@ -51,9 +51,9 @@ export class FASTButton extends Button {
         }
     }
 
-    private defaultSlottedContentChanged(oldValue, newValue): void {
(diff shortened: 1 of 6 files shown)


packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -168,6 +168,15 @@ export class Button extends FormAssociatedButton {
     private handleFormReset = () => {
         this.form?.reset();
     };
+
+    public default: HTMLSlotElement;
```
scomea, line 172: I think the prop name should be more descriptive, "button.default" in a prop list doesn't give someone much to go on.

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -168,6 +168,15 @@ export class Button extends FormAssociatedButton {
...
+    public defaultContainer: HTMLSpanElement;
+    public handleDefaultContentChange(): void {
+        this.defaultContainer.classList.toggle(
```
scomea, line 175: Same as prop name.   More descriptive would be better here IMO.

packages/web-components/fast-foundation/src/button/button.ts
chrisdholt: I think "content" or the ref is reasonable considering it maps to the part name. Otherwise, "defaultSlot" is fairly verbose. @litteredwitherrors we need to do a quick publish but would love to get this small update in so we can unblock a few things in Fluent UI. Could we make this change today?

EisenbergEffect:
Can someone provide the context for this? Is there any way to not mess with the class list?

chrisdholt:
> Can someone provide the context for this? Is there any way to not mess with the class list?

@EisenbergEffect there are scenarios where we may want to change the behavior of a button based on the type of content passed in. A concrete example, and the intent of this PR is to solve for a scenario in Fluent UI where buttons which only include icons as children have less padding than those which do not. 

Important to note that the class list here is the internal part and not the button itself. If the concern is other issues we've seen with instances such as React clobbering the class list - I don't think there should be issues with that as this is the internal button within the shadow and not the host itself.

EisenbergEffect:
Ah, yes. I missed that rather important detail 😄

scomea (approved):

chrisdholt (approved):

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -168,6 +168,15 @@ export class Button extends FormAssociatedButton {
...
     };
+
+    public content: HTMLSlotElement;
+    public contentContainer: HTMLSpanElement;
+    public handleContentChange(): void {
+        this.contentContainer.classList.toggle(
+            "content",
+            this.content.assignedNodes().length > 0
+        );
+    }
```
chrisdholt, line 179: I missed this before, but we want to do this in the component implementation. This can be a noop in foundation. I think we do something similar for slider label regarding orientation.

chrisdholt (changes requested):
One thing I realized after the fact that we'll need to change.

packages/web-components/fast-foundation/src/button/button.template.ts
```diff
@@ -45,7 +45,7 @@ export const ButtonTemplate = html<Button>`
     >
         ${startTemplate}
         <span class="content" part="content">
-            <slot></slot>
+            <slot ${ref("content")} @slotchange="${x => x.handleContentChange()}"></slot>
```
chrisdholt, line 48: The slotted directive may be more beneficial here for us - it will return an array of all slotted nodes. We ultimately want to apply a class in the event that the array has a single SVG in it.

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -168,6 +168,15 @@ export class Button extends FormAssociatedButton {
     private handleFormReset = () => {
         this.form?.reset();
     };
+
+    public content: HTMLSlotElement;
+    public contentContainer: HTMLSpanElement;
```
chrisdholt, line 173: probably want to put this on the Button

packages/web-components/fast-components/src/button/index.ts
```diff
@@ -45,15 +45,23 @@ export class FASTButton extends Button {
...
+     * @public
+     * @remarks
+     */
+    public defaultSlottedContentChanged(oldValue, newValue): void {
+        const slottedElements = this.defaultSlottedContent.filter(
+            x => x.nodeType === Node.ELEMENT_NODE
+        );
+        if (slottedElements.length === 1 && slottedElements[0] instanceof SVGElement) {
+            this.root.classList.add("icon-only");
         }
```
chrisdholt: @litteredwitherrors - should we have an else to remove the class in the case this is no longer what happens?

chrisdholt (approved):

PR #4159, opened 2020-12-02
feat: add default slot changed to anchor

litteredwitherrors:
# Description
This PR adds a feature that dynamically adds an "icon-only" class to the anchor element within the anchor component when its content is an SVG element.

## Motivation & context

This feature will give the ability to add additional styles to icon-only anchor components.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/anchor/index.ts
+++ b/packages/web-components/fast-components/src/anchor/index.ts
@@ -59,8 +59,8 @@ export class FASTAnchor extends Anchor {
     /**
      * Applies 'icon-only' class when there is only an SVG in the default slot
      *
-     * @public
-     * @remarks
+     * @internal
+     *
      */
     public defaultSlottedContentChanged(oldValue, newValue): void {
         const slottedElements = this.defaultSlottedContent.filter(
--- a/packages/web-components/fast-components/src/anchor/index.ts
+++ b/packages/web-components/fast-components/src/anchor/index.ts
@@ -67,9 +67,9 @@ export class FASTAnchor extends Anchor {
             x => x.nodeType === Node.ELEMENT_NODE
         );
         if (slottedElements.length === 1 && slottedElements[0] instanceof SVGElement) {
-            this.root.classList.add("icon-only");
+            this.control.classList.add("icon-only");
         } else {
-            this.root.classList.remove("icon-only");
+            this.control.classList.remove("icon-only");
         }
     }
 }
--- a/packages/web-components/fast-components/src/anchor/index.ts
+++ b/packages/web-components/fast-components/src/anchor/index.ts
@@ -55,6 +55,23 @@ export class FASTAnchor extends Anchor {
             this.appearance = "neutral";
         }
     }
+
+    /**
+     * Applies 'icon-only' class when there is only an SVG in the default slot
+     *
+     * @public
+     * @remarks
+     */
+    public defaultSlottedContentChanged(oldValue, newValue): void {
+        const slottedElements = this.defaultSlottedContent.filter(
+            x => x.nodeType === Node.ELEMENT_NODE
+        );
+        if (slottedElements.length === 1 && slottedElements[0] instanceof SVGElement) {
+            this.root.classList.add("icon-only");
+        } else {
+            this.root.classList.remove("icon-only");
+        }
+    }
 }
 
 /**
--- a/packages/web-components/fast-components/src/styles/patterns/button.ts
+++ b/packages/web-components/fast-components/src/styles/patterns/button.ts
@@ -94,8 +94,9 @@ export const BaseButtonStyles = css`
         display: flex;
     }
 
-    a.icon-only {
-        padding-top: 6px;
+    .control.icon-only {
+        padding: 0;
+        line-height: 0;
     }
 
     ::slotted(svg) {
--- a/packages/web-components/fast-components/src/styles/patterns/button.ts
+++ b/packages/web-components/fast-components/src/styles/patterns/button.ts
@@ -94,6 +94,10 @@ export const BaseButtonStyles = css`
         display: flex;
     }
 
+    a.icon-only {
+        padding-top: 6px;
+    }
+
     ::slotted(svg) {
         ${
             /* Glyph size and margin-left is temporary -
--- a/packages/web-components/fast-foundation/src/anchor/anchor.template.ts
+++ b/packages/web-components/fast-foundation/src/anchor/anchor.template.ts
@@ -1,4 +1,4 @@
-import { html } from "@microsoft/fast-element";
+import { html, ref, slotted } from "@microsoft/fast-element";
 import { endTemplate, startTemplate } from "../patterns/start-end";
 import { Anchor } from "./anchor";
 
@@ -38,10 +38,11 @@ export const AnchorTemplate = html<Anchor>`
         aria-owns="${x => x.ariaOwns}"
         aria-relevant="${x => x.ariaRelevant}"
         aria-roledescription="${x => x.ariaRoledescription}"
+        ${ref("root")}
     >
         ${startTemplate}
         <span class="content" part="content">
-            <slot></slot>
+            <slot ${slotted("defaultSlottedContent")}></slot>
         </span>
         ${endTemplate}
     </a>
--- a/packages/web-components/fast-foundation/src/anchor/anchor.template.ts
+++ b/packages/web-components/fast-foundation/src/anchor/anchor.template.ts
@@ -38,7 +38,7 @@ export const AnchorTemplate = html<Anchor>`
         aria-owns="${x => x.ariaOwns}"
         aria-relevant="${x => x.ariaRelevant}"
         aria-roledescription="${x => x.ariaRoledescription}"
-        ${ref("root")}
+        ${ref("control")}
     >
         ${startTemplate}
         <span class="content" part="content">
--- a/packages/web-components/fast-foundation/src/anchor/anchor.ts
+++ b/packages/web-components/fast-foundation/src/anchor/anchor.ts
@@ -85,7 +85,7 @@ export class Anchor extends FASTElement {
      *
      * Default slotted content
      *
-     * @public
+     * @internal
      */
     @observable
     public defaultSlottedContent: HTMLElement[];
--- a/packages/web-components/fast-foundation/src/anchor/anchor.ts
+++ b/packages/web-components/fast-foundation/src/anchor/anchor.ts
@@ -1,4 +1,4 @@
-import { attr, FASTElement } from "@microsoft/fast-element";
+import { attr, FASTElement, observable } from "@microsoft/fast-element";
 import { ARIAGlobalStatesAndProperties, StartEnd } from "../patterns/index";
 import { applyMixins } from "../utilities/apply-mixins";
 
@@ -80,6 +80,21 @@ export class Anchor extends FASTElement {
      */
     @attr
     public type: string;
+
+    /**
+     *
+     * Default slotted content
+     *
+     * @public
+     * @remarks
+     */
+    @observable
+    public defaultSlottedContent: HTMLElement[];
+
+    /**
+     * References the root element
+     */
+    public root: HTMLButtonElement;
 }
 
 /**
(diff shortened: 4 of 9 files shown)


packages/web-components/fast-components/src/anchor/index.ts
```diff
@@ -55,6 +55,23 @@ export class FASTAnchor extends Anchor {
...
         }
     }
+
+    /**
+     * Applies 'icon-only' class when there is only an SVG in the default slot
+     *
+     * @public
+     * @remarks
+     */
+    public defaultSlottedContentChanged(oldValue, newValue): void {
```
EisenbergEffect, line 65: I would recommend looking into using the `filter` option on the slotted directive. It works the same as for the `children` directive. Here's a link to the docs: https://www.fast.design/docs/fast-element/using-directives#the-children-directive

chrisdholt: @EisenbergEffect - I may misunderstand it, but I thought filter would only return those slotted elements. We want to support everything as slotted content here, but we only want to take action when it's an SVG and there is one item within the anchor.

packages/web-components/fast-foundation/src/anchor/anchor.template.ts
```diff
@@ -38,10 +38,11 @@ export const AnchorTemplate = html<Anchor>`
         aria-owns="${x => x.ariaOwns}"
         aria-relevant="${x => x.ariaRelevant}"
         aria-roledescription="${x => x.ariaRoledescription}"
+        ${ref("root")}
```
nicholasrice, line 41: Kinda a nit, but can we call this "control" instead of "root" so that aligns with the class and part attributes?

packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -80,6 +80,21 @@ export class Anchor extends FASTElement {
...
+     * @public
+     * @remarks
+     */
+    @observable
+    public defaultSlottedContent: HTMLElement[];
+
+    /**
+     * References the root element
+     */
+    public root: HTMLButtonElement;
```
nicholasrice, line 97: Should this be `HTMLAnchorElement`?

packages/web-components/fast-foundation/src/anchor/anchor.template.ts
chrisdholt: This is my bad - probably need a change to button as well :|

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -94,6 +94,10 @@ export const BaseButtonStyles = css`
         display: flex;
     }
 
+    a.icon-only {
+        padding-top: 6px;
+    }
```
chrisdholt, line 99: I was able to make an adjustment on this in Fluent UI to center consistently across browsers. Here's what I used:

```css
    .control.icon-only {
        padding: 0;
        line-height: 0;
    }
```

chrisdholt (changes requested):
One note on the styling to get things centered regardless of the content type. If we can update the button ref from "root" to "control" (my bad!) as part of this as well that would be great!

packages/web-components/fast-components/src/styles/patterns/button.ts
litteredwitherrors: Cool, I'll add it.

packages/web-components/fast-foundation/src/anchor/anchor.ts
litteredwitherrors: You are correct, will change.

packages/web-components/fast-foundation/src/anchor/anchor.template.ts
litteredwitherrors: Will change.

packages/web-components/fast-components/src/anchor/index.ts
litteredwitherrors: @EisenbergEffect @chrisdholt Since this is specifically for the use case of stand-alone SVG icons, would adding a filter for SVG elements in the template be too specific in FAST Foundation?

chrisdholt: Yep, I think this is a good callout - I'd prefer the generic (non-filtered) for foundation.

EisenbergEffect: Good point!

nicholasrice:
@litteredwitherrors can you add "feat:" at the beginning of your PR title so this PR gets designated correctly by conventional-commits?

packages/web-components/fast-components/src/anchor/index.ts
```diff
@@ -55,6 +55,23 @@ export class FASTAnchor extends Anchor {
             this.appearance = "neutral";
         }
     }
+
+    /**
+     * Applies 'icon-only' class when there is only an SVG in the default slot
+     *
+     * @public
+     * @remarks
```
nicholasrice, line 63: you don't need `@remarks` if there is no remark content.

packages/web-components/fast-components/src/anchor/index.ts
```diff
@@ -55,6 +55,23 @@ export class FASTAnchor extends Anchor {
             this.appearance = "neutral";
         }
     }
+
+    /**
+     * Applies 'icon-only' class when there is only an SVG in the default slot
+     *
+     * @public
```
nicholasrice, line 62: Is this an API we want surfaced to documentation?

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -182,7 +182,7 @@ export class Button extends FormAssociatedButton {
     /**
      * References the root element
      */
-    public root: HTMLButtonElement;
+    public control: HTMLButtonElement;
```
nicholasrice, line 185: this is *technically* a breaking change actually :/ We probably need to do this in an additive way or not make the change to align to "control"

litteredwitherrors: @nicholasrice @chrisdholt Gotcha. What would be the best path forward?

chrisdholt: I mean - this was ultimately published end of last week. Would it be reasonable to see if folks have even used this? Otherwise we take an issue IMO and fix later.

chrisdholt: I know *I* have used it, but perhaps we should ask others.

nicholasrice: I think my preference is to stick to the policy on this one and not break the API surface. Something like this can be done to get the best of both worlds:

```ts
 /**
  * @deprecated
  */
  public get root() { return this.control; }
  
 /**
 * References the root element
 */
  public control: HTMLButtonElement;
```

we could maybe add a note that `control` should be `@internal` as well - while it's documented now, it really *shouldn't* be because it leaks implementation details to the public API.

chrisdholt: Thanks @nicholasrice!

chrisdholt: @nicholasrice looks like this was updated - can you take another pass when you have a minute?

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -180,9 +180,17 @@ export class Button extends FormAssociatedButton {
     };
 
     /**
-     * References the root element
+     * @deprecated
      */
```
chrisdholt, line 184: @litteredwitherrors - looks like the deprecation block is missing a note so the build is broken. If we commit the below it should fix.

```suggestion
    /**
     * This has been deprecated in favor of control property
     * @deprecated
     */
```

litteredwitherrors: Sounds good. Commit away.

litteredwitherrors: Nevermind, looks like I'll have to push a change.

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -180,9 +180,18 @@ export class Button extends FormAssociatedButton {
     };
 
     /**
-     * References the root element
+     * This has been deprecated in favor of control property
+     * @deprecated
```
chrisdholt, line 184:
```suggestion
     * References the root element
     * @deprecated This has been deprecated in favor of control property
```

Going to try and commit this since it still shows failure.

packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -80,6 +80,21 @@ export class Anchor extends FASTElement {
...
+     * Default slotted content
+     *
+     * @public
+     * @remarks
+     */
+    @observable
+    public defaultSlottedContent: HTMLElement[];
+
+    /**
+     * References the root element
```
nicholasrice, line 95:
```suggestion
     * References the root element
     * @internal
```

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -180,9 +180,19 @@ export class Button extends FormAssociatedButton {
     };
 
     /**
+     * This has been deprecated in favor of control property
      * References the root element
+     * @deprecated This has been deprecated in favor of control property
```
nicholasrice, line 185: This explanation might be confusing because `control` isn't public and documented. Probably okay to just omit the message.
```suggestion
     * @deprecated
```

chrisdholt: Unfortunately this throws an error in the build if it doesn't include the message. As internal it may not...I'm open to try it, but we may need to add it back in

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -180,9 +180,19 @@ export class Button extends FormAssociatedButton {
     };
 
     /**
+     * This has been deprecated in favor of control property
```
chrisdholt, line 183:
```
suggestion
```

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -181,8 +181,17 @@ export class Button extends FormAssociatedButton {
 
     /**
      * References the root element
```
chrisdholt, line 183:
```
     * @internal
```

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -180,9 +180,18 @@ export class Button extends FormAssociatedButton {
     };
 
     /**
-     * References the root element
+     * @internal
```
chrisdholt, line 183:
```suggestion
```

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -180,9 +180,17 @@ export class Button extends FormAssociatedButton {
     };
 
     /**
-     * References the root element
```
chrisdholt, line 183:
```suggestion
     * References the root element
```

packages/web-components/fast-components/src/anchor/index.ts
chrisdholt:
```suggestion
```

packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -80,6 +80,21 @@ export class Anchor extends FASTElement {
      */
     @attr
     public type: string;
+
+    /**
+     *
+     * Default slotted content
+     *
+     * @public
+     * @remarks
```
chrisdholt, line 89:
```suggestion
```

packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -80,6 +80,20 @@ export class Anchor extends FASTElement {
      */
     @attr
     public type: string;
+
+    /**
+     *
+     * Default slotted content
+     *
+     * @public
```
nicholasrice, line 88: Same question as above here, should this be a public facing API or is this an implementation detail of the component? If it is the latter, this should be marked as `@internal`

nicholasrice (approved):

PR #4163, opened 2020-12-03
chore: ensure diff checking includes markdown files so that the generated API report is included

janechu:
# Description

Pull requests sometimes include API report updates that were not part of the pull request but instead were part of changes made in a previously merged pull request. This should catch and fail the build when the API report has not been updated to ensure it is accurate when a pull request is merged to master.

## Issue type checklist

## Process & policy checklist


nicholasrice (approved):
Nice! My only thought is should this be set and run from the project root instead of individual packages?

scomea (approved):

janechu:
@nicholasrice I did consider that but also that people are running all of these commands from within the specific package folder instead of from the root, plus we don't run a root test on the root package JSON. If we're OK doing that, we could change this but I think having it in specific folders is more intuitive during development. Thoughts?

chrisdholt (approved):

EisenbergEffect (approved):

PR #4164, opened 2020-12-04
feat: allow filtering MDN data by status

janechu:
# Description

This allows the filtering of MDN data by it's status, the default export of MDN data from the tooling is filtered to only include "standard" MDN CSS.

## Issue type checklist

## Process & policy checklist


packages/tooling/fast-tooling/src/data-utilities/mapping.mdn-data.ts
```diff
@@ -523,24 +528,54 @@ export function resolveCSSPropertySyntax(
...
-                    name: key,
-                    appliesTo: value.appliesto,
-                    syntax: resolveCSSPropertySyntax(value, key, syntaxKeys, typeKeys),
-                };
+                if (conformsToOptions(cssPropertyMapOptions, key, value)) {
+                    return {
+                        name: key,
+                        appliesTo: value.appliesto,
+                        syntax: resolveCSSPropertySyntax(
+                            value,
```
nicholasrice, line 567: Is perf a significant consideration here? If so I would recommend combining this all into the reduce function. As-is there are three iterations over the array where there could one.

nicholasrice (approved):
Just one note re perf. Not sure how big that array is but it looks like this is a build script so it may not really be an issue. Feel free to disregard the above if it isnt.

packages/tooling/fast-tooling/src/data-utilities/mapping.mdn-data.ts
janechu: Not really a consideration as it's run ad-hoc, but no reason to not make the change :)

PR #4166, opened 2020-12-04
feat: adds DI system, Configuration, and FoundationElement

nicholasrice:
# Description
This change adds several new alpha features that make progress towards https://github.com/microsoft/fast/issues/3361. All features are exported with the `@alpha` release flag, meaning they can be imported but they are undocumented and subject to API changes / removal.

The follow features are added:
- general-purpose dependency injection system
- `Configuration` implementation to manage element creation, templating, and styling. This will likely be extended to provide general app-level services to components.
- FoundationElement to connect with `Configuration`, manages getting templates and styles from `Configuration`.

A few questions for @EisenbergEffect:
1. I wasn't sure specifically who to attribute the DI code too, can you have a look at that and advise on any necessary changes?
2. You mentioned there were *tests* for the DI - can you point me to those so I can include them in this PR?

**And a general question:**
Are we okay with this being exported from the package as alpha? The intent here is to facilitate future explorations for Provider and Design System work using these tools, but it does provide opportunity for external usage. I personally think exporting under an alpha flag is sufficient but open to other thoughts.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/di/di.ts
```diff
@@ -0,0 +1,1414 @@
...
+     */
+    respectConnection?: boolean;
+}
+
+function createInterface<K extends Key, T = any>(
+    friendlyName?: string
+): DefaultableInterfaceSymbol<K, T>;
+function createInterface<K extends Key, T = any>(
+    configuration?: InterfaceConfiguration & { respectConnection: false }
+): DefaultableInterfaceSymbol<K, T>;
```
nicholasrice, line 354: @EisenbergEffect All of my changes to make re-resolution of DI interface after connection change are in this function.

EisenbergEffect:
I'll add the attribution and the tests. Please stand by. 😄

janechu (approved):

EisenbergEffect:
Ok, fixed up the docs errors. All tests passing. I think we're ready for review.

EisenbergEffect (approved):

EisenbergEffect:
Successive PRs addressed Nick's questions, added tests, etc. With the latest code we'll be able to do something like this:

In the fast or fluent component packages, the following code will replace our inheritance and decorator usage:

```
export const button = FoundationElement.configuration({
  baseName: 'button',
  template: ButtonTemplate,
  styles: FluentButtonStyles
});
```

And then the app dev will be able to  do this:

```ts
import { button, FluentDesignSystem } from '@fluentui/web-components';

FluentDesignSystem
  register(
    button()
  ).applyTo(document.body);
```

The `button` function, when invoked, creates a registry capable of configuring the DI container. The registry also has access to the design system itself, so it can customize the component definition based on that (e.g. use the proper element name prefix based on the design system or end-user configuration). When the registry is installed, it defines the element by merging the default configuration with any user-specified changes (different name, prefix, styles, template, etc.)

The `FluentDesignSystem` would provide a fluent API for configuring all the components and design system settings. In the end, it's a convenience API that configures the DI container and associates that container with the specified DOM tree. In this way DI really powers everything, but there's a nice API for getting things configured.

For our single-file script builds, we'll provide an index module that contains the design system code shown above, but lists every component. That would enable the script tag version to work the same as it does today. But, developers and designers who have customization will have much more power and flexibility.

PR #4167, opened 2020-12-07
feat: make setComponent and dependency properties and methods public

proempire:
# Description

This PR wants to solve this issue by making setComponent public, then we can override the behavior of setComponent, decide whether to make the tab focused or not.

## Motivation & context

Sometimes we need one tab automatically switch to another tab without being focused (i.e., A carousel with auto-scroll feature). In current design, we can't override the focus behavior because it's encapsulated in tabs.ts setComponent.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/tabs/tabs.ts
+++ b/packages/web-components/fast-foundation/src/tabs/tabs.ts
@@ -392,7 +392,7 @@ export class Tabs extends FASTElement {
         this.setComponent();
     };
 
-    private focusTab(): void {
+    public focusTab(): void {
         this.tabs[this.activeTabIndex].focus();
     }
 
--- a/packages/web-components/fast-foundation/src/tabs/tabs.ts
+++ b/packages/web-components/fast-foundation/src/tabs/tabs.ts
@@ -124,13 +124,13 @@ export class Tabs extends FASTElement {
      */
     public activetab: HTMLElement;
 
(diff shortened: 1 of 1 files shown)


SethDonohue (approved):

chrisdholt:
What are the accessibility implications supporting this behavior?

SethDonohue:
After further discussion I have reset my review. There is likely a better and potentially faster way to use Tabs to be able to access these methods for use and control on other components that extend it. One being  _forking_ Tabs, making any changes needed that way, and bringing those changes over to the component that is extending/using Tabs.

chrisdholt (changes requested):
Thanks for the PR @proempire - I'm happy to learn a bit more about the goals here but I'm not sure that making everything which is currently private set to public is the right move here. While I get that in a carousel implementation of Tabs one might want to auto-rotate and *not* focus the control, I think that if Carousel existed with this behavior, the value of making this public would be greatly diminished. If there are more use cases where having focus and selection be separate, I think we could consider making this change - even then, I think we'd want to look at what the right public API is and how that affects the internal controls. As it currently stands, this seems to just make everything private public so it can be changed and I don't think that's necessarily the right path forward. I think we'd want a more condensed and clear public API.

SethDonohue:
Closing as this is no longer needed.

PR #4172, opened 2020-12-07
fix(fast-foundation): ensure global aria attrs are applied

EisenbergEffect:
# Description

This PR fixes an issue where global aria attributes are not syncing state.
Fixes https://github.com/microsoft/fluentui/issues/16121

Several of the attribute mixins were applied through mixin inheritance. However, the `applyMixins` helper only copies own properties of the prototype. After some discussion @nicholasrice and I decided the best course of action was to not change the `applyMixins` function (could have some unexpected consequence for consumers) but to use it to apply the base mixins instead of using inheritance. This should provide for the highest compatibility with existing code.

Note: This is a temporary fix. We hope to provide a new feature in fast-element to allow templates to hook into observation of built-in attribute changes without re-defining the aria properties like this.

## Issue type checklist

The mixed class no longer inherits from the global base. However, all the same properties will be present. The only issue that could arise is if someone, through TypeScript, were somehow dependent on those properties being statically defined on the mixin itself, rather than applied through the mixin helper.

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/anchor/anchor.ts
+++ b/packages/web-components/fast-foundation/src/anchor/anchor.ts
@@ -87,7 +87,7 @@ export class Anchor extends FASTElement {
  *
  * @public
  */
-export class DelegatesARIALink extends ARIAGlobalStatesAndProperties {
+export class DelegatesARIALink {
     /**
      * See {@link https://www.w3.org/WAI/PF/aria/roles#link} for more information
      * @public
@@ -98,6 +98,8 @@ export class DelegatesARIALink extends ARIAGlobalStatesAndProperties {
     public ariaExpanded: "true" | "false" | undefined;
 }
 
+applyMixins(DelegatesARIALink, ARIAGlobalStatesAndProperties);
+
 /**
  * Mark internal because exporting class and interface of the same name
  * confuses API documenter.
@@ -105,5 +107,8 @@ export class DelegatesARIALink extends ARIAGlobalStatesAndProperties {
  * @internal
  */
 /* eslint-disable-next-line */
-export interface Anchor extends StartEnd, DelegatesARIALink {}
+export interface Anchor
+    extends StartEnd,
+        DelegatesARIALink,
+        ARIAGlobalStatesAndProperties {}
 applyMixins(Anchor, StartEnd, DelegatesARIALink);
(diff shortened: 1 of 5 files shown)


packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -98,12 +98,17 @@ export class DelegatesARIALink extends ARIAGlobalStatesAndProperties {
...
  * TODO: https://github.com/microsoft/fast/issues/3317
  * @internal
  */
 /* eslint-disable-next-line */
-export interface Anchor extends StartEnd, DelegatesARIALink {}
+export interface Anchor
+    extends StartEnd,
+        DelegatesARIALink,
+        ARIAGlobalStatesAndProperties {}
 applyMixins(Anchor, StartEnd, DelegatesARIALink);
```
chrisdholt, line 114: Thanks @EisenbergEffect! Quick question since I'm just looking at the PR; does what's on 101 automatically get inherited on here on 114 with the name "DelegatesARIALink"?

EisenbergEffect: Yes. Line 101 mixes the globals into DelegatesARIALink so it can be used on its own as a complete set of attributes. Then line 114 mixes the entire thing into the Anchor.

chrisdholt: Great, thanks for the quick action on this!

chrisdholt (approved):

packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -87,7 +87,7 @@ export class Anchor extends FASTElement {
  *
  * @public
  */
-export class DelegatesARIALink extends ARIAGlobalStatesAndProperties {
+export class DelegatesARIALink {
```
nicholasrice, line 90: I think since you're mixing `ARIAGlobalStatesAndProperties` into `DelegatesARIALink` you'll want to merge the declarations as well:

```ts
interface DelegatesARIALink extends ARIAGlobalStatesAndProperties {}
```

As-is the runtime type is inconsistent w/ the TS types.

EisenbergEffect: Good call. Looks like some of the tests are failing related to this. I'll get it fixed shortly.

nicholasrice (approved):

PR #4176, opened 2020-12-09
fix: select should emit change event

radium-v:
# Description
The `change` event was being emitted by the `fast-option` elements as they changed instead of the `fast-select`, which was causing performance issues when adding a lot of `fast-select` elements on the page.

This PR also includes a fix to allow regular `<option>` elements to be used in a listbox or select for compatibility. It's not recommended to use `<option>`s but at least now things won't break if you do.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/listbox/listbox.ts
+++ b/packages/web-components/fast-foundation/src/listbox/listbox.ts
@@ -1,5 +1,6 @@
-import { uniqueId } from "lodash-es";
 import { attr, DOM, FASTElement, observable } from "@microsoft/fast-element";
+import difference from "lodash-es/difference";
+import uniqueId from "lodash-es/uniqueId";
 import { ListboxOption } from "../listbox-option/listbox-option";
 import { ARIAGlobalStatesAndProperties } from "../patterns/aria-global";
 import { applyMixins } from "../utilities/apply-mixins";
@@ -67,8 +68,8 @@ export class Listbox extends FASTElement {
      * @internal
      */
     @observable
-    public options: ListboxOption[] = [];
-    public optionsChanged(prev, next): void {
+    public options: ListboxOption[];
+    public optionsChanged(prev: ListboxOption[] = [], next: ListboxOption[]) {
         if (this.$fastController.isConnected) {
             this.options.forEach(o => (o.id = o.id || uniqueId("option-")));
         }
@@ -81,9 +82,16 @@ export class Listbox extends FASTElement {
      */
     @observable
     public selectedOptions: ListboxOption[] = [];
-    protected selectedOptionsChanged(prev, next) {
+    protected selectedOptionsChanged(
+        prev: ListboxOption[] = [],
+        next: ListboxOption[]
+    ): void {
+        if (prev.length && next.length && !difference(prev, next).length) {
+            return;
+        }
+
         if (this.$fastController.isConnected) {
-            this.options.forEach(o => (o.selected = next.includes(o)));
+            this.options.map(o => (o.selected = next.includes(o)));
         }
     }
 
@@ -143,9 +151,6 @@ export class Listbox extends FASTElement {
      */
     public setSelectedOption(index: number = this.selectedIndex): void {
         const selectedOption = this.options[index];
-        if (!selectedOption) {
-            return;
-        }
 
         const selectedOptions: ListboxOption[] = [];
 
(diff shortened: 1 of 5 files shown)


packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -81,9 +82,16 @@ export class Listbox extends FASTElement {
      */
     @observable
     public selectedOptions: ListboxOption[] = [];
-    protected selectedOptionsChanged(prev, next) {
+    protected selectedOptionsChanged(
+        prev: ListboxOption[] = [],
+        next: ListboxOption[]
+    ): void {
+        if (prev.length && next.length && !difference(prev, next).length) {
```
EisenbergEffect, line 89: How expensive is the `difference` function? Should this be moved inside the `isConnected` branch?

EisenbergEffect (approved):

janechu (approved):

chrisdholt (approved):

packages/web-components/fast-foundation/src/listbox/listbox.ts
radium-v: This can be changed to use native methods. I'll update.

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -129,10 +132,12 @@ export class Listbox extends FASTElement {
     /**
      * @internal
      */
-    protected setDefaultSelectedOption(): void {
-        let selectedIndex = this.options.findIndex(el => el.selected);
-        selectedIndex = selectedIndex !== -1 ? selectedIndex : 0;
-        this.setSelectedOption(selectedIndex);
+    protected async setDefaultSelectedOption() {
+        const selectedIndex = await Promise.resolve(
```
EisenbergEffect, line 136: Seems like the Promise isn't needed here and this could be made into a non-async method.

packages/web-components/fast-foundation/src/select/select.spec.ts
```diff
@@ -88,6 +89,279 @@ describe("Select", () => {
...
+
+            expect(wasChanged).to.be.true;
+
+            expect(element.value).to.equal("three");
+
+            await disconnect();
+        });
+    });
+
+    describe("should NOT emit a 'change' event when the value changes by programmatic interaction", () => {
```
EisenbergEffect, line 338: For my own confirmation, this is the way the built-ins work as well, yes?

packages/web-components/fast-foundation/src/select/select.ts
```diff
@@ -24,10 +24,73 @@ export class Select extends FormAssociatedSelect {
...
+            this.focusAndScrollOptionIntoView();
+            this.indexWhenOpened = this.selectedIndex;
         }
     }
 
+    private indexWhenOpened: number;
+
+    private _value: string;
+
+    @volatile
```
EisenbergEffect, line 36: You shouldn't need `volatile` in this case.

EisenbergEffect (approved):
Approving pending a couple of quick minor fixes.
Thanks for filling out the tests and the attention to detail around the interactions.

packages/web-components/fast-foundation/src/listbox/listbox.ts
radium-v: This had a strong code smell when I put it in, and your comment led me towards fixing a lot of async-ish happenings throughout listbox and select.

packages/web-components/fast-foundation/src/select/select.spec.ts
radium-v: Correct - if you have a `<select>` and you change its selected value with `select.value = "any sort of value"`, it won't emit an event.

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -365,6 +393,15 @@ export class DelegatesARIAListbox {
      */
     @observable
     public ariaExpanded: "true" | "false" | undefined;
+
+    /**
+     * See {@link https://www.w3.org/WAI/PF/aria/roles#button} for more information
```
chrisdholt, line 398: Nit, I think this link would be `https://www.w3.org/WAI/PF/aria/roles#listbox`

chrisdholt (approved):

EisenbergEffect (approved):

Issue #4160, opened 2020-12-02
Button in accordion item should be marked aria-disabled when expanded in single select mode


eljefe223:
**Describe the bug; what happened?**
The button in accordion item should be marked `aria-disabled` when expanded in single select mode since the button is no longer interact-able. This is outlined in the aria spec.

```
If the accordion panel associated with an accordion header is visible, and if the accordion does not permit the panel to be collapsed, the header button element has aria-disabled set to true.
```

PR #4180, opened 2020-12-10
fix(accordion): add aria-diabled to an accordion-item when expanded in single mode

yeraycaballero:
# Description
On single mode, an expanded accordion item is not clickable, however its button it is not marked as aria-disabled. This commit adds aria-disabled to the accordion-item and it's internal button

## Motivation & context

https://github.com/microsoft/fast/issues/4160

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/accordion/accordion.ts
+++ b/packages/web-components/fast-foundation/src/accordion/accordion.ts
@@ -74,6 +74,17 @@ export class Accordion extends FASTElement {
         this.accordionItems.forEach((item: HTMLElement, index: number) => {
             if (item instanceof AccordionItem) {
                 item.addEventListener("change", this.activeItemChange);
+
+                if (this.isSingleExpandMode()) {
+                    if (this.activeItemIndex === index) {
+                        item.expanded = true;
+                        item.setAttribute("aria-disabled", "true");
+                    } else {
+                        item.expanded = false;
+                        item.removeAttribute("aria-disabled");
+                    }
+                }
+
                 if (this.isSingleExpandMode()) {
                     this.activeItemIndex !== index
                         ? (item.expanded = false)
@@ -92,6 +103,9 @@ export class Accordion extends FASTElement {
 
     private resetItems(): void {
         this.accordionItems.forEach((item: AccordionItem, index: number) => {
+            if (item.expanded === true) {
+                item.removeAttribute("aria-disabled");
+            }
             item.expanded = false;
         });
     }
@@ -108,6 +122,7 @@ export class Accordion extends FASTElement {
         if (this.isSingleExpandMode()) {
             this.resetItems();
             event.target.expanded = true;
+            event.target.setAttribute("aria-disabled", "true");
         }
         this.activeid = event.target.getAttribute("id");
         this.activeItemIndex = Array.from(this.accordionItems).indexOf(selectedItem);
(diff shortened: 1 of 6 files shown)


packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -92,6 +103,9 @@ export class Accordion extends FASTElement {
 
     private resetItems(): void {
         this.accordionItems.forEach((item: AccordionItem, index: number) => {
+            if (item.expanded === true) {
+                item.removeAttribute("aria-disabled");
```
chrisdholt, line 107: Can you elaborate on the thought behind this interaction? IE, if someone sets expanded, this would no longer be disabled? Is that pattern documented as part of the ARIA specification or somewhere?

packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -108,6 +122,7 @@ export class Accordion extends FASTElement {
         if (this.isSingleExpandMode()) {
             this.resetItems();
             event.target.expanded = true;
+            event.target.setAttribute("aria-disabled", "true");
```
chrisdholt, line 125: I'm not sure this is correct - I think we only want this disabled if it should be disabled. The control should be intractable by default. Disabled as an attribute would set aria-disabled and prevent user interaction.

chrisdholt (changes requested):
Thanks for the PR @yeraycaballero. I think there might be some confusion on the intent of `aria-disabled`. In the case a user sets disabled on the accordion (or item), we want to prevent interaction and ensure that aria-disabled is applied appropriately in the template. I don't think we want aria-disabled applied any time the control is in expanded mode.

packages/web-components/fast-foundation/src/accordion/accordion.ts
yeraycaballero: Hi @chrisdholt thanks for your reply and sorry for not getting back earlier.

Yes, it has been confusing. I'm going to try to explain my assumptions so we can clarify it a bit and you can point me in the right direction.

Initially and driven by the issue https://github.com/microsoft/fast/issues/4160 created by @eljefe223, I thought that the fast-accordion had implemented a behaviour mentioned briefly in the w3c spec that requires a panel to be expanded at all times and it can’t be closed. That matches exactly the issue, where the aria-disabled attribute was missing and that was what I tried to solve in the PR.

You can find more information about it in these links:
- https://w3c.github.io/aria-practices/#accordion
- https://w3c.github.io/aria-practices/examples/accordion/accordion.html

You can check that in the wc3 example, the selected item has got a inner button with an attribute aria-disabled=true and that attribute is missing in the fast accordion item

**Right now I think that the fact that the fast accordion items can't be closed wasn't intentional and it is a bug. It seems the the fast accordion only supports this configuration** 

1. **Multi**: allow multiple panels expanded ( AccordionExpandMode = "multi")
2. **Single**: allow one panel expanded at the time (AccordionExpandMode = "single")

**where all the items can be closed in both modes.**

If that is correct, there is a bug in the fast accordion on single mode because it doesn’t allow to close all the accordion-items. I would be pleasant to create a issue for it to send a pull request to fix it.

awentzel:
Hi @yeraycaballero, Thanks for your contribution, we'd love to get this work into master.  Do you have time to revisit and address the questions on this thread?   Thanks,  Aaron

EisenbergEffect:
@chrisdholt How would you like to handle this PR? It's been sitting since December 2020.

chrisdholt:
> @chrisdholt How would you like to handle this PR? It's been sitting since December 2020.

I think this can be closed for now, though I'd be happy to have @yeraycaballero contribute an overall fix! We should likely get some clarity on the issue as described. Thanks!

PR #4181, opened 2020-12-11
fix: tooltip position related styles not applied

scomea:
# Description
Tooltip css selectors were based on the user settings on attributes so weren't being applied when the attributes were absent.  Switched to use anchored-region classes instead.
 
## Motivation & context
noticed a bug

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):

nicholasrice:
This change reverts the work done in https://github.com/microsoft/fast/pull/4078, which I don't think is what we want. The goal of that PR was to move away from setting the class attribute on the host because clobbering of those classes happens easily when used with some UI frameworks.

I think instead that the relevant properties should always be reflected to attributes, ensuring that the attrs can always be used for styling.

scomea:
@nicholasrice In this case we want to set the styles based on the actual position of the tooltip which is most reliably retrieved by selecting based on the internal anchored region's classes.  The tooltip's "position" attribute doesn't work for this as currently implemented because if it is "undefined" the toolip dynamically places itself above/below the anchor ("position" remains undefined).  There currently is not an attribute on tooltip that advertises the actual position in the dynamic case,  probably should be, but even if we add that I'd still consider the internal anchored region the most stable indicator of how to style margins based on placement.

nicholasrice (approved):
Yep, you're right. I mistook this for requiring hoisting classes to the tooltip but that's not what is going on. Looks good.

Issue #4177, opened 2020-12-09
[Discussion] - handling tag dependencies in component templates


scomea:
When we define base component classes in 'fast-foundation' we typically include a default template which is then used in libraries that actually declare a component implementation.  For example we export a 'Badge' class and 'BadgeTemplate' from 'fast-foundation' which are used to declare the <fast-badge> custom element.

```js
import { Badge, BadgeTemplate as template } from "@microsoft/fast-foundation";
import { BadgeStyles as styles } from "./badge.styles";

/**
 * The FAST Badge Element. Implements {@link @microsoft/fast-foundation#Badge},
 * {@link @microsoft/fast-foundation#BadgeTemplate}
 *
 *
 * @public
 * @remarks
 * HTML Element: \<fast-badge\>
 */
@customElement({
    name: "fast-badge",
    template,
    styles,
})
export class FASTBadge extends Badge {}
```

This works well for simple components like Badge that only use basic html elements in their templates.  However when a component wants to include other custom elements within its own template things become problematic as the template will need to be customized every time it is used in a differently named library.  Examples of this include ['data-grid'](https://github.com/microsoft/fast/pull/4029) which needs to create child elements at runtime (ie. 'fast-data-grid' needs to create 'fast-data-grid-row' elements while a fluent version would want to create 'fluent-data-grid-row' elements), or a ['tooltip'](https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/tooltip/tooltip.template.ts) component which includes a 'fast-anchored-region' in its template when it is in the 'fast' library and presumably a 'fluent-anchored-region' in the fluent library.

The current path available to address this issue involves creating custom templates for these components rather than reuse the ones in 'fast-foundation' which is not ideal.

EisenbergEffect:
This is a tricky issue. I have thoughts on a long, medium, and short-term solutions.

## Long Term Solution

In the long-term, we'll have a new browser feature which will enable us to instantiate custom element registries and associate those with shadow dom. The registry will allow us to map tag names to element definitions, scoped to the shadow root. By combining this with our new provider system, we could have elements declare their dependencies in terms of the tag name used internal to the template, along with the base component class. Then, we can turn that into an element registry that maps the element name used in the template to the actual component class defined by the design system. If we're able to eliminate the need for class inheritance for different design systems, then this is trivially easy based on the proposed API. If we can't eliminate the inheritance, we can still achieve the same end result, just with a bit more code.

[...]

PR #4186, opened 2020-12-14
feat: add tooltip template creation

scomea:
# Description
This change removes the tooltip template with a "fast" dependency due to the internal "fast-anchored-region" and replaces it with a template creation function that returns a template using the desired prefix. 

## Motivation & context
Proposed implementation to address template/tag dependencies. [See discussion here.](https://github.com/microsoft/fast/issues/4177)

## Issue type checklist

This will break implementations with a dependency on the current tooltip template.  Unlikely this exists at this time, but possible.

## Process & policy checklist


chrisdholt (approved):
Ideally I'd like to create some more general guidance around when this pattern might be used as well as its existence. For now, I think this will be welcome to unblock folks from the issue as it currently stands.

EisenbergEffect (approved):

PR #4191, opened 2020-12-16
feat: add popover (aka flyout) component

SethDonohue:
# Description

Adds a Popover behavior component.

## Motivation & context

There is a use case for having a behavior that allows for content to be displayed in front of current content to provide additional content or actions and tied to some other element. Similar to a Dialog, but it won't be a modal, or force the user to make a decision on something, Popover will have engage the user and require the user to at least close the popover if they no longer want to view that content or provide the requested information, etc.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/popover/popover.ts
```diff
@@ -0,0 +1,634 @@
...
+
+export { PopoverPosition };
+
+/**
+ * An Popover Custom HTML Element.
+ *
+ * @public
+ */
+export class Popover extends FASTElement {
+    private static DirectionAttributeName: string = "dir";
```
scomea, line 20: I believe this is unused.

scomea:
I think we need some mechanism that updates the position of the popover in case layout changes.  Tooltips don't do this but they are transient enough that it's not such a big deal, popover is meant to contain active focusable ui and should be able to remain usable after resizing/scrolling so people will notice.  I think we should consider adding a timer that calls update() on the underlying anchored region a few times a second so that it automatically corrects layout after scrolling/resizing.  Probably an option to turn that on/off as well.   I have a working branch that does this for tooltip, including animating to the updated position.  https://github.com/microsoft/fast/tree/users/scomea/tooltip-track

SethDonohue:
> I think we need some mechanism that updates the position of the popover in case layout changes. Tooltips don't do this but they are transient enough that it's not such a big deal, popover is meant to contain active focusable ui and should be able to remain usable after resizing/scrolling so people will notice. I think we should consider adding a timer that calls update() on the underlying anchored region a few times a second so that it automatically corrects layout after scrolling/resizing. Probably an option to turn that on/off as well. I have a working branch that does this for tooltip, including animating to the updated position. https://github.com/microsoft/fast/tree/users/scomea/tooltip-track

Probably an excellent idea. I'll take a look at your branch when finishing this implementation. 
Do you think I should add something to the spec (https://github.com/microsoft/fast/pull/3691) before completing the spec, or add it after we get this component mostly sorted?

packages/web-components/fast-foundation/src/popover/popover.ts
SethDonohue: Likely, I haven't finished this just yet, but will take a look at this.

scomea:
> Do you think I should add something to the spec (#3691) before completing the spec, or add it after we get this component mostly sorted?

I think it's something we should address before we ship it, but purists would want to see the spec first.

scomea:
Another "spec" question - fixed vs. absolute positioning.  Fixed breaks out of containers but has latency on scrolling and updates needed to track the anchor. You currently have absolute positioning - is that the right choice, should developers be able to choose?

SethDonohue:
> Another "spec" question - fixed vs. absolute positioning. Fixed breaks out of containers but has latency on scrolling and updates needed to track the anchor. You currently have absolute positioning - is that the right choice, should developers be able to choose?

I'll have to look in to this further. If I recall correctly I couldn't get this to work without absolute positioning, but you just added a recent change that works with flex on anchored region, correct?

Issue #3857, opened 2020-09-08
Please add a fast-number-field component for data applications


MarcSkovMadsen:
I'm working on bringing Fast to the world of Data, Python and [Panel](https://panel.holoviz.org). See [Issue 3850](https://github.com/microsoft/fast/issues/3850).

Very often in Data you would like the user to type some numbers (integers or floats). For that Panel provides the [Spinner](https://panel.holoviz.org/reference/widgets/Spinner.html#widgets-gallery-spinner) (a bad name. Should have been `NumberInput` :-)).

I also see you would be needing it for the Component Explorer. For example for the [fast-progress](https://explore.fast.design/components/fast-progress).

![image](https://user-images.githubusercontent.com/42288570/92433368-20c28580-f19d-11ea-81b4-30a74d00a539.png)

But in Fast this is not supported. I see a Fast [number-field spec](https://github.com/microsoft/fast/blob/master/specs/number-field.md) though.

#### Solution

Design, Implement and Release a `fast-number-field` similar to the `fast-text-field`.

chrisdholt:
Hey @MarcSkovMadsen, would you mind taking a look at the current spec and providing some feedback? You can find it here: https://github.com/microsoft/fast/blob/master/specs/number-field.md#number-field

MarcSkovMadsen:
I have looked at the spec @chrisdholt . My feedback is

What I like about your components is that they are extensions of the basic ones. Like https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number. That is very important to me as then they can be used as drop in replacements. For example something web components do no support discrete steps.

![image](https://user-images.githubusercontent.com/42288570/92684146-0fac7c80-f335-11ea-8374-5a2e689b5ecb.png)

and I need discrete steps in my applications.

Furthermore I don't see (marked in red) the below in your specification

![image](https://user-images.githubusercontent.com/42288570/92684245-4edacd80-f335-11ea-8d63-4cca56ca06ed.png)

The link below does not lead to anything usefull
![image](https://user-images.githubusercontent.com/42288570/92684095-edb2fa00-f334-11ea-889e-df272dfeb7cb.png)

chrisdholt:
Thanks @MarcSkovMadsen for the feedback; would you want to contribute some of the below as improvements to the spec? Once we feel good about the spec, we - or anyone - can start work on the control!

MarcSkovMadsen:
Nope :-). I would like to help improve things. But this time i did not have the time.

[...]

PR #4204, opened 2020-12-18
feat: add fast-number-field component for data applications

robarbms:
# Description

Adds a fast-number-field component which takes numerical values that can be incremented and decremented in user defined stepped values.

## Motivation & context

Provides a text entry for numerical values requested by community.
https://github.com/microsoft/fast/issues/3857

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/number-field/number-field.definition.ts
```diff
@@ -0,0 +1,126 @@
...
+                },
+                {
+                    name: "step",
+                    description: "The step interval",
+                    type: DataType.string,
+                    default: false,
+                    required: false,
+                },
+            ],
+            slots: [
```
EisenbergEffect, line 118: There are start and end slots as well yes?

EisenbergEffect (approved):
In `sites/website/sidebars.js` we need to add this component to the list. Very excited to get this in!

packages/web-components/fast-foundation/src/number-field/number-field.template.ts
```diff
@@ -0,0 +1,75 @@
...
+                aria-label="${x => x.ariaLabel}"
+                aria-labelledby="${x => x.ariaLabelledby}"
+                aria-live="${x => x.ariaLive}"
+                aria-owns="${x => x.ariaOwns}"
+                aria-relevant="${x => x.ariaRelevant}"
+                aria-roledescription="${x => x.ariaRoledescription}"
+                ${ref("control")}
+            />
+            <div class="controls">
+                <div class="step_up" @click="${x => x.stepUp()}"></div>
```
SethDonohue, line 69: nit: I think we are trying to stick with kabob-case for multiple word classes that aren't things like `__hidden`.
This at least seems to be the pattern for *most* other multiple word classes on our components :P

SethDonohue (approved):

packages/web-components/fast-components/src/number-field/number-field.definition.ts
robarbms: Added start and end slots.

packages/web-components/fast-foundation/src/number-field/number-field.template.ts
robarbms: Good to know. I was following the pattern based on label__hidden. I've updated these to step-up and step-class.

packages/web-components/fast-foundation/src/number-field/number-field.template.ts
```diff
@@ -0,0 +1,75 @@
...
+                ?disabled="${x => x.disabled}"
+                list="${x => x.list}"
+                maxlength="${x => x.maxlength}"
+                minlength="${x => x.minlength}"
+                placeholder="${x => x.placeholder}"
+                ?readonly="${x => x.readOnly}"
+                ?required="${x => x.required}"
+                size="${x => x.size}"
+                :value="${x => x.value}"
+                type="text"
```
chrisdholt, line 43: @robarbms Are there any potential implications with having this be articulated as text semantically and via AT but having it behave strictly as a number?

robarbms: I used text type rather than number to avoid limitations of number. You can not use float values unless your steps are set to a float value(not 1.0, it needs to be a fractional like 0.1). I do number validation on the component side rather than using the in built validation of type number.

chrisdholt: Thanks! Given that shift, I wonder if we should add support for `inputmode` as an attribute here with only numeric types? Research on that here for input with type text for numbers similar to the choice we're making: https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -0,0 +1,246 @@
...
+        this.$emit("change");
+    }
+}
+
+/**
+ * Includes ARIA states and properties relating to the ARIA textbox role
+ *
+ * @public
+ */
+export class DelegatesARIANumberbox {}
```
chrisdholt, line 227: I don't think there is an actual ARIA Numberbox, so I wonder if this would simply use DelegatesARIATextbox as well.

robarbms: Updated this.

packages/web-components/fast-foundation/src/number-field/number-field.template.ts
robarbms: I've updated the inputmode. I left off pattern because I already had validation being handled on the valueChanged method using parseFloat. Their pattern seemed incomplete for data statistics and trying to maintain a pattern could be potentially problematic.

packages/web-components/fast-foundation/src/number-field/number-field.template.ts
SethDonohue: Yeah I am sure there are other opinions, which is why I say nit, no big deal.

chrisdholt (approved):

Issue #3908, opened 2020-09-16
fast-menu-item checkbox not working


MarcSkovMadsen:
In the specs the `fast-menu-item` checkbox attribute is described.

But it does not work as you can see when I click the `checked` checkbox in the component explorer

![fast-menu-item-checked](https://user-images.githubusercontent.com/42288570/93289348-75d64b00-f7de-11ea-85d4-672e49d35bba.gif)

You can see in the above video that the checked attribute actually appears. I could something changing without the video but it returned to the initial code so fast that I was not sure it was the checked attribute appearing.

The below example does not work either.

```html
<html>
    <head>
        <script type="module" src="https://unpkg.com/@microsoft/fast-components"></script>
    </head>
<body style="margin:0px">
    <fast-design-system-provider use-defaults style="min-height: 100%">
        <h2>Check Box</h2>
        <fast-menu>
            <fast-menu-item role="menuitemcheckbox">Check Box Not Checked</fast-menu-item>
            <fast-menu-item role="menuitemcheckbox" checked>Check Box Checked</fast-menu-item>
        </fast-menu>
    </fast-design-system-provider>
</body></html>
```

![image](https://user-images.githubusercontent.com/42288570/93289434-a0280880-f7de-11ea-8b07-c602d9967e07.png)

PR #4205, opened 2020-12-21
fix: add radio and checkbox visual in fast-menu-item

### LINKED ISSUE #3909: fast-menu-item radio attribute not working
The radio attribute of the fast-menu-item does not work as you can see in the below image from the component explorer.

I would have expected some indication that the third menu-item is checked.

image

khamudom:
# Description

Update the template in `menu-item`, adding `when` directives to look for the role of `menuitemcheckbox` or `menuitemradio`.
Added style selectors with the role of `menuitemcheckbox` or `menuitemradio` to layout the controls alongside the start and content slots.

without role. visual, nothing should change
image

with role
image

## Motivation & context

fixex #3908 
fixes #3909 

## Issue type checklist

## Process & policy checklist


eljefe223:
@khamudom My question here is shouldn't the radio items be in a radio group so only one can be selected? @chrisdholt for opinions

khamudom:
> @khamudom My question here is shouldn't the radio items be in a radio group so only one can be selected? @chrisdholt for opinions

@eljefe223 I agree a radio should be in a group. This change was to just add the visual of the radio indicator to show up in the item when you have a role of radio and it is checked. See issue #3909. If there are more changes I need to do than what I currently have, please let me know.

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
```diff
@@ -17,7 +17,49 @@ export const MenuItemTemplate = html<MenuItem>`
         class="${x => (x.disabled ? "disabled" : "")} ${x =>
             x.expanded ? "expanded" : ""}"
     >
-        ${startTemplate}
+        <div part="start-container" class="start-container">
```
chrisdholt, line 20: What happens when I want an icon in addition to this? I do not think this should replace start/end.

chrisdholt (changes requested):
+ @scomea who was looking at the *behaviors* which are enumerated by the W3C.

scomea:
I believe adding radio group functionality would would be follow on PR to add to the work done here.  Essentially a radio button that is not part of a radio group or is the group's only member essentially behaves like a checkbox.  What has not been specced is how we define radio groups in menus, I'm thinking a group="myRadioItems" attibute on the items, or whether we style groups in some way, but I don't think that would change the base checked/unchecked visuals.

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
khamudom: `${startTemplate}` is still there. I put it in the same div container as the radio and checkbox inputs.

scomea:
This branch adds the actual "radio-group" functionality, groups are broken up by adding role="separator" elements into the menu. Should be ready for pr shortly.

scomea:
Radio group selection behavior pr is here - https://github.com/microsoft/fast/pull/4208

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
khamudom: @chrisdholt even though I still have `${startTemplate}` alongside the radio and checkbox inputs, is this still not the right approach, or was there another concern with this change?

khamudom:
> Radio group selection behavior pr is here - #4208

This PR is now merged with @scomea changes. You can now see only one radio input is selected at a time when smoking this PR.

SethDonohue (commented):
Looks like the _checked_  and _selected_ HC colors are off? Hover looks fine.
Screen Shot 2021-01-12 at 10 10 53 AM

khamudom:
> Looks like the _checked_ and _selected_ HC colors are off? Hover looks fine.
> Screen Shot 2021-01-12 at 10 10 53 AM

Yes, this issue is already happening in master. But I will go ahead and apply a fix to this PR.

khamudom:
> Yes, this issue is already happening in master. But I will go ahead and apply a fix to this PR.

High contrast issue is fixed and pushed.
image

SethDonohue (approved):

packages/web-components/fast-foundation/src/menu/menu.ts
```diff
@@ -139,6 +142,10 @@ export class Menu extends FASTElement {
             if (item === focusIndex) {
                 this.menuItems[item].setAttribute("tabindex", "0");
             }
+            if (hasRole) {
+                this.menuItems[item].classList.remove("without-role");
```
scomea, line 146: Can you just have a single class that you toggle?  Make one state the default and add the class otherwise.

chrisdholt: I actually think that these visuals may be different in any given design system. I think we need explicit classes for checkbox or radio visuals.

chrisdholt (changes requested):

packages/web-components/fast-foundation/src/menu/menu.ts
```diff
@@ -129,6 +129,9 @@ export class Menu extends FASTElement {
 
     private setItems = (): void => {
         const focusIndex = this.menuItems.findIndex(this.isFocusableElement);
+        const hasRole =
```
chrisdholt, line 132: I'd prefer a different const name here - technically all of these could have explicit roles, including a valid role of simply `menuitem`.

packages/web-components/fast-foundation/src/menu/menu.ts
scomea: These classes control spacing in all menu items based on whether there are any radio/check items in the entire list, so it should be consistent for all menu item types.

chrisdholt: I don't necessarily agree with this - I could be swayed, but this is confusing to me. What happens if I want to style them differently? How do I accomplish that? I could apply standard spacing to both by chaining classes together in CSS. If I for some reason need them to have different styles, how would I accomplish that? 

Technically, if this just controls spacing for checkbox and radio items, do we need the class selector? It seems like we could control that in CSS using attribute selectors since those attributes will *always* be in the DOM. Perhaps this isn't actually necessary?

scomea: The purpose of the mechanism is to allow standard menu items to add the space when they share a menu with radio/checkbox items so alignement can be maintained without adding the space when there are not radio/checkbox items in the list. So somehow the menu scans it children and sets some flag that its children can use for styling.  Maybe a class isn't the best way to do this.

chrisdholt: My concern is that we're adding foundational behaviors to manage a specific visual style that may not be a requirement. If we need a hook to style radio/checkbox implementations, I'd prefer to use those names because it is more semantic and intuitive if I'm trying to target *just* a radio or checkbox, or even both. Alternatively, do we need the hook? Could we simply do something like below in CSS? 

```css
::slotted(fast-menu-item) {
    // add space
}

// can we increase specificity here?
::slotted([role="menuitemcheckbox"]),
::slotted([role="menuitemradio"]) {
    // don't add space
}
```

awentzel (approved):

EisenbergEffect:
Pinging this one since it's starting to get stale.

khamudom:
> Pinging this one since it's starting to get stale.

I will be focusing on this today.

packages/web-components/fast-foundation/src/menu/menu.ts
khamudom: this is removed

packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
```diff
@@ -108,25 +113,151 @@ export const MenuItemStyles = css`
...
+
+    :host(:hover) .radio {
+        border-color: ${accentForegroundCutRestBehavior.var};
+    }
+
+    :host([checked="true"]) .radio {
+        display: none;
+    }
+
+    :host([checked="true"]) .radio-checked {
```
chrisdholt, line 212: @khamudom what's the `.radio-checked` class here for?

packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
```diff
@@ -108,25 +113,151 @@ export const MenuItemStyles = css`
...
+        };
+        outline: none;
+        margin-inline-start: 10px;
+    }
+
+    :host(:hover) .radio {
+        border-color: ${accentForegroundCutRestBehavior.var};
+    }
+
+    :host([checked="true"]) .radio {
```
chrisdholt, line 208: Can we bind this to the `aria-checked` attribute rather than `checked`? IMO, we likely shouldn't be reflecting `checked` - it should likely only be `fromView` (no need to change in this PR)

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
chrisdholt, line 20: @khamudom what are the possible implications for this being within the start container here? I'm wondering if this should be separate from start in the template.

khamudom: this is now updated refactored. the inputs and startTemplate are separated and not in one container

packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
khamudom: updated to use aria-checked

packages/web-components/fast-foundation/src/menu-item/menu-item.template.ts
```diff
@@ -17,6 +17,36 @@ export const MenuItemTemplate = html<MenuItem>`
...
+                            />
+                        </svg>
+                    </span>
+                `
+            )}
+            ${when(
+                x => x.role === MenuItemRole.menuitemradio,
+                html<MenuItem>`
+                    <span part="radio" class="radio">
+                        <div part="radio-indicator" class="radio-indicator"></div>
```
chrisdholt, line 45: @khamudom one more nit on this - can you change this to a span? Technically span elements are permitted to only include phrasing content: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content

The only other question I have here is whether these should be slots w/ default slotted content so people can pass their own svg's into this. Also curious as to why we're using an SVG for the checkbox above but not here.

khamudom: I agree they should be a slot with default indicators.
I was mimicking our checkbox and radio component, just to make sure these inputs in the menu-item look similar to the stand-alone version. The radio component is using just a `div` element (which I will change to a `span` here) But, I don't see why it couldn't be an SVG. I will investigate find a "dot" SVG.

chrisdholt: Gotcha - can we wrap these in a slot and treat them as default slotted content for now? I think we can keep the current radio implementation, that's fine; let's just change the element to a span.

khamudom: Added default slots for the indicators and refactor/combined some selectors in the styles file. Also added some examples to show custom indicators.

image

chrisdholt (approved):

PR #4210, opened 2020-12-28
feat: add getPosition service for anchored region

scomea:
# Description
Anchored region makes use of an `intersectionObserver` instance to calculate positioning.  Currently each anchored region instance has it's own intersection observer instance.  This change moves to using a single intersection observer instance for all anchored-region instances on a page.

## Motivation & context
This is a performance improvement by removing multiple anchored region instanciations.

This is the performance graph from rendering the anchored region sample page before this change:

image

After:

image

A modest but detectable improvement.  This should be more significant in scenarios with frequent updates of multiple components.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/anchored-region/anchored-region.spec.ts
```diff
@@ -11,17 +11,21 @@ class FASTAnchoredRegion extends AnchoredRegion {}
...
-        </div>
-    `);
+                <div id="viewport" style="width: 1000px; height: 1000px;">
+                    <button id="anchor" style="width: 100px; height: 100px;">anchor</button>
+                    <fast-anchored-region
+                        viewport="viewport"
+                        anchor="anchor"
+                        id="region"
+                    >
+                        <div id="contents" style="width: 100px; height: 100px;">
```
janechu, line 21:
```suggestion
                        <div id="contents" style="width: 100px; height: 100px;"></div>
```

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -367,6 +367,8 @@ export class AnchoredRegion extends FASTElement {
     private pendingReset: boolean = false;
     private currentDirection: Direction = Direction.ltr;
 
+    private static intersectionService: IntersectionService = new IntersectionService();
```
janechu, line 370: Why not remove the association with anchored region and instead export from the `./intersection-service` a new instance that any component can import? That way this can be re-usable for other components.

scomea: I've considered that but figured we'd get it in as a private thing before we add it to our public api.  If folks think we should make it public from the start I can make that change.

packages/web-components/fast-foundation/src/anchored-region/intersection-service.ts
```diff
@@ -0,0 +1,100 @@
+/**
+ *
+ *
+ * @public
+ */
+export class IntersectionService {
+    private intersectionDetector: IntersectionObserver;
+
+    private observedElements: Element[] = [];
```
nicholasrice, line 9: You might consider using a [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of two arrays to manage element and observer callbacks. You can use the Element as a map key and the observer callback array as the value

scomea: Yes, that is better, thx.

janechu (approved):

nicholasrice (approved):

PR #4224, opened 2021-01-05
fix: slider disabled and readonly cursor styles

### LINKED ISSUE #4222: Slider shows cursor pointer when disabled or readonly
Cursor for _disabled_ and _readonly_ Slider shows as pointer.
Screen Shot 2021-01-05 at 11 35 39 AM

Go the components explorer and select Slider, then chose disabled or readonly attribute, and then hover over Slider.

Cursor should show as not-allowed cursor:
Screen Shot 2021-01-05 at 11 34 24 AM

* OS & Device: MacOS, PC
* Browser Microsoft Edge, Google Chrome

SethDonohue:
# Description
Closes #4222 

Note: There is no `.slider` class on this component's template, so the css `.slider ` selector was not targeting anything. The disabled cursor still works on the labels and the forcedColors cursor style can be removed with this change as well.

Before:
Screen Shot 2021-01-05 at 11 35 39 AM

After:
Screen Shot 2021-01-05 at 11 34 24 AM

High Contrast:
Screen Shot 2021-01-05 at 1 31 05 PM

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/slider/fixtures/base.html
+++ b/packages/web-components/fast-components/src/slider/fixtures/base.html
@@ -301,5 +301,42 @@
                 100
             </fast-slider-label>
         </fast-slider>
+        <!-- Read Only -->
+        <h4>Read Only</h4>
+        <fast-slider min="0" max="100" step="5" readonly>
+            <fast-slider-label position="0">
+                0
+            </fast-slider-label>
+            <fast-slider-label position="10">
+                10
+            </fast-slider-label>
+            <fast-slider-label position="20">
+                20
+            </fast-slider-label>
+            <fast-slider-label position="30">
+                30
+            </fast-slider-label>
+            <fast-slider-label position="40">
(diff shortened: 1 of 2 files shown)


EisenbergEffect (commented):
We recently did some changes in other components to move away from being based on host classes and instead leverage attribute selectors, as that's less invasive and less likely to conflict with frameworks that may be interacting with our components. Would it be possible to take that same approach in this case as well?

SethDonohue:
> We recently did some changes in other components to move away from being based on host classes and instead leverage attribute selectors, as that's less invasive and less likely to conflict with frameworks that may be interacting with our components. Would it be possible to take that same approach in this case as well?

Updated to use attributes instead of classes. Will do so on anything in the future!

marjonlynch (approved):
smoked, looks good

janechu (approved):

PR #4230, opened 2021-01-06
fix: added new grid for horizontal tabs with start and end elements

marjonlynch:
When utilizing the start and end slots in the tabs control, I expect that they will line up on either side of the tabs and not have the end pushed out to the width of the tabs control.

# Description

Adding another grid to encompass the horizontal tabs and a start and end slot if utilized. The start and end slot columns are auto sized so if they are not used, that space will be filled with the tabs only. This change requires no logic change in the tabs class and does not affect tabs without start/end slots.

Screen Shot 2021-01-06 at 3 31 14 PM

situation before change:
Screen Shot 2021-01-06 at 3 29 05 PM

## Motivation & context

The motivation came through needing to use the end slot for a button that should be just on the other side of the last tab.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/tabs/tabs.styles.ts
+++ b/packages/web-components/fast-components/src/tabs/tabs.styles.ts
@@ -40,14 +40,6 @@ export const TabsStyles = css`
         align-self: center;
     }
 
-    :host([orientation="horizontal"]) .start {
-        grid-column: 1;
-    }
-
-    :host([orientation="horizontal"]) .end {
-        grid-column: 3;
-    }
-
     .activeIndicator {
         grid-row: 2;
         grid-column: 1;
--- a/packages/web-components/fast-components/src/tabs/tabs.styles.ts
+++ b/packages/web-components/fast-components/src/tabs/tabs.styles.ts
@@ -29,9 +29,23 @@ export const TabsStyles = css`
         box-sizing: border-box;
     }
 
+    .tablist-container {
+        display: grid;
+        grid-template-columns: auto auto auto;
+        grid-template-rows: auto;
+    }
+
     .start,
     .end {
         align-self: center;
+    }
+
+    :host([orientation="horizontal"]) .start {
+        grid-column: 1;
+    }
+
+    :host([orientation="horizontal"]) .end {
+        grid-column: 3;
     }
 
     .activeIndicator {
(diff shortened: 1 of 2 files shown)


khamudom (approved):

SethDonohue (approved):

eljefe223:
Is this the consensus? I don't necessarily agree that the end content should butt up against the last tab. Tabs are used as layout components and could see options, sub actions and info being lined to the the right edge of the component. For instance look at the tabs across the top of this github page.

image

image

image

marjonlynch:
I think you could style it to look that way but having it the way it was before with the start and end templates wrapping both the tabs and tab panels creates a sort of coupling between the two pieces that I don't think should exist. Maybe this PR needs a bit more to allow the dev to customize the tab list container, like the grid layout. The github example is maybe just one way that devs would want tabs to work and I'm not sure that's how the majority of devs would want it to be.

chrisdholt:
> I think you could style it to look that way but having it the way it was before with the start and end templates wrapping both the tabs and tab panels creates a sort of coupling between the two pieces that I don't think should exist. Maybe this PR needs a bit more to allow the dev to customize the tab list container, like the grid layout. The github example is maybe just one way that devs would want tabs to work and I'm not sure that's how the majority of devs would want it to be.

I think it would be good to get a better understanding of the possible variations enumerated by @eljefe223 and what would be proposed based on your feedback here @marjonlynch. What are the benefits for having an API for the layout vs a dev leveraging CSS.

marjonlynch:
Also, I don't know if this is the consensus and wouldn't do anything no matter the approvals until Chris gave some feedback

marjonlynch:
yes, I would like to hear about a variation where start or end templates are needing to wrap the tab panel portion of this component

EisenbergEffect:
Should we set up a meeting to discuss this more?

chrisdholt:
> Should we set up a meeting to discuss this more?

I think that would be helpful. Similarly, I know @marjonlynch had questions about overflow which we could discuss at that time as well. I had a beneficial conversation w/ another person working in the Open-UI space yesterday on this and it was helpful to get their opinions as well.

nicholasrice (approved):

awentzel (approved):

Issue #4237, opened 2021-01-11
Create a behavior for attaching component styles based on an appearance


chrisdholt:
Many of our components have an "appearance" attribute to define a given visual treatment (anchor, button, etc). Currently for a given component all styles are attached regardless of the component appearance. So if I have an input with a "filled" appearance, the "outline" appearance is also being created and attached as styles. Ideally, we should only include those styles which are being leveraged at a given time.

Ideally we can use something fairly generic like this:

```js
import { css } from "@microsoft/fast-element";
import { AppearanceBehavior } from "@microsoft/fast-foundation";

const buttonAccentBehavior = AppearanceBehavior.with("accent");

const styles = css`
  :host {
    background: grey;
    color: black;
  }
`
.withBehaviors(buttonAccentBehavior(css`
  :host {
    background: blue;
    color: white;
  }
`))
```

EisenbergEffect:
To make sure I follow...is the idea that the behavior would observe the `appearance` attribute of the host element, and when its value equals "accent" it will add the specified stylesheet, otherwise it will not add or remove the styles?

chrisdholt:
> To make sure I follow...is the idea that the behavior would observe the `appearance` attribute of the host element, and when its value equals "accent" it will add the specified stylesheet, otherwise it will not add or remove the styles?

Essentially, yes - consider the behavior we have for something like forced colors or match media but scoped to a given attribute value rather than the state of the OS or window.

EisenbergEffect:
Cool. We can definitely do this. We can probably create a general mechanism for this so that we could even separate out the various attribute values, allowing for them to be tree-shaken.

nicholasrice:
Small clarification since the question came up when syncing w/ this with Jeff, I think we would want to map this to observable *properties* and not content attributes, is that correct?

EisenbergEffect:
I think so. If it's based on observable properties, we can use it with any property that uses our system and not just html elements. If I'm following you...

eljefe223:
Thinking something like this:

```
export function appearanceBehavior(value: string, styleSheet: ElementStyles) {
    return new PropertyStyleSheetBehavior("appearance", value, styleSheet)  
}
```

```
.withBehaviors(
  appearanceBehavior("stealth", css`...`),
  appearanceBehavior("outline", css`...`),
  etc.
)
```

Any thoughts or objections?

EisenbergEffect:
[...]

PR #4238, opened 2021-01-11
feat: Create a behavior for attaching component styles based on an appearance

eljefe223:
# Description
Closes: #4237 

Adds a behavior that allows to attach styles based on any property. Adds a behavior that attaches styles based on appearance.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
+++ b/packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
@@ -39,8 +39,7 @@ export class PropertyStyleSheetBehavior implements Behavior {
      * @internal
      */
     public unbind(source: typeof FASTElement & HTMLElement) {
-        this.unbind(source);
-
+        Observable.getNotifier(source).unsubscribe(this, this.propertyName);
         (source as any).$fastController.removeStyles(this.styles);
     }
 
--- a/packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
+++ b/packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
@@ -31,7 +31,7 @@ export class PropertyStyleSheetBehavior implements Behavior {
      */
     public bind(elementInstance: FASTElement) {
         Observable.getNotifier(elementInstance).subscribe(this, this.propertyName);
-        this.handleChange;
+        this.handleChange(elementInstance, this.propertyName);
     }
 
     /**
--- a/packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
+++ b/packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
@@ -58,7 +58,7 @@ export class PropertyStyleSheetBehavior implements Behavior {
 }
 
 /**
- * Behavior that will condiitionally apply a stylesheet based on the elements
+ * Behavior that will conditionally apply a stylesheet based on the elements
  * appearance property
  *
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
```diff
@@ -0,0 +1,72 @@
...
+        Observable.getNotifier(elementInstance).subscribe(this, this.propertyName);
+    }
+
+    /**
+     * Unbinds the behavior from the element.
+     * @param source - The element for which the behavior is unbinding.
+     * @internal
+     */
+    public unbind(source: typeof FASTElement & HTMLElement) {
+        this.unbind(source);
```
EisenbergEffect, line 42: I think you want to do `Observable.getNotifier(source).unsubscribe(this, this.propertyName);`

EisenbergEffect (changes requested):

packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
```diff
@@ -0,0 +1,71 @@
...
+        private value: any,
+        private styles: ElementStyles
+    ) {}
+
+    /**
+     * Binds the behavior to the element.
+     * @param elementInstance - The element for which the property is applied.
+     */
+    public bind(elementInstance: FASTElement) {
+        Observable.getNotifier(elementInstance).subscribe(this, this.propertyName);
```
EisenbergEffect, line 33: We should probably also check the value at this point and add the styles if appropriate, otherwise we're only adding the styles if the value changes from its initial value.

EisenbergEffect: The easiest way to do this is to just call `handleChange` directly.

EisenbergEffect (changes requested):

packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
nicholasrice: @eljefe223 you need to call the function w/ the appropriate arguments:

```
this.handleChange(elementInstance, this.propertyName);
```

packages/web-components/fast-foundation/src/utilities/property-stylesheet-behavior.ts
```diff
@@ -0,0 +1,72 @@
...
+    public handleChange(source: FASTElement, key) {
+        if (source[key] === this.value) {
+            source.$fastController.addStyles(this.styles);
+        } else {
+            source.$fastController.removeStyles(this.styles);
+        }
+    }
+}
+
+/**
```
nicholasrice, line 61: I'm wondering if the is the appropriate place to export `appearanceBehavior` from. Nothing in fast-foundation has an `appearance` property so I think I would expect this to be exported from fast-components

eljefe223: I think that makes sense, @chrisdholt thoughts?

chrisdholt: I'd agree for "appearance" and would agree we need to do a one off in Fluent for this. I think that's okay. If Fluent decided to move away from "appearance" for some other semantic, we'd want to be sure to update this accordingly rather than inheriting those semantics.

EisenbergEffect (approved):

nicholasrice (approved):

PR #4245, opened 2021-01-14
fix: add whitespace filter to text-field template to correctly hide label div with start and end content

### LINKED ISSUE #4219: text-field's label div is not properly hidden when start or end slots have content
In Text Field when there is content in the start or end slots the label div is _not_ hidden causing Text Field to have a larger height than it should.

With Start:
Screen Shot 2021-01-04 at 2 16 30 PM

With End:
Screen Shot 2021-01-04 at 2 16 48 PM

View with any Text Field that has content slotted in the `start` or `end` slots.

Should be similar to _default_ Text Field without a label where the height does not extend above the border.
Screen Shot 2021-01-04 at 2 15 55 PM

* OS & Device: MacOS on PC
* Browser Microsoft Edge, Google Chrome

SethDonohue:
# Description

Fixes #4219 

Added a white space filter tha filters out `text` nodes that have _only_ white space in them. Does so with simple nodeType, regex and length check.

Note: this helps make sure this component follows it's height property as expected when nothing is slotted in the label.

Before:
Screen Shot 2021-01-14 at 10 42 31 AM

After:
Screen Shot 2021-01-14 at 10 39 55 AM

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/text-field/text-field.template.ts
```diff
@@ -2,6 +2,18 @@ import { html, ref, slotted } from "@microsoft/fast-element";
 import { endTemplate, startTemplate } from "../patterns";
 import { TextField } from "./text-field";
 
+/***
+ * A simple whitespace filter for the template
+ * @public
+ */
+export function whitespaceFilter(value, index, array) {
+    console.log(value, value.nodeValue);
```
EisenbergEffect, line 10: Please remove the `console.log` 😄

EisenbergEffect (changes requested):

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
SethDonohue: ahah oops, luckily it was still a draft :P, removed.

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
```diff
@@ -2,6 +2,17 @@ import { html, ref, slotted } from "@microsoft/fast-element";
...
+/***
+ * A simple whitespace filter for the template
+ * @public
+ */
+export function whitespaceFilter(value, index, array) {
+    return value.nodeType === 3 &&
+        typeof value.nodeValue === "string" &&
+        !value.nodeValue.replace(/\s/g, "").length
+        ? false
+        : true;
```
radium-v, line 14:
```suggestion
    return value.nodeType === Node.TEXT_NODE &&
        typeof value.nodeValue === "string" &&
        !value.trim().length
```

radium-v (changes requested):

radium-v:
Can we change the wording to refer to it as blank space?

packages/web-components/fast-foundation/src/text-field/text-field.spec.ts
```diff
@@ -199,6 +199,66 @@ describe("TextField", () => {
 
         await disconnect();
     });
+    it("should hide the label when start and/or end content is provided", async () => {
+        const { element, connect, disconnect } = await setup();
+        const content: HTMLImageElement = document.createElement("img");
+        content.src = "https://placehold.it/16";
```
radium-v, line 205: Could this cause timeout errors when running in an environment without a network connection?

packages/web-components/fast-foundation/src/text-field/text-field.spec.ts
```diff
@@ -199,6 +199,66 @@ describe("TextField", () => {
...
+        element.appendChild(content);
+
+        expect(
+            element.shadowRoot
+                ?.querySelector("label")
+                ?.classList.contains("label__hidden")
+        ).to.be.true;
+
+        await disconnect();
+
```
radium-v, line 220: The start and end test should be split into two `it()` calls. Repetition in tests is fine, since we want to make sure all tests are isolated from each other.

packages/web-components/fast-foundation/src/text-field/text-field.spec.ts
```diff
@@ -199,6 +199,66 @@ describe("TextField", () => {
...
```
radium-v, line 232: This should be another split. Repetition in tests is fine, since we want to make sure all tests are isolated from each other.

radium-v (changes requested):

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
SethDonohue: I like the more clear `node.text_node`, add to a commit now.

packages/web-components/fast-foundation/src/text-field/text-field.spec.ts
SethDonohue: probably, It really can be anything so I'll change it.

packages/web-components/fast-foundation/src/text-field/text-field.spec.ts
SethDonohue: split up!

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
```diff
@@ -2,6 +2,17 @@ import { html, ref, slotted } from "@microsoft/fast-element";
 import { endTemplate, startTemplate } from "../patterns";
 import { TextField } from "./text-field";
 
+/***
+ * A simple whitespace filter for the template
+ * @public
+ */
```
nicholasrice, line 8: Can you add types to the function arguments?

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
```diff
@@ -2,6 +2,17 @@ import { html, ref, slotted } from "@microsoft/fast-element";
...
+export function whitespaceFilter(value, index, array) {
+    return (
+        value.nodeType === Node.TEXT_NODE &&
+        typeof value.nodeValue === "string" &&
```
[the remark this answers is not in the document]
SethDonohue: whoops, I think It should actually be `!value.nodeValue.trim().length`. We may have gotten lucky with that trim working in tests.

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
SethDonohue: Updated

SethDonohue:
@awentzel @Ibrahimmaga Any idea why the Linux validation is failing? Tests are working locally and I don't see how the Disclosure tests would fail based on this PR. Also seeing a timeout on the macos build, which I've seen randomly on other PRs

SethDonohue:
> Can we change the wording to refer to it as blank space?

Any other thoughts on this?

It does technically remove all whitespace before and after the value, so I feel like that is more accurate.
Blank space may not clearly tell others using this function what it filters out.

nicholasrice:
> > Can we change the wording to refer to it as blank space?
> 
> Any other thoughts on this?
> 
> It does technically remove all whitespace before and after the value, so I feel like that is more accurate.
> Blank space may not clearly tell others using this function what it filters out.
 Yea in this case  I think "whitespace" is more accurate; this will trim out new lines in addition to blank spaces.

packages/web-components/fast-foundation/src/text-field/text-field.template.ts
```diff
@@ -2,6 +2,22 @@ import { html, ref, slotted } from "@microsoft/fast-element";
...
+/***
+ * A simple whitespace filter for the template
+ *
+ * @param value - The Node that is being inspected
+ * @param index - The index of the node within the array
+ * @param array - The Node array that is being filtered
+ *
+ * @public
+ */
+export function whitespaceFilter(value: Node, index: number, array: Node[]) {
```
nicholasrice, line 14: Kind of nitty but I don't think this should be exported from `text-field.template.ts` because there is nothing intrinsic to text-field about this filter. It could likely be exported from `fast-element` but it would be good to get @EisenbergEffect's opinion on that.

SethDonohue: I almost added it to the utilities section as I could see this getting used on `text-area` as well as it has a similar issue when whitespace is slotted, we just haven't noticed it yet because it does not have start and end slots.

Think i should move it to /utilities for now?

SethDonohue: I moved it to utils for now, as it should be able to be used elsewhere.

radium-v (approved):

eljefe223 (approved):

EisenbergEffect (approved):

PR #4256, opened 2021-01-21
fix: tooltips are incorrectly positioned when parent is a flex container

scomea:
# Description
Anchored region was not calculating layout correctly is some situations because it did not identify the correct "containing block" element in all circumstances, for example when the parent of the tooltip was a flex container.  This change takes advantage of the existing initial frame when a tooltip is rendered hidden to calculate relative positions to also calculate the dimensions of the tooltip's containing block (the changes are actually in anchored region) by forcing the hidden tooltip to span it and looking at its dimensions.  

## Motivation & context
User reports of incorrect tooltip placement.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
```diff
@@ -11,9 +11,28 @@ FASTDesignSystemProvider;
...
 
+function onShowClick(e: MouseEvent): void {
+    toggleTooltipVisibility("tooltip-show-1");
+    toggleTooltipVisibility("tooltip-show-2");
+    toggleTooltipVisibility("tooltip-show-3");
+    toggleTooltipVisibility("tooltip-show-4");
+}
+
+function toggleTooltipVisibility(tooltipId: string): void {
+    const tooltipInstance: HTMLElement | null = document.getElementById(tooltipId);
```
m4thieulavoie, line 29: [Curious] Could've we use `querySelector` here to clean up the type castings?
```suggestion
    const tooltipInstance: FASTTooltip | null = document.querySelector(`#${tooltipId}`);
```
and a bit down
```ts
tooltipInstance.visible = !tooltipInstance.visible;
```

radium-v: I think `getElementById` is fine, but the types can be asserted instead of declared:

```suggestion
    const tooltipInstance = document.getElementById(tooltipId) as FASTTooltip | null;
```

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -621,17 +624,20 @@ export class AnchoredRegion extends FASTElement {
         this.pendingPositioningUpdate = false;
 
         let regionRect: DOMRect | ClientRect | null = null;
+        regionRect = this.applyIntersectionEntries(entries);
```
nicholasrice, line 627: what's the purpose of assigning this to `null` and then to the product of `this.applyIntersectionEntries(entries)`?

scomea: cleaned up

packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
scomea: updated

scomea:
Last update also removed use of the offset parent to identify the containing block's dimensions as the new method should always work.

EisenbergEffect (approved):

eljefe223 (approved):

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -1,5 +1,6 @@
 import { attr, DOM, FASTElement, observable } from "@microsoft/fast-element";
 import { Direction } from "@microsoft/fast-web-utilities";
+import { unset } from "lodash-es";
```
m4thieulavoie, line 3: This import seems unused

scomea: removed

PR #4258, opened 2021-01-21
feat: CI - Codeql-analysis.yml

Ibrahimmaga:
# Description
This will scan all code security and vulnerability in our repository
Scan codeql too analysis 

This is a Codeql-Analysis tool for security and vulnerability scanning
it will support languages like JavaScript, typescript, and csharp just in case

## Motivation & context
We want all code to be scan and receive alert for any vulnerability so we don't
run into problem in production

 Security and Vulnerability

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):

awentzel (changes requested):
Please add this to the Action for Fast Validation.  It does not have to occur on every push.  Weekly is enough.  Also, we used to have this so I'm surprised it was missing.  Perhaps, look at all actions and see what we had before since this was working.

Other comments inline

.github/workflows/CI - Codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master, explore-site, releases/* ]
```
awentzel, line 16: Please migrate the minimum requirements into the existing FAST Validation GH Action as a step after code coverage that could be dependant on each platform build succeeding.  This way we use a single Action to validate compliance and stability-related issues once per week.  This keeps build time down for development efforts and keeps all of our actions relatively clean and well organized by function.  We also have very specific name patterns to improve readability when investigating actions.

You can keep strategy and steps without the comments and that should make it a lot less code.

.github/workflows/CI - Codeql-analysis.yml
```diff
@@ -0,0 +1,67 @@
...
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript', 'typescript' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
```
awentzel, line 32: Jane has some code that I believe is C++ or C.  We may want to include that in the scan.  Jane can provide more details on what package that code lives in.

.github/workflows/ci-nightly.yml
```diff
@@ -26,7 +26,8 @@ jobs:
       run: |
         git config user.name ${{ env.GITHUB_PUBLISH_USER }}
         git config user.email ${{ env.GITHUB_PUBLISH_EMAIL }}
-        git remote set-url origin https://${{GITHUB_SERVICE_PAT}}:${{env.GITHUB_SERVICE_PAT}}@github.com/microsoft/fast.git
```
awentzel, line 29: Can you explain this change and how it relates to Code Security Scanning or why the change is needed?

Ibrahimmaga: This has nothing to do with Code security scanning. This workflow is for deployment. It is failing on env secrets value.  Since now continuous deployment is not in place, it's not a big concern. We will fix this next when all the testing work is finished.

Ibrahimmaga:
I will close this PR for now. It takes so long to execute. I will reopened later if necessary.

Ibrahimmaga:
will be reopened latter if necessary.

Issue #3317, opened 2020-06-16
Remove api-extractor exception from @microsoft/fast-foundation


nicholasrice:
#3316 implements a change to `packages/web-components/fast-foundation/api-extractor.json` that instructs the `ae-different-release-tags` rule to not fail in CI. This is due to [an issue in api-documenter](https://github.com/microsoft/rushstack/issues/1921) not being able to differentiate exports of the same name. Several components implement a mixin pattern where naming an interface the same as a class is required.

When the api-documenter issue is fixed we can revert the change to our `api-extractor.json` and update the release-tag reference for all relevant exports.

Ibrahimmaga:
api-extractor.json was causing problem in windows environment when I was messing around with prettier:diff to solve windows os validation. This doesn't seems a problem any longer since windows validation is working. But if someone get any exception error, let me know then we will remove the exception.

Ibrahimmaga:
closes #4281

PR #4281, opened 2021-01-27
fix: remove api-extract exception

Ibrahimmaga:
# Description

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,7 +57,7 @@ sites/website/i18n/
 
 # Docusaurus package readme's
 docs/en/packages/*/README.md
-
+**/.md
 # TypeDoc generated subsites
 docs/en/packages/*/api/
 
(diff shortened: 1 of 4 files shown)


EisenbergEffect (commented):
I can't recall what the effect of this was. Maybe @radium-v remembers. Let's double check before we remove this.

Ibrahimmaga (commented):
Removing it is not even solving the problem. I might needs to reverse the changes. John is also looking into the issue with api-extractor.

radium-v:
I don't think this is actionable yet. API Extractor version 7.8.2 [introduces a change that reports errors on `import()` statements in the generated declaration files](https://github.com/microsoft/rushstack/blob/master/apps/api-extractor/CHANGELOG.md#782), which would need to be resolved before upgrading to a newer version of the extractor.

There's also no indication that the issue tracked by #3317 is fixed on the extractor side, so the warnings still cause failures when running `api-extractor` without the `--local` flag.

Ibrahimmaga:
This PR is not solving the issue of api extractor

PR #4288, opened 2021-01-29
fix: increase sophistication of behavior binding and unbinding

### LINKED ISSUE #4242: Using a behavior that dynamically loads styles and uses other behaviors seems to not properly append behaviors
We created a behavior that dynamically loads in stylesheets based on a property. We are using this to dynamically load stylesheets based on the `appearance` property for `fast-anchor` and `fast-button`.  The behavior loads in the various styles but not the `behaviors` for the various appearances. Meaning the colors are not loading for the various anchor and button states.

nicholasrice:
# Description
This change implements more sophisticated behavior tracking to make the `Controller` aware of multiple add and remove operations of the same behavior instance. It configures the `Controller` to only bind a behavior instance the first time it is added, and to only unbind a behavior instance after it is removed the number of times it was added. This addresses issues w/ inadvertently unbinding a behavior that was added by multiple sources.

This change also addresses an issue where a `Behavior` which depends on (and adds) *another* `Behavior` would not see it's dependent behavior bound when the initial `Behavior` is bound during the `Controller`'s `onConnectedCallback`.

fixes #4242 

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-element/src/controller.ts
```diff
@@ -258,10 +272,13 @@ export class Controller extends PropertyChangeNotifier {
         }
 
         const behaviors = this.behaviors;
-
         if (behaviors !== null) {
-            for (let i = 0, ii = behaviors.length; i < ii; ++i) {
-                behaviors[i].bind(element, defaultExecutionContext);
+            const keys = behaviors.keys();
```
EisenbergEffect, line 276: How does this approach compare to a normal for/of loop? Is this faster?

EisenbergEffect (approved):

packages/web-components/fast-element/src/controller.ts
nicholasrice: I didn't actually know you could for/of a Map but of course you can. It appears to be slightly slower in general: https://jsbench.me/pbkkit8zui/1 - I Changed this to use for/of because it's more concise and faster. Thanks for calling this out!

chrisdholt (approved):

PR #4331, opened 2021-02-04
fix: accordion arrow key press moves focus from content to accordion header

### LINKED ISSUE #4326: accordion focus moves from focusable panel content to header when arrow key up or down are pressed
The focus is moving from accordion item panel focusable content back to the accordion headers when the down or up arrow keys are pressed.

Example:

see stackblitz repro: https://stackblitz.com/edit/fast-design-tku4xg?file=index.html

1. Focus on accordion header
2. Tab focus to any button in panel
3. Press arrow up or down key
4. Notice how focus moves to accordion headers

The accordion item should _not_ handle arrow key presses or focus unless the focus is _already_ on the accordion headers.
See ARIA Spec Accordion Keyboard Interaction: https://w3c.github.io/aria-practices/#keyboard-interaction
> Down Arrow (Optional): If focus is on an accordion header, moves focus to the next accordion header. If focus is on the last accordion header, either does nothing or moves focus to the first accordion header.
Up Arrow (Optional): If focus is on an accordion header, moves focus to the previous accordion header. If focus is on the first accordion header, either does nothing or moves focus to the last accordion header.

accordionkeycast

* OS & Device: MacOS, Windows, on PC
* Browser Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox

SethDonohue:
Added focusable content to default example
Added check for target to equal current target before handling accordion item key down event.

If the event.target and event.currentTarget are the same then the arrow key press happened on the accordion itself, which is the event we want to act on with arrow up/down keys.

Note: this _also_ makes focus _not_ move if it is on one of the start or end items in the accordion header. I believe this is the appropriate response, as that is focusable content so  the accordion item key down handler should not act on that. Only if the actual accordion item is focused do we want the arrow keys to move focus between accordion items.

# Description
 closes #4326 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/accordion/accordion.ts
+++ b/packages/web-components/fast-foundation/src/accordion/accordion.ts
@@ -127,25 +127,29 @@ export class Accordion extends FASTElement {
     }
 
     private handleItemKeyDown = (event: KeyboardEvent): void => {
-        const keyCode: number = event.keyCode;
-        this.accordionIds = this.getItemIds();
-        switch (keyCode) {
-            case keyCodeArrowUp:
-                event.preventDefault();
-                this.adjust(-1);
-                break;
-            case keyCodeArrowDown:
-                event.preventDefault();
-                this.adjust(1);
-                break;
-            case keyCodeHome:
-                this.activeItemIndex = 0;
-                this.focusItem();
-                break;
-            case keyCodeEnd:
-                this.activeItemIndex = this.accordionItems.length - 1;
-                this.focusItem();
-                break;
+        // only handle the keydown if the event target is the accordion item
+        // prevents arrow keys from moving focus to accordion headers when focus is on accordion item panel content
+        if (event.target === event.currentTarget) {
+            const keyCode: number = event.keyCode;
+            this.accordionIds = this.getItemIds();
+            switch (keyCode) {
+                case keyCodeArrowUp:
+                    event.preventDefault();
+                    this.adjust(-1);
+                    break;
+                case keyCodeArrowDown:
+                    event.preventDefault();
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -125,25 +125,29 @@ export class Accordion extends FASTElement {
...
-                this.activeItemIndex = 0;
-                this.focusItem();
-                break;
-            case keyCodeEnd:
-                this.activeItemIndex = this.accordionItems.length - 1;
-                this.focusItem();
-                break;
+        // only handle the keydown if the event target is the accordion item
+        // prevents arrow keys from moving focus to accordion headers when focus is on accordion item panel content
+        if (event.target === event.currentTarget) {
```
radium-v, line 130: nit: instead of wrapping the entire block, you could invert it and return when they're not equal.

radium-v (approved):

packages/web-components/fast-foundation/src/accordion/accordion.ts
chrisdholt: Yeah, would prefer a guard here rather than wrapping the entire block to execute. 

```
if (event.target !== event.currentTarget) {
    return;
}
```

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/accordion/accordion.ts
SethDonohue: Updated!

chrisdholt (approved):

PR #4336, opened 2021-02-04
chore: remove build from test steps

radium-v:
# Description

Removes `build` and related steps from corresponding `test` steps.

## Motivation & context

Our current testing actions can be broken down by type:
- Packages using `karma` and `mocha`, which require a build to complete before running tests:
  -  `fast-element`, `fast-foundation`, `fast-components`
  - Since the current workflow expects the project to be built before tests will be runnable, no changes need to be made for these.
- Packages using `jest`, which runs directly against the source files anyway:
  - `fast-tooling`, `fast-tooling-react`, `fast-animation`, `fast-colors`, `fast-web-utilities`
  - Running `build` for these packages has no effect on the test output, and automatically building after running unit tests is redundant. This can be verified by running `yarn clean:dist` before running `yarn test` in these packages.
- Packages using `webpack` which don't have any test processes in place:
  - `fast-website`, `fast-tooling-examples`, `fast-creator`, `fast-component-explorer`, `fast-color-explorer`, `fast-figma-plugin-msft`, `site-utilities` (doesn't use `webpack` but it doesn't have any tests)
  - Running tests shouldn't change the contents of the previous build, and re-building when calling `test` is confusing.

Due to the interconnected dependencies between pacakges, the `prepare` step still needs to be run before most packages will be able to run any tests (but this is the case regardless of these changes).

## Issue type checklist

## Process & policy checklist


sites/fast-component-explorer/package.json
```diff
@@ -12,7 +12,7 @@
     "prettier": "prettier --config ../../.prettierrc --write \"**/*.{ts,tsx,html}\"",
     "prettier:diff": "prettier --config ../../.prettierrc \"**/*.{ts,tsx,html}\" --list-different",
     "start": "webpack-dev-server --config=./webpack.dev.js --history-api-fallback --progress",
-    "test": "yarn convert:readme && yarn eslint && yarn build:app",
+    "test": "yarn eslint",
```
chrisdholt, line 15: I don't think the `build:app` step will be run during any existing part of our build process. Is it possible that an update to components causes a break (API change, etc) which would then not be caught by ensuring that the app builds properly. A big part of running this during the test step was because the app building in itself is *the test*. How do we address that scenario with this change?

sites/fast-color-explorer/package.json
```diff
@@ -9,7 +9,7 @@
     "prettier": "prettier --config ../../.prettierrc --write \"**/*.{ts,tsx,html}\"",
     "prettier:diff": "prettier --config ../../.prettierrc \"**/*.{ts,tsx,html}\" --list-different",
     "start": "webpack-dev-server",
-    "test": "yarn eslint && yarn build",
```
nicholasrice, line 12: For packages that have / are *sites* I don't think removing `build` is safe. There is potential for code changes or dependency updates to affect whether these packages successfully build, and there is no CI process that validates these packages build if we remove `build` from `test`.

radium-v:
This PR was hastily made in the pursuit of speeding up process times. See #4346 for discussion.

PR #4349, opened 2021-02-08
fix: add check for intersection observer support before instanciating

scomea:
# Description
We should not throw an error when intersection observer is not supported. 

## Motivation & context
Partners report testing environments have been throwing errors when code tries to create an intersection observer

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/anchored-region/intersection-service.ts
```diff
@@ -50,6 +55,10 @@ export class IntersectionService {
      * initialize intersection detector
      */
     private initializeIntersectionDetector = (): void => {
+        if (!window.IntersectionObserver) {
```
EisenbergEffect, line 58: Instead of using `window` import `$global` from `fast-element`. It's a platform safe global you can use for these sorts of checks.

EisenbergEffect (changes requested):
What environment is this being run in where there's no intersection observer?

packages/web-components/fast-foundation/src/anchored-region/intersection-service.ts
scomea: Cool, done.

chrisdholt (approved):

scomea:
@EisenbergEffect Some partners reported errors during test runs, I expect it isn't supported in their test suites and our recent changes to how we initialize intersection observer manifested it.

eljefe223 (approved):

EisenbergEffect (approved):

PR #4356, opened 2021-02-10
fix: update the event listener to be passed to a config to the message system

janechu:
# Description

This change fixes the issue where the event listener was passed through postMessage which should only accent JSON data but was passing tests as the test environment did not throw an error. It will now be passed to the message system which can be retrieved by other services after initialization.

Reliant on the completion of #4355

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/tooling/fast-tooling/src/message-system-service/message-system.service.ts
+++ b/packages/tooling/fast-tooling/src/message-system-service/message-system.service.ts
@@ -1,4 +1,5 @@
 import { MessageSystem } from "../message-system";
+import { Register } from "../message-system/message-system.props";
 import { MessageSystemServiceAction } from "./message-system.service-action";
 
 export interface IdentifiedAction {
@@ -22,7 +23,12 @@ export interface ActionNotFound {
     error: string;
 }
 
-export interface MessageSystemServiceConfig<C> {
+export interface MessageSystemServiceConfig<A, C = {}> {
+    /**
+     * The id for the service
+     */
+    id?: string;
+
     /**
      * The message system
      * used for sending and receiving shortcuts to the message system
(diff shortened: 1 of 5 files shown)


packages/tooling/fast-tooling/src/message-system-service/message-system.service.ts
```diff
@@ -81,14 +94,14 @@ export abstract class MessageSystemService<C> {
      * The service should get any config to be passed to
      * the registered actions
      */
-    abstract getActionConfig(id: string): C;
+    abstract getActionConfig(id: string): A;
```
chrisdholt, line 97: I know this is technically just a typing change, but curious if this is backwards compatible or breaking (are A and C "ducks")?

packages/tooling/fast-tooling/src/message-system-service/message-system.service.ts
```diff
@@ -81,14 +94,14 @@ export abstract class MessageSystemService<C> {
...
 
     /**
      * Returns an action with a specific ID that can be run
      */
     public action = (id: string): IdentifiedAction => {
         const action = this.registeredActions.find(
-            (action: MessageSystemServiceAction<C, unknown>) => {
+            (action: MessageSystemServiceAction<A, unknown>) => {
```
chrisdholt, line 104: Same question about ducks (or duck typing) here :)

packages/tooling/fast-tooling/src/message-system-service/message-system.service.ts
chrisdholt: *Not exported - disregard (thx @Janechu)

chrisdholt (approved):

nicholasrice (approved):

PR #4360, opened 2021-02-10
fix: whitespaceFilter filtering elements

SethDonohue:
# Description

The whitespaceFilter in utilities was filtering out anything that was not a TEXT_NODE.
It should return true if the item is an element, so that it doesn't filter elements out.

- Updated filter to check for type of non TEXT_NODE
- Added an example to Text-Field to test that using a <span> element as a label is not filtered out.
- Added tests for whitespaceFilter

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/utilities/whitespace-filter.ts
+++ b/packages/web-components/fast-foundation/src/utilities/whitespace-filter.ts
@@ -7,9 +7,9 @@
  * @public
  */
 export function whitespaceFilter(value: Node, index: number, array: Node[]) {
-    return (
-        value.nodeType === Node.TEXT_NODE &&
-        typeof value.nodeValue === "string" &&
-        !!value.nodeValue.trim().length
-    );
+    return value.nodeType !== Node.TEXT_NODE
+        ? true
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-foundation/src/utilities/whitespace-filter.ts
```diff
@@ -7,9 +7,9 @@
...
  */
 export function whitespaceFilter(value: Node, index: number, array: Node[]) {
-    return (
-        value.nodeType === Node.TEXT_NODE &&
-        typeof value.nodeValue === "string" &&
-        !!value.nodeValue.trim().length
-    );
+    return value.nodeType !== Node.TEXT_NODE
+        ? true
+        : value.nodeType === Node.TEXT_NODE &&
```
nicholasrice, line 12: this check is now redundant and can be removed.

SethDonohue: Truth. Removed

nicholasrice (approved):

chrisdholt (approved):

PR #4361, opened 2021-02-11
chore: clean up storybook stories and document storybook how-to

radium-v:
# Description

* Use storybook decorators to wrap all stories in a default `<fast-design-system-provider>`
* Remove existing provider elements from all stories
* Use side-effect imports to import components (removes the tree-shaking calls)
* Clean up some funky logic in some stories

## Motivation & context

Now that we're using an updated version of Storybook, we can use its features with more confidence. This reduces the complexity of writing a component's stories and will make it easier to start using other features offered by storybook.

## Issue type checklist

## Process & policy checklist


nicholasrice (approved):
Nice!

packages/web-components/fast-components/src/slider/fixtures/base.html
```diff
@@ -1,342 +1,340 @@
...
+    <!-- Min above 0 -->
+    <h2>Min value greater than 0</h2>
+    <div style="width: 160px;">
+        <fast-slider min="5" max="15" step="1" value="10" style="min-width: unset;">
+            <fast-slider-label hide-mark="true" position="5">
+                5
             </fast-slider-label>
-            <fast-slider-label position="1">
-                1
+            <fast-slider-label hide-mark="true" position="15">
```
janechu, line 86: Why are these values changing?

janechu (commented):
Looks good overall, only had one question in relation to changes to the stories for slider.

Edit: Also to note in future its easier to review smaller PRs as we are more likely to miss issues in large ones like this.

Edit 2: One strategy to consider if all of the changes should be performed at once is to create a feature branch and have the PRs be made against that.

packages/web-components/fast-components/src/slider/fixtures/base.html
radium-v: 🤔 that's weird. Must be a formatter issue.

janechu: Yep, must be, double checked it against the original and it looks correct.

janechu (approved):

PR #4368, opened 2021-02-16
fix: add default expanded and collapsed SVGs to accordion-item component

khamudom:
# Description
Added expanded and collapsed SVGs to the icon slots.
Removed SVGs in examples and added one example to showcase overriding the default icons.

Default
image
image

Added an example that overrides the default icon.
image
image

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
+++ b/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
@@ -33,8 +33,12 @@ export const AccordionItemTemplate = html<AccordionItem>`
             ${startTemplate}
             ${endTemplate}
             <span class="icon" part="icon" aria-hidden="true">
-                <slot name="expanded-icon" part="expanded-icon"></slot>
-                <slot name="collapsed-icon" part="collapsed-icon"></slot>
+                <slot name="expanded-icon" part="expanded-icon">
+                    ${x => x.expandedIcon()}
+                </slot>
+                <slot name="collapsed-icon" part="collapsed-icon">
+                    ${x => x.collapsedIcon()}
+                </slot>
             <span>
         </div>
         <div
--- a/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
+++ b/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
@@ -34,10 +34,10 @@ export const AccordionItemTemplate = html<AccordionItem>`
             ${endTemplate}
             <span class="icon" part="icon" aria-hidden="true">
                 <slot name="expanded-icon" part="expanded-icon">
-                    ${x => x.expandedIcon()}
+                    ${x => x.expandedIcon}
                 </slot>
                 <slot name="collapsed-icon" part="collapsed-icon">
-                    ${x => x.collapsedIcon()}
+                    ${x => x.collapsedIcon}
                 </slot>
             <span>
         </div>
--- a/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
+++ b/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
@@ -1,54 +1,80 @@
-import { html, ref } from "@microsoft/fast-element";
+import { html, ref, ViewTemplate } from "@microsoft/fast-element";
 import { endTemplate, startTemplate } from "../patterns/start-end";
 import { AccordionItem } from "./accordion-item";
+
+const expandedSVG = html<AccordionItem>`
+    <svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
+        <path
+            fill-rule="evenodd"
+            clip-rule="evenodd"
+            d="M3.78 3h12.44c.43 0 .78.35.78.78v12.44c0 .43-.35.78-.78.78H3.78a.78.78 0 01-.78-.78V3.78c0-.43.35-.78.78-.78zm12.44-1H3.78C2.8 2 2 2.8 2 3.78v12.44C2 17.2 2.8 18 3.78 18h12.44c.98 0 1.78-.8 1.78-1.78V3.78C18 2.8 17.2 2 16.22 2zM14 9H6v2h8V9z"
+        />
+    </svg>
+`;
+
+const collapsedSVG = html<AccordionItem>`
+    <svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
+        <path
+            fill-rule="evenodd"
+            clip-rule="evenodd"
+            d="M16.22 3H3.78a.78.78 0 00-.78.78v12.44c0 .43.35.78.78.78h12.44c.43 0 .78-.35.78-.78V3.78a.78.78 0 00-.78-.78zM3.78 2h12.44C17.2 2 18 2.8 18 3.78v12.44c0 .98-.8 1.78-1.78 1.78H3.78C2.8 18 2 17.2 2 16.22V3.78C2 2.8 2.8 2 3.78 2zM11 9h3v2h-3v3H9v-3H6V9h3V6h2v3z"
+        />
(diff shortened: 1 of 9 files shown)


EisenbergEffect:
Thanks for working on this @khamudom Let me advise a couple of modifications to this approach that will yield improved performance, reduced memory, and greater flexibility.

First, you'll want to extract the templates into module-level constants. The reason for this is that whenever you use the `html` function, you are creating a new template definition. This means that every instance of the accordion is going to generate a new template, rather than share a template. So, lets extract the templates so they can be re-used.

Second, instead of making these functions, I would make them `@observable` properties and set the default value of that property to the shared template. With that design, inheritance is not required to customize the icons. The properties can be set on individual instances (in vanilla JS or through a FAST binding) to change the icons.

chrisdholt:
> Thanks for working on this @khamudom Let me advise a couple of modifications to this approach that will yield improved performance, reduced memory, and greater flexibility.
> 
> First, you'll want to extract the templates into module-level constants. The reason for this is that whenever you use the `html` function, you are creating a new template definition. This means that every instance of the accordion is going to generate a new template definition, rather than share a template definition to create views from. So, lets extract the templates so they can be re-used.
> 
> Second, instead of making these functions, I would make them `@observable` properties and set the default value of that property to the shared template. With that design, inheritance is not required to customize the icons. The properties can be set on individual instances (in vanilla JS or through a FAST binding) to change the icons.

@EisenbergEffect we talked about addressing defaults with a common pattern in the coming months where teams could *change* the default slotted content as part of the provider work; do we want to establish this as a new pattern prior to that work? In other controls (such as breadcrumb, select, etc) we're setting some default slotted content already in the template. I'd personally prefer to keep that simple approach for the time being and then solve the larger issue as planned. While I think this is a creative approach, I think unless we see this as the long term approach to addressing this we should probably stay consistent.

EisenbergEffect:
In the very least, we should extract the templates, even if we want to keep the functions. In terms of the future approach, I need to do some tests, but I do think we'll have a way to let these things be configured at the design system level. That said, there's also a value to being able to set them on individual instances, like I show above. I think we could enable a pattern where inside the template, the property is used but if not present the design system configuration is used, getting the best of both worlds.

The pattern above is currently used by `FoundationElement` (part of the new design system work) to even enable overriding templates and styles on individual components where the fallback is the template/style pulled from DI.

chrisdholt:
> In the very least, we should extract the templates, even if we want to keep the functions. In terms of the future approach, I need to do some tests, but I do think we'll have a way to let these things be configured at the design system level. That said, there's also a value to being able to set them on individual instances, like I show above. I think we could enable a pattern where inside the template, the property is used but if not present the design system configuration is used, getting the best of both worlds.
> 
> The pattern above is currently used by `FoundationElement` (part of the new design system work) to even enable overriding templates and styles on individual components where the fallback is the template/style pulled from DI.

If we feel good about this approach, should we set this up for everything which currently has default slotted content? I'd much prefer consistency across implementation and if we feel confident this is a net-benefit and a good long-term direction I think we should just make the change across all components which could implement it.

packages/web-components/fast-foundation/src/accordion/accordion.spec.md
```diff
@@ -73,6 +73,32 @@ As defined by the W3C:
...
+        </svg>
+`
+
+ const collapsedIcon = html`
+    <svg>
+        <path/>
+    </svg>
+ `
+
+export class MyAccordionItem extends AccordionItem {
```
nicholasrice, line 93: While an author could absolutely create a new element that extends AccordionItem, I think a more common way to utilizes these properties would be to override the element instance *in a template*. Doing so is easier because it doesn't require defining a new element just change the properties. I think we should show this example as well:

```ts
html`
    <fast-accordion-item :expandedIcon=${expandedIcon}></fast-accordion-item>
`
```

nicholasrice (approved):

packages/web-components/fast-foundation/src/accordion-item/accordion-item.ts
```diff
@@ -1,7 +1,33 @@
...
+
+const collapsedIconTemplate = html`
+    <svg width="20" height="20" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
+        <path
+            fill-rule="evenodd"
+            clip-rule="evenodd"
+            d="M16.22 3H3.78a.78.78 0 00-.78.78v12.44c0 .43.35.78.78.78h12.44c.43 0 .78-.35.78-.78V3.78a.78.78 0 00-.78-.78zM3.78 2h12.44C17.2 2 18 2.8 18 3.78v12.44c0 .98-.8 1.78-1.78 1.78H3.78C2.8 18 2 17.2 2 16.22V3.78C2 2.8 2.8 2 3.78 2zM11 9h3v2h-3v3H9v-3H6V9h3V6h2v3z"
+        />
+    </svg>
+`;
```
chrisdholt, line 29: @EisenbergEffect any idea what the performance implications are for making these observable by default vs. just using it as default slotted content? I'd prefer to not make this change without making a change across *every other* instance we have default slotted content as well. I know we discussed a potentially different implementation as well; would the other approach be favored above this?

EisenbergEffect: The new design system approach isn't mutually exclusive with this approach. It depends on how we want to shape the extensibility. I think we should delay creating the observable in light of the future configuration-based approach though. Additionally, I also think I forgot that the slot could always be targeted, which makes the use of observable for this particular scenario much less meaningful.

@khamudom Apologies for the distraction 😢

khamudom: @EisenbergEffect made an update based on our discussion and your suggestions.

packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
```diff
@@ -1,50 +1,80 @@
...
-    <template
-        class="${x => (x.expanded ? "expanded" : "")}"
-        slot="item"
-    >
-        <div
-            class="heading"
-            part="heading"
-            role="heading"
-            aria-level="${x => x.headinglevel}"
+export function createAccordionItemTemplate(
```
chrisdholt, line 29: Wouldn't this be a breaking change? I'm confused why we need to take this approach rather than dropping the SVG into the slot by default.

khamudom: Yeah, I noted that this is a breaking change. And that was the initial approach to just add an SVG to the slot. I can still do that for a quick fix to have default icons. My understanding of the issue is when another library wants to use their own default icons, like fluent UI, they still need to keep adding an SVG to override the FAST version every time they implement the component. If this is not an issue I can take the initial approach.

chrisdholt: > My understanding of the issue is when another library wants to use their own default icons, like fluent UI, they still need to keep adding an SVG to override the FAST version every time they implement the component.
This is definitely still an issue but I think that's different than this ask. @EisenbergEffect has a PR up which would address this (thought that's being reconciled with all requirements). Definitely appreciate the extra effort here but we'd need to plan for a breaking change like this and I think we'd want to make that change across all components.

For this specifically, I think starting with a default SVG as slotted content would suffice and relieve the immediate pain.

khamudom: OK, I can definitely just add an SVG. 
Thanks!

SethDonohue:
Smoked. I don't see any regressions form master.

SethDonohue (approved):

chrisdholt (approved):

EisenbergEffect (approved):

EisenbergEffect:
@Ibrahimmaga Looks like the exact same Playwright failures on this PR as on mine.

janechu (approved):

PR #4379, opened 2021-02-18
feat: add combobox component

radium-v:
# Description
Adds a `<fast-combobox>` component.

image

There are four levels of autocomplete:
1. no autocomplete
2. "inline" autocomplete - As you type, the first matching option will automatically populate the field. The matching option will also be selected in the list.
3. "list" autocomplete - As you type, the list will filter to display only options that start with the value.
4. "both" autocomplete - inline and list methods will both be applied, meaning the list will shrink to match the filtered values, and the closest match will be automatically populated in the field.

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
@radium-v Is there a pattern for use where the list of items is dynamically populated based on what is typed into the input box? i.e. I've got thousands of items and I want to search based on what's typed, and provide potential options in the list dynamically.

packages/web-components/fast-foundation/src/combobox/combobox.template.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Combobox } from "./combobox";
+
+/**
+ * The template for the {@link @microsoft/fast-foundation#(Select:class)} component.
+ * @public
+ */
+export const ComboboxTemplate = html<Combobox>`
+    <template
+        autocomplete="${x => x.autocomplete}"
+        class="${x => (x.open ? "open" : "")} ${x =>
```
nicholasrice, line 13: I think we shouldn't reflect these attrs to classes and should encourage styling based on attr selectors because authoring systems like React can unknowingly remove classes set by the element on the element host.

packages/web-components/fast-foundation/src/combobox/combobox.ts
```diff
@@ -0,0 +1,563 @@
...
+ * @public
+ */
+export class DelegatesARIACombobox extends ARIAGlobalStatesAndProperties {
+    /**
+     * See {@link https://www.w3.org/WAI/PF/aria/roles#button} for more information
+     * @public
+     * @remarks
+     * HTML Attribute: aria-expanded
+     */
+    @observable
```
nicholasrice, line 535: should this be attr?

radium-v: I'm actually not sure if this should even be presented in this class, since it's only used on the internal control and is handled programmatically.

chrisdholt: Is this set on the combobox itself or a control in the template?

radium-v: @chrisdholt it's set on the text input control within the template (it's similarly set on the button control for Select).

I think this property is also set on the `DelegatesARIAListbox` class which doesn't seem right. I'll readdress the ARIA properties on these three components.

chrisdholt: @radium-v sounds good - in the case that the listbox *supports* that role but doesn't need it by default, I think it's okay for it to be there, but agree that it likely shouldn't be there if the host element itself needs it as part of the public facing API.

packages/web-components/fast-components/src/combobox/combobox.definition.ts
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    name: "disabled",
+                    description: "The disabled attribute",
+                    type: DataType.boolean,
+                    default: false,
+                    required: false,
+                },
+                {
+                    name: "name",
+                    description: "The name attribute",
```
chrisdholt, line 22:
```suggestion
                    title: "Name",
                    description: "This element's value will be surfaced during form submission under the provided name",
```

packages/web-components/fast-components/src/combobox/combobox.definition.ts
```diff
@@ -0,0 +1,74 @@
...
+                    type: DataType.string,
+                    default: "",
+                    required: false,
+                },
+                {
+                    name: "position",
+                    description: "The position attribute",
```
chrisdholt, line 29:
```suggestion
                    title: "Position",
                    description: ""Controls the placement of the combobox dropdown",
```

chrisdholt (commented):
We'll want to ensure that the definition file includes titles as well as descriptions now. I've added some suggestions below.

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
```diff
@@ -1,4 +1,4 @@
-import { attr, FASTElement, observable } from "@microsoft/fast-element";
+import { attr, FASTElement, observable, Observable } from "@microsoft/fast-element";
 import { isHTMLElement } from "@microsoft/fast-web-utilities";
```
EisenbergEffect, line 2: What other dependencies do we have coming from fast-web-utilities. If this is the only one, we could easily remove it since this is a pretty simple check to make inline.

EisenbergEffect (approved):
Pending Nick's note on removing the host class modification.

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
radium-v: For foundation, there are a lot of single-use constants and enums: `Direction`, `Orientation`, `KeyCodes` (as well as individual ones like `KeyCodeEscape`). There are also some functions like `wrapInBounds`, `limit`, `canUseFocusVisible`, and `DOM.queueUpdate`/`DOM.nextUpdate`.

For components, all but one are `SystemColors`, `Orientation`, and `Direction`, which are all enums. The remainder is the `RtlScrollConverter` class, which is only used in the `anchored-region` storybook story.

chrisdholt: We have quite a few - I don't think removing here will make a significant reduction. If we want to explore dropping the package as a dependency we should probably take it to an issue and determine how we can do so.

EisenbergEffect: Sounds good. I wasn't very familiar with what we were using from that package, so I just wanted to make sure, since this particular utility is pretty simple to inline if needed. No reason to change anything.

eljefe223:
image

For this example, if I type a letter that does not match I always get this result. Expected?

radium-v:
@eljefe223 not expected - fix incoming.

packages/web-components/fast-foundation/src/combobox/combobox.ts
```diff
@@ -0,0 +1,602 @@
...
+                    this.setSelectedOptions();
+                }
+                break;
+            }
+
+            default: {
+                if (key.length === 1 && key.match(/\S/)) {
+                    this.filterOptions();
+                    if (this.autocomplete === "list" || this.autocomplete === "both") {
+                        if (!this.open && this.control.value.length) {
```
janechu, line 420:
```suggestion
Some of these if statements can be consolidated since they only nest. Example:

```suggestion
                    if (
                        (this.autocomplete === "list" || this.autocomplete === "both") &&
                        (!this.open && this.control.value.length)
                    ) {
```
```

File: packages/web-components/fast-components/src/combobox/combobox.definition.ts [hunk 7]
```diff
@@ -0,0 +1,78 @@
...
+export const fastComboboxDefinition: WebComponentDefinition = {
+    version: 1,
+    tags: [
+        {
+            name: "fast-combobox",
+            title: "Combobox",
+            description: "The FAST combobox element",
+            attributes: [
+                {
+                    name: "disabled",
```
chrisdholt, line 14:
```suggestion
                    name: "disabled",
                    title: "Disabled",
```

packages/web-components/fast-components/src/combobox/combobox.definition.ts
```diff
@@ -0,0 +1,78 @@
...
+                    description: "Sets the disabled state of the combobox",
+                    type: DataType.boolean,
+                    default: false,
+                    required: false,
+                },
+                {
+                    name: "name",
```
chrisdholt, line 21:
```suggestion
                    name: "name",
                    title: "Name",
```

packages/web-components/fast-components/src/combobox/combobox.definition.ts
```diff
@@ -0,0 +1,78 @@
...
+                    description:
+                        "This element's value will be surfaced during form submission under the provided name",
+                    type: DataType.string,
+                    default: "",
+                    required: false,
+                },
+                {
+                    name: "position",
```
chrisdholt, line 29:
```suggestion
                    name: "position",
                    title: "Position",
```

chrisdholt (approved):
A couple additions in the definition file to take care of but looks good besides that!

Issue #4369, opened 2021-02-16
Design Systems VNext: Design Tokens


nicholasrice:
With the addition of Dependency Injection and new component registration methods provided by `FoundationElement` and `DesignSystem`, this issue proposes a new way to expose Design Tokens to an application using those features.

**Goals:**
- Make Design Tokens a top-level API surface to better reflect industry standards and terminology
- Remove the need of a wrapping `fast-designs-system-provider` element
- Provide mechanisms to make Design Token implementations injectable and customizable by app authors / component libraries.
- Provide mechanisms to allow Design Tokens to have complex hierarchies.
- Simplify difficulties in setting the `background-color` design token to a color recipe implementation.
# Design Tokens
Design tokens are a fundamental building block for creating design systems, and inform individual property values from the design language. Examples of design tokens are colors, font sizes, spacing units, etc.

## Requirements
1. Accessible through DOM tree
2. Can have values reflected to CSS custom properties at arbitrary DOM nodes.
3. Mutable through the DOM tree. A token value at one node may not be the value at another node.
4. Observable. It must be possible to know when a token changes.

## Implementation of a Design Token
A design token is a typed, unique symbol that can be used to resolve *values*. The generic parameter identifies the runtime type the token will resolve.

```ts
type DesignToken<T> = {};
```

## Design Token Library
The token library is responsible for *managing* design tokens. It is where token values are set and retrieved, and is responsible for emitting notifications when tokens are set. 

I was originally hoping the DI Container could be used to manage setting and getting values of Design Tokens but ran into the following issues:
1. We need to optionally reflect a token to CSS custom properties when it is set. This isn't a behavior of DI containers nor should it be, so we need *some* abstraction here.
2. DI containers don't *re-resolve* if an intermediary Container registers a value. This becomes problematic when a Design Token is set during an app lifecycle at a node it was not previously set; this is invisible to previously resolved values.

Because of these limitations and requirements, a purpose-specific resolution and notification system will be required.

**DesignTokenLibrary Interface**
```ts
interface DesignTokenLibrary {
    /**
     * Enables hierarchial properties through a tree
     */
    upstream: DesignTokenLibrary | null;
   
    subscribe(subscriber: Subscriber, ...tokens: DesignToken<any>[]);
    unsubscribe(subscriber: Subscriber, ...tokens: DesignToken<any>[]);
    get<T>(token: DesignToken<T>): T;
    alias<T>(alias: DesignToken<T>, target: DesignToken<T>): void;

[...]

Issue #3748, opened 2020-08-19
[Proposal] Enable Binding in ElementStyles


EisenbergEffect:
I had an idea for limited binding in css, so I wanted to share it and see what people thought.

## Example

With this feature in place, you could write the following:

```ts
const styles = css<MyElement>`
  :host {
    padding: ${x => x.padding};
  }
`;

@customElement({
  name: 'my-element',
  styles
})
export class MyElement extends FASTElement {
  @attr size: string = 'medium';

  get padding() {
    switch(this.size) {
      case 'small':
        return '4px';
      case 'medium':
        return '8px';
      case 'large':
        return '16px';
    }
  }
}
```

This allows the same stylesheet to be shared across all instances of `MyElement` but the value of host padding will be dependent on the state of the individual element instance.

## Technical Approach

A naive approach would be to watch bindings and when they change re-evaluate the entire stylesheet by concatenating css and replacing the old styles with the new. While this is the most flexible solution, allowing for unlimited dynamic css generation, it performs very poorly, forcing a full css parse for every change, a new stylesheet instantiation, and a dom update to add/remove sheets.

Alternatively, we could replace the binding expression with a generated, unique CSS variable and then simply update the CSS property value on the host with each change. You would end up with a single stylesheet that looked like this:

```css
:host {
  padding: var(--fast-abc123);
}
```

The limitation is that bindings could only be used in positions where CSS variables are allowed. But, the only work to update is to watch the expression and write the property when it changes.

### Implementation

To implement this, we would enhance the `css` tagged template function so that it allows instances of `Directive` in the content, just like templates do. We would then capture all arrow functions and convert them to a `CSSBindingDirective`. This directive would write a placeholder into the CSS which would be the generated variable name. It would then be used to create a behavior based on the expression and variable name. All of this could use the same `Directive` api we have for templates by implementing `createPlaceholder` and `createBehavior`. Since styles already can have behaviors associated with them, the created behavior is added to the styles. Whatever custom element the styles are added to (either as primary or dynamic styles), the behavior will be activated resulting in property subscription and CSS variable updates.

This approach also enables other CSS directives to be written as well, which could generate CSS and attach behaviors for other purposes than simple css/expression synchronization. One idea would be to change the way our CSS var behaviors are functioning. For example, in our card styles, we have something like this:

```ts
export const CardStyles = css`
    :host {
        ...
        background: ${neutralFillCardRestBehavior.var};
        ...
    }
`.withBehaviors(
[...]

PR #4383, opened 2021-02-19
feat: adds CSSDirective for use with the css tagged template literal

nicholasrice:
# Description
This PR adds a new type, `CSSDirective` that can be used to bind behaviors and interpolate CSS into the the css tagged template literal.

This lays the groundwork for CSS bindings (#3748) and the new Design Token work being discussed in #4369.

## Issue type checklist

## Process & policy checklist


awentzel (approved):

packages/web-components/fast-element/src/styles.ts
```diff
@@ -279,6 +303,13 @@ export function css(
             }
 
             styles.push(value);
+        } else if (value instanceof CSSDirective) {
```
EisenbergEffect, line 306: I had an idea when looking at the chain of conditionals. What if we put this check first? Then, we could allow the return result of `createCSS()` to actually be `ComposableStyles`. So we'd basically take the output of `value.createCSS()` and assign it to `value` so that the checks below would then handle the other types. This could let a `CSSDirective` return not only strings but also `ElementStyles` and `CSSStyleSheet` instances. Thoughts?

EisenbergEffect (approved):

packages/web-components/fast-element/src/styles.ts
nicholasrice: Yea good call, I'll update

Issue #3042, opened 2020-04-30
Add code coverage using Node V8 instead of Babel for perf improvements


awentzel:
**What feature are you requesting?**
Using the [V8 Engine](https://jestjs.io/docs/en/configuration#coverageprovider-string) to process Jest code coverage instead of Babel.

**What would be the benefit of adding this feature?**
Greater performance

**What solution would be ideal for you?**
An optimized, easy to maintain and performant solution

**What alternatives have you considered?**
n/a

**Is there any additional context for your request?**
As of this feature request it's still early so we should wait for a mature release and better documentation.

**Attach screenshots or images add detail to your request:**

PR #4388, opened 2021-02-23
chore: update to improve performance using v8 rather than babel for code coverage

Ibrahimmaga:
# Description
Indicates which provider should be used to instrument code for coverage. Allowed values are babel (default) or v8.

Note that using v8 is considered experimental. This uses V8's builtin code coverage rather than one based on Babel. It is not as well tested, and it has also improved in the last few releases of Node. Using the latest versions of node (v14 at the time of this writing) will yield better results.

 "testURL": "http://localhost",
    "transform": {
      "^.+\\.ts?$": "ts-jest"
      "^.+\\.js?$": "babel-jest" 
      we have add GITHUB_TOKEN to enable the code coverage to code climate
## Motivation & context
Here we are looking for performance.  from local host machine it has decreased total time from 105s to 75s. We will test in CI-workflow to get a better picture.
 

Performance

https://github.com/microsoft/fast/issues/3042
## Issue type checklist

This can affect the testing with Jest and Babel

## Process & policy checklist


package.json
```diff
@@ -48,6 +48,7 @@
     "prepare": "lerna run prepare",
     "test": "yarn eslint \"/**/*.{ts}\"",
     "unit-tests": "jest --maxWorkers=4",
+    "coverage": "npx jest --coverage",
```
awentzel, line 51: We're not using `npx` anywhere so I think we should stick with yarn to be consistent.

package.json
```diff
@@ -83,7 +84,7 @@
     "testURL": "http://localhost",
     "transform": {
       "^.+\\.ts?$": "ts-jest",
-      "^.+\\.js?$": "babel-jest"
+      "^.+\\.js?$": "babel-jest"
```
awentzel, line 87: I don't see any different with these lines.

.github/workflows/ci-weekly.yml
```diff
@@ -120,24 +120,16 @@ jobs:
         yarn test-firefox:pw 
         yarn test-webkit:pw
 
+    - name: Test Coverage
+      run: coverage
```
awentzel, line 124: Does 'coverage' run all of our packages?

awentzel (changes requested):
25% improvement is excellent.  I would like to have a walk-through on this PR to see it in action.  Please schedule a half-hour for me to review live.  Thanks

.github/workflows/ci-weekly.yml
Ibrahimmaga: Yes. It does now

awentzel (approved):

awentzel (approved):

.github/workflows/ci-nightly.yml
```diff
@@ -35,8 +35,8 @@ jobs:
       run: lerna run prettier:diff
 
     - name: Testing unit tests
-      run: lerna run test --stream
-
+      run: lerna run test --stream   
+
```
chrisdholt, line 39: Should these added spaces be removed?

awentzel: @chrisdholt this space should have been there before. It creates spacing between tasks which is consistent across all yaml pipelines.  So this is accurate.

chrisdholt: @awentzel I think an empty return is expected but I'm not sure about the (several) empty spaces after the text and on the next line. I won't block on it but the above is different than an empty return. Not sure if it'll cause issues or not though.

Ibrahimmaga: This won't cause any issues.

EisenbergEffect (approved):
Let's add a follow-up task to have GitHub run a coverage check any time a PR is submitted so that we can see whether individual PRs increase or reduce coverage. Or is that already in place? I see it in this PR but not others. Was it this PR that added that?

Ibrahimmaga:
Yes.  it is included in this PR. I have made the changed yesterday. For each PR, we will have a code coverage upload to code climate.

EisenbergEffect:
Excellent. Thank you!

chrisdholt (approved):

awentzel:
> Let's add a follow-up task to have GitHub run a coverage check any time a PR is submitted so that we can see whether individual PRs increase or reduce coverage. Or is that already in place? I see it in this PR but not others. Was it this PR that added that?

We refactored to ensure code coverage would happen on push to master and pull_request events to solve your concern.

Issue #3427, opened 2020-07-01
CI is not catching untracked api-report.md files


nicholasrice:
**Problem:**
Running `yarn` on a clean master branch will periodically result in untracked git changes to `api-report.md` files in web component packages. 

**Expected:**
Running `yarn` should *never* cause un-tracked changes to files.  Building `master` should always be transparent to git.

**Cause:**
`api-report.md` files are generated by the `doc` process of some packages (`/packages/web-components/**`, `fast-colors`, and `fast-animation`). These files are the product of `d.ts` files adorned with TSDoc tags and are affected by code changes that affect the *public* API of the package.

Often, a PR author will not run the processes locally to build the `api-report.md` files prior to putting a branch that affects the public API up for a pull request. The CI is checking for outdated `api-report.md` files on the CI in the `test` process, however updates to these files are not actually being caught. This is because of the following:

1. CI runs `yarn`.
    1. Installs dependencies
    2. Runs `build` for each package
    3. `build` runs `doc` and docs are build - at this point, `api-report.md` files are generated but *may* contain untracked changes in the case described above.
2. CI runs `test`
    1. `test` runs `doc:ci` which will fail is `api-report.md` files are not up-to-date. However, because `doc` has already been run, this always passes (even though they may contain *untracked* changes)
3. test passes

To mitigate this issue, I'd propose that the CI should ensure that there are no un-tracked changes as a result of building and testing a pull request.

PR #4389, opened 2021-02-23
bug: fix untracked files

Ibrahimmaga:
# Description
Running yarn on a clean master branch will periodically result in untracked git changes to api-report.md files and among other files in web component packages.  This can occur based on how the Web Component packages were cross-testing out of order.  This is not an API Report-specific issue.

 To preserve the untracked file that may delete the api-report.md file. We have removed the git clean -xdf and added script for more validation test:diff and " run yarn prepare" on fail or error.

 "test:diff:error": "echo \"Untracked files exist, try running yarn prepare to identify the culprit.\" && exit 1",
    "test:diff": "git update-index --refresh && git diff-index --quiet HEAD -- || yarn test:diff:error",
    "test:validation": "yarn test:diff",
    "test": "yarn eslint \"/**/*.{ts}\" ",

## Motivation & context
Reduce intermittent failures on the CI.

This solves the failed untracked child process on CI

https://github.com/microsoft/fast/issues/3427 
## Issue type checklist

- [x] **Bugfix**: A change that fixes an issue, link to the issue above.

## Process & policy checklist


.github/workflows/ci-weekly.yml
```diff
@@ -38,6 +32,9 @@ jobs:
 
     - name: Prepare workspaces
       run: yarn prepare
+    
+    - name: Clean and remove untracked files
+      run: git clean -xfd
```
chrisdholt, line 37: Shouldn't this run prior to prepare?

Ibrahimmaga: You must be right. I believe anywhere before the test get executed is fine. I'm still in the testing phase

.github/workflows/ci-weekly.yml
```diff
@@ -4,12 +4,9 @@ on:
   push:
```
awentzel, line 4: the push and pull_request events should be reverted and placed in a different PR since they are unrelated to this work.  I believe you have another PR already for this or coming?

awentzel (changes requested):

.github/workflows/ci-weekly.yml
Ibrahimmaga: Yes. I will have a PR for that.

.github/workflows/ci-weekly.yml
awentzel: Just to note this is a CI only change and required to prevent a lot of the inconsistent pipeline issues which break the build.

Ibrahimmaga: This doesn't work before prepare but it work right after git/checkout

chrisdholt: Confirming - I think this is what we discussed yesterday on a call, but for the untracked files (API Report) that's by design and we shouldn't be running git clean -fdx. That should fail the build consistently.

The repo should start clean so I think this PR may be unnecessary.

awentzel: @chrisdholt you are correct in that new PR runs "should" be from a clean machine.  However, my understanding is the problem comes after the packages are installed, and tested when this API file is generated.  Then my understanding is from git perspective there is an untracked file and so the pipeline sporadically breaks.  I've also, witnessed this for months in my local development environment.  It seems like this file should be ignored entirely by .gitignore.

@Ibrahimmaga can you confirm that that running `git clean -xfd` does not fail the build?  If this does not fail the build, then still I see this could be a problem in the local developer environment as well.  So what is your solution to that scenario?

.github/workflows/ci-weekly.yml
```diff
@@ -160,7 +166,7 @@ jobs:
 
     - name: Install package dependencies
       run: yarn install --frozen-lockfile --ignore-scripts
-
+
```
awentzel, line 169: Ideally, these empty spaces could be reverted as well due to the sensitivity of YAML syntax issues.

awentzel (commented):
If we need to have another 15-minute conversation I would include everyone commenting on this PR plus Nick.

awentzel (approved):

.github/workflows/ci-weekly.yml
chrisdholt: @awentzel Nick is out but we can hop on a call. The untracked file sporadically happens because a dev has not done something required of them. In this case, it's primarily random because the API has changed and they have not updated the API report which should result in a failure.

chrisdholt (dismissed):
Per comments, holding on this as I don't believe this is needed and the failures noted here are by design.

.github/workflows/ci-weekly.yml
awentzel: @chrisdholt How can we automatically update the API report such that this doesn't happen in the first place?  Does this file need to be committed?  How do we know that a build break is caused by this, and thus we know what the fix is?  That seems like something that we should eliminate from being a problem.  It costs a lot of time to troubleshoot for contributors that aren't aware of this being the issue.

awentzel: I would expect some test that determines if the API has changed, and if so, request by CLI, the action that must be taken to resolve, commit, and push.

chrisdholt: @awentzel Great questions - I'll answer below:

- We do run this automatically and update as part of both `yarn build` and `yarn test` - this is the reason that it's getting generated on the CI. Those are not being run locally by a dev. This is similar to prettier being bypassed. We can't ensure that someone *will always run this* because you can *always* bypass the command - because of this, we need to fail the build to maintain the integrity of the main branch.

- Yes, the file needs to be committed - if the file is not committed the API report which powers our API documentation gets out of date. If we do not fail the build, contributors will hit this locally when running a fresh install with `yarn`, or when bootstrapping, or when testing. In the case that we wanted to update staging after every merge to master, our docs would be out of date without this being committed along with API changes.

- We can know that a build break is caused by this because the error in the break says that API Report was generated and there are uncommitted files. We discussed on the call a few improvements that can help here. 1) We should do a brown bag to ensure our direct team knows how to investigate a build failure. Not every build failure is an issue with the build - many are issues with tests, or prettier, or the API report...can we help raise awareness for how to read the build output and error? 2) We can investigate how to provide a more prescriptive hint if API report is generated. We can troubleshoot for external contributors, but folks on our team need to know how to debug build issues.

chrisdholt: > I would expect some test that determines if the API has changed, and if so, request by CLI, the action that must be taken to resolve, commit, and push.

Yep, this is run as part of `test` which is fairly representative of what it is.

awentzel: @chrisdholt could it be as simple as being more verbose in the error message that we write to the terminal?  While "API Report was generated and there are uncommitted files" tells me the problem, but not the solution.  I would instead recommend changing this to "An API Report was generated due to a breaking API change. Please commit the unstaged files to your branch and resubmit the PR" or something to this effect?

chrisdholt: Totally agree - I didn't put the exact message, and I don't think we actually control what's happening now, but I do think that we could find a way to provide a better signal here for sure. I mentioned that on our call last week with Ibrahim, I'm just not sure the best way to break it down yet. We could break down our tests on the CI separating this check from the other - the name of the step could be "Validate API report" which would be a big signal. Perhaps that makes sense? Rather than run `yarn test` on CI, we run both steps individually. Thoughts on that? I can push a quick draft PR to gauge consensus.

awentzel: I think that makes sense.  The "signal" then would be an improvement and easy to understand since it's listed as a script.  Natural tendency should be to rerun that test locally and see if it can be reproduced.  But, I think we still need documentation if we can't add verbosity to the message, such that people understand how the API Report works and what depends on it?  My apologies if we have that.  I haven't looked for it.

chrisdholt: Yes, definitely need both - I think I'd keep the `yarn test` script as it's nice to run both for local development. If need be though on CI we could call each individually which would give a solid hint. Docs for sure need to be added for contributors on the API report, how to generate, etc.

awentzel: Proposed Fix:  We move the generic fail safe measure from each package and place at the root to call at the end of our pipeline process.  This will then, run to ensure there are **no** untracked files for **any** reason, a symptom of a bigger issue.  So this fix is not directly related to API Reports. 

So we are moving:
```
    "test:diff:error": "echo \"You have untracked files, try running yarn prepare.\" && exit 1",
    "test:diff": "git update-index --refresh && git diff-index --quiet HEAD -- || yarn test:diff:error",
``` 
From the three package.json files they exist in, into the root package.json file.  This test can be labeled as "Final Testing Validation"  If anything is found, here at the end, then we know it's a new issue and should be resolved appropriately.  The new issue could be because of a process failure, workflow failure, or implementation failure.  All things we can then work to harden.

Ibrahimmaga: This could be a good solution. instead of removing the untracked file we will just rebuild.

awentzel: These changes have been made and now ready for re-review.

.github/workflows/ci-weekly.yml
```diff
@@ -136,7 +139,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-
+
```
chrisdholt, line 142:
```suggestion
```

.github/workflows/ci-weekly.yml
```diff
@@ -152,7 +155,7 @@ jobs:
 
     - name: Install package dependencies
       run: yarn install --frozen-lockfile --ignore-scripts
-
+
```
chrisdholt, line 158:
```suggestion
```

.github/workflows/ci-weekly.yml
awentzel: @chrisdholt this is by design.  There should be an empty row between steps/tasks for easier readability.

.github/workflows/ci-weekly.yml
awentzel: By design, as empty spaces between steps/tasks make it easier to read.

.github/workflows/ci-weekly.yml
awentzel: by design, as empty rows between steps/tasks makes it easier to read.

.github/workflows/ci-weekly.yml
```diff
@@ -4,6 +4,9 @@ on:
   push: 
     branches:
     - master
+  pull_request:
+    branches:
+    - master
```
awentzel, line 9: Once this is approved a final commit should be made to remove `pull_request`.  It's just being used for testing, @Ibrahimmaga for the reminder.

nicholasrice (approved):

Issue #4381, opened 2021-02-18
Non-destructive ElementDefinition Configuration


nicholasrice:
This issue concerns the alpha `DesignSystem` and `FoundationElement` interfaces.

## Feature Description
To scaffold Design System features, we would like to be able to compose ElementStyles into a component definition's stylesheets during registration with the Design System. This would allow a component definition to expose very basic structural styling, with design opinion layered on top during registration, as well as arbitrary extensions to existing element definitions.

## Problem
It is not currently possible to extend the product of `FoundationElement.configuration()`, only override:

```ts
import { MyElementConfiguration } from "./my-element.ts";

new DesginSystem().register(MyElementConfiguration({styles: css`` /* This overrides (not extends) the existing styles */ }))
```

## Proposed Solution
There are a few options on how we could address this:

**1. Change the behavior We could change the behavior to merge styles instead of override the property. This would prevent removal of styles though, which is perhaps not what we want.**

**2. Expose an `addStyles()` method on the returned configuration function**
```ts
const myConfiguration = FoundationElement.configure({/* ... */});

myConfiguration.addStyles(myAddativeElementStyles);

new DesignSystem().register(myConfiguration());
```

**3. Expose a more general `compose()` method This would work a lot like the above, but would implement a sophisticated deep merge of the provided argument with the definition:**

```ts
const myConfiguration = FoundationElement.configure({/* ... */}, styles: css`` );

myConfiguration.compose({styles: myOtherStyles}); // -> Converts the `styles` property to an array including both stylesheets.

new DesignSystem().register(myConfiguration());
```

The benefit of this is that object-options like `elementOptions`, `attributes`, and `shadowOptions `could be additively configured.

My preference would be to go with option 3 (or something similar) because it provides the most flexibility for app authors. Thoughts?

nicholasrice:
Revisiting this now that #4390 has gone in.

The complication for this feature comes primarily from the `styles` data and potentially the `attributes` data. In practice I don't think the `attributes` data will be edited often, but it does manifest as an array like styles.

To add and remove data from the definition's `styles` property, we need to expose some mechanism to *concat* and *filter* the styles. We cannot simply expose the original definition object and require authors to mutate that data into a new definition because #4390 introduces a function type that can be assigned to `styles` that lazily defines the styles. As a library we can help handle this case ergonomically with *some* API that augments an existing definition. There are a two primary API models I see that we can use to allow mutations of definitions are:

**Static methods returning new definitions**
```ts
[...]

PR #4390, opened 2021-02-23
feat: enable FoundationElement.compose to accept template functions

EisenbergEffect:
# Description

This PR is part of the Design System vNext work and expands our capabilities for registering our web components to enable the developer to provide a function in place of template, styles, etc. The function will be executed when the element is being defined so that it has access to the configured element prefix, icon overrides, etc.

Here's an example of how this might be used to enable a component to be defined when it doesn't know the final tag name of its internal components and when it also wants to render default slotted icon content, but also enable the end-user to change the default icon for all instances of the component:

#### In @microsoft/fast-foundation

```ts
export class Foo extends FoundationElement {
  @attr baz;
}

export const FooTemplate = (context, definition) => html<Foo>`
  <${context.tagFor(Bar)} qux={x => x.baz}>
    <slot>
      ${definition.defaultIcon}
    </slot>
  </${context.tagFor(Bar)}>
`;
```

#### In @microsoft/fast-components or @fluentui/web-components

```ts
export const FooStyles = (context, definition) => css<Foo>`
  ...css here...
`;

export const foo = Foo.compose({
  baseName: 'foo',
  template: FooTemplate,
  styles: FooStyles
});
```

#### In MyApp

```ts
fluentUIDesignSystem()
  .register(
    foo({ defaultIcon: '...some html or even a template...' }),
    bar()
  )
  .applyTo(document.body);
```

>Note: I also renamed `FoundationElement.configure` to `FoundationElement.compose` and enabled it to use `this` internally so it can be called from the derived type (e.g. Foo above) and you don't need to pass the type again.

> Note: You do not need to pass a function for template or styles. You can also pass a template and styles object directly, just as in the past. Additionally, you can pass functions for element config, shadow dom options, etc. as well.

## Issue type checklist

## Process & policy checklist


chrisdholt (approved):

nicholasrice:
I like this but I think we'll want to re-approach #4381 with it. I think *technically* this PR actually enables that issue's feature, but an author might need to do some cumbersome logic to additively merge or remove data.

I think we can figure that out though, this looks great!

nicholasrice (approved):

EisenbergEffect:
@nicholasrice I just re-read your proposal. I totally didn't realize I was taking your `compose` naming scheme and interpreting it differently. If you think it makes more sense, we can change the method name back to `configure` (or to something else) and then add a `compose` method to do additive changes. Either way, I think we need what you propose in #4381 in addition to this PR so that folks can both replace the styles or simply add new styles on top of the base styles.

Do you want to just apply your work directly into this branch and then we can merge the final product? That might make it easier for you to work out the method names and the final shape of the API for all the scenarios. Whatever is easiest for you. Let me know.

nicholasrice:
@EisenbergEffect I think we can get this in and follow up with an implementation for #4381

EisenbergEffect:
@Ibrahimmaga It looks like Playwright is consistently failing. Can you advise?

janechu (approved):

Ibrahimmaga:
I have a PR that will fix the Issue. Playwright version 1.9 just got broken that what we are using on the CI. I downgraded to 1.8 and it’s working great 👍.

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #4391, opened 2021-02-24
feat: add horizontal scroll web component

robarbms:
next and previous flipper
bring this into fast foundation
DOM.queueUpdate(); instead of setTimeout

feat: adding a content scroller component

# Description

A component used to scroll horizontally through multiple cards.

http://robdor:6006/?path=/story/horizontal-scroll--horizontal-scroll

## Motivation & context

Used to scroll horizontally through a set of dynamic content. Smooth scrolling and alignment of different forms of content or cards.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/scroller/scroller.ts
```diff
@@ -0,0 +1,241 @@
...
+    ResizeObserverClassDefinition,
+} from "../anchored-region/resize-observer";
+import { ResizeObserverEntry } from "../anchored-region/resize-observer-entry";
+import { ScrollerControlStyles } from "./scroller.styles";
+
+declare global {
+    interface WindowWithResizeObserver extends Window {
+        ResizeObserver: ConstructibleResizeObserver;
+    }
+}
```
robarbms, line 17: @scomea I borrowed the code for resizeObserver from AnchoredRegion. I'm not sure if I did it right as I removed what I believe I didn't need.

SethDonohue:
Throwing out a name suggestion:

what about:
`conveyer`? It does have an er ending, but it looks like a conveyer as a conveyer typically is a long platform that moves items along it...

`turnstile`? This typically only lets one item through at a time.

scomea:
I like "gallery" myself.

chrisdholt:
> Throwing out a name suggestion:
> 
> what about:
> `conveyer`? It does have an er ending, but it looks like a conveyer as a conveyer typically is a long platform that moves items along it...
> 
> `turnstile`? This typically only lets one item through at a time.

I think my preference is to stay away from names that relate to something in the real world and stay more semantic or matter of fact. We previously had gone with `Horizontal Scroll` because it describes clearly the behavior. Conveyor for instance could describe a conveyer belt, or it could mean that it's relating to conveying information - not super clear IMO. Similar, turnstile would relate more to what a "Carousel" is IMO.

chrisdholt:
> I like "gallery" myself.

Could gallery also be represented with a single rotating instance though? Gallery seems like it could be a flat structure, it could be a list, or it could be a rotating slideshow (carousel). It is a catchy name, but I don't know that it would be clear what the intent is.

robarbms:
The closest model that I could find was most commonly called horizontal-scroller which I truncated to scroller.

packages/web-components/fast-components/src/scroller/scroller.definition.ts
```diff
@@ -0,0 +1,51 @@
...
+    version: 1,
+    tags: [
+        {
+            name: "fast-scroller",
+            title: "Scroller",
+            description: "The FAST scroller element",
+            attributes: [
+                {
+                    name: "view",
+                    description: "Default or mobile view",
```
bheston, line 14: Can this have the list of allowable values?

packages/web-components/fast-foundation/src/scroller/scroller.ts
```diff
@@ -0,0 +1,276 @@
...
+/**
+ * A Scroller Custom HTML Element
+ * @public
+ */
+export class Scroller extends FASTElement {
+    /**
+     * Reference to DOM element containing the content to scroll
+     * @public
+     */
+    public content: HTMLDivElement;
```
bheston, line 34: Might these be any html element? Is it possible to allow the template to be more flexible?

packages/web-components/fast-foundation/src/scroller/scroller.ts
bheston: Would this be worth extracting to some sort of provider for reuse?

bheston (commented):
Regarding the naming, I think the thing I was most looking for was that nature of the way it scrolls by content item rather than free scrolling like a scrollbar. "Scroller" in and of itself doesn't say that to me, but I'm not sure if "item-scroller" helps. "Item viewport" perhaps? "Scroller" is also fine if we don't find something more descriptive. It will be clear from sample usage.

packages/web-components/fast-foundation/src/scroller/scroller.ts
robarbms: I was too ambiguous here. This isn't the actual content. This is a ref used in the template, it is the content container. I'll update to make it more clear.

packages/web-components/fast-components/src/scroller/scroller.definition.ts
robarbms: Updated to show the values.

packages/web-components/fast-foundation/src/scroller/scroller.template.ts
```diff
@@ -0,0 +1,37 @@
+import { html, ref } from "@microsoft/fast-element";
+import { endTemplate, startTemplate } from "../patterns";
+import { Scroller } from "./scroller";
+
+/**
+ * @public
+ */
+export const ControlsTemplate = html<Scroller>`
+    <div ${ref("previousFlipper")} class="scroll scroll-prev">
+        <div @click="${x => x.scrollToPrevious()}" class="scroll-cntrl">
```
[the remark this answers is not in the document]
robarbms: This was mostly boilerplate code that I wrote out originally. I personally am not a fan of the term control as it's used in too many contexts. I've update these to scroll-action.

packages/web-components/fast-foundation/src/scroller/scroller.ts
robarbms: Probably a good idea. I "borrowed" this from @scomea 's work for AnchoredRegion.

packages/web-components/fast-foundation/src/scroller/scroller.styles.ts
```diff
@@ -0,0 +1,123 @@
...
+        transition: transform var(--scroller-duration) var(--scroller-easing);
+        display: inline-block;
+    }
+
+    .scroll-content ::slotted(*) {
+        display: inline-block;
+        white-space: normal;
+        vertical-align: var(--scroller-align);
+    }
+`;
```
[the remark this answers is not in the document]
chrisdholt: This is a good catch - should be composed in fast-components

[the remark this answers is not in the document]
robarbms: I saw the pattern of putting styles in fast-components, but these styles are primarily functional, not presentational. So I was going to ask. I'll move them to fast-components

packages/web-components/fast-foundation/src/scroller/scroller.template.ts
```diff
@@ -0,0 +1,37 @@
...
+            <slot name="previous-flipper"></slot>
```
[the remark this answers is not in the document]
robarbms: I had considered having a default. It would be nice if it just came with a flipper rather than having to add what would likely be the most commonly used. But they may not want flippers at all. It may be controlled by an outside handler or script. Plus I wasn't sure if this was going to be a problem moving it from FAST to FluentUI and to it's final destination, Edge feeds.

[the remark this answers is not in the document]
SethDonohue: Ah yes, if there is potential that the implementation doesn't want flippers at all that could be an issue.
How often would this be the case?
If you add a part name the the previousFlipper div then they could just hide that if they don't want flippers.

I know for FluentUI there was talk of adding methods to allow for a different default to be added to the template, but not sure where that conversation left off.

packages/web-components/fast-foundation/src/scroller/scroller.styles.ts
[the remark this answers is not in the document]
robarbms: I moved the styles to fast-components but my tests started failing. I'm assuming we don't want fast-foundation to take a dependency on fast-components so I just copied the styles needed so the tests could run. I'm just not sure with best practices here.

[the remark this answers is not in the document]
SethDonohue: Hmm I am  not sure we have done much testing for styles related things so far, so your approach is probably best for now imho.

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
```diff
@@ -0,0 +1,257 @@
...
+    <fast-card>Card number 9</fast-card>
+    <fast-card>Card number 10</fast-card>
+    <fast-card>Card number 11</fast-card>
+    <fast-card>Card number 12</fast-card>
+    <fast-card>Card number 13</fast-card>
+    <fast-card>Card number 14</fast-card>
+    <fast-card>Card number 15</fast-card>
+    <fast-card>Card number 16</fast-card>
+</fast-horizontal-scroll>
+
```
SethDonohue, line 228: You may want to add an example where the cards have tab able content? I'd suggest a simple `<button>` in the first card and another one in the last card to see how that effects things.

In experimentation, I added a `<button>` to the first and last card of the default example.
Tabbing to the button on the last card seems to mess up the scroll behavior and gets things out of sync. The component may need to listen for focus events or something to get synced back up.

chrisdholt: I just tried to repro this by adding a button to the last card - all I notice is that it pulls the card into view and focusses on it. Nothing else is focusable, so shift tab isn't going to go back to anything (there's nothing to focus on but the storybook controls. With everything focusable, this brings them into focus but doesn't "smoothly" scroll, which I think is ok. @SethDonohue what is your expected behavior here?

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.definition.ts
```diff
@@ -0,0 +1,59 @@
...
+    version: 1,
+    tags: [
+        {
+            name: "fast-horizontal-scroll",
+            title: "HorizontalScroll",
+            description: "The FAST HorizontalScroll element",
+            attributes: [
+                {
+                    name: "view",
+                    description: "Default or mobile view",
```
chrisdholt, line 14:
```suggestion
                    title: "View",
                    description: "Default or mobile view",
```

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.stories.ts
```diff
@@ -0,0 +1,8 @@
+import HorizontalScrollTemplate from "./fixtures/horizontal-scroll.html";
+import "./index";
+
+export default {
+    title: "HorizontalScroll",
```
chrisdholt, line 5:
```suggestion
    title: "Horizontal Scroll",
```

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
SethDonohue: Did you try to scroll with the flippers after tabbing to that button?

chrisdholt: Sure, that is more clear. No, I did not try and do that. We really shouldn't have to listen for events here - that seems like it would prove costly. @robarbms take a look at Seth's note - anything we can do to ensure that this doesn't get too out of whack?

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
```diff
@@ -0,0 +1,123 @@
...
+    .scroll-prev {
+        right: auto;
+        left: 0;
+    }
+
+    .scroll-next:before {
+        left: auto;
+        right: 0;
+        background: linear-gradient(to right, transparent, var(--scroll-fade-next));
+    }
```
chrisdholt, line 14: Should there be an instance of scroll-previous as well for LTR?

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
robarbms: I can see how that might be a problem. I'm using an internal position property to track so that I don't have to touch the DOM every time. I'll explore checking current position each time a scroll is called. It's going to require a fair amount of refactoring as I think I'll need to calculate each "page" position on load rather than just individual item positions.

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
robarbms: .scroll:before in the base styles act as the .scroll-prev:before for LTR and .scroll-next:before for RTL. It's easier to think of this as .scroll.scroll-next:before which overrides some of the styles used by .scroll.scroll-prev:before. If that makes sense.

chrisdholt: Yeah, I missed this - that's all :)

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
robarbms: I've redone the code so that it uses scrollLeft rather than translate. This allows you to tab through items in the horizontal-scroll. The scripts now work off what ever scroll position they are currently at.

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
```diff
@@ -0,0 +1,119 @@
...
+        --scroll-align: middle;
+        display: block;
+        position: relative;
+    }
+
+    .scroll-view {
+        overflow-x: auto;
+    }
+
+    ::-webkit-scrollbar {
```
chrisdholt, line 105: @robarbms should we include the `-moz-` prefix for mozilla?

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Scrolls items to the left
+     * @public
+     */
+    public scrollToPrevious(): void {
+        const position: number = this.getScrollPosition();
+        const stops: Array<number> = this.scrollStops;
+        if (this.isRtl) stops.sort((a: number, b: number): number => b - a);
+        const current: number = stops.findIndex(
+            (stop: number, idx: number, ary: Array<number>): boolean =>
```
chrisdholt, line 259: nit: I think verbosity is preferred for index and array for readability.

chrisdholt (approved):

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
robarbms: Updated.

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
robarbms: Added scrollbar-width: none for FireFox

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
SethDonohue: Looks good!

SethDonohue (approved):

khamudom (approved):

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+    private setStops(): void {
+        this.width = this.offsetWidth;
+        let lastStop: number = 0;
+        const scrollItems: Array<any> = [].slice
+            .call(this.children)
+            /* filter out non-default slots */
+            .filter(
+                (el: Element): boolean =>
+                    !el.getAttribute("slot") || el.getAttribute("slot") === "default"
+            );
```
radium-v, line 179:
```suggestion
        /* filter out non-default slots */
        const scrollItems = Array.from(this.children).filter(
            el => !el.getAttribute("slot") || el.getAttribute("slot") === "default"
        ) as HTMLElement[];
```

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * Scrolls items to the left
+     * @public
+     */
+    public scrollToPrevious(): void {
+        const position: number = this.getScrollPosition();
+        const stops: Array<number> = this.scrollStops;
+        if (this.isRtl) stops.sort((a: number, b: number): number => b - a);
```
radium-v, line 257: Always use multi-line `if` statements with curly braces

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+                this.isRtl
+                    ? stop <= position
+                    : stop <= position &&
+                      (index === array.length - 1 || array[index + 1] > position)
+        );
+        const right: number = stops[current + 1];
+        const nextIndex: number =
+            stops.findIndex((stop: number): boolean =>
+                this.isRtl ? stop - this.width < right : stop + this.width > right
+            ) || 0;
```
radium-v, line 269: This is very confusing to read

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+     * @public
+     */
+    public scrollToNext(): void {
+        const position: number = this.getScrollPosition();
+        const current: number = this.scrollStops.findIndex(
+            (stop: number): boolean => stop === position
+        );
+        const outOfView: number = this.scrollStops.findIndex((stop: number): boolean =>
+            this.isRtl ? stop <= position - this.width : stop >= position + this.width
+        );
```
radium-v, line 287: There's a lot going on here too which should probably be refactored

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+                case "ease-in":
+                    return Math.pow(k, 1.675);
+                case "ease-out":
+                    return 1 - Math.pow(1 - k, 1.675);
+                case "ease-in-out":
+                    return 0.5 * (Math.sin((k - 0.5) * Math.PI) + 1);
+                default:
+                    return k;
+            }
+        };
```
radium-v, line 319: Can this function be converted into either a static method or extracted to be its own thing?

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+     */
+    @attr
+    private width: number;
+
+    /**
+     * Scroll stop positions between elements
+     * @internal
+     */
+    @attr
+    private scrollStops: Array<number>;
```
radium-v, line 85:
```suggestion
    private scrollStops: number[];
```

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -0,0 +1,363 @@
...
+        );
+    }
+
+    /**
+     * Handle resize events
+     * @internal
+     */
+    private handleResize = (entries: ResizeObserverEntry[]): void => {
+        entries.forEach((entry: ResizeObserverEntry) => {
+            if (entry.target === this) this.resized();
```
radium-v, line 162: Always use multi-line if statements with curly braces

radium-v (changes requested):

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Styles used for the flipper container and gradient fade
+ * @public
+ */
+export const ActionsStyles = css`
+    .scroll-area {
+        position: relative;
+    }
+
+    div.scroll-view {
```
radium-v, line 55: Does this need to be qualified?

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    ::-webkit-scrollbar {
+        display: none;
+    }
+
+    .content-container {
+        white-space: nowrap;
+        display: inline-block;
+        transform: translate3d(0, 0, 0);
```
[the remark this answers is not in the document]
robarbms: It was initially there when I was using translateX to force hardware acceleration. I removed it with a lot of other code during a cleanup pass. Yesterday morning, the flippers were not showing up for RTL mode. I started putting code back to see if something caused it. The problem went away and seemed to be unrelated to my cleanup. I could see items shifted 500px off 0 for some reason. This was missed during the re-cleanup.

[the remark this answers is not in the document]
robarbms: Just tested. The problem resurfaces when I remove this.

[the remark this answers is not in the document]
radium-v: It looks like it's used to force a composite render layer, which can be useful in some situations. Try adding this instead:

```
position: relative;
z-index: 0;
```

radium-v:
image


If the areas for the `.scroll-prev` and `.scroll-next` should be clickable, you can use `pointer-events: none` on the `.scroll` containers and `pointer-events: auto` on their child `.scroll-action` containers. That will allow you to click through the gradient when content is underneath while still allowing the flippers to be clickable as well.

radium-v:
If you click the flipper to move to the next section while it's already in motion, the movement seems to get confused.

https://user-images.githubusercontent.com/863023/110171791-17fbe680-7db1-11eb-8200-aa8789feafd5.mp4

chrisdholt:
> image
> 
> If the areas for the `.scroll-prev` and `.scroll-next` should be clickable, you can use `pointer-events: none` on the `.scroll` containers and `pointer-events: auto` on their child `.scroll-action` containers. That will allow you to click through the gradient when content is underneath while still allowing the flippers to be clickable as well.

We'll want to be careful here though considering the card or element underneath may be fully clickable.

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
robarbms: I needed this when I load these styles using $fastController.addStyles. Seems like it is prepending rather than appending styles maybe? They weren't overriding mobile styles when needed.

packages/web-components/fast-components/src/horizontal-scroll/horizontal-scroll.styles.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    .content-container ::slotted(*) {
+        display: inline-block;
```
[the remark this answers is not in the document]
robarbms: It causes some unexpected rendering of content of the child node. Inline-block seemed closer to expected rendering.

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
```diff
@@ -0,0 +1,301 @@
...
+        margin: 20px;
+    }
+
+    .right-gradient {
+        --scroll-fade-next: var(--background-color);
+    }
+
+    .both-gradient {
+        --scroll-fade-next: var(--background-color);
+        --scroll-fade-previous: var(--background-color);
```
litteredwitherrors, line 21: Are the terms "previous" and "next" too objective, considering in RTL that may mean the cards advance to the left?

litteredwitherrors (approved):
Probably not a big deal, but the main thing that stands out that I'm not sure about is the use of "next" and "previous" throughout the component. Perhaps "left" and "right" flippers might be better?

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
robarbms: I used the names next and previous to account for RTL. The next flipper will scroll left in LTR and scroll right in RTL. I was also matching the naming used in the fast-flipper control for consistency.

packages/web-components/fast-components/src/horizontal-scroll/fixtures/horizontal-scroll.html
```diff
@@ -0,0 +1,301 @@
+<h1>Horizontal scroll</h1>
+
+<style type="text/css">
+    fast-card {
+        width: 120px;
+        height: 200px;
+        margin: 5px;
```
chrisdholt, line 7: @robarbms @radium-v I'm just seeing that the margin is coming from elements themselves. Could we leverage the slotted selector here to apply this via CSS Property to any slotted item - with exception of the *last*?

robarbms: Added --scroll-item-spacing with a default of 5px.

radium-v (approved):

PR #4398, opened 2021-02-25
bug: downgrade playwright to v1.8.0 from v1.9.0

Ibrahimmaga:
# Description
Playwight test is failing likely related to version 1.9.0
image

downgrade playwright to version 1.8.0
## Motivation & context

Ensure cross-browser validation

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/.github/workflows/ci-weekly.yml
+++ b/.github/workflows/ci-weekly.yml
@@ -47,8 +47,7 @@ jobs:
 
     - name: Test Playwright
       run: |
-        cd packages/web-components/fast-components 
-        npm install playwright@v1.8.0
+        cd packages/web-components/fast-components
         yarn test-chromium:pw 
         yarn test-firefox:pw 
         yarn test-webkit:pw
@@ -114,8 +113,7 @@ jobs:
      
     - name: Test Playwright
       run: |
-        cd packages/web-components/fast-components 
-        npm install playwright@v1.8.0
+        cd packages/web-components/fast-components
         yarn test-chromium:pw 
         yarn test-firefox:pw 
         yarn test-webkit:pw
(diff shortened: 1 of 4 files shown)


awentzel (commented):
I notice in our package.json files we are using "^1.8.0".  This fix will work for the pipeline, but, what about the developer experience.  Is playwright breaking due to this bug for the local developer as well?  It seems it should be broken as well.  If that's true, then I would instead request pegging the package.json file to use "1.8.0" instead of "^1.8.0", and then we don't need the yaml pipeline file updates.

Ibrahimmaga:
Playwright is not breaking for local development because we are using version 1.8.  For a playwright to work on another machine, we need playwright to be install on that environment. So far it's been working. But I realized without specifying the version, playwright could install 1.9.0. On this PR we are forcing the installation of 1.8 which is the stable version.  I did a little research on playwright, and most team are having issue with 1.9.0.   My first approach was to use the package.json file, but that doesn't work with playwright.  We can always create issue to improve on playwright, but this is just a hot fix "moving from unstable to a stable version of playwright"

chrisdholt:
> Playwright is not breaking for local development because we are using version 1.8. For a playwright to work on another machine, we need playwright to be install on that environment. So far it's been working. But I realized without specifying the version, playwright could install 1.9.0. On this PR we are forcing the installation of 1.8 which is the stable version. I did a little research on playwright, and most team are having issue with 1.9.0. My first approach was to use the package.json file, but that doesn't work with playwright. We can always create issue to improve on playwright, but this is just a hot fix "moving from unstable to a stable version of playwright"

@Ibrahimmaga We are not locked on that version though, so a fresh clean and install could in fact give us 1.9.0. We are opted into semver and so if the issue is with 1.9.0 we should lock for both local and ci development. https://semver.org/

awentzel:
- New version released today for `1.9.1`.  
- Upgrade playwright to peg to the new patched version in package.json file for fast-components.  
- Remove Playwright from the package.json file at the root.

I believe this will work for us.

awentzel (approved):

chrisdholt (approved):

EisenbergEffect (approved):

PR #4402, opened 2021-02-28
fix: anchored region pointer events mk2

scomea:
# Description
Second crack at this after finding the previous attempt caused regressions (it wasn't checked in).  

## Motivation & context
Users are unable to change the anchored-region's pointer events setting with the current configuration.  This change allows authors to set that and opacity after initial placement.

This change also adds a "loaded" css class.

## Issue type checklist

## Process & policy checklist\


packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -878,6 +882,9 @@ export class AnchoredRegion extends FASTElement {
 
         if (!this.initialLayoutComplete) {
             this.initialLayoutComplete = true;
+            this.classList.toggle("loaded", true);
```
nicholasrice, line 885: Is there a reason you're not doing this work in the template itself?

scomea: That would be better, done.

robarbms (approved):

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
```diff
@@ -6,7 +6,7 @@ import { AnchoredRegion } from "./anchored-region";
  * @beta
  */
 export const AnchoredRegionTemplate = html<AnchoredRegion>`
-    <template>
+    <template class="${x => (x.initialLayoutComplete ? "loaded" : "")}">
         ${when(
             x => x.initialLayoutComplete,
```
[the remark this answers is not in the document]
scomea: Yes, still needed, children could override and still have pointer events on which could cause inappropriate mouse events if the region isn't placed yet. This also mitigates children than take focus on load which is a fairly common pattern in pop ups, if an element takes focus before it is placed there can be unwanted browser auto-scroll behaviors triggered.

PR #4412, opened 2021-03-03
fix: data-grid stores generated templates

scomea:
# Description
Store the templates to avoid creation functions being called more than once

## Motivation & context
Avoid creating unneccessary template instances.

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):
@scomea Technically, this change isn't needed because the decorator will run once and cache the created instance. The place where it tends to matter is when a template is used inside of another template. In that case you don't want to create the template within the template because every time the parent template is instantiated the child template will get re-defined. So, in that case you want to factor it out and re-use it. Let me know if this makes sense.

Nothing wrong with this refactor of course. But wanted to let you know that it wasn't technically needed.

scomea:
@EisenbergEffect Good to know.  I'm thinking the easiest general approach is to prefer passing the template unless there is a specific reason to pass the function.

scomea:
@EisenbergEffect I think the row/cell commit I added fits the template within a template scenario?

EisenbergEffect:
Yes. Definitely. That second set of updates is the scenario where it matters. Spot on.

chrisdholt (approved):

SethDonohue (approved):

robarbms (approved):

PR #4433, opened 2021-03-08
feat: add auto-update attribute for anchored region

scomea:
# Description

Normally anchored region will only recalculate placement based on either attribute/prop changes, a resize event on the anchor element, or a call to update().  With this change we add an "auto-update" attribute to anchored region that allows for an additional mode - 'auto', which  will prompt the anchored region to additionally check it's placement when the window resizes, the viewport resizes and on any scroll event on the page.

## Motivation & context
Needed to support document level scaling menus.

## Issue type checklist

- [c] I have modified an existing component

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
@@ -1,5 +1,10 @@
 import { attr, DOM, FASTElement, observable } from "@microsoft/fast-element";
-import { Direction, eventVisibilityChange } from "@microsoft/fast-web-utilities";
+import {
+    Direction,
+    eventVisibilityChange,
+    eventScroll,
+    eventResize,
+} from "@microsoft/fast-web-utilities";
 import { getDirection } from "../utilities";
 import { IntersectionService } from "./intersection-service";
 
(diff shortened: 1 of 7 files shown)


SethDonohue:
Think you can add any tests for this? Not sure if playwright would make it any easier now.

SethDonohue (approved):

scomea:
@SethDonohue Yes, I need to devote a chunk of time very soon to see if that's doable with the new system, it should be.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -370,13 +412,20 @@ export class AnchoredRegion extends FASTElement {
     private pendingReset: boolean = false;
     private currentDirection: Direction = Direction.ltr;
 
+    // defines how big a difference in pixels there must be between states to
```
nicholasrice, line 415: Would exposing this as a public property be useful to app authors?

nicholasrice (approved):

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
scomea: perhaps, but my thinking is that I'd like to delay adding specific api's to control this until we see a need

awentzel (approved):

PR #4438, opened 2021-03-09
feat: add expand/collapse icon to menu items

scomea:
# Description
Ads expand and collapse icons to menu items.

## Motivation & context
An expected feature of nested menus.

## Issue type checklist

## Process & policy checklist


packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -42,14 +42,12 @@ export class MenuItem extends FASTElement {
...
-            const submenu: Element | undefined = this.getSubmenu();
-            if (submenu === undefined) {
+            if (this.submenu === undefined) {
                 return;
             }
             if (this.expanded === false) {
-                (submenu as Menu).collapseExpandedItem();
+                (this.submenu as Menu).collapseExpandedItem();
             } else {
                 this.currentDirection = getDirection(this);
```
robarbms, line 52: I haven't used getDirection so I am guessing at the purpose here. Is it to inherit direction from the host menu component? From a performance standpoint it makes sense to only set it if the submenu is open. But with the indicator having RTL specific styles, would this need to be called in connectedCallback now?

robarbms (changes requested):

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -83,17 +81,18 @@ export class MenuItem extends FASTElement {
     }
 
     /**
+     * reference to the anchored region
+     *
      * @internal
      */
     @observable
-    public submenu: Element | undefined;
+    public submenuRegion: AnchoredRegion;
```
chrisdholt, line 89: @scomea is this a breaking change?

scomea: It's marked internal so I'd say that means folks shouldn't be taking dependencies on it, but it is an unnecessary change so I'll revert it to minimize churn

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
scomea: No, we get is specifically to populate the dir attribute of the anchored region when it is created.  All the css layout stuff just works.

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -254,11 +259,13 @@ export class MenuItem extends FASTElement {
      *
      * @internal
      */
-    private getSubmenu = (): undefined | Element => {
+    private updateSubmenu = (): void => {
         const domChildren: Element[] = Array.from(this.children);
```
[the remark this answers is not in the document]
scomea: yep

chrisdholt (approved):

robarbms (approved):

PR #4441, opened 2021-03-09
feat: add gestures utility

mlijanto:
# Description

Initial merge to add PointerGestureHandler, a utility library for device-agnostic gesture detection.

## Issue type checklist

## Process & policy checklist


packages/utilities/fast-gestures/src/utilities/get-rotation-angle.ts
```diff
@@ -0,0 +1,4 @@
+export function getRotationAngle(x1: number, y1: number, x2: number, y2: number): number {
+  const angle: number = (Math.atan2(y1 - y2, x1 - x2) * 180) / Math.PI + 180;
```
nicholasrice, line 2: Minor nit; I don't think you need to allocate a variable here

packages/utilities/fast-gestures/src/utilities/get-coordinate-type.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    case PointerCoordinateType.Page:
+      return { x: e.pageX, y: e.pageY };
+
+    case PointerCoordinateType.Offset:
+      return { x: e.offsetX, y: e.offsetY };
+
+    case PointerCoordinateType.Screen:
+      return { x: e.screenX, y: e.screenY };
+
```
nicholasrice, line 17: `PointerCoordinateType.Page` and `default` can be combined

packages/utilities/fast-gestures/docs/api-report.md
```diff
@@ -0,0 +1,49 @@
...
+//
+// @public (undocumented)
+export type IPointerGestureConfig = Partial<Record<PointerGestureFeature, boolean>>;
+
+// Warning: (ae-forgotten-export) The symbol "IPointerGestureInfo" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type PointerGestureCallback = (pointerGestureInfo: IPointerGestureInfo) => void | null;
+
+// Warning: (ae-forgotten-export) The symbol "Disposable" needs to be exported by the entry point index.d.ts
```
nicholasrice, line 27: Can you make sure these ae-forgotten-export subjects are exported?

nicholasrice:
Awesome, thanks @mlijanto! This looks great. I'm noticing a break in the build when looking for the entry `/src/index.ts` file, do you know what that is about? 

https://github.com/microsoft/fast/pull/4441/checks?check_run_id=2079900033

packages/utilities/fast-gestures/src/interfaces/pointer-gesture-info.ts
```diff
@@ -0,0 +1,7 @@
+import { PointerTarget } from "../pointer-target";
+import { IPointerInfo } from "./pointer-info";
+
+export interface IPointerGestureInfo {
```
chrisdholt, line 4: @mlijanto can we drop "I" as a prefix to the interfaces across the PR?

packages/utilities/fast-gestures/docs/api-report.md
mlijanto: Sure. I've addressed all the `ae-forgotten-export` warnings. Thanks!

mlijanto:
> Awesome, thanks @mlijanto! This looks great. I'm noticing a break in the build when looking for the entry `/src/index.ts` file, do you know what that is about?
> 
> https://github.com/microsoft/fast/pull/4441/checks?check_run_id=2079900033

Thanks for the review, @nicholasrice! Looks like the build break is caused by the empty `@microsoft/fast-physics` package. It can't find an entry point for the package. Is there a way we can skip building that package for now?

packages/utilities/fast-gestures/src/interfaces/pointer-gesture-info.ts
mlijanto: @chrisdholt I've removed the "I" prefix from most of the interfaces, but there are some cases where removing the prefix would create a conflicting name between the interface and the class that implements it. Could I leave the "I" prefix in these cases, or do you have any guidance or pattern I should follow?

chrisdholt: Can you point to those scenarios? I see similar names but I don't see any with conflicts - I could be missing it.

mlijanto: There's a few of them, such as `IPoint` and `Point`, `IEvent` and `Event`. But this is no longer an issue. Rob and Nick gave a guidance to add the prefix "Default" to the implementations so I could remove the "I" prefix from all the interfaces. Thanks!

awentzel:
It looks like this is ready to be completed.  Pinging for review.

nicholasrice (approved):

mlijanto:
> It looks like this is ready to be completed. Pinging for review.

There's still a couple of more works need to be done after our last discussion with Rob and Nick. I'll need to replace the dependencies on codes taken from the VSCode project with our own and write some more unit tests.

EisenbergEffect:
@mlijanto Where are we at on this? Is this ready for a review/merge/publish? Or does it need more work?

PR #4458, opened 2021-03-12
fix: correct data grid header cell tag name generation

chrisdholt:
# Description
When using the foundation component to create a custom data grid implementation, it was noticed that the header cells were still using `fast` prefixes rather than `foo`. I believe this was caused by an incorrect name in the opening tag for the function which generates the prefixed template.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/data-grid/data-grid-row.template.ts
+++ b/packages/web-components/fast-foundation/src/data-grid/data-grid-row.template.ts
@@ -14,7 +14,7 @@ function createCellItemTemplate(prefix: string): ViewTemplate {
 
 function createHeaderCellItemTemplate(prefix: string): ViewTemplate {
     return html`
-    <${prefix}-data-grid-cell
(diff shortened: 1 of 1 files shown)


EisenbergEffect (approved):

packages/web-components/fast-foundation/src/data-grid/data-grid-row.template.ts
```diff
@@ -14,7 +14,7 @@ function createCellItemTemplate(prefix: string): ViewTemplate {
 
 function createHeaderCellItemTemplate(prefix: string): ViewTemplate {
     return html`
-    <${prefix}-data-grid-cell
+    <${prefix}-data-grid-header-cell
```
scomea, line 17: This should be the opposite, remove the "header" from the closing tag.  I'm surprised this didn't cause issues but in our samples the first tag "wins".

scomea (changes requested):

PR #4461, opened 2021-03-13
fix: combobox interaction problems in firefox

radium-v:
# Description

Both of these issues were present in Firefox:
* Clicking the combobox causes the mouse to enter a broken "text selection" mode
* Tabbing through the combobox causes both the combobox and its text input to receive focus

## Motivation & context
These interaction quirks are only present in Firefox.

To reproduce:

"text selection" bug:
1. Click a combobox
2. Click on a listbox option
3. Without clicking again, hover over the text input for the combobox.

Expected: Moving the mouse cursor over the text input without the mouse button held down should not select text in the text input.
Actual: Content in the text input is selected from the position of the caret to the position of the mouse cursor.

"tabbing" bug:
1. Tab to the combobox
2. Press tab again once focused on the combobox

Expected: Focus should move away from the combobox.
Actual: Focus moves to the text input inside the combobox.

## Issue type checklist

## Process & policy checklist


scomea (approved):

packages/web-components/fast-components/src/combobox/combobox.styles.ts
```diff
@@ -15,10 +15,6 @@ export const ComboboxStyles = css`
         user-select: none;
     }
 
-    :host(:active) .selected-value {
-        user-select: none;
```
scomea, line 19: I wonder if it makes more sense setting "functional" styles like this should be done in foundation (ie. in component or template) vs. in the styled component packages.

radium-v: The block here still needs to be removed, but that would probably make sense for the block above on line 15.

nicholasrice (approved):

PR #4472, opened 2021-03-16
chore: add beachball changes to test pipeline failures

awentzel:
Continuous Publishing with Beachball 
# Description
Update to test why publish yml file is not being picked up by GH Actions
package owners who want to automate the bumping of versions based on change files with beachball, you'll need to provide some information for your Continuous Integration (CI) systems. These one-time setup steps will be unique for different CI system naturally, but the general idea remain the same. beachball publish needs write access to with are the git repo and npm registry.

## Motivation & context
Nightly publish to npm registry

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -27,7 +27,6 @@ jobs:
       run: |
         git config user.name ${{ env.GITHUB_PUBLISH_USER }}
         git config user.email ${{ env.GITHUB_PUBLISH_EMAIL }}
-        git remote set-url origin https://${{env.GITHUB_SERVICE_PAT}}:${{env.GITHUB_SERVICE_PAT}}@github.com/microsoft/fast.git
     
     - name: Install package dependencies / prepare workspaces
       run: yarn install --frozen-lockfile
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -14,7 +14,7 @@ jobs:
     env:
       NPM_PAT: ${{ secrets.PACK_TOKEN }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      GITHUB_SERVICE_PAT: ${{ secrets.GH_PAT }}
+      GITHUB_PAT: ${{ secrets.GH_PAT }}
       GITHUB_SERVICE_USER: "Microsoft FAST Builds"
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-nightly.yml
```diff
@@ -28,19 +28,23 @@ jobs:
       run: |
         git config --global user.name "${{ env.GITHUB_SERVICE_USER }}"
         git config --global user.email "${{ env.GITHUB_SERVICE_EMAIL }}"
-    
+   
     - name: Install package dependencies / prepare workspaces
       run: yarn install --frozen-lockfile
+    
+    - name: check for change
```
chrisdholt, line 35: This might be best as "check change" as this step is ensuring there is a change file - if not, we'd exit right away.

```suggestion
    - name: check change
```

.github/workflows/ci-nightly.yml
```diff
@@ -28,19 +28,23 @@ jobs:
...
+      run: yarn check
 
     - name: Testing Prettier format
       run: lerna run prettier:diff
 
     - name: Testing unit tests
       run: lerna run test --stream   
-      
+
+    - name: look for change file
```
chrisdholt, line 44: This one is probably more semantic and helps to differentiate with "check for changed files" as we want to exit if there are *now* generated files.
```suggestion
    - name: check for changed files
```

chrisdholt: Looks like I was mistaken on what's being executed here. This isn't running a beachball script, the script in question is actually checking git status...https://github.com/microsoft/fluentui/blob/master/scripts/tasks/check-for-modified-files.ts

Ibrahimmaga: I agree. I just remove it. I think we need to check the change in daily build though.

chrisdholt: Yep - the PR instance should have a single check after yarn gets dependencies installed.

.github/workflows/ci-nightly.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
+      NPM_PAT: ${{ secrets.PACK_TOKEN }}
```
awentzel, line 15: PACK_TOKEN should be reverted back to the correct me for our secret npm token name.

awentzel:
@Ibrahimmaga why do we now have two PRs for Beachball?

awentzel:
aborting this one to use the latest #4487

PR #4480, opened 2021-03-19
fix: update react-dev-utils to 11.0.4 & is-svg@^4.2.2:

Ibrahimmaga:
Dependabot cannot update react-dev-utils to a non-vulnerable version
# Description
Upgrade react-dev-utils to version 11.0.4 and is-svg@^4.2.2:or later. For example:

react-dev-utils@^11.0.4:
  version "11.0.4"

is-svg@^4.2.2
this module performs a quick-and-dirty check. It's fast, but in certain cases it will give incorrect results.
Returns true for an SVG-like string that isn't well-formed or valid: <svg><div></svg>
If you want to make certain that your SVG is valid, try parsing it with libxmljs.

Vulnerable versions: >= 0.4.0, < 11.0.4
Patched version: 11.0.4
## Motivation & context
resolve dependabot alerts

## Issue type checklist

## Process & policy checklist


janechu (approved):

janechu:
Bug fixes should be preceded by "fix" not "bug" and should be lowercase. I've corrected the PR title to reflect.

nicholasrice (approved):

awentzel:
Shouldn't there be package bumps also in package.json file or files?

Ibrahimmaga:
> Shouldn't there be package bumps also in package.json file or files?

we are not using beachball on this PR.

awentzel:
> > Shouldn't there be package bumps also in package.json file or files?
> 
> we are not using beachball on this PR.

What I mean is, generally when upgrading dependencies, which are packages in themselves, that would be done on package.json file and then by running yarn install those dependencies would update the yarn.lock file which would then get committed.  I don't see any dependencies being updated with this PR.  So, how did yarn.lock then get created?

Ibrahimmaga:
> > > Shouldn't there be package bumps also in package.json file or files?
> > 
> > 
> > we are not using beachball on this PR.
> 
> What I mean is, generally when upgrading dependencies, which are packages in themselves, that would be done on package.json file and then by running yarn install those dependencies would update the yarn.lock file which would then get committed. I don't see any dependencies being updated with this PR. So, how did yarn.lock then get created?

I can't see. I'm sure I have updated package.json let me double check

Ibrahimmaga:
> > > > Shouldn't there be package bumps also in package.json file or files?
> > > 
> > > 
> > > we are not using beachball on this PR.
> > 
> > 
> > What I mean is, generally when upgrading dependencies, which are packages in themselves, that would be done on package.json file and then by running yarn install those dependencies would update the yarn.lock file which would then get committed. I don't see any dependencies being updated with this PR. So, how did yarn.lock then get created?
> 
> I can't see. I'm sure I have updated package.json let me double check

both of them are not in package.json. perhaps we don't need it there. I usually do the global search but they are not in package.son. They are no in master neither.

PR #4487, opened 2021-03-23
chore: update beachball configuration and testing

Ibrahimmaga:
Beachball testing Scenarios
# Description
The testing will provide end to end testing with beachball

1. test for non change file
2. test for files that can trigger beachball
3. check for changes in package.json
4. ci-workflow

Adds and updates beachball package and configuration to project to handle the publishing of NPM public node packages on a nightly cadence through the CLI.

Files reside in `./build/releasing/*`, `package.json`, and `.github/workflows/ci-weekly.yml`

# Testing locally
Using scripts as noted in `package.json`  (@awentzel to test)
[x] 1. change any files and stage the commit  (@awentzel to test)
[x] 2. run the commands for `yarn change`, `yarn check`.  (@awentzel to test)
[x] 3. run the commands for `yarn publish` and `yarn public-ci` (@awentzel to test)
[] 3. run the commands for `yarn publish` and `yarn public-ci` (@chrisdholt   to test)

# Testing CI
[x] 1. test connection to GitHub API using PAT  (@awentzel to test)
[x] 2. test run tasks passes until NPM publish task  (@awentzel to test)
[ ] 3. change `ci-nightly.yml` node task to match that in package.json `yarn publish-ci` (@chrisdholt  to test)
[ ] 3. change `ci-nightly.yml` to use `on` `push` to master and on master branch (@chrisdholt  to test)

#Final CI test and PR change prior to merging
[ ] 1. change `ci-nightly.yml` to use `on` cron for master branch then commit this PR  (@chrisdholt  to test)

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/.github/workflows/ci-daily.yml
+++ b/.github/workflows/ci-daily.yml
@@ -37,6 +37,9 @@ jobs:
     - name: Install package dependencies
       run: yarn install --frozen-lockfile --ignore-scripts --prefer-offline
 
+    - name: Check for file changes
+      run: yarn checkchange
+
     - name: Testing Prettier format
       run: yarn format:check
 
--- a/.github/workflows/ci-daily.yml
+++ b/.github/workflows/ci-daily.yml
@@ -37,7 +37,7 @@ jobs:
     - name: Install package dependencies
       run: yarn install --frozen-lockfile --ignore-scripts --prefer-offline
 
-    - name: Check for file changes
+    - name: Check for the presence of changed files inside ./change
       run: yarn checkchange
 
     - name: Testing Prettier format
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -27,7 +27,6 @@ jobs:
       run: |
         git config user.name ${{ env.GITHUB_PUBLISH_USER }}
         git config user.email ${{ env.GITHUB_PUBLISH_EMAIL }}
-        git remote set-url origin https://${{env.GITHUB_SERVICE_PAT}}:${{env.GITHUB_SERVICE_PAT}}@github.com/microsoft/fast.git
     
     - name: Install package dependencies / prepare workspaces
       run: yarn install --frozen-lockfile
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -14,7 +14,7 @@ jobs:
     env:
       NPM_PAT: ${{ secrets.PACK_TOKEN }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      GITHUB_SERVICE_PAT: ${{ secrets.GH_PAT }}
+      GITHUB_PAT: ${{ secrets.GH_PAT }}
       GITHUB_SERVICE_USER: "Microsoft FAST Builds"
       GITHUB_SERVICE_EMAIL: "fastdna@service.microsoft.com"
 
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -40,6 +40,7 @@ jobs:
       
     - name: Publish NPM changed packages
       run: |
+        yarn add beachball -W
         yarn publish-ci
       # git reset --hard origin/master
       # npm login -n ${{ env.NPM_PAT }} --access public -y
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
+      NPM_PAT: ${{ secrets.PACK_TOKEN }}
       GITHUB_SERVICE_PAT: ${{ secrets.GH_PAT }}
       GITHUB_SERVICE_USER: "Microsoft FAST Builds"
       GITHUB_SERVICE_EMAIL: "fastdna@service.microsoft.com"
@@ -26,7 +27,7 @@ jobs:
       run: |
         git config user.name ${{ env.GITHUB_PUBLISH_USER }}
         git config user.email ${{ env.GITHUB_PUBLISH_EMAIL }}
-        git remote set-url origin https://${{GITHUB_SERVICE_PAT}}:${{env.GITHUB_SERVICE_PAT}}@github.com/microsoft/fast.git
+        git remote set-url origin https://${{env.GITHUB_SERVICE_PAT}}:${{env.GITHUB_SERVICE_PAT}}@github.com/microsoft/fast.git
     
     - name: Install package dependencies / prepare workspaces
       run: yarn install --frozen-lockfile
@@ -39,6 +40,6 @@ jobs:
       
     - name: Publish NPM changed packages
       run: |
-        npm run beachball-publish -- -b origin/master -n $(npmToken) --access public -y
+        npm run beachball-publish -- -b origin/master -n ${{ env.NPM_PAT }} --access public -y
         git reset --hard origin/master
     
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -13,6 +13,7 @@ jobs:
 
     env:
       NPM_PAT: ${{ secrets.PACK_TOKEN }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GITHUB_SERVICE_PAT: ${{ secrets.GH_PAT }}
       GITHUB_SERVICE_USER: "Microsoft FAST Builds"
       GITHUB_SERVICE_EMAIL: "fastdna@service.microsoft.com"
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -11,6 +11,13 @@ jobs:
   build_linux:
     runs-on: ubuntu-latest
 
+    env:
+      NPM_PAT: ${{ secrets.PACK_TOKEN }}
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_SERVICE_PAT: ${{ secrets.GH_PAT }}
+      GITHUB_SERVICE_USER: "Microsoft FAST Builds"
+      GITHUB_SERVICE_EMAIL: "fastdna@service.microsoft.com"
+
     steps:
     - uses: actions/checkout@v2
 
@@ -19,8 +26,8 @@ jobs:
 
     - name: Authenticate Git for Pushes
       run: |
-        git config user.name ${{ env.GITHUB_PUBLISH_USER }}
-        git config user.email ${{ env.GITHUB_PUBLISH_EMAIL }}
+        git config user.name ${{ env.GITHUB_SERVICE_USER }}
+        git config user.email ${{ env.GITHUB_SERVICE_EMAIL }}
     
     - name: Install package dependencies / prepare workspaces
       run: yarn install --frozen-lockfile
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -26,8 +26,8 @@ jobs:
 
     - name: Authenticate Git for Pushes
       run: |
-        git config user.name ${{ env.GITHUB_SERVICE_USER }}
-        git config user.email ${{ env.GITHUB_SERVICE_EMAIL }}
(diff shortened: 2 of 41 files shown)


nicholasrice:
@Ibrahimmaga is this PR ready for review? If not, could you convert it to a draft PR so folks know to ignore it for now?

Ibrahimmaga:
> @Ibrahimmaga is this PR ready for review? If not, could you convert it to a draft PR so folks know to ignore it for now?

yes you can review

change/@microsoft-fast-components-9902ce21-7d35-46de-8c68-28f33a5e69de.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "add 4th option for template tittle",
```
chrisdholt, line 3: nit: title* (spelling).

change/@microsoft-fast-components-9902ce21-7d35-46de-8c68-28f33a5e69de.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
chrisdholt, line 2: Is this an example or a test? If not, I think we'll want this to be "none"

Ibrahimmaga: Just for testing purpose

nicholasrice:
@Ibrahimmaga okay great, thanks! Can you update the PR title to align to our PR naming conventions?

.github/workflows/ci-daily.yml
```diff
@@ -37,6 +37,9 @@ jobs:
     - name: Install package dependencies
       run: yarn install --frozen-lockfile --ignore-scripts --prefer-offline
 
+    - name: Check for file changes
```
chrisdholt, line 40:
```suggestion
    - name: Check for change files
```
A bit of a nit, but I think we should be specific here. This checks for the presence of "change" files for the purpose of ensuring we have metadata necessary to generate a changelog and version correctly. So, not technically "file changes" as the PR has that *always*, we really want to look for the specific files.

package.json
```diff
@@ -39,9 +39,10 @@
   "scripts": {
     "bump": "beachball bump",
     "change": "beachball change",
-    "check": "beachball check",
+    "checkchange": "beachball check  --scope \"!packages/*\" --changehint \"Run 'yarn change' to generate a change file\"",
+    "check": "beachball check ",
     "publish": "beachball publish",
-    "publish-ci": "beachball publish -n $(npm.token)",
+    "publish-ci": "beachball publish -b origin/master",
```
nicholasrice, line 45: Does beachball actually make commits to master like Lerna does?

Ibrahimmaga: I have not used Learna before. After you fill out the scope for file changed, it will automatically commit. But always check with git status to make sure.

package.json
```diff
@@ -39,9 +39,10 @@
   "scripts": {
     "bump": "beachball bump",
     "change": "beachball change",
-    "check": "beachball check",
+    "checkchange": "beachball check  --scope \"!packages/*\" --changehint \"Run 'yarn change' to generate a change file\"",
```
chrisdholt, line 42: I'm checking, but I think we'll want to remove the `!` before "packages". I *think* that is saying, "not that folder".

package.json
chrisdholt: @nicholasrice this link presents the workflow in detail: https://microsoft.github.io/beachball/overview/getting-started.html

package.json
chrisdholt: Yeah, I think we want to have `\"packages/*\"`: https://github.com/microsoft/beachball/blob/master/src/monorepo/getScopedPackages.ts#L12

Ibrahimmaga: makes sense

change/@microsoft-fast-components-9902ce21-7d35-46de-8c68-28f33a5e69de.json
chrisdholt: Cool - I'm going to "hold" while we test just to prevent this from being merged *with this in*. I think once we can confirm this is working as expected we'll need to publish w/ Lerna, merge this immediately after, and that will start enforcing these files. 👍

chrisdholt (changes requested):
Holding on this while we test to ensure that the "test" change files don't accidentally get checked in :)

It won't get checked. bug good to block just in case.

Ibrahimmaga:
> Are we at a point where we should set up a meeting with the team to go over procedures? Also, I think there was an old PR with documentation updates related to this...

Yes. I will go over this with Aaron in a few minutes, then we will update you

package.json
nicholasrice: @chrisdholt So yes? What I'm trying to understand is if there is a race condition between someone manually merging a PR prior and beachball's publishing commit.

chrisdholt: I think there could be, though it seems like that's acknowledged and they try to account for that discrepancy (sorry for another link): https://microsoft.github.io/beachball/cli/publish.html

> In most version bumping systems that automates git repo version with npm registry, they assume that the source code is fresh by the time they push the changes back to the git repository. This is almost never the case when the build takes more than a few minutes! So, beachball fetches latest before we push back to the target branch. In large systems, it has been observed that without the git fetch, it becomes a source of conflict.

FWIW (which may be very little), we've been using this on the Fluent UI side with both automated nightly publishes as well as manual publishes during the day. We haven't hit an issue with those *yet*...but I acknowledge, *yet*.

nicholasrice: Gotcha okay, well hopefully in practice it won't be an issue.

chrisdholt: @nicholasrice totally agree :). I do think there are a couple things which limit the risk (though there isn't inherently *no risk*):

1. Only contributors with write access have merge permissions. We can manage through policy in terms of *manual* publishes...if we we need to manually publish we ask folks to hold PR's.

2. We setup the nightly publish at a time which is *inconvenient* for merging anyways. Perhaps something like 2AM PST (5AM ET), etc...

chrisdholt: For 2, I think folks who *can merge* would be aware of this and ideally try and avoid merging at that time

chrisdholt: Lastly (I hope), If we see a manual publish has already been kicked off - we don't run it again while it's already going. 

I think those simple guidelines would help avoid the issue

.github/workflows/ci-daily.yml
awentzel: We could also include "Check for the presence of changed files inside `./change`".  The entire detail should be added to documentation we wrote today @Ibrahimmaga.

.github/workflows/ci-nightly.yml
```diff
@@ -1,6 +1,9 @@
 name: CI - FAST Publish
 
 on:
+  push:
+    branches:
+    
   pull_request:
```
awentzel, line 7: I would still prefer this run once nightly unless this would pose an issue.  How do you feel about this, @chrisdholt?   If we're all in agreement, then we should replace the `pull_request` event with a cron schedule and only for the master branch.

I do think that in our `ci-daily.yml` we can include the `yarn checkchange` to provide indication for contributors that more work is required before pushing to master to prepare for publishing.

awentzel (commented):
Please add the documentation, plus any additional details as Chris has noted, to the https://www.fast.design/docs/community/contributor-guide for awareness.

awentzel:
@Ibrahimmaga please update this PR title to describe what change is being implemented instead of your branch name.  This is useful for when a changelog file is created, the community can scan and quickly understand how FAST is improving.

.github/workflows/ci-nightly.yml
```diff
@@ -12,6 +15,8 @@ jobs:
     runs-on: ubuntu-latest
 
     env:
+      NPM_PAT: ${{ secrets.PACK_TOKEN }}
```
awentzel, line 18: PACK_TOKEN is not correct, use the one as shown for secret npm.

.github/workflows/ci-nightly.yml
chrisdholt: I think I'm tracking - I would expect publish to *only* run on a cron with a method of manually triggering the build in emergencies or urgent cases only. *Ideally* a manual deployment would be triggered via an ADO build agent behind permissions or something similar so that it doesn't require any individual to run it from their own CLI.

+ @EisenbergEffect if he has thoughts as well

awentzel (approved):

.github/workflows/ci-nightly.yml
awentzel: @chrisdholt, confirmed, we will be converting this to use Cron as soon as it's approved and as a final step before completing the PR.  We can put this in now if you want?

chrisdholt: @awentzel let's include as part of this PR if we can. 

Once this goes in, every current/existing PR which includes files under `/packages/*` *should* fail at `yarn checkchange`. Once this is ready, we should pause merging and publish. Immediately after the publish is complete, we should merge this in. That kicks off the official shift from publishing w/ Lerna to publishiing w/ Beachball.

EisenbergEffect: My original thought was to use NPM release channels.

* Canary Channel: Either triggered on every merge to master or by a nightly cron job. I don't have a strong opinion on which one. Should be at least once per day though.
* Beta Channel: Released once per week via a cron job.
* Stable Channel: Manually released by Chris, myself or other DRI at a planned time, typically in conjunction with messaging (e.g. blog post, tweets, conference, etc.).

Here's some information: https://github.com/danillouz/til-npm-release-channels
I think the biggest question was how to make this work with beachball.

packages/web-components/fast-components/src/listbox/scenarios/index.html
```diff
@@ -3,5 +3,6 @@
         <fast-option value="1">Option 1</fast-option>
         <fast-option value="2">Option 2</fast-option>
         <fast-option value="3">Option 3</fast-option>
```
nicholasrice, line 5: is this change related?

nicholasrice: Ah nm I see the change test

Ibrahimmaga: yes. we can leave the change if that okay. I added option 4 or just remove it. I believe no harm was done on this.

.github/workflows/ci-nightly.yml
Ibrahimmaga: right now it is set to trigger 1 publish per night on cron job. Chris will know a little more.

package.json
```diff
@@ -39,9 +39,10 @@
   "scripts": {
     "bump": "beachball bump",
     "change": "beachball change",
-    "check": "beachball check",
+    "checkchange": "beachball check  --scope \"packages/*\" --changehint \"Run 'yarn change' to generate a change file\"",
```
chrisdholt, line 42: I was playing around with this locally - I think what we want is below. It should check for the presence of change files in all packages but not the sites, which would be what we want IMO. If there are changes *only* to sites, we don't need change files. Any changes under packages AND those would require them.

```suggestion
    "checkchange": "beachball check --scope \"!sites/*\" --changehint \"Run 'yarn change' to generate a change file\"",
```

chrisdholt: For some reason passing in packages/ explicitly doesn't work, so I the above is correct.

.github/workflows/ci-nightly.yml
chrisdholt: I think we'll want to ensure this meets Rob's expectations as well. We definitely can use a nightly publish. If we need to release those on the NPM cadence, I'm not sure this work accommodates that.

awentzel: I'd prefer to first get our publishing working and testing as an automated replacement of today's manual process.  Once that's in place and verified for a week or so we can most certainly discuss NPM channels.

- We may require contributing to beachball to get semver 2 meta data feature which to my knowledge has not been completed, such that we can properly tag the channel per previous GH Issues
- We may want to wait for GH/NPM Release capability to update.

In either case, that would be a good next step to discuss after we get the base automation in place.

EisenbergEffect: Sounds good. Just want to make sure the goal is clear.

Issue #3953, opened 2020-09-23
bug: components should call preventDefault on keyboard actions which have caused a ui change


scomea:
We should do a pass to ensure that when we use a keyboard event to trigger a ui change we should mark that event with preventDefault() so that upstream consumers of bubbled events can avoid triggering a second ui change on the same event.

For example key handling in tree view item does not mark the event arrowLeft event when it triggers ui changes:

```js

    public handleKeyDown = (e: KeyboardEvent): void | boolean => {
        if (e.target !== e.currentTarget) {
            return true;
        }

        switch (e.keyCode) {
            case keyCodeArrowLeft:
                this.handleArrowLeft();
                break;
```
If a parent element in the dom also captures the event it has no way to know not to act on it.  This is most obvious with how the browser reacts to some key strokes, like scrolling on space/home, but it can also affect js code where user keystrokes could inadvertantly trigger multiple changes in the ui at once.

tech4GT:
Hi @scomea is this issue still active? I would like to work on this 😄

scomea:
@tech4GT Sure.  The important thing to remember is that template bound events invoke prevent default on the event by default (https://www.fast.design/docs/fast-element/declaring-templates#events) unless the call returns true which can be confusing.  Looking at the current state of the handleKeyDown() handler in tree item it always returns true and preventDefault has been added to the up/down key event cases. The fix here would be to return false for any handled key event and true for unhandled keystrokes.  ie. Enter and arrow left/right should also be marked with preventDefault (return false)

tech4GT:
@scomea I see! That makes sense! So the solution here would be to have a default case within the switch which returns true rather than breaking out and returning true every single time, correct?

Something like this
```js
public handleKeyDown = (e: KeyboardEvent): void | boolean => {
        if (e.target !== e.currentTarget) {
            return true;
        }

        switch (e.keyCode) {
            case keyCodeArrowLeft:
                this.collapseOrFocusParent();
                break;
            case keyCodeArrowRight:
                this.expandOrFocusFirstChild();
                break;
            case keyCodeArrowDown:
                // preventDefault to ensure we don't scroll the page
                e.preventDefault();
                this.focusNextNode(1);
                break;
            case keyCodeArrowUp:
                // preventDefault to ensure we don't scroll the page
                e.preventDefault();
                this.focusNextNode(-1);
                break;
            case keyCodeEnter:
                // In single-select trees where selection does not follow focus (see note below),
[...]

PR #4489, opened 2021-03-24
fix: add return values to prevent default on handled keys in the keyDown handler within the tree item component

tech4GT:
Signed-off-by: tech4gt <varun.gupta1798@gmail.com>

# Description

## Motivation & context

#3953 

## Issue type checklist

## Process & policy checklist


scomea (approved):
We should be returning false whenever we use the event (arrow keys + enter) and false otherwise.  One way would be for each case except the default one to return false.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -183,22 +183,21 @@ export class TreeItem extends FASTElement {
...
                 // In single-select trees where selection does not follow focus (see note below),
                 // the default action is typically to select the focused node.
                 this.handleSelected(e);
                 break;
+            default:
+                return true;
         }
-
-        return true;
+        return false;
```
radium-v, line 200: `return false` prevents default, stops propagation, and stops any further function execution.

Since this is the end of the function anyway, it should either just `return` (void) or not have a final return statement.

tech4GT: Ah I thought of that, but ended up adding the statement anyway. I'll remove it! Thanks! :D

EisenbergEffect: Here's the piece of code you all want to look at: https://github.com/microsoft/fast/blob/master/packages/web-components/fast-element/src/templating/binding.ts#L368

radium-v: Well this is embarrassing 😳 the `return false` behavior I described is only relevant in the context of jQuery's `.on()` events, and may have been a thing for really old browsers (like, pre-2009). Returning false in a current-day event handler does nothing special. Either way, removing it here resolves exactly this sort of ambiguity, which is always good!

radium-v (approved):

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -183,22 +183,20 @@ export class TreeItem extends FASTElement {
                 break;
             case keyCodeArrowDown:
                 // preventDefault to ensure we don't scroll the page
-                e.preventDefault();
                 this.focusNextNode(1);
                 break;
             case keyCodeArrowUp:
                 // preventDefault to ensure we don't scroll the page
```
awentzel, line 189: If we remove prevent default, should we also remove the comment for it in the line above?

tech4GT: I left it in so as to not remove the information that it in fact prevents default (or returns true now) to prevent the page from scrolling. Do you think probably modifying it might be a better option?

tech4GT: Updated the comment :D

scomea:
This needs "Run 'yarn change' to generate a change file" done before it can be merged.

tech4GT:
> This needs "Run 'yarn change' to generate a change file" done before it can be merged.

@scomea Shall I run it or, would it be run on master and I should wait for that?

scomea:
@tech4GT Yes, you should run it locally and update the branch, after that I'll get it in.  (I'll click GitHub's "Update Branch" buttons to get a branch mergable, but beyond that I leave it to the change authors)

EisenbergEffect:
Ping @scomea as a reminder on final review and merge of this fix.

EisenbergEffect (approved):

EisenbergEffect:
@scomea @chrisdholt Looks like this is fully approved and has been sitting here waiting to be merged for a long while. Is this still applicable and can it be merged?

scomea:
> @scomea @chrisdholt Looks like this is fully approved and has been sitting here waiting to be merged for a long while. Is this still applicable and can it be merged?

No longer applicable.  Tree item does not even handle key events now.  Can be closed IMO.

chrisdholt:
> > @scomea @chrisdholt Looks like this is fully approved and has been sitting here waiting to be merged for a long while. Is this still applicable and can it be merged?
> 
> No longer applicable. Tree item does not even handle key events now. Can be closed IMO.

Yep, this was one reason this was sitting. We had a “breaking” change to move the events wholesale.

PR #4498, opened 2021-03-25
fix: add display:grid to flipper slots to prevent container from changing glyph position

### LINKED ISSUE #4497: Inherited font-sizes move glyph positions in fast-flipper
When `<fast-flipper>` is a child of any element with a font-size, the slots inherit the parent font-size which messes up the glyph positioning. You can even see it in the docs page: https://www.fast.design/docs/components/flipper/ where it the glyph is offset due to the 15.4px font-size.

- Go to the [example codepen](https://codepen.io/brookdurant/pen/vYgNEKa)
- Anywhere `<fast-flipper>` or `<fluent-flipper>` are children of elements that have a font-size and don't have a font-size explicitly set on themselves

I would expect default flipper behavior to be **either**: 
- the flipper to be unaffected by parent `font-size` (preferable)
- the entire flipper and its children to resize based on inherited `font-size`

flipperError

I think the flipper behavior is different that the button behavior. Should they be consistent?

One option for fix could be setting `font-size: 0;` on the flipper by default.

brookdozer:
# Description
Set `font-size:0` in flipper.styles.ts for the `.next` and `.previous` slot classes so the glyphs will render in their proper positions regardless of container's inheritable font-size

## Motivation & context
Fixes #4497 

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-components/src/flipper/flipper.styles.ts
+++ b/packages/web-components/fast-components/src/flipper/flipper.styles.ts
@@ -50,7 +50,7 @@ export const FlipperStyles = css`
 
     .next,
     .previous {
-        position: absolute;
+        position: relative;
         ${
             /* Glyph size and display: grid are temporary - 
             replace when adaptive typography is figured out */ ""
--- a/packages/web-components/fast-components/src/flipper/flipper.styles.ts
+++ b/packages/web-components/fast-components/src/flipper/flipper.styles.ts
@@ -50,13 +50,13 @@ export const FlipperStyles = css`
 
     .next,
     .previous {
-        position: relative;
+        position: absolute;
         ${
-            /* Glyph size and font-size is temporary - 
+            /* Glyph size and display: grid are temporary - 
             replace when adaptive typography is figured out */ ""
         } width: 16px;
         height: 16px;
-        font-size: 0;
+        display: grid;
     }
 
     :host([disabled]) {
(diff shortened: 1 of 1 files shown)


scomea:
Wouldn't it be weird if someone adds content to a slot and we apply font size 0?  Can we specify the height of something instead?

chrisdholt:
> Wouldn't it be weird if someone adds content to a slot and we apply font size 0? Can we specify the height of something instead?

This is a good call out if someone wanted to pass localized text. What if we applied the font size to slotted SVG’s?

‘’’css
::slotted(SVG) {
    font-size: 0;
    line-height: 0;
}
’’’

brookdozer:
Great comments, @scomea and @chrisdholt, I think we should move discussion to the bug, #4497. The `::slotted(svg) {}` won't work because the flipper content isn't always slotted.

radium-v (changes requested):
Setting a `display` value on `.previous` and `.next`, or on slotted `svg`s would be better and more intuitive than resetting the `font-size`/`line-height`.

radium-v (approved):

scomea (approved):

EisenbergEffect (approved):

EisenbergEffect:
@bheston See the comment in the PR above. I've noticed a number of similar comments in other components. How far are we from having the adaptive typography worked out enough to start resolving these "todos"?

PR #4502, opened 2021-03-26
fix: add support for slots in horizontal-scroll content

robarbms:
# Description

Allows for custom components to slot in content into the horizontal-scroll component.

## Motivation & context

When slotting in content into the horizontal-scroll, it would see it as 1 child node. This change lets the horizontal-scroll component to set the slotted content as the scroll content.

## Issue type checklist

## Process & policy checklist


SethDonohue:
I ran the stories and I can't see the difference in either. The slotted items look the same and functionality was the same. What was happening before this change that you noticed the difference?

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -205,11 +205,32 @@ export class HorizontalScroll extends FASTElement {
...
     };
 
+    /**
+     * Looks for slots and uses child nodes instead
+     * @internal
+     */
+    private updateScrollStops(): void {
+        const updatedItems: HTMLElement[] = [];
+
+        this.scrollItems.map(item => {
```
nicholasrice, line 215: Minor nit, I'd recommend using `forEach` instead of `map` here because you're not actually returning a value, which is the defining feature between `map` and `forEach`

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -205,11 +205,32 @@ export class HorizontalScroll extends FASTElement {
...
+            if (item instanceof HTMLSlotElement) {
+                item.assignedElements().forEach(child => {
```
nicholasrice, line 217: You can push all these into `updatedItems` w/o the extra loop by using `apply`:

```ts
updatedItems.push.apply(updatedItems, item.assignedElements())
```

nicholasrice (approved):

robarbms:
> I ran the stories and I can't see the difference in either. The slotted items look the same and functionality was the same. What was happening before this change that you noticed the difference?

I don't have an example checked in. This is in the case that you want to use a horizontal-scroll component and slot items from the parent container into the horizontal-scroll element. The horizontal-scroll component was seeing the single slot element so setting up the scroll stops didn't work.

I have a working sample on my box that I'm not checking in. I've created a fast-stripe component that uses the horizontal-scroll in the template.

http://robdor:6007/?path=/story/horizontal-scroll--horizontal-scroll

scomea (approved):

chrisdholt (approved):

PR #4505, opened 2021-03-26
fix: labToLCH problem for colors where a and b are near zero

eigenvector32:
# Description
If converting a LAB color to LCH and both A and B are very near zero but not equal to zero (eg: floating point errors
resulting in them being -0.00000001 or whatever) then the H value of the resulting LCH color will be very different from
what it would be if they were exactly equal to zero. This fix sets H to 0 if both A and B are near the discontinuity at 0.

Also adds an option to manually create a Histogram for use in quantize rather than having one created automaticially.

## Motivation & context

This fix to labToLCH can come up when a previous operation results in a color that is almost exactly black but floating point errors mean its A and B are not exactly 0. EG: If a previous blend should have produced black but the numbers haven't been rounded off. 

The option to quantize based on a Histogram can be useful if you want to re-use a Histogram several times for perf reasons
or if you want to alter the Histogram to add or remove color data prior to quantizing. 

## Issue type checklist

## Process & policy checklist

- [X] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.


eigenvector32:
Note that the test data which needed updating all has to do with pure greyscale colors (those where red, green and blue are all equal). For LCH colors where C is ~0 the H parameter doesn't actually mean anything. It's like HSL where L = 0. 

The changes for the saturation results on a pure greyscale color however do indicate that there was previously a bug there which was fixed by the change to labToLCH. Calling saturateViaLCH on a greyscale color should not change it.

nicholasrice (approved):

chrisdholt (approved):

PR #4515, opened 2021-03-30
fix: remove the use of detail as this is a synthetic event value and will not be compatible with React 17

janechu:
# Description

Removed the check using `detail` and resolve whether a click is a single or double click using only `setTimeout`.

## Motivation & context

The use of the synthetic events `detail` would occasionally result in console logs being generated from React. Due to this also being removed in React 17 the double click detection is remediated to work solely from `setTimeout`.

## Issue type checklist

## Process & policy checklist


EisenbergEffect:
Possibly a different approach: https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event

janechu:
@EisenbergEffect unfortunately I have an action for a single click as well as a double click. For React (not sure about native events), it will fire the double click once and the single click twice if a double click is used.

EisenbergEffect (approved):

PR #4523, opened 2021-03-31
fix: value of 0 for progress is indeterminate

### LINKED ISSUE #4517: Progress value of 0 is indeterminate

When progress value is set to `0` progress goes into its `indeterminate` state. `0` should be a `determinate` value.

https://stackblitz.com/edit/fast-progress?file=src/progress-app.ts

eljefe223:
# Description
Closes #4517 

Value of `0` for progress is indeterminate this is incorrect, `0` is a `determinate` value. this is also outline in the progress spec

```
- `value`, the value of the progress, if not passed the progress will be put into its "indeterminate" state.
```

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/progress/progress.template.ts
+++ b/packages/web-components/fast-foundation/src/progress/progress.template.ts
@@ -15,7 +15,7 @@ export const ProgressTemplate: ViewTemplate<BaseProgress> = html`
         class="${x => (x.paused ? "paused" : "")}"
     >
         ${when(
-            x => x.value,
+            x => typeof x.value === "number" && !Number.isNaN(x.value),
             html<BaseProgress>`
                 <div class="progress" part="progress" slot="determinate">
                     <div
@@ -27,7 +27,7 @@ export const ProgressTemplate: ViewTemplate<BaseProgress> = html`
             `
         )}
         ${when(
-            x => !x.value,
+            x => typeof x.value !== "number" || Number.isNaN(x.value),
             html<BaseProgress>`
                 <div class="progress" part="progress" slot="indeterminate">
                     <slot class="indeterminate" name="indeterminate">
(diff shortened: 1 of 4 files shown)


scomea (approved):

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/progress/progress.template.ts
```diff
@@ -27,7 +27,7 @@ export const ProgressTemplate: ViewTemplate<BaseProgress> = html`
             `
         )}
         ${when(
-            x => !x.value,
+            x => typeof x.value !== "number" || Number.isNaN(x.value),
```
janechu, line 30: Shouldn't the `nullableNumberConverter` convert value to `null` if the value is `null`, `undefined` or `NaN`? Therefore should you be checking for `null` value not `NaN`?

https://github.com/microsoft/fast/blob/master/packages/web-components/fast-element/src/components/attributes.ts#L87

nicholasrice: Good catch, @eljefe223 sorry for the miss-guidance on this.  A simple null check should be fine, but also update the TS type for the property to be `number | null` so the property type matches the converter's return type.

packages/web-components/fast-foundation/src/progress/progress.template.ts
```diff
@@ -15,7 +15,7 @@ export const ProgressTemplate: ViewTemplate<BaseProgress> = html`
         class="${x => (x.paused ? "paused" : "")}"
     >
         ${when(
-            x => x.value,
+            x => typeof x.value === "number" && x.value !== null,
```
janechu, line 18: `null` will never be a `number` so you can just do:

```suggestion
            x => typeof x.value === "number",
```

packages/web-components/fast-foundation/src/progress/progress.template.ts
```diff
@@ -27,7 +27,7 @@ export const ProgressTemplate: ViewTemplate<BaseProgress> = html`
             `
         )}
         ${when(
-            x => !x.value,
+            x => typeof x.value !== "number" || x.value === null,
```
janechu, line 30: Same situation here:

```suggestion
            x => typeof x.value !== "number",
```

janechu (approved):

PR #4534, opened 2021-04-03
fix: allows a float to be accurately used for the step attribute

janechu:
# Description

This change assumes that all `step` attributes are floating points and adjusts them to be integers during calculations to overcome issues with doing math with floats.

## Motivation & context

An issue arose when attempting to use the `<fast-slider>` with a `step` attribute of `0.1`. Math in JavaScript with regards to integers is straight forward, however once floats are used it is possible to get floating point issues.

## Issue type checklist

## Process & policy checklist

### DIFF
--- a/packages/web-components/fast-foundation/src/slider/slider.ts
+++ b/packages/web-components/fast-foundation/src/slider/slider.ts
@@ -447,12 +447,29 @@ export class Slider extends FormAssociatedSlider implements SliderConfiguration 
         if (isNaN(value)) {
             value = this.min;
         }
+
+        /**
+         * The following logic intends to overcome the issue with math in JavaScript with regards to floating point numbers.
+         * This is needed as the `step` may be an integer but could also be a float. To accomplish this the step  is assumed to be a float
+         * and is converted to an integer by determining the number of decimal places it represent, multiplying it until it is an
+         * integer and then dividing it to get back to the correct number.
+         */
         let constrainedValue: number = value - this.min;
-        const remainderVal: number = constrainedValue % Number(this.step);
+        let remainderValue: number = constrainedValue % Number(this.step);
+        const stepString: string = this.step + "";
+        const decimalPlacesOfStep: number = !!(this.step % 1)
+            ? stepString.length - stepString.indexOf(".") - 1
+            : 0;
+        const roundedConstrainedValue: number = Math.round(constrainedValue / this.step);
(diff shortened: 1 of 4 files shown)


packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -447,12 +447,29 @@ export class Slider extends FormAssociatedSlider implements SliderConfiguration
         if (isNaN(value)) {
             value = this.min;
         }
+
+        /**
+         * The following logic intends to overcome the issue with math in JavaScript with regards to floating point numbers.
+         * This is needed as the `step` may be an integer but could also be a float. To accomplish this the step  is assumed to be a float
+         * and is converted to an integer by determining the number of decimal places it represent, multiplying it until it is an
```
janechu, line 454:
```suggestion
         * This is needed as the `step` may be an integer but could also be a float. To accomplish this the step is assumed to be a float
         * and is converted to an integer by determining the number of decimal places it has, multiplying it until it is an
```

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -447,12 +447,29 @@ export class Slider extends FormAssociatedSlider implements SliderConfiguration
...
+         */
         let constrainedValue: number = value - this.min;
-        const remainderVal: number = constrainedValue % Number(this.step);
+        let remainderValue: number = constrainedValue % Number(this.step);
+        const stepString: string = this.step + "";
+        const decimalPlacesOfStep: number = !!(this.step % 1)
+            ? stepString.length - stepString.indexOf(".") - 1
+            : 0;
+        const roundedConstrainedValue: number = Math.round(constrainedValue / this.step);
+        const stepMultiplier: number = Math.pow(10, decimalPlacesOfStep);
```
scomea, line 464: We could save a few cycles by caching "stepMultiplier" into a private and only recalculating when step changes.

scomea (approved):

robarbms:
I ran into a similar problem with the number-field component. I used Number toPrecision() to resolve JS calculation errors. I used precision to the 12th decimal since the component may be needed for scientific purposes which was sufficient to catch errors. Errors seemed to happen around the 14th decimal position.

chrisdholt (approved):
This looks good to me - would be great to get a few other eyes on this from @microsoft/fast-collaborators

EisenbergEffect (approved):

PR #4535, opened 2021-04-03
feat: add toolbar component

radium-v:
# Description

Adds the `<fast-toolbar>` component.

## Issue type checklist

## Process & policy checklist


packages/utilities/fast-web-utilities/src/key-codes.ts
```diff
@@ -143,3 +143,23 @@ export const keyCodeTilde: number = 192;
...
+export const keyArrowLeft = "ArrowLeft";
+export const keyArrowRight = "ArrowRight";
+export const keyArrowUp = "ArrowUp";
+
+export const keyEnter = "Enter";
+export const keyEscape = "Escape";
+export const keySpace = " ";
+export const keyTab = "Tab";
+
+export enum ArrowKeys {
```
scomea, line 160: Can this be a type instead of an enum?

packages/utilities/fast-web-utilities/src/key-codes.ts
```diff
@@ -143,3 +143,23 @@ export const keyCodeTilde: number = 192;
 export const keyCodeWindowsLeft: number = 91;
 export const keyCodeWindowsOpera: number = 219; // Opera
 export const keyCodeWindowsRight: number = 92;
+
+/**
+ * String values for use with KeyboardEvent.key
+ */
+export const keyArrowDown = "ArrowDown";
```
scomea, line 150: Type as string?

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
```diff
@@ -136,7 +136,7 @@ export class RadioGroup extends FASTElement {
         super.connectedCallback();
         this.direction = getDirection(this);
         this.setupRadioButtons();
-        this.parentToolbar = this.parentElement?.closest('[role="toolbar"]');
+        this.parentToolbar = this.closest<HTMLElement>('[role="toolbar"]');
         this.isInsideToolbar =
```
scomea, line 140: I'm wondering if we should rename this to  something like "isTabStop" and make it a settable public api because there may be other situations where an author may want to do this, like say a radio group in a grid cell where authors want to follow "only one of the focusable elements contained by the grid is included in the page tab sequence." (https://www.w3.org/TR/wai-aria-practices/#grid).  We could then also consider consolidating detection in the parent (ie. toolbar) and have it check it's children for the presence of "isTabStop" and act accordingly.  Seems more scaleable.

radium-v: This situation is unique to a `radiogroup` inside of a `toolbar`: https://w3c.github.io/aria-practices/#for-radio-group-contained-in-a-toolbar

scomea: I don't believe it is unique.  Data grids are typically a single tab stop yet can contain cells that contain multiple widgets which presumably could include a radio group, so in that situation an author would also want to suppress the behavior of radio groups adding child radio buttons to the tab queue.  There's even mention of it being valid for complex grid-cells possibly having their own contained tab queue like a modal dialog when a user has navigated "into" it.  Implementing this such that any parent, be it toolbar, grid-cell or something else can turn this behavior on or off seems more flexible to me.

radium-v:
There's a weird bug that's affecting `<input type="radio">` elements:

1. Tab into a toolbar
2. Use the arrow keys to focus a radio input
3. Mark the radio checked with the space key
4. In the same toolbar, use the arrow keys to focus an unchecked radio input
5. Tab away from the toolbar

At this point, there's no way to tab back into the toolbar unless a control is focused a different way (like clicking).

scomea:
@radium-v Should I be able to repro with your current branch?  Tabbing/selecting radio buttons seems to work fine with your examples.

radium-v:
Here's a screen recording of the bug. My inputs are tab, right, right, space, right, tab, shift+tab.

https://user-images.githubusercontent.com/863023/116447022-65725e00-a80c-11eb-858c-8b35a6248f36.mp4

packages/utilities/fast-web-utilities/src/key-codes.ts
radium-v: This is used as a collection of values so a type won't work in its place.

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
radium-v: While there are similarities between keyboard navigation behaviors for composite widgets, this situation is specifically unique as defined in the [`radiogroup` section of the ARIA spec](https://w3c.github.io/aria-practices/#keyboard-interaction-15). For that reason, this property and its surrounding behaviors are only relevant when a `fast-radio-group` is inside of a `fast-toolbar`.

scomea: If you look though the [grid section of the Aria spec](https://w3c.github.io/aria-practices/#grid) it specifically mentions a radio group as a valid child of a complex cell, and discusses a few approaches to keyboarding and how the tab key could behave in these scenarios.  So someone putting a radio group in a data grid cell could very likely want to take it out of the tab order.  This seems like the same functionality toolbar needs so why not a mechanism that accommodates both?

packages/utilities/fast-web-utilities/src/key-codes.ts
scomea: We use types for collections of values in a bunch of places:
ie. `export type VerticalPosition = "top" | "bottom" | "unset";`
Apparently there are perf benefits.

packages/web-components/fast-foundation/src/radio-group/radio-group.ts
chrisdholt: @scomea can you open an issue to look at exploring and possibly adding the behavior based on a deeper investigation into something like a layout grid (which we don't currently have). My concern here is that we'll end up w/ some scope creep outside the toolbar itself. I'm all for exploring if we can update radio to leverage this in data grid as well, I just want to be careful to randomize as part of this PR and make it dependent on that change. Thoughts?

scomea: Data grid currently supports complex cells which could have their own internal navigation queues.  Conceptually, I think that any component that add things to the tab queue automatically we'd probably want to be able to suppress that behavior (I'm thinking it will come up when someone wants to do a nested grid, for example). Also, I think it's a better design to have the parent suppress the behavior of the child (ie. toolbar suppresses the behavior in radio group) than the child try to determine the right behavior by detecting the parent (having to update the child for every new scenario isn't great).  I'll open an issue, but it seems like something we should standardize on.

Garbee: When I first looked over the code I also felt the child determining behavior weird. Components should be instructed on how to act rather than self-determining how to act in any given context.

I don't feel this should block merge though. This seems to be one of the first times this has come up (although I haven't deeply reviewed the other component code at all) and it's fine enough to land and able to be refined. As brought up with Datagrid, once more higher-order components are built these situations will be hit more. So these are good early sample cases to chomp on.

packages/web-components/fast-components/src/toolbar/toolbar.definition.ts
```diff
@@ -0,0 +1,37 @@
+import type { WebComponentDefinition } from "@microsoft/fast-tooling/dist/esm/data-utilities/web-component";
+import { DataType } from "@microsoft/fast-tooling";
+import { Orientation } from "@microsoft/fast-web-utilities";
+
+export const fastToolbarDefinition: WebComponentDefinition = {
```
chrisdholt, line 5: @janechu made an update to definition files the other day, I think this may need to update

packages/web-components/fast-foundation/src/toolbar/toolbar.options.ts
```diff
@@ -0,0 +1,49 @@
...
+            rtl: Increment.next,
+        },
+    },
+    ArrowRight: {
+        horizontal: {
+            rtl: Increment.previous,
+            ltr: Increment.next,
+        },
+    },
+};
```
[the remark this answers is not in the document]
radium-v: Since this is specific to the toolbar, I moved it into the main file and renamed it to `ToolbarArrowKeyMap`.

packages/web-components/fast-foundation/src/toolbar/toolbar.options.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Enum for previous/next increments.
+ *
+ * @internal
+ */
+export enum Increment {
+    previous = -1,
+    next = 1,
+}
```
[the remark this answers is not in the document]
radium-v: My enum usage was a little overzealous so I removed the `Increment` enum.

packages/utilities/fast-web-utilities/src/key-codes.ts
radium-v: I updated the `ArrowKeys` enum to now be a const with a combined type.

chrisdholt (approved):

packages/web-components/fast-foundation/src/toolbar/toolbar.template.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The template for the {@link @microsoft/fast-foundation#(Toolbar:class)} component.
+ *
+ * @public
+ */
+export const ToolbarTemplate: ViewTemplate<Toolbar> = html`
+    <template
+        aria-label="${x => x.ariaLabel}"
+        aria-labelledby="${x => x.ariaLabelledby}"
```
scomea, line 14: Is there a reason developers can't set the label and labelledby attributes themselves since this is on the outer node of the component?

scomea (approved):

litteredwitherrors:
@radium-v When I activate a toolbar child element via click or keyboard that child's `tab-index` changes to 0. Is that the intended behavior?

chrisdholt:
> @radium-v When I activate a toolbar child element via click or keyboard that child's `tab-index` changes to 0. Is that the intended behavior?

I'll let @radium-v advise here, but I think I *would* expect this to a certain extent...click is a focusable event, so I'd expect focus to be brought to that element. The kicker would be if I click and then hit tab, do I move to the next element? If you do, then I think we are gravy :)

radium-v:
> @radium-v When I activate a toolbar child element via click or keyboard that child's `tab-index` changes to 0. Is that the intended behavior?

@litteredwitherrors the toolbar uses a roving tabindex pattern, and keyboard tabbing only handles entering and exiting the toolbar. navigation between items in the toolbar is handled with the arrow keys. 

> if I click and then hit tab, do I move to the next element? If you do, then I think we are gravy :)

@chrisdholt clicking then hitting tab moves focus out of the toolbar and onto the next focusable object on the page. If you then shift+tab to go back, focus moves back to the previously focused toolbar control. 🍛

chrisdholt:
> > @radium-v When I activate a toolbar child element via click or keyboard that child's `tab-index` changes to 0. Is that the intended behavior?
> 
> @litteredwitherrors the toolbar uses a roving tabindex pattern, and keyboard tabbing only handles entering and exiting the toolbar. navigation between items in the toolbar is handled with the arrow keys.
> 
> > if I click and then hit tab, do I move to the next element? If you do, then I think we are gravy :)
> 
> @chrisdholt clicking then hitting tab moves focus out of the toolbar and onto the next focusable object on the page. If you then shift+tab to go back, focus moves back to the previously focused toolbar control. 🍛

I was thinking I mistakenly said "tab" here when I meant, "arrow to the next focusable item". I'm assuming that works based on what I see here :)

PR #4542, opened 2021-04-05
update contributing documentation with `yarn change` guidance

chrisdholt:
# Description
This PR updates the contributing documentation to include details about generating change files as part of the PR process.

## Issue type checklist

## Process & policy checklist


EisenbergEffect (approved):
Do we need to provide any guidance as to how the comment should be written/structured?

chrisdholt:
> Do we need to provide any guidance as to how the comment should be written/structured?

I thought about including that here - the nice thing is that this goes into review, so it should be reviewed as part of the PR. I think we can add some guidance if you feel it would be helpful.

EisenbergEffect:
Good point about it being part of the PR and thus reviewable. Maybe we wait and see what happens. We can always add the guidance if we see a need.

janechu (approved):

awentzel (approved):

PR #4557, opened 2021-04-07
fix: ensure DesignToken works with falsey values

nicholasrice:
# Pull Request

## 📖 Description
I noticed when doing some implementation tests that a `DesignToken` set to a *falsey* value would be un-resolvable for a node, throwing an exception. This changes resolution logic to correctly handle *falsey* values, so that values like `0`, `""`, `NaN`, and `null` are all valid values for a `DesignToken`.

This change also extends the type-guard to dis-allow `DesignToken<undefined>` and `DesignToken<void>`, because the `undefined` value is used to determine if a value should be resolved from a parent node.

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-foundation-bf618f1d-588a-4fef-8ee0-aadb2142c79b.json
+++ b/change/@microsoft-fast-foundation-bf618f1d-588a-4fef-8ee0-aadb2142c79b.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "ensure DesignToken works with falsey values",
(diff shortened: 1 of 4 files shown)


chrisdholt (approved):
There's a ping incoming on the discord channels, but you'll want to run `yarn change` from the root of the repo as we have officially moved to leveraging Beachball for any future publishes.

nicholasrice:
@chrisdholt ah hah! Woot!! Will do

EisenbergEffect (approved):

PR #4562, opened 2021-04-08
DRAFT: Radio buttons auto populate empty aria-label

scomea:
## 📖 Description
I think it would be good if basic radio buttons with simple text labels automatically did the right thing as far as Aria attributes is concerned.  This change adds a mechanism that uses the text content of a radio button as the aria-label attribute if the author does not set it themselves.  It just makes accessibility more automatic.

### 🎫 Issues
Thinking of ways to make accessibility easier.

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-components-c6f32a1d-4483-40e5-b738-fb6db238c071.json
+++ b/change/@microsoft-fast-components-c6f32a1d-4483-40e5-b738-fb6db238c071.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "fix aria labelling",
+  "packageName": "@microsoft/fast-components",
+  "email": "scomea@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-foundation-484f5a4c-a1b2-4f52-a501-80d25d693300.json
+++ b/change/@microsoft-fast-foundation-484f5a4c-a1b2-4f52-a501-80d25d693300.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "fix aria labelling",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "scomea@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/web-components/fast-components/src/checkbox/fixtures/base.html
+++ b/packages/web-components/fast-components/src/checkbox/fixtures/base.html
@@ -41,7 +41,21 @@
     <fast-checkbox checked>Oranges</fast-checkbox>
 </fieldset>
 
-<h2>Visual vs audio label</h2>
+<h2>Aria</h2>
 <fast-checkbox>
-    <span aria-label="Audio label">Visible label</span>
+    Labeled by content
 </fast-checkbox>
+
+<div style="display: flex; flex-direction: column; margin-top: 12px;">
+    <h3 id="externallabel">External label</h3>
+    <fast-checkbox aria-labeledby="externallabel">
+        Content not used as aria label
+    </fast-checkbox>
+</div>
+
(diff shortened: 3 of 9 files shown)


chrisdholt:
I think the issue here is likely more nuanced than what is expected - and I likely should've caught this with Radio...the HTML label element only works with HTML radios, it does not work with ARIA radios. Given that, I think rather than using the `<label>` element here we should likely just use a span and try to correlate the text to the radio as the label using aria-labelledby. Given that the radio is visually labelled, we should likely use that rather than aria-label. 

> Each radio element is labelled by its content, has a visible label referenced by aria-labelledby, or has a label specified with aria-label.

This method prevents needing to create an additional attribute. In the case we have text being slotted in, that is mapped internally so everything "just works". Similarly, if someone wants to use the radio without a visible label (ARIA label only) that can still be passed to the host which has the role of radio. This should also prevent the need to grab the contents which have been slotted and parse them, which seems costly.

I definitely see the problem, but I think we can get away with a cheaper and less intrusive solution. Good catch on this!

scomea:
Why does the span need to exist?  Wouldn't this suffice?
<fast-radio>
     This description is visible and read by aria
</fast-radio>

chrisdholt:
> Why does the span need to exist? Wouldn't this suffice?
> `<fast-radio> This description is visible and read by aria </fast-radio>`

That would be the implementation side - on the template we would swap the existing `label` to be a `span` with aria-labelledby. In this case, we can use a scoped ID (within the shadow DOM) to correlate the text being slotted (as it currently is today) to the radio itself.

So the change only exists in the template and appears as so:

image

With this change, if there is no slotted content, it's up to the author to add aria-label. However, if they include slotted content, it's automatically tied to the radio as is noted in the spec. So, the API stays the same and intuitive - it just works now :)

scomea:
I'll give that a try and update the PR.  Note that with this approach I believe the automatically set "aria-labeledby" will override an author set "aria-label".  What I'd expect is  - use author set aria-labeledby if it exists, otherwise use author set aria-label, otherwise try to generate based on content.

chrisdholt:
> I'll give that a try and update the PR. Note that with this approach I believe the automatically set "aria-labeledby" will override an author set "aria-label". What I'd expect is - use author set aria-labeledby if it exists, otherwise use author set aria-label, otherwise try to generate based on content.

My code was super quick to rely what I was describing above, so it may not be exactly what's needed :). In my pseudo code we aren't adding ARIA label if there isn't slotted content, so that shouldn't apply it. In the case you are calling out, we have author supplied aria-labelledby (also valid) and we should fall back to that in the case it exists. I think that's fair and a good point. In case it's null I'd assume it doesn't exist.

In the case where an author passes visible text AND an ARIA label to the host element, I would expect that the aria-label actually supersedes the shadowed implementation of aria-labelledby - simply because of how ARIA label works, but we should test that being the case. Ideally the platform makes this part easy, but definitely a good thing to test.

scomea:
@chrisdholt The issue with that approach is that the aria-labeledby attribute is on the outer component in the light dom and the label element is in the shadow dom so id does not resolve.

chrisdholt:
> @chrisdholt The issue with that approach is that the aria-labeledby attribute is on the outer component in the light dom and the label element is in the shadow dom so id does not resolve.

I think something else is going on - this should resolve fine IMO. While the value is being reflected to the light DOM, it's in the template and available to the host. Everything in the current fixtures file (including aria-labelledby) is reading out the label along w/ the radio using VoiceOver.

brookdozer:
Nice solution! It's easy to understand and performs well with Voiceover on a Mac. Thanks for adding checkbox!

awentzel:
@scomea, what's the status of this PR?  As a Draft PR that's 11 months old should it remain here or should it be abandoned or completed?

Issue #3136, opened 2020-05-14
Add Visual Studio Code workspace to project root


radium-v:
**What feature are you requesting?**
It would be great if we could add a [`.vscode` workspace directory](https://code.visualstudio.com/docs/getstarted/settings) to the project root.

**What would be the benefit of adding this feature?**
It seems that our team primarily uses VSC, and this would allow us to recommend extensions for syntax highlighting and also set some editor configurations 

**What alternatives have you considered?**
There are some more environment-agnostic options like `editorconfig`, which could work as an alternative option for users who don't use VSC.

**Is there any additional context for your request?**
Certain VSC settings can conflict or interfere with the tools we use - for instance, Prettier doesn't catch things in all files, and things like indentation or EOL newline can be enforced to reduce the dependence on Prettier.

radium-v:
Here are some defaults I'm thinking about:

```json
{
  "[typescript]": {
    "editor.tabSize": 4
  },
  "editor.trimAutoWhitespace": true,
  "files.eol": "\n",
  "files.trimTrailingWhitespace": true,
  "files.insertFinalNewline": true,
  "files.trimFinalNewlines": true,
  "javascript.format.enable": false,
  "javascript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false,
  "js/ts.implicitProjectConfig.experimentalDecorators": true,
  "npm.packageManager": "yarn",
  "typescript.enablePromptUseWorkspaceTsdk": true,
  "typescript.format.enable": false,
  "typescript.preferences.quoteStyle": "double"
}
```

awentzel:
I'm in favor of adding this as I've been using for a while myself. We can add "cSpell.words" and "cSpell.ignoreWords" to eliminate common workspace spelling nuances.  I also will require this potentially in the future for different Azure packages that can be managed within VSCode.

We should add a task to meet with the team and see what we can determine are safe defaults we'd want to implement across packages.

radium-v:
This is resolved via #4574. @awentzel, can you create a new issue for adding spellcheck?

awentzel:
Per note above by Radium-v. Resolving this PR.

PR #4564, opened 2021-04-08
feat: vs code file auto save

Ibrahimmaga:
Save / Auto Save#
# Pull Request

## 📖 Description
By default, VS Code requires an explicit action to save your changes to disk, Ctrl+S.
However, it's easy to turn on Auto Save, which will save your changes after a configured delay or when focus leaves the editor. With this option turned on, there is no need to explicitly save the file. The easiest way to turn on Auto Save is with the File > Auto Save toggle that turns on and off save after a delay.
There is a scenarios  when you lunch vs code from a terminal and your old settings may not be there. This Pr will auto save all files within the workspace.
For more control over Auto Save, open .vscode settings and find the associated settings:

### 🎫 Issues
https://github.com/microsoft/fast/issues/3136

### General

### Component-specific

### DIFF
--- a/.github/.vscode/settings.json
+++ b/.github/.vscode/settings.json
@@ -1,0 +1,4 @@
+{  
+  "files.autoSave": "afterDelay",
+  "files.autoSaveDelay": 1000,  
+}


.github/.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{  
+  "files.autoSave": "afterDelay",
+  "files.autoSaveDelay": 1000,  
+}
```
chrisdholt, line 4: I'm not sure this is something we should turn on for everyone who wants to work in the repo. While I think it may be helpful for some, I'm not sure I see this as intuitive or expected behavior. If there is something I don't want to save, I need to remember to explicitly *un-save* it somehow or revert the behavior.

This seems more like something that should be added in user settings rather than as default behavior for the repo.

chrisdholt (changes requested):

.github/.vscode/settings.json
Ibrahimmaga: This issue was created a long time ago. I didn't think It was necessary. But, can we guarantee that everyone will do the settings always for autosave?. As we always forget to run yarn or commit before pushing to master. 
@John Kreitlow created the issue and he has suggested some workspace config settings. but I only find the auto save necessary.  Let see what he thinks about the PR.  if not necessary still, we can close it.

radium-v: I added a [comment to the issue](https://github.com/microsoft/fast/issues/3136#issuecomment-816007158) with some suggestions for defaults. We don't want to set anything that could interfere with a developer's regular way of working, we only want to enforce some formatting-specific settings to reduce issues when working in the project.

awentzel: @radium-v given you originally created this issue, I think, would you mind specifying the defaults you were thinking that makes sense and we can just do those for now?  Thanks

radium-v: The link in my previous comment points to a set of suggested defaults I added to the issue. While there shouldn't be any harm in adding them, I would like to get more input from the team before committing anything.

radium-v:
Closing and moving to #4574.

Issue #4556, opened 2021-04-07
rfc: follow up on beachball publish


Ibrahimmaga:
# 💬 RFC

beachball publish failed due to access restriction
## 🔦 Context
![image](https://user-images.githubusercontent.com/14119043/113899223-6ff78580-9781-11eb-8661-7d28564bef50.png)

1. try to use --access public command and verify our credentials
2. have Chris use his credential
3. contact Ken from beachball team
4. research npm publish for organization
     Organizations allow teams of contributors to read, write, and public and private packages. Organizations are free when they 
     publish public packages. When organizations publish private packages, an npm Teams subscription is required. For more 
     information on npm Teams pricing, see our products page.
     https://docs.npmjs.com/adding-members-to-your-organization 
     https://docs.npmjs.com/cli/v7/using-npm/scope
5. binary path require configuration
    we will use "scripts-prepend-node-path=true" in .npmrc folder and it will solve the problem
6. npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@microsoft%2ffast-foundation - Not found
    The PUT is use to insert into an API.  Here when publish is writing to npm, it need to have a permission. We need to provide 
     npm token with write access.
     .npmrc and write //registry.npmjs.org/:_authToken=${NPM_TOKEN} into it
 7. Update $NPM_PAT to make sure it has read and write access policies
 8. npm config set registry https://registry.npmjs.org/ in npmrc

awentzel:
I'm not sure the 4th item is relevant since we're part of Microsoft organization, we shouldn't have to worry about pricing plans, and we've already registered our service account with our organization NPMjs admin.

Ibrahimmaga:
Ken confirm that we should use --access public command  to not be restricted

awentzel:
closed by #3889

PR #4578, opened 2021-04-12
fix: add -r to publish and set node-path to true

Ibrahimmaga:
# Pull Request

## 📖 Description
1. binary path require configuration
    we will use "scripts-prepend-node-path=true" in .npmrc folder and it will solve the problem
2. npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@microsoft%2ffast-foundation - Not found
    The PUT is use to insert into an API. Here when publish is writing to npm, it need to have a permission. We need to provide
3. npm token with write access.
   .npmrc config communicate with npm and write to //registry.npmjs.org/:_authToken=${NPM_TOKEN}. However this will expose the token, So I moved it to the workflow and use -r as beachball publish docs suggest.

4. Update $NPM_PAT to make sure it has read and write access policies

### 🎫 Issues
#4556

### General

### Component-specific

### DIFF
--- a/.github/workflows/cd-publish.yml
+++ b/.github/workflows/cd-publish.yml
@@ -13,7 +13,7 @@ env:
 jobs:
   build_linux:
     runs-on: ubuntu-latest
-    if: contains('chrisdholt,janechu,nicholasrice', github.actor)
+    if: contains('chrisdholt,janechu,nicholasrice,awentzel,EisenbergEffect', github.actor)
 
     steps:
     - uses: actions/checkout@v2
@@ -42,5 +42,5 @@ jobs:
    
     - name: Publish NPM changed packages
       run: |
-        yarn publish-ci -y  --access public $NPM_PAT 
+        yarn publish-ci -y -r --access public $NPM_PAT 
      
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,3 @@
-registry=https://registry.npmjs.org/
+scripts-prepend-node-path=true
 package-lock=false
 


.github/workflows/cd-publish.yml
```diff
@@ -42,5 +42,5 @@ jobs:
    
     - name: Publish NPM changed packages
       run: |
-        yarn publish-ci -y  --access public $NPM_PAT 
+        yarn publish-ci -y -r --access public $NPM_PAT
```
EisenbergEffect, line 45: Just want to make sure that `$NPM_PAT` isn't a typo.

EisenbergEffect: Oh, probably "public access token". I thought it was a typo for "path". Ignore me.

EisenbergEffect (approved):

.github/workflows/cd-publish.yml
Ibrahimmaga: sounds like a typo. it is without the $. but we need $ in CI. some people use ${NPM_PAT}, which does't work. So far $NPM_PAT is working for us.

nicholasrice (approved):

.github/workflows/cd-publish.yml
awentzel: `-r` is default, so it should not be needed.

awentzel: I'm not seeing that NPM_PAT has been updated for read/write permissions.  Last update was October.  
Screen Shot 2021-04-12 at 10 29 57 AM

Ibrahimmaga: let updated

Ibrahimmaga: we can replace -r with "https://registry.npmjs.org/:_authToken=$NPM_PAT." we just can't put it in .npmrc for security reason.

awentzel: You can leave the code as is with `-r` then as I agree it shouldn't be in .npmrc.  Thanks for the clarification.

awentzel (approved):

PR #4580, opened 2021-04-12
feat: add default value mechanism to DesignToken

nicholasrice:
# Pull Request

## 📖 Description
This PR adds a `withDefault()` method to a `DesignToken` to associate a default value to a design token. This allows tokens to be used even if the token value hasn't been set for some parent context, reducing unanticipated errors and providing a better authoring experience. 

### General

### Component-specific


packages/web-components/fast-foundation/src/design-token/design-token.ts
```diff
@@ -330,16 +343,20 @@ class DesignTokenNode<T> {
...
+            return null;
+        }
+
         if (this.target !== document.body && this.target.parentNode) {
             const container = DI.getOrCreateDOMContainer(this.target.parentElement!);
             try {
                 return container.get(DesignTokenNode.channel(this.token));
             } catch (e) {
-                return null;
+                return DesignTokenNode.for(this.token, defaultElement);
```
EisenbergEffect, line 356: I wonder if we should consider adding a method to the container such as `tryGet` that doesn't throw if the service is not found, but returns null instead.

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/design-token/design-token.ts
nicholasrice: Yea, that's a good call. I'll add that and work it in here.

nicholasrice: Actually `Container` has a `has()` method, is it preferable to just test prior to `get()` if the author wants to avoid the error?

nicholasrice: I suppose that would be twice the resolution work; I'll just add it

nicholasrice: Actually that looks a little more involved than I first thought, I'll take it as a follow up

EisenbergEffect: There might not be an easy way to add it at all. I haven't looked in a while. Just a passing thought. If we can find a way, it's a potential improvement for later.

nicholasrice: I think there is, but `jitRegister` throws as well as `get()` so there are a few code-paths to patch up. https://github.com/microsoft/fast/issues/4582 is tracking and I'll add a TODO in this block

janechu (approved):

PR #4609, opened 2021-04-17
Adjust button padding with --corner-radius changes

ben-girardet:
# Pull Request

I suggest to adjust the padding of a button when the `--corner-radius` value becomes larger than 10. This allow the content of the button to render nicely even with rounded corners.

button-corner2

becomes:

button-corner

### General

### Component-specific


EisenbergEffect:
Very cool!

EisenbergEffect:
@ben-girardet Can you run `yarn change`? That will generate the changelog metadata that we need to merge the PR.

chrisdholt:
> Very cool! Technically, the PR looks fine. We'll need to make sure that this is fine from the perspective of the components team and design folks. Ping @chrisdholt

Yep, very cool - would be great to have @bheston take a look here and consider if this can work with the density story he's been investigating.

ben-girardet:
> @ben-girardet Can you run `yarn change`? That will generate the changelog metadata that we need to merge the PR.

I sure can. Just wanted to get some feedacks first to see if the idea is in line with how `density` is supposed to work. Looking forward for @bheston feedback and learn more about `density` and `design-unit`.

ben-girardet:
Small ping here. Any updates on reviews ?
No hurries on my side, but would when this gets processed (either approved or rejected) that will help me to learn some things about the density / design-unit pattern.

EisenbergEffect:
@ben-girardet Apologies for the delay 🙇‍♂️  This somehow got lost in the ebb and flow of work. Brian is going to take a look and get you some feedback soon.

bheston:
Hi @ben-girardet. Sorry for the delay. As Rob implied my notification got buried. Also sorry in advance for the long response. :)

Nice idea. I know that's a common technique some UI designs employ to adjust for visual weight of the content relative to the container.

As Chris mentioned, the density story is evolving since the React days, and the current component styles don't portray enough of the picture to see where it came from. Briefly, the new idea of the density system is to treat separately the scales of content size in relation to vertical and horizontal spacing. This would apply both inside and out of a component, so button padding as you illustrate here as well as space between buttons in something like a toolbar. Padding is separated from content sizing so density truly functions as how tight the components are in an experience, and less as a scaling factor as the previous React implementation provided.

The notion of the `design-unit` will likely persist, as that provides a simple way to define your grid for alignment, but also for branding or framework interop convenience.

What I've been thinking of recently is to structure density on separate horizontal and vertical adjustments. Base default density would be `0` for both. Our buttons currently have `12px` of horizontal padding. Our design unit is `4px`. That means our base horizontal multiplier is `3`. Adjusting the horizontal density `-1` would result in `8px` padding instead. Adjusting to `+1` would produce `16px`.

The current `10px` you see is actually the result of what was previously a function, and a lot of hard-coded assumptions. It's the target `12px` padding minus an assumed `2px` for border. The actual value is more like `(horizontal-multiplier + horizontal-density) * design-unit - border-width`.

Where this gets complicated, and directly relates to what you've noticed, is how the various sizes involved come together in the whole system. Some of the cases become unlikely, but we try to solve the systems to rationally support anything within reasonable limits. For instance, if the horizontal padding were already rather large (either through density adjustment or base multiplier), the additional padding relative to the corner radius becomes less of an issue. It also creates an opinion that the padding should always be related to the corner radius. From a design system perspective that might be true for other attributes, for instance, maybe the problem isn't as noticeable on a light background neutral button compared to the higher contrast of an accent button. Another example is the icon button, which shouldn't increase padding when fully rounded. Or when the button text size is 100px. These relationships are why it's been hard to lock down a resilient system.

Stylistically I like the change, but I'd want to make sure we apply it consistently across the components, and that it doesn't affect other scenarios like composited components or the icon button example. I would also want to make sure we could carry it forward once the density work is implemented.

From a design perspective I'll add this for one of our designers who has been looking at the component visual update to consider.

I'd love to keep the chat going and work on the density updates with the community if you or others are interested.

ben-girardet:
Thanks @bheston for this thorough answer. I must admit I don't yet grasp everything you said (I'm not familiar with react and what density was supposed to be in react for example). But I'm with you with the vision you mentioned: 

> we try to solve the systems to rationally support anything within reasonable limits

As I understand, this PR should stay stale for a while as there is some more work to do regrading `density`. I believe this must reach a point where the concepts are more precisely defined and then we can come back to this corner radius question and see how to make it fit. As you said, there are probably other cases (other than buttons) where that might apply as well.

I would be happy to follow / contribute in further discussions about density. Maybe some visual sketches can help too. 

FYI: I'm experimenting on my own with extending the FAST frame design system. I'm putting together a small app for testing purposes. This allow me to try existing components and create new ones (like a toolbar or some flowing containers). The idea is to have a few different common design (homepage, blog page, admin dashboard, ...) and see how components interacts with each others in those scenarios. It's still WIP but I can share it when ready if that can be of any interest... or maybe you already have something like this...?

chrisdholt:
> Thanks @bheston for this thorough answer. I must admit I don't yet grasp everything you said (I'm not familiar with react and what density was supposed to be in react for example). But I'm with you with the vision you mentioned:
> 
> > we try to solve the systems to rationally support anything within reasonable limits
> 
> As I understand, this PR should stay stale for a while as there is some more work to do regrading `density`. I believe this must reach a point where the concepts are more precisely defined and then we can come back to this corner radius question and see how to make it fit. As you said, there are probably other cases (other than buttons) where that might apply as well.
> 
> I would be happy to follow / contribute in further discussions about density. Maybe some visual sketches can help too.
> 
> FYI: I'm experimenting on my own with extending the FAST frame design system. I'm putting together a small app for testing purposes. This allow me to try existing components and create new ones (like a toolbar or some flowing containers). The idea is to have a few different common design (homepage, blog page, admin dashboard, ...) and see how components interacts with each others in those scenarios. It's still WIP but I can share it when ready if that can be of any interest... or maybe you already have something like this...?

Would love to see this and have you share it!

ben-girardet:
I will share this soon. Still adding a few pieces...

Meanwhile I would like to ask a question - quite related to the density / spacing discussion above. 

I've noticed in [styles/size.ts](https://github.com/microsoft/fast/blob/eeb625e346a54da4c1f338eb90341a6e2d9ddb83/packages/web-components/fast-components/src/styles/size.ts) file that there is the following export: 

```ts
export const heightNumber =
    "(var(--base-height-multiplier) + var(--density)) * var(--design-unit)";
```

I was wondering what's the benefit of having this as a TS variable that one can use in CSS Styles instead of attaching a CSS custom property in the fast design provider such as `--heightNumber: (var(--base-height-multiplier) + var(--density)) * var(--design-unit)` ?

I'm sure there is a specific reason to it... and would like to learn more.

EisenbergEffect:
Ping @bheston to continue the convo here or move to an issue.

bheston:
The `density` conversation continued in our Discord design channel. I was going to try to summarize it here, but it's kind of long.

Nick's initial answer to the last question about `heightNumber` is:
> IIRC we didn't want to elevate heightNumber to the design system because it's really an implementation detail of the design system. Storing it as a string const allows re-use of the core concept without making the entity itself configurable.

I don't recall any other factors for this implementation decision.

Regarding @ben-girardet's sample app, I'd love to see that as well. I've started to work on probably a simpler sample app in the Color Explorer as a way to prove out what adjusting the color recipes does. I figured it would probably best live as an extracted sample component we could use for implementation illustration as well as a sample throughout the various tools. I was going for an email/chat type interface (Teams-ish) with a few tabs including one with the FAST Frame from our main site. Really anything that can show all the components and some relationships in between.

Lastly, regarding the density work, we're finishing up some of the design token features, and I've been focusing on that, but plan to switch to density afterward. I'll figure out a way we can look at some of that in an open source design model - somehow :)

EisenbergEffect:
Closing this PR in favor of picking up the conversation in the Adaptive Density RFC: https://github.com/microsoft/fast/issues/5510

PR #4611, opened 2021-04-19
fix: add separator on breadcrumb-item when a new item is added

### LINKED ISSUE #4420: A separator is not added to Breadcrumb item when new one is appended to the list. fast-foundation > Breadcrumb.ts
**Description**

A separator is not added to Breadcrumb item when new one is appended to the list

This is something I haven't been able to reproduce it using the component explorer editor, so I added a runnable example in stackblitz instead. https://stackblitz.com/edit/fast-design-breadcrumb?file=src/app.ts

This example renders a breadcrumb with three items. 
There is a button, which will add a new item when clicked. That new item will be added at the end of the list, but the separator between the second last item and the last item won't be rendered

Notice that this issue doesn't happen when the component is created, but when the slotted items change.

**Steps to reproduce it**

1. Run the example in stackblitz
2. You should see a breadcrumb with three items and a button to add new items
3. click on the button
4. Notice that a new item has been added as last child but there is no separator between this new item and the pevious one

image
image

**Expected behaviour**

Once a new breadcrumb item is added to the list as last item. I expect to see a separator between the last previous item and the new last item.

* OS Device: MacOS
* Browser: Google Chrome
* Additional context:

BreadcrumbItem.ts: lines 24-28

On the slottedBreadcrumbItemsChanged callback there is code to remove the separator of the last breadcrumb item,
but there is not logic to add the separator when there is a new last item

khamudom:
# Pull Request

## 📖 Description

I created the `setItemSeparator` method that will set the separator to true. I also combined `removeLastItemSeparator` with `setItemSeparator`. This was added to fix the issue where if a breadcrumb-item is dynamically added, a separator will also be added to the item if it is not the last breadcrumb-item.
I also added an example showing this scenario in the fixtures.

### 🎫 Issues

closes #4420 

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
+++ b/packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
@@ -28,6 +28,7 @@ export const BreadcrumbItemStyles = css`
     .listitem {
         display: flex;
         align-items: center;
+        width: max-content;
     }
 
     .separator {
--- a/packages/web-components/fast-components/src/breadcrumb/breadcrumb.stories.ts
+++ b/packages/web-components/fast-components/src/breadcrumb/breadcrumb.stories.ts
@@ -23,7 +23,7 @@ function addItem(): void {
 }
 
 addons.getChannel().addListener(STORY_RENDERED, (name: string) => {
-    if (name.toLowerCase().startsWith("breadcrumb")) {
+    if (name.toLowerCase() === "breadcrumb--breadcrumb") {
         const button = document.getElementById("add-button") as HTMLElement;
         button.addEventListener("click", addItem);
     }
--- a/packages/web-components/fast-components/src/breadcrumb/breadcrumb.stories.ts
+++ b/packages/web-components/fast-components/src/breadcrumb/breadcrumb.stories.ts
@@ -1,6 +1,33 @@
+import { STORY_RENDERED } from "@storybook/core-events";
+import addons from "@storybook/addons";
 import BreadcrumbTemplate from "./fixtures/base.html";
 import "../breadcrumb-item";
 import "./index";
+
+function addItem(): void {
+    const breadcrumbElement = document.getElementById("mybreadcrumb");
+    const items = breadcrumbElement?.querySelectorAll("fast-breadcrumb-item");
+
+    if (items === undefined) {
+        return;
+    }
+    const lastItem = items[items.length - 1];
+    const item: any = document.createElement("fast-breadcrumb-item");
+    item.setAttribute("href", "#");
+    item.textContent = `Breadcrumb item ${items.length + 1}`;
+
+    if (lastItem.parentNode === null) {
+        return;
+    }
+    lastItem.parentNode.insertBefore(item, lastItem.nextSibling);
+}
+
+addons.getChannel().addListener(STORY_RENDERED, (name: string) => {
+    if (name.toLowerCase().startsWith("breadcrumb")) {
+        const button = document.getElementById("add-button") as HTMLElement;
+        button.addEventListener("click", addItem);
+    }
+});
 
 export default {
     title: "Breadcrumb",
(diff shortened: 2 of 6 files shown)


packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -25,11 +25,20 @@ export class Breadcrumb extends FASTElement {
                 this.slottedBreadcrumbItems.length - 1
             ];
 
+            this.resetItemSeparator();
             this.removeLastItemSeparator(lastNode);
```
chrisdholt, line 29: Could these two be combined to be a bit more efficient? Consider that you run through each item and set all separators to true and *then* you go find the last one and change it. Could these be combined into a single method which is more efficient?

chrisdholt (approved):
Approve with one suggestion to combine these methods to be more efficient. Perhaps this can be done while you are iterating each item...

packages/web-components/fast-components/src/breadcrumb-item/breadcrumb-item.styles.ts
```diff
@@ -28,6 +28,7 @@ export const BreadcrumbItemStyles = css`
     .listitem {
         display: flex;
         align-items: center;
+        width: max-content;
```
chrisdholt, line 31: Is this a new addition?

khamudom: yeah, it's an addition to fixing the non-href item. I noticed from running the stackblitz example that if the window was small enough, it would make the item, which is not an href, wrap itself.
image

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
khamudom: yeah, I will combine this

khamudom: updated

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
```diff
@@ -25,15 +25,20 @@ export class Breadcrumb extends FASTElement {
...
-    private removeLastItemSeparator(lastNode: HTMLElement): void {
-        if (lastNode instanceof BreadcrumbItem) {
-            (lastNode as BreadcrumbItem).separator = false;
-        }
+    private setItemSeparator(lastNode: HTMLElement): void {
+        this.slottedBreadcrumbItems.forEach((item: HTMLElement) => {
+            if (item instanceof BreadcrumbItem) {
+                (item as BreadcrumbItem).separator = true;
+            }
+            if (lastNode instanceof BreadcrumbItem) {
```
marjonlynch, line 38: Do you need to do the lastNode check for each iteration of the loop or just once outside the loop?

packages/web-components/fast-components/src/breadcrumb/breadcrumb.stories.ts
```diff
@@ -1,7 +1,34 @@
...
+
+    if (items === undefined) {
+        return;
+    }
+    const lastItem = items[items.length - 1];
+    const item: any = document.createElement("fast-breadcrumb-item");
+    item.setAttribute("href", "#");
+    item.textContent = `Breadcrumb item ${items.length + 1}`;
+
+    if (lastItem.parentNode === null) {
```
marjonlynch, line 19: I know this is just a helper function for your story but why not just say if (lastItem.parentNode !== null) then do the thing rather than have multiple return statements. I guess this is a preference thing so disregard if you don't agree

khamudom: thanks. refactored

packages/web-components/fast-foundation/src/breadcrumb/breadcrumb.ts
khamudom: updated

marjonlynch (approved):

Ibrahimmaga (approved):

PR #4619, opened 2021-04-20
fix: removed disabled style on anchor web component

### LINKED ISSUE #4595: fix: remove disabled stylings in anchor component
# 🐛 Bug Report

I noticed that adding the `disabled` attribute to the anchor component styles it, where it has no disabled logic whatsoever.

You can track the discussion in [discord](https://discord.com/channels/449251231706251264/449251231706251266/832239386195394560)

## 💻 Repro or Code Sample

You can try it [here](https://explore.fast.design/components/fast-anchor)

## 🤔 Expected Behavior

Nothing :)

## 😯 Current Behavior

Anchor has a disabled UI, whereas it has no logic for disabled

## 💁 Possible Solution

Remove the disabled styles inherited from the button component

## 🌍 Your Environment

* OS & Device: OSX
* Browser Chrome
* Version Latest

khamudom:
# Pull Request

## 📖 Description

The anchor component should not have a `disabled` style, so I moved all the disabled selectors out of the common button styles file and added them to the ` button.styles` file. 

### 🎫 Issues

closes #4595 

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/button/button.styles.ts
+++ b/packages/web-components/fast-components/src/button/button.styles.ts
@@ -1,18 +1,117 @@
 import { css } from "@microsoft/fast-element";
 import {
+    disabledCursor,
+    forcedColorsStylesheetBehavior,
+} from "@microsoft/fast-foundation";
+import { SystemColors } from "@microsoft/fast-web-utilities";
+import {
     AccentButtonStyles,
+    accentFillRestBehavior,
     BaseButtonStyles,
     LightweightButtonStyles,
+    neutralFillRestBehavior,
+    neutralFillStealthRestBehavior,
     OutlineButtonStyles,
     StealthButtonStyles,
 } from "../styles/index";
 import { appearanceBehavior } from "../utilities/behaviors";
 
 export const ButtonStyles = css`
-    ${BaseButtonStyles}
+    ${BaseButtonStyles} :host ([disabled]) {
+        opacity: var(--disabled-opacity);
+        background-color: ${neutralFillRestBehavior.var};
+        cursor: ${disabledCursor};
+    }
 `.withBehaviors(
-    appearanceBehavior("accent", AccentButtonStyles),
-    appearanceBehavior("lightweight", LightweightButtonStyles),
-    appearanceBehavior("outline", OutlineButtonStyles),
-    appearanceBehavior("stealth", StealthButtonStyles)
+    forcedColorsStylesheetBehavior(
+        css`
+            :host([disabled]),
+            :host([disabled]) .control {
+                forced-color-adjust: none;
+                background-color: ${SystemColors.ButtonFace};
+                border-color: ${SystemColors.GrayText};
+                color: ${SystemColors.GrayText};
+                cursor: ${disabledCursor};
+                opacity: 1;
+            }
+        `
+    ),
+    appearanceBehavior(
+        "accent",
+        css`
+            ${AccentButtonStyles} :host ([appearance="accent"][disabled]) {
+                background: ${accentFillRestBehavior.var};
+            }
+        `.withBehaviors(
+            forcedColorsStylesheetBehavior(
+                css`
+                    :host([appearance="accent"][disabled]) .control,
+                    :host([appearance="accent"][disabled]) .control:hover {
+                        background: ${SystemColors.ButtonFace};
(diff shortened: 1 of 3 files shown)


scomea (approved):

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -187,21 +171,17 @@ export const AccentButtonStyles = css`
         color: ${accentForegroundCutRestBehavior.var};
     }
 
-    :host([appearance="accent"]:hover) {
+    :host([appearance="accent"]:hover:not([disabled])) {
```
chrisdholt, line 174: I'm slightly concerned at the complexity and specificity of the selectors here. My assumption is that this is due to the fact that we're adding these styles to the base button. What if we shared the styles we're applying to the base button across each of these appearances? Would this complex selector still be needed?

khamudom: I removed the :not pseudo class. what I decided to do was add hover and active on the disabled selectors for each appearance

chrisdholt (approved):

nicholasrice:
Anchor doesn't support the `disabled` content attribute but it is valid to not have an `href` content attribute. I *sort of* consider that the anchor's equivalent of disabled. Is that a case we want to represent with anchor?

khamudom:
> Anchor doesn't support the `disabled` content attribute but it is valid to not have an `href` content attribute. I _sort of_ consider that the anchor's equivalent of disabled. Is that a case we want to represent with anchor?

Not sure if this will answer your question, but this work is to just remove the disabled look to anchor.

chrisdholt:
> Anchor doesn't support the `disabled` content attribute but it is valid to not have an `href` content attribute. I _sort of_ consider that the anchor's equivalent of disabled. Is that a case we want to represent with anchor?

@nicholasrice In this case would we still want to attach styles since this is technically just "text" in that case? I think we're headed towards a rabbit hole where guidance will be our best friend, but I think this is a worth while conversation. I'm curious what others think as well here - would what expected behavior be in this scenario? Styles or no styles? Disabled styles for non-hypertext?

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -187,7 +171,7 @@ export const AccentButtonStyles = css`
         color: ${accentForegroundCutRestBehavior.var};
     }
 
-    :host([appearance="accent"]:hover) {
+    :host([appearance="accent"]:hover:) {
```
robarbms, line 174: Typo?

packages/web-components/fast-components/src/button/button.styles.ts
```diff
@@ -1,18 +1,148 @@
...
+                        color: ${SystemColors.GrayText};
+                    }
+                `
+            )
+        )
+    ),
+    appearanceBehavior(
+        "lightweight",
+        css`
+            :host([appearance="lightweight"][disabled]),
```
robarbms, line 70: This doesn't appear to be needed since it's already declared by the default styles for :host([appearance="lightweight"]) on lines 301-304 in patterns/button.ts

packages/web-components/fast-components/src/button/button.styles.ts
```diff
@@ -1,18 +1,148 @@
...
+            :host([appearance="lightweight"][disabled]:hover),
+            :host([appearance="lightweight"][disabled]:active) {
+                background-color: transparent;
```
robarbms, line 73: background-color: transparent doesn't look like its needed as well. It should be the default for all of these states as they are not changed for :hover or :active pseudo states in patterns/button.ts.

packages/web-components/fast-components/src/styles/patterns/button.ts
khamudom: thanks. a PR will up shortly to fix this

packages/web-components/fast-components/src/button/button.styles.ts
khamudom: this is still needed to overwrite the hover and active state on base button

packages/web-components/fast-components/src/button/button.styles.ts
khamudom: yes, this is not needed

PR #4623, opened 2021-04-21
feat: refactor color recipes

### LINKED ISSUE #3833: Adaptive Color System Refactor
# FAST Color Refactor
The issue outlines proposed changes to how the color system works in FAST.

The primary changes are to the data types used by color algorithms, but there are also some necessary changes to the DesignSystemProvider to make the recipe implementations more efficient.

## Goals
- Improve performance by minimizing color parsing and stringifying
- Open the system up to use *any* color model, not just hex codes.
- Associate palette source directly with palette
    - Should avoid confusion of accent-palette and accent-base-color needing to be kept in sync
- Simplify implementation of contrast-based functions and make them easier to leverage
- Color recipes should be more portable by not requiring the entire design-system as an argument - should only expect what data they need.

## Non-goals
This proposal doesn't address *how* recipes get registered to the provider, nor how to replace implementations of recipes used by components for all or some components. https://github.com/microsoft/fast/issues/3361 is a better place to discuss those more general features.
 
### Swatch
A swatch is a simple color structure representing a color that can be used in CSS. It has the following interface:
```ts
interface Swatch {
    /**
     * The relative luminance of the swatch
     */
    readonly relativeLuminance: number; // between 0 and 1
    
    /**
     * A string representing the color that can be used in CSS
     */
    readonly css: string;
    
    /**
     * Returns the contrast ratio between this swatch
     * and a reference swatch
     */
    static contrast(color: Swatch): number; 
}
```

### Palette
A palette is a structure that representing a collection of `Swatch` objects derived from or related to a single source Swatch. It also provides several methods for finding swatches relative to other swatches.

```ts
interface Palette {
    /**
     * The color that the palette is derived from
     */
    readonly source: Swatch;
    
    /**
     * All swatches in the palette
     */
    readonly swatches: Swatch[];
    
    /**
     * Returns a swatch from the palette that most closely matches
     * the contrast ratio provided to a provided reference.
     * 
     * @remarks
     * We need to know where to start searching
     * and which direction to search in - isDarkMode?
     */
    matchContrast(reference: Swatch, contrast: number): Swatch;
    
    /**
     * Returns the index of the palette that most closely matches
     * the relativeLuminance of the provided swatch
     */
    closestIndexOf(reference: Swatch): number;
     
     /**
      * Gets a swatch by index. Index is clamped to the limits
      * of the palette so a Swatch will always be returned.
      */
    get(index: number): Swatch;
}
```

## Color Recipes
The most significant change to color recipes is that they will not memoize the design system object itself, rather the provided arguments.

Additionally, they will all need to be re-factored to use the new data structures.

### DesignTokenResolver
A `DesignTokenResolver` type will be added. It is a simple structure that is bound to a provider and observes dependent properties. When the properties change, the value field is updated.
```ts
interface DesignSystemResolver<T, S> {
  observedProperties: Array<keyof T>; // The properties on the context the resolver cares about
  value: string;
  name: string;
  bind(context: T): void; // Hooks up property observation and sets value
  unbind(context: T): void; // Remove property observation
}
```

Color recipes will have `DesignSystemResolver` implementations, where `value` is updated to be the product of the recipe and the `observedProperties` are the properties of the DesignSystem that are necessary to calculate the recipe.

When a resolver attaches to a context, all properties in the `observedProperties` list on the context will be observed with the reaction to observed property changes updating the `value` field.
> `observedProperties` could probably be an implementation detail, I'm not sure a explicit property is needed

The resolver is bound to the provider, and the provider subscribes to changes to the resolvers's `value` field. When the value changes, the corresponding CSS Custom Property is updated.

nicholasrice:
# Pull Request

## 📖 Description
This PR refactors the existing color recipes loosely following https://github.com/microsoft/fast/issues/3833. Doing so decouples the recipe implementations from the `FASTDesignSystem` interface in favor of specific granular inputs.

It adds two key types, `Palette` and `Swatch` that are used by recipes. A `Palette` represents a collection of `Swatch` object, and a `Swatch` represents a color in a `Palette`. Two concrete implementations of these base types are also added, `PaletteRGB` and `SwatchRGB`, which are used by the color recipes.

closes #3833

### General

### Component-specific


packages/web-components/fast-components/src/color-vNext/palette.ts
```diff
@@ -0,0 +1,148 @@
...
+ * A {@link Palette} representing RGB swatch values.
+ * @public
+ */
+export class PaletteRGB implements Palette<SwatchRGB> {
+    /**
+     * {@inheritdoc Palette.source}
+     */
+    public readonly source: SwatchRGB;
+    public readonly swatches: ReadonlyArray<SwatchRGB>;
+    private lastIndex: number;
```
bheston, line 53: Minor point, but what about 'count' or 'length' or something instead? I keep reading this as "previous index" I think because I know there's some ramp-walking logic involved here.

packages/web-components/fast-components/src/color-vNext/palette.ts
```diff
@@ -0,0 +1,148 @@
...
+        return binarySearch(source, condition, startSearchIndex, endSearchIndex);
+    }
+
+    /**
+     * {@inheritdoc Palette.get}
+     */
+    public get(index: number): SwatchRGB {
+        return (
+            this.swatches[index] ||
+            this.swatches[clamp(index, 0, this.swatches.length - 1)]
```
bheston, line 106:
```suggestion
            this.swatches[clamp(index, 0, this.lastIndex)]
```

packages/web-components/fast-components/src/color-vNext/recipes/neutral-divider.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * The neutralDivider color recipe
+ * @param palette - The palette to operate on
+ * @param background - The background of the divider
+ * @param offset - The offset from the background
+ *
+ * @internal
+ */
+export function neutralDivider(palette: Palette, background: Swatch, offset: number) {
```
bheston, line 13:
```suggestion
export function neutralDivider(palette: Palette, reference: Swatch, delta: number) {
```

packages/web-components/fast-components/src/color-vNext/recipes/neutral-divider.ts
```diff
@@ -0,0 +1,17 @@
...
+    const direction = directionByIsDark(background);
```
bheston, line 14:
```suggestion
    const direction = directionByIsDark(reference);
```

packages/web-components/fast-components/src/color-vNext/recipes/neutral-divider.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    return palette.get(palette.closestIndexOf(background) + direction * offset);
```
bheston, line 16:
```suggestion
    return palette.get(palette.closestIndexOf(reference) + direction * offset);
```

bheston, line 16:
```suggestion
    return palette.get(palette.closestIndexOf(reference) + direction * offset);
```

bheston, line 16:
```suggestion
    return palette.get(palette.closestIndexOf(reference) + direction * delta);
```

packages/web-components/fast-components/src/color-vNext/recipes/neutral-fill.ts
```diff
@@ -0,0 +1,34 @@
...
+export function neutralFill(
+    palette: Palette,
+    reference: Swatch,
+    restDelta: number,
+    hoverDelta: number,
+    activeDelta: number,
+    focusDelta: number,
+    selectedDelta: number
+) {
+    const backgroundIndex = palette.closestIndexOf(reference);
```
bheston, line 23:
```suggestion
    const referenceIndex = palette.closestIndexOf(reference);
```

packages/web-components/fast-components/src/color-vNext/recipes/neutral-foreground-hint.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Palette } from "../palette";
+
+/**
+ * The neutralForegroundHint color recipe
+ * @param palette - The palette to operate on
+ * @param background - The background of the divider
+ *
+ * @internal
+ */
+export function neutralForegroundHint(palette: Palette, background: Swatch) {
```
bheston, line 11:
```suggestion
export function neutralForegroundHint(palette: Palette, reference: Swatch) {
```

packages/web-components/fast-components/src/color-vNext/recipes/neutral-outline.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export function neutralOutline(
+    palette: Palette,
+    reference: Swatch,
+    restDelta: number,
+    hoverDelta: number,
+    activeDelta: number,
+    focusDelta: number
+) {
+    const backgroundIndex = palette.closestIndexOf(reference);
```
bheston, line 16:
```suggestion
    const referenceIndex = palette.closestIndexOf(reference);
```

bheston (approved):
All the suggestions are just for consistency. Recipes look good as a first step toward the DI style work.

EisenbergEffect (approved):
Really excited about this work. @nicholasrice Can you add documentation for this like you did for the design tokens? Even better if you can connect the dots between the two.

packages/web-components/fast-components/src/color-vNext/palette.ts
nicholasrice: The 'endIndex' is referring to where to stop looking in the array, not necessarily the *last* index in the array.  The item at that index is commonly referred to as ['last'](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop) so I think I'll leave this as-is. It also isn't part of the public API, so it doesn't affect consumers.

janechu (approved):

nicholasrice:
@EisenbergEffect I documented to Palette and Swatch interfaces and implementations but not the recipes themselves since those are actually internal. We'll end up exporting a function that wraps the recipes that implements the `DerivedDesignToken` type, which I think we can document use of.

EisenbergEffect:
Sounds good!

PR #4651, opened 2021-04-28
feat: create color design tokens

nicholasrice:
# Pull Request

## 📖 Description
This PR integrates new color recipes into DesignTokens to be used in CSS. It replaces all instances of color behaviors in components in `@microsoft/fast-components`

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
+++ b/packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
@@ -11,13 +11,12 @@ import {
     density,
     designUnit,
     focusOutlineWidth,
+    neutralFocus,
     outlineWidth,
     typeRampMinus1FontSize,
     typeRampMinus1LineHeight,
 } from "../design-tokens";
 import {
-    neutralDividerRestBehavior,
-    neutralFocusBehavior,
     neutralForegroundActiveBehavior,
     neutralForegroundFocusBehavior,
     neutralForegroundHoverBehavior,
@@ -84,9 +83,9 @@ export const accordionItemStyles = (context, definition) =>
 
     .button:${focusVisible}::before {
         outline: none;
-        border: calc(${outlineWidth} * 1px) solid ${neutralFocusBehavior.var};
+        border: calc(${outlineWidth} * 1px) solid ${neutralFocus};
         box-shadow: 0 0 0 calc((${focusOutlineWidth} - ${outlineWidth}) * 1px)
-            ${neutralFocusBehavior.var};
+            ${neutralFocus};
     }
 
     :host([expanded]) .region {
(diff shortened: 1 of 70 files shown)


EisenbergEffect:
How was the perf looking in the end? Any additionally follow-up work planned for that?

nicholasrice:
@EisenbergEffect the color recipes themselves seem fine for perf, but the frequent resolution of token values (colors often depend on lots of tokens, and each element often depends on lots of colors) is causing a significant perf regression. The problem grows linearly with the number of elements. I'm going to hold off on merging this until I can do a deeper investigation and find a resolution.

EisenbergEffect:
Could we store the token values themselves in a central storage service, but have it be something that is created on the fly as needed? For example, set up a singleton storage on the element where the design system is initialized and have everything reference that. But, if someone needs to set a token value somewhere on a child node, we dynamically create a new storage service on that node and switch the system over from singleton mode to dom hierarchy mode and then have consumers locate the storage system in the same way that it located the design system provider in the past.

nicholasrice:
We *sort of* do that already but there might be some short-circuiting we can do during resolution. The core use-case is for tokens to be set *frequently* throughout the DOM hierarchy (think fill/background color & palettes) so in practice the system will enter the more inefficient mode rather quickly. I do have it set up where each token is isolated though, so we could see some performance gains if we only branch the overridden token to that mode.

EisenbergEffect:
I can't remember the internal implementations details, so forgive me if this is rehashing or we've already tried this. But, what if we only acted on "write" of a token and never on "read". The idea would be that the token simply injects the CSS variable and nothing else. Then, the design system writes all the token values, performing all calculations. We do this as part of the initial registration. Then, if someone writes a new token value to a DOM node, we publish an event up the tree to find the parent node where tokens were written. We then pull down all the tokens and write them all to new the node where the new value is. That would cause all css variables to trickle down. Would something like this work? My thought is that the writing of custom properties is less frequent than the need to access them in styles. And then maybe we establish some practices/patterns related to this. So, for example, with cards, we encourage folks to override the token on the parent node that contains the list of cards, rather than on individual cards, so we get less frequent needs to write all tokens. We might also be able to reduce the number of events to find token storage by doing a fast path to first check the parentNode before firing an event. Just tossing ideas out...

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -58,17 +52,17 @@ export const accordionItemStyles = (context, definition) =>
...
 
     .button:hover {
-        color: ${neutralForegroundHoverBehavior.var};
+        color: ${neutralForegroundRest};
     }
 
     .button:active {
-        color: ${neutralForegroundActiveBehavior.var};
+        color: ${neutralForegroundRest};
     }
```
chrisdholt, line 66: @nicholasrice should these all be neutralForegroundRest or should we also have Hover and Active here?

nicholasrice: In practice `neutralForeground` never actually had hover/active states. They were all the same, so I just made a single neutralForegroundRest recipe.

chrisdholt: makes sense!

chrisdholt (approved):

nicholasrice:
I ended up adding an optional argument for Fill recipes to provide a the swatch context to calculate from. If no swatch is provided, the fillColor is retrieved from the target element. This is ready for review now; In a follow up PR I would go through and delete old assets and update documentation.

EisenbergEffect (approved):

PR #4665, opened 2021-05-04
feat: improve design token performance

nicholasrice:
# Pull Request

## 📖 Description
This PR addresses significant performance issues uncovered with the initial implementation of `DesignTokens`, specifically removing the behavior that using a `DesignToken` as a `CSSDirective` causes value resolution for *each token instance in the stylesheet*. It removes the `addCustomPropertyFor()` and `removeCustomPropertyFor()` methods from `DesignToken`.

Instead, CSS custom properties are written for an element in two cases:
1. A value is *set* for an element via `DesignToken.setValueFor()`
2. A token that some other `DesignToken` *depends* on is set via `DesignToken.setValueFor()`

Setting CSS custom properties in these two cases ensures that all a DesignToken can always be used as a CSS directive, and using the CSSDirective simply interpolates the CSS variable string (no value resolution).

A few things notable:
1. Because using a `DesignToken` is an inert action, we don't really know *where* tokens are being used in CSS. To guarantee any usage anywhere is accurate, we must write *all* CSS custom properties that depend on another token value whenever that value is set. For instance, all neutral recipes depend on the `neutralPalette` `DesignToken`. If `neutralPalette` is set for some arbitrary node, *all neutral recipes are emitted to CSS custom properties for that node*. This is an area we can do some performance work on; it could be more performant to do a *little* work when a `DesignToken` is used in a stylesheet and only write out the token when we know that the CSS custom property is actually being used.
2. The decision-tree to fully reconcile when and where CSS custom properties should exist is complex and not fully implemented. Tokens are set for arbitrary elements and dependency graphs for derived token values can and do change throughout the tree. Token values can have *many* dependencies, making it hard to track exactly which dependent token is causing a dependee to write out it's CSS custom property (and thus when to *remove* the CSS custom property). This implementation largely avoids tearing down CSS custom properties for elements after they've been required at some point in time, instead relying on the value resolution mechanism to ensure the *value* of the CSS custom property is always correct (even if it is no longer necessary). All this really means is that we can get into a case where a CSS custom property is set to the same value for a parent/child element pair when the CSS custom property only existing on the parent would suffice. In practice I don't think this will happen that often (It can happen if a dependent token is set and then deleted for a node).

**Todo:** 
- [ ] Implement mechanism to sync the defaultElement with the DesignSystem's element context.
- [ ] Add ability for a token to opt out of CSS custom property creation (neutralPalette, accentPalette)
- [x] Set up inspection for `rawValue` so that we can determine if the node for which a dependent token is set is still the same as the token value that has the dependency.
- [x] Ensure we only explicitly write the CSS custom property once for a node, and otherwise rely on `value` change handlers to update the property.
- [ ] Catch when a token being set is using the same token to derive the value. When that happens, use the parent element.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/design-token/README.md
+++ b/packages/web-components/fast-foundation/src/design-token/README.md
@@ -4,7 +4,7 @@
 The FAST Design Token implementation is designed to provide first-class support for Design Tokens and make setting, getting, and using Design Tokens simple.
 
 ## What is a Design Token
-A Design Token is a semantic, named variable used to describe a Design System. They often describe design concepts like typography, color, sizes, UI spacing, etc. FAST encourages checking out [https://github.com/design-tokens/community-group#design-tokens](the Design Tokens Community Group) for more information on Design Tokens themselves.
+A Design Token is a semantic, named variable used to describe a Design System. They often describe design concepts like typography, color, sizes, UI spacing, etc. FAST encourages checking out [the Design Tokens Community Group](https://github.com/design-tokens/community-group#design-tokens) for more information on Design Tokens themselves.
 
 ## Using Design Tokens in FAST
 This usage guide walks through creating and using a Design Token *background-color*, which is represented in code as a hexadecimal color string.
@@ -54,23 +54,7 @@ A default value can be set for a token, so that the default value is returned fr
 fillColor.withDefault("#FFFFFF");
 ```
 ### 6. Emit a token to a CSS Custom Property
-A Design Token can be made available in CSS through CSS custom properties. The custom property value will be set to the token's value for the supplied target element.
-
-```ts
-fillColor.addCustomPropertyFor(descendent); // --fill-color: #FFFFFF;
-```
-
-If the value of the token *changes* for the target element, the CSS custom property will be updated to the new value:
-
-```ts
-fillColor.setValueFor(descendent, "#F7F7F7"); // --fill-color: #F7F7F7;
-```
-
-The CSS custom property can also be removed through a parallel method:
-
-```ts
-fillColor.removeCustomPropertyFor(descendent);
-```
+DesignToken emits a token to CSS automatically whenever the value is set for an element. In the case when a DesignToken is assigned a (derived value)[#derived-design-token-values], the CSS custom property will also be emitted when any dependent tokens change.
 
 #### Values with a 'createCSS' method
 It is sometimes useful to be able to set a token to a complex object but still use that value in CSS. If a DesignToken is assigned a value with a `createCSS` method on it, the product of that method will be used when emitting to a CSS custom property instead of the Design Token value itself:
@@ -92,7 +76,6 @@ const value = {
     }
 }
 fancyFillColor.setValueFor(descendent, value)
-fancyFillColor.addCustomPropertyFor(descendent); // --fancy-fill-color: rgb(255, 0, 0);
 ```
 
 ## Using Design Tokens in CSS
--- a/packages/web-components/fast-foundation/src/design-token/design-token.spec.ts
+++ b/packages/web-components/fast-foundation/src/design-token/design-token.spec.ts
@@ -344,10 +344,7 @@ describe("A DesignToken", () => {
 
             expect(token.getValueFor(target)).to.equal(12)
             
-            // TODO: should it throw here? It throws because writing the CSS custom property
-            // requires inspecting the value, which can't resolve (see expect below). Seems a bit
-            // odd that just deleting a value would throw though.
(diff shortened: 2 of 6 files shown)


EisenbergEffect:
Thanks for the notes @nicholasrice and for the hard work on coming up with a performant implementation. Neither of the notable items seem like big concerns. Do you want/need to implement everything on the todo list before merging or are some of the items options/doable at a later time?

nicholasrice:
I want to at least tackle the middle 3 with this PR. The other two are isolated features that can be done as a followup

EisenbergEffect:
One other question: how should we treat this related to the color tokens PR? Is this in addition to that? Is the other ready to move out of draft mode? Let me know what you want me to look at first and I'll get on it.

nicholasrice:
This should probably go in first. There are a few small changes that should happen in that PR in response to this one

chrisdholt (approved):

packages/web-components/fast-foundation/src/design-token/README.md
```diff
@@ -4,7 +4,7 @@
 The FAST Design Token implementation is designed to provide first-class support for Design Tokens and make setting, getting, and using Design Tokens simple.
 
 ## What is a Design Token
-A Design Token is a semantic, named variable used to describe a Design System. They often describe design concepts like typography, color, sizes, UI spacing, etc. FAST encourages checking out [https://github.com/design-tokens/community-group#design-tokens](the Design Tokens Community Group) for more information on Design Tokens themselves.
+A Design Token is a semantic, named variable used to describe a Design System. They often describe design concepts like typography, color, sizes, UI spacing, etc. FAST encourages checking out [the Design Tokens Community Group](https://github.com/design-tokens/community-group#design-tokens) for more information on Design Tokens themselves.
```
EisenbergEffect, line 7: Love the callout to the CG here.

packages/web-components/fast-foundation/src/design-token/README.md
```diff
@@ -54,23 +54,7 @@ A default value can be set for a token, so that the default value is returned fr
...
-```ts
-fillColor.setValueFor(descendent, "#F7F7F7"); // --fill-color: #F7F7F7;
-```
-
-The CSS custom property can also be removed through a parallel method:
-
-```ts
-fillColor.removeCustomPropertyFor(descendent);
-```
+DesignToken emits a token to CSS automatically whenever the value is set for an element. In the case when a DesignToken is assigned a (derived value)[#derived-design-token-values], the CSS custom property will also be emitted when any dependent tokens change.
```
EisenbergEffect, line 57: I think our community will really appreciate this explanation. Great!

packages/web-components/fast-foundation/src/design-token/design-token.spec.ts
```diff
@@ -1,12 +1,12 @@
...
-import { css, DOM, FASTElement, Observable } from "@microsoft/fast-element";
+import { css, DOM, FASTElement, html, Observable } from "@microsoft/fast-element";
 import { expect } from "chai";
 import { DesignSystem } from "../design-system";
 import { FoundationElement } from "../foundation-element";
 import { DesignToken } from "./design-token";
 
 new DesignSystem().register(
-    FoundationElement.compose({ type: class extends FoundationElement { }, baseName: "custom-element" })()
+    FoundationElement.compose({ type: class extends FoundationElement { }, template: html`<slot></slot>`, baseName: "custom-element" })()
```
EisenbergEffect, line 9: I think you'll need to update this code after you update the branch since the DesignSystem API changed a bit. See the DesignSystem tests for some examples. Shouldn't be difficult to update.

packages/web-components/fast-foundation/src/design-token/design-token.ts
```diff
@@ -404,22 +418,37 @@ class DesignTokenNode<T> {
...
+        this.handleChange = noop as () => void;
+        this._rawValue = value;
+
+        if (!this.useCSSCustomProperty) {
+            this.useCSSCustomProperty = true;
         }
 
-        this._rawValue = value;
+        if (this.bindingObserver) {
+            const records = this.bindingObserver.records();
```
EisenbergEffect, line 429: Nifty 😄

EisenbergEffect (approved):
Great work @nicholasrice ! This was a challenging project that required the harmonization of several complex feature requirements with ergonomic APIs, all while overcoming performance challenges. As a bonus, you've managed to implement this in a very readable and well-factored way. This is going to be a huge benefit to our design systems over the long haul. Congrats. 🎊

PR #4666, opened 2021-05-04
feat: Anchored region uses transform for positioning

scomea:
## 📖 Description

This change modifies anchored region to use transforms for positioning rather than top/right/bottom/left. This drastically improves performance, makes the scaling experience better by getting rid of elements being anchored to the reverse edge and makes the math a bit simpler too.

I also removed the updateAnchorOffset() call from the api as it seemed extraneous (this change would be part of a version bump).

The storybook sample page has also been beefed up to provide clearer examples and some explanations. It has way more auto-tracking regions active than I'd expect to see in a page so how it performs is a point of interest. Note that you can scroll to the bottom and click to open a bunch of sample menus.

## 📑 Test Plan
Anchored region needs a bunch of PW tests added.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
@@ -6,7 +6,10 @@ import type { AnchoredRegion } from "./anchored-region";
  * The template for the {@link @microsoft/fast-foundation#(AnchoredRegion:class)} component.
  * @beta
  */
-export const AnchoredRegionTemplate: ViewTemplate<AnchoredRegion> = html`
+export const anchoredRegionTemplate: (
+    context,
+    definition
+) => ViewTemplate<AnchoredRegion> = (context, definition) => html`
     <template class="${x => (x.initialLayoutComplete ? "loaded" : "")}">
         ${when(
             x => x.initialLayoutComplete,
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
@@ -10,7 +10,7 @@ export const anchoredRegionTemplate: (
     context,
     definition
 ) => ViewTemplate<AnchoredRegion> = (context, definition) => html`
-    <template class="${x => (x.initialLayoutComplete ? "loaded" : "")}">
+    <template class="${x => (x.regionVisible ? "loaded" : "")}">
         ${when(
             x => x.initialLayoutComplete,
             html<AnchoredRegion>`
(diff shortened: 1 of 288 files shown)


packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
```diff
@@ -10,7 +10,7 @@ export const anchoredRegionTemplate: (
     context,
     definition
 ) => ViewTemplate<AnchoredRegion> = (context, definition) => html`
-    <template class="${x => (x.initialLayoutComplete ? "loaded" : "")}">
+    <template class="${x => (x.regionVisible ? "loaded" : "")}">
```
chrisdholt, line 13: do we need the prefix of `region` here since this is on the region itself?

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
```diff
@@ -359,62 +360,51 @@ export class AnchoredRegion extends FoundationElement {
     @observable
     public initialLayoutComplete: boolean = false;
 
+    /**
+     * indicates that region contents are now loaded and visible
+     *
+     * @internal
+     */
+    @observable
+    public regionVisible: boolean = false;
```
chrisdholt, line 369:
```suggestion
    public visible: boolean = false;
```

scomea: "visible" is already a property on html element with a different meaning, so I don't think having an identical name is a good idea.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
scomea: There's already a 'visible' prop with a different meaning.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.ts
chrisdholt: Do you have a link to this? I’m not seeing this on the generic element or HTML element properties.

scomea: And note that is labelled as "@internal"

scomea: Sorry, I was apparently confused about the existence of a visible prop.

scomea: I think "contentRendered" might be a more accurate name for this.

packages/web-components/fast-foundation/src/anchored-region/anchored-region.template.ts
scomea: went with "contentRendered" instead

PR #4667, opened 2021-05-04
Small Modifications to Improve Startup Performance

EisenbergEffect:
# Pull Request

## 📖 Description

This PR contains a few small fixes aimed at reducing library size and improving startup.

* The `ElementStyles` API defined instance `withKey` and static `find` methods for storing and looking up styles by key. This feature was not used by us and was an exploratory feature that is now superseded by DI capabilities. This PR removes that code. The API was unfortunately not marked as `@alpha`, which is my fault. That said, I think there's little risk in removing this in a minor release. I have queried the community to see if anyone is using it. I think it's unlikely since the API isn't explained in our documentation and again, it's not used by us.
* In the `AdoptedStyleSheetsStyles` implementation, the logic to construct stylesheets has been moved from the constructor to a property that lazily creates the stylesheets. This should improve startup time specifically in the case where an experience has a number of components, styles, or behaviors that are included in the bundle but are not needed for the very first render. Imagine you have 10 components in your bundle, but only 5 are needed for the first paint. This change will allow the second 5 components to delay their stylesheet creation until they are actually rendered.
* In the `Controller` we were previously using the `observable` decorator to make the `isConnected` property observable. This was the only use of a decorator in `fast-element`. I switched this to a manually implemented getter/setter pair in order to prevent the TS compiler from injecting the runtime code needed for this scenario. The result was an approximate .6kb reduction in library size *after* minification and gzipping.

## 👩‍💻 Reviewer Notes

Changes are isolated to `element-styles.ts` and `controller.ts`. The changes to styles are primarily a move of code from the constructor to a property. The change to the controller is just a shift from decorator to getter/setter

## 📑 Test Plan

No new tests were added. The goal here was to make an implementation change to reduce size and improve perf without changing the functionality.

### General

## ⏭ Next Steps

These changes were just small items that popped out at me in the course of other work. I don't have any other items like this now but I'll continue to make small improvements like this as they occur to me.

Note: We do have a strategic emphasis on improving performance. Most of those are larger tasks, not low-hanging fruit like the changes in this PR.


packages/web-components/fast-element/docs/api-report.md
```diff
@@ -171,7 +171,8 @@ export class Controller extends PropertyChangeNotifier {
     readonly element: HTMLElement;
     emit(type: string, detail?: any, options?: Omit<CustomEventInit, "detail">): void | boolean;
     static forCustomElement(element: HTMLElement): Controller;
-    readonly isConnected: boolean;
```
nicholasrice, line 174: We don't want `isConnected` publicly assignable do we? Should we shadow this so there is only a public getter?

EisenbergEffect: What do you think is the best way to accomplish that from a TS perspective?

nicholasrice (approved):
Just one question on adding the public `isConnected` getter, otherwise approved.

packages/web-components/fast-element/docs/api-report.md
nicholasrice: It would involve a third property, but I think you could use an IIFE to hide it from the private API as well. I think this would work.

```ts
private _isConnected: boolean;
public get isConnected(): boolean {
    return this._isConnected;
}
constructor() {
    // other stuff
        (() => {
            let connected = false;
            Object.defineProperty(this, "_isConnected", {
                get() {
                    Observable.track(this, "isConnected");
                    return connected
                },
                set(value: boolean) {
                    Observable.notify(this, "isConnected");
                    connected = value;
                },
            })
        })()
}
```

EisenbergEffect: That makes me feel a bit uncomfortable. I'm wondering if I can use `@internal` on a setter only? TS wouldn't let me make the property setter private...but I think they enabled that in a later version of TS. Let me play around with it a bit.

chrisdholt (approved):

packages/web-components/fast-element/docs/api-report.md
EisenbergEffect: Nope...doesn't look like the `@internal` trick will work.

EisenbergEffect: I think we are trying too hard 🤣  How about this?

```ts
public get isConnected(): boolean {
  Observable.track(this, "isConnected");
  return this._isConnected;
}

private setIsConnected(value: boolean) {
  this._isConnected = value;
  Observable.notify(this, "isConnected");
 }
```

nicholasrice: I'm good with whatever so long as we're not saying `isConnected` is publicly mutable :)

chrisdholt (approved):

PR #4668, opened 2021-05-04
 fix: select a11y issues

m4thieulavoie:
# Pull Request

## 📖 Description

This PR addresses a few a11y issues that axe raised in an accessibility audit I did today. To give a bit of context, I went from 62 to 42 axe issues in the Select storybook page.

### General

### Component-specific


packages/web-components/fast-foundation/src/select/select.template.ts
```diff
@@ -23,16 +23,16 @@ export const SelectTemplate: ViewTemplate<Select> = html`
         <div
             aria-activedescendant="${x => (x.open ? x.ariaActiveDescendant : null)}"
             aria-controls="listbox"
-            aria-expanded="${x => x.ariaExpanded}"
             aria-haspopup="listbox"
+            aria-labelledby="selected-value"
             class="control"
             part="control"
-            role="button"
+            role="textbox"
```
m4thieulavoie, line 30: Axe was looking for a child with the `textbox` role (as per the [combobox spec](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html))

packages/web-components/fast-foundation/src/select/select.template.ts
```diff
@@ -23,16 +23,16 @@ export const SelectTemplate: ViewTemplate<Select> = html`
         <div
             aria-activedescendant="${x => (x.open ? x.ariaActiveDescendant : null)}"
             aria-controls="listbox"
-            aria-expanded="${x => x.ariaExpanded}"
             aria-haspopup="listbox"
+            aria-labelledby="selected-value"
```
m4thieulavoie, line 27: This DOM element was requiring a label (or `aria-labelledby`) in order to be a11y compliant

packages/web-components/fast-foundation/src/select/select.template.ts
chrisdholt: Ah - @m4thieulavoie I think Axe is outdated. The combobox from ARIA 1.1 and 1.2 are significantly different. Should this be `combobox` rather than textbox? This doesn't fully support free input of text, so I'm not sure that textbox is accurate. Technically I think this is the select-only combobox: https://w3c.github.io/aria-practices/#examples-2

packages/web-components/fast-foundation/src/select/select.template.ts
chrisdholt: I think this is fine, but we should run this via ARIA 1.2 rather than 1.1 as the implications are significant. Is there a way for you to validate against ARIA 1.2 in AXE instead of 1.1?

packages/web-components/fast-foundation/src/select/select.template.ts
m4thieulavoie: Oh, let me check the axe version. However, the host already has the `combobox` role. So perhaps `button` was the right one here

m4thieulavoie: @chrisdholt you were absolutely right. Axe has [this opened issue](https://github.com/dequelabs/axe-core/issues/2505) to add support for the 1.2 combobox.
I'll close this PR as it seems to be a false alarm on our end!

chrisdholt: All good! Once that AXE update is out, we'd love you to take another pass :)

PR #4673, opened 2021-05-05
doc: update the component spec and establish new process checkpoints

EisenbergEffect:
# Pull Request

## 📖 Description

This PR updates our component spec template with two important additions:

1. We want to explicitly incorporate the visual design work as part of the overall spec process.
2. We want to formalize our process a bit more, reflected in the checkpoints called out in the Process section.

## 👩‍💻 Reviewer Notes

Please re-read the entire spec template and consider it in its entirety.

### General

### DIFF
--- a/specs/template.md
+++ b/specs/template.md
@@ -1,5 +1,13 @@
 # Spec Title
 
+## Process
+
+- [ ] **Checkpoint 1** - Schedule a spec review meeting to cover the Overview section before proceeding to work on the Design section. Be sure to include relevant stakeholders such as members of the tools team.
+
+- [ ] **Checkpoint 2** - Schedule a spec review meeting before beginning the implementation work. Be sure to invite relevant stakeholders such as design system owners.
+
+- [ ] **Checkpoint 3** - Submit the component PR and schedule a component code review meeting or present the component in one of our Engineering Open Mic sessions. Address feedback from the review, obtain final approvals through GitHub, and merge.
+
 ## Overview
 
 *The name of the component, along with a high-level description.*
@@ -10,7 +18,7 @@
 
 ### Use Cases
 
-*Primary use cases for this component. These should be based on typical patterns seen in the industry, requests from customers and feedback from tooling.*
+*Primary use cases for this component. These should be based on typical patterns seen in the industry and requests from customers.*
 
 ### Non-goals
 
@@ -111,7 +119,7 @@ Outline the component structure with a diagram of its visual tree (shadow dom). 
 
 ### Test Plan
 
-*What is the plan for testing the component, if different from the normal path? Note that the normal plan includes unit tests for basic state/behavior as well as e2e tests to validate the specific user stories described above.*
+*What is the plan for testing the component, if different from the normal path? Note that the normal plan includes unit tests for basic state/behavior as well as end-to-end tests to validate the specific user stories described above.*
 
 ### Tooling
 
@@ -122,16 +130,7 @@ Outline the component structure with a diagram of its visual tree (shadow dom). 
 *What additions or changes are needed for user documentation and demos? Are there any architectural/engineering docs we should create as well, perhaps due to some interesting technical challenge or design decisions related to this component?*
 
 ---
-
-## Process
-
-### Checkpoints
-
-- [ ] **Checkpoint 1** - Schedule a spec review meeting to cover the Overview section before proceeding to work on the Design section. Be sure to include relevant stakeholders such as members of the tools team.
-
-- [ ] **Checkpoint 2** - Schedule a spec review meeting before beginning the implementation work. Be sure to invite relevant stakeholders such as design system owners.
-
-- [ ] **Checkpoint 3** - Submit the component PR and schedule a component code review meeting or present the component in one of our Engineering Open Mic sessions. Address feedback from the review, obtain final approvals through GitHub, and merge.
+## Appendix
 
 ### Resources
 
--- a/specs/template.md
+++ b/specs/template.md
@@ -10,7 +10,7 @@
 
 ### Use Cases
 
-*Primary use cases for this component. These should be based on typical patterns seen in the industry, requests from customers (internal and external), and feedback from tooling.*
+*Primary use cases for this component. These should be based on typical patterns seen in the industry, requests from customers and feedback from tooling.*
 
 ### Non-goals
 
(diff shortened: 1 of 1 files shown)


specs/template.md
```diff
@@ -49,15 +49,21 @@
...
 - *Slot Names*
 - *Host Classes*
 - *Slotted Content/Slotted Classes*
 - *CSS Parts*
 
+*Work closely with the visual design partner to co-develop the API and anatomy along side the visual design.
+
+### Visual Appearance
+
+*Provide Figma files and other design assets. Be sure to account for the various component states, including hover, active, etc. as well as validity, and appearance variants. Consider all officially supported design systems as well as known community scenarios as appropriate. Work closely with engineering to co-develop the visual design along side the API and anatomy.*
```
chrisdholt, line 65: I may want to tweak this a bit for nuance. Technically the only officially supported design system from a visual appearance standpoint is our own. I think perhaps the intent here is to ensure that our template and class are flexible to accommodate other design systems (which may be included as "prior art" or industry examples above), but I don't think we should try and design for systems the community has *extended* off FAST. I'm not sure there's a way for us to accommodate that.

specs/template.md
```diff
@@ -117,10 +123,20 @@
...
+## Process
+
+### Checkpoints
+
+- [ ] **Checkpoint 1** - Schedule a spec review meeting to cover the Overview section before proceeding to work on the Design section. Be sure to include relevant stakeholders such as members of the tools team.
+
+- [ ] **Checkpoint 2** - Schedule a spec review meeting before beginning the implementation work. Be sure to invite relevant stakeholders such as design system owners.
+
+- [ ] **Checkpoint 3** - Submit the component PR and schedule a component code review meeting or present the component in one of our Engineering Open Mic sessions. Address feedback from the review, obtain final approvals through GitHub, and merge.
+
```
chrisdholt, line 135: What would you think about moving this towards the top? So at a glance, folks can see where this is "in process"

specs/template.md
```diff
@@ -10,11 +10,11 @@
 
 ### Use Cases
 
-*Primary use cases for this component.*
+*Primary use cases for this component. These should be based on typical patterns seen in the industry, requests from customers (internal and external), and feedback from tooling.*
```
chrisdholt, line 13: I think we can drop "internal". As far as the project is concerned, all customers are "external" with exception of tooling. Perhaps we can just clarify a bit? Thinking more on this, it seems like we've added the explicit language here to ensure we don't move on without addressing this requirement. While I definitely get the intent of adding it here explicitly, I think it just feels odd to me to call it out as special given tooling gets a vote at the table and can actually hold the spec to ensure it gets in there. 

I'd like to work on a suggestion but I don't have a concise thing *right now*

specs/template.md
janechu: I'll take a stab at the re-wording.

```suggestion
*Provide Figma files and other design assets. Be sure to account for the various component states, including hover, active, etc. as well as validity, and appearance variants. Focus primarily on the officially supported design system as well as known community scenarios as appropriate. Consider other popular design systems during this process and, where possible, ensure that common design features that may not be part of the officially supported design system can be accommodated. Work closely with engineering to co-develop the visual design along side the API and anatomy.*
```

specs/template.md
awentzel: I agree with excluding "internal and external", since both cover the complete intent already, it's just extra words not extra meaning imho.

specs/template.md
awentzel, line 65: Appear to be using this sentence twice.  Can it be consolidated into just a single usage and placed beneath the Anatomy and Appearance parent markdown title as a note?

specs/template.md
janechu:
```suggestion
*Primary use cases for this component. These should be based on typical patterns seen in the industry, requests from customers and feedback from tooling.*
```

specs/template.md
```diff
@@ -105,7 +111,7 @@
 
 ### Test Plan
 
-*What is the plan for testing the component, if different from the normal path?*
+*What is the plan for testing the component, if different from the normal path? Note that the normal plan includes unit tests for basic state/behavior as well as e2e tests to validate the specific user stories described above.*
```
awentzel, line 114: I tend to not use abbreviations since I suspect most people understand e2e, however, some may not so I'd recommend spelling it out.

awentzel (approved):

specs/template.md
janechu: I would agree with this, I think it could be at the very top, before the spec layout begins. It would also be informative at first glance, if we leave it in the spec, to inform on the status of the spec.

specs/template.md
chrisdholt: I'm open to the proposed rewording from @janechu, but I do want to push on the fact that just as @janechu is adding feedback to the spec template, she can hold or add *requirements* to any given spec. I get the explicit nod, but it feels odd to me and makes tooling feel disconnected from the overall process. If tooling is required to sign-off on the spec, is this wording necessary?

janechu: Should we have some wording like "determine if additional requirements and feedback are necessary from tooling"? Because tooling is only impacted with certain components and we don't need to be involved in the process for all of them.

chrisdholt: @janechu @EisenbergEffect would the language in the first "checkpoint" address this? I see a nod to tooling there and I think if this isn't fully captured there we could update the language for the checkpoint to be clear. The intent of that checkpoint is to ensure that we've captured the use cases. Ensuring the tooling team is invited to that ensures they have a voice and check to validate their scenarios are included. If all scenarios are not, an action item would be to go update the spec to include all relevant needs from the tooling team *can be met*. Thoughts?

specs/template.md
EisenbergEffect: I put it in two sections in the event they are not both read by both people. So, I want to remind the engineering to work with the designer and the designer to work with the engineer, basically.

specs/template.md
EisenbergEffect: Makes sense. I'll shuffle things a bit.

EisenbergEffect:
Ok, I believe I've addressed all feedback. Ready for a second pass when folks are available. Thanks for helping me out on this!

chrisdholt (approved):
Thanks @EisenbergEffect - this is a great update to both the spec and process!

PR #4675, opened 2021-05-05
fix: selected option is not visible when dropdown is opened in fast-select

### LINKED ISSUE #4581: fix: selected option is not visible when dropdown is opened in fast-select
# 🐛 Bug Report

## 💻 Repro or Code Sample

1. Create fast-select
2. Pass in a valid value for `selectedIndex`
3. Open dropdown to see all options
4. Dropdown does not open to selected value

## 🤔 Expected Behavior

When opening the dropdown, the selected value should be focused and in view.

## 😯 Current Behavior

The component seems to be trying to focus and scroll before the dropdown is fully rendered, so when the user opens the dropdown they just see the top of the list.

image

radium-v:
# Pull Request

## 📖 Description

Ensure that listbox-based components properly scroll the currently selected option into view.

### 🎫 Issues

fixes #4581.

## 👩‍💻 Reviewer Notes

The `DOM.queueUpdate` and `DOM.nextUpdate` didn't seem to work here. I suspect it has something to do with how our queueing process works.

## 📑 Test Plan

The playwright test uses `closeTo` since it's possible for different browsers to render the element differently. I used a delta of 16 so things like box model and font size aren't likely to cause the test to fail.

### General

### Component-specific


change/@microsoft-fast-components-f057933c-e206-4ff8-ac2a-a7fcde8beb70.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: selected option is not visible when dropdown is opened in fast-select",
```
chrisdholt, line 3: nit - not necessary to change, but we technically don't need to prefix these descriptions :)
```suggestion
  "comment": "selected option is not visible when dropdown is opened in fast-select",
```

change/@microsoft-fast-foundation-4e5a154f-162b-48d4-84fb-df1ddc03e213.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: selected option is not visible when dropdown is opened in fast-select",
```
chrisdholt, line 3: Same note as above, leaving the suggestion only
```suggestion
  "comment": "selected option is not visible when dropdown is opened in fast-select",
```

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -160,18 +160,38 @@ export class Listbox extends FASTElement {
...
+            requestAnimationFrame(() => {
+                this.firstSelectedOption.scrollIntoView({ block: "nearest" });
+            });
         }
     }
 
+    /**
+     * @internal
+     */
+    private shouldSkipFocus: boolean = false;
```
chrisdholt, line 172: Could you add a quick comment to the tsdoc on when this might be the case?

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -160,18 +160,38 @@ export class Listbox extends FASTElement {
     protected focusAndScrollOptionIntoView(): void {
         if (this.contains(document.activeElement) && this.firstSelectedOption) {
             this.firstSelectedOption.focus();
-            this.firstSelectedOption.scrollIntoView({ block: "nearest" });
+            requestAnimationFrame(() => {
+                this.firstSelectedOption.scrollIntoView({ block: "nearest" });
```
chrisdholt, line 164: Options don't appear to be implemented in Safari - do we know if this errors or gracefully fails in that scenario? My preference is definitely to keep this here, just curious if the absence of it in Safari causes any issues.

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -160,18 +160,38 @@ export class Listbox extends FASTElement {
...
+    }
+
+    /**
+     * Prevents `focusin` events from firing before `click` events when the
+     * element is unfocused.
+     *
+     * @internal
+     */
+    public mousedownHandler(e: MouseEvent): boolean | void {
+        this.shouldSkipFocus = !this.contains(document.activeElement);
```
chrisdholt, line 193: Would the reference to document.activeElement possibly be incorrect if this its in the shadow root?

chrisdholt (approved):

packages/web-components/fast-foundation/src/listbox/listbox.ts
scomea, line 164: I've occasionally seen some unexpected scrolls of parent scrolling containers with scrollIntoView and switched to a more targeted `  this.menuElement.scrollTo(0, focusedOption.offsetTop);` in my picker prototype.  Not sure if that's an issue here.

packages/web-components/fast-components/src/select/select.pw.spec.ts
```diff
@@ -250,4 +250,48 @@ describe("FASTSelect", function () {
...
+
+    describe("when opened", function () {
+        it("should scroll the selected option into view", async function () {
+            const element = (await this.page.waitForSelector(
+                "fast-select"
+            )) as ElementHandle<FASTSelect>;
+
+            await element.evaluate(element => {
+                element.innerHTML = "";
+                for (let i = 0; i < 50; i++) {
```
nicholasrice, line 262: Thank you for adding tests!

nicholasrice (approved):

PR #4705, opened 2021-05-12
chore: fixing incorrect usage of setTimeout in tests

nicholasrice:
# Pull Request

## 📖 Description
HorizontalScroll tests had several tests  using the `await` expression with `Window.setTimeout`. Amazingly it didn't error, but it also doesn't do what the tests were expecting and causing [intermittent build failures](https://github.com/microsoft/fast/pull/4699/checks?check_run_id=2567537408). 

This PR changes those `await setTimeout()` calls to `await timeout()`, where `timeout` is a new test utility that resolves after the provided timeout.

It also disables a few tests that I *believe* are inaccurate. Working with @robarbms to determine if that is correct.

### General

### Component-specific


chrisdholt (approved):
Thanks @nicholasrice!

scomea (approved):

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/disclosure/disclosure.spec.ts
```diff
@@ -17,21 +18,15 @@ async function createDisclosure() {
...
+
 
 describe("Disclosure", () => {
     describe("User interaction", () => {
         it("should toggle the content using `toggle()`", async () => {
             const { element, connect, disconnect } = await createDisclosure();
             await connect();
             element.toggle();
-            await macrotask();
+            await timeout();
```
SethDonohue, line 29: So does having a timeout of *0* effectively just create an empty task/promise to simulate some work being done?

nicholasrice: A timeout of 0 will queue a JavaScript "task" (sometimes called a macro task) with no delay. This pattern provides all the queued micro tasks (like template changes) an opportunity to execute prior to moving onto the next statement. https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide#tasks_vs_microtasks is a good read to understand what is going on.

PR #4709, opened 2021-05-12
feat: styling and text alignment for menu

eljefe223:
# Pull Request

## 📖 Description
This change fixes styling inconsistencies and adds logic to apply classes based on the presence of various children a menu item supports to ensure the content text is always aligned.

###Text Only
image

###Icon or Checkbox or Radio
image
image

###Icon and Checkbox or Radio
image

###Sub Menu Open
image

### 🎫 Issues

Aligns `fast-menu` and `fast-menu-item` with desired design

### General

### Component-specific


packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
```diff
@@ -50,6 +51,33 @@ export const menuItemStyles = (context, definition) =>
         border: calc(${focusOutlineWidth} * 1px) solid transparent;
     }
 
+    :host(.indent-0) {
```
chrisdholt, line 54: just tossing it out - does `align-0` make more sense here than indent, or do we feel indent is sufficiently descriptive of what's happening? Would it help to document this behavior as part of the Foundation Readme so that folks understand the intent when they are extending foundation?

packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
```diff
@@ -108,14 +135,6 @@ export const menuItemStyles = (context, definition) =>
         justify-content: center;
     }
     
-    ::slotted(svg) {
-        ${
-            /* Glyph size and margin-left is temporary - 
-            replace when adaptive typography is figured out */ ""
-        } width: 16px;
-        height: 16px;
-    }
```
chrisdholt, line 117: I know where the removal of this came from, but I don't think we should make that change *just* here. If we're planning on making updates to accommodate both 20x20 and 16x16 SVG's we should do that in a separate branch as an entire feature. Otherwise, we are piecemeal removing things without fully validating the work. Can we add this back in and open an issue to track the update for 20x20 icons?

packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
```diff
@@ -125,15 +144,19 @@ export const menuItemStyles = (context, definition) =>
         fill: ${neutralForegroundRest};
     }
 
-    :host([aria-haspopup="menu"]),
-    :host([role="menuitemcheckbox"]),
-    :host([role="menuitemradio"]) {
+    :host(:not(.indent-2)[aria-haspopup="menu"]),
+    :host(:not(.indent-2)[role="menuitemcheckbox"]),
+    :host(:not(.indent-2)[role="menuitemradio"]) {
```
chrisdholt, line 149: `:not()` can get fairly expensive - are there potential alternatives here which might be more performant?

chrisdholt (changes requested):

chrisdholt (approved):

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
```diff
@@ -16,6 +16,12 @@ import { MenuItemRole } from "./menu-item.options";
 
 export { MenuItemRole };
 
+/**
+ * Types of button appearance.
```
nicholasrice, line 20: This description doesn't seem accurate to me. Also, does this really need to be public?

nicholasrice (approved):
Looks good, just a question about whether we need to document`MenuItemColumn`

packages/web-components/fast-foundation/src/menu-item/menu-item.ts
eljefe223: I'll update the comment, but men does import it

packages/web-components/fast-components/src/menu-item/menu-item.styles.ts
eljefe223: Was able to remove most of the nots.

chrisdholt (approved):

PR #4711, opened 2021-05-12
add CSS partial tagged template literal

### LINKED ISSUE #4633: feat: add partial CSS tagged-template literal to fast-element
# 🙋 Feature Request
`@microsoft/fast-element` exposes the `css` tagged-template literal to create CSS documents that support's `ElementStyles`, `CSSDirective` and `string` interpolation. In implementation, each `ElementStyles` instances get's converted to a `CSSStyleSheet` instance, and this works for most purposes.

There are [some cases though](https://github.com/microsoft/fast/pull/4631/files#diff-7cf742765fdd7ef14a385bed41069673a57260992fb0adcd4587cb044e52b316), where a CSS block being interpolated relies on `CSSDirective` interpolation but *does not* result in a full, valid `CSSStyleSheet`.

I want to propose adding a `cssPartial` tagged-template literal that can be interpolated with strings and `CSSDirective` instances, and that returns a `CSSDirective`. This will allow using `CSSDirective`s in re-usable CSS blocks that may not be valid CSS in-and-of themselves, such as property values and collections of rules with no selector:

**Example usage**:
```ts
// 'baseHeightMultiplier', 'density', and 'designUnit' are all 'CSSDirective` instances
const heightNumber: CSSDirective = cssPartial`(${baseHeightMultiplier} + ${density}) * ${designUnit}`;

const myStyles = css`
    host: {
        height: ${heightNumber};
    }
`;
```

nicholasrice:
# Pull Request

## 📖 Description
This PR adds a new CSS tagged template literal called `cssPartial`. `cssPartial` allows abstracting CSS blocks that are not valid CSS in-and-of themselves into a structure usable by `ElementStyles`. `cssPartial` allows interpolation of `CSSDirective` instances as well as strings, and itself returns a `CSSDirective` that can be interpolated into `ElementStyles`.

closes #4633

### General

### Component-specific


chrisdholt (approved):

packages/web-components/fast-element/docs/guide/leveraging-css.md
```diff
@@ -128,6 +128,71 @@ Rather than simply concatenating CSS strings, the `css` helper understands that
 You can also pass a CSS `string` or a [CSSStyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet) instance directly to the element definition, or even a mixed array of `string`, `CSSStyleSheet`, or `ElementStyles`.
 :::
 
+### Partial CSS
+There are times when you may want to create reusable blocks of *partial* CSS, where the abstraction is not valid CSS in and of itself, such a groups of CSS properties or a complex value. To do that, you can use the `cssPartial` tagged template literal:
```
EisenbergEffect, line 132:
```suggestion
There are times when you may want to create reusable blocks of *partial* CSS, where the abstraction is not valid CSS in and of itself, such as groups of CSS properties or a complex value. To do that, you can use the `cssPartial` tagged template literal: 
```

EisenbergEffect (approved):

packages/web-components/fast-element/docs/guide/leveraging-css.md
chrisdholt: nice - my mind walked right past this!

EisenbergEffect:
@nicholasrice One quick thought/question...is there any way to factor out and share code between the `css` and `cssPartial` helpers? If we add bindings in CSS, I'm thinking we'd need to add it in both places as well, so I'm wondering if there's some way to share the basic parsing/collecting code.

nicholasrice:
We can probably share some, though `css` handles a few extra cases which makes re-use a bit tricky.

nicholasrice:
Hmm I just realized `CSSDirective` returns `ComposableStyles` and not just string, so this won't work as-is. Converting to draft

chrisdholt (approved):

nicholasrice:
@EisenbergEffect Okay I updated this to support all valid return types from `CSSDirective` and also to re-use style collection logic from `css`. 

The main difference between `css` and `cssPartial` is that `css` creates a new CSSStyleSheet for every string in the reduced styles, whereas `cssPartial` will zip-up all strings into a single string and return that from the resulting `CSSDirective`. This means it supports some rather strange syntaxes such as:

**This sets a host's color to blue:**
```ts
class MyDirective extends CSSDirective {
    createCSS() {
        return css`:host { color: blue; }`
    }
}
const partial = cssPartial`color: ${new MyDirective} green`;

css`
    :host {
        ${partial}
    }
`
```

A few questions:
1. Are there general concerns with enabling the above scenario?
2. I currently have `cssPartial` restricted to embedding `string | CSSDirective` (no `ElementStyles or CSSStyleSheet`), but this can obviously be circumvented by returning an `ElementStyles` or `CSSStyleSheet` from a `CSSDirective`. Should I just open up the API?
3. If we changed the behavior of `css` to work like `cssPartial` I think we could technically support both scenarios with one template function, though I think we would loose some of the `CSSStyleSheet` re-use that `ElementStyles` is optimizing for internally. We probably *don't* want to do this, but opening the conversation just in case having one function that does both is preferred.

EisenbergEffect:
Can you expand a bit more on your third point above? I think it would be nice to have a single function, but not sure I fully understand what we'd give up.

nicholasrice:
https://github.com/microsoft/fast/blob/master/packages/web-components/fast-element/src/styles/element-styles.ts#L150 is where ElementStyles converts all style *strings* into CSSStyleSheet instances, caching the instances. That logic is why we can't provide incomplete or invalid CSS to `css` (such as ``` css`color: red` ```). When we try to populate a `CSSStyleSheet` with `"color: red"`, the browser rightfully doesn't know how to convert that to a CSS rule because there is no selector, so it just does nothing.

 If, instead, `ElementStyles` zipped all of the strings into *one* string (or at least all immediate siblings, for instance `[string, string, CSSStyleSheet, string, CSSStyleSheet] => [string, CSSStyleSheet, string, CSSStyleSheet]`), then including a incomplete rule like the above into another `ElementStyles` would just cause the strings to get zipped together prior to `CSSStyleSheet` hydration.

I *think* doing so would reduce the amount of CSSStyleSheet re-use we could do though, because we can no-longer always leverage `CSSStyleSheets` for the individual style strings in a `CSSStyleSheet`.

EisenbergEffect:
Ah, yes. That makes sense. It seems like it could be a perf cost since the browser won't be sharing the sheets or be able to dedupe them internally. So, probably we don't want to do that.

Hmm. I'm not coming up with anything better than the `cssPartial` template function at the moment. If ElementStyles or a CSSStyleSheet are returned, will that break things?

nicholasrice:
Returned from where? `cssPartial`?

EisenbergEffect:
Yes. You mentioned expanding the types to allow that I think. If those were allowed, would it break things?

nicholasrice:
No, it would work the same. It would just make it a little easier to write strange CSS like the above.

EisenbergEffect:
Ok. I think it's ok then. Let's just not put the strange stuff in the docs 😉

nicholasrice:
heh yea okay. I'll update the function.

packages/web-components/fast-element/docs/guide/leveraging-css.md
```diff
@@ -128,6 +128,71 @@ Rather than simply concatenating CSS strings, the `css` helper understands that
...
 :::
 
+### Partial CSS
+There are times when you may want to create reusable blocks of *partial* CSS, where the abstraction is not valid CSS in and of itself, such as groups of CSS properties or a complex value. To do that, you can use the `cssPartial` tagged template literal: 
+
+```ts
+import { css, cssPartial } from "@microsoft/fast-element";
+
+const partial = cssPartial`color: red;`;
+const styles = css`:host{ ${cssPartial} }`;
```
janechu, line 138:
```suggestion
const styles = css`:host{ ${partial} }`;
```

packages/web-components/fast-element/docs/guide/leveraging-css.md
```diff
@@ -128,6 +128,71 @@ Rather than simply concatenating CSS strings, the `css` helper understands that
...
+    }
+  }
+}
+```
+
+### Usage in ElementStyles
+The `CSSDirective` can then be used in an `ElementStyles`, where the CSS string from `createCSS()` will be interpolated into the stylesheet, and the behavior returned from `createBehavior()` will get bound to the element using the stylesheet:
+
+```ts
+const styles = css`:host {${new RandomWidth()}}`;
```
janechu, line 193: Is the intended result that `CSSDirective` class will execute the `crreateCSS()` in the constructor?

janechu (approved):
Couple clarifying questy

packages/web-components/fast-element/docs/guide/leveraging-css.md
nicholasrice: `createCSS()` gets called by `ElementStyles`, so that happens implicitly

PR #4722, opened 2021-05-17
update to exclude tooling from beachball publishing

awentzel:
Publishing with Beachball is failing because tooling is using `latest` versions.  When beachball attempts to publish, the latest version for tools changes to an actual version, then is unable to bump because that version already exists.

By excluding tools, for now, we can avoid this issue and continue to publish.  @janechu can provide additional details as to the temporary nature of this exclusion.

For exclusion details refer to https://microsoft.github.io/beachball/concepts/groups.html
### 🎫 Issues
* Beachball fails thinking it's publishing to an already existing version

## 📑 Test Plan

Will manually deploy a publish to determine if this in fact resolves our issue.

### General

### Component-specific

### DIFF
--- a/build/releasing/index.ts
+++ b/build/releasing/index.ts
@@ -6,8 +6,12 @@ export const config: BeachballConfig = {
     groups: [
         {
             name: "Microsoft FAST",
-            include: ["packages/utilities/*", "packages/web-components/*"],
-            exclude: ["packages/tooling/*"],
+            include: [
+                "packages/tooling/*",
+                "packages/utilities/*",
+                "packages/web-components/*",
+            ],
+            exclude: ["packages/tooling/fast-tooling"],
             disallowedChangeTypes: ["major"],
         },
     ],
--- a/build/releasing/index.ts
+++ b/build/releasing/index.ts
@@ -6,11 +6,8 @@ export const config: BeachballConfig = {
     groups: [
         {
             name: "Microsoft FAST",
-            include: [
-                "packages/tooling/*",
-                "packages/utilities/*",
-                "packages/web-components/*",
-            ],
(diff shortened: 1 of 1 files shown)


build/releasing/index.ts
```diff
@@ -6,11 +6,8 @@ export const config: BeachballConfig = {
     groups: [
         {
             name: "Microsoft FAST",
-            include: [
-                "packages/tooling/*",
-                "packages/utilities/*",
-                "packages/web-components/*",
-            ],
+            include: ["packages/utilities/*", "packages/web-components/*"],
+            exclude: ["packages/tooling/*"],
```
chrisdholt, line 10: Do we want to exclude all tooling from versioning together? I think we just want to exclude the version which is explicitly pinned to `latest`.

```suggestion
            include: ["packages/tooling/*", "packages/utilities/*", "packages/web-components/*",],
            exclude: ["packages/tooling/fast-tooling"],
```

chrisdholt: @janechu can advise, but I think we just want this to exclude that single package

build/releasing/index.ts
```diff
@@ -6,11 +6,9 @@ export const config: BeachballConfig = {
     groups: [
         {
             name: "Microsoft FAST",
-            include: [
-                "packages/tooling/*",
-                "packages/utilities/*",
-                "packages/web-components/*",
-            ],
+            include: ["packages/utilities/*", "packages/web-components/*"],
```
nicholasrice, line 9: I think you want to delete the first `include` field

awentzel: I guess the web merge doesn't work so well.  Resolved.

nicholasrice (approved):

chrisdholt (approved):

PR #4732, opened 2021-05-19
add API support for setting default slotted content as part of component composition

### LINKED ISSUE #4724: feat: Add ability to render elements in number field's step controls
# 🙋 Feature Request

In order to match the design language of the rest of our component toolkit built on top of FAST, we need to be able to render our own arrow icons within the number field's step controls in place of the default arrow icons.

Given that we only need to customize the arrow icons themselves, I believe we only need the ability to slot content inside the `.step-up` and `.step-down` divs that house the default arrow icons and handle the click event.

Requested markup from `.number-field-template.ts`
```
 <div class="controls" part="controls">
  <div class="step-up" part="step-up" @click="${x => x.stepUp()}">
    <slot></slot>
  </div>
  <div class="step-down" part="step-down" @click="${x => x.stepDown()}">
    <slot></slot>
  </div>
</div>
```

## 💻 Examples

Default icons:
CleanShot 2021-05-17 at 16 51 31@2x

Example of Figma component w/ custom styling + product-specific icons
CleanShot 2021-05-17 at 17 05 17@2x

cc @hawkticehurst

chrisdholt:
# Pull Request

## 📖 Description
This PR adds support for default slotted content for certain components (primarily those with visual state or behavior indicators). During this I took the opportunity to close #4724.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/accordion-item/index.ts
+++ b/packages/web-components/fast-components/src/accordion-item/index.ts
@@ -1,4 +1,3 @@
-import { html } from "@microsoft/fast-element";
 import {
     AccordionItem,
     AccordionItemOptions,
@@ -19,7 +18,7 @@ export const fastAccordionItem = AccordionItem.compose<AccordionItemOptions>({
     baseName: "accordion-item",
     template,
     styles,
-    collapsedIcon: html`
+    collapsedIcon: `
         <svg
             width="20"
             height="20"
@@ -33,7 +32,7 @@ export const fastAccordionItem = AccordionItem.compose<AccordionItemOptions>({
             />
         </svg>
     `,
-    expandedIcon: html`
+    expandedIcon: `
         <svg
             width="20"
             height="20"
--- a/packages/web-components/fast-components/src/accordion-item/index.ts
+++ b/packages/web-components/fast-components/src/accordion-item/index.ts
@@ -1,3 +1,4 @@
+import { html } from "@microsoft/fast-element";
 import {
     AccordionItem,
     accordionItemTemplate as template,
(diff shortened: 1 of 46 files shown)


packages/web-components/fast-components/src/accordion-item/index.ts
```diff
@@ -13,10 +15,38 @@ import { accordionItemStyles as styles } from "./accordion-item.styles";
  * @remarks
  * HTML Element: \<fast-accordion-item\>
  */
-export const fastAccordionItem = AccordionItem.compose({
+export const fastAccordionItem = AccordionItem.compose<AccordionItemOptions>({
     baseName: "accordion-item",
     template,
     styles,
+    collapsedIcon: html`
```
EisenbergEffect, line 22: The default option can be a simple string here, which would be more performant in the end. That doesn't prevent the end user from overriding it with a custom template. Since there doesn't appear to be an need for template features, I would just remove the `html` tag from the front of all of these.

packages/web-components/fast-components/src/breadcrumb-item/index.ts
```diff
@@ -13,10 +14,11 @@ import { breadcrumbItemStyles as styles } from "./breadcrumb-item.styles";
  * @remarks
  * HTML Element: \<fast-breadcrumb-item\>
  */
-export const fastBreadcrumbItem = BreadcrumbItem.compose({
+export const fastBreadcrumbItem = BreadcrumbItem.compose<BreadcrumbItemOptions>({
     baseName: "breadcrumb-item",
     template,
     styles,
+    separator: "/",
```
EisenbergEffect, line 21: Nice!

EisenbergEffect (changes requested):
This is going to be very cool. I hadn't realized how many of our components can be improved by this. One recommendation I have it to change the defaults to not use `html`. Since none of them actually use templating features, the defaults can just be simple strings, which will improve performance. That doesn't prevent the end user from overriding that and providing a template that does use bindings, etc. if they want.

packages/web-components/fast-components/src/accordion-item/index.ts
EisenbergEffect:
```suggestion
    collapsedIcon:`
```

chrisdholt: Thanks @EisenbergEffect - updated to just leverage strings here rather than templates. I think this is a good start for the default scenarios. Over time we can add or update additional components as well.

EisenbergEffect (approved):

PR #4744, opened 2021-05-20
fix tooling libraries to use the new design system registry and updated named imports

janechu:
# Pull Request

## 📖 Description

This change updates to use the new Design System registry as well as named imports from `@microsoft/fast-components` as well as updates the dependencies to not be on `latest` but the current version. This change will be mirrored in another PR to master so that vNext can be unblocked and master can be subsequently unblocked.

## 👩‍💻 Reviewer Notes

This change appears to run fine on the manual apps and web components that are registered to the design system are showing up. Let me know if there is a better method, we are making the assumption that either the user is not going to be on a separate version of the Design System and will be using the same class with this change.

### General

### Component-specific

### DIFF
--- a/packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
+++ b/packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
@@ -7,11 +7,11 @@ import {
 } from "@microsoft/fast-colors";
 import { attr, DOM, observable } from "@microsoft/fast-element";
 import { isNullOrWhiteSpace } from "@microsoft/fast-web-utilities";
-import { FASTTextField } from "@microsoft/fast-components";
+import { fastTextField } from "@microsoft/fast-components";
+import { DesignSystem } from "@microsoft/fast-foundation";
 import { FormAssociatedColorPicker } from "./color-picker.form-associated";
 
-// Prevent tree shaking
-FASTTextField;
+DesignSystem.getOrCreate().register(fastTextField());
 
 /**
  * Simple class for storing all of the color picker UI observable values.
--- a/packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
+++ b/packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
@@ -7,11 +7,16 @@ import {
 } from "@microsoft/fast-colors";
 import { attr, DOM, observable } from "@microsoft/fast-element";
 import { isNullOrWhiteSpace } from "@microsoft/fast-web-utilities";
-import { fastTextField } from "@microsoft/fast-components";
-import { DesignSystem } from "@microsoft/fast-foundation";
 import { FormAssociatedColorPicker } from "./color-picker.form-associated";
 
-DesignSystem.getOrCreate().register(fastTextField());
+/**
+ * This is currently experimental, any use of the color picker must include the following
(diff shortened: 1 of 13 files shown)


packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
```diff
@@ -7,11 +7,11 @@ import {
...
 import { attr, DOM, observable } from "@microsoft/fast-element";
 import { isNullOrWhiteSpace } from "@microsoft/fast-web-utilities";
-import { FASTTextField } from "@microsoft/fast-components";
+import { fastTextField } from "@microsoft/fast-components";
+import { DesignSystem } from "@microsoft/fast-foundation";
 import { FormAssociatedColorPicker } from "./color-picker.form-associated";
 
-// Prevent tree shaking
-FASTTextField;
+DesignSystem.getOrCreate().register(fastTextField());
```
EisenbergEffect, line 14: My recommendation would be to remove this. Just make one call to getOrCreate and pass all the components to it in a central location.

EisenbergEffect (changes requested):

packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
janechu: I can insist that if anyone uses these `experimental` components that they have to include all dependencies themselves, but as far as I'm aware it's possible that before release we may not have a way of having implicit registrations occurring. This is not ergonomic for someone who's only taking a dependency on tooling and would expect to only register, for example, the color picker.

Imagining for a moment that the `<Form />` relied on the FAST components and aside from importing and using it, the app author also had to register all of its web component dependencies separately. I would in this case have preferred to wait on adoption until we had an ergonomic solution. Fortunately only `experimental` components are using the FAST components at this time.

packages/tooling/fast-tooling/src/web-components/color-picker/color-picker.ts
```diff
@@ -7,11 +7,16 @@ import {
...
 
-// Prevent tree shaking
-FASTTextField;
+/**
+ * This is currently experimental, any use of the color picker must include the following
+ * imports and register with the DesignSystem
+ *
+ * import { fastTextField } from "@microsoft/fast-components";
+ * import { DesignSystem } from "@microsoft/fast-foundation";
+ * DesignSystem.getOrCreate().register(fastTextField());
```
nicholasrice, line 18: I think you *also* want to change this component to follow the model set in fast-foundation, where ColorPicker extends FoundationElement  the package export is actually the product of `ColorPicker.compose()`.

This gives ColorPicker oppertunity to get the *tag* for any dependent components from context:

```
const ColorPickerTemplate = (context: ElementDefinitionContext) => {
   return html` .... <${context.tagFor(TextField}`></${context.tagFor(TextField)}>`
}

janechu: This is passing build, I need to also convert the other web components such as the html renderer to use the updated format. I think it might be more timely to merge this as-is with a follow up PR.

nicholasrice: cool sounds good

EisenbergEffect (approved):

PR #4748, opened 2021-05-20
Feat: add calendar component

robarbms:
# Pull Request

## 📖 Description
This adds a basic calendar component. This will be used for a date picker component.

## 👩‍💻 Reviewer Notes

Localization is particularly difficult because of all the market, calendar and numbering permutations. I have localization mostly baked in. I wasn't able to find an exact list of what calendar and numbering types each market uses. I haven't tested localization defaults outside of the en-US market. 

## 📑 Test Plan

Need to figure out how to test other markets, calendar types and numbering systems.

### General

### Component-specific

## ⏭ Next Steps

Find a solution for calendars that don't align with the Gregorian calendar system.


EisenbergEffect:
@robarbms Would you mind dropping a few screenshots in? I probably won't get a chance to pull down the code and run it, but I'd love to see how this is looking.

robarbms:
> @robarbms Would you mind dropping a few screenshots in? I probably won't get a chance to pull down the code and run it, but I'd love to see how this is looking.

@EisenbergEffect 
**Default calendar** - Note: I added a sans-serif font to all calendars for the screenshots because times new roman is too hard to look at.
calendar-SS

**Long weekday labels** - It supports the browser settings of long, short and narrow. (Monday, Mon, M). Month labeling also supports long, short and narrow display types. (December, Dec, D)
calendar-expanded-SS

**Compact view** - An example of a smaller calendar width and using narrow weekdays.
calendar-compact-SS

**Border styling** - An example showing a lot of the custom styling. It's possible to do a little more but going too crazy made it hard to look at so I kept it to a limited color pallet. 
calendar-custom-SS

**With events** - The template adds dynamic slots with the dates(MM-DD-YYYY) as the slot names so it's possible to add custom events for each date.
calendar-events-SS

**Thai Buddhist calendar** - An example of a Thai Buddhist calendar. Shows off localization. European is pretty easy. Thai is a little trickier as they are on the year 2464. But the Buddhist calendar has been aligned with the Gregorian calendar.
calendar-thai-SS

**Arabic calendar** - An example of an Arabic calendar and RTL. This might need some implementation changes. I'm just using CSS direction on the component. I have code that should in theory identify RTL calendars but I don't know enough about all the different calendar types to confidently identify them.
calendar-arabic-SS

EisenbergEffect:
AWESOME! Thanks @robarbms 🎊

Ibrahimmaga:
they all look great

change/@microsoft-fast-components-c15c784b-9a60-441a-88dc-f032d990e5bb.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
nicholasrice, line 2: I don't think this is a major change; that would result in a breaking version change on publish. Can you downgrade this to minor?

change/@microsoft-fast-foundation-8b1801b3-ce1d-4623-af8d-b8b388be0ddd.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "major",
```
nicholasrice, line 2: same note on major as above here.

packages/web-components/fast-components/src/calendar/calendar.styles.ts
```diff
@@ -0,0 +1,65 @@
...
+        font-size: 1.5em;
+        padding: 5px 0;
+        text-align: center;
+    }
+
+    .days {
+        border-left: var(--cell-border);
+        border-top: var(--cell-border);
+        display: grid;
+        grid-template-columns: 1fr 1fr 1fr 1fr 1fr 1fr 1fr;
```
nicholasrice, line 31: Just FYI in case you weren't aware, you can write this as `repeat(7, 1fr)` to condense and make this a little easier to understand at a glance.

packages/web-components/fast-foundation/src/calendar/calendar.ts
```diff
@@ -0,0 +1,238 @@
...
+/**
+ * Calendar component
+ * @public
+ */
+export class Calendar extends FASTElement {
+    /**
+     * Today's date
+     * @internal
+     */
+    private today: Date = new Date();
```
nicholasrice, line 39: If the calendar loads at 11:59PM and the user interacts at 12:00am, does this need to update to the next day?

robarbms: It should only be used during the rendering cycle. The only case I can see is if the calendar is updated, triggering a re-rendering. It would be working at a static Date during class construction. I've replaced it to use new Date() at their calls so that it will be updated on re-renders.

scomea:
Should we call this something more specific than calendar?  It feels like it's really a "calendar month display" (?) and I could imagine someone using this to render months in a "full year" calendar component.

robarbms:
> Should we call this something more specific than calendar? It feels like it's really a "calendar month display" (?) and I could imagine someone using this to render months in a "full year" calendar component.

I considered having a full year displayed as an option, implemented a couple different ways. The first was if the developer set the year but not the month attribute, it could show the whole year. I liked having it default to the current month as this would be the most common use case and it would be nice if the developer could get the latest month calendar without the use of a client or server side script. The second way was to create a view attribute with it defaulted to the month view. This has the benefit of being added later without effecting current implementations. Here are some early explorations of a year calendar [Codepen example](https://codepen.io/kungfukarl/pen/NWdaXZQ)

This could also be done pretty easily through a script:

`
    ((year) => {
        const container = document.getElementById("year");
        const addMonth = (_, index) => {
            const monthCalendar = document.createElement("fast-calendar");
            monthCalendar.className = "compact";
            monthCalendar.month = index + 1;
            if(year) monthCalendar.year = year;
            container.appendChild(monthCalendar);
        }
        new Array(12).fill(null).forEach(addMonth);
    })();
`
I included this as an example on the Storybook page as well.

scomea:
What about accessibility?  I feel like I should be able to keyboard navigate and have narrator spout out days and events, etc...

robarbms:
> What about accessibility? I feel like I should be able to keyboard navigate and have narrator spout out days and events, etc...

There are no inherent interactions with the calendar component. It's just a a rendering of the calendar. Interactions in the Storybook are just examples of how to implement getting a click or changing the calendar month. Tabbing to flippers or tabbing to events would be handled by the dev slotting them.

I'm not sure how to approach tabbing when adding the onclick function to the calendar. It would need to add the ability to add focus to calendar days.

scomea:
> > What about accessibility? I feel like I should be able to keyboard navigate and have narrator spout out days and events, etc...
> 
> There are no inherent interactions with the calendar component. It's just a a rendering of the calendar. Interactions in the Storybook are just examples of how to implement getting a click or changing the calendar month. Tabbing to flippers or tabbing to events would be handled by the dev slotting them.

From an accessibility POV being able to navigate the content is important, and I'm not sure devs can do that well on their own.  Making each event a tab stop isn't ideal.  I'm thinking the interaction model for calendar would be very similar to what we have for grid - ie.  A single tab stop, either on the first date or the last focused one, then arrow keys to navigate the dates, for each date narrator would go through a spiel about the date and events,  "enter key" enters the date item and allows selection of events.   Or maybe the month then "enter" to navigate dates?  They even use a calendar as the example for an accessible grid at MDN - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Grid_Role

packages/web-components/fast-components/src/calendar/README.md
```diff
@@ -0,0 +1,4 @@
+# fast-calendar
+An implementation of a calendar as a web-component.
+
+For more information view the [component specification](../../../fast-foundation/src/calendar/calendar.spec.md).
```
nicholasrice, line 4: Does this file exist somewhere that I'm not seeing?

nicholasrice: Oh nm I see it in another PR

robarbms:
> > > What about accessibility? I feel like I should be able to keyboard navigate and have narrator spout out days and events, etc...
> > 
> > 
> > There are no inherent interactions with the calendar component. It's just a a rendering of the calendar. Interactions in the Storybook are just examples of how to implement getting a click or changing the calendar month. Tabbing to flippers or tabbing to events would be handled by the dev slotting them.
> 
> From an accessibility POV being able to navigate the content is important, and I'm not sure devs can do that well on their own. Making each event a tab stop isn't ideal. I'm thinking the interaction model for calendar would be very similar to what we have for grid - ie. A single tab stop, either on the first date or the last focused one, then arrow keys to navigate the dates, for each date narrator would go through a spiel about the date and events, "enter key" enters the date item and allows selection of events. Or maybe the month then "enter" to navigate dates? They even use a calendar as the example for an accessible grid at MDN - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Grid_Role

I've implemented a data-grid system. I worked off of data-grid rather than fast-data-grid so that I didn't take on the styles which conflicted with the calendar.

scomea:
Was going to give it a test drive but having issues building - 
image

robarbms:
> Was going to give it a test drive but having issues building -
> image

Sorry. There were changes to data-grid and design-tokens from vNext that I needed to fix. I'm still polishing things up but I pushed the changes needed to build and run Calendar.

scomea:
Slick.  Only major issue I see is that we need to work on what assistive tech says as you navigate the calendar (ie. the aria things).

packages/web-components/fast-foundation/src/calendar/calendar.template.ts
```diff
@@ -0,0 +1,77 @@
...
+ * A calendar day template
+ * @public
+ */
+export const CalendarDayTemplate: ViewTemplate<CalendarDateInfo> = html`
+    <div
+        part="day"
+        aria-label="${(x, c) =>
+            ("getMonth" in c.parent ? c : c.parentContext).parent.getMonth(
+                x.month
+            )} ${x => x.day}"
```
robarbms, line 37: @scomea added the month and day example: "July 26" as the aria label for each cell.

robarbms: Ran into some issues with aria-labels for abbreviated weekday names. I added labels but seems to read off the abbreviated if it's an actual word. Reads "sat" for Sat but reads "Friday" for Fri. I might need to play around with the abbr attribute.

eljefe223:
Noticing a few minor styling issues.

The text appears to be off center in the selected date:
image

Text is clipped in some of the event titles:
image

 I personally feel that the selected date style seems broken because the border doubles up:
image

Perhaps we have @bheston revisit this?

khamudom:
For high contrast, could you add,
```

**Comment by khamudom** [comment]
To fix some of high contrast issues, you can add,
```
.withBehaviors(
    forcedColorsStylesheetBehavior(
        css`
            :host {
                --selected-day-outline: 1px solid ${SystemColors.Highlight};
            }
            .today .date {
                background: ${SystemColors.Highlight};
                color: ${SystemColors.HighlightText};
                fill: currentcolor;
            }
            .day.selected {
                background: ${SystemColors.Canvas};
                color: ${SystemColors.CanvasText};
                fill: currentcolor;
            }
        `
    )
);
```
This will fix the selected states
before
image
image
after
image
image
This will fix the high contrast within the host element. There are other high contrast issues in other examples, this will have to be addressed along with the styles in the fixture.

File: packages/web-components/fast-foundation/src/calendar/calendar.template.ts [hunk 7]
```diff
@@ -0,0 +1,167 @@
+import { ExecutionContext, html, repeat } from "@microsoft/fast-element";
```
scomea, line 1: "ExecutionContext" is unused

packages/web-components/fast-foundation/src/calendar/calendar.ts
```diff
@@ -0,0 +1,339 @@
...
+/**
+ * Calendar component
+ * @public
+ */
+export class Calendar extends FoundationElement {
+    /**
+     * date formatter utitlity for getting localized strings
+     */
+    @observable
+    public dateFormatter: DateFormatter = new DateFormatter();
```
scomea, line 55: Should this be marked as internal?

packages/web-components/fast-foundation/src/calendar/calendar.ts
```diff
@@ -0,0 +1,339 @@
...
+     * The number of miliseconds in a day
+     * @internal
+     */
+    private oneDayInMs: number = 86400000;
+
+    /**
+     * An evaluation of whether or not the calendar is rendered as RTL
+     * @internal
+     */
+    public isRTL(locale: string = this.locale): boolean {
```
scomea, line 154: Where is this used?

packages/web-components/fast-foundation/src/calendar/calendar.ts
scomea: I don't think it needs to be observable either, an internal public prop should do.

packages/web-components/fast-foundation/src/calendar/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./calendar";
+export * from "./calendar.template";
+export * from "./date-formatter";
```
scomea, line 3: I don't think we need to export `date-formatter` as it looks to be a utility specifically intended for the calendar base class and not a public api?

robarbms: I will need it for the date-picker as well. But I can pull that in through a file path as it will only be needed in fast-foundation.

packages/web-components/fast-foundation/src/calendar/calendar.ts
robarbms: It was just used in testing. Removed this.

packages/web-components/fast-foundation/src/calendar/index.ts
robarbms: When I remove this, there are lots of warnings about missing exports. Part of it is that all of the types that Calendar relies on.

bheston:
Regarding the overall design, since we're embarking on the visual update we should add this one for @jareyno to specify. One way to address this for selection, which may or may not be contiguous, would be to use a fill instead of an outline. Perhaps maintain the square shape to differentiate it from the circle used to mark the current day. That seems common for calendar controls.

jareyno:
@bheston That sounds good. I'll look into it

robarbms:
> To fix some of high contrast issues, you can add,
> 
> ```
> .withBehaviors(
>     forcedColorsStylesheetBehavior(
>         css`
>             :host {
>                 --selected-day-outline: 1px solid ${SystemColors.Highlight};
>             }
>             .today .date {
>                 background: ${SystemColors.Highlight};
>                 color: ${SystemColors.HighlightText};
>                 fill: currentcolor;
>             }
>             .day.selected {
>                 background: ${SystemColors.Canvas};
>                 color: ${SystemColors.CanvasText};
>                 fill: currentcolor;
>             }
>             .week-day {
>                 background-color: ${SystemColors.Canvas};
>             }
>         `
>     )
> );
> ```
> 
> This will fix the selected states and the background on week-day
> before
> image
> image
> image
> 
> after
> image
> image
> image
> 
> This will fix the high contrast within the host element. There are other high contrast issues in other examples, this will have to be addressed with the styles in the fixture.

I've updated for high contrast. Thanks!

robarbms:
> I'm having an issue running this in Safari and Firefox - can you look into what might be happening there and if we need fallbacks somewhere?

This was an issue on date construction differences. Chrome and IE allow for ```new Date('10-7-2021')```. FireFox doesn't and I had to use a more universal constructor ```new Date(2021, 10, 7)```

PR #4768, opened 2021-05-24
chore(docs): inline text editing layer for htmlrender in creator 

williamw2:
# Pull Request

## 📖 Description

Adding a markdown spec proposal for a new inline editing HTML Render Layer component to be integrated into Creator.

### General

### Component-specific


packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/inline-edit.spec.md
```diff
@@ -0,0 +1,33 @@
...
+When the user is done, any event that causes the `textarea` to lose focus to include typing Tab or Enter, or clicking outside of the `textarea`
+will be considered a confirmation. The target text node's contents will be replaced with the contents of the `textarea` and the `textarea` input 
+will be removed. The Data Dictionary will be updated with the new content and a `data` message will be sent via the `MessageSystem`.
+If the Esc key is typed instead then the target text node's content will be restored to its original value and the `textarea` input will be 
+removed. 
+
+## Out of scope
+Inline editing is not intended to be a full function text editing / formating experience (at this time) and will only allow for simple text 
+input. It is also not meant to be an HTML editor and any markup typed into the inline editor will be escaped so it appears as text. Markup
+editing can be accomplished in the Monico editor.
```
janechu, line 33:
```suggestion
editing can be accomplished in the Monaco editor.
```

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/inline-edit.spec.md
```diff
@@ -0,0 +1,33 @@
...
+element. While the user is typing within this area keyboard/mouse interactions will work normally (letters, numbers, special characters, 
+arrow key navigation, mouse click curser placement, highlighting, copy/paste etc.) with the exceptions of Tab, Enter and Esc. Tab and Enter 
+will perform a confirmation action which will remove the `textarea`, replace the text node with the new text and transmit a data message
+announcing the update. The Esc key will cancel the text edit, removing the `textarea` and restoring the text node to the original value.
+
+## Implementation
+When the `HTMLRender` component detects a `dblclick` event it determine if the even occured on a text node and walk up the `composedPath` 
+until it finds the containing element. It will then send a `ActivityType.DoubleClick` alert to all of the child layers. 
+The `HTMLRenderLayerInlineEdit` layer will respond to this alert by finding the position and styles (background color, font color, size, family etc)
+and will create a `textarea` input element with the same styles placed directly over the target location. It will also save a backup of the
```
janechu, line 21: > directly over the target location

Hmmm, is there a concern that adding or removing lines could move the content below up/down the page (or should)? If we don't do this do we think it will be a problem, especially in the case of removing lines. I suppose if the `textarea` does not resize smaller, we won't have an issue.

williamw2: I mention in the Implementation section that the text node will be updated as the user types and the textarea will resize to match. So it will resize as lines are added/removed but I should probably mention that it would need some minimum size. Especially in situations like when an element with no text is double clicked.

EisenbergEffect (approved):

janechu (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/inline-edit.spec.md
williamw2: Fixed

Issue #4739, opened 2021-05-19
feat: FAST should have a standard mechanism for suppressing the automatic tab queue behaviors of some components


scomea:
Some components automatically add themselves or child items to the tab queue by setting tabindex values on elements.  For example radio group and data grid do this.   However there are cases when these components may find themselves nested in other components that may want to suppress this behavior, for example a radio group within a toolbar, or anything nested in a complex data grid cell.  We should have a scalable standardized pattern for this.

This came up regarding toolbar and radio group (https://github.com/microsoft/fast/pull/4535) which has a very specific mechanism to address this.  Ideally we can come up with a saleable approach that does not depend on the child component to know of and be able to identify scenarios where it should not be adding things to the tab queue, but rather shifts that task to the higher level parent component.

janechu:
Unfortunately `@microsoft/fast-components` has been deprecated for some time and `@microsoft/fast-foundation` is being deprecated per #6955.

PR #4775, opened 2021-05-25
feat: nested grid

scomea:
## 📖 Description

Yes, your grids can have grids.

image

This pr adds a nested grid example and some minor fixes/changes to data-grid components to get the keyboarding to work correctly (ie. arrow keys to move around the grid, "Enter" to access a child grid, "esc" to move back to the parent).

The one notable change is adding a "no-tabbing" attribute to data grid so that nested grids don't add themselves to the tab queue which is a proposed fix for this issue: https://github.com/microsoft/fast/issues/4739.  

### 🎫 Issues
https://github.com/microsoft/fast/issues/4739

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/data-grid/data-grid.ts
+++ b/packages/web-components/fast-foundation/src/data-grid/data-grid.ts
@@ -356,6 +356,10 @@ export class DataGrid extends FoundationElement {
         // only observe if nodes are added or removed
         this.observer.observe(this, { childList: true });
 
+        if (this.noTabbing) {
+            this.setAttribute("tabindex", "-1");
+        }
+
         DOM.queueUpdate(this.queueRowIndexUpdate);
     }
 
--- a/packages/web-components/fast-foundation/src/data-grid/data-grid.ts
+++ b/packages/web-components/fast-foundation/src/data-grid/data-grid.ts
@@ -63,7 +63,7 @@ export interface ColumnDefinition {
      * When headerCellInternalFocusQueue is false this function is called when the cell is first focused
      * to immediately move focus to a cell element, for example a cell that is a checkbox could move
      * focus directly to the checkbox.
-     * When headerCellInternalFocusQueue is true this function is called when the user hits Enter or F2
+     * When ueuheaderCellInternalFocusQe is true this function is called when the user hits Enter or F2
      */
     headerCellFocusTargetCallback?: (cell: DataGridCell) => HTMLElement;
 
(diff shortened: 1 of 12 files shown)


awentzel:
Changed the name to remove, "Draft" as it's already attributed as a GH Draft.

awentzel:
Hi @scomea, is this work ongoing?  Is this work still necessary?  As a Draft, I'd recommend either pushing non-draft changes and finishing the review or abandoning until it's ready.

scomea:
I'll update with a real pr shortly.

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -63,7 +63,7 @@ export interface ColumnDefinition {
      * When headerCellInternalFocusQueue is false this function is called when the cell is first focused
      * to immediately move focus to a cell element, for example a cell that is a checkbox could move
      * focus directly to the checkbox.
-     * When headerCellInternalFocusQueue is true this function is called when the user hits Enter or F2
+     * When ueuheaderCellInternalFocusQe is true this function is called when the user hits Enter or F2
```
robarbms, line 66: nit: 
```suggestion
     * When headerCellInternalFocusQueue is true this function is called when the user hits Enter or F2
```

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
```diff
@@ -126,6 +126,32 @@ export class DataGrid extends FoundationElement {
...
+     * @public
+     * @remarks
+     * HTML Attribute: no-tabbing
+     */
+    @attr({ attribute: "no-tabbing", mode: "boolean" })
+    public noTabbing: boolean = false;
+    private noTabbingChanged(): void {
+        if (this.$fastController.isConnected) {
+            if (this.noTabbing) {
+                this.setAttribute("tabIndex", "-1");
```
robarbms, line 142: Should this just not set tabindex? This would allow the developer to still control the tabindex if they wanted to. They might want to elevate a particular cell in the tabbing order? Not sure of use cases, just wondering if we can make it more flexible if the case is needed.

robarbms (approved):

packages/web-components/fast-foundation/src/data-grid/data-grid.ts
scomea: I think we want to keep the data-grid as a focusable entity whether or not it is in the tab queue.

PR #4785, opened 2021-05-26
Remove Design System Provider infrastructure

nicholasrice:
# Pull Request
This PR migrates Design System Provider(DSP) to Design Token infrastructure.

## 📖 Description
The DSP still exists as a Web Component primarily to ease migration, however we've also gotten feedback that the component is useful so folks are free to use it if they wish. This PR removes DSP infrastructure from fast-foundation though, and refactors the DSP in fast-components to be one based on Design Tokens.

It also updates the FAST website where appropriate to leverage new component and Design Token infrastructure.

Opening as a draft at the moment because I discovered a few bugs that need to be addressed outside this PR. Once those fixes go in I'll open un-draft this PR.

### General

### Component-specific


sites/fast-website/src/app/index.ts
```diff
@@ -1,4 +1,13 @@
+import { DesignSystem } from "@microsoft/fast-foundation";
+import * as components from "@microsoft/fast-components/dist/esm/custom-elements";
+import { siteColorSwatch } from "./components";
 import "./css/style.css";
+DesignSystem.getOrCreate().register(
+    ...Object.values(components).map(x => (x as () => void)()),
```
EisenbergEffect, line 6: I wonder if we should have a special export from custom-elements that implements Registry so that we could pass it directly to the register method without the odd syntax. Is it worth doing that for convenience?

EisenbergEffect (approved):
Looks like you got a lot done. What's left to get the site up and running?

sites/fast-website/src/app/index.ts
nicholasrice: Oops forgot to update this, good catch. I was going to only register what was being used in the site.

Good question. It could be, maybe a `customElements` or `components` named export that is the entire component collection?

sites/fast-website/src/app/index.ts
```diff
@@ -1,4 +1,13 @@
...
+    siteColorSwatch({ prefix: "site" })
```
nicholasrice, line 7: @EisenbergEffect I found a bit of an odd use-case we might want to address. The site's color swatch extends FAST's Radio component, which is a Foundation component. Migrating that straight over to vNext infra causes the component to blow up because foundation is looking for a presentation in the Design System, even though the component is actually being defined via `@customElement`. I'm wondering if there is a way to make this case easier, where an author wants to *extend* a component with a `FoundationElement` base-class and define it using `@customElement`

packages/web-components/fast-components/src/button/fixtures/button.html
```diff
@@ -1,80 +1,82 @@
-<h1>Button</h1>
+<fast-design-system-provider>
```
nicholasrice, line 1: note-to-self, revert this

nicholasrice:
@EisenbergEffect I don't think anything, it should be good to go when this gets merged, though it would probably benefit from a second pair of eyes to validate.

sites/fast-website/src/app/index.ts
EisenbergEffect: I think we could fix this pretty easily. In FoundationElement, if there is no. presentation object present, just do nothing. Then the base connect implementation should kick in and look up the template and the styles with the definition object that comes from the decorator. Does that sound right?

EisenbergEffect: Actually, it's more tricky than that. We need to change the way the presentation object is looked up internally, but we could probably make that work. How fast do we need this do you think? I might be able to do that on Friday.

sites/fast-website/src/app/index.ts
EisenbergEffect: What about something called `allComponents`?

EisenbergEffect: We could follow the same pattern in fluent.

sites/fast-website/src/app/index.ts
nicholasrice: Not fast, just a case I encountered here and I'm sure it will happen again. I'll write up an issue to track.

EisenbergEffect: I had another look at the code and I think it's just a couple of small changes. I'll see if I can get my PR-for-the-week in tomorrow to enable this 🤣

nicholasrice: nice, if not I can look into it next week. https://github.com/microsoft/fast/issues/4794 is tracking

nicholasrice:
Closing this as rebasing is a nightmare. Will re-open w/ a different branch.

PR #4786, opened 2021-05-26
Fix Design Token Parent Element Retrieval

nicholasrice:
# Pull Request
Fixes a bug caused by asserting `HTMLElement.parentElement` was not null.

## 📖 Description
An invalid assertion was causing a runtime error in cases where a DesignToken was being set for an element that was a direct child of a ShadowRoot. In this case, the `parentNode` property is not null (it is a document fragement) but `parentElement` is null. 

This PR updates that logic to use the "composedParent", which correctly accounts for the case when the parentNode is a shadowRoot.

### General

### Component-specific


EisenbergEffect (approved):

bheston (approved):
I was running into this same issue, which resolved when I patched with this update.

janechu (approved):

PR #4805, opened 2021-05-29
add a layout web component for css properties

janechu:
# Pull Request

## 📖 Description

This creates a CSS layout web component in `@microsoft/fast-tooling` that can be used by itself or with the experimental custom css control in `@microsoft/fast-tooling-react`. It primarily uses the CSS flexbox to update CSS either by the change event or an `onChange` property assigned to the element.

This PR also contains the integration into the Creator design tool.

Screen Shot 2021-05-28 at 5 49 02 PM

## 👩‍💻 Reviewer Notes

Running the Creator site will show off the control in the right pane, to view it by itself run `yarn start` to use the manual testing app in `@microsoft/fast-tooling` package.

## 📑 Test Plan

New unit tests have been added for this component.

### General

### Component-specific

### DIFF
--- a/packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.template.ts
+++ b/packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.template.ts
@@ -535,243 +535,238 @@ const columnGap = html`
  */
 export const CSSLayoutTemplate = html<CSSLayout>`
     <template>
-        <div class="root">
+        <div class="control-region">
+            <fast-switch @change="${(x, c) => x.handleToggleCSSLayout()}">
+                Enable Flexbox
+            </fast-switch>
+        </div>
+        <div class="flexbox-controls${x => (x.flexEnabled ? ` active` : "")}">
             <div class="control-region">
-                <fast-switch @change="${(x, c) => x.handleToggleCSSLayout()}">
-                    Enable Flexbox
-                </fast-switch>
+                <label for="fast-tooling-css-flex-direction">Direction</label>
+                <div class="control-radio-region">
+                    ${repeat(
+                        x => x.flexDirectionOptions,
+                        html<string, CSSLayout>`
+                            <div
+                                class="${(x, c) => c.parent.flexDirectionName} ${x =>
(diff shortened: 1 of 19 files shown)


EisenbergEffect:
Could you post a screenshot? 😁

EisenbergEffect:
Sweet!

EisenbergEffect (approved):
As I was reviewing, I realized...I can't remember reviewing another web components you've written. Is this your first FAST Web Component @janechu ?

EisenbergEffect:
We should get this in, but there may be some way to reuse some FAST components to improve this a bit in the future. In particular, I'm thinking of replicating the pattern for how the color swatch was built for the FAST site. That's kind of similar to the radio group-like situation with the icon buttons here. We shouldn't block on that though. We can revisit when there's more time.

packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.template.ts
```diff
@@ -0,0 +1,779 @@
...
+    </svg>
+`;
+
+/**
+ * The template for the css-layout component.
+ * @public
+ */
+export const CSSLayoutTemplate = html<CSSLayout>`
+    <template>
+        <div class="root">
```
chrisdholt, line 538: I could be missing an underlying need for this, but from what I can tell, you technically could remove this and the ‘template’ and use ‘host:’ where your root selector is in CSS

chrisdholt (approved):

janechu:
> As I was reviewing, I realized...I can't remember reviewing another web components you've written. Is this your first FAST Web Component @janechu ?

It very well might be, I've reviewed and looked at web components but I also can't remember actually writing one before this.

packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.template.ts
janechu: Nope, this was some leftover bits while I was revving that now no longer serve a purpose, I'll remove.

Issue #4713, opened 2021-05-13
fix: should be able to tab to the flipper control in horizontal-scroll component


robarbms:
# 🐛 Bug Report

## 🤔 Expected Behavior

Should be able to set tabbing on flippers.

## 😯 Current Behavior
Click behavior is on a wrapper surrounding the flipper slots.

## 💁 Possible Solution

Moving the click behavior from the flipper slot wrappers to the flippers themselves will allow developers to use their own flippers , click and tabbing behaviors.

## 🔦 Context
This will future proof the component for custom flippers.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #4807, opened 2021-05-29
Add tabbing to horizontal scroll component flippers

robarbms:
# Pull Request

## 📖 Description

Users should be able to use the keyboard to navigate the horizontal-scroll component.

### 🎫 Issues

[fix: should be able to tab to the flipper control in horizontal-scroll component ](https://github.com/microsoft/fast/issues/4713)

## 👩‍💻 Reviewer Notes

Converted the flipper wrappers to buttons so that they could get the click / enter key events bubble from the flippers. Disabled tab on the wrapper and enabled tabbing on the flippers.

### General

### Component-specific

## ⏭ Next Steps

Add default flippers using the fast-flipper component. I'll create a separate issue.

### DIFF
--- a/packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
+++ b/packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
@@ -10,6 +10,29 @@ export const HorizontalScrollTemplate: ViewTemplate<HorizontalScroll> = html`
     <template class="horizontal-scroll">
         ${startTemplate}
         <div class="scroll-area">
+            ${when(
+                x => x.view !== "mobile",
+                html<HorizontalScroll>`
+                    <div
+                        class="scroll scroll-prev"
+                        part="scroll-prev"
+                        ${ref("previousFlipper")}
+                    >
+                        <button
+                            class="scroll-action"
+                            @click="${x => x.scrollToPrevious()}"
+                            tabindex="-1"
+                        >
+                            <slot name="previous-flipper">
+                                <fast-flipper
+                                    direction="previous"
+                                    aria-hidden="false"
+                                ></fast-flipper>
+                            </slot>
+                        </button>
+                    </div>
+                `
+            )}
             <div
                 class="scroll-view"
                 @scroll="${x => x.scrolled()}"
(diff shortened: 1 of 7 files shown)


scomea:
It looks like we're relying on authors to set "aria-hidden" on the flippers they provide.  Should the scroll component force or default that setting to false if unset?

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
```diff
@@ -10,6 +10,24 @@ export const HorizontalScrollTemplate: ViewTemplate<HorizontalScroll> = html`
...
+                    <div
+                        class="scroll scroll-prev"
+                        part="scroll-prev"
+                        ${ref("previousFlipper")}
+                    >
+                        <button
+                            class="scroll-action"
+                            @click="${x => x.scrollToPrevious()}"
+                            tabindex="-1"
+                        >
```
chrisdholt, line 25: If I slot a flipper we have buttons inside buttons essentially - I think we discussed potentially doing this in the feature branch to leverage the new API for default slotted content. I think that makes this story a lot easier.

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
```diff
@@ -27,23 +45,18 @@ export const HorizontalScrollTemplate: ViewTemplate<HorizontalScroll> = html`
...
                         class="scroll scroll-next"
                         part="scroll-next"
                         ${ref("nextFlipper")}
                     >
-                        <div class="scroll-action" @click="${x => x.scrollToNext()}">
+                        <button
+                            class="scroll-action"
+                            @click="${x => x.scrollToNext()}"
+                            tabindex="-1"
+                        >
```
chrisdholt, line 57: Same note as above - I think we should target the feature branch on this and leverage the default slotted content API's. In that case, we can also slot the actual flippers themselves.

chrisdholt (changes requested):
I think if we target the feature branch we can leverage the default slotted content API's and create a really robust story here. I'm still of the mind that by *default* the flippers should be hidden from AT. If we leverage the new API for default slotted content, we can actually do some clever things to make bringing these into the tab index.

1. We can default the actual flippers, which is more reuse
2. We can expose an attribute that can be used to toggle visibility of the defaulted flippers
3. Because the scroll handlers are public methods, we can leverage those on the default slotted content - when someone slots their own, they can still leverage it.

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
robarbms: @chrisdholt  I wasn't sure if I should wait until vNext was merged back into master.

PR #4822, opened 2021-06-03
add avatar

litteredwitherrors:
# Pull Request

## 📖 Description

This PR adds FAST avatar.

### General

### Component-specific


packages/web-components/fast-components/src/avatar/avatar.styles.ts
```diff
@@ -0,0 +1,78 @@
...
+            line-height: var(--avatar-size, var(--avatar-size-default));
+            display: block;
+            min-height: var(--avatar-size, var(--avatar-size-default));
+        }
+
+        ::slotted(fast-badge) {
+            position: absolute;
+            display: block;
+            bottom: 0;
+            right: 0;
```
nicholasrice, line 76: Does this need to change in RTL to `left`?

packages/web-components/fast-foundation/src/avatar/avatar.spec.md
```diff
@@ -0,0 +1,115 @@
...
+|`badge`| Slot for fast badge       | `fast-badge` |
+|`media`| Slot for images and icons | `img`, `svg` |
+
+### Anatomy and Appearance
+
+*Template*
+```js
+<div
+    class="coin ${x => x.shape}"
+    part="coin"
```
nicholasrice, line 58: Where does "coin" come from? Is that a term that is being used widely in other component libraries?

I think I'd prefer a more semantic name like "backplate" to "coin", because to me it better conveys the purpose of the part.

packages/web-components/fast-components/src/avatar/avatar.styles.ts
litteredwitherrors: Good, catch. I believe it should. I'll update.

packages/web-components/fast-foundation/src/avatar/avatar.spec.md
litteredwitherrors: When I initially wrote the spec for avatar I was under the impression that the term was used in other avatars. After reviewing again, no so much.

Backplate works. I'll update.

packages/web-components/fast-components/src/avatar/index.ts
```diff
@@ -0,0 +1,68 @@
...
+export const imgTemplate = html<Avatar>`
+    ${when(
+        x => x.imgSrc,
+        html`
+            <img
+                src="${x => x.imgSrc}"
+                alt="${x => x.alt}"
+                slot="media"
+                class="image"
+                part="image"
```
chrisdholt, line 39: should we update these (and the styles) to both be "media"?

chrisdholt (approved):

nicholasrice (approved):

PR #4853, opened 2021-06-10
chore: update the creator with the new design tokens and remove the design-system-provider component

janechu:
# Pull Request

## 📖 Description

This updates the Creator app by removing the `<fast-design-system-provider>` component as a wrapper and uses the design tokens when updates occur.

## 👩‍💻 Reviewer Notes

This change does not update the theme, @bheston do you have a reference as to what needs to happen for theme switching between light and dark mode to occur?

### General

### Component-specific

### DIFF
--- a/sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
+++ b/sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
@@ -1,12 +1,12 @@
 import {
     accentPalette,
     baseLayerLuminance,
-    cornerRadius,
+    controlCornerRadius,
     disabledOpacity,
     fillColor,
-    focusOutlineWidth,
-    outlineWidth,
+    focusStrokeWidth,
     PaletteRGB,
+    strokeWidth,
     SwatchRGB,
 } from "@microsoft/fast-components";
 import { parseColorHexRGB } from "@microsoft/fast-colors";
@@ -45,13 +45,13 @@ export function mapFASTComponentsDesignSystem(
                 break;
             }
             case "control-corner-radius":
-                cornerRadius.setValueFor(element, value);
+                controlCornerRadius.setValueFor(element, value);
                 break;
             case "stroke-width":
-                outlineWidth.setValueFor(element, value);
+                strokeWidth.setValueFor(element, value);
                 break;
             case "focus-stroke-width":
-                focusOutlineWidth.setValueFor(element, value);
+                focusStrokeWidth.setValueFor(element, value);
                 break;
             case "disabled-opacity":
                 disabledOpacity.setValueFor(element, value);
--- a/sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
+++ b/sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
@@ -11,22 +11,33 @@ import {
 } from "@microsoft/fast-components";
 import { parseColorHexRGB } from "@microsoft/fast-colors";
 
-export function mapFASTComponentsDesignSystem(designSystem: {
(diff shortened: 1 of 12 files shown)


sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
```diff
@@ -0,0 +1,50 @@
...
+                if (base) {
+                    accentPalette.withDefault(
+                        PaletteRGB.create(SwatchRGB.create(base.r, base.g, base.b))
+                    );
+                }
+
+                break;
+            }
+            case "control-corner-radius":
+                cornerRadius.withDefault(value);
```
janechu, line 37: @bheston presumably your changes to the names of these will have to come first, `controlCornerRadius`, `strokeWidth`, `focusStrokeWidth`.

sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
```diff
@@ -0,0 +1,50 @@
...
+    SwatchRGB,
+} from "@microsoft/fast-components";
+import { parseColorHexRGB } from "@microsoft/fast-colors";
+
+export function mapFASTComponentsDesignSystem(designSystem: {
+    [key: string]: any;
+}): void {
+    Object.entries(designSystem).forEach(([attribute, value]: [string, any]) => {
+        switch (attribute) {
+            case "fill-color":
```
bheston, line 19: Is there any way these can come from the design token declaration `name` in case they ever change?

sites/fast-creator/app/web-components/index.tsx
```diff
@@ -121,8 +126,8 @@ export function getColorPickerControls(
     updateHandler: (updatedData: { [key: string]: unknown }) => void
 ): StandardControlPlugin[] {
     return [
-        getColorPickerControl("background-color", updateHandler),
-        getColorPickerControl("accent-base-color", updateHandler),
+        getColorPickerControl("fill-color", updateHandler),
+        getColorPickerControl("accent-color", updateHandler),
```
bheston, line 130: This isn't a realized property in the new tokens, but if you're mapping to the palette I suggest we keep the "base" aspect in this name.

bheston (approved):

bheston:
> This change does not update the theme, @bheston do you have a reference as to what needs to happen for theme switching between light and dark mode to occur?
The light and dark mode work like they have before, but I'm not sure where it needs to appear in this PR. Instead of the DSP the token value is attached to any element, so probably a container that replaces your previous DSP. You can set the `baseLayerLuminance` token to `LightMode`, `DarkMode`, or really any value between `0` and `1`.
Then you need the `fillColor` set to a value from one of the layer recipes, typically `L1` for the primary layer. I believe that works now without the need to create that shadow design system we had previously, that is, one that sets the base layer luminance value to provide to the fill layer recipe.
I feel like this falls short of your question, but hopefully it helps. I don't think I have a current sample to point to though.

sites/fast-creator/app/configs/library.fast.design-system.mapping.ts
nicholasrice: FWIW the design tokens themselves have a 'name' property that could be used, though there is no such token "accent-color"

janechu: Some of these are obfuscating what is actually happening under the hood since a 1 to 1 UX pattern with design tokens in this case is undesirable.

janechu:
Awaiting the merge of #4854

nicholasrice (approved):

Issue #4808, opened 2021-05-30
feat: add default flippers to horizontal-scroll component


robarbms:
# 🙋 Feature Request

Add default flippers, fast-flipper, to the horizontal-scroll component.
## 🤔 Expected Behavior

This will mean that the developer would get fast-flippers if none are defined.

## 😯 Current Behavior

The developer must slot in a pair of flippers.

PR #4870, opened 2021-06-14
Adding keyboard interactions to horizontal-scroll

robarbms:
# Pull Request

## 📖 Description

Adds default previous and next flippers to horizontal-scroll with tabbing enabled.
Adds click behaviour for enter key on focused flippers.
Adds arrow key navigation to horizontal-scroll.

### 🎫 Issues

https://github.com/microsoft/fast/issues/4808
https://github.com/microsoft/fast/issues/4713

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/horizontal-scroll/index.ts
+++ b/packages/web-components/fast-components/src/horizontal-scroll/index.ts
@@ -1,3 +1,4 @@
+import { html } from "@microsoft/fast-element";
 import {
     HorizontalScroll as FoundationHorizontalScroll,
     HorizontalScrollOptions,
@@ -37,8 +38,17 @@ export const fastHorizontalScroll = HorizontalScroll.compose<HorizontalScrollOpt
     baseName: "horizontal-scroll",
     template,
     styles,
-    nextFlipper: `<fast-flipper @click=${x =>
-        x.scrollToNext()} aria-hidden="false"></fast-flipper>`,
-    previousFlipper: `<fast-flipper @click=${x =>
-        x.scrollToPrevious()} direction="previous" aria-hidden="false"></fast-flipper>`,
+    nextFlipper: html`
+        <fast-flipper
+            @click="${x => x.scrollToNext()}"
+            aria-hidden="false"
+        ></fast-flipper>
+    `,
+    previousFlipper: html`
+        <fast-flipper
+            @click="${x => x.scrollToPrevious()}"
+            direction="previous"
+            aria-hidden="false"
+        ></fast-flipper>
+    `,
 });
--- a/packages/web-components/fast-components/src/horizontal-scroll/index.ts
+++ b/packages/web-components/fast-components/src/horizontal-scroll/index.ts
@@ -1,5 +1,6 @@
 import {
     HorizontalScroll as FoundationHorizontalScroll,
+    HorizontalScrollOptions,
     horizontalScrollTemplate as template,
 } from "@microsoft/fast-foundation";
 import {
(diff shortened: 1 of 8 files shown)


packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
```diff
@@ -36,7 +42,9 @@ export const horizontalScrollTemplate: (
                         ${ref("previousFlipper")}
                     >
                         <div class="scroll-action" @click="${x => x.scrollToPrevious()}">
```
chrisdholt, line 44: I think we want to pull the click events off this with this change, yes?

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
```diff
@@ -45,7 +53,9 @@ export const horizontalScrollTemplate: (
                         ${ref("nextFlipper")}
                     >
                         <div class="scroll-action" @click="${x => x.scrollToNext()}">
```
chrisdholt, line 55: I think we want to pull the click events off this with this change, yes?

packages/web-components/fast-components/src/horizontal-scroll/index.ts
```diff
@@ -32,8 +33,12 @@ export class HorizontalScroll extends FoundationHorizontalScroll {
  * @remarks
  * HTML Element: \<fast-horizontal-scroll\>
  */
-export const fastHorizontalScroll = HorizontalScroll.compose({
+export const fastHorizontalScroll = HorizontalScroll.compose<HorizontalScrollOptions>({
     baseName: "horizontal-scroll",
     template,
     styles,
+    nextFlipper: `<fast-flipper @click=${x =>
+        x.scrollToNext()} aria-hidden="false"></fast-flipper>`,
```
chrisdholt, line 41: Do you think we should set `aria-hidden` to x => x.hiddenFromAT` (I think that does what's expected...)

nicholasrice (approved):

packages/web-components/fast-components/src/horizontal-scroll/index.ts
robarbms: Wouldn't that set it back to the flipper default of true?

chrisdholt: Good call - we could create an attribute which is slightly different for the Horizontal scroll perhaps and tie to that? Just a thought...

robarbms: Extended the aria-hidden to the horizontal-scroll. It made sense to just pass down the same attribute since only the flippers would use assistive technology.

packages/web-components/fast-components/src/horizontal-scroll/index.ts
```diff
@@ -32,8 +34,21 @@ export class HorizontalScroll extends FoundationHorizontalScroll {
...
  */
-export const fastHorizontalScroll = HorizontalScroll.compose({
+export const fastHorizontalScroll = HorizontalScroll.compose<HorizontalScrollOptions>({
     baseName: "horizontal-scroll",
     template,
     styles,
+    nextFlipper: html`
+        <fast-flipper
+            @click="${x => x.scrollToNext()}"
+            aria-hidden="${x => x.flippersHiddenFromAT}"
```
chrisdholt, line 44: love it!

chrisdholt, line 44: love it!

chrisdholt (approved):

chrisdholt:
Thanks @robarms I love seeing the new default control work w/ custom templates and attrs!

PR #4872, opened 2021-06-15
Update recipe design tokens

bheston:
# Pull Request

## 📖 Description

- Updated recipes to use design tokens instead of DI
- Made color recipe types less specific

### General

### Component-specific


packages/web-components/fast-components/src/color/recipes/accent-foreground.ts
```diff
@@ -1,19 +1,19 @@
...
-    palette: PaletteRGB,
+    palette: Palette,
     reference: Swatch,
     contrastTarget: number,
     restDelta: number,
     hoverDelta: number,
     activeDelta: number,
     focusDelta: number
-): Record<"rest" | "hover" | "active" | "focus", SwatchRGB> {
+): Record<"rest" | "hover" | "active" | "focus", Swatch> {
```
nicholasrice, line 16: I'm curious why you're converting all of these to `Swatch` from `SwatchRGB`; I thought we needed RGB info to convert these to alpha colors eventually?

packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -16,16 +17,30 @@ import {
...
 import { neutralStroke as neutralStrokeAlgorithm } from "./color/recipes/neutral-stroke";
-import { SwatchRGB } from "./color/swatch";
 import { StandardLuminance } from "./color/utilities/base-layer-luminance";
 import { accentBase, middleGrey } from "./color/utilities/color-constants";
+import { SwatchFamily } from "./color/recipe";
+
+/** @public */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export interface Recipe<
+    T extends string | number | boolean | BigInteger | null | Array<any> | symbol | {}
```
nicholasrice, line 34: This typing seems a bit odd. What is the reason for enumerating all of these types and asserting T must extend them? Can this just be `T`?

packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -357,37 +373,40 @@ const accentFillByContrast = (contrast: number) => (
...
-    (element: HTMLElement, fill?: SwatchRGB) => ReturnType<typeof accentFillAlgorithm>
->("accent-fill", builder =>
-    builder.instance(accentFillByContrast(ContrastTarget.normal))
-);
+export const accentFillRecipe = create<ColorFamilyRecipe>({
+    name: "accent-fill-recipe",
+    cssCustomPropertyName: null,
+}).withDefault({
+    evaluate: (element: HTMLElement, reference?: Swatch): SwatchFamily =>
+        accentFillByContrast(ContrastTarget.normal)(element),
```
nicholasrice, line 381: I think you need to pass `reference` to the resolved color recipe, otherwise the second argument is pointless

packages/web-components/fast-components/src/color/recipes/accent-foreground.ts
bheston: Per our chat, my initial goal here was to simplify the interface to avoid restrictions to future consumers. Potentially we could support other color models or something.
I had forgotten about making this change in the context of transparency support, but I wasn't specific enough in our chat long ago about needing the RGB info. The underlying issue with the previous system was that we passed around solid hex values everywhere. We could parse those and use them in child component recipes, but as soon as a container needed to use a transparent color, there was no longer a way to parse that or use it to create derivative colors.
Now that we have the more descriptive palette and swatch model, and we're getting an actual object any time we evaluate a recipe, I believe there's a small augmentation we can add to Swatch to allow transparent colors to be rendered in css, but passed around with sufficient information to make derivative colors.
We agreed to leave this change in, as even if we end up needing to diverge to fully support transparency, the addition of RGB information alone is not sufficient (at least in my current assumed model) to do what we would need to,

packages/web-components/fast-components/src/design-tokens.ts
bheston: I copied that from the `DesignToken` interface. I thought it was more than was probably necessary, but my attempt was to say that a `Recipe` could evaluate to anything that a `DesignToken` could be. Incredibly open to suggestions or a better way to do this.

packages/web-components/fast-components/src/design-tokens.ts
bheston: Good catch.

packages/web-components/fast-components/src/design-tokens.ts
nicholasrice: I think this can just be `T`. The limitation exist in DesignToken solely to exclude function value generics (so that it can interpret derived values), but that limitation doesn't exist here.

PR #4877, opened 2021-06-17
fix: add update the state of the CSS control when the control based on the inline string value

janechu:
# Pull Request

## 📖 Description

This change passes the `value` from the CSS inline style and maps it to the non-shorthand CSS declarations. This will not set shorthand declarations as it will not parse them, this is to be investigated at a later date.

## 👩‍💻 Reviewer Notes

Run the manual app in `fast-tooling-react` and change the CSS values, then navigate to a sub object and navigate back, the values should re-populate.

This solves an issue where updating the navigation and navigating back did not set the CSS values the CSS is currently mapped to.

## 📑 Test Plan

There is currently no test plan as this is an experimental control and uses web components, which currently do not work with our unit testing for React.

### General

### Component-specific


EisenbergEffect (approved):
Approved pending at least one test 😏

janechu:
> Approved pending at least one test 😏

I can't test this, see my description, currently attempting to test this with Jest causes errors that on precursory attempts to diagnose did not reveal any solutions.

nicholasrice (approved):

Issue #4776, opened 2021-05-25
<fast-number-field> does not apply read only to step-up/step-down


vnbaaij:
# 🐛 Bug Report

When setting a Number field to read only, the value can still be changed with the up/down buttons 

## 💻 Repro or Code Sample
Go to https://explore.fast.design/components/fast-number-field and change the Code to `<fast-number-field placeholder="name" value="1" readonly></fast-number-field>`
You can't change the value by typing into the field directly, but you can still change the value with the up/down buttons

## 🤔 Expected Behavior

The up/down buttons should be disabled (or not shown)

## 😯 Current Behavior
The up/down buttons change a read only value

PR #4881, opened 2021-06-18
fix: number-field improved value handling and event management

robarbms:
# Pull Request

## 📖 Description

Centralizes logic to validate the value being updated. Ensures that it's a number and that it's between the min and max values. Sets the proxy value and fires "input" and "change" events for the component.

Ensures that the value is within range if the min or max attributes change.

When the number-field has the readonly or disabled attribute, the step up and step down controls still work. This piggy-backs off of the hideStep attribute to remove the controls when the readonly or disabled attribute is enabled.

### 🎫 Issues

https://github.com/microsoft/fast/issues/4776

### General

### Component-specific


packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -169,38 +171,38 @@ export class NumberField extends FormAssociatedNumberField {
...
+                value = this.max;
+            }
+
+            value = parseFloat(value.toPrecision(12)).toString();
         }
 
-        return parseFloat(value.toPrecision(12)).toString();
+        if (value != this.value) {
+            this.value = value;
+            if (this.proxy instanceof HTMLElement) {
```
scomea, line 200: Shouldn't this be checking for HTMLInputElement as that has a "value" prop and HTMLElement does not?

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -169,38 +171,38 @@ export class NumberField extends FormAssociatedNumberField {
...
+                this.proxy.value = this.value;
+            }
+            this.$emit("input");
+            this.$emit("change");
```
scomea, line 204: What does the "input" event mean?  Beforehand it was only emitted when using the step methods, now on every change?

robarbms: HTML number fields fire both the input and change events when ever the user interacts with the field or the step controls. I'm just trying to make both events available per an issue logged. I moved the value updates, validation and event firing to ensure consistency between the different methods to update.

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: It should be HTMLInputElement. Updated

scomea (approved):

EisenbergEffect (approved):

PR #4882, opened 2021-06-18
add avatar to vNext

litteredwitherrors:
# Pull Request

This PR adds FAST Avatar.

### General

### Component-specific


chrisdholt (approved):

packages/web-components/fast-components/src/avatar/fixtures/base.html
```diff
@@ -0,0 +1,118 @@
...
+    <fast-avatar
+        src="https://via.placeholder.com/32"
+        alt="Fang's profile image"
+        link="#"
+        name="Fang Law"
+        shape="square"
+        fill="accent-secondary"
+    ></fast-avatar>
+
+    <h3>Square avatar with with initials and badge</h3>
```
robarbms, line 87: Should be 'Square avatar with initials'

robarbms (approved):

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,89 @@
...
+     * Indicates the Avatar should have a name.
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: name
+     */
+    @attr public name: string;
+    protected nameChanged(): void {
+        if (!this.initials) {
+            const result = this.name?.match(/\b\w/g) || [];
```
nicholasrice, line 30: I think you need to track a clean/dirty state here. As is, if no initials are set, then name changes (still no initials explicitly set by author) then the initials property won't update with the name change.

I also wonder if this violates the "keep content in light DOM" guiding principle a bit too much. Should the the avatar just expect text content instead of having competing initials & name properties?

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -0,0 +1,30 @@
...
+            x.fill ? `background-color: var(--avatar-fill-${x.fill});` : void 0}"
+    >
+        <a
+            class="link"
+            part="link"
+            href="${x => (x.link ? x.link : void 0)}"
+            style="${x => (x.color ? `color: var(--avatar-color-${x.color});` : void 0)}"
+        >
+            <slot name="media" part="media">${definition.media || ""}</slot>
+            <span class="name" part="name">${x => x.initials}</span>
```
nicholasrice, line 26: In relation to my comment below about the relationship between `initials` and `name`, this feels odd to me, that we call this part `name` but use `initials`. Why does the component accept both pieces of data if it only really uses `initials`?

@chrisdholt It would be good to get your thoughts here, but my thinking is that we probably should just accept content in the form of a default slot and leave the content formatting up to app authors.

packages/web-components/fast-foundation/src/avatar/avatar.ts
```diff
@@ -0,0 +1,89 @@
...
+    @attr public initials: string;
+
+    /**
+     * Indicates the Avatar should have a color fill.
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: fill
+     */
+    @attr public fill: string;
```
bheston, line 51: This is probably a bit of reworking, but I feel like these two colors should be based on the token infrastructure somehow. For instance, the badge color should probably adjust like an accent color for light or dark mode. This may be similar to how we handle a custom color for card, though I think it can be simpler than that if we create a new token that is a copy of the way accent fill works.

I see I didn't realize this on the presence badge PR, so I'll adjust there. Possibly something we want to clean up later, but I think it would be breaking.

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
litteredwitherrors: I'm inclined to agree, after spending more time with web components since initially writing this, it does seem like bad form. I went ahead and removed both `initials` and `name` from `avatar` in favor of using the default slot for text content.

packages/web-components/fast-foundation/src/avatar/avatar.ts
chrisdholt: Could this still come in as an attribute and we update it in a future PR to adopt? While behavior is different, I see it as a feature which is additive to support accurate palettes...

For transparency, I'm curious if this implementation work needs to block vNext

packages/web-components/fast-components/src/avatar/index.ts
```diff
@@ -0,0 +1,68 @@
...
+     * Indicates the Avatar should have alt text
+     *
+     * @public
+     * @remarks
+     * HTML Attribute: alt
+     */
+    @attr public alt: string;
+}
+
+export const imgTemplate = html<Avatar>`
```
chrisdholt, line 30: I think you'll want to add tsdocs here as `@public` for the documentation - that's where the build error is coming from.

PR #4895, opened 2021-06-24
fix: resolve security and depencencies

Ibrahimmaga:
## 📖 Description
CVE-2021-29060
moderate severity
Vulnerable versions: < 1.5.5
Patched version: 1.5.5
In the npm package color-string, there is a ReDos (Regular Expression Denial of Service) vulnerability regarding an exponential time complexity for linearly increasing input lengths for hwb() color strings.

CVE-2021-23368
moderate severity
Vulnerable versions: >= 7.0.0, < 7.0.36
Patched version: 7.0.36
The npm package postcss from 7.0.0 and before versions 7.0.36 and 8.2.10 is vulnerable to Regular Expression Denial of Service (ReDoS) during source map parsing.

CVE-2021-33587
high severity
Vulnerable versions: >= 4.0.0, <= 5.0.0
Patched version: 5.0.1
The css-what package 4.0.0 through 5.0.0 for Node.js does not ensure that attribute parsing has Linear Time Complexity relative to the size of the input.

CVE-2021-33502
high severity
Vulnerable versions: < 4.5.1
Patched version: 4.5.1
The normalize-url package before 4.5.1, 5.x before 5.3.1, and 6.x before 6.0.1 for Node.js has a ReDoS (regular expression denial of service) issue because it has exponential performance for data: URLs.

CVE-2021-24033
moderate severity
Vulnerable versions: >= 0.4.0, < 11.0.4
Patched version: 11.0.4
react-dev-utils prior to v11.0.4 exposes a function, getProcessForPort, where an input argument is concatenated into a command string to be executed. This function is typically used from react-scripts (in Create React App projects), where the usage is safe. Only when this function is manually invoked with user-provided values (ie: by custom code) is there the potential for command injection. If you're consuming it from react-scripts then this issue does not affect you.

### General

### Component-specific


awentzel:
I'd actually prefer each security update to have 1 PR focusing on that update.  It's best practice because If we do these all at once and a problem occurs, we can't revert just the offender.  We end up reverting all of them or worse, having to create a new PR to fix forward.  Gut instinct is to revert the PR that broke the system and that's generally the fastest way to resolve wider impact issues across team productivity.

Ibrahimmaga:
I totally agree with you. Let do that instead then.

Ibrahimmaga:
This will not be considered. A better approach is to resolve dependencies individually in case we need to revert the changes.

PR #4899, opened 2021-06-24
fix: upgrade css-what to version 5.0.1

Ibrahimmaga:
# Pull Request

## 📖 Description

CVE-2021-33587
high severity
Vulnerable versions: >= 4.0.0, <= 5.0.0
Patched version: 5.0.1
The css-what package 4.0.0 through 5.0.0 for Node.js does not ensure that attribute parsing has Linear Time Complexity relative to the size of the input.

### General

### Component-specific


robarbms (approved):

yarn.lock
```diff
@@ -8438,7 +8438,7 @@ css-select@^2.0.0, css-select@^2.0.2:
   integrity sha512-Dqk7LQKpwLoH3VovzZnkzegqNSuAziQyNZUcrdDM401iY+R5NkGBXGmtO05/yaXQziALuPogeG0b7UAgjnTJTQ==
   dependencies:
     boolbase "^1.0.0"
-    css-what "^3.2.1"
+    css-what "^5.0.1"
```
chrisdholt, line 8441: Changes like these are just going to get blown away in the next install - I don't think this or lines 8451, 8462, should be changing

chrisdholt (changes requested):

yarn.lock
awentzel: @chrisdholt I'm not exactly following what the issue here is.  Are there updates on the way that you know about that will resolve this security issue by upgrading the package version?  This is a high severity security issue that needs to be updated per the details in the description.  Do you have an alternative solution?  Why should these not be changed?  What's the impact beyond security?

radium-v: These lines are part of the resolution metadata for the dependency graph. Let's take this entry for example:

```
css-select@^2.0.0, css-select@^2.0.2:
  version "2.1.0"
  resolved "https://registry.yarnpkg.com/css-select/-/css-select-2.1.0.tgz#6a34653356635934a81baca68d0255432105dbef"
  integrity sha512-Dqk7LQKpwLoH3VovzZnkzegqNSuAziQyNZUcrdDM401iY+R5NkGBXGmtO05/yaXQziALuPogeG0b7UAgjnTJTQ==
  dependencies:
    boolbase "^1.0.0"
    css-what "^3.2.1"
    domutils "^1.7.0"
    nth-check "^1.0.2"
```

The first line is the list of entries for this resolution. Any packages that depend on `css-select` and specifically call for versions `^2.0.0` or `^2.0.2` are resolved to this entry.

The second line is the version of the resolution. In this case, both `^2.0.0` and `^2.0.2` are being resolved to match `2.1.0`, which is the latest version that meets the specified ranges from line 1.

The `resolved` and `integrity` lines are the archive URL and an integrity hash to ensure that the files downloaded (or restored from local cache) are exactly the same every time they're installed.

The last part is the `dependencies` list. These are the names and versions that this specific resolved version of the package says it requires. In our case, `css-select@2.1.0` will _always_ depend on `boolbase@^1.0.0`, `css-what@3.2.1`, `domutils@^1.7.0`, and `nth-check@^1.0.2`. This will never change. If this changes, then by principle, the `resolved`, `version` and `integrity` all must be different as well.

In the top-level `package.json`, the `resolutions` field is used to force all matching entries to resolve to one specific version range. Our dependency graph relies on three distinct versions of `css-what`:
```
css-what@2.1:
  version "2.1.3"
  resolved "https://registry.yarnpkg.com/css-what/-/css-what-2.1.3.tgz#a6d7604573365fe74686c3f311c56513d88285f2"
  integrity sha512-a+EPoD+uZiNfh+5fxw2nO9QwFa6nJe2Or35fGY6Ipw1R3R4AGz1d1TEZrCegvw2YTmZ0jXirGYlzxxpYSHwpEg==

css-what@^3.2.1:
  version "3.4.2"
  resolved "https://registry.yarnpkg.com/css-what/-/css-what-3.4.2.tgz#ea7026fcb01777edbde52124e21f327e7ae950e4"
  integrity sha512-ACUm3L0/jiZTqfzRM3Hi9Q8eZqd6IK37mMWPLz9PJxkLWllYeRf+EHUSHYEtFop2Eqytaq1FizFVh7XfBnXCDQ==

css-what@^4.0.0:
  version "4.0.0"
  resolved "https://registry.yarnpkg.com/css-what/-/css-what-4.0.0.tgz#35e73761cab2eeb3d3661126b23d7aa0e8432233"
  integrity sha512-teijzG7kwYfNVsUh2H/YN62xW3KK9YhXEgSlbxMlcyjPNvdKJqFx5lrwlJgoFP1ZHlB89iGDlo/JyshKeRhv5A==
```

Adding an entry to the `resolutions` field will force all three of these to be combined into one entry, and it would look something like this:

```
css-what@2.1, css-what@^3.2.1, css-what@^4.0.0, css-what@^5.0.1:
  version "5.0.1"
  resolved "https://registry.yarnpkg.com/css-what/-/css-what-5.0.1.tgz#3efa820131f4669a8ac2408f9c32e7c7de9f4cad"
  integrity sha512-FYDTSHb/7KXsWICVsxdmiExPjCfRC4qRFBdVwv7Ax9hMnvMmEjP9RfxTEZ3qPZGmADDn2vAKSo9UcN1jKVYscg==
```

But you see that the first line for the entry still lists all of the versions; they just point to one entry now. This is why the values under other `dependencies:` shouldn't be manually altered, because they'll still point to the correct entry as long as it exists and resolves to a version. Yarn will automatically notice when a resolved entry has incorrect values for its dependencies, and will replace the values to match _what the resolved package says its dependencies are_. It'll then verify that the entry matches its integrity hash.

Now if that `resolutions` change does indeed cause other dependencies to be removed or changed, those changes should be included since Yarn will correct any missed changes automatically. But again because of the integrity hash, that would mean that the entry itself must have updated as well (at least the `version`, `resolved`, and `integrity`).

radium-v:
As mentioned in cheeriojs/cheerio#1924, this only affects `css-what >=4.0.0<=5.0.0`. Removing and regenerating the entries for `svgo@2.3.0` and `cheerio@^1.0.0-rc.3` should be all that's needed to eliminate `css-what@^4.0.0`.

yarn.lock
awentzel: Great explanation, thanks @radium-v

awentzel:
> As mentioned in [cheeriojs/cheerio#1924](https://github.com/cheeriojs/cheerio/issues/1924), this only affects `css-what >=4.0.0<=5.0.0`. Removing and regenerating the entries for `svgo@^2.3.0` and `cheerio@^1.0.0-rc.3` should be all that's needed to eliminate `css-what@^4.0.0`.

@Ibrahimmaga if you don't have any questions go ahead and update per @Ibrahimmaga guidance above and let's try and get this into master today or tomorrow.  Thanks

Ibrahimmaga:
This will be closed. Previous PR has fixed the update for css-what.

Ibrahimmaga:
This will be closed. Previous PR has fixed the update for css-what.

awentzel:
@Ibrahimmaga please add the link to the associated PR.

PR #4902, opened 2021-06-24
Design token cleanup

bheston:
# Pull Request

## 📖 Description

Updated the fillColor type handling in the design-system-provider.

Also cleaned up token sequence in code. The second commit is **only** a sequence change.

## 👩‍💻 Reviewer Notes

Review the first commit. The second one is sequence change only.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/color/recipes/neutral-stroke-divider.ts
+++ b/packages/web-components/fast-components/src/color/recipes/neutral-stroke-divider.ts
@@ -3,14 +3,14 @@ import { Palette } from "../palette";
 import { directionByIsDark } from "../utilities/direction-by-is-dark";
 
 /**
- * The neutralDivider color recipe
+ * The neutralStrokeDivider color recipe
  * @param palette - The palette to operate on
  * @param reference - The reference color
  * @param delta - The offset from the reference
  *
  * @internal
  */
-export function neutralDivider(
+export function neutralStrokeDivider(
     palette: Palette,
     reference: Swatch,
     delta: number
--- a/packages/web-components/fast-components/src/design-tokens.ts
+++ b/packages/web-components/fast-components/src/design-tokens.ts
@@ -41,39 +41,7 @@ export type InteractiveColorRecipe = Recipe<InteractiveSwatchSet>;
 
 const { create } = DesignToken;
 
-/** @public */
-export const accentFillRestDelta = create<number>("accent-fill-rest-delta").withDefault(
-    0
-);
-/** @public */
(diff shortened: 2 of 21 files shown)


EisenbergEffect (approved):

packages/web-components/fast-components/src/color/recipes/neutral-stroke-divider.ts
```diff
@@ -3,14 +3,14 @@ import { Palette } from "../palette";
...
- * The neutralDivider color recipe
+ * The neutralStrokeDivider color recipe
  * @param palette - The palette to operate on
  * @param reference - The reference color
  * @param delta - The offset from the reference
  *
  * @internal
  */
-export function neutralDivider(
+export function neutralStrokeDivider(
```
chrisdholt, line 13: I think this broke the color explorer - do we need to change it anywhere else in addition to there?

bheston: That's still not breaking in a local build of the whole repo. I must not be running the same thing as CI. Fixed. Color Explorer is cheating a bit, I don't think there's anywhere else in this repo doing that.

chrisdholt (approved):

packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -961,5 +920,56 @@ export const neutralLayer4 = create<Swatch>(
...
+export const neutralOutlineActive = neutralStrokeActive;
+/** @public @deprecated Use neutralStrokeFocus */
+export const neutralOutlineFocus = neutralStrokeFocus;
+/** @public @deprecated Use neutralStrokeDividerRest */
+export const neutralDivider = neutralStrokeDividerRest;
+/** @public @deprecated Use neutralLayer1 */
+export const neutralLayerL1 = neutralLayer1;
+/** @public @deprecated Use neutralLayer2 */
+export const neutralLayerL2 = neutralLayer2;
+/** @public @deprecated Use neutralLayer3 */
```
nicholasrice, line 972: Can we talk about removing these? It seems a bit odd to me to introduce *new* exports that are already deprecated. If the goal is to ease migration, can we just provide a migration doc that maps v1 assets to v2 assets?

bheston: Yeah, that was the goal. I was thinking about that too. I realized they were new. I guess what they would allow is for some find and replace around "Behavior.var" and then at least things would compile. I still like the idea I proposed about some sort of migration wrapper, but I guess a doc is fine.

nicholasrice (approved):

PR #4906, opened 2021-06-24
Date picker spec

robarbms:
# Pull Request

## 📖 Description

Outlining the construction of a date-picker component.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Work in progress.

Proof of concept: https://codepen.io/kungfukarl/pen/4ce294a9acbe5756cd9699c8665f89f6

## 📑 Test Plan

N/A

### General

### Component-specific


ecdanchik:
Is the ability to mark a specific date in the calendar control as disabled / unavailable part of this spec? This may be implicit, but wanted to be sure.

robarbms:
> Is the ability to mark a specific date in the calendar control as disabled / unavailable part of this spec? This may be implicit, but wanted to be sure.

I have been working off existing date-picker controls for FluentUI and W3C.

[FluentUI date-picker](https://developer.microsoft.com/en-us/fluentui#/controls/web/datepicker)
[W3C date input](https://www.w3.org/wiki/HTML/Elements/input/date)

They don't outline any way to specifically target single dates as being disabled / unavailable.

I will look at the feasibility of adding a disable day(s) feature.

claviska:
> I will look at the feasibility of adding a disable day(s) feature.

We should consider how disabled dates might coexist with [date range selections](https://github.com/microsoft/fluentui/issues/14175), should either be supported. In this case, allowing min and max dates makes sense, but disabling arbitrary dates could lead to a confusing user experience.

specs/date-picker/date-picker.md
```diff
@@ -0,0 +1,214 @@
...
+- *A customer using the component on a web page.*
+- *A developer building an app with the component and interacting through HTML/CSS/JavaScript.*
+- *A designer customizing the component.*
+
+### API
+*Component Name*
+- `fast-date-picker`
+
+*Props/Attrs*
+- `value` - A predefined date or one picked by the user.
```
bheston, line 62: Is this component appropriate for a range of dates from an accessibility perspective? That's a common need, but not sure if we plan to address this here either initially or in the future.

specs/date-picker/date-picker.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Anatomy 
+
+**Structure:**
+
+```html
+<div class="date-picker">
+  <div class="label"></div>
+  <div class="date-input">
+    <input type="text" placeholder="${x => x.placeholder}" value=${x => x.value} />
```
bheston, line 93: We should allow `start` and `end` content as well.

specs/date-picker/date-picker.md
```diff
@@ -0,0 +1,214 @@
...
+    <span class="calendar-icon"
```
bheston, line 94: We just switched the 'clear' button to a native html button in the search component. Hopefully we can follow the same pattern here for consistency of styles. I see I may be making an assumption about the difference between clicking the icon and into the text portion. Are we going to show the calendar on default focus into the control, or only on the icon? If only on clicking the icon then I think the pattern defined in search is appropriate here.

bheston (approved):

chrisdholt:
@robarbms can we make sure this is up to date and get it "review ready" so we can get it merged. Considering we're a ways down the road ideally its not too much to get this up and in.

specs/date-picker/date-picker.md
```diff
@@ -0,0 +1,285 @@
...
+- `type` - Type of picker. Supports all of the HTML types. `date` | `month` | `week` | `time` | `datetime-local`
+- `selected-dates` - Dates to be selected.
+- `disabled-dates` - Dates that cannot be selected.
+
+*Methods*
+
+
+*Events*
+
+- `change` - callback fired when the entered date has changed.
```
nicholasrice, line 93: I'm noticing that the platform emits `change` events "...when an when an alteration to the element's value is *committed by the user.*" (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event). Can we align to that behavior? I think it will make the component easier to use if it models native control behavior where possible.

We could also implement the [input event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) behavior, which is more like what you are describing here, to get the best of both worlds.

specs/date-picker/date-picker.md
```diff
@@ -0,0 +1,285 @@
...
+- `placeholder` - Text to appear when a date isn't present.
+- `name` - the name of the control
+- `appearance` - enum
+    - outline (default)
+    - filled
+- `readonly` - If the date is editable.
+- 'disabled` - If the control is disabled.
+- `required` - The date field is required or optional.
+- `min` - Earliest allowed date.
+- `max` - Latest allowed date.cx
```
nicholasrice, line 81:
```suggestion
- `max` - Latest allowed date.
```

specs/date-picker/date-picker.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **Year view:** Opens a year selection view. Used for quickly changing the calendar view, month select view or for selecting a year when type="year".
+
+- **Localization:** Formats that date for the local market. MM--DD-YYYY, DD--MM-YYYY, etc. It supports country, language, numbering type and calendar type.
+
+### Risks and Challenges
+
+*Notable risks or challenges associated with implementing the component. Would we need to make any breaking changes in order to achieve this component's goals?*
+
+- Many different calendar types, even within the same market.
```
nicholasrice, line 47: Do you have an idea of how much code supporting the different calendar types in one control will be? My general concern is bloating out a single component w/ myriad support options and loading un-necessary code for markets and use-cases that don't need it. If the code to support them is small it is likely a non-issue, but if not, perhaps we can brainstorm some ways to support the myriad calendars in a way that still keeps individual implementations lightweight.

nicholasrice (approved):
Thanks @robarbms. I'd like to see the alignment of change / input events before we move forward with this (unless there are reasons not to that I'm not considering), but otherwise it looks good.

specs/date-picker/date-picker.md
robarbms: I reduced it to what I believe to be a minimum viable product. I'm using the built-in internationalization functions of the browser for getting dates, month names, etc. This will cover all of the markets that use the Gregorian calendars. But I think there will be problems with markets using other calendars like Buddhist, Islamic, Hebrew, etc. There are 13 defined calendar types in the browser. But it's more complicated than that. A specific calendar can be different in different markets. For example, there are different leap years with long and short months for the Buddhist calendar in Cambodia than the Buddhist calendar in Burma. 

The only way that I might be able to differentiate is to walk backwards and forward from the current date looking for month and year changes. This might be slow, so I left it as a future nice to have feature.

specs/date-picker/date-picker.md
robarbms: I was going to proxy these from the fast-text-field, but I started thinking, these events should just bubble. I shouldn't have to emit anything. I can call this in the spec though.

EisenbergEffect (approved):
Please add comments to the unfilled sections or remove them if there's nothing worth adding there.

specs/date-picker/date-picker.md
robarbms: It was bubbling the change event from the text-field component but not when picking. Updated the date-picker and the spec to reflect this.

robarbms:
I've resolved the change and input issues and updated it in the spec.

bheston (approved):

Issue #4038, opened 2020-10-15
bug: dialog does not track all possible updates to internal tab queue


scomea:
The dialog component updates it's list of tabbable elements based on adding/removing child elements but not other cases, like elements having tab index changed programmatically.  This [fix](https://github.com/microsoft/fast/pull/4036) addressed some pressing issues with the dialog tab queue but we should revisit the issue:

- is there a different approach that does not have an external dependency (tabbable)?
- we should be more resilient about catching changes to the tab queue, either by watching for more dom updates or maybe updating the list of tabbable items on each keystroke?

tlmii:
Per discord discussion, I'm adding the details of my scenario that falls under the "... but not other cases, ..." comment in the OP's comment.

In my case, my dialog is hidden initially, so when tabbable is called to determine which children should be in the tabbableElements list, it returns nothing. That's expected behavior with [tabbable](https://github.com/focus-trap/tabbable):

> Any of the above will not be considered tabbable, though, if any of the following are also true about it:
> - has a negative tabindex attribute
> - has a disabled attribute
> - **either the node itself _or an ancestor of it_ is hidden via display: none or visibility: hidden** <---- My issue
> - is nested under a closed \<details\> element (with the exception of the first \<summary\> element)
> - is an <input type="radio"> element and a different radio in its group is checked

But there's no way to get past that with the fast dialog because there's no later request for tabbable elements once the dialog is no longer hidden.

scomea:
we now update the tab queue on each tab keystroke, so this is resolved.

PR #4911, opened 2021-06-25
fix: dialog focus queue should detect focusable fast elements

scomea:
## 📖 Description

This change modifies the dialog component's tab queue to be able to recognize focusable fast elements using a mechanism similar to what we use in toolbar.  Also, we no longer track the tab queue continuously but rather update it from scratch on each tab keystroke. 

Note that this approach wouldn't handle a fast components with multiple tabbable children in its shadow dom, but I don't believe we do that anywhere.

### 🎫 Issues
https://github.com/microsoft/fast/issues/4038

## 📑 Test Plan
Added a basic playwright test, should expand that to run through more combos of possible child types. 

### General

### Component-specific

## ⏭ Next Steps
I believe we should also a a mechanism for authors to have more control over how we detect the first/last elements of the tab queue.  We know our solution has holes (ie. won't work with components from other libs) so we should enable developers to roll their own if they need to.


radium-v (approved):
Always great to see new playwright tests!

nicholasrice:
> Note that this approach wouldn't handle a fast components with multiple tabbable children in its shadow dom, but I don't believe we do that anywhere.

Can we document that limitation somewhere so it doesn't bite us or package consumers later?

robarbms (approved):

scomea:
Added a blurb to the spec about focus queue

chrisdholt (approved):

PR #4913, opened 2021-06-26
Improve listbox, combobox, and select accessibility with screen readers

### LINKED ISSUE #4824: fix: Additional role=combobox in Combobox web component
# 🐛 Bug Report

Two role="combobox" are added to the Combobox web component, one in the shadowroot on the input element, and another on the template/host. The one on the template/host I don't know how it gets added, the one on the input element it's obvious from the code and it has the accompanying aria-expanded attribute as expected.

I would expect only one to be there and the missing aria-expanded attribute seems to be an a11y violation.

## 💻 Repro or Code Sample

https://explore.fast.design/components/fast-combobox

## 🤔 Expected Behavior

Only one role="combobox" should be rendered per webcomponent

## 😯 Current Behavior

two are being rendered:

image

## 💁 Possible Solution

From experimenting, it looks like the Combobox is inheritting from ListBox which defines the role class variable as @attr, not sure if there's a way to undo that, but that is causing the role to be rendered in the dom on the host and then the combobox uses that to also render the input's role. We would either need to undo that attr directive or not inherit from ListBox

## 🔦 Context

Getting accessibility violations in the a11y insights tool. Looks like NVDA is also not reading the expanded attribute correctly either.

## 🌍 Your Environment

* OS & Device: Windows
* Browser Edge and Chromium

radium-v:
# Pull Request

## 📖 Description

Improves the screen reader capabilities for the listbox, combobox, and select components.

### 🎫 Issues

Closes #4824.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/combobox/fixtures/base.html
+++ b/packages/web-components/fast-components/src/combobox/fixtures/base.html
@@ -203,8 +203,8 @@
     <fast-option>Louisiana</fast-option>
     <fast-option>Maine</fast-option>
     <fast-option>Maryland</fast-option>
-    <fast-option>Massachussets</fast-option>
-    <fast-option>Michigain</fast-option>
+    <fast-option>Massachusetts</fast-option>
+    <fast-option>Michigan</fast-option>
     <fast-option>Minnesota</fast-option>
     <fast-option>Mississippi</fast-option>
     <fast-option>Missouri</fast-option>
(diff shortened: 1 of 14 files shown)


radium-v:
In Edge on Windows, there's currently a bug with the `<fast-select>` component where it reads out the selected element twice:

1. Start Narrator.
2. Tab to the "choose a T-Shirt size" control on the "Select" fixture.
3. Open the control with the space bar.
4. Use arrow keys to choose the second option.

The screen reader announces "Medium selected. Medium, 2 of 4, selected."

PR #4918, opened 2021-06-28
Fixed foregroundOnAccent recipe to work in all states

bheston:
# Pull Request

## 📖 Description

The issue was that there was a previous relationship between accentFill and foregroundOnAccent that tried to maintain the same foreground color regardless of fill. This was always problematic with the fill offset deltas as it was possible to specify deltas so far apart that no single foreground color could meet contrast. Going forward I think we want a more layered approach, where the fill color is based on that component's container color, and the foreground color is based on that fill color. For example, it's possible for the foreground to flip between black and white given a large fill delta.

The underlying issue here is that setting the `background-color` in styles does not adjust the context for all other color recipes as is necessary for proper dependent colors. That is, both recipes will be evaluated based on the containing elements fillColor. So the foregroundOnAccent tokens added in this PR assume that the fill under them is the matching accentFill, which is passes appropriate to the state of each foreground recipe. Ideally the recipes all function independently, but that requires the fillColor context to be updated at the component level rather than set in css. This is what we do in `Card` and we know we need to do it everywhere. We need to evaluate if there are performance or other optimization considerations before applying that globally though.

### General

### Component-specific


ben-girardet:
I've tested your PR and the accent foreground on rest is fixed 👍 

Following up on your comment above, I've tested to set a very large delta for the Hover Fill in order to have a light hover fill and dark rest fill. In this scenario the foreground stays the same for both states, meaning that one state is not correctly contrasted. For most cases this should work as its pretty uncommon to have such big deltas. However I wonder if this could be better handled by combining the fill / foreground recipes.

In other word, instead of having two algorithm (`accentFillAlgorithm` and `foregroundOnAccentAlgorithm`), it might be an idea to have one `accentAlgorithm` returning both `accentFillRecipe` and `accentForegroundRecipe`. 

This would allow the algorithm to compute the foreground against each respective fills.

_This suggestion is based on my little knowledge of the overall color / token design. It might not make sense... but maybe it does_

nicholasrice (approved):

bheston:
> I've tested to set a very large delta for the Hover Fill in order to have a light hover fill and dark rest fill. In this scenario the foreground stays the same for both states

Strange, that's exactly what I tested with. I used our default pink and a light yellow to test both directions. I didn't notice a case where it wasn't flipping. Just to be sure, what accent color and delta value did you use to test? I'll check again.

> However I wonder if this could be better handled by combining the fill / foreground recipes.

The recipes are currently designed to be as independent as possible. There are some interdependencies in the current visual design, but I think many of those will be going away with an upcoming visual refresh.

The further idea with the recipes would support something like you're describing, though each attribute still needs to be rendered as its own css variable. For instance, we're looking at how the recipes evolve into "modules" that can apply whatever styling is necessary to define the visual element. A concrete example might be something like a focus treatment that adds a border and a box-shadow. Ideally the recipe and token infrastructure becomes inverted in this model, where a single recipe can expose multiple tokens with some sort of wrapper used to apply them consistently in styles. Line height and font size are another paired example that might apply to this design.

ben-girardet:
I'll get back to you in an hour or so with accent and delta values for testing.

Regarding recipes and tokens, I agree that there is a need for a mechanism to set multiple token values at once. From my experience extending FAST and using it in different projects, I regularly need to change the overall accent recipes. If would help to have some sort of "entry point" in order to extend a whole group of tokens such as all neutrals or all accents (as they are strongly coupled anyway).

I also love the idea of being able to add recipes for focus and such. Maybe even usable for things like ripple effects...

ben-girardet:
> Strange, that's exactly what I tested with. I used our default pink and a light yellow to test both directions. I didn't notice a case where it wasn't flipping. Just to be sure, what accent color and delta value did you use to test? I'll check again.

@bheston 

Fill color: #FFF8F0
Accent color: #510638
Base Layer Luminance: 1
Accent Fill Hover Delta: 70 (I know it's silly, testing edge case)

Result: 
button-hover

bheston:
@ben-girardet I added this code to try your values (I'm testing on the Button page in Storybook)

```
    const fill = parseColorHexRGB("#FFF8F0")!;
    const fillSwatch = SwatchRGB.create(fill.r, fill.g, fill.b);
    fillColor.setValueFor(document.body, fillSwatch);
    baseLayerLuminance.setValueFor(document.body, 1);
    const base = parseColorHexRGB("#510638")!;
    const swatch = SwatchRGB.create(base.r, base.g, base.b);
    accentPalette.setValueFor(document.body, PaletteRGB.create(swatch));
    accentFillHoverDelta.setValueFor(document.body, 70);
```

And got this result:

image

The base layer luminance shouldn't affect this, but I do see that your container is neutral and mine is the FFF8F0 you mentioned. Is it possible to make a repro I can copy into Storybook?

ben-girardet:
I have inconsistent results and can't give you a good repro. At the moment the `var(--foreground-on-accent-rest)` _is not defined_ and I don't know why. It might be due to the way I test this PR (I built it locally and copy/paste the `dist` in the `node_modules` of a project. Might not be the best way.

I suggest to go ahead with this PR and I'll run some more tests once its part of the package. If I discover anything wrong I'll report it.

ben-girardet:
PS: I like the fact the `document.body` is the root of Design Token. In my project I'm still using a `fast-design-system-provider` but if I can remove it => it'll be a great benefit 👍

EisenbergEffect (approved):

PR #4920, opened 2021-06-29
fix design token contains logic

nicholasrice:
# Pull Request
Fixes a bug in DesignToken that was causing CSS custom properties not to get emitted in certain scenarios.

## 📖 Description
DesignToken contains logic to determine when to emit CSS custom properties for tokens that depend on other DesignTokens, and that logic depends on knowing where in the DOM hierarchy the token is being set. Previously, `Node.contains()` was being used to determine if one element contained another, however `Node.contains()` returns false when the nodes are not in the same DOM. This causes a problem when a token is set for a node in a shadow DOM.

This change introduces `composedContains`, a method for determining if an element is contained by another element that accounts for shadow DOM composition, using that method instead to determine if one node is a child of another.

### General

### Component-specific


packages/web-components/fast-foundation/src/utilities/composed-contains.ts
```diff
@@ -0,0 +1,26 @@
+import { composedParent } from "./composed-parent";
```
EisenbergEffect, line 1: No change now...but wondering if we should move both of these helpers to fast-element in the future. They seem pretty handy.

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/utilities/composed-contains.ts
nicholasrice: Yea I think that or `fast-web-utilities` would make sense

janechu (approved):

PR #4931, opened 2021-07-01
feat: anchored region uses tranforms for positioning

scomea:
## 📖 Description

Anchored region positioning performance is improved by relying on translate transforms rather than left/right/top/bottom values.  This major change also adds a "viewport-lock" attribute which keeps the region within the viewport on the specified axis, useful for keeping anchored elements such as tooltips in view.  Additionally the sample page has been significantly improved with better examples and descriptions.

### General
- [ x ] I have included a change request file using `$ yarn change`
- [ x ] I have added tests for my changes.
- [ x ] I have tested my changes.
- [ x ] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific


SethDonohue:
I don't see anything in the code that stands out to me, and I do like the new examples and all the explanation that comes with it!

I do see that the "scaling menu" doesn't seem to scale, or maybe I am missing what it is supposed to do? Does it just scale based on the number of items in it? The tooltip for it is clipped by its container in my view.

SethDonohue (approved):

scomea:
> I do see that the "scaling menu" doesn't seem to scale, or maybe I am missing what it is supposed to do? Does it just scale based on the number of items in it? The tooltip for it is clipped by its container in my view.

Does it not get a scrollbar when you resize the window so the menu no longer fits?

image

packages/web-components/fast-components/src/anchored-region/anchored-region.stories.ts
```diff
@@ -19,74 +17,103 @@ addons.getChannel().addListener(STORY_RENDERED, (name: string) => {
...
+                }
+            });
 
-                scalingViewportPreviousXValue = target.scrollLeft;
-                scalingViewportPreviousYValue = target.scrollTop;
+        document
+            .querySelectorAll("[id^=anchor-menu-many]")
+            .forEach((el: HTMLButtonElement) => {
+                el.addEventListener("click", (e: MouseEvent) => {
+                    const menuNum = el.id.substr(16, el.id.length - 16);
```
robarbms, line 107: nit: the substr length parameter isn't needed since it's just going to the end of the id.

packages/web-components/fast-components/src/anchored-region/anchored-region.stories.ts
```diff
@@ -19,74 +17,103 @@ addons.getChannel().addListener(STORY_RENDERED, (name: string) => {
...
+                    const menu: FoundationAnchoredRegion = document.getElementById(
+                        `menu-many${menuNum}`
+                    ) as FoundationAnchoredRegion;
+                    if (menu.style.display === "none") {
+                        menu.style.display = "";
+                    } else {
+                        menu.style.display = "none";
+                    }
```
robarbms, line 115: Should this use computedStyle? It will only recognize the display if it's applied directly to the element and not if through a stylesheet.

nit: This could also be condensed into:
`menu.style.display = menu.style.display === "none" ? "" : "none";`

robarbms (approved):

nicholasrice (approved):

chrisdholt (approved):

PR #4932, opened 2021-07-02
feat: convert FAST component library to be a dynamic import

janechu:
# Pull Request

## 📖 Description

This PR adds logic where adding the FAST components library prompts the preview application to dynamically load and register components from the library. Without this action the components have not been included and will not render.

## 👩‍💻 Reviewer Notes

Run the Creator application and try to add a Card, this should not render correctly as the components have not been imported and registered. Then add the FAST components library using the button in the Libraries tab, the Card should render correctly. Additionally you can add the library first then the card, this should result in the correct rendering of the Card.

### General

### Component-specific

## ⏭ Next Steps

- Make the components unavailable to be added until the library they are comprised of are added
- Add in the Fluent UI component library using the same methods the FAST library inclusion is employing

### DIFF
--- a/sites/fast-creator/app/configs/library.fast.import.ts
+++ b/sites/fast-creator/app/configs/library.fast.import.ts
@@ -1,0 +1,6 @@
+import * as FASTComponents from "@microsoft/fast-components";
+
+/**
+ * Prevent tree-shaking
+ */
+FASTComponents;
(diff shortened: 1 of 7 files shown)


chrisdholt (approved):
LET'S GO!!!

EisenbergEffect (approved):
I wonder if this won't create some confusion and result in some bugs reports when there is no bug. For example, people trying to add components or something when no library is available. Should we think about following this up with some UX work that helps in this area?

janechu:
@EisenbergEffect See my note on next steps, it requires a new feature for the `MessageSystem` that I'm currently working on.

PR #4934, opened 2021-07-02
feat: adding preview mode to creator

williamw2:
# Pull Request

## 📖 Description
Adding preview mode toggle to creator. When creator is set to preview mode the left and right columns and the code view are hidden and the viewer container style is changed to accurately reflect the rendering markup. In addition the HTMLRender navigation and inline edit functionality is suppressed so that the user can interact with rendered markup.

### General

### Component-specific


packages/tooling/fast-tooling/src/web-components/html-render/html-render.ts
```diff
@@ -124,6 +127,48 @@ export class HTMLRender extends FoundationElement {
...
+                    this.interactiveMode = action[1] !== "preview";
+                    if (this.interactiveMode) {
+                        this.updateLayers(
+                            this.layerActivityId,
+                            ActivityType.releaseFocus,
+                            null,
+                            null,
+                            null
+                        );
+                        // give everything time to update positions
```
janechu, line 146: In the future we might want to consider listening for DOM updates and having a setTimeout that triggers a certain amount of time after the last update has been fired.

sites/fast-creator/app/creator.props.ts
```diff
@@ -13,6 +13,10 @@ export enum NavigationId {
     libraries,
 }
 
+export enum DisplayMode {
+    default,
+    preview,
```
janechu, line 18: We may have more display modes in the future, so we might want to stay away from "default". Also, assigning these to strings means that this is a little more future-proof, in case this is ever stored somewhere and these values get moved around for some reason.

```suggestion
    interactive = "interactive",
    preview = "preview",
```

sites/fast-creator/app/creator.tsx
```diff
@@ -689,11 +730,39 @@ class Creator extends Editor<{}, CreatorState> {
...
+                displayMode: newState ? DisplayMode.preview : DisplayMode.default,
+            },
+            () => {
+                // Send message
+                this.fastMessageSystem.postMessage({
+                    type: MessageSystemType.custom,
+                    options: {
+                        originatorId: creatorOriginatorId,
+                        action: newState
+                            ? "displayMode::preview"
```
janechu, line 758: I would store these as const, maybe in the `app/utilities/shared.ts` if both the `creator.tsx` and `preview.tsx` will end up using them.

sites/fast-creator/app/preview/preview.tsx
```diff
@@ -18,12 +18,14 @@ import {
...
 } from "@microsoft/site-utilities";
-import { Direction } from "@microsoft/fast-web-utilities";
+import { classNames, Direction } from "@microsoft/fast-web-utilities";
 import {
     mapFASTComponentsDesignSystem,
     setupFASTComponentDesignSystem,
 } from "../configs/library.fast.design-system.mapping";
 import { elementLibraries } from "../configs";
+import { creatorOriginatorId } from "../creator";
+import { DisplayMode } from "../creator.props";
```
janechu, line 28: We should place `DisplayMode` and `creatorOriginatorId` in the `app/utilities/shared.ts` because if we import from `creator.tsx` we will end up bulking up the distributed preview application when webpack bundles the scripts.

janechu (changes requested):

sites/fast-creator/app/creator.props.ts
williamw2: Done

sites/fast-creator/app/creator.tsx
williamw2: Done

sites/fast-creator/app/preview/preview.tsx
williamw2: Done

packages/tooling/fast-tooling/src/web-components/html-render/html-render.ts
williamw2: Agreed.

janechu (approved):

EisenbergEffect (approved):

PR #4935, opened 2021-07-02
Update documentation for vNext APIs

nicholasrice:
# Pull Request

## 📖 Description
This PR updates the website documentation to document the new Design System and Design Token APIs.

There are a few sections I wasn't sure how to address so I'll tag the appropriate folks in comments - those should get filled out before this is merged. I also opened this as a feature branch in case we want to iterate on the docs structure on this branch instead of merging.

### General

### Component-specific


sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+const myElementTemplate = (
+context: ElementDefinitionContext,
+definition: FoundationElementDefinition
+) => html`<!-- ... -->`;
+```
+
+Let’s take a closer look at what these two arguments give us.
+
+#### What is the `ElementDefinitionContext`?
+// TODO Rob - I'm not sure how to explain this argument well or what all of it's capabilities are
```
nicholasrice, line 57: @EisenbergEffect I wasn't exactly sure how to describe the ElementDefinitionContext - can you take a stab at it?

sites/website/src/docs/design-systems/fast-frame.md
```diff
@@ -4,228 +4,297 @@ title: The FAST Frame Design System
...
+| Plus 1             | `typeRampPlus1FontSize`         | `typeRampPlus1LineHeight`        |
+| Plus 2             | `typeRampPlus2FontSize`         | `typeRampPlus2LineHeight`        |
+| Plus 3             | `typeRampPlus3FontSize`         | `typeRampPlus3LineHeight`        |
+| Plus 4             | `typeRampPlus4FontSize`         | `typeRampPlus4LineHeight`        |
+| Plus 5             | `typeRampPlus5FontSize`         | `typeRampPlus5LineHeight`        |
+| Plus 6 (largest)   | `typeRampPlus6FontSize`         | `typeRampPlus6LineHeight`        |
+-------------------------------------------------
+
+#### Sizing
+- `baseHeightMultiplier`: TODO - Bheston
```
nicholasrice, line 162: @bheston I wasn't sure how to best describe these design tokens - can you take a stab at it?

sites/website/src/docs/design-systems/fast-frame.md
```diff
@@ -4,228 +4,297 @@ title: The FAST Frame Design System
...
+To accomplish these goals, FAST Frame makes heavy use of algorithmic colors called Recipes. Recipes are a combination of an algorithm and input values that produce a desired result. Just as you can bake different types of cookies with different combinations of sugar, butter, flour, and salt, you can produce different design system treatments by altering recipe values (measurements) or algorithms (instructions).
 
-#### Palettes
-Each color recipe operates on a palette. A palette is an array of hexadecimal colors ordered from light to dark by [relative luminance](https://www.w3.org/WAI/GL/wiki/Relative_luminance#:~:text=WCAG%20definition%20of%20relative%20luminance,and%201%20for%20lightest%20white). By default, FAST components leverage the `neutralPalette` and the `accentPalette`.
```
nicholasrice, line 58: @bheston Thanks for writing the docs for this color section!

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+id: creating-a-component-library
+title: Creating a Component Library
+sidebar_label: Creating a Component Library
+custom_edit_url: https://github.com/microsoft/fast/edit/master/sites/website/src/docs/design-systems/creating-a-component-library.md
+---
+
+FAST exposes powerful abstractions for component library authors to allow ample opportunity for application authors to configure, augment, and override a FAST component. For example:
+- Custom element tag names are easily configured and disambiguated.
+- Custom elements are explicitly registered to avoid tree-shaking.
+- templates, styles, default slotted content, and arbitrary component configurations are all configurable by application authors.
```
awentzel, line 11: Should this sentence be Capitalized for 'templates'?

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+Let’s take a closer look at what these two arguments give us.
+
+#### What is the `ElementDefinitionContext`?
+// TODO Rob - I'm not sure how to explain this argument well or what all of it's capabilities are
+
+Importantly, the ElementDefinitionContext can be used to inspect the tag-name for other components using the `ElementDefinitionContext.tagFor()` method. This allows usage of other library components in a template or stylesheet without knowing ahead of time what the tag-name for the element will be:
+
+```ts
+import { html } from "@microsoft/fast-element";
+import { FASTButton } from "@microsoft/fast-components";
```
bashmish, line 63: I can't find `FASTButton` export anymore, looks like a typo here and in some other places

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+During component registration with the Design System, an application author can provide a `defaultContent` field as a configuration, which is used by the template if it exists.
+
+### Compose and Export Registration
+Lastly, create and export the registration function so that application authors can register the component in their Design System:
+
+```ts
+export const myElementRegistration = MyElement.compose({
+  baseName: ‘my-element’,
```
bashmish, line 98: maybe good to find an alternative example name which doesn't require a dash in the `baseName`, because `my` looks like a base prefix which it's not, especially given the exported const `myElementRegistration` has it too

compare to `fastButton` which has:

```js
export const fastButton = Button.compose({
    baseName: "button",
    // ...
});
```

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+During component registration with the Design System, an application author can provide a `defaultContent` field as a configuration, which is used by the template if it exists.
+
+### Compose and Export Registration
+Lastly, create and export the registration function so that application authors can register the component in their Design System:
+
+```ts
+export const myElementRegistration = MyElement.compose({
```
bashmish, line 97: shouldn't you pass `MyElementDefinition` interface?

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+  baseName: ‘my-element’,
+  template,
+  styles,
+  defaultContent: "Click me!"
+})
+```
+
+Note that the registration is composed with a "defaultContent" value. This value can be provided during composition or as a conditional in the template; either approach will work.
```
bashmish, line 105: I don't get neither `during composition` nor `as a conditional in the template`

the example below `myElementRegistration({defaultContent: "Don’t click me!"})` was expected, but I can't relate it to any of these 2 approaches

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Registering Library Components in an Application
+To register the component, an application author will import the registration and register it in their DesignSystem, overriding any properties as necessary:
+
+```ts
+import { myElementRegistration } from "your-package";
+import { DesignSystem } from "@microsoft/fast-foundation";
+
+DesignSystem.getOrCreate().register(
+	myElementRegistration({defaultContent: "Don’t click me!"})
```
bashmish, line 115: I think it's OK for App Dev to register a component with a certain prefix, but the rest should be of LD concern, especially all defaults like this

even if this is technically possible and has some rare use-cases, I would not recommend describing this use case without a precaution

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,117 @@
...
+  template,
+  styles,
+  defaultContent: "Click me!"
+})
+```
+
+Note that the registration is composed with a "defaultContent" value. This value can be provided during composition or as a conditional in the template; either approach will work.
+
+### Registering Library Components in an Application
+To register the component, an application author will import the registration and register it in their DesignSystem, overriding any properties as necessary:
```
bashmish, line 108: I'd like to raise a big concern here

to me a DesignSystem is smth that a Library Dev creates, while an App Dev imports it
this technically means that App Dev should not do `DesignSystem.getOrCreate()`
but this remark is less about the docs here, and more about the APIs themselves
I expected a library package to be able to export an instance of `DesignSystem` and still be able to connect this instance to a node, translating to smth like

```js
// library side
export const designSystem = new DesignSystem()
  .register(/* register all components here with all defaults */)
  .withPrefix('my');

// app side
import { designSystem } from 'my-library';
designSystem.provideFor(myAppRoot);
```

I realise that this is not how all organisation do it, but it's one of the common approaches I saw in practice and I don't understand how to best achieve it with FAST right now

bashmish: one of the ideas of how it can be done with current APIs is to export a function from the lib like this (and this is where my comment comes closer to docs which can give some guidance here):

```js
// library side
export const provideDesignSystem = (rootElement) => {
  DesignSystem
    .getOrCreate(rootElement)
    .register(/* register all components here with all defaults */).withPrefix('my');
};

// app side
import { provideDesignSystemFor } from 'my-library';
provideDesignSystem(myAppRoot);
// or for the entire document
provideDesignSystem();
```

sites/website/src/docs/design-systems/fast-frame.md
```diff
@@ -4,228 +4,297 @@ title: The FAST Frame Design System
...
+
+```html
+<fast-button>Click me!</fast-button>
+```
+
+## Configuring Components
+FAST Frame components are highly configurable, so let’s look at some of the opportunities for configuration that can be leveraged. If you’d like to understand these APIs further, check out the [creating a component library](/docs/design-systems/creating-a-component-library) documentation. 
+
+### Configuring the DesignSystem
+The `DesignSystem` is the entry-point for most component configuration. It can be used control the options for which the custom element is defined in the browser, which template and stylesheet to use, and how to disambiguate custom elements that are defined for a single element tag name.
```
bashmish, line 42: It can be used **to** control

sites/website/src/docs/design-systems/fast-frame.md
```diff
@@ -4,228 +4,297 @@ title: The FAST Frame Design System
...
+j
```
bashmish, line 43: typo?

sites/website/src/docs/design-systems/fast-frame.md
```diff
@@ -4,228 +4,297 @@ title: The FAST Frame Design System
...
+If two elements registered with the DesignSystem share the same tag name, the name can be disambiguated using `DesignSystem.withElementDisambiguation()`, allowing opportunity to provide a new name if there are conflicts:
+
+```ts
+DesignSystem
+    .getOrCreate()
+    .withElementDisambiguation((name: string, type: Constructable<{}>) => {
+        return `disambiguated-${name}`;
+    })
+    .register(/* ... */
+```
```
bashmish, line 86: I think I understand quite well why this is needed, but I could not understand how I'm supposed to use this in real life, until I read the section ```What is the `ElementDefinitionContext`?```  which demos

```ts
const template = (context: ElementDefinitionContext) => {
	const buttonTag = context.tagFor(FASTButton);
	return html`<${buttonTag}>Click me!</${buttonTag}>`
}
```

and yet I'm not sure I get all of this right
maybe link this section to the other one and give a full picture

sites/website/src/docs/design-systems/overview.md
```diff
@@ -4,61 +4,14 @@ title: What is the Design System?
...
+## How does FAST facilitate development of Design Systems?
+### Highly Configurable Web Components
+[FAST Frame](/docs/design-systems/fast-frame) is a library of highly configurable Web Components. The visual design can be impacted in myriad ways using the exported [Design Tokens](docs/design-systems/design-tokens), element templates and styles can be augmented or overridden, and shadow roots can be opened or closed as necessary for your application. If you’re looking to get up and running with using FAST Web Components, [start here.](/docs/design-systems/fast-frame)
 
-## How the Design System is used
-There are two ways that Design System values (Design Tokens) are used:
-1. In CSS
-2. In JavaScript
+### Foundation Components
+By exporting raw JavaScript Custom Element classes, `@microsoft/fast-foundation` handles the business-logic driving a large catalog of UI components. These Custom Element classes can taken as-is and composed with your templates and styles to match your application or library requirements, reducing development time and overhead of maintaining the core business logic of the component.
```
bashmish, line 14: can **be** taken as-is

sites/website/src/docs/design-systems/creating-a-component-library.md
EisenbergEffect: I think of the library dev as "defining" the design system and the app dev as "instantiating" it. The reason we have the `.getOrCreate` API is that we need to ensure that only one design system exists for any given Node in the DOM. So, we "get or create" the design system instance associated with the provided DOM Node. If none is provided, then it defaults to the document body. Would it be meaningful to add this explanation as a note in the docs?

bashmish: looks like it's not only about terminology here, but about different layers and what is expected to be done on each of them

for example `.withPrefix('my')` is not smth I want App Dev to do, but currently it's part of the "instantiating" and as such it's suggested to App Devs

I'm also starting to think that a real-world example of `getOrCreate(otherElement)` should help understand this API better
all examples in this PR (hope I didn't miss anything) are not touching upon why I would like to scope my Design System instance, and this is much related to my earlier question in Discord about why Design Tokens have their own scope (`token.setValueFor(otherElement)`) and how they are related to a Design System and it's scope

EisenbergEffect: The `provideDesignSystem` function you show is I think a good pattern. I had intended us to ship a function like this for FAST and Fluent that baked in the default prefix at least. I think we may have missed that in the release. We should patch that up.

In any case, the library dev should provide a function like that which sets up the minimal configuration for the intended design system. One of the issues with calling `register` in this case though, is that that causes a static linking of all the components, which will prevent tree shaking. The intent is that the app dev picks which components they want to register based on what they are using in their particular app. So, it should look something like this:

```ts
provideFASTDesignSystem()
  .register(
    fastButtoin()
  );
```

I'm not sure if "provide" is the best word or not. Let's continue discussion here and figure that out, then we can follow up with a PR to both FAST and Fluent to add these functions, so that at least the prefixes are set to defaults properly. That will also give us a place to do any other minimal configuration setup in the future without breaking folks by forcing them to add new configuration in successive incremental releases.

EisenbergEffect: /cc @chrisdholt This would be a non-breaking change. We'd add some functions to each library. The old mechanism would still work but we'd update docs to show this API instead as a best practice.

bashmish: > "provide" is the best word or not

I was thinking of `initDesignSystem` or just `getOrCreateDesignSystem`, but decided to use `provideDesignSystem` since it was used in the old wrapper custom element `<fast-design-system-provider>`

the old `<fast-design-system-provider>` was also responsible for design tokens, so it contained everything AFAICT, and `provide` term was a perfect match for what it was doing, whereas I don't fully understand all the implications of calling `DesignSystem.getOrCreate`, so find it difficult to suggest a good name

EisenbergEffect: I think that's a decent justification for using the term "provide" here. I'm warming up to it. Let's see if anyone else has thoughts before we commit.

sites/website/src/docs/design-systems/creating-a-component-library.md
chrisdholt: This is a good catch, this would now be `fastButton` which is a function. @nicholasrice as FYI

sites/website/src/docs/design-systems/creating-a-component-library.md
chrisdholt: I think this is fair feedback, thanks for the thoughts @bashmish!

sites/website/src/docs/design-systems/creating-a-component-library.md
chrisdholt: I'm not sure I agree that this is solely a library's concern - in several cases we've had feedback that folks want to change the default slotted content across the entire app in a single location. That's really the intent behind this. I think it's more common than you might think. For instance, I want to change the default glyph for the tree view item expand/collapse glyph. Without the above API (and corresponding doc) you really can't make that change holistically across the app without either wrapping the component and passing new content via a slot as a new "element", or changing the slotted content everywhere the element shows up.

EisenbergEffect: I think this scenario may just need further clarification in the docs.

sites/website/src/docs/design-systems/creating-a-component-library.md
EisenbergEffect: I'm going to try to address most of the general feedback so hopefully we can just do one more pass with Nick when he returns.

sites/website/src/docs/design-systems/creating-a-component-library.md
chrisdholt: > /cc @chrisdholt This would be a non-breaking change. We'd add some functions to each library. The old mechanism would still work but we'd update docs to show this API instead as a best practice.

@EisenbergEffect I have a use case I'm curious about with the above. Lets say I'm a library author extending the FAST or Fluent design system with my own components, but I *want* my own prefix - IE, I don't want to prefix my components with Fluent, but I do want to retain that prefix for the Fluent components. Is there a mechanism we can have in `provideFluentDesignSystem`, where I register my own components under the same design system bug with a custom prefix?

EisenbergEffect: Yes. It would look like this:

```ts
provideFluentDesignSystem(someChildElement)
  .register(
    fluentButton()
  ).withPrefix('min')
  .register(
    myComponent()
  );
```

EisenbergEffect: Or, you could do it like this:

```ts
provideFluentDesignSystem()
  .register(
    fluentButton()
  );

DesignSystem.getOrCreate()
  .withPrefix('mine')
  .register(
    myComponent()
  );
```

sites/website/src/docs/design-systems/creating-a-component-library.md
bashmish: @chrisdholt I didn't say App Devs don't want it :D

not to disagree, but to share my experience: the situation I had many times is when a Feature Dev comes and asks to add a possibility to change smth because his Feature UXer asked to add that or Feature Dev saw it on designs and the library doesn't have the API to achieve it

then you ask a Global UXer (who is usually very strict as his goal is to ensure consistency across the app or multiple apps) and it turns out it was not a good idea to change it on a feature level and the Feature UXer and Global UXer align to prevent the need for deviations from an existing standard

most of the time such alignment results in no change in the library or how library is instantiated in the app, but there are some exceptions, hence I see it as a rare use-case, even though there are parts which are more likely to be accepted for changes on a feature level than others

maybe in this docs a better example is needed to justify the flexibility

chrisdholt: Totally understand, and I've been on all sides myself :)

I think with regard to our goals, we want to aid in extensibility and enable easily updating or changing for any reason - greater alignment perhaps for some, experimentation with ease for others, etc. I can say with confidence that we have folks across the spectrum you're describing :). I think @EisenbergEffect's call out on clarity will help, and to your point, I think perhaps the example can improve to help articulate a reason someone *may* want to do this. Would that help?

sites/website/src/docs/design-systems/creating-a-component-library.md
bashmish: I'm thinking of

```ts
// library
export const provideMyDesignSystem = (childElement?: HTMLElement) => {
  const ds = DesignSystem.getOrCreate(childElement).withPrefix('my');

  // ... setup whatever other defaults are necessary ...

  return {
    register: ds.register,
  };
};

// export all components registration functions
export { myAccordion, myButton }; 
```

if one needs to have full control, there is always a possibility to import `DesignSystem` from '@microsoft/fast-foundation' directly and setup everything from scratch, as individual components are necessary to export for tree-shaking anyway according to an earlier remark

sites/website/src/docs/design-systems/creating-a-component-library.md
EisenbergEffect: I reworked a good bit of this section. Have a read through the article again and let me know if things are any clearer.

sites/website/src/docs/design-systems/creating-a-component-library.md
EisenbergEffect: I reworked a good portion of the article. Have a read through the latest and see if things are any clearer.

sites/website/src/docs/design-systems/creating-a-component-library.md
EisenbergEffect: That could absolutely be the pattern that you follow with your library, and we could provide guidance on that. I don't think we'd want to only export `register` from our own component libraries (FAST and Fluent), as that would cut off some scenarios we know people need that extra configuration for. But, I do think this makes sense.

bashmish: Fully agree!

I came to the following layers when analysing a bigger picture and see I overlooked 2 of them which give more sense to certain APIs:

1. foundation components layer: maximum flexibility, sensible defaults
2. company components layer: use the flexibility of the foundation layer, but enforce certain things (e.g. prefix) and close them for modification, export individual component for app to decide what to register
3. app setup layer: register only the components needed by the app
4. feature level: only use the components available in the app

What is restricted and allowed on layers 2-4 is up to the company building their Design System, but the foundation layer of components you are building in FAST, including `@microsoft/fast-components` should give this flerxibility

This brings me to the next question/suggestion: although be default almost everything should be customisable on your 1st layer, there should be ways to restrict customisation of anything on layers 2-3, while preserving customisation for what is needed to remain customisable. Docs might help to understand this better, but maybe some new APIs will be needed as well.

PS: I'm now translating my vision into a Design System starter kit and will be able to tell more once I make more progress with it. E.g. if some APIs are missing or if smth is hard to achieve.

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,145 @@
...
+import { DesignSystem } from "@microsoft/fast-foundation";
+
+DesignSystem.getOrCreate()
+  .withPrefix('your')
+  .register(
+	  counter({ defaultButtonContent: "Please count." })
+  );
+```
+
+The advantage of setting `defaultButtonContent` here is that it overrides the `Counter`'s default slotted content for **all** instances of `your-counter`. Instead of having to define it everywhere in the app that the button is used, the app author provides the configuration once in a centralized location. This can be extremely useful in scenarios where components contain default icons that the app author may need to swap.
```
bashmish, line 145: is it also how some things expected to be localized/translated? if yes, good to add a note here too

robarbms (approved):

sites/website/src/docs/design-systems/creating-a-component-library.md
EisenbergEffect: I'm going to work to get the `provideFASTDesignSystem()` API in soon. Then I'll update the docs to use that throughout. I like the content of you've got here on layers. So, I'll also add a guidance section to the creating libraries doc and include some of that along with the other things we've been talking about. Thanks for all the time you've taken to provide this feedback @bashmish

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -0,0 +1,145 @@
...
+  context: ElementDefinitionContext,
+  definition: FoundationElementDefinition
+) => html`<!-- ... -->`;
+```
+
+Let’s take a closer look at what the two arguments to these lazily evaluated functions give us.
+
+#### What is the `ElementDefinitionContext`?
+
+The first argument is the `ElementDefinitionContext`. This context object provides information that is available *during* the attempt to register the element. For example, it will tell you the HTML element tag `name` that the component will be defined as. You can also see what the default `shadowRootMode` of the `DesignSystem` is. Most of the other APIs are used internally by `FoundationElement.compose` (see below) to define the element, but one particular API of note is the `tagFor()` method of the context. This API is used to inspect the tag-name of other components and allows usage of other library components in a template or stylesheet without knowing ahead of time what the tag-name for the element will be. Here's an example of how the reusable `Counter` component would leverage this to ensure that the `Button`'s tag name is correct, even if the app developer decides to rename `Button` in some way.
```
bashmish, line 68: would be great to also document how `tagFor` is used in the CSS context, like in https://github.com/microsoft/fast/pull/4963/files

sites/website/src/docs/design-systems/creating-a-component-library.md
bashmish: @EisenbergEffect thanks for constructive communication about it! Feel free to move some of my remaining remarks into future PRs, maybe this PR is good enough for now to update docs on the FAST website and help people onboard with version 2.x.x.

EisenbergEffect: I've captured these notes so that I can follow up with a PR to improve the ergonomics and accompany that with docs updates. Going to resolve this feedback for now.

EisenbergEffect (approved):

EisenbergEffect:
@nicholasrice Shell we merge?

nicholasrice:
@EisenbergEffect I am going to address the last two comments and then we'll be good. Also about to put up a PR unblocking doc-site publishing.

EisenbergEffect:
Awesome! Thank you @nicholasrice Ping me with the new PR and let me know when we're already. @awentzel After we merge this and the upcoming PR, we'll want to do a quick review of the stage site and then if that looks good we'll want to turn around a release as soon as we can so that the docs are updated for our customers. So, might be later today or perhaps tomorrow for that release depending on how things go.

sites/website/src/docs/design-systems/creating-a-component-library.md
nicholasrice: I think this *could* be used for some some localization solutions but not all, and I want to be careful about not prescribing that solution here. Good thought but I'm going to leave this as-is.

sites/website/src/docs/design-systems/creating-a-component-library.md
nicholasrice: good call, updating.

awentzel (approved):

chrisdholt (approved):

EisenbergEffect:
@bashmish Here's a PR to make things a bit nicer based on your feedback: https://github.com/microsoft/fast/pull/4973

After we merge that we can update these documents here to both use this new API as well as provide guidance to library authors on using this same pattern. I've got your notes from above saved to use in creating that guidance.

Issue #4941, opened 2021-07-07
fix: base class discovery in context.tagFor()


m4thieulavoie:
# 🐛 Bug Report

[More detail here](https://discord.com/channels/449251231706251264/482296787021856768/862336132255842334)

Maybe a bit of a far-fetched issue, but for each of our components we apply a mixin to them. What that effectively means is that for example, instead of creating an `AnchoredRegion`, we create a `CustomAnchoredRegion`. With the new v2 release, this line doesn't resolve.

```
<${context.tagFor(AnchoredRegion)}
```

I would like to propose a new argument `baseClass` for the `compose()` method that would take the actual FAST base class.
Another idea could also be to have a fallback if `tagFor` returns undefined that would try `basePrefix-component-name` ?

## 💻 Repro or Code Sample

No repro, can always provide one if needed

## 🤔 Expected Behavior

FAST should resolve the base class, no matter how many mixins there are in the way

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

chrisdholt:
closed w/ #4943

PR #4943, opened 2021-07-07
fix: allow for base class

m4thieulavoie:
** Discussion PR, the goal is to get traction and kick off a discussion here **

# Pull Request

## 📖 Description

This PR adds the ability to specify the `baseClass` of a newly created FAST component. This is useful when using a typescript mixin on top of FAST foundation components. Since under the hood FAST is looking for the exact base class name, this allows to specify and override the default value.

```
class MyCustomSwitch extends Switch {}

export const fastSwitch = MyCustomSwitch.compose<SwitchOptions>({
    ...,
    baseClass: Switch
});
```

### 🎫 Issues

- https://github.com/microsoft/fast/issues/4941

### General

### Component-specific


packages/web-components/fast-foundation/src/foundation-element/foundation-element.ts
```diff
@@ -216,7 +221,7 @@ export class FoundationElementRegistry<
         const prefix = definition.prefix || context.elementPrefix;
         const name = `${prefix}-${definition.baseName}`;
 
-        context.tryDefineElement(name, this.type, x => {
+        context.tryDefineElement(name, overrideDefinition.baseClass ?? this.type, x => {
```
EisenbergEffect, line 224: I *think* that passing the base class here instead of the actual class may cause some problems with the platform element definition. What if we allow an optional third argument to `tryDefineElement` that takes an `options` object that allows passing in an additional type for name mapping?

m4thieulavoie: I think you're right, and either way it would be a lot safer to implement, I'll add it! 👍

m4thieulavoie: @EisenbergEffect I ended up breaking the parameter list for an object for scalability, let me know what you think about it. I also chose to register the actual class _and_ the baseClass (if any) for better flexibility. I could really use your review at this point just to make sure what I'm doing still makes sense 😅

m4thieulavoie:
@EisenbergEffect got the chance to have a look? I would love to hear your initial thoughts on this one

EisenbergEffect:
I've got it on my list. It's looking like next week though due to my current commitments. I haven't forgotten it though and want to get a solution in.

packages/web-components/fast-foundation/docs/api-report.md
```diff
@@ -1016,6 +1016,7 @@ export class FoundationElement extends FASTElement {
 // @public
 export interface FoundationElementDefinition {
     readonly attributes?: EagerOrLazyFoundationOption<(AttributeConfiguration | string)[], this>;
+    baseClass?: Constructable;
```
EisenbergEffect, line 1019: Should we also enable an array? Do we like `baseClass` or would we prefer something like `alias` or `foundation`?

packages/web-components/fast-foundation/src/design-system/design-system.ts
```diff
@@ -76,6 +76,12 @@ export interface ElementDefinitionContext {
  */
 export type ElementDefinitionCallback = (ctx: ElementDefinitionContext) => void;
 
+interface ElementDefinitionParams
```
EisenbergEffect, line 79: I think we should probably export this and then add documentation comments. We'll also need the `@public` indicator in the comment to satisfy our docs generation stuff.

packages/web-components/fast-foundation/src/design-system/design-system.ts
```diff
@@ -300,6 +300,9 @@ class DefaultDesignSystem implements DesignSystem {
 
                     elementTypesByTag.set(elementName!, type);
                     elementTagsByType.set(type, elementName!);
+                    if (baseClass) {
```
EisenbergEffect, line 303: Yep. I think it's probably this simple and should address the problem. I think this will help with a couple of Fluent/FAST components as well that need an additional registration.

packages/web-components/fast-foundation/src/foundation-element/foundation-element.ts
```diff
@@ -33,6 +33,11 @@ export interface FoundationElementDefinition {
      */
     baseName: string;
 
+    /**
+     * The actual FAST base class of the component.
```
EisenbergEffect, line 37:
```suggestion
     * The actual FAST base class of the component if different from the class used to compose.
```

packages/web-components/fast-foundation/src/design-system/design-system.ts
```diff
@@ -95,11 +101,7 @@ export interface DesignSystemRegistrationContext {
      * @param callback - A callback to invoke if definition will happen.
      * @public
      */
-    tryDefineElement(
-        name: string,
-        type: Constructable,
-        callback: ElementDefinitionCallback
-    );
+    tryDefineElement(ctx: ElementDefinitionParams);
```
EisenbergEffect, line 104: Though not used directly by consumers (as far as I know), this would technically be a breaking change. Do we feel like we can get away with that on a minor release? /cc @chrisdholt

EisenbergEffect: Note: This is a better design as it would be more open to extension in the future without breaks. Just want to make sure that we're ok doing this on a minor release. I seriously doubt it will affect anyone, but again, technical breaking.

EisenbergEffect (changes requested):
Sorry it took me so long to dig in here. I definitely want to address this issue with a solution very close to what's here, so I think this can move out of a Draft state now. I've noted a few things we should discuss before merging.  Also, before merging, we'll want at least one unit test that checks this functionality. If you aren't sure how to write that, let me know and I can probably jump in and add a commit with that after the rest of the issues are addressed.

packages/web-components/fast-foundation/docs/api-report.md
m4thieulavoie: Are you proposing doing something like the following?
```suggestion
    baseClass?: Constructable | Constructable[];
```

For the name, I'll leave it up to you guys. However, `foundation` sounds a bit vague to me

EisenbergEffect: I agree. I think that was a bad idea. We'll probably keep this as `baseClass` and if we need we can make it single or array, but we can add the array support in a later PR if someone really needs it.

m4thieulavoie: Sounds good with me!

packages/web-components/fast-foundation/src/design-system/design-system.ts
m4thieulavoie: Should be fixed now 👍

EisenbergEffect:
@m4thieulavoie Ready for me to take another look?

m4thieulavoie:
> @m4thieulavoie Ready for me to take another look?

Yes! That's why I requested back your review with the GitHub feature a couple of days ago 👍

EisenbergEffect:
Oops! Sorry about that! A lot of emails fly through my inbox...I may have missed the alert for that one. It's on my list to review now 😄

packages/web-components/fast-foundation/src/design-system/design-system.ts
```diff
@@ -90,16 +108,10 @@ export interface DesignSystemRegistrationContext {
 
     /**
      * Used to attempt to define a custom element.
-     * @param name - The name of the element to define.
-     * @param type - The type of the constructor to use to define the element.
-     * @param callback - A callback to invoke if definition will happen.
+     * @param ctx - The definition context.
```
EisenbergEffect, line 111: I think we need to fix this param description since this is technically not the definition context.

packages/web-components/fast-foundation/src/foundation-element/foundation-element.ts
```diff
@@ -216,43 +221,56 @@ export class FoundationElementRegistry<
         const prefix = definition.prefix || context.elementPrefix;
         const name = `${prefix}-${definition.baseName}`;
 
-        context.tryDefineElement(name, this.type, x => {
-            const presentation = new DefaultComponentPresentation(
-                resolveOption(definition.template, x, definition),
-                resolveOption(definition.styles, x, definition)
-            );
+        context.tryDefineElement({
```
EisenbergEffect, line 224: Should we Omit the `baseClass` from the `OverrideFoundationElementDefinition` as well?

EisenbergEffect (changes requested):
Almost there. One thing we need to fix in the docs comments and then a question I have for consideration.

m4thieulavoie:
> Almost there. One thing we need to fix in the docs comments and then a question I have for consideration.

Thanks! I'm AFK until monday but I'll correct those once I'm back!

packages/web-components/fast-foundation/src/design-system/design-system.ts
m4thieulavoie: I had a bit of a hard time finding a proper description for this one. I could use some suggestions, otherwise I'll come up with something on monday 👍

vaudy: Maybe just using `Element Definition`? @EisenbergEffect any idea?

packages/web-components/fast-foundation/src/foundation-element/foundation-element.ts
m4thieulavoie: I'm not sure I follow the question 🤔

m4thieulavoie: Nevermind, just got the sense of the question (I need cafeine 😅 ). I think so yeah!

m4thieulavoie: Actually we can't `omit` it. `overrideDefinition` has the `OverrideFoundationElementDefinition` type
image

packages/web-components/fast-foundation/src/design-system/design-system.ts
m4thieulavoie: I highly inspired from @vaudy 's suggestion, but feel free to suggest anything else 👍

packages/web-components/fast-foundation/src/foundation-element/foundation-element.ts
EisenbergEffect: I'm a bit confused because the code below does exactly this:

```ts
baseClass: overrideDefinition.baseClass
```

But it seems like that should never come from the override definition but only from the original definition.

m4thieulavoie: I think I see what you mean. The line in your previous comment is required as `baseClass` is passed along _twice_ in order to get from the `compose()` call all the way to the actual condition added in the other file changed in this PR.
I hope I was clear! :) However if I missed something definitely let me know!

EisenbergEffect:
@nicholasrice Could you take another look at this? I think @m4thieulavoie has addressed all the issues.

packages/web-components/fast-foundation/src/design-system/design-system.ts
```diff
@@ -90,16 +108,10 @@ export interface DesignSystemRegistrationContext {
 
     /**
```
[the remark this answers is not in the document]
nicholasrice: yep, you can deprecate overload signatures :)

nicholasrice (approved):

m4thieulavoie:
@EisenbergEffect @nicholasrice I had a pretty big merge conflict to fix, so please make sure to double-check the changes before merging in in case I messed up something 👍

EisenbergEffect (changes requested):
Unfortunately, I don't think the tests for the improved disambiguation are correct. Can you update those from what's in our main branch?

m4thieulavoie:
> Unfortunately, I don't think the tests for the improved disambiguation are correct. Can you update those from what's in our main branch?

I think I would need more guidance on this, they seem to be correct as far as I can tell. Can you maybe point out the tests are are erroned @EisenbergEffect ?

EisenbergEffect:
Some of the assertions are missing and also the setting of `callbackCalled` in one of the tests. Here's the source to replace the two incorrect tests:

```ts
    it("Should only call callbacks for duplicate elements by default", () => {
        const elementName = uniqueElementName();
        const customElement = class extends HTMLElement {};
        const host = document.createElement("div");
        const system = DesignSystem.getOrCreate(host);
        let callbackCalled = false;

        expect(() => {
            system.register(
                {
                    register(container: Container) {
                        const context = container.get(DesignSystemRegistrationContext);
                        context.tryDefineElement(elementName, customElement, x =>
                            x.defineElement()
                        );
                    },
                },
                {
                    register(container: Container) {
                        const context = container.get(DesignSystemRegistrationContext);
                        context.tryDefineElement(
                            elementName,
                            class extends HTMLElement {},
                            x => {
                                x.defineElement();
                                callbackCalled = true;
                            }
                        );
                    },
                }
            );
        }).not.to.throw();

        expect(customElements.get(elementName)).to.equal(customElement);
        expect(callbackCalled).to.be.true;
    });

    it("Can completely ignore duplicates", () => {
        const elementName = uniqueElementName();
        const customElement = class extends HTMLElement {};
        const host = document.createElement("div");
        const system = DesignSystem.getOrCreate(host)
            .withElementDisambiguation(() => ElementDisambiguation.ignoreDuplicate);
        let callbackCalled = false;

        expect(() => {
            system.register(
                {
                    register(container: Container) {
                        const context = container.get(DesignSystemRegistrationContext);
                        context.tryDefineElement(elementName, customElement, x =>
                            x.defineElement()
                        );
                    },
                },
                {
                    register(container: Container) {
                        const context = container.get(DesignSystemRegistrationContext);
                        context.tryDefineElement(
                            elementName,
                            class extends HTMLElement {},
                            x => {
                                x.defineElement();
                                callbackCalled = true;
                            }
                        );
                    },
                }
            );
        }).not.to.throw();

        expect(customElements.get(elementName)).to.equal(customElement);
        expect(callbackCalled).to.be.false;
    });
```

m4thieulavoie:
> Some of the assertions are missing and also the setting of `callbackCalled` in one of the tests. Here's the source to replace the two incorrect tests:
> 
> ```ts
>     it("Should only call callbacks for duplicate elements by default", () => {
>         const elementName = uniqueElementName();
>         const customElement = class extends HTMLElement {};
>         const host = document.createElement("div");
>         const system = DesignSystem.getOrCreate(host);
>         let callbackCalled = false;
> 
>         expect(() => {
>             system.register(
>                 {
>                     register(container: Container) {
>                         const context = container.get(DesignSystemRegistrationContext);
>                         context.tryDefineElement(elementName, customElement, x =>
>                             x.defineElement()
>                         );
>                     },
>                 },
>                 {
>                     register(container: Container) {
>                         const context = container.get(DesignSystemRegistrationContext);
>                         context.tryDefineElement(
>                             elementName,
>                             class extends HTMLElement {},
>                             x => {
>                                 x.defineElement();
>                                 callbackCalled = true;
>                             }
>                         );
>                     },
>                 }
>             );
>         }).not.to.throw();
> 
>         expect(customElements.get(elementName)).to.equal(customElement);
>         expect(callbackCalled).to.be.true;
>     });
> 
>     it("Can completely ignore duplicates", () => {
>         const elementName = uniqueElementName();
>         const customElement = class extends HTMLElement {};
>         const host = document.createElement("div");
>         const system = DesignSystem.getOrCreate(host)
>             .withElementDisambiguation(() => ElementDisambiguation.ignoreDuplicate);
>         let callbackCalled = false;
> 
>         expect(() => {
>             system.register(
>                 {
>                     register(container: Container) {
>                         const context = container.get(DesignSystemRegistrationContext);
>                         context.tryDefineElement(elementName, customElement, x =>
>                             x.defineElement()
>                         );
>                     },
>                 },
>                 {
>                     register(container: Container) {
>                         const context = container.get(DesignSystemRegistrationContext);
>                         context.tryDefineElement(
>                             elementName,
>                             class extends HTMLElement {},
>                             x => {
>                                 x.defineElement();
>                                 callbackCalled = true;
>                             }
>                         );
>                     },
>                 }
>             );
>         }).not.to.throw();
> 
>         expect(customElements.get(elementName)).to.equal(customElement);
>         expect(callbackCalled).to.be.false;
>     });
> ```

Oh right, GH made it really hard to compare, it should be fixed now 👍

EisenbergEffect (approved):

EisenbergEffect:
Awesome! Thanks @m4thieulavoie We'll get this in as soon as CI finishes up.

EisenbergEffect:
It's in! This should go out in this evening's release.

PR #4958, opened 2021-07-13
chore(docs): image upload web component for creator

williamw2:
# Pull Request

## 📖 Description
Adding a spec for a new image upload web component for use with creator.

### General

### Component-specific


packages/tooling/fast-tooling/src/web-components/image-upload/image-upload.spec.md
```diff
@@ -0,0 +1,21 @@
...
+## Interaction
+The component will appear as a button that when clicked will open the system file picker UI configured for image file types. Once a file is chosen
+the component will perform the configured action on that file (i.e. convert it to an ObjectURL, upload it to a server, etc) and return the string URL
+that can be used in the `src` attribute of an img tag or as a CSS `background-image`.
+
+## Implementation
+The component will be a form-associated control that will use an input type=file element as its base. In addition a `ImageFileAction` interface will be defined that 
+specifies methods for taking action on the image file once it is selected. The component will expect an `onchange` callback function as well as an instance
+of the `ImageFileAction` interface. When an image file is selected by the user the component will execute the code provided by the `ImageFileAction` class 
+which will produce a URL string which will then be returned through the `onchange` callback function.
```
janechu, line 16: Some interesting formatting happened here.

```suggestion
The component will be a form-associated control that will use an input type=file element as its base. In addition a `ImageFileAction` interface will be defined that  specifies methods for taking action on the image file once it is selected. The component will expect an `onchange` callback function as well as an instance of the `ImageFileAction` interface. When an image file is selected by the user the component will execute the code provided by the `ImageFileAction` class  which will produce a URL string which will then be returned through the `onchange` callback function.
```

chrisdholt (commented):
One question on this @williamw2 and @janechu - would a "file picker" meet the requirements here? I don't want to randomize, but I wonder if we should open an issue/request for a file picker component so we can ultimately leverage that here...just a thought/question

packages/tooling/fast-tooling/src/web-components/image-upload/image-upload.spec.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Out of scope
+The initial implementation of the component will only include one implementation of the `ImageFileAction` interface which will utilize the `URL.createObjectURL()` 
+API to store the image locally and return an `Object URL`. In the future additional `ImageFileAction` classes may be easily implemented to perform tasks such
+as uploading the file to a server and returning a remote URL.
```
janechu, line 21: Also formatting here is a bit wonky. Can you include any links to MDN that might be helpful?

```suggestion
The initial implementation of the component will only include one implementation of the `ImageFileAction` interface which will utilize the `URL.createObjectURL()`  API to store the image locally and return an `Object URL`. In the future additional `ImageFileAction` classes may be easily implemented to perform tasks such as uploading the file to a server and returning a remote URL.
```

janechu:
@chrisdholt I think that's valid, there is a complication in that in the future we might want to update the UI to show a thumbnail of the image. We could consider creating the image-upload first then abstracting any useful features to share the code. We could also change the name to image-picker for consistency.

janechu (approved):

Ibrahimmaga (approved):

PR #4961, opened 2021-07-15
chore: update the exporting of services and web components in tooling and match the new pathing in sites

janechu:
# Pull Request

## 📖 Description

This change updates the export for `@alpha` web components and services from the `@microsoft/fast-tooling` package so they can be imported from the root location. It also corrects a few mis-typed const names that should be lowercase and it normalizes some file structures for easier exporting.

This change affects the Creator site and Explorer site as the pathing has been updated to reflect that these services and web components can now be imported from the root of the package.

Due to issues resulting from exporting web components from the root, jest test have been updated with mock files and the `jest-environment-jsdom-sixteen` package has been added to allow for more modern global methods assumed to be available in browsers.

## 👩‍💻 Reviewer Notes

It should not be necessary to run the sites as the build gate should break if the pathing is incorrect.

### General

### DIFF
--- a/packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.style.ts
+++ b/packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.style.ts
@@ -1,6 +1,6 @@
 import { css } from "@microsoft/fast-element";
 
-export const HTMLRenderLayerInlineEditStyles = css`
+export const htmlRenderLayerInlineEditStyles = css`
     .edit-region {
         position: fixed;
         top: 0;
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,5 @@
 *.spec.ts
+*.spec.tsx
 **/__test-images__
 **/__tests__
 **/.tmp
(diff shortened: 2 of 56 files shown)


EisenbergEffect (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.style.ts
```diff
@@ -1,6 +1,6 @@
 import { css } from "@microsoft/fast-element";
 
-export const HTMLRenderLayerInlineEditStyles = css`
+export const htmlRenderLayerInlineEditStyles = css`
```
robarbms, line 3: Is there a reason for renaming this? It doesn't match other variable declarations (HTMLRenderOriginatorId, HTMLRenderLayerInlineEditStyles, HTMLRenderLayerInlineEditTemplate). Is there a pattern for using title casing versus camel casing?

robarbms (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.style.ts
janechu: Oh, looks like I just missed a few. I will take another pass at this afterwards but in future the consistency on naming should be that classes and interfaces are PascalCase and variables are camelCase. We may have missed a few but that has been the established pattern thus far.

janechu: On second thought, I've included all naming changes in this PR since I've already made a few fixes.

chrisdholt (approved):

PR #4963, opened 2021-07-16
update component styles to leverage context.tagFor() for nested element styles

chrisdholt:
# Pull Request

## 📖 Description
This PR introduces an update to make styles more portable to other design system prefixes by leveraging the `context.tagFor()` API for stylesheets which reference slotted or nested components.

### 🎫 Issues
None have presented themselves with my testing

### General

### Component-specific


janechu (approved):

EisenbergEffect (approved):

robarbms (approved):

scomea (approved):

litteredwitherrors (approved):

radium-v (commented):
This feels weird, since now if I want to use the `<fast-tooltip>` component on its own, I must have an import dependency on the `AnchoredRegion` from `fast-foundation`. But the tag for `AnchoredRegion` doesn't exist in foundation, it exists in components. Is the `baseName` required to compose the element, or am I missing something else that goes on under the hood?

EisenbergEffect:
There's an improvement coming to enable using the foundation class to look up the tag.

chrisdholt:
> This feels weird, since now if I want to use the `<fast-tooltip>` component on its own, I must have an import dependency on the `AnchoredRegion` from `fast-foundation`. But the tag for `AnchoredRegion` doesn't exist in foundation, it exists in components. Is the `baseName` required to compose the element, or am I missing something else that goes on under the hood?

Base Name is required to compose the element, prefix defaults to fast but can be enumerated. The key here is that our components already support `context.tagFor(AnchoredRegion)` to ensure that `fast-anchored-region` can become `foo-anchored-region` or `bar-anchored-region`. If our styles are explicitly written for `fast-` they will be unusable if someone tries to register with a custom prefix for their project.

The import dependency bit has already started up in conversation where components can register their dependencies here. I think that's an DX and ergonomic improvement. @eisenbergeffect can advise on that status

EisenbergEffect:
Actually, ignore my comment from above. But, yes we're working on a few additional things in this area. The idea of this is to be able to create templates and styles based on components even if we don't know the final name of the component.

nicholasrice (approved):

PR #4965, opened 2021-07-17
feat: add fluent UI to Creator

janechu:
# Pull Request

## 📖 Description

This change adds the Fluent UI web component library to Creator. There is a slight issue in Fluent UI which means the build will fail until a publish in Fluent UI fixes the issue. See https://github.com/microsoft/fluentui/issues/18987.

## 👩‍💻 Reviewer Notes

Once this is unblocked by https://github.com/microsoft/fluentui/issues/18987 you will see that two buttons now show up in the left tab, FAST components and Fluent UI components. Both can be added, there are a few caveats however, we are using the Design Tokens from FAST for editing in the right pane. This means that when adding Fluent UI components, the accent color is available but will not affect any components. Additionally, if both component libraries are added, there may be some discrepancies between packages. There is also an issue where both libraries have components that are named the same. This will need to be resolved by prefixing or affixing the library name. See next steps for a future resolution.

### General

## ⏭ Next Steps

- Allow libraries to add an additional pane to the right rail for configuration, this will be used for the Design Tokens for FAST and Fluent UI component libraries but should be open ended enough to allow for any extra library configuration to take place
- Prefix or affix the library name to the component name in the Form and Navigation


sites/fast-creator/app/configs/fast/library.fast.registry.ts
```diff
@@ -0,0 +1,13 @@
+import { DesignSystem } from "@microsoft/fast-foundation";
+import { allComponents } from "@microsoft/fast-components";
+import { setupFASTComponentDesignSystem } from "./library.fast.design-system.mapping";
+
+export function registerFASTComponents(): void {
+    setupFASTComponentDesignSystem(document.body);
+
+    DesignSystem.getOrCreate().register(
```
EisenbergEffect, line 8: Before calling `register` you can call `withPrefix` to set the prefix for all registered components, rather than setting it on each individual component.

sites/fast-creator/app/configs/fluent-ui/library.fluent-ui.registry.ts
```diff
@@ -0,0 +1,13 @@
+import { DesignSystem } from "@microsoft/fast-foundation";
+import { allComponents } from "@fluentui/web-components";
+import { setupFluentUIComponentDesignSystem } from "./library.fluent-ui.design-system.mapping";
+
+export function registerFluentUIComponents(): void {
+    setupFluentUIComponentDesignSystem(document.body);
+
+    DesignSystem.getOrCreate().register(
```
EisenbergEffect, line 8: See comment above on setting prefix. Also, if you have to set this to "fluent" that seems like a bug in our fluent web components. I think I know what's going on there. I'll put together a PR to fix it in the source.

EisenbergEffect (changes requested):
A minor tweak or two. The code also shows me that we've got a bug in the core libraries. So, I'll create a PR to fix that situation in fast and fluent shortly. We can update this code to use the improved API when ready but this can merge without that if we adjust the prefix API as noted.

nicholasrice (approved):
looks good aside from the already highlighted issues.

sites/fast-creator/app/configs/fluent-ui/library.fluent-ui.registry.ts
janechu: Changed to withPrefix for now. Can you link the PR as a comment to this when ready? I'll follow and remove the withPrefix once the fix is in.

EisenbergEffect: We merged the PR for FAST. I'll be submitting the PR for Fluent today. I'll link when I've got it up. Essentially, you'll be able to do something like this:

```ts
provideFluentDesignSystem().register(allComponents);
```

Here's the link to the FAST PR that was merged, for reference https://github.com/microsoft/fast/pull/4973

EisenbergEffect: Here's the related Fluent PR: https://github.com/microsoft/fluentui/pull/19056

janechu: When that PR is merged and the package is published we will update the registration logic. Thanks @EisenbergEffect!

EisenbergEffect (approved):

PR #4966, opened 2021-07-19
Improving DX - adding Gitpod, because no project should be left behind

shaal:
# Pull Request

## 📖 Description

Gitpod provides automated dev environments in the cloud (and free to use for public repos).
By adding `.gitpod.yml` file to this repo, we will improve developer experience when working on FAST repo, without modifying current way developers working on it.

When clicking on "Open in Gitpod" button, a user gets a whole development environment in their browser. 
Users no longer need to wait for things like `yarn install` (which can take about 8 minutes). That's because any compile/make process is running ahead of time, during Gitpod's prebuild process, which runs automatically when a new branch or PR is created.

This is an additive change, so it wouldn't affect anyone who wants to continue work on their local machine, but will lower the barriers for anyone who wants to contribute.

Outside of changes that are possible in this PR, in order to automatically generate prebuilds, this Github app https://github.com/apps/gitpod-io should be configured to run on this repo. 

### General

### Component-specific


chrisdholt (changes requested):
We always welcome contributions, so thanks for taking the time to create this PR! While I appreciate the initiative, this feels more appropriate to start as an issue. I’m not sure I fully understand the benefit to the project, the requirements for running, and what the implications are from a workflow and maintenance standpoint, etc. Is this an alternative to other/existing tools, etc? If you’d like to open an issue to discuss adding this, I’m more than happy to dig in with our team here. Without a discussion, I’m not comfortable at this time signing off on adding additional tooling, etc.

PR #4973, opened 2021-07-20
feat(fast-design-system): improve ergonomics of the fast design system

EisenbergEffect:
# Pull Request

## 📖 Description

Based on community feedback and some potential error cases, I've added a `provideFASTDesignSystem` function as the default way of getting a design system instance. Additionally, I've made `allComponents` into a registry itself so that it can be passed directly to the design system. Here's what the code would now be:

```ts
provideFASTDesignSystem().register(allComponent);
```

## 👩‍💻 Reviewer Notes

This is an additive change to mostly improve ergonomics but to also provide a more standard setup function which we can augment the content of in the future without breaking consumers.

## 📑 Test Plan

Existing tests passing. New APIs are consumed in the process of running the storybook, and so are validated there.

### General

## ⏭ Next Steps

As soon as this PR is in, we need to do two pretty quick follow up tasks:

* Update the FluentUI Web Components with the same pattern e.g. `provideFluentUIDesignSystem()`
* Update our documentations to use these functions rather than `DesignSystem.getOrCreate()...`


packages/web-components/fast-components/src/custom-elements.ts
```diff
@@ -151,4 +155,19 @@ export const allComponents = {
     fastToolbar,
     fastTreeView,
     fastTreeItem,
+    register(container?: Container) {
```
nicholasrice, line 158: way nicer!

chrisdholt (approved):
Sweet!

nicholasrice (approved):

Issue #4727, opened 2021-05-18
Missing Multiselect Component


ameyasubhedar:
Fast components is missing a multi select component. There is a Radio Group, Select, List, but that enables only single select.

A check box group or a list with check box components would be ideal for multi selections.

EisenbergEffect:
@radium-v Can you share a bit about the work going on with multi-select in the select component so we can better understand if that will address this request?

chrisdholt:
> @radium-v Can you share a bit about the work going on with multi-select in the select component so we can better understand if that will address this request?

While we await @radium-v's response, I will note that my hope is that the multiselect behavior in Select will help solve this problem. If not, I'd like to hear and understand more about what a "checkbox" group would do? To me, that sounds primarily like an HTML form element. There's no additional behaviors I would expect to be part of that control besides styles to lay it out visually from what I can tell, so it would be helpful to understand the scenario you're looking to solve for if not Multiselect in Select. Thanks!

radium-v:
I'm currently adding the `multiple` functionality to `listbox`, which will also enable it in `select`.

From a data standpoint, a checkbox group is currently achievable with `<fast-checkbox>` elements in a form. Any checked checkboxes in a form with the same `name` will be returned with [`FormData.getAll`](https://developer.mozilla.org/en-US/docs/Web/API/FormData/getAll) and will also appear when the form is submitted (for `GET` requests, the checked values are all appended to the URL like `?check-group=1&check-group=2` etc.).

```html
<form id="demo" onchange="result.value=new FormData(demo).getAll('check-group').toString()">
    <div>
        <fast-checkbox name="check-group" value="1">1</fast-checkbox>
        <fast-checkbox name="check-group" value="2">2</fast-checkbox>
        <fast-checkbox name="check-group" value="3">3</fast-checkbox>
        <fast-checkbox name="check-group" value="4">4</fast-checkbox>
    </div>
    <output name="result" for="demo"></output>
</form>
```
(Here's a more comprehensive demo: https://stackblitz.com/edit/typescript-k5c5hn?file=index.html)

This is technically the same behavior that `radio` elements exhibit, with the additional constraint that only one radio in a group can be selected at a time.

ameyasubhedar:
Thanks @radium-v , what about using the check box group in the Select Dropdown. We have common use case where we present dropdowns to the user and then they can check multiple options.

[...]

Issue #4190, opened 2020-12-16
Add multiple mode to listbox/select


radium-v:
**What feature are you requesting?**
Listbox/Select should support `multiple` and `size` properties, which will allow users to choose more than one option at a time. This will require support for related ARIA attributes as well as keyboard navigation changes. See https://www.w3.org/TR/wai-aria-practices-1.1/#listbox_roles_states_props for more info on ARIA role support.

From a presentation perspective, a select in multi-mode should look similar to our current listbox component. The button and its indicator should not be displayed, and the dropdown should be presented. The ability to open and close the select should have no effect in multi-mode.

The `size` property should be added to allow for changing the number of visible options. I'm not sure if this property should affect the listbox when in single mode. It's also unclear how this property should affect a listbox with arbitrary slotted content like containers, images, horizontal rules, etc.

PR #4974, opened 2021-07-21
add multi-selection mode to listbox and select

radium-v:
# Pull Request

## 📖 Description

Adds the `multiple` and `size` attributes to `<fast-listbox>` and `<fast-select>`.

### 🎫 Issues

#4190
#4727

### General

### Component-specific


packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
```diff
@@ -66,97 +68,141 @@ export class ListboxOption extends FoundationElement {
...
     public selected: boolean = this.defaultSelected;
-    protected selectedChanged(): void {
-        if (this.$fastController.isConnected) {
-            if (!this.dirtySelected) {
-                this.dirtySelected = true;
-            }
+    protected selectedChanged(prev: unknown, next: boolean): void {
+        if (!this.dirtySelected) {
+            this.dirtySelected = true;
+        }
```
robarbms, line 130: Nit: this can be just this.dirtySelected = true since it only does it when it's not true.

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
```diff
@@ -66,97 +68,141 @@ export class ListboxOption extends FoundationElement {
...
         }
     }
 
+    public get form(): HTMLFormElement | null {
+        return this.proxy?.form;
+    }
+
+    private get hasProxy() {
+        return this.proxy instanceof HTMLOptionElement;
+    }
```
robarbms, line 82: Checking for the proxy before accessing it always kind of bugged me. I also wondered, what if there were other reasons to not update the proxy, we'd have to change it in multiple places. This fixes that problem, but then shouldProxy() would be a more descriptive name.

It still leaves us with a bunch of if(this.hasProxy()) {} wrappers in code. Possibly a this.setProxyAttribute(attribute, value); method that we could blindly call? And then we could put what ever logic we want in the setProxyAttribute) method including this.proxy istanceof HTMLOptionElement. Just food for thought.

This needs a boolean return type.

robarbms (changes requested):

scomea:
I'm confused about states here:

image

I think the "focused but not selected" shouldn't look focused when focus leaves the component.

radium-v:
> I think the "focused but not selected" shouldn't look focused when focus leaves the component.

@scomea This should now be fixed.

> When I first tab to the multi-select component narrator says "combobox, collapsed, has popup", then I tab again to get to the menu where I get a description that names out the currently selected item by name followed by "selection has x items" where x is the total number of items not the number selected. Using arrow keys to navigate the select menu doesn't prompt narrator descriptions (multi listbox does). The aria stuff needs attention I think, and we should only have one tab stop.

There are accessibility issues with select being tracked by #4824. I have a draft PR #4913 that I'll update soon, now that this PR is in review.

ben-girardet:
Can't wait to use the select component with multi-selection feature available :-)

PR #4979, opened 2021-07-21
Add CLI

litteredwitherrors:
# Pull Request

## 📖 Description

This adds the FAST CLI.

### Testing instructions:

Navigate to the package:
```
cd packages/utilities/fast-cli
```

Install dependencies:
```
yarn install
```

Create a NPM symlink:
```
npm link
```

You can now run the `fast` command.

### General

### Component-specific


EisenbergEffect:
@litteredwitherrors I just wanted to check in and see if you felt you had a set of actions items to take this forward? Excited to be able to get this to the community!

EisenbergEffect:
One think I think we should look into before releasing anything is how we can build off of the `npm init` comment. I recently learned that it matches package pattern names to run scaffolders. So we could enable this with no previously installed packages `npm init @microsoft/fast`. Here's a link to the documentation https://docs.npmjs.com/cli/v6/commands/npm-init Perhaps, rather than a "CLI" we create an advanced `create-*` package for FAST. Part of what that sets up could also be some npm/yarn commands for generators as well. This seems like it could be an even smoother more ecosystem compatible to go. Thoughts?

litteredwitherrors:
> One thing I think we should look into before releasing anything is how we can build off of the `npm init` command. I recently learned that it matches package pattern names to run scaffolders. So we could enable this with no previously installed packages `npm init @microsoft/fast`. Here's a link to the documentation [docs.npmjs.com/cli/v6/commands/npm-init](https://docs.npmjs.com/cli/v6/commands/npm-init) Perhaps, rather than a "CLI" we create an advanced `create-*` package for FAST. Part of what that sets up could also be some npm/yarn commands for generators as well. This seems like it could be an even smoother, more ecosystem compatible way to go. Thoughts?

Okay, after looking into this a little bit, the implications here are that we would have to create separate repos for different kinds of projects (i.e. FAST design system, FAST app) as well as maintain those individually. So we'd end up with repos like `@microsoft/create-fast-design-system` and `@microsoft/create-fast-app`, as well as any additional project types that make be requested.

Currently makes handles this for us by providing the UI wizard to generate different types of projects programmatically based on the Makes skeleton repo, which is currently hosted in the `fast-cli` package.  We could move the skeleton template to its own repo and register it as a package a user could run something like `npx makes @microsft/create-fast` and generate a new project without having to download the CLI. This would also solve an issue we're currently having with having in the CLI where the FAST repo won't ignore the skeleton's `package.json`.

The `fast new` command is actually just running `npx makes @microsft/create-fast` under the hood to generate a new project, but a user would have the option to run that command if they merely wanted to generate a new fast project without installing the CLI.

EisenbergEffect:
Would we need separate repos or just separate packages?

litteredwitherrors:
> Would we need separate repos or just separate packages?

If it was a single repo I'm not sure how we would handle delivering different folder structures on request.

EisenbergEffect:
I hadn't gotten that far. I was just curious about whether we could go this path. My main thinking is that this would feel more native to the ecosystem and potential be less friction.

EisenbergEffect:
@chrisdholt @janechu  Should we close out this PR?

chrisdholt:
Closing to pursue an updated direction.

PR #4984, opened 2021-07-22
fix: toolbar focus keyboarding sequence

### LINKED ISSUE #4983: fix: tab sequence of start/end slots in toolbar out of sink when focus brought back to toolbar
# 🐛 Bug Report

In toolbar when any focusable content is slotted in the `start` or `end` slots and that content is the last focused content, then when returning focus back to the toolbar the focus is brought to the incorrect slotted item.

## 💻 Repro or Code Sample

stackblits:
https://fast-design-rybc3w.stackblitz.io/

1) focus on "before toolbar button"
2) tab to the "after toolbar button"
3) shirt-tab (reverse tab) to toolbar, notice that focus is on first default slotted item and not the "end slotted button" as it should be.

## 🤔 Expected Behavior

Should tab back to the last focused item if there is focusable content in the start or end slots:

toolbarfocussequencereversetab

The arrow keys should navigate to all focusable slotted content within the toolbar including start and end.  Seen here the arrow key is pressed after tabbing to the end slotted content, which leads to the next default slotted content item being focused.
toolbar-arrow-key-sequence

## 💁 Possible Solution
 
Ensure that the focusinHandler only controls focus when focus returns to the default slotted content. 
I am working on a PR for this fix now.

## 🌍 Your Environment

* OS & Device: MacOS, Windows,
* Browser Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox

SethDonohue:
# Pull Request

## 📖 Description
Fixes a focus sequence bug on the Toolbar

### 🎫 Issues
#4983 

## 👩‍💻 Reviewer Notes

This changes the template of the toolbar slightly to ensure that the default slotted content is what is focused correctly.
Could this be a breaking change?

## 📑 Test Plan
**WIP**  Looking at updating the tests for this component now

### General

### Component-specific


packages/web-components/fast-foundation/src/toolbar/toolbar.template.ts
```diff
@@ -23,19 +23,24 @@ export const toolbarTemplate: (
...
-                    property: "slottedItems",
-                })}
-            ></slot>
+            <span
+                class="content"
+                part="content"
+                @focusin="${(x, c) => x.focusinHandler(c.event as FocusEvent)}"
+                @click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
+                @keydown="${(x, c) => x.keydownHandler(c.event as KeyboardEvent)}"
+            >
```
chrisdholt, line 36: I think there is an assumption that start/end content *isn't* meant to be part of the toolbar - is that assumption accurate? I'm not sure - considering the role of toolbar is on the host, I would assume that start/end are "regions" here I can still drop slotted content.

chrisdholt (changes requested):
Let's talk through the intent of start/end here - I think we're making an assumption that it's not interactive content, which seems like a big assumption. 

I wonder if what we need to do is include the contents of the default slot as well as start/end in order...

SethDonohue:
> Let's talk through the intent of start/end here - I think we're making an assumption that it's not interactive content, which seems like a big assumption.
> 
> I wonder if what we need to do is include the contents of the default slot as well as start/end in order...

I have updated to include this, and found it was cleaner to do this by adding the start/end slots directly into the toolbar template and class rather than importing them and extended the StartEnd class.

packages/web-components/fast-components/src/toolbar/fixtures/base.html
```diff
@@ -93,6 +97,30 @@ <h2>FAST components</h2>
         <fast-option>Option 3</fast-option>
     </fast-select>
     <fast-checkbox>Checkbox</fast-checkbox>
+    <fast-radio-group slot="end">
+        <fast-radio value="filter">Filter</fast-radio>
+        <fast-radio value="search">
+            <fast-text-field placeholder="Search"></fast-text-field>
```
brookdozer, line 103: Keyboarding to the `fast-text-field` is inconsistent when nested in the `fast-radio`, but that's not relevant to the keyboard focus (which works great for me across all browsers).

packages/web-components/fast-components/src/toolbar/fixtures/base.html
```diff
@@ -93,6 +97,30 @@ <h2>FAST components</h2>
...
+    <fast-radio-group>
+        <fast-radio checked>One</fast-radio>
+        <fast-radio>Two</fast-radio>
+        <fast-radio>Three</fast-radio>
+    </fast-radio-group>
+    <fast-button>Button</fast-button>
+    <fast-radio-group slot="end">
+        <fast-radio value="filter">Filter</fast-radio>
+        <fast-radio value="search">
+            <fast-text-field placeholder="Search"></fast-text-field>
```
brookdozer, line 120: Same as above: keyboarding to the `fast-text-field` is inconsistent when nested in the `fast-radio`. Probably not relevant to the keyboard focus fix, (which works great for me across all browsers).

packages/web-components/fast-components/src/toolbar/fixtures/base.html
SethDonohue: yes, this is when it is nested inside of the fast-radio. The fast-radio must be selected and then you can tab to the text-field.

I did find an issue while looking at this related to the initial tab sequence that I am looking in to.

packages/web-components/fast-components/src/toolbar/toolbar.styles.ts
```diff
@@ -69,6 +69,11 @@ export const toolbarStyles: (
             margin-inline: 0;
         }
 
+        .start__hidden,
+        .end__hidden {
+            display: none;
+        }
```
chrisdholt, line 75: Can you elaborate on why these are necessary to add as well as why they are *defaulted* and *hardcoded* into the template?

packages/web-components/fast-foundation/src/toolbar/toolbar.template.ts
```diff
@@ -29,14 +28,30 @@ export const toolbarTemplate: (
...
         <slot name="label"></slot>
         <div class="positioning-region" part="positioning-region">
-            ${startTemplate}
+            <span class="start start__hidden" part="start" ${ref("startContainer")}>
+                <slot
+                    name="start"
+                    ${slotted({
+                        filter: elements(),
+                        property: "startSlottedItems",
+                    })}
```
chrisdholt, line 37: It seems like this could be generally useful to add across all instances - is there a reason we couldn't do that?

SethDonohue: True, I'll take a look at doing this, and the implications of it. Should be fine since components using start and end should also extend the StartEnd class.

packages/web-components/fast-components/src/toolbar/toolbar.styles.ts
SethDonohue: I added these so that the start and end don't show up at all, instead of being a 0x0 span when nothing is slotted. Is either way preferred or does it not really matter since they are a 0x0 div?

To get them to show hidden from the start they were added to the template, but  `this.startSlottedItemsChanged()` and `this.endSlottedItemsChanged();` could be added to the connectedCallback to run at the start, which would then toggle the __hidden classes as needed.

Would you recommend I remove all the __hidden and related code if the 0x0 span is an acceptable result?

packages/web-components/fast-foundation/src/toolbar/toolbar.template.ts
SethDonohue: So if we add this we can't really use it, can we? I added this to the StartEnd and tried to access it in the Toolbar class, but it shows the startSlottedItems as empty. I believe this is expected because it is a mixin and not a class extension.

I originally added this to add the __hidden classes based on if items were slotted, but that's not necessary, so I will likely remove this and anything related to my additions for __hidden functionality.

SethDonohue: Ok, after further investigation having the ${slotted...} directive in the StartEnd template results in there being nothing in the startSlottedItems *inside* the Toolbar class AND inside the StartEnd class? 
Is this because StartEnd is a mixin, not an extended class?

this.start.assignedNodes() shows items just fine in the Toolbar class.

SethDonohue: I am going to remove this for now since it was just for the __hidden and I can use this.start.assignedNodes() to get the nodes for proper keyboard navigation.

packages/web-components/fast-components/src/toolbar/toolbar.styles.ts
SethDonohue: removed __hidden and related code, not necessary

packages/web-components/fast-foundation/src/toolbar/toolbar.ts
```diff
@@ -198,6 +189,31 @@ export class Toolbar extends FoundationElement {
...
+     * get all the slotted elements
+     * @internal
+     */
+    protected getAllSlottedItems(): (HTMLElement | Node)[] {
+        return [
+            ...this.start.assignedNodes(),
+            ...this.slottedItems,
+            ...this.end.assignedNodes(),
+        ];
+    }
```
radium-v, line 202: * This can be a getter
* switching to `assignedElements()` means we don't have to handle weird types or worry about text nodes

```suggestion
    protected get allSlottedItems(): Element[] {
        return [
            ...this.start.assignedElements(),
            ...this.slottedItems,
            ...this.end.assignedElements(),
        ];
    }
```

packages/web-components/fast-foundation/src/toolbar/toolbar.ts
```diff
@@ -198,6 +189,31 @@ export class Toolbar extends FoundationElement {
...
+
+    /**
+     * Prepare the slotted elements which can be focusable.
+     *
+     * @internal
+     */
+    protected buildFocusableElements(): void {
```
radium-v, line 209: `reduceFocusableElements` might be a clearer indication of what's going on with this method.

packages/web-components/fast-foundation/src/toolbar/toolbar.ts
```diff
@@ -183,12 +174,12 @@ export class Toolbar extends FoundationElement {
...
 
-        const incrementer = this.getDirectionalIncrementer(key);
-        if (!incrementer) {
+        const incrementor = this.getDirectionalIncrementor(key);
+        if (!incrementor) {
             return !(e.target as HTMLElement).closest("[role=radiogroup]");
         }
 
-        const nextIndex = this.activeIndex + incrementer;
+        const nextIndex = this.activeIndex + incrementor;
```
radium-v, line 182: I can't find any concrete difference in usage for `incrementer` or `incrementor`, and most sources just point to the Wiktionary entries:

[Incrementer](https://en.wiktionary.org/wiki/incrementer):
> * Any device or circuit that carries out an incrementation operation.
> * _Related terms: incrementor_

[Incrementor](https://en.wiktionary.org/wiki/incrementor):
> * That which increments; a computer programming instruction that increments a value.
> * _Related terms: incrementer_

🤷🏻

radium-v (approved):
Looks good and runs well! I left a few non-blocking suggestions.

packages/web-components/fast-foundation/src/toolbar/toolbar.ts
SethDonohue: ahah ahh yes, i just was using my spellcheck, i'l leave this change out.

packages/web-components/fast-foundation/src/toolbar/toolbar.ts
SethDonohue: awesome! updated to this :)

packages/web-components/fast-foundation/src/toolbar/toolbar.ts
SethDonohue: agreed!

packages/web-components/fast-foundation/src/toolbar/toolbar.template.ts
SethDonohue: I am trying to repro this issue in a stackblitz, but am having issues. Tracking it on my todo list.

chrisdholt (approved):
Looks like a great fix, thanks @SethDonohue!

PR #4986, opened 2021-07-23
Moved aria-hidden attributes to the template

bheston:
# Pull Request

## 📖 Description

The motivation for this was to allow the template to describe the parts of the component that shouldn't be provided to AT so consumers that choose to override these indicators don't break AT or need to copy excessive attributes.

Full disclosure, I attempted to test this but was not able to get NVDA to announce the svgs with or without the aria-hidden attribute. I would love a second opinion on when these are necessary and how I might get them to announce.

### General

### Component-specific


packages/web-components/fast-foundation/src/checkbox/checkbox.template.ts
```diff
@@ -27,10 +27,10 @@ export const checkboxTemplate: (
             x.checked ? "checked" : ""} ${x => (x.indeterminate ? "indeterminate" : "")}"
     >
         <div part="control" class="control">
-            <slot name="checked-indicator">
+            <slot name="checked-indicator" aria-hidden="true">
```
EisenbergEffect, line 30: I haven't seen aria attributes used directly on a slot element. Are we sure that has the desired effect?

bheston: I wondered that too. How do I get the element to show up in AT at all? Then I can test if it works on slots. I figured it did because they are just an element and the spec says it applies to everything on and below the element. Slots are somewhat special though, so I could see some obscure reason it doesn't work. Our other patterns seems to be to place this on a span around the slot.

chrisdholt: The contents here are set in the compose call in fast-components. The default content for foundation is empty.

bheston: Thanks Chris for confirming. I removed the aria-hidden tag for any child of an element with [applicable roles](https://w3c.github.io/aria-practices/#children_presentational).

change/@microsoft-fast-components-042a7206-4ebb-4e25-b0a3-b85315222aa9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Moving aria-hidden attributes to the template",
```
chrisdholt, line 3:
```suggestion
  "comment": "remove unnecessary aria-hidden attributes",
```

change/@microsoft-fast-foundation-4fbf4c4d-4a17-410a-94f8-b762fe40e435.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Moving aria-hidden attributes to the template",
```
chrisdholt, line 3:
```suggestion
  "comment": "remove unnecessary aria-hidden attributes",
```

chrisdholt (approved):
Just a nit on the change file itself.

EisenbergEffect (approved):

Issue #4944, opened 2021-07-07
fix: fast-tooltip always positions above the anchor on long pages where the anchor has to be scrolled into view


caseyfaber:
# 🐛 Bug Report

The spec for FAST Tooltip says "When the position is undefined the tooltip is placed above or below the anchor based on available space." However, on long pages where you scroll, if you have to scroll to the anchor element (so if it is "below the fold"), it will always position above the anchor, even if you scroll down and there is no space in the viewport above.

## 💻 Repro or Code Sample

Here is a sample with a lot of text - https://codesandbox.io/s/fast-tooltip-with-scrolling-page-93nvs?file=/index.html

If you scroll down to the _second_ button that is in the middle of all the paragraphs (labeled "Button 2"), and position it at the top of the screen, it will still reveal above the anchor.

## 🤔 Expected Behavior

The tooltip should position above or below the anchor based on current viewport space, even if the anchor is farther down the page and you have to scroll to bring it into view.

## 😯 Current Behavior

Tooltip reveals above the anchor if the anchor is on a page that it has to be scrolled to get to.

![image](https://user-images.githubusercontent.com/2295203/124828014-91f0d780-df2b-11eb-9e7b-5d938ad4ab91.png)

## 🌍 Your Environment

* OS & Device: Windows 10
* Browser: Chrome, Edge
* Version: @microsoft/fast-components 1.21.8

EisenbergEffect:
Ping @scomea for further review/investigation.

PR #4990, opened 2021-07-26
feat:  add viewport lock to tooltip

scomea:
## 📖 Description
 -  Tooltip now has "horizontal-viewport-lock" and "vertical-viewport-lock" attributes that when set to true will keep the tooltip in the viewport
 - fix issue with the tooltip's contain css value by changing from "layout" to "size" as the layout setting prevented rendering in some cases.
 - Use the default viewport from anchored region rather than set locally (document.body was incorrect) and causing the tooltip to not always pick the tallest space to display in.

### 🎫 Issues
fixes https://github.com/microsoft/fast/issues/4944

## 📑 Test Plan
- added base test for element default setting

### General

### Component-specific

## ⏭ Next Steps
Want to fix issues with multiline tooltips.


packages/web-components/fast-foundation/src/tooltip/tooltip.ts
```diff
@@ -459,7 +475,6 @@ export class Tooltip extends FoundationElement {
         if (!this.tooltipVisible) {
             return;
         }
-        this.viewportElement = document.body;
```
robarbms, line 462: What sets viewportElement now? I don't see it being set explicitly anywhere. Is it deferred to the anchored region and is it needed anymore?

robarbms (approved):

robarbms:
I smoke tested this. Looks good.

packages/web-components/fast-foundation/src/tooltip/tooltip.ts
scomea: Falling back to the default set by anchored region.

EisenbergEffect (approved):

chrisdholt (approved):

PR #4992, opened 2021-07-26
Added support for styling slider value on the track

bheston:
# Pull Request

## 📖 Description

Adds support for a visual indicator of the selected value in the slider track area.

Note the "Next Steps" below.

### General

### Component-specific

## ⏭ Next Steps

There are two ways the components currently handle RTL. Some look for `dir=rtl` and others use the design token via the directional behavior. This change is currently not working as expected in rtl. The slider values go from right to left, but there is no way to style the track value differently because the style sheet helper expects the design token. Since the two could be out of sync I didn't want to add that here in favor of resolving rtl support consistently.

### DIFF
--- a/packages/web-components/fast-components/src/slider/slider.styles.ts
+++ b/packages/web-components/fast-components/src/slider/slider.styles.ts
@@ -19,7 +19,19 @@ import {
     neutralStrokeHover,
     neutralStrokeRest,
 } from "../design-tokens.js";
-import { heightNumber } from "../styles/index.js";
+import { DirectionalStyleSheetBehavior, heightNumber } from "../styles/index.js";
+
+const ltr = css`
+    .track-start {
+        left: 0;
+    }
+`;
+
+const rtl = css`
+    .track-start {
+        right: 0;
+    }
+`;
 
 /**
  * Styles for Slider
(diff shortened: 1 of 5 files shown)


packages/web-components/fast-foundation/src/slider/slider.template.ts
```diff
@@ -32,6 +32,9 @@ export const sliderTemplate: (
         <div part="positioning-region" class="positioning-region">
             <div ${ref("track")} part="track-container" class="track">
                 <slot name="track"></slot>
+                <div part="track-start" class="track-start" style="${x => x.position}">
+                    <slot name="track-start"></slot>
+                </div>
```
chrisdholt, line 37: This appears to be a breaking change - can we discuss the need/requirement for this?

chrisdholt (changes requested):

packages/web-components/fast-foundation/src/slider/slider.template.ts
bheston: Is it breaking to add a new part? The requirement here was to add necessary elements to display the value, with the intent being to make it easy to customize the treatment if desired. For the latter reason I avoided using something like `before` or `after` elements to make it more formal. Especially since the `position` attribute used currently to set the edge makes some assumptions that I wanted to make more official in a future update. for instance, I'd like to add `track-end` as well, but was unable to style where it should "start" without modifying more of the code.

chrisdholt: @bheston with our update to support default slotted content, couldn't position be set on the default slotted content (which seems like it could be this)?

bheston: It looks like the default slot is for `slider-labels`. Also, I had to bind the `position` in the template, which I'm not sure is possible through styling or without extending the `Slider` code. Even if it weren't used for labels, it seems that adding the position to the default slot would restrict its use.

chrisdholt: @bheston, sorry to be clear, the `track-start` slot could host the DIV in `fast-components` itself and bind to the data (it supports ViewTemplate's). An example of this is in Horizontal Scroll. I *think* that would keep this fairly non-breaking for anyone inheriting Foundation. By "default", I meant having content inside a slot by default.

chrisdholt: @bheston, I see that `track-start` is new - my bad. Let me look again

EisenbergEffect:
Ping.

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/slider/slider.template.ts
awentzel: @chrisdholt are we still holding on this PR?

awentzel:
@bheston what's the status of this PR?  It's very stale would love to merge or abandon, whichever is most appropriate.  Thanks

packages/web-components/fast-foundation/src/slider/slider.template.ts
chrisdholt: If we drop this template in Fluent and VSCode and neither breaks that's a good enough validation that this doesn't break implementations. If it renders incorrectly, yes we should hold for the visual refresh.

bheston: Here's FAST before and after:

image

Here's Fluent UI before and after:

image

It turns out the styles are already in place. Thanks copy and paste :)

VS Code didn't implement a slider, but now at least I'm setup ;)

Considering this is only additive I still see no reason it's not safe. We shouldn't be responsible for css selectors that assume too much about the underlying dom.

awentzel (approved):

packages/web-components/fast-components/src/slider/slider.styles.ts
```diff
@@ -88,6 +89,13 @@ export const sliderStyles: FoundationElementTemplate<ElementStyles, SliderOption
     .thumb-cursor:active {
         background: ${neutralForegroundRest};
     }
+    .track-start {
+        background: ${accentForegroundRest};
+        position: absolute;
+        height: 100%;
+        left: 0;
```
nicholasrice, line 96: Does the usage of the `left` property here need to change in RTL documents?

bheston: Yep, good catch. Resolved.

chrisdholt (approved):

PR #4993, opened 2021-07-26
docs(components): add/update and expand content

EisenbergEffect:
# Pull Request

## 📖 Description

This PR updates all component and integration docs (and a few others) to show the new component composition and design system APIs. Several missing components were added as well.

### General

## ⏭ Next Steps

A couple of follow up PRs include:

* Re-organizing the documentation table of contents based on feedback and to support content growth
* Adding a new "Apps and Experiences" section to the docs, which will contain docs on DI, Router, etc.


sites/website/src/docs/design-systems/fast-frame.md
```diff
@@ -11,29 +11,21 @@ FAST Frame ([`@microsoft/fast-components`](https://www.npmjs.com/package/@micros
...
 // ...
-import { fastButton, fastMenu } from "@microsoft/fast-components";
-
-// ...
-designSystem.register(
-    fastButton(),
-    fastMenu()
-);
-
+import {
```
nicholasrice, line 18: nit: indentation

packages/web-components/fast-foundation/src/skeleton/README.md
```diff
@@ -24,9 +37,27 @@ _Example:_
...
 ```
 
+### Pattern
+
+A URL for an image asset may be passed to the `pattern` attribute. In this mode, the `fast-skeleton` component is used as a container for a transparent SVG that may express a more complex placeholder
+
+```html
+<fast-skeleton
+    style="
+        --neutral-fill-rest: #e1dfdd;
```
nicholasrice, line 47: Do we need to add the `--neutral-fill-rest` definition here? We don't generally recommend setting CSS custom properties manually that are controlled by DesignToken.

EisenbergEffect: Not sure. If I recall, I took that directly from the fixtures.

packages/web-components/fast-foundation/src/horizontal-scroll/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+```ts
+import {
+    provideFASTDesignSystem,
+    fastHorizontalScroll
+} from "@microsoft/fast-components";
+
+provideFASTDesignSystem()
+    .register(
+        fastHorizontalScroll()
```
nicholasrice, line 22: Does this need `fastFlipper()` registered as well?

EisenbergEffect: Yes. Good catch. Update incoming.

packages/web-components/fast-foundation/src/skeleton/README.md
EisenbergEffect: I'm going to remove it. It's not present in explorer.

nicholasrice (approved):

robarbms (approved):

packages/web-components/fast-foundation/src/tabs/README.md
```diff
@@ -5,85 +5,105 @@ sidebar_label: tabs
...
-        <fast-tab-panel id="dessertsPanel">
-            <ol>
-                <li><fast-anchor href="#" appearance="hypertext">Tiramisu</fast-anchor></li>
-                <li><fast-anchor href="#" appearance="hypertext">Spumoni</fast-anchor></li>
-                <li><fast-anchor href="#" appearance="hypertext">Limoncello and Ice Cream with Biscotti</fast-anchor></li>
-            </ol>
-        </fast-tab-panel>
-    </fast-tabs>
-</fast-design-system-provider>
+<fast-tabs activeId="entrees">
```
janechu, line 31:
```suggestion
<fast-tabs activeid="entrees">
```

sites/website/sidebars.js
```diff
@@ -21,32 +21,35 @@ module.exports = {
...
                         "components/flipper",
+                        "components/horizontal-scroll",
                         "components/listbox",
                         "components/listbox-option",
                         "components/menu",
-                        "components/menu-item",
                         "components/number-field",
                         "components/progress",
                         "components/radio",
                         "components/radio-group",
```
janechu, line 44: In a similar vein to these other changes where we are consolidating some components that need to be explained in the context of a wrapping component, should radio button exist inside radio-group?

sites/website/src/docs/community/writing-documentation.md
```diff
@@ -31,7 +31,7 @@ The required fields are:
 * **sidebar_label**: The label displayed in the table of contents.
 * **custom_edit_url**: A URL where the document source can be edited.
 
-Beneath the YAML metadata block, article's body is written with [GitHub-flavored Markdown syntax](https://github.github.com/gfm/). Since our documentation system will turn the `title` metadata into the article header, the content should begin  with a short introductory paragraph, followed by an h2 header for each section of the document. You may use h3 headers beneath h2 headers but avoid deeper sub-sections.
+Beneath the YAML metadata block, the article's body is written with [GitHub-flavored Markdown syntax](https://github.github.com/gfm/). Since our documentation system will turn the `title` metadata into the article header, the content should begin  with a short introductory paragraph, followed by an h2 header for each section of the document. You may use h3 headers beneath h2 headers but avoid deeper sub-sections.
```
janechu, line 34:
```suggestion
Beneath the YAML metadata block, the article's body is written with [GitHub-flavored Markdown syntax](https://github.github.com/gfm/). Since our documentation system will turn the `title` metadata into the article header, the content should begin with a short introductory paragraph, followed by an h2 header for each section of the document. You may use h3 headers beneath h2 headers but avoid deeper sub-sections.
```

sites/website/src/docs/design-systems/creating-a-component-library.md
```diff
@@ -139,17 +139,79 @@ Here we provide the base component name of `counter`, which will be combined wit
...
+
+```ts
+export function provideSpecialDesignSystem(element?: HTMLElement): DesignSystem {
+    return DesignSystem.getOrCreate(element).withPrefix("special");
+}
+```
+
+The underlying `DesignSystem.getOrCreate` API gets the design system directly attached to the provided element. If one does not exist, it creates one. By default, if no element is provided, the design system is created on the `document.body`. Using this function abstracts those details and provides a much improved set of ergonomics to those using your components, while also giving you the opportunity to bake in certain configuration yourself, such as defining the default element prefix.
+
+Now, consumers of your components can setup your design system in their application with the button component as follows:
```
janechu, line 210:
```suggestion
Now consumers of your components can setup your design system in their application with the button component as follows:
```

sites/website/src/docs/resources/glossary.md
```diff
@@ -5,7 +5,12 @@ sidebar_label: Glossary
 custom_edit_url: https://github.com/microsoft/fast/edit/master/sites/website/src/docs/resources/glossary.md
 ---
 #### Design System
-In FAST, the Design System is a collection of properties that informs visual rendering. It generally contains information about fonts, type-ramps, colors, etc.
+
+A Design System can generally be thought of as a collection of resources for interactive media that promotes brand alignment. While that definition is intentionally broad, in UI development, Design Systems generally manifest as component libraries surrounded by usage guidance and design principles.
```
janechu, line 9:
```suggestion
A Design System can generally be thought of as a collection of resources for interactive media that promotes brand alignment. While that definition is intentionally broad, in UI development Design Systems generally manifest as component libraries surrounded by usage guidance and design principles.
```

janechu (approved):

packages/web-components/fast-foundation/src/dialog/README.md
```diff
@@ -5,7 +5,25 @@ sidebar_label: dialog
 custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/src/dialog/README.md
 ---
 
-A web component implementation of a [dialog](https://w3c.github.io/aria-practices/#dialog_modal) which can either be modal or non-modal.
+As defined by the [W3C](https://w3c.github.io/aria-practices/#dialog_modal):
+
+> A dialog is a window overlaid on either the primary window or another dialog window. Windows under a modal dialog are inert. That is, users cannot interact with content outside an active dialog window. Inert content outside an active dialog is typically visually obscured or dimmed so it is difficult to discern, and in some implementations, attempts to interact with the inert content cause the dialog to close.
```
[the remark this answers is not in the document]
janechu: Ah, I see this is quoted from the link above, removing this comment.

packages/web-components/fast-foundation/src/checkbox/README.md
```diff
@@ -7,31 +7,74 @@ custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-comp
 
 An implementation of a [checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox) as a form-connected web-component.
```
chrisdholt, line 8: Should we call out that this is an ARIA implementation? There are different requirements, such as the `<label>` element not being applicable to ARIA checkboxes (and Radios)

packages/web-components/fast-foundation/src/dialog/README.md
```diff
@@ -5,7 +5,25 @@ sidebar_label: dialog
...
+>
+> Like non-modal dialogs, modal dialogs contain their tab sequence. That is, Tab and Shift + Tab do not move focus outside the dialog. However, unlike most non-modal dialogs, modal dialogs do not provide means for moving keyboard focus outside the dialog window without closing the dialog.
```
chrisdholt, line 12: One of the reasons that modal/non-modal was initially included was because the W3C definition trends strongly towards "modal" which without a careful reading could create some incorrect assumptions. I'm good with this change, but something to be aware of.

packages/web-components/fast-foundation/src/divider/README.md
```diff
@@ -7,25 +7,40 @@ custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-comp
 
 A web component implementation of a [horizontal rule](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr).
```
chrisdholt, line 8: No change from you here, but while we're updating should add some context in terms of API? For example, it's technically an hr but we also provide an API to set a role of presentation which effectively removes those semantics altogether.

sites/website/sidebars.js
EisenbergEffect: Radios don't have to be inside a group so they are broken out. But the menu item can really only be inside of a menu. So, those are combined into the same doc. It was really strange having it on its own I think.

packages/web-components/fast-foundation/src/horizontal-scroll/README.md
```diff
@@ -0,0 +1,125 @@
+---
+id: horizontal-scroll
+title: fast-horizontal-scroll
+sidebar_label: horizontal-scroll
+custom_edit_url: https://github.com/microsoft/fast/edit/master/packages/web-components/fast-foundation/src/horizontal-scroll/README.md
+---
+
+An implementation of a content scroller as a web-component.
```
chrisdholt, line 8: I'm wondering if we should add docs here with regard to what to expect in terms of behavior. For instance, the reason that flippers are typically hidden by default is that they are duplicate content when items within the scroll region contain interactive content. If there's only visual or presentational content within this, flippers should be exposed.

chrisdholt (approved):

packages/web-components/fast-foundation/src/button/README.md
```diff
@@ -5,35 +5,58 @@ sidebar_label: button
...
+    Button,
+    buttonTemplate as template,
+} from "@microsoft/fast-foundation";
+import { buttonStyles as styles } from "./my-button.styles";
 
-// Button
-@customElement({
-    name: "fast-button",
+export const myButton = Button.compose({
+    baseName: "button",
```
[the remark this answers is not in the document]
EisenbergEffect: We've got some docs on the basics, but there are differences in certain components that need to be documented, particularly when it comes to default content overrides. There may be other options in the future as well. Previously, we had the equivalent of this section for each component, so I just updated that to no longer be completely broken/incorrect information.

packages/web-components/fast-foundation/src/disclosure/README.md
```diff
@@ -5,36 +5,56 @@ sidebar_label: disclosure
...
-            in The CW's Arrowverse franchise, first introduced in the 2012 pilot episode
-            of the television series Arrow, and later starring in Legends of Tomorrow.
-        </div>
-    </fast-disclosure>
-</fast-design-system-provider>
+<fast-disclosure appearance="lightweight">
+    <span slot="start">👩🏻‍🦳</span>
+    <strong slot="title">Read about White Canary</strong>
+    <div>
+        Sara Lance, also known by her alter-ego White Canary, is a fictional character
```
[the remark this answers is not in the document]
EisenbergEffect: Probably. I think this was a community contributed example. I could switch it to be about FAST.

sites/website/sidebars.js
chrisdholt: @EisenbergEffect with ARIA radios this is actually not the case - the radios definitely do need to be a child of an element with a role of radiogroup. Radio's which use the native input w/ type="radio" can be linked together with special magic, but from an ARIA standpoint that's not possible or valid; these are only tied together under the same parent. There is ongoing conversations in Open UI around this, but it would require changes from the ARIA working group and likely whatwg.

packages/web-components/fast-foundation/src/divider/README.md
EisenbergEffect: My preference is to get this merged and release the docs, since what we have published now is completely broken for our latest release. Then, I think we should establish an ideal component doc template. Something that includes things like properties, attributes, slots, events, examples, referenced design tokens, etc. Once we have the template, I think we should have the individual contributors who worked on each component go submit PRs to update the related docs. In this way we can have a consistent approach but also split up the work. I can tell you that these basic corrections took a lot of time to do across the whole set. I don't think any one person has the bandwidth to do all the ideal updates right now.

packages/web-components/fast-foundation/src/horizontal-scroll/README.md
EisenbergEffect: See comment above.

packages/web-components/fast-foundation/src/checkbox/README.md
EisenbergEffect: @chrisdholt Can you paste me the content that you want to put here instead? Or submit an insert suggestion that I can easily merge? You know better than I what the wording should be.

sites/website/sidebars.js
EisenbergEffect: But they don't need to be the child of our radio-group component specifically, correct? I can update the radio example. This would also mean that the example that we've had on our site for many months now is incorrect...

EisenbergEffect: Improved the examples for radio and radio-group.

sites/website/src/docs/design-systems/fast-frame.md
EisenbergEffect: Fixed

packages/web-components/fast-foundation/src/accordion/README.md
```diff
@@ -5,53 +5,102 @@ sidebar_label: accordion
...
+    fastAccordion,
+    fastAccordionItem
+} from "@microsoft/fast-components";
+
+provideFASTDesignSystem()
+    .register(
+        fastAccordion(),
+        fastAccordionItem({
+            collapsedIcon: `...your collapsed icon...`,
+            expandedIcon: `...your expanded icon...`,
```
bashmish, line 44: good to see more examples like this 👍

PR #4998, opened 2021-07-27
Register design-system-provider

bheston:
# Pull Request

## 📖 Description

We kept the DSP for backwards compat, and encourage consumers to transition to setting design token properties directly, though it still is rather convenient, but confusing to use since it doesn't have a visible interface and is not registered by default.

This simple change registers the dsp by default so any experiences using it don't have to register it separately.

### General

### Component-specific


chrisdholt:
It looks like this actually just adds it where it was missing prior and includes it with `allComponents` for easier registration - I think that's fine. Folks will still need to register this in their application though, which I think is expected. It would be odd if this was somehow "different" and "special", especially since the component is technically deprecated.

chrisdholt (approved):

nicholasrice (approved):

EisenbergEffect (approved):

janechu (approved):

Issue #4930, opened 2021-07-01
feat: Publicly Expose `fixture<T>` in `@microsoft/fast-tooling`


RehanSaeed:
# 🙋 Feature Request

`@microsoft/fast-tooling` has [`fixture<T>`](https://github.com/microsoft/fast/blob/e85551b6f4c3c79f5c5d8500a0ed2b5988a8def6/packages/tooling/fast-tooling/src/web-components/fixture.ts#L110) which seems like a really useful helper function for tests. It's not currently publicly exposed, can that be changed?

## 🤔 Expected Behavior

Allow us to import [`fixture<T>`](https://github.com/microsoft/fast/blob/e85551b6f4c3c79f5c5d8500a0ed2b5988a8def6/packages/tooling/fast-tooling/src/web-components/fixture.ts#L110).

## 😯 Current Behavior

[`fixture<T>`](https://github.com/microsoft/fast/blob/e85551b6f4c3c79f5c5d8500a0ed2b5988a8def6/packages/tooling/fast-tooling/src/web-components/fixture.ts#L110) is not publicly exposed, so can't import it.

## 💁 Possible Solution

Add it to the `@microsoft/fast-tooling` package.

EisenbergEffect:
@RehanSaeed I've been thinking of creating a new `@micrsofot/fast-testing` package which could container this and other helpers present and future. What do you think of that idea?

There's some trick with circular dependencies. For example, the helpers leverage foundation but are also needed to test foundation. If put into another package, I'm not sure how that would work exactly.

RehanSaeed:
> @RehanSaeed I've been thinking of creating a new @microsoft/fast-testing package which could contain this and other helpers present and future. What do you think of that idea?

Seems like a good idea.

rajsite:
+1 we're trying to leverage it for testing our components as well.

While you're digging in there could the type for [Fixture](https://github.com/microsoft/fast/blob/928d73621c62064992189270a920b2c6d6d9e19e/packages/web-components/fast-foundation/src/test-utilities/fixture.ts#L85) be updated from:

```ts
export interface Fixture<TElement = HTMLElement> {
    // ...
    connect(): Promise<void>;
    disconnect(): Promise<void>;
}
```

to:

```ts
export interface Fixture<TElement = HTMLElement> {
    // ...
    connect: () => Promise<void>;
    disconnect: () => Promise<void>;
}
```

[...]

PR #5016, opened 2021-07-30
refactor(fixture): mild change to typings to enable typical code pattern

EisenbergEffect:
# Pull Request

## 📖 Description

At the request of the community, I've made a small type change to our `fixture` test helper. This should be a non-breaking change but it's technically a change to the API, so I've marked this as "minor" change.

### 🎫 Issues

* Related to #4930  but does not fully address the need yet. See Next Steps below.

## 👩‍💻 Reviewer Notes

This is a change to the shape of the interface only, primarily for the purpose of enabling type-safe destructuring of the `Fixture` interface without lint warnings.

## 📑 Test Plan

Existing tests and TS compilation should suffice in this case.

### General

## ⏭ Next Steps

To fully address the community need, we should make our testing utilities officially available, along with documentation on the site. There is a challenge of making these utilities available, related to where we put them. At present, they are part of the `fast-foundation` library. This is because they were originally used to write tests for foundation, and they are also dependent on foundation types. To make them available, it would be better to have a `fast-testing` package. However, this creates a circular dependency issue in our monorepo due to the fact that `fast-testing` would depend on `fast-foundation` and `fast-foundation` would depend on `fast-testing`. A solution would be to move *all* of our tests out of our packages into a dedicated test project within the monorepo. This would break the circular dependency, as only the test project would depend on the `fast-testing` package. There are other advantages to this, such as more simply running all tests for the monorepo and better optimizing the test build time. At present, I don't think this is possible, due to the presence of the tooling in our monorepo, which has a different test setup than the packages. However, with our plan to move the tooling out, this would then enable use to move forward. So, steps would be as follows:

1. Move the tooling into its own monorepo
2. Work to ensure that individual test package setup is exactly the same across all packages in the fast monorepo
3. Move all the tests to a central location, probably a `test` folder at the root of the monorepo, and making adjustments to have a single test setup with commands to run "sets" of tests
4. Create the `fast-testing` package and then refactor to base all centralized tests on that package

I'm open to other ideas, but I don't see a way to break the circular dependency between foundation and test without moving the tests. I do think this will yield other benefits as well, especially server build time. It would be a change in workflow, though not to something uncommon (I've followed this pattern in other projects as well.)

### DIFF
--- a/change/@microsoft-fast-foundation-61f51210-4c16-4b3f-8984-f6a68f6e54d8.json
+++ b/change/@microsoft-fast-foundation-61f51210-4c16-4b3f-8984-f6a68f6e54d8.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "refactor(fixture): mild change to typings to enable typical code pattern",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "roeisenb@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-foundation-61f51210-4c16-4b3f-8984-f6a68f6e54d8.json
+++ b/change/@microsoft-fast-foundation-61f51210-4c16-4b3f-8984-f6a68f6e54d8.json
@@ -1,7 +1,7 @@
 {
-  "type": "minor",
+  "type": "none",
   "comment": "refactor(fixture): mild change to typings to enable typical code pattern",
   "packageName": "@microsoft/fast-foundation",
   "email": "roeisenb@microsoft.com",
-  "dependentChangeType": "patch"
+  "dependentChangeType": "none"
 }
--- a/packages/web-components/fast-foundation/src/test-utilities/fixture.ts
+++ b/packages/web-components/fast-foundation/src/test-utilities/fixture.ts
@@ -82,7 +82,7 @@ export interface Fixture<TElement = HTMLElement> {
      * Yields control to the caller one Microtask later, in order to
      * ensure that the DOM has settled.
      */
-    connect(): Promise<void>;
+    connect: () => Promise<void>;
 
     /**
      * Removes the {@link Fixture.parent} from the DOM, causing the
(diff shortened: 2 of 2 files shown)


change/@microsoft-fast-foundation-61f51210-4c16-4b3f-8984-f6a68f6e54d8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
chrisdholt, line 2: @EisenbergEffect I don't think this is exported from the package yet, so this would technically be a "none" change as it doesn't affect the distributed packages.

With that said, we do have a *request* to export this from our packages for use which would support the `minor` version bump denoted here. Do we want to do that or should we change this (and then the `dependentChangeType`) to none?

chrisdholt (approved):

change/@microsoft-fast-foundation-61f51210-4c16-4b3f-8984-f6a68f6e54d8.json
EisenbergEffect: I don't think we should do that. We shouldn't really have a testing-specific function as part of a runtime library. We need to extract that to its own thing, which has implications (see Next Steps).

I can change it to "none" though.

awentzel:
From a pipeline perspective, this could help with improve build times across all repositories.  I'm curious, architecturally how this would work.  Are we talking about unit testing, integration testings, etc?  How would that be separated from the packages they are testing?  I feel like I'm missing something.  Conceptually, it sounds great.

EisenbergEffect:
@awentzel This would move all tests, both unit and integration, into a single place. For reference, this is what we did for Aurelia 2. In the packages folder we have a `__tests__` folder where all test are located, broken down by package and then also a separate place where the integration tests are: https://github.com/aurelia/aurelia/tree/master/packages/__tests__

robarbms (approved):

PR #5019, opened 2021-07-30
fix: fix image load bug in fast-tooling html renderer navigation component

williamw2:
# Pull Request

## 📖 Description
Delay in the loading of images was causing a race condition that could result in an element containing an image being highlighted before the image was loaded and the final size of the element was set. This would result in the navigation layer highlight being in the wrong position or size.
The fix is to add a ResizeObserver to the body to trigger a redraw of the highlight whenever the size of the content changes.
*Note: The ResizeObserver is not supported in the FAST package's current Typescript version. This required the use of a temporary stopgap of the resize-observer files (copied from /packages/web-components/fast-foundation/anchored-region). Typescript has added support for the vanilla ResizeObserver and once the packages are upgraded these files can be removed. 

### General

### Component-specific


packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
```diff
@@ -35,9 +48,18 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
...
+    private resizeDetector: ResizeObserverClassDefinition | null = null;
+
     connectedCallback() {
         super.connectedCallback();
 
+        this.resizeDetector = new ((window as unknown) as WindowWithResizeObserver).ResizeObserver(
+            this.handleWindowChange
+        );
+
+        this.resizeDetector.observe(document.body);
```
robarbms, line 61: This seems like it will be called ALOT during page loading. We might want to think about debouncing handleWindowChange().

robarbms (changes requested):

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/resize-observer.ts
```diff
@@ -0,0 +1,17 @@
...
+    callback: ResizeObserverCallback
+) => ResizeObserverClassDefinition;
+
+export declare class ResizeObserverClassDefinition {
+    constructor(callback: ResizeObserverCallback);
+    public observe(target: Element): void;
+    public unobserve(target: Element): void;
+    public disconnect(): void;
+}
+
```
robarbms, line 13: All of these types and interfaces exist in ..anchored-region/resize-observer.ts, you should import those rather than duplicate the code. As an aside, we should probably move resize-observer somewhere more globally, like ../resize-observer/index.ts or into a definition file.

janechu: I had advised duplicating since we are moving this code to a new repository in a few weeks, even though we could reference it it is located in anchored-region and we would have to specifically drill down into the package when pathing to reference. We will be moving this code and upgrading our TypeScript dependency then deleting these files so I think the duplication, due to it being temporary, is harmless for now.

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
williamw2: There is a debouncer. Line 85 starts the setTimeout. Also the update logic only happens when an element is actively highlighted or selected which does not occur until until after the markup is rendered.

janechu (approved):

robarbms (approved):

PR #5024, opened 2021-08-03
fix menu default slot assignment

### LINKED ISSUE #4795: FAST Menu cannot be used in named slots.
# 🐛 Bug Report

Currently, `Menu` cannot be slotted into other components using named slots.
Internally `Menu` checks to see if it is a submenu and either sets the slot attribute to `submenu` or `undefined`, thus removing any user-defined value for the slot attribute.

## 💻 Repro or Code Sample

https://stackblitz.com/edit/fast-design-qqi4ue?file=index.html

## 🤔 Expected Behavior

Menu should retain user-defined values for the `slot` attribute.

## 😯 Current Behavior

Menu overrides the user value to either `submenu` if it is a descendent to another menu, or `undefined`

## 💁 Possible Solution

Possible fix could be to first check if `slot` is already set and to then use that instead of setting to undefined.
However, this could potentially create some conflicts with the submenu behavior and user expectations in cases where developers are creating custom nested menus and are trying to set the `slot` to something other than `submenu`.

## 🔦 Context

Creating a generic popup-like component with multiple named slots. A workaround is currently achieved by using a default slot when using `Menu`, but this is not ideal.

## 🌍 Your Environment

Windows 10
Edge Canary(latest)
FAST-Element(latest)
FAST-Foundation(latest)
FAST-Components(latest)

chrisdholt:
# Pull Request

## 📖 Description
The menu previously forced either a resolution for submenu's or undefined which prevented alternative slotting of the menu, which definitely should be supported. The initial approach I considered was simply falling back to the slot value itself; however, when I really considered the right behavior, I felt it was more important to honor the slot attribute above all else, and have the "special magic" of automatic nested submenu support be the secondary behavior.

### 🎫 Issues
Closes #4795 

### General

### Component-specific


EisenbergEffect (approved):
It's important to realize that this will only be evaluated once because `slot` is not observable by the templating engine. So, if someone adds a slot later on, it will not update.

chrisdholt:
> It's important to realize that this will only be evaluated once because `slot` is not observable by the templating engine. So, if someone adds a slot later on, it will not update.

Ohhhh, this is good feedback. Perhaps I’ll document this and spin off a follow-up issue to investigate that. Thanks @EisenbergEffect

nicholasrice (approved):

PR #5026, opened 2021-08-03
DRAFT: flyout config

scomea:
## 📖 Description

I'm looking at ways to "package up" different flyout configurations into ```FlyoutConfig``` objects so we have a standardized way of describing the behavior of a flyout relative to its anchor.  This would allow for some reusability as a number of components with flyouts will want to have similar flyout behaviors (ie. "above", "below", "widest area", "tallest area", etc...), as well as present developers with simple choices to get the behavior they want.

Additionally, our current "flyout tech" is the anchored region component which has a particular set of properties that govern its behavior which are encapsulated in a proposed ```AnchoredRegionConfig```, but at some point we expect to transition to native technologies (html/css) to position these elements.  The idea is that each ```FlyoutConfig``` would wrap a corresponding ```AnchoredRegionConfig```, and that when we get a new positioning technology we can add a new type of configuration object for that technology to the flyout configurations.  This could give us some flexibility during that transition.

This is my first crack at the idea, looking for feedback.

### DIFF
--- a/packages/web-components/fast-foundation/src/anchored-region/anchored-region-config.ts
+++ b/packages/web-components/fast-foundation/src/anchored-region/anchored-region-config.ts
@@ -80,10 +80,10 @@ export interface AnchoredRegionConfig {
 }
 
 /**
- * A flyout that always places itself above the anchor, has
+ * A region that always places itself above the anchor, has
  * a width to match the anchor, and is sized vertically by content
  */
-export const Above: AnchoredRegionConfig = {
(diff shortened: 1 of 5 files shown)


scomea:
This doc discusses the proposed css positioning scheme that could be the future alternative to anchored region:
https://github.com/MicrosoftEdge/MSEdgeExplainers/blob/main/CSSAnchoredPositioning/explainer.md

PR #5035, opened 2021-08-04
fix: security-update-bump tar to 4.4.15 or later

Ibrahimmaga:
# Pull Request

## 📖 Description
GHSA-3jfq-g458-7qm9
high severity
Vulnerable versions: >= 4.0.0, < 4.4.14
Patched version: 4.4.14
Impact
Arbitrary File Creation, Arbitrary File Overwrite, Arbitrary Code Execution

node-tar aims to prevent extraction of absolute file paths by turning absolute paths into relative paths when the preservePaths flag is not set to true. This is achieved by stripping the absolute path root from any absolute file paths contained in a tar file. For example /home/user/.bashrc would turn into home/user/.bashrc.

This logic was insufficient when file paths contained repeated path roots such as ////home/user/.bashrc. node-tar would only strip a single path root from such paths. When given an absolute file path with repeating path roots, the resulting path (e.g. ///home/user/.bashrc) would still resolve to an absolute path, thus allowing arbitrary file creation and overwrite.

Patches
3.2.2 || 4.4.14 || 5.0.6 || 6.1.1

The tar version have been bump to 6.0.0 or later. However we are still keeping tar version 4.4.14 because @evocateur/pacote@^9.6.3 depends on it. 

### 🎫 Issues
Users may work around this vulnerability without upgrading by creating a custom entry method that sanitizes the entry. path or a filter method that removes entries with absolute paths. But, are encouraged to upgrade to the latest patch versions, rather than attempt to sanitize tar input themselves.

### General

### Component-specific

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2565,7 +2565,7 @@
   dependencies:
     fs-extra "^8.1.0"
     ssri "^6.0.1"
-    tar "^4.4.8"
+    tar "^6.1.0"
 
 "@lerna/github-client@3.22.0":
   version "3.22.0"
@@ -2756,7 +2756,7 @@
     figgy-pudding "^3.5.1"
     npm-packlist "^1.4.4"
     npmlog "^4.1.2"
-    tar "^4.4.10"
+    tar "^6.1.0"
     temp-write "^3.4.0"
 
 "@lerna/package-graph@3.18.5":
@@ -15692,7 +15692,7 @@ node-gyp@^5.0.2:
     request "^2.88.0"
     rimraf "^2.6.3"
     semver "^5.7.1"
-    tar "^4.4.12"
+    tar "^6.1.0"
     which "^1.3.1"
 
 node-gyp@^7.1.0:


janechu (approved):

yarn.lock
```diff
@@ -2565,7 +2565,7 @@
   dependencies:
     fs-extra "^8.1.0"
     ssri "^6.0.1"
-    tar "^4.4.8"
+    tar "^6.1.0"
```
radium-v, line 2568: These entries will be reverted by Yarn.

radium-v (changes requested):
~This should be handled by adding an entry to the `resolutions` list in the root-level `package.json`, then running `yarn --ignore-scripts`.~

EDIT: Since the version has to be above `4.4.15`, you can delete the just entry for `tar@^4.4.10, tar@^4.4.12, tar@^4.4.8:` and regenerate the lockfile, which will replace the entry with the latest version for `4.4.15`.

radium-v:
This should also be addressed for the `tar@^6.0.2, tar@^6.1.0:` entry.

yarn.lock
Ibrahimmaga: the reason tar "^4.4.8" is left there is we are using  @evocateur/pacote which is using tar version 4.4.8, until they change their tar version, Yarn will always keep tar 4.4.8.

PR #5037, opened 2021-08-05
fix: update html render navigation to use custom title if present

williamw2:
# Pull Request

## 📖 Description
Updating the HTMLRender Navigation component to use the dataSetName in the pill for the selected element if it exists using the schemaDictionary title as the fallback.

### General

### Component-specific


packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
```diff
@@ -96,12 +97,19 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
         return new OverlayPosition(pos.top, pos.left, pos.width, pos.height);
     }
 
+    private getTitleForDictionaryId(dataDictionaryId): string {
```
nicholasrice, line 100: Looks like the return type should be `string | null`

williamw2: Would be nice if the compiler would catch stuff like that. Fixed.

janechu (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
```diff
@@ -96,12 +97,19 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
         return new OverlayPosition(pos.top, pos.left, pos.width, pos.height);
     }
 
+    private getTitleForDictionaryId(dataDictionaryId): string | null {
+        if (this.schemaDictionary && this.dataDictionary) {
+            if (this.dataDictionary[0][dataDictionaryId].data[dataSetName])
+                return this.dataDictionary[0][dataDictionaryId].data[dataSetName];
+            else
```
radium-v, line 104: Missing braces around the else statement

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
```diff
@@ -96,12 +97,19 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
         return new OverlayPosition(pos.top, pos.left, pos.width, pos.height);
     }
 
+    private getTitleForDictionaryId(dataDictionaryId): string | null {
```
radium-v, line 100: This method could be cleaned up with [nullish coalescing operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) and [optional chaining operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

radium-v (changes requested):

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
williamw2: Done

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
williamw2: No longer relevant.

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
```diff
@@ -96,12 +97,17 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
...
+        const dataDictionaryEntry = this.dataDictionary?.[0][dataDictionaryId];
+        return (
+            dataDictionaryEntry.data[dataSetName] ??
+            this.schemaDictionary?.[dataDictionaryEntry.schemaId].title ??
+            null
+        );
```
radium-v, line 106:
```suggestion
        return (
            this.schemaDictionary?.[this.dataDictionary?.[0][dataDictionaryId].schemaId]
                .title ?? null
        );
```

williamw2: This suggestion breaks the functionality. The function needs to return the dataDictionary dataSetName if it exists otherwise the schema title if it exists otherwise null.

radium-v (approved):

PR #5057, opened 2021-08-13
feat: detect if user have set preventDefault on fast-button

iahu:
# Pull Request
set timeout to detect if user have set `preventDefault` for fast-button, so that we can cancel the next actions, which just like the native button can prevent event (say as submit a form) at anytime.

## 📖 Description
I notice that, the native button can prevent default (like submit a form) anytime, but fast-button can't.

So I add a  setTimeout to detect if user have set `preventDefault` for fast-button, so that we can cancel the next actions, which just like the native button can prevent event (say as submit a form) at anytime.

### 🎫 Issues
addEventListener `click` after `DOMContentLoaded` and set `e.preventDefault()`

```html
<form action="#">
  <input type="text" name="text" />
  <fast-button type="submit" id="btn">submit</fast-button>
</form>
```

```js
document.querySelector('#btn').addEventListener('click', (e) => {
  e.preventDefault()
})
```
click the submit button will submit form as well.

## 👩‍💻 Reviewer Notes
I think html spec should provide a better way to detect `preventDefault` on custom element define lifetime. but now `setTimeout` look like is the best choice.

### DIFF
--- a/packages/web-components/fast-foundation/src/button/button.ts
+++ b/packages/web-components/fast-foundation/src/button/button.ts
@@ -150,7 +150,7 @@ export class Button extends FormAssociatedButton {
     /**
      * Submits the parent form
      */
-    private handleSubmission = () => {
+    private handleSubmission = (e: Event) => {
         if (!this.form) {
             return;
         }
@@ -161,22 +161,28 @@ export class Button extends FormAssociatedButton {
             this.attachProxy();
         }
 
-        // Browser support for requestSubmit is not comprehensive
-        // so click the proxy if it isn't supported
-        typeof this.form.requestSubmit === "function"
-            ? this.form.requestSubmit(this.proxy)
-            : this.proxy.click();
+        setTimeout(() => {
+            if (e.defaultPrevented) return;
 
-        if (!attached) {
-            this.detachProxy();
-        }
+            // Browser support for requestSubmit is not comprehensive
+            // so click the proxy if it isn't supported
+            typeof this.form?.requestSubmit === "function"
+                ? this.form.requestSubmit(this.proxy)
+                : this.proxy.click();
+
+            if (!attached) {
+                this.detachProxy();
+            }
+        });
     };
 
     /**
      * Resets the parent form
      */
-    private handleFormReset = () => {
-        this.form?.reset();
+    private handleFormReset = (e: Event) => {
+        setTimeout(() => {
+            !e.defaultPrevented && this.form?.reset();
+        });
     };
 
     public control: HTMLButtonElement;
--- a/packages/web-components/fast-foundation/src/button/button.ts
+++ b/packages/web-components/fast-foundation/src/button/button.ts
@@ -1,4 +1,4 @@
-import { attr, observable } from "@microsoft/fast-element";
+import { attr, observable, DOM } from "@microsoft/fast-element";
 import { ARIAGlobalStatesAndProperties, StartEnd } from "../patterns/index";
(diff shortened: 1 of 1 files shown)


scomea:
Not sure if I exactly understand intent here, but I think a better approach would be for the button check if the click event has been handled (defaultPrevented) and not act if it has, and authors who wish to act on the click event before form submission could set an event listener with "usecapture= true"?  https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

nicholasrice:
@scomea unless I'm misunderstanding you, I think that will still be too early. useCapture will catch the event on the way down, while authors will generally preventDefault() in the bubble phase (after capture phase). 

@iahu thanks for filing the PR, this is definitely a bug. I'm not super comfortable using a timeout for this purpose though because it pushes handling to a new task which I think is behaviorally different than the native button. I think what we should *really* do is catch the event on the document root and only request submission if default hasn't been prevented. We might have to get a little crafty though, cause I don't think we want the listener to fire for *every* click event in the document if we can avoid it.

scomea:
> @scomea unless I'm misunderstanding you, I think that will still be too early. useCapture will catch the event on the way down, while authors will generally preventDefault() in the bubble phase (after capture phase).

But isn't that the point of capture phase?  To allow authors to handle an event before bubbling?

nicholasrice:
It is, but I don't think most authors would expect to have to use capture in this case. I'm hesitant to make that the solution here because it doesn't align with the native button element's submission behavior. I think I would expect, as an author, to be able to call .preventDefault() during either phase and have it successfully cancel the form submission.

scomea:
Additionally - at some point an author's form mechanism may be custom enough to justify not using a fast button in "submit" or "reset" mode and just use it as a regular button to call their own form submit logic.

scomea:
> I think I would expect, as an author, to be able to call .preventDefault() during either phase and have it successfully cancel the form submission.

I think `DOM.queueUpdate` would be preferable to a timeout if we need to go that route.

scomea:
Would it work to add an event listener on the document when the button first catches the event and catch that same event when it bubbles up?

nicholasrice:
@scomea yea that is what I was playing with earlier today, the only issue then is we need a way to catch .stopPropagation() and .stopImmediatPropagation(). It's a little wonky but we can assign those methods to custom functions that handle tear down and then invoke call the Event's native method

iahu:
OK, I can update this PR to follow your discuss result.

iahu:
> @scomea yea that is what I was playing with earlier today, the only issue then is we need a way to catch .stopPropagation() and .stopImmediatPropagation(). It's a little wonky but we can assign those methods to custom functions that handle tear down and then invoke the Event's native method

`stopPropagation` and `stopImmediatPropagation` just work as except. no need to spy the native implementation,

and, I notice `DOM. queueUpdate` use `requestAnimationFrame` which just like `setTimeout`, but I have update it.

scomea:
Here's a thought, what if we put our click event listeners for submit/reset buttons on the form instead of the button?  That would give authors opportunity to handle the event as it bubbled. Would that match the native behavior closely enough?

iahu:
> Here's a thought, what if we put our click event listeners for submit/reset buttons on the form instead of the button? That would give authors opportunity to handle the event as it bubbled. Would that match the native behavior closely enough?

It seem not works, if authors set `stopPropagation()` on `fast-button` click event.

EisenbergEffect:
Closing this with no author response in almost 8 months.

PR #5074, opened 2021-08-16
fix: make resize observer target configurable in html render navigation component

### LINKED ISSUE #5063: Fix uncached image de-sync of the navigation outline
# Description

The outline created by the active item in the canvas area of the Creator should account for image loading (pre-cache).

## Requirements

- Check for images loading in so that the active navigation outline is in the correct coordinates
- Ensure that this is happening before the image has been cached by the browser

williamw2:
# Pull Request

## 📖 Description

Defaulting to the body resulted in scenarios that did not necessarily trigger the resize observer. This PR adds a "resizeobserverselector" attribute to the HTMLRenderNavigation component so that the resize observer target can be specified by the implementing code to ensure that the observer is attached to an element that will resize as content changes.
The new attribute is optional and the component will default to "document.body" if it is not supplied or if the selector specified does not match anything in the document.
This attribute is then used in Creator to specify the containing element which resizes to fit its children and should always trigger the observer whenever the content changes.

### 🎫 Issues
fixes #5063 

### General

### Component-specific

### DIFF
--- a/packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
+++ b/packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
@@ -71,7 +71,9 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
 
         this.resizeDetector.observe(
             (this.resizeobserverselector
-                ? document.querySelector(this.resizeobserverselector)
+                ? (this.getRootNode() as Element).querySelector(
+                      this.resizeobserverselector
+                  )
                 : null) ?? document.body
         );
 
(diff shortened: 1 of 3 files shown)


janechu (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
```diff
@@ -59,7 +69,11 @@ export class HTMLRenderLayerNavigation extends HTMLRenderLayer {
             this.handleWindowChange
         );
 
-        this.resizeDetector.observe(document.body);
+        this.resizeDetector.observe(
+            (this.resizeobserverselector
+                ? document.querySelector(this.resizeobserverselector)
```
EisenbergEffect, line 74: Just to be safe, rather than referencing `document` here, you may want to call `this.getRootNode()`. That will get you the document if the component is in the light dom but if it happens to be used inside of another shadow dom, it will get you the shadowRoot. You should be able to perform the querySelector on either (though you might have to cast in TypeScript).

EisenbergEffect (approved):
Looks good. I made a small recommendation to make the component usable in more contexts.

packages/tooling/fast-tooling/src/web-components/html-render-layer-navigation/html-render-layer-navigation.ts
williamw2: Done.

Issue #4976, opened 2021-07-21
feat: add ability to configure defaultElement for DesignToken


tlmii:
# 🙋 Feature Request

I would like to be able to specify the root/default element for my FAST application, so that all Design Tokens I set default behaviors for will only impact descendants of that element. 

## 🤔 Expected Behavior

I should be able to - somehow - specify an element on the page that will act as the root of my design system. That element would have all of the CSS custom properties applied to it for the descendants to pick up.

## 😯 Current Behavior

The body element is currently hard-coded as the default element for Design Tokens:

https://github.com/microsoft/fast/blob/2382e1c080fc27819e914d28e4e3be31ac35d595/packages/web-components/fast-foundation/src/design-token/design-token.ts#L20

This means that all of the CSS custom properties are added to the body element

## 💁 Possible Solution

@nicholasrice [mentioned a possible API for this](https://discord.com/channels/449251231706251264/482296787021856768/866744415610208296):

`DesignToken.configure({ defaultElement: HTMLElement })` or something more specific like `DesignToken.setDefaultTarget(element: HTMLElement)`

though @EisenbergEffect mentioned concerns about timing and subsequent discussion indicated a need to potentially modify the timing of when `withDefault` actually took effect (in essence deferring it until we were more confident that a method like the above `configure` or `setDefaultTarget` would have been called).

## 🔦 Context

Our scenario involves running a FAST mini application that is injected into a running page. We don't have knowledge of what that running page might have on it, and it could be another FAST application. That being said, being able to isolate the impact of our running code to just the specified root element and its descendants would be a cleaner solution even if the page _isn't_ running FAST.

EisenbergEffect:
I've got a few ideas around this. They center around changing how defaults are handled for tokens and also in connecting token initialization to our `DesginSystem` object.

First, I think we would make some changes to `DesignToken#withDefault`. I think we should probably store the default value on its own and then only apply it to *design token roots* (a new concept). So, we would have an implementation that looks something like this:

```ts
public withDefault(value: DesignTokenValue<T> | DesignToken<T>) {
  this.defaultValue = value;
  tokensWithDefaults.push(this); //todo: guard against multiple adds of this token
  
  for (const root of tokenRoots) {
    DesignTokenNode.for(this, root).set(value);
  }
      
  return this;
}
```

[...]

PR #5077, opened 2021-08-16
fix: set set baseLayerLuminance according to fillColor

### LINKED ISSUE #5060: fix: menu component not adapting colors properly
# 🐛 Bug Report

The Menu, Menu Item, and Listbox (Listbox Element, Select, and Combobox) don't properly adapt their colors when `fill-color` changes.

## 💻 Repro or Code Sample

In `fast-components`:
1. Modify the file `.storybook/preview.js` to set the value for the `fillColor`:
    ```js
    import * as FAST from "../src/index-rollup";
    import { parseColor } from "@microsoft/fast-colors";
    FAST.fillColor.setValueFor(document.body, FAST.SwatchRGB.from(parseColor("#FFFFFF")));
    // ...
    ```
2. Run `yarn start` to start Storybook.
3. Navigate to the Menu story in the sidebar.

## 🤔 Expected Behavior

The Menu should be rendered with a light color scheme.

## 😯 Current Behavior

Menu:
Screenshot 2021-08-13 150341

Listbox:
Screenshot 2021-08-13 151435

Select:
image

Combobox:
image

## 🔦 Context

These components all use `neutralLayerFloating`, as well as other `neutral` values that don't update when in light mode.

## 🌍 Your Environment

* OS & Device: Windows 10 PC
* Browser Microsoft Edge
* Version 92.0.902.73

radium-v:
# Pull Request

## 📖 Description

Sets the default value for `baseLayerLuminance` to use the `fillColor`, to determine if light or dark mode should be set.

### 🎫 Issues

Fixes #5060.

## 👩‍💻 Reviewer Notes

I'm not 100% familiar with the new design system, so this may not be the most solid approach to setting the default `baseLayerLuminance`.

## 📑 Test Plan

To confirm this work, the line in `.storybook/preview.js` can be changed to any color value:

```js
const backgroundColor = "#FFF";
```

### General

### Component-specific

## ⏭ Next Steps

Tests should be added so that the light/dark mode luminance is set properly in various situations.

An added bonus: once we update Storybook and add `@storybook/addons-essential`, we will be able to hook into SB's built-in functionality for changing the background color.

### DIFF
--- a/change/@microsoft-fast-components-292ddd91-67af-49c5-8911-fde8eca3e3ea.json
+++ b/change/@microsoft-fast-components-292ddd91-67af-49c5-8911-fde8eca3e3ea.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "set baseLayerLuminance according to fillColor",
+  "packageName": "@microsoft/fast-components",
+  "email": "john.kreitlow@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/web-components/fast-components/.storybook/preview-head.html
+++ b/packages/web-components/fast-components/.storybook/preview-head.html
@@ -5,7 +5,6 @@
         display: flex;
         flex-direction: column;
         height: 100%;
-        background-color: var(--fill-color);
     }
     #root {
         padding: 20px;
--- a/packages/web-components/fast-components/.storybook/preview.js
+++ b/packages/web-components/fast-components/.storybook/preview.js
@@ -1,5 +1,18 @@
 import "../src/index-rollup";
+import * as FAST from "../src/index-rollup";
+import { parseColor } from "@microsoft/fast-colors";
 
 export const parameters = {
     layout: "fullscreen",
 };
+
+const root = document.body;
+
+// By default, we grab the current background color to set the fillColor token.
+// This value can be changed to see the components in light mode.
+const backgroundColor = getComputedStyle(root).backgroundColor;
+
+FAST.fillColor.setValueFor(root, FAST.SwatchRGB.from(parseColor(backgroundColor)));
+
+root.style.backgroundColor = FAST.fillColor.createCSS();
+root.style.color = FAST.neutralForegroundRest.createCSS();
--- a/packages/web-components/fast-components/src/design-tokens.ts
+++ b/packages/web-components/fast-components/src/design-tokens.ts
@@ -1,29 +1,30 @@
 import { DesignToken } from "@microsoft/fast-foundation";
 import { Direction } from "@microsoft/fast-web-utilities";
 import { Palette, PaletteRGB } from "./color/palette";
-import { Swatch } from "./color/swatch";
+import { InteractiveSwatchSet } from "./color/recipe";
 import { accentFill as accentFillAlgorithm } from "./color/recipes/accent-fill";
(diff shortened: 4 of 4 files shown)


EisenbergEffect:
Definitely want @bheston to review this one.

bheston:
This is an interesting thought around reciprocal tokens, but it's not the intended sequence. The intended flow is `baseLayerLuminance` -> `neutralLayerX` -> `fillColor`, so setting luminance to light or dark cascades the colors all the way down. The intent being that the system is the driver, not specific color values.

I actually have a change coming in to fix this relationship I describe above.

bheston (changes requested):
This is an interesting thought around reciprocal tokens, but it's not the intended sequence. The intended flow is `baseLayerLuminance` -> `neutralLayerX` -> `fillColor`, so setting luminance to light or dark cascades the colors all the way down. The intent being that the system is the driver, not specific color values.

I actually have a change coming in to fix this relationship I describe above.

bheston:
Here is my aforementioned PR: https://github.com/microsoft/fast/pull/5091

The issue I think you may have been trying to solve here came up again today, which is when someone is using only a component or two from FAST and not configuring the entire design system or using `fillColor` or layer recipes. In that case, I think some interop like this is beneficial. I'm not sure if there's a way we can detect that or some convenience method, but say I put a `fast-button` onto my dark grey website, I probably want the button to take the luminance value from that container. I might also want my light blue site to pick up the blue as the neutral palette base.

Any thoughts on a way we might add that at component registration or detect that it's in some non-design-system container and bootstrap the design system a bit?

chrisdholt:
> Here is my aforementioned PR: #5091
> 
> The issue I think you may have been trying to solve here came up again today, which is when someone is using only a component or two from FAST and not configuring the entire design system or using `fillColor` or layer recipes. In that case, I think some interop like this is beneficial. I'm not sure if there's a way we can detect that or some convenience method, but say I put a `fast-button` onto my dark grey website, I probably want the button to take the luminance value from that container. I might also want my light blue site to pick up the blue as the neutral palette base.
> 
> Any thoughts on a way we might add that at component registration or detect that it's in some non-design-system container and bootstrap the design system a bit?

Question for clarity - can you elaborate on what you mean by "non-design-system container"? Do you mean the FAST DSP or something else? Every design system registers to some DOM node I believe, so there would always be some aspect of a container and I think if you're pulling in a single component you likely need to configure the design system accordingly. I think that's fairly straight forward as it is now, but we can certainly look at improved ergonomics for it. I think I just want to make sure we aren't expecting DSP's to be the go-to use case :)

bheston:
> can you elaborate on what you mean by "non-design-system container"

Yeah, exactly what you said. I didn't mean the DSP or needing a DSP for anything to work, I was trying to explain the case of a single FAST button somewhere within a DOM of other non-FAST components. The tokens all hang from somewhere though, currently the document body. This is fine even though nothing would consume them in the hierarchy until that button. This shouldn't be an issue as long as we're not stepping on other css variables, or perhaps the css variables need to be namespaced.

The ergonomics idea I'd like to improve is that notion of setting up the design system for a container that otherwise has no notion of the design system (i.e., legacy dom, not using tokens, etc.)

EisenbergEffect:
Dropping this here as its an issue that's related to the direction the conversation has turned: https://github.com/microsoft/fast/issues/4976

PR #4846, opened 2021-06-09
fix: increment/decrement number-field value with arrow keys

yeraycaballero:
# Pull Request

## 📖 Description
An HTML input type number allows to increment/decrement the value using arrow up/down keys. This commit adds that behaviour to number-field.

## 👩‍💻 Reviewer Notes

I added a keydown handler to the internal control, which calls stepUp or stepDown if the keycode is arrow-up or arrow-down. This commit includes a test, but it can tested manually in storybook or components explorer

### General

- [x] I have included a change request file using `$ yarn change`
- [x] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
- [x] I have read the [CONTRIBUTING](https://github.com/Microsoft/fast/blob/master/CONTRIBUTING.md) documentation and followed the [standards](https://www.fast.design/docs/community/code-of-conduct/#our-standards) for this project.

### Component-specific

- [ ] I have added a new component
- [x] I have modified an existing component
- [ ] I have updated the [definition file](https://github.com/Microsoft/fast/blob/master/packages/web-components/fast-components/CONTRIBUTING.md#definition)
- [ ] I have updated the [configuration file](https://github.com/Microsoft/fast/blob/master/packages/web-components/fast-components/CONTRIBUTING.md#configuration)

PR #5081, opened 2021-08-17
Add up/down support to number field

claviska:
# Pull Request

## 📖 Description

Adds support for pressing the up/down arrow keys to step the number field up/down. I'm considering it a "bug" rather than a feature since this brings the FAST component to parity with `<input type="number">`.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

- Open the number field demo
- Focus on any number field
- Press up/down
- Verify min/max/step are honored

### General

### DIFF
--- a/packages/web-components/fast-foundation/src/number-field/number-field.ts
+++ b/packages/web-components/fast-foundation/src/number-field/number-field.ts
@@ -261,6 +261,24 @@ export class NumberField extends FormAssociatedNumberField {
     public handleChange(): void {
         this.$emit("change");
     }
+
+    /**
+     * Handles the internal control's `keydown` event
+     * @internal
+     */
+    public handleKeyDown(event: KeyboardEvent): boolean {
+        if (event.key === "ArrowUp") {
+            this.stepUp();
+            return false;
+        }
+
+        if (event.key === "ArrowDown") {
+            this.stepDown();
+            return false;
+        }
+
+        return true;
+    }
 }
 
 /**
--- a/change/@microsoft-fast-foundation-6ed466c1-f107-4f38-9b7d-87527dbfdeab.json
+++ b/change/@microsoft-fast-foundation-6ed466c1-f107-4f38-9b7d-87527dbfdeab.json
@@ -1,6 +1,0 @@
-{
-  "type": "minor",
-  "packageName": "@microsoft/fast-foundation",
-  "email": "corylaviska@microsoft.com",
-  "dependentChangeType": "patch"
-}
--- a/change/@microsoft-fast-foundation-6ed466c1-f107-4f38-9b7d-87527dbfdeab.json
+++ b/change/@microsoft-fast-foundation-6ed466c1-f107-4f38-9b7d-87527dbfdeab.json
@@ -1,0 +1,6 @@
+{
+  "type": "minor",
+  "packageName": "@microsoft/fast-foundation",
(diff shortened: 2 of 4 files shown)


scomea (approved):

robarbms (approved):

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -261,6 +261,24 @@ export class NumberField extends FormAssociatedNumberField {
     public handleChange(): void {
         this.$emit("change");
     }
+
+    /**
+     * Handles the internal control's `keydown` event
+     * @internal
+     */
+    public handleKeyDown(event: KeyboardEvent): boolean {
+        if (event.key === "ArrowUp") {
```
robarbms, line 270: You should include the arrow key codes and check using those.

```
import { keyCodeArrowDown, keyCodeArrowUp } from "@microsoft/fast-web-utilities";
```

Then check the keycode:

```
if (event.keyCode === keyCodeArrowUp) {
    ...
}
```

robarbms (changes requested):

packages/web-components/fast-foundation/src/number-field/number-field.ts
claviska: As [`KeyboardEvent.keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), I opted to use a [widely supported syntax](https://caniuse.com/keyboardevent-key) to achieve the same thing without importing an additional utility.

May I ask why we're still using key codes now that the browsers bake more intuitive values into `KeyboardEvent.key`?

I'm happy to make the change if that's what it takes — just seems like a step backwards.

scomea: > As [`KeyboardEvent.keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), I opted to use a [widely supported syntax](https://caniuse.com/keyboardevent-key) to achieve the same thing without importing an additional utility.

This is correct, we should moving away from keyCode.

robarbms (approved):

packages/web-components/fast-foundation/src/number-field/number-field.ts
scomea, line 270: Actually, if you could import the key identifier strings from the shared resource like we do in radio group that would match the new pattern. 
https://github.com/microsoft/fast/blob/79f62ae08fd5db598272ca907ba03c4afa81a49f/packages/web-components/fast-foundation/src/radio-group/radio-group.ts#L2

yeraycaballero:
Hi  @claviska. I'm happy to see you here :). This pull request seems to be the same as one I did a while ago. You can check it here https://github.com/microsoft/fast/pull/4846

packages/web-components/fast-foundation/src/number-field/number-field.ts
claviska: Updated in b28918d.

chrisdholt: An issue to track alignment across the components: https://github.com/microsoft/fast/issues/5089 :)

chrisdholt (approved):

Issue #5069, opened 2021-08-14
Remove the accent color picker from the top bar of Creator


janechu:
# Description

The accent color is no longer a top level item for the default fluent web components and should not be top level next to the LTR/RTL control and the theme control.

## Requirements

- Confer with design that this should be the case
- Remove the control from the top bar

williamw2:
The idea to remove the control from the top bar came about because it did not change the appearance of Fluent components. However, after consulting with design this may be a bug in the Fluent library or with the Creator's implementation.  @chrisdholt is investigating. In the mean time we will leave this issue open until investigations are complete.

janechu:
Closing via #5086 as this enables the color picker for fluent web components and the design consensus is that it is a top level item, with theme and RTL switching.

PR #5086, opened 2021-08-18
Default to the fluent ui design system and tokens

janechu:
# Pull Request

## 📖 Description

This change switches the defaults of the design system and design tokens to use the Fluent UI defined schemas and design tokens.

### 🎫 Issues

Related to #5069 

## 👩‍💻 Reviewer Notes

You should now see the Fluent UI default accent color (which was the only difference currently) in the top bar as well as the design tokens tab.

### General

## ⏭ Next Steps

- Discussions on the effect of having multiple component libraries with their own design tokens, some of which may conflict, to follow.


EisenbergEffect:
@janechu does this mean that for the mean time the accent color picker won't work for the FAST components?

janechu:
> @janechu does this mean that for the mean time the accent color picker won't work for the FAST components?

It will, but it does mean at present we cannot make them independent of each other, but right now the tokens are aligned. I'm up for a longer discussion on how to handle this long term (name spacing?) but it seems like an issue that should be solved from a library POV.

EisenbergEffect (approved):

PR #5087, opened 2021-08-18
feat(fast-element): make the task queue resilient to task exceptions

EisenbergEffect:
# Pull Request

## 📖 Description

This PR attempts to make the DOM task queue more resilient by adding a try/catch block around each task execution. If an error is caught, it is queued and then scheduled to be thrown at a later time where it will not affect the task queue execution.

### 🎫 Issues

No official issue for this.

## 👩‍💻 Reviewer Notes

Just a small bit of code added to the queue system. The interesting bit is what's done with the error itself, since we don't want to swallow it but we can't allow it to be thrown during the course of draining the queue.

## 📑 Test Plan

I'm going to add some tests for this mechanism before converting this PR out of Draft. We should probably also run some performance tests before merging. I'm not sure whether extracting the task execution to its own function performs better or worse than leaving it inline. There may be some v8 optimization issues depending on how this is done. We should experiment with that. I'd also like to see what the effect is of having the try/catch vs. removing it, so we know if there's going to be a slow down no matter what.

### General

## ⏭ Next Steps

Before this PR is complete, we need to add unit tests and do some perf explorations.


EisenbergEffect:
@nicholasrice When you return to work, can you pick this up and use our new benchmarking infrastructure to test a few things out? See notes above.

PR #5091, opened 2021-08-18
Updated the default fill color to neutral layer 1

bheston:
# Pull Request

## 📖 Description

The `fillColor` design token was set to a fixed, dark position on the neutral palette. This ignored the `baseLayerLuminance` design token because it was not possible to only change that and have the design system respond. The ideal flow is that the luminance drives the layer colors and the layer colors are used as backgrounds for fixed sections of the app, even if there is only a single section (layer 1).

Also updated the Storybook wrapper to adjust the navigation text based on the updated fill color.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/design-tokens.ts
+++ b/packages/web-components/fast-components/src/design-tokens.ts
@@ -294,11 +294,144 @@ export const accentPalette = create<Palette>({
     cssCustomPropertyName: null,
 }).withDefault(PaletteRGB.create(accentBase));
 
+// Neutral Layer Card Container
 /** @public */
-export const fillColor = create<Swatch>("fill-color").withDefault(element => {
-    const palette = neutralPalette.getValueFor(element);
-    return palette.get(palette.swatches.length - 5);
+export const neutralLayerCardContainerRecipe = create<ColorRecipe>({
+    name: "neutral-layer-card-container-recipe",
+    cssCustomPropertyName: null,
+}).withDefault({
+    evaluate: (element: HTMLElement): Swatch =>
+        neutralLayerCardContainerAlgorithm(
+            neutralPalette.getValueFor(element),
+            baseLayerLuminance.getValueFor(element),
+            neutralFillLayerRestDelta.getValueFor(element)
+        ),
 });
+
+/** @public */
+export const neutralLayerCardContainer = create<Swatch>(
+    "neutral-layer-card-container"
+).withDefault((element: HTMLElement) =>
+    neutralLayerCardContainerRecipe.getValueFor(element).evaluate(element)
+);
+
+// Neutral Layer Floating
+/** @public */
+export const neutralLayerFloatingRecipe = create<ColorRecipe>({
+    name: "neutral-layer-floating-recipe",
+    cssCustomPropertyName: null,
+}).withDefault({
+    evaluate: (element: HTMLElement): Swatch =>
+        neutralLayerFloatingAlgorithm(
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -294,12 +294,145 @@ export const accentPalette = create<Palette>({
...
+/** @public */
+export const neutralLayer4 = create<Swatch>(
+    "neutral-layer-4"
+).withDefault((element: HTMLElement) =>
+    neutralLayer4Recipe.getValueFor(element).evaluate(element)
+);
+
+/** @public */
+export const fillColor = create<Swatch>("fill-color").withDefault(element =>
+    neutralLayer1.getValueFor(element)
```
EisenbergEffect, line 433: Should fill color be dependent on the idea of a *current layer*? Or is fill color always based on neutralLayer1?

EisenbergEffect (approved):

packages/web-components/fast-components/src/design-tokens.ts
bheston: It's not always layer 1. Under the current model it's up to someone to manually set the correct fill for the given element though. I think we structured it this way to make it easier to piece things together, for instance to build a nav component that's always "layer 2" and a toolbar that's "layer 3" and allow you to compose your app structurally however makes sense rather than force nesting and infer layer boundaries. I think that's what you were asking anyway.

So someone might do `fillColor.setValueFor(navContainer, neutralLayer2)` or the same within a component similar for how we do Card.

EisenbergEffect: I think what I was wondering is should we have `fillColor` be something like this?

```ts
export const fillColor = create<Swatch>("fill-color").withDefault(element => {
    const currentLayer = layer.getValueFor(element);
    return currentLayer.getValueFor(element);
});
```

This would introduce a token that you can set to indicate what layer a DOM tree is in, and then base fill off of that.

bheston: Ah, that's interesting. Would `layer` be the color recipe `neutralLayer1` or a mapping like `"1"`?

I want to get @nicholasrice thoughts on this as well, but it seems most beneficial to me if we have other tokens that might hang off `layer`, like perhaps that's how someone could apply different elevation values automatically (a "Floating" layer uses elevation 32, etc.)

Although, as I started writing that example, that felt very similar to the style-set model we've been considering in the modular styling approach. For instance, an "accent plate" might be a set consisting of "accent fill", "neutral border", and "foreground on accent" for content.

That seems like a great model though, keeping sets as tokens. I'm not sure the structure this should take right now, but I think it's a good consideration. I suggest we use the model I have in code for now since it works and it's less abstracted, but continue with thoughts on the evolution of compound or dependent styling.

EisenbergEffect: Rather than having it return a number, I was thinking it returned a token. It's a token that specifies which token to use. I have no idea if this is a good idea technically. Nick would definitely need to weigh in.

nicholasrice: I think whether we did this or not should depend on if we need to coordinate other tokens to the layer or not. If so, then I think creating a layer token to communicate the contextual layer makes a lot of sense, but otherwise it's probably an un-needed abstraction.

nicholasrice (approved):

Issue #5068, opened 2021-08-14
Ensure Monaco Editor actions "delete", "copy/paste" and "undo/redo" are reflected in other Message System services


janechu:
# Description

The actions "delete", "copy/paste" and "undo/redo" in the Monaco Editor do not appear to be reflected properly in the other Message System services.

## Requirements

- Ensure that deleting entire blocks of HTML is properly reflected in messages to the Message System (especially check the style custom elements such as flexbox are reflecting this change)
- Ensure that Copy/Paste actions are also reflected in messages properly to the Message System
- Ensure the Undo/Redo actions are also reflected in messages to the Message System

PR #5092, opened 2021-08-18
fix: resolves an issue where the CSS layout would not update when style was updated in the Monaco Editor

janechu:
# Pull Request

## 📖 Description

This makes the CSS layout controlled in the sense that the CSS string value is re-interpreted when it is passed down and applied without allowing a control to dictate it's own value.

### 🎫 Issues

Partially solves #5068

## 👩‍💻 Reviewer Notes

Build the project and run Creator, you should be able to see changes reflected in the CSS layout control when the style attribute on the current element is updated.

## 📑 Test Plan

Unit tests have been updated.

### General

## ⏭ Next Steps

- @williamw2 to update the color picker to also be controlled via the value attribute.


packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.ts
```diff
@@ -1,4 +1,5 @@
 import { observable } from "@microsoft/fast-element";
+import { keyCodeSpace } from "@microsoft/fast-web-utilities";
```
scomea, line 2: Nit: should move away from keyCodes.

packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.ts
```diff
@@ -139,6 +140,32 @@ export class CSSLayout extends FormAssociatedCSSLayout {
...
 
+    /**
+     * Handles the css-layout controls being enabled by pressing space
+     * @internal
+     */
+    public handleKeypressCSSChange = (
+        observablePropertyName: ObservableFlexboxCSSPropertyName,
+        e: KeyboardEvent
+    ): void => {
+        switch (e.keyCode) {
```
scomea, line 151: This will preventDefault() on all keypresses, consider returning true for those you aren't handling.

scomea (approved):

williamw2 (approved):

packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.ts
janechu: Yep, there's no utility to replace it yet though.

packages/tooling/fast-tooling/src/web-components/css-layout/css-layout.ts
janechu: That's fine, that's a desirable outcome since this is all self contained, we don't expect any other events to be emitted than a change event fired from the component.

PR #5120, opened 2021-08-29
feat(dialog): implement <dialog> interface

bennypowers:
# Pull Request

## 📖 Description
adds `close` and `cancel` events to dialog. See https://github.com/material-components/material-web/issues/1583

## 👩‍💻 Reviewer Notes
I was unable to get the repo up and running due to some mismatched ramda types (see discord chat)

### General

### Component-specific


packages/web-components/fast-foundation/src/dialog/dialog.ts
```diff
@@ -96,6 +96,8 @@ export class Dialog extends FoundationElement {
      */
     public dismiss(): void {
         this.$emit("dismiss");
```
chrisdholt, line 98: @bennypowers Thanks for this! @EisenbergEffect thoughts on how we might *deprecate* the `dismiss` event on line 98? If we can mark it deprecated somehow we can add a comment above to remove in the next major version.

bennypowers: sounds like a good use for custom-elements.json

EisenbergEffect: @bennypowers Does the custom-elements.json have a way to indicate that something is deprecated? If not, perhaps we just leave the event in there for now with a comment that it's deprecated, and then remove any documentation on this either on our site or the spec. Then, we can stop publishing the event in a future version.

EisenbergEffect: Looking more at the code, I'm inclined to deprecate the `dismiss` and `hide` methods, keeping their old event and state behaviors in place. Then add new `cancel()` and `close()` methods with the improved event patterns.

bennypowers: there are two options in custom-elements.json
1) use a non-standard field and consume it in your own tooling. I opened https://github.com/webcomponents/custom-elements-manifest/issues/84 to track the idea in the spec
2) Prepend `**DEPRECATED**` to description members

`HTMLDialogElement` has `close()`, `show()`, and `showModal()`

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement

awentzel: @bennypowers @EisenbergEffect @chrisdholt how are we feeling about this PR? Are there changes that still need to be made? As a stale Draft status it seems we should either remove Draft and work to get it updated to suit the comments or abandon?  Thoughts?

chrisdholt: I'd love to have this still come in, however I know that @bennypowers was running into build issues that we weren't able to repro. Definitely happy to find a way to help get beyond that if the issue still persists, but if I recall it was quite frustrating and led to this sitting in the ether.

bennypowers: Happy to take another crack at this

bennypowers: Maintainers should please permit CI to run

bennypowers: > @bennypowers Thanks for this! @EisenbergEffect thoughts on how we might _deprecate_ the `dismiss` event on line 98? If we can mark it deprecated somehow we can add a comment above to remove in the next major version.

Also, since this thread started, [`deprecated` got added to the custom-elements manifest schema](https://github.com/webcomponents/custom-elements-manifest/pull/89)

EisenbergEffect: @williamw2 Since you are investigating custom elements manifest, it would be great if we could support the new `deprecated` feature. It might be good to sync up with @nicholasrice on some of this as well. I think there's a possibility of using the API extractor metadata from our code to generate the custom elements manifest. There was recent work which Nick contributed to, to enable custom documentation tags. I *think* it's all merged and ready for use now. If we can use that, we can add custom tags for slots, deprecated, etc. and then use the API extractor metadata to generate the manifest from our code.

bennypowers:
Having rebased and taken a look at the state of the PR, I'm still mostly sure this is code-complete, but found I was unable to run the lint or test steps after an hour of fiddling. I leave that in maintainers hands, and am happy to make any tooling-related fixes necessary.

chrisdholt:
> Having rebased and taken a look at the state of the PR, I'm still mostly sure this is code-complete, but found I was unable to run the lint or test steps after an hour of fiddling. I leave that in maintainers hands, and am happy to make any tooling-related fixes necessary.

I'll pull this down today and see if we can't get things running locally. If all looks good and passes here, I don't see an issue. Thanks @bennypowers!

packages/web-components/fast-foundation/src/dialog/dialog.ts
nicholasrice: Yep, docs for that feature are here: https://api-extractor.com/pages/configs/tsdoc_json/

nicholasrice (approved):

EisenbergEffect:
@chrisdholt I'm assigning this to you so it doesn't get lost in the mix.

chrisdholt:
This is so odd, I'm not able to repro the failures locally. I'm going to check out this branch and push a PR to see if it still repro's...

packages/web-components/fast-foundation/src/dialog/dialog.spec.ts
```diff
@@ -262,6 +262,52 @@ describe("Dialog", () => {
...
+            expect(wasDismissed).to.equal(true);
+
+            await disconnect();
+        });
+
+        it("should fire a 'close' event when its button is clicked", async () => {
+            const { element, connect, disconnect } = await setup();
+
+            const button = document.createElement('button');
+            button.textContent = 'close';
```
chrisdholt, line 290:
```suggestion
            button.textContent = 'close';
            button.addEventListener("click", () => (element as unknown as Dialog).hide());
```

chrisdholt (commented):
@bennypowers I think the issue here with the tests may be that the button's click event in the "close" test doesn't do anything. So the test is failing accurately - `wasDismissed` is not being called. I'm not sure why that's causing everything else to fail as well, but it is. 

I think given the intent of when this event fires after looking at the Dialog class (when the `hide()` method is called), adding the suggestion on the above lines should fix the tests. If you want to push that change I think we'll see all green and be good to finally get this in. Thanks for bearing with us on this one :)

chrisdholt (approved):

chrisdholt:
🎉 Thanks @Bennypowers!

bennypowers:
So much green!

Thanks @chrisdholt for fixing this 🤗

PR #5140, opened 2021-09-01
support default slotted content start end

### LINKED ISSUE #3376: Add a way to default start and end templates in web components
I would like the ability to default the content of the start and end templates when using `endTemplate` and `startTemplate`.

This would allow us to continue to use the shared library in cases where we want to default content.

### LINKED ISSUE #5116: fix: Start/end slots cannot be populated with default value in FoundationElementDefinition
# 🐛 Bug Report

Unlike slots defined within given components, it does not appear to be possible to populate the [start/end slots](https://github.com/microsoft/fast/blob/e498f1e35232169d2bf80e366350d0cde812d76d/packages/web-components/fast-foundation/src/text-field/text-field.template.ts#L79) with default values when using `compose` to register a component.

## 💻 Repro or Code Sample

N/A - the behavior I'm expecting isn't supported. I'm filing this as a bug instead of a feature request because I'd assume that all slots being accessed the same way is desirable, but feel free to move this to a feature request if that feels like a better fit.

## 🤔 Expected Behavior

I was expecting to be able to populate the end slot with default content using an approach in `compose` similar to the `checkedIndicator` in the [checkbox component](https://github.com/microsoft/fast/blob/e498f1e35232169d2bf80e366350d0cde812d76d/packages/web-components/fast-components/src/checkbox/index.ts#L21), as an example.

## 😯 Current Behavior

I am able to populate the end slot from within the `connectedCallback` in my component using `this.end.innerHTML`, but this doesn't feel as clean.

## 💁 Possible Solution

Add support in the [start/end templates](https://github.com/microsoft/fast/blob/062b6642246a8bf54a1371daeee4328660672f3f/packages/web-components/fast-foundation/src/patterns/start-end.ts#L33) to provide a default value via the `definition` object used in a given component's template and add these items to the FoundationElementDefinition.

## 🔦 Context

In my case, I am adding an icon to the text-field component (a component we're building which extends the `FoundationTextField`) that is displayed when the text field is in an invalid state.

## 🌍 Your Environment

* OS & Device: MacOS
* Browser Google Chrome
* Version 92.x

chrisdholt:
# Pull Request

## 📖 Description
This PR adds support for adding default slotted content to start/end slots where supported. To enable this we did need to add new types for certain control options.

The original templates have been marked as deprecated and the new templates have been named to specify they are *slot* templates. All components have been updated to use the new templates.

Closes #5116 

### 🎫 Issues
When providing default slotted content to start/end the assignedNodes() array is empty so to accommodate this I've done an initial pass to set the `start` class by default if the definition exists. From my testing, the slotchange will still fire and remove if necessary.

### General

### Component-specific

## ⏭ Next Steps
<s>The change to the template as a function to support definition is *breaking*. To offset this, we could update the template name and use it internally within our components while maintaining the existing start and end templates until the next major version where they can be removed. Thoughts welcome here on how to best support both.</s>

### DIFF
--- a/change/@microsoft-fast-components-6ac044cd-7329-49d7-a97e-ed9bb5a62b2a.json
+++ b/change/@microsoft-fast-components-6ac044cd-7329-49d7-a97e-ed9bb5a62b2a.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add start/end slot definitions to support default slotted content",
+  "packageName": "@microsoft/fast-components",
+  "email": "chhol@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-foundation-36508b94-303a-41ad-9c37-dfe77c29c392.json
+++ b/change/@microsoft-fast-foundation-36508b94-303a-41ad-9c37-dfe77c29c392.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add start/end slot definitions to support default slotted content",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "chhol@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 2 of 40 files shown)


EisenbergEffect (approved):

EisenbergEffect:
I think we should keep the original template and mark it as deprecated so that it's not a breaking change. Since our actual templates use the new start/end template internally, the old one should be tree-shaken away unless someone has taken an explicit dependency on it when building their own components.

scomea (approved):

eljefe223 (approved):

eljefe223 (approved):

nicholasrice (approved):

PR #5141, opened 2021-09-01
lock yarn to a specific version

### LINKED ISSUE #4493: Lock yarn to a specific version
Use `yarn policies set-version` to enforce a single version of Yarn across the monorepo.

This would ensure developers are all using the same version of Yarn while working within the project.

1. Add an `engines` property to the root-level `package.json` to ensure that everyone has at least yarn version 1.13 (the first version to support local instances):
    ```json
    "engines": {
        "yarn": ">=1.13 <2"
    }
    ```

2. Run `yarn policies set-version` in the root directory. This will download the latest 1.x version of Yarn and place it in the `.yarn/releases` directory. This file and directories should be committed and pushed.

3. Add this line to the root-level `.gitattributes` file to prevent Github from including the Yarn source as part of the [language breakdown graph](https://github.com/github/linguist/blob/master/docs/overrides.md):
    ```gitattributes
    .yarn/releases/*.cjs linguist-vendored
    ```

These changes should only be made for the root-level monorepo and should not be replicated within the individual packages. **The `engines` field in particular should only be needed in the main `package.json` since we don't want it to be a _directive_ as a downstream dependency.**

Developers will still need a global instance of `yarn`. The global instance automatically checks the `.yarnrc` file and switches to use the local version referenced by the `yarn-path` field.

* `yarn policies` command: https://classic.yarnpkg.com/en/docs/cli/policies
* `yarn-path` field for `.yarnrc`: https://classic.yarnpkg.com/en/docs/yarnrc#toc-yarn-path
* `engines` field for `package.json`: https://classic.yarnpkg.com/en/docs/package-json#toc-engines

chrisdholt:
# Pull Request

## 📖 Description
Implements a single version policy for yarn for development.

closes #4493 

### General

### Component-specific


EisenbergEffect (approved):

scomea (approved):

.gitattributes
```diff
@@ -1,2 +1,5 @@
 # Set the default behavior, in case contributors don't have core.autocrlf set.
 * text=auto
+
+# prevent Github from including yarn source
+.yarn/releases/*.cjs linguist-vendored
```
radium-v, line 5: The file added is `.yarn/releases/yarn-1.22.11.js` but this is referencing `.yarn/releases/*.cjs`.

radium-v (changes requested):
When I run `yarn policies set version 1` it downloads the file as `yarn-1.22.11.cjs`. This might be due to the version of Node used when installing (I'm on 14.17.5), though I'm not sure.

.gitattributes
chrisdholt: I went based on your suggestion. I'll edit to add both. As a heads up, I'm using the same and seeing the same. Probably best to support both.
image

chrisdholt:
> When I run `yarn policies set version 1` it downloads the file as `yarn-1.22.11.cjs`. This might be due to the version of Node used when installing (I'm on 14.17.5), though I'm not sure.

Due to multiple versions of yarn existing and the primary global being outdated. Fixed and pushed.

radium-v (approved):

PR #5147, opened 2021-09-01
fix: improve CSSDesignToken performance

nicholasrice:
# Pull Request

## 📖 Description
This PR improves runtime performance of CSSDesignToken implementations. Specifically, it implements a new strategy for attaching the CSS custom properties that emit to elements. Prior to this change, each property -> value pair created a new ElementStyles instance that was re-used across components. The downside to that approach is that in cases where we need to attach a lot of ElementStyles to a FASTElement as a result of DesignToken changes (think Card, where fillColor is set for every card and many recipes depend on fillColor), it meant a significant number of ElementStyles were being attached. And because the adoptedStyleSheets array is not mutable, it required significant array creation via object spreading, causing considerable negative performance  impacts.

This change implements a new approach where each DesignToken target gets a single stylesheet attached to it, and changes to that sheet are done using the CSSOM. This reduces the number of sheet attachments that must be performed and significantly speeds up card-scenarios. 

It also has the added benefit of improving the debug experience:

**New Approach Debugger Output:**

image

**Old Approach Debugger Output:**

image

### General

### Component-specific


packages/utilities/fast-benchmarks/benchmarks/dt-add-remove/index.ts
```diff
@@ -0,0 +1,26 @@
+import { Card, fastCard } from "@microsoft/fast-components";
+import { DOM, Observable } from "@microsoft/fast-element";
+import { DesignSystem } from "@microsoft/fast-foundation";
+
+DesignSystem.getOrCreate().register(fastCard());
+
+export default new Benchmark(
+    "queue-update",
+    function (deferred: any): void {
```
nicholasrice, line 9: This benchmark of adding a creating a card, adding it to the DOM, and then removing it from the DOM is seeing ~68% ops/s improvement:

```
hz change: 835.1125631303578
hz percentage change: 68.03439047252512
mean change: -0.00032984796150013643
mean percentage change: -40.48837281535488
```

EisenbergEffect (approved):
Nice!

chrisdholt (approved):

Issue #5121, opened 2021-08-30
fix: Programmatic focus in Firefox


jumarroq:
# 🐛 Bug Report

Calling "focus" on web components does not always work the same as it does on Chrome or Edge Chromium. This seems to depend on how nested the component is and repros when circumventing FAST code by directly calling focus on the button you want to have focused.

## 💻 Repro or Code Sample

Have yet to find this repro in the FAST repo. This has been first seen in Firefox using this URL: https://www.msn.com/en-us/feed?ocid=windirect&pcsonly=true&item=flights%3Anone&item=spalink%3Alatest%2Fdebug

Find a card in the river
Select "Fewer Stories like this" under the card action menu. 
![image](https://user-images.githubusercontent.com/68871578/131374238-c320d44e-e772-4bc9-8418-33b7c26f3680.png)
"Not interested in this story" should be focused like in Edge, but instead focus is set to the browser.
![image](https://user-images.githubusercontent.com/68871578/131374311-610c3a5e-c160-428d-b70c-c285eb61b4ee.png)

## 🤔 Expected Behavior

After selecting Fewer Stories like this, the first button should have focus set to it:
![image](https://user-images.githubusercontent.com/68871578/131374515-bf6f1912-6688-4e3e-894f-2c5784e23502.png)

## 😯 Current Behavior

Focus is set to the browser instead of to the button.

## 💁 Possible Solution

Looks to be a browser difference with Firefox and perhaps how it handles shadow dom and focus. I was able to take our code out of the equation by setting focus on the button manually using setInterval(() => {button.focus()}, 500), which actually changed the behavior but didn't totally fix the issue. With that, the card itself was receiving focus every 500ms, but not the button. If I hit tab, I would go into the card and then the button would always get focused properly every 500ms. It's as if Firefox didn't know the button was there until I entered it. Before and after that the chain of activeElements (going through the shadow DOM) looked the same.

## 🌍 Your Environment

* Windows 10
* Browser [FireFox]
* Version [91]

chrisdholt:
Thanks @jumarroq.

@EisenbergEffect could this be related to delegatesFocus not being fully supported? It seems odd that focus would move the the *window* though...

EisenbergEffect:
I do agree that the browser focus is unexpected, but maybe that's what FF does when it doesn't know what to do? I seems like it could be related to delegatesFocus but I'm not sure.

jumarroq:
[...]

PR #5150, opened 2021-09-03
Using a manual focus to workaround delegatesFocus issues on Firefox

jumarroq:
# Pull Request

## 📖 Description

Working around the lack of delegatesFocus support on Firefox specifically for the FastButton and FastAnchor which rely on it to pass focus to their inner button and anchor elements. This is done only when delegatesFocus is undefined which indicates that delegatesFocus isn't supported (at least that's how it works in Firefox).  When delegetesFocus is undefined I manually set focus on the inner element otherwise I do nothing and let the default behavior handle the focus.

This is a workaround and is not perfect, there is still a difference in behavior when autofocus is used. When autofocus is set on an element on the page, that will happen before the focus is set. In the story book for the dialog you can see that where delegatesFocus is supported the autoFocused element gets focus properly, but in Firefox, the first button gets focus as a result of the Dialog's "focus first element" logic. I don't think this edge case is worth trying to fix as it would generally be a bad idea to have two elements that are supposed to get automatic focus on the page anyways.

### 🎫 Issues

https://github.com/microsoft/fast/issues/5121

## 👩‍💻 Reviewer Notes

I would appreciate help testing for any browsers outside of Chrome, Edge Chromium, or Firefox as needed.

## 📑 Test Plan

Tested using the Dialog in the fast-components storybook

### General

### Component-specific

## ⏭ Next Steps

This code should be cleaned up once FireFox officially supports delegatesFocus


packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -94,6 +94,17 @@ export class Anchor extends FASTElement {
      * References the root element
      */
     public control: HTMLAnchorElement;
+
+    /**
+     * Manually calls the control's focus when delegatesFocus is unsupported
+     */
+    public handleUnsupportedDelegatesFocus(): boolean {
```
chrisdholt, line 101: I think if we call this focus, then we can probably keep from needing to add the `<template>` tag and alter the DOM structure (which would be breaking). That *should* enable us to just do the following:

```html
<fast-anchor id="anchor">Foo</fast-anchor>
```
```ts
const anchor = document.getElementById("anchor");

anchor.focus();
```

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -187,6 +193,17 @@ export class Button extends FormAssociatedButton {
...
+    /**
+     * Manually calls the control's focus when delegatesFocus is unsupported
+     */
+    public handleUnsupportedDelegatesFocus(): boolean {
+        if (this.shadowRoot && this.shadowRoot.delegatesFocus === undefined) {
+            this.control.focus();
+            return false;
+        }
+        return true;
+    }
```
chrisdholt, line 206: I think if we call this focus, then we can probably keep from needing to add the `<template>` tag and alter the DOM structure (which would be breaking). That *should* enable us to just do the following:

```html
<fast-button id="button">Foo</fast-button>
```
```ts
const button = document.getElementById("button");

button.focus();
```

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -3,6 +3,12 @@ import { ARIAGlobalStatesAndProperties, StartEnd } from "../patterns/index";
 import { applyMixins } from "../utilities/apply-mixins";
 import { FormAssociatedButton } from "./button.form-associated";
 
+declare global {
+    interface ShadowRoot {
+        delegatesFocus: boolean;
+    }
+}
+
```
chrisdholt, line 11: Do we need to declare this on the global? I think this likely exists already...

chrisdholt (changes requested):
Thanks for the contribution @jumarroq! I left a couple comments on things I think we'll want to adjust. The primary thing we'll want to avoid here is changing the DOM structure of the template to handle this. I think we may be able to make this change without needing to adjust the DOM for each of these by exposing a public method for `focus` which when called on the custom element would focus the control.

jumarroq:
> Thanks for the contribution @jumarroq! I left a couple comments on things I think we'll want to adjust. The primary thing we'll want to avoid here is changing the DOM structure of the template to handle this. I think we may be able to make this change without needing to adjust the DOM for each of these by exposing a public method for `focus` which when called on the custom element would focus the control.

This was my initial attempt, the reason I ended up changing the template was to not override the focus method for compatible browsers, simply not doing the manual focus didn't work as we were overriding the default behavior and then the default behavior was gone. In order to call the default focus method for the compatible browsers I considered doing something like `HTMLInputElement.prototype.focus.apply(this);` so that might work, I just need to figure out how to get the prototype for the ShadowRoot I think. If you know more about that I'd appreciate guidance.

packages/web-components/fast-foundation/src/button/button.ts
jumarroq: It doesn't, at least my build was breaking without this, is there anywhere else I can add this?

packages/web-components/fast-foundation/src/button/button.ts
jumarroq: See my other reply. I definitely would rather do that to keep it simple, but I don't quite know how to avoid messing with the compatible browsers.

scomea:
Given that Firefox looks to be about to fix this on their side (https://phabricator.services.mozilla.com/D123858) do we intend this as a stopgap or a permanent thing?

jumarroq:
> Given that Firefox looks to be about to fix this on their side (https://phabricator.services.mozilla.com/D123858) do we intend this as a stopgap or a permanent thing?

I would have it just be a stopgap since it's a little messy and we only want this for firefox for the time being.

packages/web-components/fast-foundation/src/button/button.ts
chrisdholt: Sure, I see this having not been resolved by TS either yet

packages/web-components/fast-foundation/src/button/button.ts
chrisdholt: I think the problem with the current approach is that it fundamentally breaks the template in a way that is not backwards compatible and is ultimately for (important) but small scenario. I think we need to do this in a way that does not alter the template, which adds complexity but makes it much easier to treat this as a fix and one which can be removed easily over time.

jumarroq: @chrisdholt  does this approach that I suggested make sense? 

"Another approach I considered but felt it might be too wasteful was to add an event listener to the element within the button.ts/anchor.ts I can just listen for the focus event there and then use my focus handler. If that's fine too, that's probably the most straight forward solution."

chrisdholt: I think this is probably reasonable for the time being - less than ideal, but all of this is less-than-ideal, which is why it's a stopgap :)

jumarroq:
@chrisdholt @scomea 

I realized in testing that the approach for detecting if delegatesFocus is available doesn't work because the shadow root is closed in the relevant scenarios. It wasn't closed in the story book which is why I didn't catch this sooner. 

Do you have any suggestions to work around this? Checking the user agent string is probably the most straight forward way. 

We might also be able to get if delegatesFocus is available by writing some code to create a dummy element with delegatesFocus true and use the result of that to set some variable we can then reference, but that's obviously more convoluted and not sure if it would work.

scomea:
Do we know of any consequences to simply having a focus() handler that just calls focus() on the control element when called?

jumarroq:
> Do we know of any consequences to simply having a focus() handler that just calls focus() on the control element when called?

@scomea  It does cause some different behavior when an autofocus button is on the page and some other button gets focus called on it. Common scenario for this would be if a dialog has an autofocus button and a different button is the first tabbable button. Ordinarily the autofocus'd button would get focus first, but when I overwrite the focus function, the focus happens AFTER the autofocus that the browser does so the autofocus'd button effectively doesn't get focus. That's the only difference I saw in my testing.

jumarroq:
@scomea actually it also changes the behavior of when the focus indicator appears. When opening a dialog that sets the focus, the focus indicator only appears when the user uses the keyboard to open the dialog. A mouse user will not show the focus indicator for some reason (I think it's because the dialog itself doesn't properly have focus), though the first button will be effectively focused as narrator will focus on the right element and keyboard actions will act normally.

jumarroq:
@chrisdholt @scomea I found out a way to detect if we're in firefox without using a UA string if that's any better. Apparently we can just check:

typeof InstallTrigger !== 'undefined'

InstallTrigger is something only available in firefox.

chrisdholt:
> @chrisdholt @scomea I found out a way to detect if we're in firefox without using a UA string if that's any better. Apparently we can just check:
> 
> typeof InstallTrigger !== 'undefined'
> 
> InstallTrigger is something only available in firefox.

Sounds good - let's be sure to add a comment for these code blocks and why they exist so we can pull it out once support reaches a reasonable benchmark.

packages/web-components/fast-foundation/src/anchor/anchor.template.ts
```diff
@@ -8,43 +8,45 @@ import type { Anchor } from "./anchor";
...
-        aria-roledescription="${x => x.ariaRoledescription}"
-        ${ref("control")}
-    >
-        ${startTemplate}
-        <span class="content" part="content">
-            <slot ${slotted("defaultSlottedContent")}></slot>
-        </span>
-        ${endTemplate}
-    </a>
+    <template @focus="${x => x.handleUnsupportedDelegatesFocus()}">
```
scomea, line 11: If we added the event listener with code in the component's connected handler I think we could avoid needing to mess with the template at all.

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -187,6 +193,17 @@ export class Button extends FormAssociatedButton {
     }
 
     public control: HTMLButtonElement;
+
+    /**
+     * Manually calls the control's focus when delegatesFocus is unsupported
+     */
+    public handleUnsupportedDelegatesFocus(): boolean {
```
scomea, line 200: This handler should be marked as "internal".

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -8,6 +8,15 @@ import { applyMixins } from "../utilities/apply-mixins";
 import type { FoundationElementDefinition } from "../foundation-element";
 import { FormAssociatedButton } from "./button.form-associated";
 
+/**
+ * Define InstallTrigger as part of the Window to check for its presence in FireFox for working around the missing delegatesFocus support
+ */
+declare global {
+    interface Window {
+        InstallTrigger: any;
+    }
```
scomea, line 17: In other places that we did something like this we've declared a new interface and use that in the tests, seems safer than touching the Window interface directly:    

`interface WindowWithInstallTrigger extends Window {
        InstallTrigger: any;
    }`

jumarroq: Do you have an example with that? I guess I would then do something like 

`(window as WindowWithInstallTrigger).InstallTrigger`?

scomea (approved):

change/@microsoft-fast-foundation-377b942b-ea56-4464-beeb-d6bb80232402.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Using a manual focus to workaround delegatesFocus issues on Firefox",
```
chrisdholt, line 3: If we aren't going to address the issue holistically across the repo, can you scope the message here to just the components touched as part of this exercise?

packages/web-components/fast-foundation/src/anchor/anchor.ts
```diff
@@ -105,6 +105,33 @@ export class Anchor extends FoundationElement {
...
+     * This should be removed as soon as FireFox supports delegatesFocus properly,
+     * the check for override does not work for any browser outside of FireFox.
+     * Relevant PR on the Firefox browser: https://phabricator.services.mozilla.com/D123858
+     */
+    private handleUnsupportedDelegatesFocus = () => {
+        // InstallTrigger is only defined in FireFox
+        if (
+            typeof ((window as unknown) as WindowWithInstallTrigger).InstallTrigger !==
+            "undefined"
+        ) {
```
radium-v, line 129: Since this only affects buttons/anchors when `delegatesFocus` doesn't work, we could just check to see if `delegatesFocus` is set and apply the focus regardless of the support:

```suggestion
        if (this.$fastController.definition.shadowOptions?.delegatesFocus) {
```

While `shadowRoot.delegatesFocus` may not be available in `closed` mode, the `$fastController` definition will always be available regardless of the mode used.

jumarroq: Yea, but I don't see how the $fastController helps us. I saw that you can see the shadowOptions that were passed in, but that doesn't help us because the option passed in is the same. Is there a property in $fastController we can use?

scomea: Isn't `$fastController.definition.shadowOptions?.delegatesFocus` our construct and could still be true even if Firefox ignores the setting?

EisenbergEffect: Yes. That's whatever config is passed when the component is defined. It doesn't get altered based on platform capabilities.

radium-v: Another potential check could be to see if `window.ShadowRoot?.prototype.hasOwnProperty("delegatesFocus")` - Firefox and Safari 14 return `false`, and Chrome and Edge return `true`.

jumarroq: That sounds good to me, I can make that change if we're confident in it, though I'd rather not churn on this PR much more if we've already got a working solution for FireFox.

radium-v (approved):

chrisdholt (approved):

chrisdholt:
Thanks @jumarroq - we'll work on getting this merged and published!

packages/web-components/fast-foundation/src/anchor/anchor.ts
chrisdholt: Thanks for working through this all and finding a solid solution that will work cross browser!

PR #5157, opened 2021-09-07
fix(design-system): make FoundationElement.compose safe to use directly

EisenbergEffect:
# Pull Request

## 📖 Description

This PR adds a small bit of logic to detect the scenario where a `FoundationElement` is trying to define a component directly for `FoundationElement` rather than a subclass. The code will automatically insert a subclass into the hierarchy so that:

1. Other elements based on `FoundationElement` don't behave unpredictably.
2. It's now easier to define presentation-only elements that have no need for a special class.

To demonstrate point 2, here's a before and after from a demo.

Before
```ts
export class Attribution extends FoundationElement {

}

export const fluentAttribution = Attribution.compose({
	baseName: "attribution",
	template: html`...`,
	styles: css`...`
});
```

After
```ts
export const fluentAttribution = FoundationElement.compose({
	baseName: "attribution",
	template: html`...`,
	styles: css`...`
});
```

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Just a simple condition added to check the type being defined. Not much to see really.

## 📑 Test Plan

A new unit test was added for this scenario.

### General

## ⏭ Next Steps

I'm wondering if we should add a similar check for `FASTElement` but at a lower level. Not sure if it's needed or not.


change/@microsoft-fast-foundation-fb2bff9c-3129-4433-bf60-2b85c82258d4.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix(design-system): don't allow FoundationElement to be defined directly",
```
chrisdholt, line 3: I think I'm reading this wrong, but this reads as we are *preventing* - should we drop "don't" in favor of just "allow FoundationElement..." so it reads as this is supported*.

packages/web-components/fast-foundation/src/design-system/design-system.ts
```diff
@@ -3,6 +3,7 @@ import {
     FASTElementDefinition,
     PartialFASTElementDefinition,
 } from "@microsoft/fast-element";
+import { FoundationElement } from "..";
```
chrisdholt, line 6: I *think* we'll want a `/` here - it may be fine but I know we've seen issues related to pathing without one previously.

```suggestion
import { FoundationElement } from "../";
```

chrisdholt (approved):

change/@microsoft-fast-foundation-fb2bff9c-3129-4433-bf60-2b85c82258d4.json
EisenbergEffect: I mean, you can no longer have this happen: `customElements.define('foo-bar', FoundationElement)`. Not sure the best way to worth that.

packages/web-components/fast-foundation/src/design-system/design-system.ts
EisenbergEffect: VSCode. I'll fix shortly.

nicholasrice (approved):

packages/web-components/fast-foundation/src/design-system/design-system.ts
EisenbergEffect: This revealed some circular module issues. Working to fix that now.

Issue #5089, opened 2021-08-18
feat: update all components to reference `event.key` instead of `event.keyCode` (deprecated)


chrisdholt:
# 🙋 Feature Request
The API for `event.keyCode` is deprecated and likely not one which we should continue relying upon. In it's place, we should likely replace with `event.key`. Another alternative would be `event.code`, but there may be certain implications with regard to keyboard navigation which could yield unexpected results.

This task exists to track an update of any existing use of `event.keyCode` and replace it with `event.key`.

PR #5162, opened 2021-09-07
fix: use `key` instead of `keycode` to identify keystrokes

scomea:
## 📖 Description

Remove dependencies on keycode from FAST code as `keycode` is deprecated.  Keeping the `keycode` related utilities so this won't be a breaking change, we should remove those on the next major version.

### 🎫 Issues

fixes https://github.com/microsoft/fast/issues/5089

### General

### Component-specific


packages/utilities/fast-web-utilities/src/key-codes.ts
```diff
@@ -1,6 +1,6 @@
 /*
  * Key Code values
- * @deprecated - use individual keycode exports
+ * @deprecated - KeyCodes are deprecated, use individual string key exports
```
nicholasrice, line 3: Two things here:

1. It doesn't look like this is actually jsdoc / tsdoc syntax, so the deprecation comments are getting lost to the d.ts files when this is built. Can you change:

```ts
/*
 * ...
 */
```

to this:
```ts
/**
 * ....
 */
```
You should see the `@deprecated` flag light up with syntax highlighting. Once you do, you should see the `@deprecated` in the emitted d.ts files.

2. For the const exports, you're actually only deprecating `keyCodeAlt`, not the entire block. I know it's a pain, but would you add the `@deprecated` to *every* export so that authors will see the warning in their editors?

janechu (approved):

janechu:
Thanks for including the tooling updates!

packages/utilities/fast-web-utilities/src/key-codes.ts
nicholasrice: Thanks!

nicholasrice (approved):

chrisdholt (approved):

Issue #5132, opened 2021-08-31
Commit inline edited text on defocus


janechu:
# Description

Other design editors currently follow the UX where editing text and defocusing (for any reason) causes the text to be commited.

## Requirements

- Allow any defocus (even by Escape key) to commit the text that is being edited

PR #5163, opened 2021-09-08
feat: add a message queue to the message system

janechu:
# Pull Request

## 📖 Description

This change creates a queue for messages received by the message system, the messages are discharged in the order they are received to prevent async issues where messages may execute in the wrong order.

### 🎫 Issues

Related to #5132

## 👩‍💻 Reviewer Notes

Build and run the FAST Creator, messages should get sent per normal and no functionality should be missing.

## 📑 Test Plan

Unit tests added for the MessageSystem and updated for the web worker.

### General


williamw2 (approved):

packages/tooling/fast-tooling/src/message-system/webworker.ts
```diff
@@ -3,7 +3,7 @@
 import { getMessage } from "./message-system.utilities";
 
 onmessage = function (e: MessageEvent): void {
-    if (e.data && e.data.type !== undefined) {
+    if (e.data && e.data?.[0]?.type) {
```
nicholasrice, line 6: I think this is checking `e.data` twice right? If so, this could be reduced.

packages/tooling/fast-tooling/src/message-system/message-system.ts
```diff
@@ -79,23 +85,54 @@ export default class MessageSystem<C = {}> {
...
+        this.sendNextMessage();
     };
 
+    /**
+     * Fire the messages in the order they have been received when they are made available
+     */
+    private sendNextMessage = (): void => {
+        if (this.messageQueue[1][0] && this.messageQueue[0][this.messageQueue[1][0]]) {
+            const updatedEvent = new MessageEvent("message", {
+                data: this.messageQueue[0][this.messageQueue[1][0]].data[0],
```
nicholasrice, line 115: You might consider storing `this.messageQueue[0][this.messageQueue[1][0]` as a const so you you don't need to index the array so many times

packages/tooling/fast-tooling/src/message-system/webworker.ts
janechu: Whoops, you're right

nicholasrice (approved):

packages/tooling/fast-tooling/src/message-system/webworker.ts
janechu, line 6:
```suggestion
    if (e.data?.[0]?.type) {
```

PR #5169, opened 2021-09-08
feat: add option to html renderer inline edit to commit changes on blur

williamw2:
# Pull Request

## 📖 Description

This adds an option to the html-render-layer-inline-edit component to commit text changes whenever the text area loses focus and not just when the Enter key is pressed. This feature is also activated by default for the Creator.
There are also a number of bug fixes that were exposed by this change including to the tab navigation in html-render and update messages in the monaco editor in site-utilities. 

### 🎫 Issues

closes: #5132 

## 👩‍💻 Reviewer Notes

To see the change in action start the Creator and add several elements that contain text (like "p"). Double-click the text to edit it. Make a change to the text and do something to cause the textarea to lose focus (click something else, tab out, double-click something else, etc). The change should be committed and reflected in the UI.

### General

### Component-specific


change/@microsoft-fast-tooling-087591f1-07f0-4d70-a05f-787ffb87ff29.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "adding option to commit inline edit changes on blur",
```
janechu, line 3:
```suggestion
  "comment": "added option to commit inline edit changes on blur",
```

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.ts
```diff
@@ -17,6 +22,14 @@ export class HTMLRenderLayerInlineEdit extends HTMLRenderLayer {
     @attr({ mode: "boolean" })
     public autoselect: boolean;
 
+    /**
+     * Specifies when changes to the text should be committed.
+     * Only on "Enter" keypress or any time the textarea loses focus.
+     * Default: CommitMode.onBlur
+     */
+    @attr
+    public commitmode: CommitMode;
```
janechu, line 31: Should we allow mutliple types of changes? `CommitMode[]`? Also wondering if this should be `commitMode` to follow your casing below.

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.ts
```diff
@@ -63,9 +84,11 @@ export class HTMLRenderLayerInlineEdit extends HTMLRenderLayer {
 
     public handleTextInput(e: KeyboardEvent) {
         if (e.key === "Enter") {
-            this.commitEdit();
-            e.preventDefault();
-            return false;
+            if (this.commitmode === CommitMode.onEnter) {
```
janechu, line 87: should this be included in the enclosing `if`? Combined with my change above you could use `if (e.key === "Enter" && this.commitMode.includes(CommitMode.onEnter))`

janechu (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.ts
janechu: That was a bad suggestion, since this is an attr, we could do separate attrs `commitOnBlur` and `commitOnEnter` if we wanted to allow multiple commit modes.

EisenbergEffect (approved):

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.ts
williamw2: As per our offline discussion I have decided to change the commit modes to "onEnterOnly" and "onBlurOrEnter". The default behavior is now for the text changes to be committed any time the text area loses focus or if the Enter key is pressed. The "onEnterOnly" option will only commit when Enter is pressed and cancels the change on blur.

packages/tooling/fast-tooling/src/web-components/html-render-layer-inline-edit/html-render-layer-inline-edit.ts
williamw2: You were right, but this condition has now been removed (see above comment).

janechu (approved):

PR #5181, opened 2021-09-11
convert css custom property on select listbox element to design token

chrisdholt:
# Pull Request

## 📖 Description
**Problem** - The issue is outlined in the discord thread which can be found here. TLDR; combobox and select set values to the style attribute of the listbox DOM node within the template. In the event tokens are set for parts of the shadow root, the setting of the style attribute in the template invalidate the token updates and cause the node to reset w/ just the templated changes.

**Alternative solution** - We could merge the style object.

**Proposed solutions** - I've provided two proposed solutions, each in their own commit. 

- _Proposal 1_: The first proposal creates a design token at the Foundation level. Because the logic for creating the css custom property currently exists at the Foundation level, this maintains that expectation by creating a design token. With that said, this creates a pattern for creating, exporting and depending on design tokens at the Foundation level. This is less than desirable.
- _Proposal 2_: The second proposal extends both select and listbox, super's the `setPosition` logic and creates a token at the `component library` level. This approach is preferred as it maintains visual styling and tokens at the design system level. This approach does create a potential breaking change for teams creating their own select and combobox using the template and rely on the current `--max-height` custom property in their own styles.

### 🎫 Issues
See above. We'll also want to decide where/how to best document this change.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/docs/api-report.md
+++ b/packages/web-components/fast-components/docs/api-report.md
@@ -21,7 +21,7 @@ import { Button as Button_2 } from '@microsoft/fast-foundation';
 import { Card as Card_2 } from '@microsoft/fast-foundation';
 import { Checkbox } from '@microsoft/fast-foundation';
 import { CheckboxOptions } from '@microsoft/fast-foundation';
-import { Combobox } from '@microsoft/fast-foundation';
+import { Combobox as Combobox_2 } from '@microsoft/fast-foundation';
 import { ComboboxOptions } from '@microsoft/fast-foundation';
 import { Container } from '@microsoft/fast-foundation';
 import { DataGrid } from '@microsoft/fast-foundation';
@@ -54,7 +54,7 @@ import { ProgressRingOptions } from '@microsoft/fast-foundation';
 import { Radio } from '@microsoft/fast-foundation';
 import { RadioGroup } from '@microsoft/fast-foundation';
 import { RadioOptions } from '@microsoft/fast-foundation';
-import { Select } from '@microsoft/fast-foundation';
+import { Select as Select_2 } from '@microsoft/fast-foundation';
 import { SelectOptions } from '@microsoft/fast-foundation';
 import { Skeleton } from '@microsoft/fast-foundation';
 import { Slider } from '@microsoft/fast-foundation';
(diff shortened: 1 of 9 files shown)


bheston:
Good find. Is the issue specifically because the token is bound within `style`? That is, do the boolean bindings for `disabled` and `hidden` also cause the template to re-render? Or even one of the aria properties elsewhere in the template. Is the template processor replacing the entire content of the custom element, or only updating the parts that changed?

It looks like we're only using `maxHeight`  as an observable, so if we convert it to a design token I think we don't still need the observable property. That seems alright in limited use, but it does feel a bit heavy if that becomes the only way we can get a value that might change into a template that might also get style or local overrides.

chrisdholt:
> Good find. Is the issue specifically because the token is bound within `style`? That is, do the boolean bindings for `disabled` and `hidden` also cause the template to re-render? Or even one of the aria properties elsewhere in the template. Is the template processor replacing the entire content of the custom element, or only updating the parts that changed?
> 
> It looks like we're only using `maxHeight` as an observable, so if we convert it to a design token I think we don't still need the observable property. That seems alright in limited use, but it does feel a bit heavy if that becomes the only way we can get a value that might change into a template that might also get style or local overrides.

So, these don't necessarily require a re-render of the entire node. Observable (and @EisenbergEffect or @nicholasrice can correct me here) allows us to target each specific observable property on the node and only make updates as necessary. The specific issue here was with the style object as that was *changing* on toggle and thus triggering the changed even which would update accordingly with the `maxHeight`. So, it's only updating the parts that changed, it just so happens that was what was happening with the style object and this CSS custom property.

I think we still want `maxHeight` as an observable property so that it can listen to changes. I thought about creating the design token on the class itself, but then I thought about how that likely doesn't easily allow reuse within the stylesheet across both Combobox and Select. This is the primary reason that I name-spaced the token itself. This doesn't have to be **the** solution, but it is _a solution_ and I think it's a potential pattern. With that said, I do have it as a draft to gather feedback like this!

nicholasrice:
I don't understand the issue, can you clarify? What is the "style object"? How is setting the style attribute invalidating the tokens?

chrisdholt:
> I don't understand the issue, can you clarify? What is the "style object"? How is setting the style attribute invalidating the tokens?

Here's a link to a bit of the offending code: https://github.com/microsoft/fast/blob/b2283a1a2c2f35a71cf8f87e0d65cce2b7f532be/packages/web-components/fast-foundation/src/select/select.template.ts#L59

Here, we have the style attribute of the listbox node being set inline of the template to provide a variable for the listbox height. When @bheston was targeting that listbox node in the shadow and updating the token values those values would set perfectly fine. When the state of the listbox changes, it modifies the variable which rewrites the style attribute which causes that style object to update/change and remove the token updates that had just been done.

chrisdholt:
Closing this in favor of an alternative approach!

Issue #4825, opened 2021-06-04
fix: slider's thumb stays active after you go outside the container element


UtmostCreator:
# 🐛 Bug Report

If you hold down the left mouse button (LMB) on a thumb and start dragging it, and after you move your cursor outside the container and release the LMB; if you try to move your cursor back to the container (where the slider is, the thumb stays active so as if you weren't releasing the mouse button.

## 💻 Repro or Code Sample

[official example link](https://explore.fast.design/components/fast-slider)
code:
`<fast-slider></fast-slider>
`
## 🤔 Expected Behavior

After releasing the LMB the thumb of a slider should not move

## 😯 Current Behavior

After releasing the LMB the thumb of a slider outside the container, the thumb stays active and moves (when you entering the container); it stops only after you click inside the container.

## 💁 Possible Solution

detect releasing of a button to prevent the button from being inactive (as it is being held down w/o releasing)

## 🔦 Context
—

## 🌍 Your Environment
It can be reproduced only in the chrome browser.
Firefox works fine. 91.0.4472.77 

* OS & Device: Win 10 20h2, OS build 19042.985;
* Browser Google Chrome ( 93.0.4532.0; 91.0.4472.77);
* Version latest (from official web-page components.

chrisdholt:
@robarbms I wanted to check in and see how this is coming along.

PR #5184, opened 2021-09-13
fix: sliders mouseup is not firing when leaving the window

robarbms:
# Pull Request

## 📖 Description

When leaving the window, the slider will not register a mouseup event.

### 🎫 Issues

https://github.com/microsoft/fast/issues/4825

## 👩‍💻 Reviewer Notes

I'm capturing mouseleave on the window document. I found that mouseout fired while still in the window. Perhaps it's firing for the shadow DOMs mouseout? The window element wouldn't fire on mouseleave so I'm using window.document. I attach it on mousedown and remove it on mouseup per the existing pattern.

### General

### Component-specific


packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -465,6 +466,7 @@ export class Slider extends FormAssociatedSlider implements SliderConfiguration
             this.setupTrackConstraints();
             (e.target as HTMLElement).focus();
             window.addEventListener("mouseup", this.handleWindowMouseUp);
```
scomea, line 468: I think we should be making sure these listeners are removed on disconnect.

scomea:
are the horizonal scroll changes related to the slider issue?

packages/web-components/fast-foundation/src/slider/slider.ts
robarbms: This is weird. It's a different PR that I have out. I'll resubmit a corrected PR.

nicholasrice (approved):

nicholasrice (changes requested):
I think something got messed up with the commit history on this PR. can you fix that before merging?

robarbms:
> are the horizonal scroll changes related to the slider issue?

Another PR got merged some how. I've undone the errant changes

packages/web-components/fast-foundation/src/slider/slider.ts
robarbms: There were a lot of listeners, with listeners attaching other listeners. Rather than risk missing one, or a new one being added and by one of the methods, I created a 'remove' flag for each so that I could use the same methods to add and remove event listeners to ensure parity. It feels pretty heavy handed, but I feel like this way would be less prone to errors.

chrisdholt (approved):

chrisdholt:
@robarbms FYI - looks like there are a couple of conflicts with the Horizontal Scroll changes going in.

robarbms:
> @robarbms FYI - looks like there are a couple of conflicts with the Horizontal Scroll changes going in.

There was a whitespace conflict 😑

nicholasrice:
It looks like git history got messed up on this, would you mind resolving prior to merge?

packages/web-components/fast-foundation/src/slider/slider.ts
```diff
@@ -453,26 +460,32 @@ export class Slider extends FormAssociatedSlider implements SliderConfiguration
...
-
-            const controlValue: number =
-                this.orientation === Orientation.horizontal
-                    ? e.pageX - document.documentElement.scrollLeft - this.trackLeft
-                    : e.pageY - document.documentElement.scrollTop;
-
-            this.value = `${this.calculateNewValue(controlValue)}`;
+    /**
+     *
+     * @param e - MouseEvent or null. If there is no event handler it will remove the events
```
nicholasrice, line 469:
```suggestion
     * @param e - MouseEvent or null. If there is no event object provided it will remove the events handlers
```

nicholasrice (approved):
approved pending the git commit history being cleaned up.

Issue #4308, opened 2021-02-02
docs: approaches to customizing components


EisenbergEffect:
We need an article or section of the documentation that lays out the various strategies for customizing a component and provides guidance for which path to take in different scenarios.

usrrname:
Totally agreed. I make this up as I go!

janechu:
This is being addressed in the documentation update for `@microsoft/fast-element` v2.

PR #5186, opened 2021-09-14
docs: new design system example/tutorial

EisenbergEffect:
# Pull Request

## 📖 Description

A simple example doubling as a tutorial showing how to use some of the Design System and Design Token APIs.

### 🎫 Issues

Partial progress towards #4308

## 👩‍💻 Reviewer Notes

Have a read through the source and comments to make sure there are no typos or silly oopsies.
@nicholasrice For some reason the recipe demo isn't working. Can you take a look at that?

## 📑 Test Plan

Give it a run and play around with the demos in the main.ts file.

### General

## ⏭ Next Steps

* Add a readme to the examples folder.
* Remove the Blazor sample and use the readme to point to our fast-blazor repo
* Review the Todo App and add the same level of commenting throughout
* Maybe add some links throughout the main docs site pointing to specific demos where appropriate


examples/design-system-tutorial/src/components/navigation.ts
```diff
@@ -0,0 +1,61 @@
+import {
+    fillColor,
+    neutralLayer2,
+    neutralLayer3,
+    neutralLayer4,
+    neutralStrokeRest,
+    strokeWidth,
+} from "@fluentui/web-components";
```
chrisdholt, line 8: Are our docs the best place for a Fluent example? Does that create clarity or confusion? Perhaps we can enumerate more of the intent in the readme? If this is intended to help those on the Fluent side, I think perhaps we should consider having it exist there.

examples/design-system-tutorial/src/main.ts
```diff
@@ -0,0 +1,132 @@
...
+    PaletteRGB,
+    provideFluentDesignSystem,
+    StandardLuminance,
+    SwatchRGB,
+} from "@fluentui/web-components";
+import { parseColorHexRGB } from "@microsoft/fast-colors";
+import { html } from "@microsoft/fast-element";
+import { demoCard, fluentAttribution, Navigation } from "./components";
+
+Navigation; // Prevent tree shaking.
```
chrisdholt, line 24: Should we add additional comments here as to *why* this exists? IE, because it uses the decorator syntax, we need to create a hard reference to the class in the file to ensure it's not tree shaken away. That isn't necessary for the others due to....

examples/design-system-tutorial/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+A heavily-commented demo of using FAST's Adaptive UI Design System and Design Tokens.
+
+## Setup
+
+`npm install`
+
+## Run
+
+`npm start`
```
chrisdholt, line 11: Should we use npm here? Everything else in the repo is going to reference yarn.

examples/design-system-tutorial/index.html
```diff
@@ -0,0 +1,224 @@
...
+            }
+
+            fluent-demo-card:first-child {
+                grid-column: 2;
+            }
+
+            fluent-demo-card:nth-of-type(2) {
+                grid-column: 3;
+            }
+        </style>
```
chrisdholt, line 56: Nit - could this be in a CSS file? I'd love to see that (this) commented as well.

nicholasrice:
This isn't *only* about this PR but I noticed that none of the projects in "examples" are a yarn workspace. Should they be? Is there a reason we're installing all deps from the registry for these projects?

examples/design-system-tutorial/src/components/card/card.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * We can provide custom configuration for a component by defining a type as
+ * shown below and then using it in template or style factories along with the
+ * call to FoundationElement.compose.
+ */
+export type DemoCardDefinitionOptions = FoundationElementDefinition & {
+    likeButton?: ViewTemplate;
+};
+
+/**
```
nicholasrice, line 13: The recipe example doesn't work because Card doesn't use `neutralFillLayer` anywhere to derive it's fill color.

You'll need to change this definition to something like:

```ts
export class DemoCard extends Card {
    connectedCallback(): void {
        super.connectedCallback();

        const { parentElement } = this;

        if (parentElement instanceof HTMLElement) {
            fillColor.setValueFor(this, (source: HTMLElement) =>
                neutralFillLayerRecipe
                    .getValueFor(this)
                    .evaluate(this, fillColor.getValueFor(parentElement))
            );
        }
    }
}
```

I would probably change them both to use `neutralLayerFloating` and `neutralLayerFloatingRecipe` respectively, I think that is the recipe card's are using in the fluent system. @bheston to confirm.

examples/design-system-tutorial/src/components/navigation.ts
nicholasrice: Is it more clear if we just use fast-components? I don't think we should whole-sale just move this over to fluent.

FWIW we use fluent today in the todo example, so this isn't introducing a new pattern.

examples/design-system-tutorial/index.html
```diff
@@ -0,0 +1,174 @@
...
+<html lang="en">
+    <head>
+        <meta charset="utf-8" />
+        <title>Design System Tutorial</title>
+        <link href="styles.css" rel="stylesheet" />
+    </head>
+    <body>
+        <div id="root">
+            <demo-navigation>
+                <fluent-tree-view>
```
bheston, line 11: If this is going into FAST, should these be `fast` components?

examples/design-system-tutorial/src/components/attribution.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+export const fluentAttribution = FoundationElement.compose({
+    baseName: "attribution",
+    template: html`
+        <slot name="icon"></slot>
+        <slot></slot>
+    `,
+    styles: css`
+        ${display("flex")} ::slotted (*) {
```
bheston, line 30: Probably for further discussion, but I strongly dislike this being one line in our style sheets. It was highly confusing to me at first as I thought they were related. My concern is that someone new to this will think the same. These produce two selectors so should be two separate lines.

examples/design-system-tutorial/src/components/card/card.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+import { css } from "@microsoft/fast-element";
+import { DesignToken, display } from "@microsoft/fast-foundation";
+
+/**
+ * Not only can we have system-wide design tokens, but we can
+ * have component-specific tokens as well. This token enables
+ * changing the corner radius of cards independently of other controls.
+ * However, by default it aliases to the controlCornerRadius token.
+ */
+export const cardCornerRadius = DesignToken.create<number>(
```
bheston, line 21: An example like this is important, but I'm concerned this specific one might be confusing. We have `layerCornerRadius` that's applied to cards by default, so I'd like to avoid the suggestion that to have cards different than "controls" that you need to create new token. This is why in our design token examples page I created something obviously new, I think I called it "special".

What about giving the card `padding` instead and creating a `cardPadding` token? That starts to creep into density, but it's a realistic use case that's come up with cards used across different experiences.

examples/design-system-tutorial/src/components/card/card.styles.ts
```diff
@@ -0,0 +1,77 @@
...
+    "card-corner-radius"
+).withDefault(controlCornerRadius);
+
+export const demoCardStyles = css`
+    ${display("grid")} :host {
+        --elevation: 4;
+        grid-template-rows: calc(${designUnit} * 39px) 1fr auto calc(${designUnit} * 2px);
```
bheston, line 28: I generally recommend avoiding magic numbers in calcs, and for using the `px` for the unit only (so ` * 1px`). I can let it slide here, but I think the height of that first row is aligned with the `max-height` of the slotted `heading` below: `calc(${typeRampPlus1LineHeight} * 2)`.

examples/design-system-tutorial/src/components/card/card.ts
bheston: Cards use `neutralFillLayerRecipe` by default. Floating is used for dialogs or flyouts. Note that this recipe is _not_ one of the fixed `neutralLayer...` (no "Fill") recipes which work based on `baseLayerLuminance`.

examples/design-system-tutorial/src/components/navigation.ts
bheston: I put the same content above. Now that I've gone through more of the code it looks like "fluent" -> "fast" is the strongest option, and one I suggest we take.

examples/design-system-tutorial/src/main.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * In addition to being able to set default values for design tokens, you
+ * can also set a token value so that it only applies to a specific part of
+ * the DOM. Use the setValueFor method of a token to set its value only for the
+ * DOM node provided and its descendants.
+ *
+ * Uncomment the code below to set the fill color only for the first card.
+ */
+
+// fillColor.setValueFor(firstCard, (target: HTMLElement) => neutralPalette.getValueFor(target).get(73));
```
bheston, line 94: We should avoid palette indexes. How about one of the neutralLayer recipes instead.

examples/design-system-tutorial/src/main.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ *
+ * Below we change the neutralFillLayerRecipe to simply grab a fixed color lookup from
+ * the neutral palette. Changing this recipe will affect all tokens based on the recipe.
+ *
+ * Uncomment the code below to see how changing the neutralFillLayerRecipe affects the design.
+ */
+
+// const recipe: ColorRecipe = {
+//     evaluate: (target: HTMLElement) => neutralPalette.getValueFor(target).get(73)
```
bheston, line 117: Per my previous comment, maybe here use the `Palette.colorContrast` function providing `fillColor` as the `reference`. I'd like people to think about the system and not array values that might change.

bheston (approved):

examples/design-system-tutorial/src/components/attribution.ts
nicholasrice: prettier does this - no good way around at this point.

examples/design-system-tutorial/src/components/card/card.styles.ts
nicholasrice: Going to leave this as-is in the interest of not re-working this too much

examples/design-system-tutorial/src/main.ts
nicholasrice: I think this example is primarily about illustrating the basic principles of how these design tokens work, and plugging in recipes obfuscates what is happening. I'm going to leave this as-is (but change it to index a light color now that this is fast-based and not fluent).

examples/design-system-tutorial/src/main.ts
nicholasrice: also going to leave this as-is. `Palette.colorContrast()` requires 3 arguments and I fear that getting the appropriate args and using that function obfuscates what we're trying to show, which is what a "recipe" is (a function that returns a value).

nicholasrice (approved):

examples/design-system-tutorial/index.html
```diff
@@ -0,0 +1,174 @@
...
+                            The Atlantic
+                            <img
+                                src="//img-s-msn-com.akamaized.net/tenant/amp/entityid/BBY4G4r?w=16&h=16&q=100&m=6&f=png&u=t"
+                                alt="The Atlantic"
+                                slot="icon"
+                            />
+                        </fast-attribution>
+                        <img
+                            src="https://www.bing.com/th?id=ORMS.bbec3ae474a1dbbf78c519c6d136c959&pid=Wdp&w=300&h=225&qlt=60&c=1&rs=1"
+                            alt=""
```
chrisdholt, line 36: @nicholasrice @EisenbergEffect We can probably take this as a follow-up, but we should include alt text here.

chrisdholt (approved):

Issue #5047, opened 2021-08-09
fix: style and template functions have incorrect definition types


EisenbergEffect:
# 🐛 Bug Report

## 💻 Repro or Code Sample

The following code produces a compile error because the definition type that is available at override time has less properties than the one that is used at compose time. The compose time type is what our functions are currently based one.

```ts
provideFASTDesignSystem()
    .register(
        fastButton({
            styles: (ctx, def) => css`
                ${buttonStyles(ctx, def)}
                /* add your style augmentations here */
            `
        })
    )
```

## 🤔 Expected Behavior

This should not be a compile error, but should work exactly as above.

## 😯 Current Behavior

At present, a cast to `any` is required on `def`.

## 💁 Possible Solution

We should define an aliased type as `Omit<FoundationElementDefinition, 'baseName'>` and then use that to also build a function type. These should be exposed by `@microsoft/fast-foundation` and then used for all style and template factory function definitions in both `@microsoft/fast-components` and `@fluentui/web-components`.

## 🔦 Context

We have an important note in our documentation informing the community of the need for the cast and our plan to fix the types. We should try to do this as soon as we can spare the time. It's a low-risk, non-breaking change, that enables core scenarios without hacks.

cnunciato:
> At present, a cast to any is required on def.

Thank you! Newbie hitting this, happy to find a workaround here.

lizy0329:
In my case, styles field have incorrect definition types

`Type 'ElementStyles' is not assignable to type 'EagerOrLazyFoundationOption<ComposableStyles | ComposableStyles[], FoundationElementDefinition> | undefined'.`

![1.png](https://i.loli.net/2021/08/31/R3W2MDTsowrxvjI.png)

chrisdholt:
> In my case, styles field have incorrect definition types
> 
> `Type 'ElementStyles' is not assignable to type 'EagerOrLazyFoundationOption<ComposableStyles | ComposableStyles[], FoundationElementDefinition> | undefined'.`
> 
> ![1.png](https://camo.githubusercontent.com/a1b3c6562750230af7927672055a607ef55e8ab2c96e315e4da0e2293fd58ef7/68747470733a2f2f692e6c6f6c692e6e65742f323032312f30382f33312f523357324d4454736f777278766a492e706e67)

I think in this case the expectation is that it's a function which returns the css. What happens when you provide the following?

```ts
styles: (context, definition) => css`
    :host {
        background: #ff7ec9;
    }
`,
```

[...]

PR #5187, opened 2021-09-14
fix type issue for FoundationElementTemplate

radium-v:
# Pull Request

## 📖 Description

Fixes the `TS2344` errors caused by an type incompatibility:

```
Type 'K & FoundationElementDefinition' does not satisfy the constraint 'FoundationElementDefinition'.
Types of property 'template' are incompatible.
```

## 📑 Test Plan

The error should not be present in either `packages/web-components/fast-foundation/dist/fast-foundation.d.ts` or `packages/web-components/fast-foundation/dist/dts/foundation-element/foundation-element.d.ts` after building.

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-foundation-84ca8ddb-ca59-43d3-a8cf-c6da286230da.json
+++ b/change/@microsoft-fast-foundation-84ca8ddb-ca59-43d3-a8cf-c6da286230da.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "fix type issue for FoundationElementTemplate",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "john.kreitlow@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


EisenbergEffect:
Is this the fix for this issue? https://github.com/microsoft/fast/issues/5047 And if so could we also address the style typings?

SethDonohue:
> Is this the fix for this issue? #5047 And if so could we also address the style typings?

This is a fix related to en experience getting typecheck errors during build of fast-foundation. Chris mentioned it may be related to #5047?

eljefe223 (approved):

EisenbergEffect:
This doesn't fix #5047 but seems like a needed change on its own.

EisenbergEffect (approved):
With this change `FoundationElementTemplate` is equivalent to `LazyFoundationOption`. We should probably replace the use of `FoundationElementTemplate` with `LazyFoundationOption` throughout our codebase and then deprecate `FoundationElementTemplate`. The type is also a misnomer as it appears to not have had anything specifically to do with templates.

Approving this PR to move it forward under deadlines but we should probably immediately follow up with the above improvements.

This PR doesn't fix #5047. We'll need additional work for that issues, probably involving some more complex types. Needs experimentation.

SethDonohue (approved):

nicholasrice (approved):

awentzel (approved):

PR #5190, opened 2021-09-15
add lighthouse on CI

Ibrahimmaga:
# Pull Request

## 📖 Description
This change will add a lighthouse to CI such that we can monitor compliance and performance-related metrics. This is a minor step in the right direction and can be evolved further in the future.  

### 🎫 Issues
Feature Improvement not issue.

## 👩‍💻 Reviewer Notes
We should either approve, which I'm in favor of. OR disapprove and abandon as Ibrahim is no longer supporting the project.  But, I believe this is still of value to the project.

### General

### Component-specific

### DIFF
--- a/.github/workflows/ci-validate-platforms.yml
+++ b/.github/workflows/ci-validate-platforms.yml
@@ -236,6 +236,5 @@ jobs:
         with:
           urls: |
             https://www.fast.design/
-          budgetPath: ./budget.json # test performance budgets
           uploadArtifacts: true # save results as an action artifacts
           temporaryPublicStorage: true # upload lighthouse report to the temporary storage
--- a/.github/workflows/ci-validate-platforms.yml
+++ b/.github/workflows/ci-validate-platforms.yml
@@ -240,6 +240,7 @@ jobs:
           branch: ${{ github.ref }}
           outputDirectory: /tmp/artifacts
           emulatedFormFactor: "Desktop"
+          emulatedFormFactor: "Mobile"
           urls: "https://www.fast.design/"
 
       - name: Upload artifacts
--- a/.github/workflows/ci-validate-platforms.yml
+++ b/.github/workflows/ci-validate-platforms.yml
@@ -239,8 +239,7 @@ jobs:
           author: ${{ github.actor }}
           branch: ${{ github.ref }}
           outputDirectory: /tmp/artifacts
-          urls: 'https://www.foo.software,https://www.foo.software/contact'
-          
+          urls: "https://www.fast.design/"
       - name: Upload artifacts
         uses: actions/upload-artifact@master
         with:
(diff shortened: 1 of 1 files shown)


awentzel:
@Ibrahimmaga, What is the status on this PR?

Ibrahimmaga:
> @Ibrahimmaga, What is the status on this PR?

Let me run the CI and see what result we are getting

Ibrahimmaga:
> @Ibrahimmaga, What is the status on this PR?

I like the tool, but the performance number doesn't add up when comparing to the lighthouse. sounds misleading. The only thing that matches is best practice and accessibility
image
image

awentzel:
@Ibrahimmaga,

> I like the tool, but the performance number doesn't add up when comparing to the lighthouse. sounds misleading. The only thing that matches is best practice and accessibility

The way I'm interpreting this behavior is by design.  I imagine there are different metrics to test on mobile vs desktop. For example, in mobile-first scenarios, much of the UI will not be generated to improve performance and simplify the user experience.  In a desktop scenario, there is much more real estate to work with and so the UI often encompasses everything..  There are differences by design.

Have you filed an issue on the GitHub Action project to get an explanation and understand what's going on and why this behavior is different?  I assume there is documentation or a way to reach the author to get answers?

awentzel:
Status update.  We ran into an issue with the original GH Action we were to use.  We are now evaluating two other Actions to accomplish the same thing. We will continue to use this PR to test and determine which will work best with out environment.

Ibrahimmaga:
> Status update. We ran into an issue with the original GH Action we were to use. We are now evaluating two other Actions to accomplish the same thing. We will continue to use this PR to test and determine which will work best with our environment.

When using another Github action, Numbers are a little better except the performance that is in the 20 %. when using lighthouse on google chrome, the performance is 89%.
image

awentzel (commented):
This is better than what we offer today so It can probably be approved after adding the other sites we want to analyze.  

* www.fast.design
* explore.fast.design
* color.fast.design

Also, if there is a way to print the output summary to the console so it's apparent in the PR itself in addition to creating the link to click to review that would be great.

awentzel:
> This is better than what we offer today so It can probably be approved after adding the other sites we want to analyze.
> 
> * [www.fast.design](http://www.fast.design)
> * explore.fast.design
> * color.fast.design
> 
> Also, if there is a way to print the output summary to the console so it's apparent in the PR itself in addition to creating the link to click to review that would be great.

The "other" sites is now a moot point, we primarily care about www.  I propose approving and pushing this in, then improving latest as needed.  Additionally, there are conversations about migrating from subdomain structure to subdirectories to reduce maintenance effort and cost.

awentzel (approved):

EisenbergEffect:
Closing this for now. We can revisit again later.

PR #5193, opened 2021-09-16
calculate monitor refresh rate in horizontal-scroll

radium-v:
# Pull Request

## 📖 Description

- Adds a method to calculate the screen refresh rate / target frames per second for the horizontal scroller component.
- Moves `ResizeObserver` and `IntersectionService` files to `src/utilities` in `fast-foundation`
- Adds parts to internal components of the horizontal scroller template

## 👩‍💻 Reviewer Notes

Firefox may reduce the precision of timestamps to resist device fingerprinting: https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp#reduced_time_precision

## 📑 Test Plan

Tested in Chrome, Edge, and Firefox at 60hz, 120hz, 165hz, and 240hz.

### General

### Component-specific


packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.template.ts
```diff
@@ -45,7 +46,7 @@ export const horizontalScrollTemplate: FoundationElementTemplate<
                         part="scroll-prev"
                         ${ref("previousFlipperContainer")}
                     >
-                        <div class="scroll-action">
+                        <div class="scroll-action" part="scroll-action-previous">
```
chrisdholt, line 49: This is great - thank you!

chrisdholt (approved):
Thanks for this @radium-v!

chrisdholt:
Also, I pulled this down, built, smoked - definitely smoother on my mac :)

eljefe223 (approved):
Smooth like butter, Thanks for doing this John!

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -346,6 +337,8 @@ export class HorizontalScroll extends FoundationElement {
      * @public
      */
     public scrollToPrevious(): void {
+        this.getRefreshRate();
```
robarbms, line 340: Does refresh rate need to be recalculated every time the scrolls happen?

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -495,9 +541,7 @@ export class HorizontalScroll extends FoundationElement {
      * @internal
      */
     private getEasedFactor(easing: ScrollEasing, progress: number): number {
-        if (progress > 1) {
-            progress = 1;
-        }
+        progress = progress > 1 ? 1 : progress;
```
robarbms, line 544: This works but if you wanted it even more explicit.

`progress = Math.min(1, progress)`

robarbms (approved):

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
radium-v: This was the only way I could think of to ensure that Firefox returns a value that's as close to the actual refresh rate as it will allow. Chrome and Edge have no issues and always return the correct number.

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
radium-v: Calling `Math.min` is slower than an inline ternary.

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -371,6 +364,8 @@ export class HorizontalScroll extends FoundationElement {
      * @public
      */
     public scrollToNext(): void {
+        this.getRefreshRate();
```
nicholasrice, line 367: Super clever! I'm curious how this part works though. If `this.getRefereshRate()` is effectively async because of rAF, won't fps have the potential to be stale here? Are we just counting on the fact that it will get accurate eventually?

nicholasrice (approved):

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
radium-v: I have a fix incoming that will ensure that `this.getRefreshRate()` completes first, which should fix the stale value problem.

Issue #4925, opened 2021-06-30
feat: add a selectable attribute to/in TreeItem


m4thieulavoie:
[More detail here](https://discord.com/channels/449251231706251264/482296787021856768/859796174130642984)

# 🙋 Feature Request

My team would like to be able to provide a custom non-selectable item as a child of the TreeItem. In our specific context, we want to use an Empty State component that is custom to our design system.

## 🤔 Expected Behavior

A new attribute allows for us to disable the selection of a given TreeItem

Here's an example of what we want to be possible
```
TreeView
  TreeItem
    TreeItem
    TreeItem
    TreeItem
  TreeItem
    EmptyState <----- Should not be toggleable or interactive
```

## 😯 Current Behavior

Any kind of TreeItem is selectable no matter what, with the exception of being `disabled`

chrisdholt:
Thanks @m4thieulavoie for the feature request. I think it would be great to hear a bit more about your specific scenario. Our current tree view and therefore tree items are built to follow the specification outlined in the [ARIA practices documentation](https://w3c.github.io/aria-practices/#TreeView), though I do see that the `treeitem` role inherits `aria-disabled` - so it seems like there should be something called out in the spec for this.

We may need to do some investigation to see how and if we can light something up like this up in an accessible way.

m4thieulavoie:
@chrisdholt what my team wants to achieve is have a TreeView for some "groups" of contact. Each item on the root level is effectively a group that expands to contacts under that group. However, there are cases where the contact list is empty and they'd like to show an EmptyState component (effectively an icon/heading/paragraph wrapped in a web component) under the header.

My initial hacking session with the TreeView and TreeItem from FAST made me realize it is absolutely possible to achieve. The only blocker left is the fact that the TreeItem can be selected, which adds styling and triggers an event unnecessarily. Hence this me opening up this issue :) I like the idea @scomea brought (see above discord link) to add a `isSelectable` attribute to disable this selection behavior at will.

I hope it was a bit clearer!

chrisdholt:
> @chrisdholt what my team wants to achieve is have a TreeView for some "groups" of contact. Each item on the root level is effectively a group that expands to contacts under that group. However, there are cases where the contact list is empty and they'd like to show an EmptyState component (effectively an icon/heading/paragraph wrapped in a web component) under the header.
> 
[...]

PR #5194, opened 2021-09-16
fix: ensure tree item selection is only managed in tree view

eljefe223:
# Pull Request

## 📖 Description
This PR achieves the following:

1. Organizes selection and focus logic to be managed by the `TreeView`
2. Allows Disabled `TreeItem` to be focusable
3. Fixes Keyboarding bugs

### 🎫 Issues

Closes: #4925 

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
@@ -179,19 +179,6 @@ describe("TreeItem", () => {
         await disconnect();
     });
 
-    it("should NOT set a tabindex when disabled is `true`", async () => {
-        const { element, connect, disconnect } = await setup();
-
-        element.disabled = true;
-
-        await connect();
-
-        expect(element.hasAttribute("tabindex")).to.equal(false);
-        expect(element.getAttribute("tabindex")).to.equal(null);
-
-        await disconnect();
-    });
-
     it("should set a tabindex when `focusable` is true", async () => {
         const { element, connect, disconnect } = await setup();
 
@@ -347,32 +334,6 @@ describe("TreeItem", () => {
             await DOM.nextUpdate();
 
             expect(wasClicked).to.equal(true);
-
-            await disconnect();
-        });
-
-        it("should toggle the selected state when the component is clicked", async () => {
-            const { element, connect, disconnect } = await setup();
-            const nestedItem = document.createElement("fast-tree-item");
-
-            element.appendChild(nestedItem);
-
-            await connect();
-            await DOM.nextUpdate();
-
-            element.click();
-
-            await DOM.nextUpdate();
-
-            expect(element.selected).to.equal(true);
-            expect(element.getAttribute("aria-selected")).to.equal("true");
-
-            element.click();
-
-            await DOM.nextUpdate();
-
-            expect(element.selected).to.equal(false);
-            expect(element.getAttribute("aria-selected")).to.equal("false");
 
             await disconnect();
         });
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
@@ -338,7 +338,7 @@ describe("TreeItem", () => {
             await disconnect();
         });
 
-        it("should NOT toggle the selected state when the element is clicked when disabled", async () => {
+        it("should NOT set selected state when the element is clicked when disabled", async () => {
             const { element, connect, disconnect } = await setup();
             const nestedItem = document.createElement("fast-tree-item");
 
--- a/packages/web-components/fast-foundation/src/tree-view/tree-view.ts
+++ b/packages/web-components/fast-foundation/src/tree-view/tree-view.ts
@@ -83,7 +83,7 @@ export class TreeView extends FoundationElement {
             !this.contains(relatedTarget)
         ) {
             const treeView = this as HTMLElement;
-            if (this.currentFocused instanceof TreeItem) {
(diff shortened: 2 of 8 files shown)


packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -47,20 +57,36 @@ export class TreeView extends FoundationElement {
...
     private treeItems: HTMLElement[];
 
     public handleBlur = (e: FocusEvent): void => {
         const { relatedTarget, target } = e;
+        if (
+            (relatedTarget === null || relatedTarget instanceof HTMLElement) &&
+            target instanceof HTMLElement &&
+            !this.contains(relatedTarget)
+        ) {
+            const treeView = this as HTMLElement;
```
nicholasrice, line 71: I don't think you need to store `this` in a const - just call `this.setAttribute("tabindex", "0")`

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -47,20 +57,36 @@ export class TreeView extends FoundationElement {
...
+
+    public handleFocus = (e: FocusEvent): void => {
+        const { relatedTarget, target } = e;
 
-        /**
-         * Clean up previously focused item's tabindex if we've moved to another item in the tree
-         */
         if (
-            relatedTarget instanceof HTMLElement &&
+            (relatedTarget === null || relatedTarget instanceof HTMLElement) &&
```
nicholasrice, line 80: why are you moving forward with the `.contains()` call if `relatedTarget` is null? Shouldn't this check for `!==` and skip the `.contains()` check if it is?

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -88,24 +114,91 @@ export class TreeView extends FoundationElement {
...
+                // In single-select trees where selection does not follow focus (see note below),
+                // the default action is typically to select the focused node.
+                this.handleSelected(e.target as HTMLElement);
+                break;
             default:
                 return true;
         }
     };
 
+    private focusNextNode(delta: number, item: HTMLElement): void {
```
nicholasrice, line 169: this method doesn't seem to do anything if `item` is an `HTMLElement`. Should you just expect a `TreeItem` instead?

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -123,41 +216,49 @@ export class TreeView extends FoundationElement {
         }
     };
 
+    private handleSelected(item: HTMLElement): void {
```
nicholasrice, line 219: same question here - maybe just expect a TreeItem?

packages/web-components/fast-foundation/src/tree-item/tree-item.spec.ts
```diff
@@ -179,19 +179,6 @@ describe("TreeItem", () => {
...
-        element.disabled = true;
-
-        await connect();
-
-        expect(element.hasAttribute("tabindex")).to.equal(false);
-        expect(element.getAttribute("tabindex")).to.equal(null);
-
-        await disconnect();
-    });
-
```
chrisdholt, line 194: @eljefe223 I think we'll want to have comparable tests in Tree View to ensure we these are still working. While I get that the test may not be 1:1, ideally we have checks to ensure feature parity is working.

eljefe223: Right, i think we can add a test in tree view to ensure the tree item sets selected, but this specific test is no longer relevant since `disabled` items are now focusable.

chrisdholt: Thanks for the reminder (insert faceplam emoji 🙂 )

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -88,24 +111,89 @@ export class TreeView extends FoundationElement {
...
                 if (this.treeItems && this.treeItems.length) {
                     TreeItem.focusItem(this.treeItems[this.treeItems.length - 1]);
+                    this.treeItems[this.treeItems.length - 1].setAttribute(
+                        "tabindex",
+                        "0"
+                    );
+                }
+                break;
+            case keyArrowLeft:
+                e.preventDefault();
```
scomea, line 127: You don't need the preventDefaults here as the template binding does that by default.

scomea (approved):

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -123,41 +207,47 @@ export class TreeView extends FoundationElement {
...
+                this.currentSelected.selected = false;
+                this.currentSelected.setAttribute("tabindex", "-1");
+            }
+            if (!this.currentSelected) {
+                this.slottedTreeItems.forEach((item: HTMLElement) => {
+                    if (item instanceof TreeItem) {
+                        item.setAttribute("tabindex", "-1");
+                    }
+                });
+            }
```
chrisdholt, line 223: When testing this, I found one subtle issue. If I select a node, and then I move to a *disabled* node and try to select it - the disabled node correctly **does NOT** receive selection, but the previous node *loses* selection. It seems like in the case of a disabled node, nothing should change from a selection state (the old node should retain selection).

eljefe223: Should be fixed now, the tree item was preventing the selection event from firing and not entering the code path.

nicholasrice (approved):

chrisdholt (approved):

PR #5204, opened 2021-09-20
feat(di): add registerWithContext and use in design system

EisenbergEffect:
# Pull Request

## 📖 Description

This PR adds a new `registerWithContext` method to the `Container` interface. It allows the caller to pass in a `context` object which will be passed to every registry that is provided for the call. The new API enables a general way to provide context to registries without having to _permanently_ put a _temporary_ context into the DI container and force the registries to each _individually_ resolve it. This can be seen as both a performance optimization and a bug fix for key scenarios.

With this API in place, it was then used inside the `DesignSystem#register` method to provide a `DesignSystemRegistrationContext` to every component registry passed to the `DesignSystem`. This allows the context to be more correctly provided only within the context of the registration call (not permanently). It also improves startup performance by removing the need for every component registry to separately resolve the context during registration.

This does present a small breaking change. The `const` `DesignSystemRegistrationContext` was removed and now only the interface remains and it cannot be resolved through DI. The constant was used internally by `FrameworkElement.compose` to resolve the context from the dependency injection container. While it is technically possible that someone used this directly, it is not likely as this was intended for internal use and was not documented (an oversight that it was ever marked public). I don't know why anyone would use this directly either so I think this is safe to change in a minor release.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

The primary change is the addition of the `registerWithContext` method and its use inside of the DesignSystem's `register` method.

## 📑 Test Plan

Existing tests were fixed and additional tests were added to cover the `registerWithContext` API.

### General

## ⏭ Next Steps

None at this time.


EisenbergEffect:
Note, build errors should resolve once this PR is merged https://github.com/microsoft/fast/pull/5202

chrisdholt (approved):

Issue #5196, opened 2021-09-16
fix: allowing decimal values in fast number field


ben-girardet:
# 🐛 Bug Report

Currently, the `<fast-number-field>` doesn't handle decimal values properly.

- [x] With a decimal step, the up and down buttons work properly
- [ ] The field doesn't allow for the `.` symbol to be entered
- [ ] The `inputmode` is set as `numeric` which is intended for integer only values

## 🤔 Expected Behavior

This component should properly handled both integer and decimal values.

## 💁 Possible Solution

I'm happy to submit a PR to fix this issue, but I would like to discuss the right course of action first. I suggest:

1. Adding an `@attr` for `inputmode` allowing for `numeric` and `decimal` values.
2. I'm wondering if the `type` of the input should be `number`. Problem is: it will display the native up/down buttons but they can be hidden with CSS.

Would be happy to get some feedback before to jump on the PR.

atmgrifter00:
Personally, I feel that there shouldn't be a strict distinction between what kind of numeric the number field supports (integer vs. decimal), and instead it should allow a user to _always_ enter a decimal value (this works just fine on the standard HTML input element set to 'number' type).

Instead, if the goal is to restrict what the user can ultimately see in the input field, it would be nice if a user could specify formatting options (e.g. number of digits of precision, scientific notation, etc...).

PR #5205, opened 2021-09-20
fix: allowing decimal values in fast number field

robarbms:
# Pull Request

## 📖 Description

parseFloat(value) was preventing entering a float value. "1." would parse to "1" and "." would empty out the value.

### 🎫 Issues

https://github.com/microsoft/fast/issues/5196

## 👩‍💻 Reviewer Notes

I'm doing a number to number compare to keep 1. == 1 instead of "1." == 1. I've also added checks for values starting with a decimal, plus or minus sign so that they don't get parsed out before a number is entered.

## 📑 Test Plan

Created tests to ensure that floats, negative and positive values could be manually entered.

### General

### Component-specific

## ⏭ Next Steps

none


packages/web-components/fast-foundation/src/number-field/number-field.spec.ts
```diff
@@ -831,6 +831,86 @@ describe("NumberField", () => {
...
+
+            element.setAttribute("value", "11a");
+
+            expect(element.value).to.equal("11");
+        });
+
+        it("should allow float number entry", async () => {
+            const { element, connect, disconnect } = await setup();
+
+            await connect();
```
nicholasrice, line 858: I'm noticing that this implementation does not have behavioral parity with the native `<input type="number" />` when it comes to these tests: https://stackblitz.com/edit/web-platform-sre4og?file=index.html Some of the native behaviors seem a little odd to me (such as "+3" attribute resulting in empty string whereas typing it as a user results in the value '3' ) but others seem intuitive to me ("-", ".", and "+" all emit empty string values); What was the impetus for emitting those three as stateful values?

robarbms: It's not strictly necessary. I was adding for the negative scenario and added positive for completeness.

As for entering "-" and ".", it might make sense to parseFloat the result to the proxy value. This will allow the user to start a value with "-" or "." but the stored proxy will be empty string in the case that the form gets submitted.

nicholasrice: I don't think that approach will work. The proxy is only used for form-submission in cases where `ElementInternals` is not supported, so this will lead to a mis-match in behavior between different browsers. It would also make `FASTNumberField.value` inconsistent with the value submitted with the form for those browsers, which I think would be unintuitive for authors.

Can we store the typed / attribute / displayed value in a different property and set the stateful value as the result of parseFloat of that value (or empty string if the result is `NaN`)?

nicholasrice: I'm not suggesting we don't show what was typed to the user. I'm wondering if we can divorce what is *displayed* from the `FASTNumberField.value`. For example:

1. The user types "." so that "." is displayed
2. -> FASTNumberField.value is "" because "." cannot be successfully parsed to a number
3. The user types "3" so that ".3" is displayed
4. -> FASTNumberField.value is now ".3" because ".3" can successfully be parsed to a number
5. The user moves the cursor and deletes the "." so that "3" is displayed
6. -> FASTNumberField.value is now "3" because "3" can successfully be parsed to a number

The above seems to me to be how the native control functions.

atmgrifter00: FWIW, I agree with @nicholasrice on how the value updates according to a user's input. Note that this would also allow features such as numeric formatting, where the displayed value might be "1M", and the FASTNumberField.value is 1000000 (and the user could have entered either "1M" or 1000000 for the same result).

robarbms: I've updated it to use a displayText for the input field. This allows for typing out negative values, decimal values and even exponentials.

nicholasrice (approved):

chrisdholt (approved):

Issue #5083, opened 2021-08-17
rfc: binding modes, interpolating non-strings in templates, and event capture


EisenbergEffect:
# 💬 RFC

While our templating engine has been very open for extension through directives, up until now, our in-template binding capabilities have not been extensible.

## 🔦 Context

A number of scenarios around bindings have arisen over the previous months, all centered on lack of extensibility in this part of the system. Examples include:

1. One Time Bindings - Often times there's a desire to interpolate a model's value into HTML but with no desire to observe the property for updates. Due to this shortcoming, these scenarios sill had to create binding observers, resulting in unnecessary use of memory and slowed startup time.
2. Static Template Values - The `html` tagged template literal could not handle interpolation of literal objects, numbers, booleans, etc. This meant that all such scenarios forced the developer to have to write an expression for the static value `x => staticValue`. Unfortunately, taking this approach also caused the engine to setup a binding observer where one was not needed.
3. Static Template Composition - The `html` tagged template literal enabled composing other templates into the html content. However, this suffers from the same problem as the static template values, forcing all such compositions to setup observers when they are not needed.
4. Event Capture Mode - The event binding syntax did not enable the `useCapture` option or any customization of events.
5. TwoWay Bindings and Custom Handling - There was no easy way to extend the system to enable two-way binding for data-entry scenarios, or any other custom binding behavior.

It should be noted that numbers 1, 4, and 5 are technically possible today but require massive duplication of all the code in the binding system and making alterations based on the specific scenarios. Not only is this very wasteful, but it requires in-depth knowledge of the templating directive and behavior systems as well as the observation system.

## 💻 Proposal

I'm proposing that we rework the internals of the binding system and introduce the concept of `BindingMode`. A `BindingMode` is nothing more than a map of *template contexts* to `HTMLBindingBehavior` constructors. Here is some code to show what that would look like:

```ts
export type HTMLBindingBehavior = {
    new (directive: HTMLBindingDirective, target: Node): Behavior;
};

export interface BindingMode {
  attribute?: HTMLBindingBehavior;
  booleanAttribute?: HTMLBindingBehavior;
  property?: HTMLBindingBehavior;
  content?: HTMLBindingBehavior;
  class?: HTMLBindingBehavior;
  event?: HTMLBindingBehavior;
}
```

[...]

PR #5207, opened 2021-09-20
Draft: events that detach on disconnectedCallback

robarbms:
# Pull Request

## 📖 Description

There may be potential for lots of events not getting removed when an element is removed from the page. This allows FASTElements to track events attached to itself or by itself. It will automatically detach these when disconnectedCallback is called. It also exposes a list of events that it's tracking which can be searched or modified.

### 🎫 Issues

NA

## 👩‍💻 Reviewer Notes

Concept for better event management by FASTElements.

### General

### Component-specific

## ⏭ Next Steps

Get feedback


EisenbergEffect:
I think a better approach might be to use our Behaviors. Since we've got behaviors for events already, we could probably create a helper over top of that and reduce the code quite a bit. I think we should consider this after the cleanup of the Binding Behaviors https://github.com/microsoft/fast/issues/5083 planned for 2.0. With that in place we could do this much more efficiently. For example, I think it would require less code and we wouldn't need to add additional lifecycle methods.

(We're trying to keep fast-element as close to 10k as possible, so I'm wary of adding completely new things for just one scenario. That's why I lean more towards seeing if we can do this with behaviors, so we can leverage that entire system.)

robarbms:
> I think a better approach might be to use our Behaviors. Since we've got behaviors for events already, we could probably create a helper over top of that and reduce the code quite a bit. I think we should consider this after the cleanup of the Binding Behaviors #5083 planned for 2.0. With that in place we could do this much more efficiently. For example, I think it would require less code and we wouldn't need to add additional lifecycle methods.
> 
> (We're trying to keep fast-element as close to 10k as possible, so I'm wary of adding completely new things for just one scenario. That's why I lean more towards seeing if we can do this with behaviors, so we can leverage that entire system.)

That makes sense. And the size had me concerned as well. I looked for ways to reduce the size but this was as small as I could get it and make it worthwhile. I'll look at what it might look like using behaviors.

PR #5217, opened 2021-09-22
fix: add aria-disabled to an expanded accordion-item in single mode

khamudom:
# Pull Request

## 📖 Description

Added the functionality to automatically add and remove `aria-disabled` attribute to the accordion-item in the single-mode scenario.
 
In the case when an accordion is in single-mode, only one accordion-item is opened at a time, the active accordion-item now has `aria-disabled` attribute set on it. This is added because once the accordion item is opened, it does not have the capability to close itself. According to the W3C spec, the item should have aria-disabled set to true.

Here is an example of single-mode accordion from the W3C website.
https://w3c.github.io/aria-practices/examples/accordion/accordion.html#:~:text=at%20a%20time.-,Example,-Open%20In%20CodePen

### 🎫 Issues

fixes: #4160

### General

### Component-specific


packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -70,12 +70,28 @@ export class Accordion extends FoundationElement {
         this.$emit("change");
     };
 
+    private findExpandedItem(): AccordionItem | null {
+        for (let item: number = 0; item < this.accordionItems.length; item++) {
+            if (this.accordionItems[item].hasAttribute("expanded")) {
```
scomea, line 75: Not only should it have the attribute, but you should verify it is 'true'

packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -70,12 +70,28 @@ export class Accordion extends FoundationElement {
...
+                return this.accordionItems[item] as AccordionItem;
+            }
+        }
+        return null;
+    }
+
     private setItems = (): void => {
+        if (this.accordionItems.length === 0) {
```
scomea, line 83: We should also be zeroing out the values set by this function to match the "no items" state.  For example, this.activeid and this.activeItemIndex.

packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -70,12 +70,28 @@ export class Accordion extends FoundationElement {
...
+            return;
+        }
         this.accordionIds = this.getItemIds();
```
scomea, line 86: You can get rid of the `this.accordionIds = this.getItemIds();` line in the `accordionItemsChanged` function as that function immediately calls this one, so we're getting item id's more than we need to.  Not specifically part of your change but since you're in the neighborhood...

khamudom: removed

packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -108,11 +123,17 @@ export class Accordion extends FoundationElement {
 
     private activeItemChange = (event): void => {
         const selectedItem = event.target as HTMLElement;
+        this.activeid = event.target.getAttribute("id");
         if (this.isSingleExpandMode()) {
             this.resetItems();
             event.target.expanded = true;
+            event.target.setAttribute("aria-disabled", "true");
```
eljefe223, line 130: We are storing the `event.target` above as `selectedItem` you should be using it here

chrisdholt:
@khamudom curious about the concept of a *disabled* panel outside this scenario. Is there a realistic scenario where a panel is disabled and appears so? In this case (single expand) this seems to solve the instance where one *must* be expanded and therefore it becomes "disabled" from interaction; however, I'm curious about the larger disabled story amidst this change.

khamudom:
> @khamudom curious about the concept of a _disabled_ panel outside this scenario. Is there a realistic scenario where a panel is disabled and appears so? In this case (single expand) this seems to solve the instance where one _must_ be expanded and therefore it becomes "disabled" from interaction; however, I'm curious about the larger disabled story amidst this change.

Looking at the W3C spec, I did not see any description regarding the disabled state on the overall component. Disabled was only mentioned in the case of an opened panel in single mode, where the panel is not disabled, but the accordion item is. That is why the spec suggests we add `aria-disabled="true"` on the accordion item. I am wondering if the overall disabled was not originally added or addressed on the component because it was not mentioned in the spec. @eljefe223  am I assuming this correctly?

packages/web-components/fast-foundation/src/accordion/accordion.ts
```diff
@@ -57,25 +57,40 @@ export class Accordion extends FoundationElement {
...
     private setItems = (): void => {
+        if (this.accordionItems.length === 0) {
+            return;
+        }
         this.accordionIds = this.getItemIds();
         this.accordionItems.forEach((item: HTMLElement, index: number) => {
             if (item instanceof AccordionItem) {
                 item.addEventListener("change", this.activeItemChange);
                 if (this.isSingleExpandMode()) {
+                    const expandedItem: AccordionItem | null =
```
nicholasrice, line 90: Do you need to do this work for every item in accordionItems? Seems like it really only needs to happen once since it doesn't use the item or the index from the loop

packages/web-components/fast-foundation/src/accordion/accordion.ts
khamudom: updated

packages/web-components/fast-foundation/src/accordion/accordion.ts
khamudom: I moved the expandedItem check out of the loop

nicholasrice (approved):

PR #5221, opened 2021-09-23
Updated the color blocks and swatches to web components

bheston:
# Pull Request

## 📖 Description

Starting to migrate the Color Explorer to web components.

## 👩‍💻 Reviewer Notes

Color Explorer only.

## 📑 Test Plan

Make sure color explorer functions as it does currently.

### General


sites/fast-color-explorer/app/components/color-block.ts
```diff
@@ -0,0 +1,439 @@
...
+    name: "app-color-block",
+    template,
+    styles,
+})
+export class AppColorBlock extends FASTElement {
+    @attr index: number;
+
+    @attr component: ComponentTypes;
+
+    @attr({ attribute: "background-color" }) backgroundColor: string;
```
chrisdholt, line 436: should this be `fill-color`?

sites/fast-color-explorer/app/components/swatch.ts
```diff
@@ -0,0 +1,200 @@
...
+        width: 100%;
+        padding: 4px 0;
+        box-sizing: border-box;
+        color: ${neutralForegroundHint};
+        font-size: 12px;
+        grid-column-gap: 16px;
+        justify-items: start;
+    }
+    :host(.fill) {
+    }
```
chrisdholt, line 44: can this empty block be removed?

sites/fast-color-explorer/app/components/swatch.ts
```diff
@@ -0,0 +1,200 @@
...
+    :host(.foreground) .icon {
+        border: 1px solid black;
+    }
```
chrisdholt, line 47: Just a note that if these are still rendering into React here *is* a chance that the classes could be clobbered. I wonder if we should tie these to the attributes if they exist...

packages/web-components/fast-components/src/color/utilities/color-constants.ts
```diff
@@ -13,11 +13,9 @@ export const black = SwatchRGB.create(0, 0, 0);
 /**
  * @internal
  */
-export const middleGrey = SwatchRGB.create(0.5, 0.5, 0.5);
+export const middleGrey = SwatchRGB.from(parseColorHexRGB("#808080")!);
```
nicholasrice, line 16: does this change result in a different gray?

bheston: It does change things ever so slightly. It parses as 0.501... iirc, which affects some of the calculations. I believe we made this change in another place before, but this makes the ramp exactly the same the first time and when you go back and click on the 808080 swatch.

sites/fast-color-explorer/app/components/color-block.ts
bheston: It does apply to fill-color, but I was keeping it separate from that. I went with 'color' since this is the 'color-block'.

sites/fast-color-explorer/app/components/swatch.ts
bheston: These are only rendeded within another web component. Does that still apply if they are not direct children of a React component?

sites/fast-color-explorer/app/components/swatch.ts
bheston: Yeah, I can remove this and the one below. Mostly left them there to show they existed, but this is not a component I expect to be reused elsewhere.

sites/fast-color-explorer/app/components/swatch.ts
chrisdholt: You should be fine - I mentioned it primarily because I wasn't sure if this was rendering directly in React or in another WC

bheston: Good to know. The plan is to get rid of all React entirely. Soon. :)

packages/web-components/fast-components/src/color/utilities/color-constants.ts
nicholasrice: bummer, okay

nicholasrice (approved):

chrisdholt (approved):

PR #5223, opened 2021-09-23
fix: when orientation changes, the styles are updated properly on tab chil…

marjonlynch:
# Pull Request

## 📖 Description

When the orientation attribute changes we want the styles to be updated for the child tabs as it is necessary to maintain the proper grid css settings for vertical or horizontal mode. In the process of adding orientation changed functionality another issue was discovered whereby setting tab.style[gridProperty] was actually translated to grid-area and caused the tabs to be misaligned so that is also fixed in this PR by explicitly emptying grid-colum and grid-row before setting the new value.

before:
Screen Shot 2021-09-24 at 8 55 12 AM

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/tabs/tabs.ts
+++ b/packages/web-components/fast-foundation/src/tabs/tabs.ts
@@ -42,7 +42,16 @@ export class Tabs extends FoundationElement {
      */
     @attr
     public orientation: TabsOrientation = TabsOrientation.horizontal;
-
+    /**
+     * @internal
+     */
+    public orientationChanged(): void {
+        if (this.$fastController.isConnected) {
+            this.setTabs();
+            this.setTabPanels();
+            this.handleActiveIndicatorPosition();
+        }
+    }
     /**
      * The id of the active tab
      *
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-foundation/src/tabs/tabs.ts
```diff
@@ -162,7 +171,11 @@ export class Tabs extends FoundationElement {
     }
 
     private setTabs = (): void => {
-        const gridProperty: string = this.isHorizontal() ? "gridColumn" : "gridRow";
+        const gridHorizontalProperty: string = "grid-column";
+        const gridVerticalProperty: string = "grid-row";
```
robarbms, line 175: Shouldn't these be "gridColumn" and "gridRow" since they are being used on the JavaScript style object?

robarbms (approved):

packages/web-components/fast-foundation/src/tabs/tabs.ts
marjonlynch: this is an interesting question, this does work, but maybe it would be better to use it the way you suggest. I will test it out and change to that if it works the same

marjonlynch: it did work so I Updated it, thank Rob

nicholasrice (approved):

janechu (approved):

packages/web-components/fast-foundation/src/tabs/tabs.ts
```diff
@@ -162,7 +171,11 @@ export class Tabs extends FoundationElement {
     }
 
     private setTabs = (): void => {
-        const gridProperty: string = this.isHorizontal() ? "gridColumn" : "gridRow";
+        const gridHorizontalProperty: string = "gridColumn";
```
awentzel, line 174: Could these string values be good candidates for enum?

marjonlynch: This is a good point but considering only two values here, I don't anticipating that changing and these are the strings that were originally here I feel like it's best to keep it at strings for now. If there was a utility enum with all the possible values available there that would be awesome, another PR though I think.

PR #5225, opened 2021-09-23
fix: add manual focus on components with delegatesFocus to support firefox

khamudom:
# Pull Request

## 📖 Description

Added a function to manually set focus on the inner element when delegatesFocus is undefined.

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-components-ad8eb6c8-9106-4890-ad00-49f44659b24a.json
+++ b/change/@microsoft-fast-components-ad8eb6c8-9106-4890-ad00-49f44659b24a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "add function to support programmatic focus for firefox and added test",
+  "packageName": "@microsoft/fast-components",
+  "email": "khamu@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-components-ad8eb6c8-9106-4890-ad00-49f44659b24a.json
+++ b/change/@microsoft-fast-components-ad8eb6c8-9106-4890-ad00-49f44659b24a.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "add function to support programmatic focus for firefox and added test",
+  "comment": "added fallback function for browsers that do not support delegatesFocus",
   "packageName": "@microsoft/fast-components",
   "email": "khamu@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-fast-foundation-3ae546c3-c283-4503-a384-fa5da550a2a9.json
+++ b/change/@microsoft-fast-foundation-3ae546c3-c283-4503-a384-fa5da550a2a9.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "add function to support programmatic focus for firefox and added test",
+  "comment": "added fallback function for browsers that do not support delegatesFocus",
   "packageName": "@microsoft/fast-foundation",
   "email": "khamu@microsoft.com",
   "dependentChangeType": "patch"
--- a/change/@microsoft-fast-foundation-3ae546c3-c283-4503-a384-fa5da550a2a9.json
+++ b/change/@microsoft-fast-foundation-3ae546c3-c283-4503-a384-fa5da550a2a9.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "add function to support programmatic focus for firefox and added test",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "khamu@microsoft.com",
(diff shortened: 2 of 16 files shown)


change/@microsoft-fast-components-ad8eb6c8-9106-4890-ad00-49f44659b24a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "add function to support programmatic focus for firefox and added test",
```
chrisdholt, line 3: I think this could be more clear - this is only necessary where delegatesFocus isn't supported which is technically both Firefox and iOS safari. We don't need to call out both by name, but I think explaining that this relates to mitigating the lack of delegatesFocus would help.

change/@microsoft-fast-foundation-3ae546c3-c283-4503-a384-fa5da550a2a9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "add function to support programmatic focus for firefox and added test",
```
chrisdholt, line 3: Same note above

packages/web-components/fast-components/src/avatar/avatar.pw.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    it("receive focus when focused programatically", async function () {
+        const element = await this.page.waitForSelector("fast-avatar");
+
+        await this.page.evaluateHandle(element => element.focus(), element)
+
+        expect(await this.page.evaluate(
+            () => document.activeElement?.id
+        )).to.equal(await element.evaluate(node => node.id));
```
chrisdholt, line 50: Does this test that activeElement is the internal anchor or is it the root element?

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
```diff
@@ -25,6 +25,7 @@ export const avatarTemplate: (
             part="link"
             href="${x => (x.link ? x.link : void 0)}"
             style="${x => (x.color ? `color: var(--avatar-color-${x.color});` : void 0)}"
+            ${ref("control")}
```
chrisdholt, line 28: I'm wondering if this should be `link` to match with the "part". I have a proposal hopefully coming this week regarding adding refs and aligning to the named parts.

chrisdholt (changes requested):
Looks like we're potentially missing a few where this could be an issue - some of these may not be, but we should test.

- Combobox
- Toolbar
- Text Area

packages/web-components/fast-foundation/src/avatar/avatar.template.ts
khamudom: I agree, but it looks like there is an attr that is already called `link`
image

change/@microsoft-fast-components-ad8eb6c8-9106-4890-ad00-49f44659b24a.json
khamudom: updated

change/@microsoft-fast-foundation-3ae546c3-c283-4503-a384-fa5da550a2a9.json
khamudom: updated

packages/web-components/fast-components/src/avatar/avatar.pw.spec.ts
khamudom: activeElement is the root element

packages/web-components/fast-foundation/src/combobox/combobox.ts
```diff
@@ -610,6 +612,24 @@ export class Combobox extends FormAssociatedCombobox {
...
+        if (
+            window.ShadowRoot &&
+            !window.ShadowRoot.prototype.hasOwnProperty("delegatesFocus") &&
+            this.$fastController.definition.shadowOptions?.delegatesFocus
+        ) {
+            this.focus = () => {
+                this.control.focus();
+            };
+        }
+    };
```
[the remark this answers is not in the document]
khamudom: The tabindex in the template of combobox is preventing focus to be set on the input.
I also manually tested the other components and they look good.

scomea:
Is this still required?   Everybody seems to support this now.

https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus

EisenbergEffect:
@chrisdholt Can we close this?

chrisdholt:
> @chrisdholt Can we close this?

I haven't seen any traction or issues arising out of this, I think we can close this out.

PR #5228, opened 2021-09-24
update: beachball config

Ibrahimmaga:
## 📖 Description
Beachball should prevent major bumping when running beachball change. This PR is to change the current beachball configuration to allows beachball to have only 3 configurations options "none, patch, and minor" as shown in this image

image

### 🎫 Issues
The issue is, beachball was not reading the old configuration and the developer can bump to the major version without documentation. This Pr is just removing that option.

### General

### Component-specific

- [] I have updated the [definition file](https://github.com/Microsoft/fast/blob/master/packages/web-components/fast-components/CONTRIBUTING.md#definition)


beachball.config.js
```diff
@@ -1,2 +1,19 @@
-require("./build/releasing/ts-node-register");
-module.exports = require("./build/releasing/index");
+/*require("./build/releasing/ts-node-register");
```
awentzel, line 1: You can remove the code comments on rows 1 &2.

awentzel (approved):
Approving, but, recommending to manually publish as soon as we can to perform some regression testing and ensure this continues to work on schedule.

beachball.config.js
```diff
@@ -1,2 +1,16 @@
...
+        ".github/",
+        ".prettierrc",
+        ".vscode/",
+        "docs/**",
+        "jest..js",
+        "src/e2e/",
+        "src/tests/",
+        "src/fixtures/**",
+        // This one is especially important (otherwise dependabot would be blocked by change file requirements)
+        "yarn.lock",
```
chrisdholt, line 14: We previously included a configuration to ensure this only affected those things under packages. This will be the most specific config - It seems we are missing parity with the current config. Is that accurate? If a change is made in “sites”, does this expect a change file?

chrisdholt: @Ibrahimmaga can you help me understand how or why the above has been resolved? There's not a comment or a reference to anything, so from what I can tell - this is still a valid and outstanding question.

chrisdholt (changes requested):

beachball.config.js
awentzel: I don't see this as resolved.

awentzel (changes requested):
Please respond to all comments with clear explanation and path to resolution.

Ibrahimmaga:
> Please respond to all comments with clear explanation and path to resolution.

Hi @christ the old configuration is name: "Microsoft FAST",
include: [
"packages/tooling/",
"packages/utilities/",
"packages/web-components/*",
],
disallowedChangeTypes: ["major"],
we are saying to include those packages, by default beachball includes all packages. So this is not necessary unless we want to exclude a package. It is a repetition. If there is any package we want to exclude let me know. as right now beachball will treat everything that has package.json the same. and you will expect the picture above "minor, patch and none" when you run beachball change.

Ibrahimmaga:
image
exclude command throw error
with the include command we see Major showing
image

without the Group configuration Major bump is not showing
image

beachball.config.js
chrisdholt: Tested with the groups change by updating something within sites and didn't hit any issues. I have raised one other comment above that I think is likely something to address here.

chrisdholt: @Ibrahimmaga I just pulled this down with the groups addition and it worked - is there a reason it's disappeared now?

beachball.config.js
```diff
@@ -1,2 +1,16 @@
-require("./build/releasing/ts-node-register");
-module.exports = require("./build/releasing/index");
+module.exports = {
+    disallowedChangeTypes: ["major"],
+    ignorePatterns: [
+        ".ignore",
+        ".github/",
+        ".prettierrc",
+        ".vscode/",
+        "docs/**",
```
chrisdholt, line 8: @Ibrahimmaga @awentzel I *think* we'll want to leave off docs for things *within* the packages folder. Docs would include the API report here and we *definitely* want to throw on changed files if that is changed and hasn't been included. Any docs within sites are excluded with the change below, but I don't think we want to ignore packages which have a `docs/` folder due to the API report.

awentzel (approved):

sites/site-utilities/src/web-components/pragma.ts
```diff
@@ -5,5 +5,4 @@ import val from "@skatejs/val";
  * Define the wrapped createElement function - this should be imported and
  * declared as the JSX pragma wherever Web Components are rendered by React
  */
-
 export default val(createElement);
```
chrisdholt, line 8: @Ibrahimmaga we likely don't wan to check in this change here - I assume it was a test

chrisdholt (approved):

PR #5229, opened 2021-09-24
feat: add fast-react-wrapper package

EisenbergEffect:
# Pull Request

## 📖 Description

This PR adds a new utility package that enables wrapping Web Components in a React wrapper, for more seamless integration into existing React codebases.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

The source code is in a single file. The wrapper itself isn't that complicated. Its main purpose is to determine which React props should be handled by React and which should be manually applied to the underlying custom element. It also manages refs and custom events.

The wrapper works with any web component but has special handling for FAST components and Design Systems. Have a read through of the updated documentation for an explanation of the usage options.

## 📑 Test Plan

Unit tests have been added for FASTElement-based, FoundationElement-based, and VanillaJS Web Components.

### General

## ⏭ Next Steps

Next step would be to explore whether it's worth generating a single file with the wrappers preconfigured, including their custom event configuration. We really need to look at unifying our API metadata, custom-elements.json, vs code json, and React wrapper generation approaches I think. In theory, we should be able to put everything into docs tags in the source and generate all the other files based on that.


packages/utilities/fast-react-wrapper/README.md
```diff
@@ -0,0 +1,33 @@
...
+yarn add @microsoft/fast-react-wrapper
+```
+
+Within your JavaScript or TypeScript code, you can then and use the wrapper like this:
+
+```ts
+import React from 'react';
+import { provideReactWrapper } from '@microsoft/fast-react-wrapper';
+
+const { wrapper } = provideReactWrapper(React);
```
nicholasrice, line 28:
```suggestion
const { wrap } = provideReactWrapper(React);
```

nicholasrice (approved):
Looks great! Does the implementation infer types from the provided element constructor to enable prop intelisense?

EisenbergEffect:
@nicholasrice Yes, everything should be typed and have tooling support. The only thing that's needed is the optional `events` configuration to add the event properties if they are not standard events known by React. Maybe there's some way we can infer those down the road, but we don't have a way to do that yet.

robarbms (approved):

EisenbergEffect:
@nicholasrice @janechu In my tests, I imported from our testing utilities in fast-foundation. If I import from the `src` I get typings but at runtime the module can't be found. If I import from `dist` everything works at runtime but there are no types. Any ideas for how to get both to work? I was hoping we could turn this into some official documentation for using our test utilities without having to move them out of foundation (that would get messy since they are used by foundation and dependent upon it.) Any thoughts?

awentzel (approved):

nicholasrice:
Not sure on the typings thing but I can spend a few minutes to investigate

EisenbergEffect:
I'd definitely appreciate that. Thanks Nick!

PR #5235, opened 2021-09-27
Proposed: add refs to all dom nodes

### LINKED ISSUE #5234: rfc: add refs to all shadow DOM nodes in component templates
# 💬 RFC - Add refs to all Shadow DOM nodes in component templates

The templates and classes included as part of FAST Foundation exist to provide markup and functionality for a majority of use cases and serve as building blocks for not only the FAST components, but for any component which may require the same underlying role, behavior, etc. While it is almost impossible to handle any and all scenarios a user may be able to consider with a given component, we can provide an extensible model for them to build off. A good and extremely simple example of this is the concept of "appearance" in the FAST and Fluent Design Systems. The foundation classes do not include this concept as they are quite specific to a given design language. There are different semantics, implementations, restrictions - the concept of appearance is one that is *not* foundational or "globally applicable" across all design systems *OR* across all instances of a given type of component - some things which are buttons do not have the need for multiple appearances. With this in mind, we take an approach where we extend the class to enable and support our unique scenarios.

This RFC recommends an additional improvement to the extensibility of our classes and templates. Currently, refs only exist on the control for things which we *know* exist at the foundation level. This creates a limitation to our ability to use the templates. While I can easily extend the class to include certain new attributes and properties such as the `appearance` example above, I cannot do this for references to the template nodes. In instances where I need a reference, I'm essentially relegated to either swapping out the template to include refs where needed OR querying the shadow DOM from within the class which while somewhat reasonable seems less-than-ideal.

One of our core values for these Foundation controls is that they are for broad use and not specifically to support our own scenarios - `fast-components` is just another implementation of foundation, it does not get special treatment. With that said, traditionally, anytime that we've added features or functionality which requires the addition of a ref, we are able to easily add it. In this case, we have the benefit because of our monorepo to make these additions for our own ease of use. It seems reasonable that rather than adding refs whenever we find a use case, we extend the API to include these references which opens up more flexibility and opportunity for the community to take advantage of them. This also means that in the future if we want to use the Web Animations API to transition seamlessly in and out of showing/hiding the Dialog we don't need to go add refs to accomplish this. Folks using Foundation also get access to this immediately and without needing to file an issue and elevate their use case, etc.

## Implementation
The implementation of this would be fairly straight forward. Any node which is not a root node or slot (which is projected into the light DOM so refs can be added by end users) would now have a reference from the component class. Refs would be named according to the CSS Part which I think is intuitive as it's already an established API for providing authors the opportunity to reach in and style via CSS. Refs would all be typed as well, which is an added benefit.

## 💻 Examples

While not illustrative of the entire component set, I've started a [draft PR](https://github.com/microsoft/fast/pull/5235) to illustrate what this would look like across templates.

chrisdholt:
# Pull Request

## 📖 Description

This PR adds refs to all ShadowDOM nodes within a component template

### 🎫 Issues

If accepted, would close #5234

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
+++ b/packages/web-components/fast-foundation/src/accordion-item/accordion-item.template.ts
@@ -24,6 +24,7 @@ export const accordionItemTemplate: (
             part="heading"
             role="heading"
             aria-level="${x => x.headinglevel}"
+            ${ref("heading")}
         >
             <button
                 class="button"
@@ -40,7 +41,7 @@ export const accordionItemTemplate: (
             </button>
             ${startSlotTemplate(context, definition)}
             ${endSlotTemplate(context, definition)}
-            <span class="icon" part="icon" aria-hidden="true">
+            <span class="icon" part="icon" aria-hidden="true" ${ref("icon")}>
                 <slot name="expanded-icon" part="expanded-icon">
                     ${definition.expandedIcon || ""}
                 </slot>
@@ -55,6 +56,7 @@ export const accordionItemTemplate: (
             id="${x => x.id}-panel"
             role="region"
             aria-labelledby="${x => x.id}"
+            ${ref("region")}
         >
             <slot></slot>
         </div>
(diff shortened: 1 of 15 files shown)


chrisdholt:
Closing this and will open an issue to track a helper to get shadow root el's by part name or comparable. Thanks all!

Issue #5125, opened 2021-08-31
fix: IIFE bundle throws an exception when using fluent web components


nmetulev:
# 🐛 Bug Report

When bundling the `@fluentui/web-components` package in an iife bundle, and when that bundle is referenced in the `<head>`, the bundle throws an exception when loaded

## 💻 Repro or Code Sample

This project repros the issue: https://github.com/nmetulev/fast-iife-repro

## 🤔 Expected Behavior

Bundling the components in an IIFE bundle should not cause an exception in runtime.

## 😯 Current Behavior

The [design token class](https://github.com/microsoft/fast/blob/c4fe4d0706dae02bc2d4c8ea37dafe6c79c03075/packages/web-components/fast-foundation/src/design-token/design-token.ts#L22) assumes `document.body` is always valid. However, when used in an IIFE bundle in the `<head>` of a document, the script runs before the body is available and `document.body` is null and later throws an exception.

Exception:

```
Uncaught TypeError: Invalid value used as weak map key
    at WeakMap.set (<anonymous>)
    at new DesignTokenNode (bundle.js:11484)
    at Function.getOrCreate (bundle.js:11500)
    at DesignTokenImpl.setValueFor (bundle.js:11307)
    at DesignTokenImpl.withDefault (bundle.js:11318)
    at bundle.js:19489
    at bundle.js:26028
DesignTokenNode @ bundle.js:11484
getOrCreate @ bundle.js:11500
setValueFor @ bundle.js:11307
withDefault @ bundle.js:11318
(anonymous) @ bundle.js:19489
(anonymous) @ bundle.js:26028
```

## 💁 Possible Solution

There might be other solutions, but these two came to mind:
* NOP when `document.body` is null 
* await for `document.body` to load

## 🔦 Context

We are using the fluent components in the Microsoft Graph Toolkit and our bundles are iife currently. Adding the fluent components breaks our bundles.

## 🌍 Your Environment

* OS & Device: Windows
* Browser: Edge
* Version: "@fluentui/web-components": "^1.3.8"

nicholasrice:
I think long-term the solutions discussed in https://github.com/microsoft/fast/issues/4976 would address this issue, but in the meantime, I think we can defer assignment of default values until the document is ready. Doing so will be transparent to the public API.

nmetulev:
Thanks, that sounds like a great solution!

nicholasrice:
[...]

PR #5241, opened 2021-09-29
refactor DesignToken to not emit to document.body

nicholasrice:
…'t emitted to the body

# Pull Request

## 📖 Description
This PR refactors how `DesignToken.withDefault()` emits CSS custom properties. Prior to this change, token values set with `.withDefault()` would be emitted to `document.body` using the `HTMLElement.style.setProperty()` API. That approach is problematic because:
1. document.body is not defined tokens are set via a script executing prior to the document being fully parsed, as can be  the case if the script is loaded in the `document.head`
2. It emits a lot of inline styles to the document.body, which clutters that element (this isn't a problem per-se, but more of a nuisance). 

With this change, default DesignToken values will be emitted to CSS custom properties in one of two ways.
1. When constructable styles are supported, a `CSSStyleSheet` will be created and added to the `document.adoptedStyleSheets` array.
2. When constructable styles are *not* supported, a `HTMLStyleElement` will be created and added to the `document.head`.

### 🎫 Issues
fixes https://github.com/microsoft/fast/issues/5125

## 👩‍💻 Reviewer Notes
There is a discussion point I've added below to discuss one of the implementation details. I'll leave this PR in draft-state until that point has been discussed / resolved.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
+++ b/packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
@@ -6,7 +6,10 @@ import {
     FASTElement,
     observable,
     Observable,
+    TargetedHTMLDirective,
 } from "@microsoft/fast-element";
+
+export const defaultElement = document.createElement("div");
 
 function isFastElement(element: HTMLElement | FASTElement): element is FASTElement {
     return element instanceof FASTElement;
@@ -17,16 +20,8 @@ interface PropertyTarget {
     removeProperty(name: string);
 }
 
-/**
- * Handles setting properties for a FASTElement using Constructable Stylesheets
- */
-class ConstructableStyleSheetTarget implements PropertyTarget {
-    private target: PropertyTarget;
-    constructor(source: HTMLElement & FASTElement) {
-        const sheet = new CSSStyleSheet();
-        this.target = (sheet.cssRules[sheet.insertRule(":host{}")] as CSSStyleRule).style;
-        source.$fastController.addStyles(ElementStyles.create([sheet]));
-    }
+abstract class QueuedStyleSheetTarget implements PropertyTarget {
+    protected abstract target: PropertyTarget;
 
     public setProperty(name: string, value: string) {
         DOM.queueUpdate(() => this.target.setProperty(name, value));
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
```diff
@@ -17,25 +20,64 @@ interface PropertyTarget {
...
+
+        this.style = document.createElement("style") as HTMLStyleElement;
+        document.head.appendChild(this.style);
+        const { sheet } = this.style;
+
+        if (sheet) {
+            const index = sheet.insertRule(":root{}");
+            this.target = (sheet.rules[index] as CSSStyleRule).style;
+        } else {
+            throw new Error("This should never get thrown");
```
nicholasrice, line 76: note-to-self: we probably shouldn't throw here, though AFAIK we can't hit this.

packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
```diff
@@ -17,25 +20,64 @@ interface PropertyTarget {
...
 
+class HeadStyleElementStyleSheetTarget extends QueuedStyleSheetTarget {
+    protected target: PropertyTarget;
+    private readonly style: HTMLStyleElement;
+
+    constructor() {
+        super();
+
+        this.style = document.createElement("style") as HTMLStyleElement;
+        document.head.appendChild(this.style);
```
nicholasrice, line 69: I'm looking for feedback on this point in particular. It feels a little clunky to me to just shove this element into the document head but maybe I'm overthinking it. I'd *like* a way to provide an insertion-point to make this configurable, but right now `.withDefault()` will emit to a CSS custom property immediately so this work will happen prior to an author getting an opportunity to configure DesignToken in cases where a library is defining the default. In this case I suppose we could *move* the node, and maybe that is acceptable? The options as I see them are

1. Determine this is not a problem and leave this as-is
2. Implement some configuration API to provide an insertion-point for this stylesheet. For now we would move the sheet if it is already created, but in the future we could hold emitting default values until some `DesignToken.init()` API is invoked.
3. Look for a magic comment-node like `<!-- fast-design-token-target -->` and emit this element immediately after that node.
4. 2 & 3

EisenbergEffect: I think we could go with this as the first version to fix the bug. Then we can follow up with an API to enable declaring the insertion point. It seems like we can probably call that API and mirror it from the DesignSystem. At that point we could delay inserting until component registration, or just insert it and move it. I think I'd avoid magic comments for now. I'd prefer the explicit API. If we need a comment as an anchor point to enable that, I'm fine with that as an internal implementation thing.

EisenbergEffect (approved):

packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
nicholasrice: Yea sounds good. We can add a mechanism to the feature discussed in #4976

PR #5242, opened 2021-09-29
feat: Emit scrollend in horizontal-scroll

bschmaltz:
# Pull Request

## 📖 Description

Emit scrollend events when horizontal-scroll is done scrolling. This feature is needed, so I can track the visibility of ads in carousel.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
+++ b/packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
@@ -406,7 +406,7 @@ export class HorizontalScroll extends FoundationElement {
 
         if (this.speed < 1) {
             this.scrollContainer.scrollLeft = newPosition;
-            this.$emit("scrollend", newPosition);
+            this.$emit("scrollend");
             return;
         }
 
@@ -420,7 +420,7 @@ export class HorizontalScroll extends FoundationElement {
 
         if (stepCount < 1) {
             this.scrolling = false;
-            this.$emit("scrollend", newPosition);
+            this.$emit("scrollend");
             return;
         }
 
@@ -434,7 +434,6 @@ export class HorizontalScroll extends FoundationElement {
         steps.push(newPosition);
 
         this.move(steps, this.frameTime);
-        this.$emit("scrollend", newPosition);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -432,6 +434,7 @@ export class HorizontalScroll extends FoundationElement {
         steps.push(newPosition);
 
         this.move(steps, this.frameTime);
+        this.$emit("scrollend", newPosition);
```
robarbms, line 437: I think you'll want to move this inside of the this.move() after it checks if there are any remaining steps. After line 455.

bschmaltz: O, that makes sense. I originally read it as a recursive function, but I now see it is called within the requestAnimationFrame callback.

Going to drop the newPosition since I don't have a use for it, so I don't want to bother passing that to move.

robarbms (approved):

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -406,6 +406,7 @@ export class HorizontalScroll extends FoundationElement {
 
         if (this.speed < 1) {
             this.scrollContainer.scrollLeft = newPosition;
+            this.$emit("scrollend");
```
nicholasrice, line 409: I do find it a little odd that the control is firing a `scrollend` event even when it doesn't consider itself scrolling (`this.scrolling === false`). My initial thought was that you could make `this.scrolling` observable and emit a scroll start/end event depending on the transition (`true` -> `false` === scrollend, `false -> true` === scrollstart) but that wouldn't work because scrolling is only set if speed > 1? I'm not familiar enough with this control to know why that is.

Anyway, if we're adding a "scrollend" event I think for parity we should probably emit a "scrollstart" event.

robarbms: I considered adding more events like scrollstart and scrollend but Ben sent the PR out for this. I figured, if it was needed, you could attach an onclick event handler to get the scrollstart, there was just no good way to get the scrollend.

Using this.scrolling makes a lot of sense though. It was initially just used to prevent scrolling from being initiated while it was currently scrolling. We'd just need to make sure that we set this.scrolling to true and then false during an instant scroll (speed=0).

Issue #5246, opened 2021-10-01
fix: [what is the issue?] in [where is the issue?]


atmgrifter00:
# 🐛 Bug Report

I believe a [recent change](https://github.com/microsoft/fast/pull/5205) broke the inc/dec buttons of the `fast-number-field`.

## 💻 Repro or Code Sample

Literally just use `@microsoft/fast-foundation@2.16.4` (or later) along with a compatible `fast-components` version, put a `fast-number-field` down and try to use the inc/dec buttons. They won't update the value.

## 🤔 Expected Behavior

Pressing the inc/dec buttons should cause the value to update.

## 😯 Current Behavior

The value does not update.

## 💁 Possible Solution

A new `displayText` field has been added to the `NumberField` which I believe the template is getting its display text from, yet it isn't updated anywhere except the `connectedCallback` function, which seems wrong. Why aren't we just updating this in `updateValue()`?

## 🌍 Your Environment

Windows Desktop
@microsoft/fast-foundation@2.16.4
@microsoft/fast-components@2.8.3

atmgrifter00:
I'll mention that I have a fix in work for this (along with tests). I don't mind picking up the torch here.

atmgrifter00:
PR #5250 is up.

Issue #5249, opened 2021-10-04
fix: range coercion in number field makes data entry difficult


brianehenry:
# 🐛 Bug Report

When using the number field with a minimum value, many valid values are difficult or impossible to type in.

## 💻 Repro or Code Sample

`<fast-number-field min="10"></fast-number-field>`

1. Using the above configuration, try typing in "100". After typing the first character, the field will coerce the value, and after typing "100" the field will display "1000".
2. With a value already existing in the field, try backspacing in order to enter a different value. The field will not let you backspace through the whole value, as it coerces back to "10" on each keystroke.

## 🤔 Expected Behavior

Should be able to enter valid values in the field without having to fight range coercion.

## 😯 Current Behavior

Minimum values are enforced on every keystroke, making data entry difficult.

## 💁 Possible Solution

Coerce field value to range on `change` or `blur` instead of `input`.

## 🔦 Context

Trying to use number field with range configuration.

## 🌍 Your Environment

* OS & Device: Windows PC
* Browser: Microsoft Edge
* Version: @microsoft/fast-foundation 2.12.0

PR #5250, opened 2021-10-04
fix: NumberField increment/decrement buttons no longer update display value

atmgrifter00:
# Pull Request

## 📖 Description

A [recent change](https://github.com/microsoft/fast/pull/5205) to the NumberField inadvertently broke the increment/decrement buttons from changing the displayed value when pressed. More broadly, any programmatic change of the value will not update the displayed value appropriately.

### 🎫 Issues

#5246 

## 📑 Test Plan

I've added a couple of unit tests that focus on the `stepUp` and `stepDown` methods updating the displayText as expected. While the issue is technically broader than this, this seems like a reasonable set of tests.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/number-field/number-field.ts
+++ b/packages/web-components/fast-foundation/src/number-field/number-field.ts
@@ -209,6 +209,7 @@ export class NumberField extends FormAssociatedNumberField {
         }
 
         if (value != this.value) {
+            this.displayText = value;
             this.value = value.toString();
             this.$emit("input");
             this.$emit("change");
--- a/change/@microsoft-fast-foundation-eab39a41-affe-455e-8298-279c09489ebc.json
+++ b/change/@microsoft-fast-foundation-eab39a41-affe-455e-8298-279c09489ebc.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Fix inc/dec not updating displayText",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "26874831+atmgrifter00@users.noreply.github.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-react-wrapper-a959ca07-d92b-4cb9-a857-97447e6390a9.json
+++ b/change/@microsoft-fast-react-wrapper-a959ca07-d92b-4cb9-a857-97447e6390a9.json
@@ -1,7 +1,0 @@
-{
-  "type": "minor",
-  "comment": "feat: add fast-react-wrapper package",
-  "packageName": "@microsoft/fast-react-wrapper",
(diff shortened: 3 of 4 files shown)


atmgrifter00:
I'm not sure why I had to delete that other change file, but it was present after my fork, and `yarn change` complained about its presence.

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -209,6 +209,7 @@ export class NumberField extends FormAssociatedNumberField {
         }
 
         if (value != this.value) {
+            this.displayText = value;
```
robarbms, line 212: This won't work. It breaks typing in values that aren't valid while you type them. Such as trying to type **-.5e10** isn't possible with this. We'd have to make adjustments to displayText in stepUp and stepDown methods.

We could update the displayText in **valueChanged()** to catch changes to value by script, but that could update it during the **updateValue()** method when we don't want it to.

One way to deal with it all is to create a flag, **updatingValue** that is set to true when **updateValue()** is called. It would set it to false after it is finished. Then we could just set **this.displayText** to **this.value** when **this.updatingValue** is equal to false. This would update when changed by script and would fix **stepUp** and **stepDown**.

robarbms (changes requested):

packages/web-components/fast-foundation/src/number-field/number-field.ts
atmgrifter00: Cool...you want me to try and tackle this?

robarbms: I can get this. It would fix updating with scripts and increment and decrement. I'd like to build more tests around this to catch all of these edge cases as well.

atmgrifter00: Also, just so there's no confusion, my change does not seem to inhibit the text entry you outlined.

robarbms: It was lagging and would sometimes replace the value with 0 when I was playing around with entering numbers.

atmgrifter00: Ah, I see. Yeah, after playing with it a bit, the input does seem a little more fussy with my change present (not impossible, but definitely fussier). Still, I feel that the inevitable fix for [this Issue](https://github.com/microsoft/fast/issues/5249) might address that.

robarbms: I created a number-field with a min="10". With the new displayText property, it won't coerce the value to the min while typing. The internal value will actually be 10, but it will show what the user is typing now.

PR #5277, opened 2021-10-12
remove throttle fallback function in fast-animation and fast-web-utilities

radium-v:
# Pull Request

## 📖 Description

Removes the `throttle` function in `fast-web-utilities` and reworks `fast-animation` to always use `requestAnimationFrame`.

## 👩‍💻 Reviewer Notes

This also fixes a warning which was present while building webpack in `fast-animation`.

## 📑 Test Plan

All tests and Playwright steps should build and run successfully.

### General

### DIFF
--- a/change/@microsoft-fast-animation-48d45321-123e-497b-8f48-691f5af5b78c.json
+++ b/change/@microsoft-fast-animation-48d45321-123e-497b-8f48-691f5af5b78c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "use requestAnimationFrame in fast-animation",
+  "packageName": "@microsoft/fast-animation",
+  "email": "john.kreitlow@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-web-utilities-4ac26361-b8f3-407a-b56c-705d8fa6493d.json
+++ b/change/@microsoft-fast-web-utilities-4ac26361-b8f3-407a-b56c-705d8fa6493d.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "remove throttle function",
+  "packageName": "@microsoft/fast-web-utilities",
+  "email": "john.kreitlow@microsoft.com",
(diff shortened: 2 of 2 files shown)


chrisdholt:
Can you elaborate on other potential implications here? For instance, if we want to use `throttle` in `fast-foundation` or `fast-components` in the future, we'll need to take a dependency on `fast-animation`. That seems like the wrong chain. I want a utility, not an animation package. While I get that `fast-animation` is using this code _now_, but this doesn't bode well for potential future use cases. 

It seems like this may be a quick fix to address test-runner scenarios leveraging Node environments. While I understand that, I'm concerned about this:

> Each environment that uses throttle will need to create its own method by importing makeThrottler and binding to its respective performance.now.

I feel that there _must_ be some kind of alternative here to accommodate the playwright issue.

chrisdholt:
> > Can you elaborate on other potential implications here? For instance, if we want to use `throttle` in `fast-foundation` or `fast-components` in the future, we'll need to take a dependency on `fast-animation`.
> 
> In this case, `throttle` is only used in one place in `fast-animation`, so I made it a local const instead of exporting it. Other projects which could potentially use `throttle` would just need to do what `fast-animation` does:
> 
> ```ts
> import { makeThrottler } from "@microsoft/fast-web-utilities";
> const throttle = makeThrottler(performance.now.bind(performance), setTimeout);
> ```
> 
> Though it's unlikely that this would be needed anyway, since `fast-animation` only uses `throttle` as a fallback when `requestAnimationFrame` isn't available:
> 
> ```ts
> if (this.subscriptions.length === 0) {
>     if (this.useRequestAnimationFrame) {
>         window.addEventListener("scroll", this.requestFrame);
>     } else {
>         // If we can't use window.requestAnimationFrame, just throttle the update method
>         this.update = throttle(this.update, 1000 / 60); // 60fps
>         window.addEventListener("scroll", this.update);
>     }
> }
> ```

Ah, I see now that you moved just that portion. Should we update the README with that guidance in `fast-web-utilities`?

chrisdholt:
Woo hoo! Thanks @radium-v!

PR #5284, opened 2021-10-13
feat: adds DesignToken root configurability

### LINKED ISSUE #4976: feat: add ability to configure defaultElement for DesignToken

nicholasrice:
# Pull Request

## 📖 Description
This PR adds a mechanism to control where the CSS custom property for a `CSSDesignToken`'s default value is emitted. The feature supports emitting default values to any number of root HTMLElement's or Documents.

I would consider this feature *technically breaking* because it changes the behavior from automatic emission of default values to only emitting after manual registration occurs. This will only be an issue for authors that aren't using `DesignSystem` though, because because this PR also updates the `DesignSystem.register()` method to register the `DesignToken` root, also adding a configuration method to `DesignSystem` to configure the root itself.

### 🎫 Issues
fixes #4976

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
+++ b/packages/web-components/fast-foundation/src/design-token/custom-property-manager.ts
@@ -176,23 +176,26 @@ export class RootStyleSheetTarget implements PropertyTarget {
         RootStyleSheetTarget.properties[name] = value;
 
         for (const target of RootStyleSheetTarget.roots.values()) {
-            PropertyTargetManager.getOrCreate(target).setProperty(name, value);
+            PropertyTargetManager.getOrCreate(
+                RootStyleSheetTarget.normalizeRoot(target)
+            ).setProperty(name, value);
         }
     }
 
     public removeProperty(name: string): void {
         delete RootStyleSheetTarget.properties[name];
         for (const target of RootStyleSheetTarget.roots.values()) {
-            PropertyTargetManager.getOrCreate(target).removeProperty(name);
+            PropertyTargetManager.getOrCreate(
+                RootStyleSheetTarget.normalizeRoot(target)
+            ).removeProperty(name);
         }
     }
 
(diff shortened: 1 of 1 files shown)


nicholasrice:
@EisenbergEffect I didn't implement the WeakRef implementation because DesignTokenNode already uses a hard reference to it's source HTMLElement, but if you think we should replace all of those with WeakRef implementations we can do that.

Issue #5119, opened 2021-08-27
rfc: separate 'value' and 'initial-value' concepts for FAST form-associated controls


nicholasrice:
# 💬 RFC: Form Controls and the 'value' Property
## How FAST's Form Controls Work Today
Today, all FAST form controls support a 'value' JavaScript property and 'value' HTML attribute. The behavior expressed by all FAST form controls matches the [platform behavior](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-value), where setting the 'value' HTML attribute specifies the **initial value** of the control and the stateful value is expressed by element instances's 'value' JavaScript property (`<input value="initial-value" />` vs `myInputElement.value`). In cases where the control has a stateful value change (user interaction or programatic adjustment), the stateful 'value' JavaScript property will not match the 'value' HTML attribute.

## The Problem
Many UI frameworks (React, Vue, Blazor, etc) do not provide explicit mechanisms to configure how data is bound to HTML element instances. Instead, they rely on heuristics to determine when to bind data to HTML attributes, event listeners, and JavaScript properties. This has become a large pain-point with FAST's form controls because most of these UI frameworks special-case native form elements so that they work correctly with the framework.

When these frameworks render a custom form element from FAST for which the author is binding a value, the framework does not special-case the 'value' property for the custom element (special casing is generally done by tag-name), causing the framework to set the 'value' HTML attribute. As noted above, this effectivly sets the *inital* value and not necessiarly the *stateful* value, which can cause the control not to update when authors expect it to.

## Proposed Solution
The pragmatic solution would be for FAST to adjust the component interface to work in a way that better interfaces with these frameworks. This would greatly reduce developer frustration when using FAST components with these frameworks. Two changes would need to be made:

**1. Implement a 'initial-value' HTML attribute**
The 'intial-value' HTML attribute would be responsible for what the current 'value' attribute is responsible for and would be behaviorly the same as the existing 'value' attribute, namely configuring the inital value of the control and reacting to form reset callbacks.

> A parallel change would need to be made to the 'checked' attribute & property for fast-checkbox and fast-radio

We could also consider 'default-value', which is the name React has used to fill this purpose: https://reactjs.org/docs/uncontrolled-components.html#default-values

**2. Make the 'value' HTML attribute and JavaScript property solely responsible for element state**
[...]

PR #5298, opened 2021-10-20
feat: adds a currentValue property to form-associated

nicholasrice:
# Pull Request

## 📖 Description
This PR implements a new property and attribute, `currentValue` and `current-value` respectivly, to the `FormAssociated` mixin. This change is discussed in depth in https://github.com/microsoft/fast/issues/5119 but in short, this change will allow UI libraries that don't support explicitly IDL attribute binding to bind to the `current-value` attribute instead of the `value` attribute of FAST form controls, which will properly set the state of the form control.

## 👩‍💻 Reviewer Notes
I've implemented the new `current-value` attribute to mirror the property, which means the stateful value will always be in the document. Is this what we want? I also considered the 'from-view' attribute mode where we wouldn't reflect changes to the property back to the attribute value, because then the stateful value wouldn't always be reflected to the document and the element attributes would be less cluttered. 

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-foundation-b5e8fa1f-0071-4e55-a53b-ebcfd85a9576.json
+++ b/change/@microsoft-fast-foundation-b5e8fa1f-0071-4e55-a53b-ebcfd85a9576.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "adds a currentValue property to form-associated to facilitate attribute-based data binding for form values",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "chhol@microsoft.com",
+  "dependentChangeType": "patch"
+}


vnbaaij:
> > > Thank you @nicholasrice! Should we add any docs specifically in regards to this additional attribute? Can also be a follow-up as I realize that doesn't really _fit_ with our current docs around Form Association.
> > 
> > 
> > We definitely should, though I'm not aware of any current docs on it because the export is alpha. Unless I'm missing where it is? I can take a task to write a docs page for it if we indeed don't have a dedicated section, otherwise I'll update it.
> 
> Perhaps we add the docs to the framework-specific integration docs? Definitely we can add this to the Blazor wrappers and documentation as well.

In working this into the Blazor wrappers, we can make it so that on the Blazor side the property is still called `Value` (and accompanied by `ValueChanged` event). Think we need to see how it works when implemented and then maybe no docs are needed (because it works just as Blazor users would expect.

PR #5333, opened 2021-10-28
feat: add rating spec

khamudom:
# Pull Request

## 📖 Description

Specifications for the Rating component

### General

### Component-specific

### DIFF
--- a/specs/rating/rating.md
+++ b/specs/rating/rating.md
@@ -1,0 +1,162 @@
+# Rating
+
+## Overview
+
+The Rating component is used to provide feedback from a user's opinion on a particular product or experience.
+
+### Use Cases
+
+Scenario 1
+Carl has purchased a coffee machine online and he wants to share his opinion about the machine with others. In addition to writing a review, Carl also rates the product by giving it a rating of 4 out of 5 stars.
+
+Scenario 2
+Wendy just finished her companies survey. At the end of the survey, she is asked to rate how she felt about the experience. She is given a collection of emojis to choose from about how she felt.
+
+### Prior Art/Examples
+
+- [Office Fabric](https://developer.microsoft.com/en-us/fluentui#/controls/web/rating)
+- [Ant Design](https://ant.design/components/rate/)
+- [Atlassian](https://atlaskit.atlassian.com/packages/design-system/rating)
+- [Windows (UWP)](https://docs.microsoft.com/en-us/windows/apps/design/controls/rating)
+
+---
+
+### API
+
+Extends [RadioGroup](../radio-group/radio-group.spec.md).
+
+_Component name:_
+
+- `fast-rating`
+
+_Attributes:_
+
+- `disabled` - A boolean to remove interaction with the component.
+- `readonly` - A boolean to remove interaction with the component.
+- `value` - The current value of the item selected.
+- `mode` - enum
+  - multiple: This is the default mode. A mode where multiple items are highlighted for rating.
+  - single: A mode where the a user chooses only one item to rate.
+
+_Slots:_
+
+- `default` - Default slot is the slotted items.
+- `label` - Optional label. Some examples would be to indicate rating counts, or average ratings.
+
+### Anatomy and Appearance
+
+The Rating component template is extended from RadioGroup.
+
+_Template:_
+
+```html
+<template
+    role="radiogroup"
+    aria-disabled="${x => x.disabled}"
+    aria-readonly="${x => x.readOnly}"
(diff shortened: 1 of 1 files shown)


specs/rating/rating.md
```diff
@@ -0,0 +1,162 @@
...
+
+- When focused is on the component, the reader will announce the component, the current item and the the total amount of items. Example: "rating, 1 of 5 stars".
+As focused is moved to the next item, the reader will read out the current value and the amount of total items. Example: "2 of 5 stars, 3 of 5 stars, 4 of 5 stars, 5 of 5 stars".
+
+
+
+# Rating item
+
+## Overview
+The `fast-rating-item` is to used inside the `fast-rating` component.
```
chrisdholt, line 109: I don't love the `item` portion here. I'm curious - what makes up a "rating"? Is it the group together or the individual thing? If it's the group, does "item" make sense? It's better than something specific like "star" as that modifies the type of rating, but I'm curious if we can come up with a better semantic than this.

chrisdholt: Also, if this is just a radiogroup and radio - perhaps this only needs to be a `fast-components` implementation rather than something in foundation?

specs/rating/rating.md
```diff
@@ -0,0 +1,162 @@
...
+        @click="${(x, c) => x.clickHandler(c.event as MouseEvent)}"
+    >
+        <div class="rating-icons" part="rating-icons">
+            <slot name="unchecked-icon" part="unchecked-icon" aria-hidden="true">
+                ${definition.uncheckedIcon || ""}
+            </slot>
+            <slot name="checked-icon" part="checked-icon" aria-hidden="true">
+                ${definition.checkedIcon || ""}
+            </slot>
+        </div>
```
chrisdholt, line 154: Can you elaborate a bit about how checked / unchecked maps to rating - is there a method of addressing partial ratings with this approach?

specs/rating/rating.md
khamudom: My thought would be that the rating group component makes up the rating system. Since it will handle the items and the items could be any element, that has a role of radio. I went with the group and items system to follow the patterns of the other components that have a group and items, like menu and breadcrumb. But that does not mean Rating has to be the same. Just wanted to follow that pattern.
I did initially investigate just to have the Rating component and the template would use a repeat directive. I am happy to continue this approach if it makes more sense.  
I also had a similar thought that anyone can just use the RadioGroup and Radio and build their own rating system. But in this case, I guess one can say FAST Rating has already done the work for them. Maybe? Because it would have the functionalities to make it work as a rating system, accessibility, and other features already built.

specs/rating/rating.md
khamudom: I ran through a couple of names for these slots. I had front-icon/back-icon. empty-icon/filled-icon. I am not tied to unchecked-icon/checked-icon, but the reasoning for these names is because these items were basically radio items, so I went with a name that corresponds with the input being checked or unchecked.

specs/rating/rating.md
scomea, line 109: "is to used"?

EisenbergEffect (approved):
Approved pending Nick's feedback.

Thoughts for v2:

It seems that individual items may each have unique checked and unchecked icons. So, we may want to consider whether there's some API that a developer can use to set all of that up so it doesn't need to be repeated everywhere a particular rating component is used. For example, configure the component with 5 different emojis in one place.

One additional thought is that individual items could have more than a checked/unchecked state. A common example is a rating control in read-only mode where any item might show empty, half, or full stars.

awentzel:
@khamudom what's the status on this PR?  How much time remains to complete?  We can always abandon if you no longer thing this work is necessary.  Thanks

chrisdholt:
> @khamudom what's the status on this PR? How much time remains to complete? We can always abandon if you no longer thing this work is necessary. Thanks

@Khamudom can advise on if he wants to keep this rolling but considering we don't have anyone relying on this _currently_ I think we should just close. We can always revisit, but for now I don't see a reason to finalize the spec unless we plan to pick this up.

khamudom:
> > @khamudom what's the status on this PR? How much time remains to complete? We can always abandon if you no longer thing this work is necessary. Thanks
> 
> @khamudom can advise on if he wants to keep this rolling but considering we don't have anyone relying on this _currently_ I think we should just close. We can always revisit, but for now I don't see a reason to finalize the spec unless we plan to pick this up.

We should close for now and pick it up again when I or another has some bandwidth to build the component.

PR #5348, opened 2021-11-01
docs: fix Blazor, webpack, and cdn links

EisenbergEffect:
# Pull Request

## 📖 Description

Once again Webpack has made a breaking change in a non-major release that affects the most basic dev server setup. This updates our documentation so that our community isn't blocked.

Additionally, all CDN links are updated to jsdelivr. Some additional docs have been added to Blazor based on feedback from the community.

### 🎫 Issues

No issues. Just discovered this fielding a customer's issues over the weekend.

## 📑 Test Plan

n/a

### General

## ⏭ Next Steps

Would be great to have a way to find these issues ahead of customers. I'm not sure how best to do this without automated testing that just runs a version of this tutorial every day to see if Webpack broke things.


chrisdholt:
> @awentzel Is there anyway that we can exclude documentation changes from needing a change file for the repo? It's a huge barrier to entry. It means that even fixing the smallest typo requires someone to clone the repo, install everything, make the change, run the changelog generation, branch, commit, PR, etc. Without the change file requirement, anyone could edit directly in browser (linked from our docs) and automatically have GH create the branch and PR. None of that extra work required. It's what I tried to do with this PR, but obviously doesn't work. So, now I've got to go through those other steps. I don't have time for a while. So, this is going to just sit here and our documentation will remain incorrect until I get around to having time to do this. Definitely would like to address this if we can.

I think the issue is that this lives in Foundation so the default rule isn't grabbing this. 

I'll add though, that while I think not including change files for the integration docs makes sense, I'm not sure I'd exclude _every_ markdown file from requiring a change file. Certain changes to README for the root or for foundation components may be beneficial to require change files for. Might be good to figure out exactly what constitutes a change that should be included or trigger changes or not.

nicholasrice (approved):

awentzel (approved):

awentzel:
> @awentzel Is there anyway that we can exclude documentation changes from needing a change file for the repo? It's a huge barrier to entry. It means that even fixing the smallest typo requires someone to clone the repo, install everything, make the change, run the changelog generation, branch, commit, PR, etc. Without the change file requirement, anyone could edit directly in browser (linked from our docs) and automatically have GH create the branch and PR. None of that extra work required. It's what I tried to do with this PR, but obviously doesn't work. So, now I've got to go through those other steps. I don't have time for a while. So, this is going to just sit here and our documentation will remain incorrect until I get around to having time to do this. Definitely would like to address this if we can.

Yes, I believe in `beachball.config.js` we can include an ignorePattern value for `*.md` if we wanted or any directory for that matter.  I think this makes sense. Also, we still deploy documentation to staging after merge to master.  I just remembered I'll add dispatch capability as well so code owners can deploy to production at any time. 

@chrisdholt @EisenbergEffect what ignorePattern would you like to use here?  I can make the change and do some testing?

EisenbergEffect:
Added a number of other docs fixes to this PR. Ready for review again.

awentzel (approved):

janechu (approved):

Issue #5322, opened 2021-10-25
fix: Selecting disabled tree items in the FAST TreeView issues 'selected-change' events


atmgrifter00:
# 🐛 Bug Report

As the description states, when clicking on a tree item inside a FAST TreeView the 'selected-change' event [will be emitted](https://github.com/microsoft/fast/blob/45f000f038302d31636bccd85cf20a366aad7a0b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts#L133), even if the TreeItem has the 'disabled' flag set, which will prevent it from becoming "selected" (having the `selected` attribute applied to it).

This seems incorrect.

## 🤔 Expected Behavior

Clicking on a disabled TreeItem does not result in that TreeItem issuing a `selected-change` event.

## 😯 Current Behavior

Clicking on a disabled TreeItem results in a `selected-change` event being issued, despite the fact that the TreeItem `selected` state does not change.

## 💁 Possible Solution

Check if the TreeItem is disabled before emitting the `selected-change` event.

## 🔦 Context

This isn't really an issue for us, as we have an extended TreeItem that ultimately allows us to provide us the behavior we expect.

## 🌍 Your Environment

Windows desktop. Chrome.

chrisdholt:
@rajsite just an update - @eljefe223 is running an audit and addressing all instances where disabled controls still emit events. This should be addressed as part of that.

rajsite:
@chrisdholt thanks! I appreciate the update :)

eljefe223:
Have a PR up that addresses the above https://github.com/microsoft/fast/pull/5421

This fixes the selected change event from being fired, however I think we should investigate moving the click event that assigns the `selected-change` event from the root template to the `positioning-region`. This is would ensure that only the click event fires on the item itself and not when a Tree Item's children are clicked.

Click event fires on entire region rather that just the red highlighted:

![image](https://user-images.githubusercontent.com/37851214/144108288-e7d649dc-7daa-43f0-91f8-6b921798d154.png).

Another reason this is an issue is if the Tree Item does not span the entire width of the Tree clicking outside the Tree Item would select it's parent.

With this change we could simply prevent `pointer-events` via css on a disabled TreeItem:

```css
fast-tree-item([disabled]) {
    pointer-events: none;
}
```

chrisdholt:
> Have a PR up that addresses the above #5421
> 
[...]

PR #5350, opened 2021-11-01
DRAFT fix: rework tree view events

scomea:
## 📖 Description

The current tree view event model results in a tree item emitting a selection event prior to its selection state actually changing, which is not what a developer would expect.  With this change tree view listens for click events in order to determine when an item is selected, and items only emit a selected event when their selected state actually changes.  Worth noting this approach should enable the tree to respond to developer's setting the selected state of an item directly.

Additionally, tree view no longer sets the tabindex on tree items as the tree itself was already adding itself to the tab queue so setting it on the items was redundant.

Also added comments/labels to a number of functions.

## 👩‍💻 Reviewer Notes

This is a breaking change because any change to the timing of events and state changes could break some custom implementations as did the previous change that changed the order of setting prop and emitting change event.

## 📑 Test Plan
Need to add tests around events.

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-foundation-96d671a6-621c-4790-8217-b319bdc883d3.json
+++ b/change/@microsoft-fast-foundation-96d671a6-621c-4790-8217-b319bdc883d3.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "tree view events",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "scomea@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -73,12 +73,27 @@ export class TreeItem extends FoundationElement {
      */
     public expandCollapseButton: HTMLDivElement;
 
+    /**
+     * Whether the item is focusable
+     *
+     * @internal
+     */
     @observable
     public focusable: boolean = false;
 
+    /**
+     *
+     *
+     * @internal
+     */
     @observable
     public childItems: HTMLElement[];
 
+    /**
+     * The slotted child tree items
+     *
+     * @internal
+     */
     @observable
     public items: HTMLElement[];
     private itemsChanged(oldValue, newValue): void {
@@ -93,16 +108,25 @@ export class TreeItem extends FoundationElement {
     }
 
     /**
+     * Indicates if the tree item is nested
+     *
      * @internal
      */
     @observable
     public nested: boolean;
 
+    /**
+     *
+     *
+     * @internal
+     */
     @observable
     public renderCollapsedChildren: boolean;
 
     /**
      * Places document focus on a tree item
+     *
+     * @public
      * @param el - the element to focus
      */
     public static focusItem(el: HTMLElement) {
@@ -110,12 +134,31 @@ export class TreeItem extends FoundationElement {
         el.focus();
     }
 
+    /**
+     * Whether the tree is nested
+     *
+     * @public
+     */
+    public readonly isNestedItem = (): boolean => {
+        return isTreeItemElement(this.parentElement as Element);
+    };
+
+    /**
+     * Handle expand button click
+     *
+     * @internal
+     */
     public handleExpandCollapseButtonClick = (e: MouseEvent): void => {
         if (!this.disabled && !e.defaultPrevented) {
             this.expanded = !this.expanded;
         }
     };
 
+    /**
+     * Gets number of children
+     *
+     * @internal
+     */
     public childItemLength(): number {
         const treeChildren: HTMLElement[] = this.childItems.filter(
             (item: HTMLElement) => {
@@ -124,10 +167,6 @@ export class TreeItem extends FoundationElement {
         );
         return treeChildren ? treeChildren.length : 0;
     }
-
-    public readonly isNestedItem = (): boolean => {
-        return isTreeItemElement(this.parentElement as Element);
-    };
 }
 
 /**
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -66,6 +66,11 @@ export class TreeItem extends FoundationElement {
     @attr({ mode: "boolean" })
     public disabled: boolean;
 
+    /**
+     *  Reference to the expand/collapse button
+     *
+     * @internal
+     */
     public expandCollapseButton: HTMLDivElement;
 
     @observable
--- a/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
+++ b/packages/web-components/fast-foundation/src/tree-item/tree-item.ts
@@ -37,6 +37,11 @@ export class TreeItem extends FoundationElement {
      */
     @attr({ mode: "boolean" })
     public expanded: boolean = false;
+    private expandedChanged(): void {
+        if (this.$fastController.isConnected) {
+            this.$emit("expanded-change", this);
+        }
+    }
 
     /**
      * When true, the control will appear selected by user interaction.
@@ -46,6 +51,11 @@ export class TreeItem extends FoundationElement {
      */
     @attr({ mode: "boolean" })
     public selected: boolean;
+    private selectedChanged(): void {
+        if (this.$fastController.isConnected) {
+            this.$emit("selected-change", this);
+        }
+    }
 
     /**
      * When true, the control will be immutable by user interaction. See {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled | disabled HTML attribute} for more information.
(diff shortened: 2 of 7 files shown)


packages/web-components/fast-foundation/src/tree-item/tree-item.ts
```diff
@@ -46,6 +51,11 @@ export class TreeItem extends FoundationElement {
      */
     @attr({ mode: "boolean" })
     public selected: boolean;
+    private selectedChanged(): void {
+        if (this.$fastController.isConnected) {
+            this.$emit("selected-change", this);
```
nicholasrice, line 56: I *think* this behavior is inconsistent with the existing behavior, in that this event now emits for `true` -> `false` as well as `false` -> `true` mutations. I believe the existing implementation only emits for `false` -> `true` mutations (I could be mistaken here, and if so please tell me :)). Is the deviation here intentional?

packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -20,223 +20,296 @@ import { FoundationElement } from "../foundation-element";
...
-        return this.slottedTreeItems.some((node: HTMLElement) => {
-            return isTreeItemElement(node) && node.querySelector("[role='treeitem']");
-        });
-    }
+    /**
+     * The tree item that is designated to be in the tab queue.
+     *
+     * @internal
+     */
+    @observable
```
nicholasrice, line 65: Does this need to be observable? I don't see any observers or change handlers

scomea: no, I'll take that out.

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
scomea: Not quite.  Since a recent change the tree item emits "selected-change" on click, every click, before the parent tree actually changes the 'selected' prop of the item.  If a dev hooked up a listener on that item they could easily get the change event prior to the state actually changing which is not ideal.  That being said, should we only be emitting for true?  I don't see an upside to that myself.

nicholasrice: Yea I understand this is mostly to fix timing, but my point is this also introduces a new behavior - that is the part I'm wary of. I don't know if we should only be emitting for true, but emitting for false will be behaviorally different. IMO fixing a regression to the event timing isn't "breaking". Adding event emission in the opposite selection case seems to me to be "breaking".

scomea: I don't think that's the case.  This is the handler invoked on click/keydown from months ago, I didn't filter out deselection.

`` private handleSelected(e?: Event): void {
        this.selected = !this.selected;
    }``

nicholasrice: This is the code prior to the timing regression: https://github.com/microsoft/fast/blob/3ad50c8eec98b3156a4559586ca76e4abebefad8/packages/web-components/fast-foundation/src/tree-item/tree-item.ts#L299

That handler is only invoked on item click and "enter" key input though, not whenever `selected` is changed as this code is.

scomea: The initial implementation changed the selected prop and emitted a "selected-change" event from the tree item on click/keystroke.  The parent tree listened for and reacted to that event.  The change last month modified that such that the decision as to whether a clicked element should become selected or not was made at the tree level as a result of the same "selected-change" event as previous, but now the event was emitted prior to the actual state change.  The proposal here is to have the tree listen for click events directly and set selection based on that, which in turn emits the "selected-change" event based on the property change.  This also means that the tree can handle developers setting the selected state of an item directly.

scomea: I think it's reasonable to argue that the original implementation emitted the event on property change just as much as on user interaction as the two were closely meshed.

nicholasrice: I understand the goals. Timing aside, my pushback is specific to the fact that this pull-request causes `tree-item`s to emit a `selected-change` event both when becoming *selected* and *deselected*, whereas the existing behavior is that they only emit a `selected-change` event when becoming *selected*.

nicholasrice: Actually I guess that's not quite true. It would emit for deselection in cases of user-interaction of the item itself.

scomea: > I understand the goals. Timing aside, my pushback is specific to the fact that this pull-request causes `tree-item`s to emit a `selected-change` event both when becoming _selected_ and _deselected_, whereas the existing behavior is that they only emit a `selected-change` event when becoming _selected_.

Ok, so there is a subtle difference. It is not that a tree item that is clicked on while selected now emits a "selection-changed" event for becoming deselected, but rather that items who become deselected for reasons other than being clicked on directly would now emit as well.

nicholasrice: Hmm yea that makes sense. It's still not my favorite because everyone listening for these events on some parent like the tree-view is going to have to do work to determine which item is being selected and which is being deselected, and that the handler will be invoked twice for each *act of selection*, but theoretically they should already be doing that. I no longer think it's breaking though.

scomea: I'm thinking we should add a selection changed event on the tree view for folks who don't want don't want to be so atomic about it.  Ie.  we'd emit it once when the selected item has changed and all the child item state is set.

rajsite:
I'll just poke my head in and mention that one thing we have been implementing on top of the foundation tree-view and tree-item is different selection modes. For example in one selection mode an item is always selected and you can't have an unselected state (think like a radio button group). In another we only have leaf nodes be selectable (think a tree to show items inside categories where the categories are tree-items that are not selectable, only the leaf tree-items are selectable).

Currently we rely on listening for selected events and doing things like stopping propagation or re-firing an event elsewhere. I haven't thought about exactly what it would look like but it would be nice to have a cleaner way to override selection logic and implement our own behavior.

scomea:
> Currently we rely on listening for selected events and doing things like stopping propagation or re-firing an event elsewhere. 

You'll want to pay attention to the changes in this pr as it does change how the events around selection works.   Notably every tree item emits when it changes selected state, either true or false, and the bug where selection changed events were emitted prior to the value for selected being actually changed is now fixed.  

> I haven't thought about exactly what it would look like but it would be nice to have a cleaner way to override selection logic and implement our own behavior.

Well, this change does disentangle selection from ui events such that you should be able to intercept clicks/keystrokes on the tree items yourself, mark the event with preventDefault so tree view ignores it and then set the selected state of the items yourself, which should now trigger selected-change events which Tree view should recognize.  Ideally devs should be able to control a tree view by setting expanded/selected/etc... directly on the tree items.  Does that align to what you're needing?

rajsite:
> you should be able to intercept clicks/keystrokes on the tree items yourself

I think what I was imagining was a higher-level abstraction where we wouldn't have to be aware of the details of the user interactions (intercepting specific click or keyboard events; dealing with things like not breaking links in tree-items has been finicky to have to consider) but instead be able to able implement / override specific actions. ie ToggleSelect would be invoked by space / enter / click / focus capture and I can override the action and reject that selection and select something else or allow it. Other overridable actions scanning the events handled might be SelectPrevious / SelectNext / SelectFirst / SelectLast, etc.

> mark the event with preventDefault so tree view ignores it and then set the selected state of the items yourself, which should now trigger selected-change events which Tree view should recognize. Ideally devs should be able to control a tree view by setting expanded/selected/etc... directly on the tree items. Does that align to what you're needing?

It's very possible we can migrate our implementation from the old event behavior to the new behavior. I think it's difficult to know what issues we'll have without trying just trying it. Unfortunately I don't have an estimate of when we can play with it.

rajsite:
Also, a bug from our team related to tree events and disabled states if it's interesting for this PR: https://github.com/microsoft/fast/issues/5322

scomea:
> Also, a bug from our team related to tree events and disabled states if it's interesting for this PR: #5322

Yes, that should be addressed by this change.

chrisdholt:
> > Also, a bug from our team related to tree events and disabled states if it's interesting for this PR: #5322
> 
> Yes, that should be addressed by this change.

@scomea you and @eljefe223 can race; he's fixing all disabled instances, but obviously whoever gets there first will resolve. Either way, it's in process :)

packages/web-components/fast-foundation/src/tree-item/tree-item.ts
atmgrifter00: @scomea, just so we're clean, it would now be considered an anti-pattern (or, more aptly, _incorrect_) to, in a derived `TreeItem`, handle the `selected-change` event and change the current `selected` state to something else. The reason for doing this would be to prevent a user from de-selecting a currently selected tree item (by clicking on it).

Instead, it seems like the correct implementation for this behavior would be to override the `click` event, update the `selected` state as necessary, and stop the propagation of it if needed. Is this right?

scomea: @atmgrifter00 I generally consider calling stopPropagation to bad form and would call preventDefault on the initial click event instead.  The parent tree should ignore click events marked as default prevented. But yes, authors should be able to listen from click events on tree items, handle it with their own code and set the selected state of tree items as they see fit. The parent tree should update to reflect the selected state changes on child tree items.

atmgrifter00:
Any update on when this might go through? Thanks!

EisenbergEffect (approved):

change/@microsoft-fast-foundation-96d671a6-621c-4790-8217-b319bdc883d3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "tree view events",
+  "packageName": "@microsoft/fast-foundation",
+  "email": "scomea@microsoft.com",
+  "dependentChangeType": "patch"
```
nicholasrice, line 6: Should this be *minor* since, while it fixes a broken behavior, still introduces a new behavior?

nicholasrice (approved):
Approved with one question above.

change/@microsoft-fast-foundation-96d671a6-621c-4790-8217-b319bdc883d3.json
chrisdholt: @scomea this change should be all that's needed IMO

chrisdholt (approved):

Issue #5339, opened 2021-10-28
Turn on `noImplicitAny` and `strictPropertyInitialization` for `fast-components`


chrisdholt:
As part of #5303 we turned off @typescript-eslint/typedef globally as it was overwritten across multiple packages. As part of that, we turned on noImplicitAny and strictPropertyInitialization for all packages. Turning this on revealed several issues related to both tsconfig settings in fast-components and so it was turned off explicitly for that package. The errors should be fixed and addressed as part of this issue and the code should be tested to ensure that there are no breaks or side effects of the change. This issue exists to track turning that back on and resolving the errors.

m4thieulavoie:
@chrisdholt I'll look at this one too 👍

m4thieulavoie:
@chrisdholt https://github.com/microsoft/fast/pull/5354 is ready for review too :ok_hand:

PR #5354, opened 2021-11-01
remove noImplicitAny from fast-components

m4thieulavoie:
# Pull Request

## 📖 Description

This PR removes the `noImplicitAny` tsc rule in `fast-components`. Mostly straight forward, but I'm maybe not 100% sure about some of my changes. Feel free to challenge or propose changes 👍 

### 🎫 Issues

(1/2) for https://github.com/microsoft/fast/issues/5339

### General

### Component-specific


packages/web-components/fast-components/src/data-grid/data-grid.stories.ts
```diff
@@ -387,7 +387,7 @@ function cellTemplateButtonClick(cell: DataGridCell): void {
     ) {
         return;
     }
-    const newRowData: object = { ...cell.rowData };
+    const newRowData: any = { ...cell.rowData };
```
m4thieulavoie, line 390: I left `any` here since this is a storybook file, but I can definitely update if someone has a cleaner approach 👍

packages/web-components/fast-components/src/design-system-provider/index.ts
```diff
@@ -127,7 +127,7 @@ const backgroundStyles = css`
 
 function designToken<T>(token: DesignToken<T>) {
     return (source: DesignSystemProvider, key: string) => {
-        source[key + "Changed"] = function (
+        (source as any)[key + "Changed"] = function (
```
m4thieulavoie, line 130: This one is probably the one I wish I could fix the most, but really couldn't think of any cleaner way 😢

packages/web-components/fast-components/src/select/select.pw.spec.ts
```diff
@@ -200,15 +200,15 @@ describe("FASTSelect", function () {
...
+                            await this.page.exposeFunction("sendEvent", (type: string) =>
                                 expect(type).to.equal(eventName)
                             );
 
                             await element.evaluate((node, eventName) => {
                                 node.addEventListener(
                                     eventName,
                                     async (e: CustomEvent) =>
-                                        await window["sendEvent"](e.type)
+                                        await (window as any)["sendEvent"](e.type)
```
m4thieulavoie, line 211: Since this is a test file, I assume `any` is okay-ish here. Feel free to propose something cleaner 👍

packages/web-components/fast-components/src/data-grid/data-grid.stories.ts
```diff
@@ -215,7 +215,7 @@ const buttonCellTemplate = html<DataGridCell>`
                 x.columnDefinition === null ||
                 x.columnDefinition.columnDataKey === null
                     ? null
-                    : x.rowData[x.columnDefinition.columnDataKey]}
+                    : x.rowData[x.columnDefinition.columnDataKey as keyof object]}
```
radium-v, line 218: This should be a safer type casting:
```suggestion
                    : x.rowData[
                          x.columnDefinition.columnDataKey as keyof typeof x.rowData
                      ]}
```

It looks like TS is treating `object` as `{}`, as evidenced by the emitted error:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
  No index signature with a parameter of type 'string' was found on type '{}'.ts(7053)
```

@chrisdholt perhaps we could set `"suppressImplicitAnyIndexErrors": true` which would disable these errors in the tsconfig? Casting as `object` or `keyof typeof x.rowData` seems a little excessive.

packages/web-components/fast-components/src/data-grid/data-grid.stories.ts
radium-v: I think this, and the cast `as object` above point to `rowData` being incorrectly typed in `DataGridCell`. Something like `Record<string, unknown>` should work here:

```suggestion
    const newRowData: Record<string, unknown> = { ...cell.rowData };
```

@scomea is there a better type for `DataGridCell.rowData` than `object | null`?

packages/web-components/fast-components/src/design-system-provider/index.ts
radium-v: Suppressing this implicit casting error by explicitly casting `as any` seems like a bug. The further I go the more I think setting `"suppressImplicitAnyIndexErrors": true` may be the better option.

radium-v (changes requested):

packages/web-components/fast-components/src/data-grid/data-grid.stories.ts
m4thieulavoie: I agree with above ^ Having to do the above seems a tad aggressive 😅  I'll wait for the discussion to be had before tackling this specific change

scomea:
Any change in the typing of rowData would likely have to start with changing the type of base grid's 'rowsData' type which is 
```public rowsData: object[] = [];```

File: packages/web-components/fast-components/src/anchor/index.ts [hunk 5]
```diff
@@ -50,7 +50,7 @@ export class Anchor extends FoundationAnchor {
      * @internal
      *
      */
-    public defaultSlottedContentChanged(oldValue, newValue): void {
+    public defaultSlottedContentChanged(oldValue: string, newValue: string): void {
```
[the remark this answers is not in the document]
nicholasrice: wouldn't it be `HTMLElement[] | undefined`?

packages/web-components/fast-components/src/button/index.ts
```diff
@@ -43,7 +43,7 @@ export class Button extends FoundationButton {
      * @public
      * @remarks
      */
-    public defaultSlottedContentChanged(oldValue, newValue): void {
+    public defaultSlottedContentChanged(oldValue: string, newValue: string): void {
```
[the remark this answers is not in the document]
nicholasrice: same question as above, don't we know this is either `undefined` (un-initialized) or `HTMLElement[]` (initialized)?

packages/web-components/fast-components/src/anchor/index.ts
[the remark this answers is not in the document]
radium-v: `HTMLElement[] | undefined` would be better. I suggested `unknown` in this case since the underlying value for `defaultSlottedContent` isn't filtered until it reaches this function, and it's actually `Node[]` when coming from the `slotted` directive. The filter in this function should either be moved to the `slotted` call in the template, or the property `defaultSlottedContent` should be changed to type `Node[]` in the base class.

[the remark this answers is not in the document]
radium-v: On further investigation, this specific function is handling cases where only a single `SVGElement` is present, so `HTMLElement[]` isn't the right type.

packages/web-components/fast-components/src/anchor/anchor.pw.spec.ts
```diff
@@ -14,7 +14,7 @@ describe("FASTAnchor", function () {
         this.documentHandle = await this.page.evaluateHandle(() => document);
 
         this.setupHandle = await this.page.evaluateHandle(
-            (document) => {
+            (document: any) => {
```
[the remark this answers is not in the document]
m4thieulavoie: Yep, that did the trick. Thanks for that!

EisenbergEffect (approved):

packages/web-components/fast-components/src/anchor/index.ts
```diff
@@ -50,7 +50,7 @@ export class Anchor extends FoundationAnchor {
      * @internal
      *
      */
-    public defaultSlottedContentChanged(oldValue, newValue): void {
+    public defaultSlottedContentChanged(oldValue: unknown, newValue: SVGElement): void {
```
bheston, line 53: The slot content is _often_ an SVG, but this should not be required. The slots should allow any content, including plain text or any other component.

Also, does the `unknown / something` convention assume that it's always null or undefined and then it's _something_? What if the content is unset?

chrisdholt: Unknown is actually unknown :) - not a great explanation, I know. It’s actually semantic in Typescripts case. In terms of the new value, agree that this could be anything. In the method we are filtering to apply logic when it’s only an svg, but it could be any HTML Element or a text node..

chrisdholt:
@m4thieulavoie I'm going to close this given the future of fast components, probably the easiest way to resolve any outstanding questions :). Thanks!

PR #5355, opened 2021-11-02
remove strictPropertyInitialization in components

m4thieulavoie:
# Pull Request

## 📖 Description

This PR removes the `strictPropertyInitialization` rule from `fast-component`'s `tsconfig` file.

### 🎫 Issues

https://github.com/microsoft/fast/issues/5339

### General

### Component-specific


packages/web-components/fast-components/src/avatar/index.ts
```diff
@@ -20,7 +20,7 @@ export class Avatar extends FoundationAvatar {
      * HTML Attribute: src
      */
     @attr({ attribute: "src" })
-    public imgSrc: string;
+    public imgSrc = "";
```
robarbms, line 23: Any reason to not keep the typings, especially initialized, they won't be null or undefined?
```suggestion
    public imgSrc: string = "";
```

robarbms (approved):

packages/web-components/fast-components/src/avatar/index.ts
m4thieulavoie: I was unsure on what y'all prefer on this one (I've been at companies that prefer either one). Happy to put it back 😄

m4thieulavoie:
```suggestion
    public imgSrc: string = "";
```

packages/web-components/fast-components/src/avatar/index.ts
```diff
@@ -29,7 +29,7 @@ export class Avatar extends FoundationAvatar {
      * @remarks
      * HTML Attribute: alt
      */
-    @attr public alt: string;
+    @attr public alt = "";
```
m4thieulavoie, line 32:
```suggestion
    @attr public alt: string = "";
```

packages/web-components/fast-components/src/disclosure/index.ts
```diff
@@ -17,11 +17,11 @@ export class Disclosure extends FoundationDisclosure {
     /**
      * Disclosure default height
      */
-    private height: number;
+    private height = 0;
```
m4thieulavoie, line 20:
```suggestion
    private height: number = 0;
```

packages/web-components/fast-components/src/disclosure/index.ts
```diff
@@ -17,11 +17,11 @@ export class Disclosure extends FoundationDisclosure {
     /**
      * Disclosure default height
      */
-    private height: number;
+    private height: number = 0;
     /**
      * Disclosure height after it's expanded
      */
-    private totalHeight: number;
+    private totalHeight = 0;
```
m4thieulavoie, line 24:
```suggestion
    private totalHeight: number = 0;
```

packages/web-components/fast-components/src/design-system-provider/index.ts
```diff
@@ -188,7 +188,7 @@ export class DesignSystemProvider extends FoundationElement {
         converter: swatchConverter,
     })
     @designToken(fillColor)
-    public fillColor: Swatch;
+    public fillColor: Swatch = fillColor.getValueFor(this);
```
nicholasrice, line 191: Doing this is going to cause a lot of un-necessary work for each design-system-provider instance so I don't think we should do this. 

I'd propose updating the property types to allow for undefined, thus not requiring initialization (at least I think that is how it works).  If that *doesn't* work, I'd propose disabling that rule for this file.

m4thieulavoie: That should be done 👍

m4thieulavoie:
@chrisdholt can you have a look at this one? :pray:

packages/web-components/fast-components/src/avatar/index.ts
```diff
@@ -20,7 +20,7 @@ export class Avatar extends FoundationAvatar {
      * HTML Attribute: src
      */
     @attr({ attribute: "src" })
-    public imgSrc: string;
+    public imgSrc: string = "";
```
chrisdholt, line 23: Should this perhaps be null or undefined?

chrisdholt (approved):
Would be great to get @nicholasrice or @EisenbergEffect’s eyes in this as well. Thanks @m4thieulavoie!

packages/web-components/fast-components/src/avatar/index.ts
m4thieulavoie: Yeah good call, it should be consistent with the remaining of the PR :+1:

nicholasrice (approved):

packages/web-components/fast-components/src/anchor/index.ts
```diff
@@ -25,7 +25,7 @@ export class Anchor extends FoundationAnchor {
      * HTML Attribute: appearance
      */
     @attr
-    public appearance: AnchorAppearance;
+    public appearance: AnchorAppearance = "neutral";
```
EisenbergEffect, line 28: For some reason I thought we couldn't do this and we had to check in the connectedCallback. Maybe it was an old bug that was fixed...I though there was some sort of timing issue with upgrades or binding. Something like that. @nicholasrice Does any of this ring a bell for you?

nicholasrice: hmmm maybe? I didn't see the classList mutation, that may be why it was done this way. It is probably best to keep the assignment in the property initialization and just move the class to the template.

EisenbergEffect (approved):

PR #5356, opened 2021-11-02
fix type compilations and dupes for style exports, update picker styles to expected convention

### LINKED ISSUE #5346: fix: component style exports cause type complications in fast-components
# 🐛 Bug Report

For `fast-components`, we're exporting the final style constants by creating a new exported const when we haven't changed anything in the original alias. I haven't come across any components which modify their aliased `styles` import, so I don't know if the reassignment is necessary in most cases.

## 💻 Repro or Code Sample

```ts
import { componentStyles as styles } from "./component.styles"; // <- aliased import

export const fastComponent = Component.compose({
    baseName: "component",
    template,
    styles, // <- object property shortcut
    // ...
})

// this doc block should move to ./component.styles.ts
/**
 * Styles for component
 * @public
 */
export const componentStyles = styles; // <- creates duplicated type definition
```

## 😯 Current Behavior

Duplicated type definitions are presented in both `index.ts` as well as in the `*.style.ts` file when built. Also, since the tsdoc block is only present in `index`, meaning the entire style file effectively has no documentation.

## Suggested Fix

We should change the style export to use `as`  instead of creating a new const, and the tsdoc comment should be moved to the `*.styles.ts` file:

```ts
import { componentStyles as styles } from "./component.styles";

export const fastComponent = Component.compose({
    baseName: "component",
    template,
    styles,
    // ...
})

export { styles as componentStyles };

// alternatively, we could do this:
// export { componentStyles } from "./component.styles";
```

```ts
/**
 * Styles for {@link fastComponent}
 * @public
 */
export const componentStyles = // ...
```

chrisdholt:
# Pull Request

### 🎫 Issues
Fixes #5346

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/picker/index.ts
+++ b/packages/web-components/fast-components/src/picker/index.ts
@@ -33,12 +33,6 @@ export const fastPicker = Picker.compose({
 });
 
 /**
- * Styles for Picker
- * @public
- */
-export const PickerStyles = pickerStyles;
-
-/**
  * Base class for Picker
  * @alpha
  */
@@ -59,12 +53,6 @@ export const fastPickerMenu = PickerMenu.compose({
 });
 
 /**
- * Styles for PickerMenu
- * @public
- */
-export const PickerMenuStyles = pickerMenuStyles;
-
-/**
  *  Component that displays available picker menu options
  *
  *
@@ -77,12 +65,6 @@ export const fastPickerMenuOption = PickerMenuOption.compose({
     template: pickerMenuOptionTemplate,
     styles: pickerMenuOptionStyles,
 });
-
-/**
- * Styles for PickerMenuOption
- * @public
- */
-export const PickerMenuOptionStyles = pickerMenuOptionStyles;
 
 /**
  * Component that displays the list of selected picker items along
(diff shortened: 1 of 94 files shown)


packages/web-components/fast-components/src/picker/index.ts
```diff
@@ -118,8 +94,4 @@ export const fastPickerListItem = PickerListItem.compose({
     styles: pickerListItemStyles,
 });
 
-/**
- * Styles for PickerListItem
- * @public
- */
-export const PickerListItemStyles = pickerListItemStyles;
+export { pickerStyles, pickerListItemStyles, pickerMenuOptionStyles, pickerMenuStyles };
```
chrisdholt, line 97: *Technically* this is a break. These are functions and should be exported as such. Picker itself is in alpha, but these were marked public. I'm happy to add deprecated versions of those removed here.

EisenbergEffect (approved):

packages/web-components/fast-components/src/picker/index.ts
EisenbergEffect: Probably not an issue. I don't think picker is documented anywhere, so likely folks don't really know this exists.

robarbms (approved):

packages/web-components/fast-components/src/picker/index.ts
chrisdholt: My thought as well 👍

radium-v (approved):

PR #5362, opened 2021-11-03
update workflows for consistency to match new implementations on new repositories

awentzel:
# Pull Request

## 📖 Description
Brings over more intuitive workflow naming to speed up finding status and notification service to discord for all continuous delivery workflows.

### 🎫 Issues
* Some users were not totally clear on what CD or CI acronyms are
* Some names are inherent. For example, we know it's FAST, why use it in the name
* Notifications were never working, this implements a known working solution

## 👩‍💻 Reviewer Notes
Relatively low-risk initial changes such that contributors or consumers of FAST have consistent paradigms in place when searching in our workflows, GitHub Actions.

## 📑 Test Plan
FAST Creator and FAST Labs are good sources to see these same changes working.

### General

### Component-specific

### DIFF
--- a/.github/workflows/cd-color-stage.yml
+++ b/.github/workflows/cd-color-stage.yml
@@ -1,4 +1,4 @@
-name: CD - Color Explorer
+name: Deploy Color Explorer
 
 on:
   push:
@@ -25,12 +25,12 @@ jobs:
 
     - name: Install package dependencies / prepare workspaces
       run: yarn install --frozen-lockfile
-    
+
     - name: Build & Prepare Web Application
       run: |
         cd ${{ env.AZURE_WEBAPP_BUILD_PATH }}
         yarn build
-        cp ../site-utilities/statics/server/package.json ../site-utilities/statics/server/server.js www 
+        cp ../site-utilities/statics/server/package.json ../site-utilities/statics/server/server.js www
         cd www
         npm i
         ls -lta
--- a/.github/workflows/cd-color-stage.yml
+++ b/.github/workflows/cd-color-stage.yml
@@ -63,4 +63,4 @@ jobs:
         webhook_token: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_TOKEN }}
         color: '#DE2D6D'
         username: 'FAST DevOps Color Explorer Bot'
-        message: 'Deployment has completed to Staging on https://github.com/microsoft/fast/actions/workflows/cd-publish.yml'
+        message: 'Deployment has completed to Staging on https://github.com/microsoft/fast/actions/workflows/cd-color-stage.yml'
--- a/.github/workflows/cd-color-stage.yml
+++ b/.github/workflows/cd-color-stage.yml
@@ -50,3 +50,17 @@ jobs:
         app-name: ${{ env.AZURE_WEBAPP_PASSIVE_STAGE_NAME }}
         package: ${{ env.AZURE_WEBAPP_DIST_PATH }}
         slot-name: ${{ env.AZURE_WEBAPP_SLOT_NAME }}
+
+  notify:
+    runs-on: ubuntu-latest
+    needs: build_deploy_linux
+
+    steps:
+    - name: Notify on Discord
+      uses: appleboy/discord-action@master
+      with:
+        webhook_id: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_ID }}
+        webhook_token: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_TOKEN }}
+        color: '#DE2D6D'
+        username: 'FAST DevOps Color Explorer Bot'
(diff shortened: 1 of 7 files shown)


.github/workflows/cd-publish.yml
```diff
@@ -1,4 +1,4 @@
-name: CI - FAST Publish
+name: Release Packages
```
chrisdholt, line 1:
```suggestion
name: Release NPM Packages
```
If we're looking for clarity I might suggest being specific about the type of packages.

.github/workflows/cd-static-cdn.yml
```diff
@@ -8,15 +8,29 @@ on:
...
+
+    steps:
+    - name: Notify on Discord
+      uses: appleboy/discord-action@master
+      with:
+        webhook_id: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_ID }}
+        webhook_token: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_TOKEN }}
+        color: '#DE2D6D'
+        username: 'FAST DevOps CDN Bot'
+        message: 'Deployment has completed to CDN on https://github.com/microsoft/fast/actions/workflows/cd-static-cdn.yml'
```
chrisdholt, line 36: Do these (various releases) all run simultaneously? Do we have any concerns about the bots being noisy?

.github/workflows/ci-daily.yml
```diff
@@ -1,4 +1,4 @@
-name: CI - FAST
+name: Validate
```
chrisdholt, line 1: Is Validate clear enough? Is this just validating no breaks on the main branch?

chrisdholt (approved):

.github/workflows/cd-static-cdn.yml
awentzel: This notifies our private #ms-internal-notifcations channel which may be noisy though only delivery workflows trigger notifications.

Pros:
* I will be notified on delivery and can easily click a link to view pass/fail of deployments to proactively resolve to breaks or blocking issues.
* Anyone can easily click and become familiar with how we release and in what workflows
* Elevates visibility for leadership curious on what's most recently deployed.
* Can easily be ignored in this channel for those not concerned.
* It's almost like a breaking news feed, I suppose.

awentzel: I believe we could also do some updates to include the event, the branch, and path whereby the path would could be restricted to only deploy if files required or dependencies of a site are changed.  So this would reduce noise if it became a problem.

.github/workflows/ci-daily.yml
awentzel: I believe so, and yes.

.github/workflows/cd-static-cdn.yml
chrisdholt: Thanks for the follow up @awentzel. This is probably fine for now, but I do like the enhancement above. Would be great over time to have builds conditional on changed dependencies for sites and perhaps packages, etc.

.github/workflows/cd-color-stage.yml
```diff
@@ -50,3 +50,17 @@ jobs:
...
+
+    steps:
+    - name: Notify on Discord
+      uses: appleboy/discord-action@master
+      with:
+        webhook_id: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_ID }}
+        webhook_token: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_TOKEN }}
+        color: '#DE2D6D'
+        username: 'FAST DevOps Color Explorer Bot'
+        message: 'Deployment has completed to Staging on https://github.com/microsoft/fast/actions/workflows/cd-publish.yml'
```
janechu, line 66: Not sure if this is right, is this more accurate?

```suggestion
        message: 'Deployment has completed to Staging on https://github.com/microsoft/fast/actions/workflows/cd-color-stage.yml'
```

.github/workflows/cd-explore-stage.yml
```diff
@@ -49,4 +49,18 @@ jobs:
...
+
+    steps:
+    - name: Notify on Discord
+      uses: appleboy/discord-action@master
+      with:
+        webhook_id: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_ID }}
+        webhook_token: ${{ secrets.DISCORD_NOTIFICATION_WEBHOOK_TOKEN }}
+        color: '#DE2D6D'
+        username: 'FAST DevOps Component Explorer Bot'
+        message: 'Deployment has completed to Staging on https://github.com/microsoft/fast/actions/workflows/cd-publish.yml'
```
janechu, line 66:
```suggestion
        message: 'Deployment has completed to Staging on https://github.com/microsoft/fast/actions/workflows/cd-explore-stage.yml'
```

janechu (approved):

nicholasrice (approved):

Issue #5359, opened 2021-11-03
fix: NumberField does not visually reflect the component's 'value' when 'value' is assigned programatically


nicholasrice:
# 🐛 Bug Report
When `NumberField.value` is assigned via JavaScript, the component's `value` is not reflected to the inner `<input />` control's `value`. The effect of this is that there is no visual representation of a assigned value (or changed value). 

## 💻 Repro or Code Sample
Min repro here: https://stackblitz.com/edit/typescript-7hcdkp?file=index.ts

## 🤔 Expected Behavior
Assigning a `value` to the control via JavaScript should visually reflect the assigned `value`.

radium-v:
The `control.value` doesn't seem to update to match the value when it's forced to change after the min/max range changes:

1. set `min` to 0 and `max` to 100
2. set `value` to 100
3. set `max` to 10

The `value` and `currentValue` return 10 but the `control.value` stays at 100 until I interact with the control.

PR #5368, opened 2021-11-05
fix: enables setting number-field value through script

robarbms:
# Pull Request

## 📖 Description

Updating the number-field value with script was not updating the value displayed in the number-field text box.

### 🎫 Issues

https://github.com/microsoft/fast/issues/5359

## 👩‍💻 Reviewer Notes

I had to enabling updating the text fields value when the number-fields value had changed. I had to prevent it from changing the text fields value while the user was typing as they need to be able to enter invalid values while typing. For example the text fields value can be `-0.` while the number-fields value will be `0` so that they can enter `-0.5` without having the number-field auto-correcting them.

I also set up a flag to prevent valueChanged() calling back into itself while updating this.value.

## 📑 Test Plan

I created a test for changing the number-fields value through script.

### General

### Component-specific


EisenbergEffect (approved):

radium-v:
The `control.value` doesn't seem to update to match the value when it's forced to change after the min/max range changes:

1. set `min` to 0 and `max` to 100
2. set `value` to 100
3. set `max` to 10

The `value` and `currentValue` return 10 but the `control.value` stays at 100 until I interact with the control.

nicholasrice (approved):

robarbms:
> The `control.value` doesn't seem to update to match the value when it's forced to change after the min/max range changes:
> 
> 1. set `min` to 0 and `max` to 100
> 2. set `value` to 100
> 3. set `max` to 10
> 
> The `value` and `currentValue` return 10 but the `control.value` stays at 100 until I interact with the control.

Good catch.

I'm adding a check if the `previousValue` is null. This is an indication that the min or max has been updated. I could make it more specific. I could call `this.valueChanged('minMaxUpdate', this.value)` from in minChanged() and maxChanged(). But I can't think of a use case where a null check would cause a problem.

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -158,28 +177,47 @@ export class NumberField extends FormAssociatedNumberField {
...
+            if (isNaN(value)) {
+                value = "";
+            } else {
+                value = Math.min(value, this.max ?? value);
+                value = Math.max(value, this.min ?? value).toString();
+            }
+
+            if (this.proxy instanceof HTMLInputElement) {
+                this.proxy.value = value;
+            }
```
radium-v, line 204: The `valueChanged` supermethod handles this

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -158,28 +177,47 @@ export class NumberField extends FormAssociatedNumberField {
...
+
+            this.value = value;
+
+            if (previousValue !== undefined) {
+                this.$emit("input");
+                this.$emit("change");
+            }
+            this.updatingValue = false;
+            super.valueChanged(previousValue, nextValue);
```
radium-v, line 220: Should the `$emit` check happen after calling the supermethod?

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -158,28 +177,47 @@ export class NumberField extends FormAssociatedNumberField {
...
-        this.value = value.toString();
-
-        if (this.proxy instanceof HTMLInputElement) {
-            this.proxy.value = this.value;
-        }
-
-        if (previousValue !== undefined) {
-            this.$emit("input");
-            this.$emit("change");
+        if (previousValue != nextValue && this.updatingValue === false) {
```
radium-v, line 192: `valueChanged` is handled as an observable, so this function is only called when `value` changes (or when `min`/`max` change). Are there any cases where this function is executed and `previousValue` and `nextValue` are equal?

packages/web-components/fast-foundation/src/number-field/number-field.ts
```diff
@@ -119,13 +119,22 @@ export class NumberField extends FormAssociatedNumberField {
...
+     * @internal
+     */
     public maxChanged(previousValue, nextValue): void {
         this.max = Math.max(nextValue, this.min ?? nextValue);
         const min = Math.min(this.min, this.max);
         if (this.min !== undefined && this.min !== min) {
             this.min = min;
         }
-        this.valueChanged(this.value, this.value);
+        this.valueChanged(null, this.value);
```
radium-v, line 137: I'm not a fan of this since `prevValue` _isn't_ null, it's the value of `this.value`. I think the contents of `valueChanged` should be extracted into their own clamping function, which can be called by `minChanged`, `maxChanged`, and `valueChanged` on an independent basis.

radium-v (changes requested):

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: It does. I just need to pass value into the supermethod so that it doesn't use next which may not be valid.

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: I was previously passing the same values into valueChanged from minChanged and maxChanged to trigger validation. You can see that in the code that I deleted.

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: It could happen after the calling super. In the case that they are submitting form on change maybe? The only thing that happens in the super is setting the proxy which is only for form submission. I'll move the emit.

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: I was previously calling `this.valueChanged(this.value, this.value)` because it was technically accurate.

The big problem is validating and changing this.value from inside this.valueChanged without re-triggering this.valueChanged.

I will abstract the validation and call into it from minChanged, maxChanged and valueChanged.

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: I moved the emits to the end of valueChanged().

packages/web-components/fast-foundation/src/number-field/number-field.ts
robarbms: I've pull out the validation code to it's own function `getValidValue()` to make sure the value can be parsed into a valid number and is between the min and max. `valueChanged()` calls into it as does `minChanged()` and `maxChanged()`. The big difference here is that typing in the text field now passes a flag to `valueChanged()` telling it to not override the contents of `this.control.value`.

EisenbergEffect:
How are we looking on this one? Would love to get this merged and released today if possible.

robarbms:
> How are we looking on this one? Would love to get this merged and released today if possible.

I think it's good. I reworked how it deals with typing so that it doesn't override user input while typing. Basically calling into valueChanged() with a flag just for text input so that the other use cases can just set this.value and the default behavior updates the text input. Much cleaner. I was just waiting on signoff from John. I'll ping him and get this in today.

radium-v (approved):

PR #5377, opened 2021-11-08
feat: update recipes, design tokens and utility files

khamudom:
# Pull Request

## 📖 Description

Before making changes to the components I am updating certain recipes, the design token, and the utility files. These are global changes to most, if not all, of the components.

Updated dark mode to 0.09, this aligns with the Figma design
Updated the accent color to use `#80DEEA`
Update `focusStrokeOuter` to 5.5
Update base height to 8
Update`cornerCornerRadius` to 6
Update deltas on  a few design-tokens.
Add layer `layerCornerRadius`. This is used on components like menu and card.
Add `contrast-set-recipe` file. This will be used in the `strokeControlStrongRecipe`

I am using the button example to give an idea of the changes
before
before
after
after

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/color/recipes/contrast-set-recipe.ts
+++ b/packages/web-components/fast-components/src/color/recipes/contrast-set-recipe.ts
@@ -1,0 +1,45 @@
+import { Palette } from "../palette";
+import { InteractiveSwatchSet } from "../recipe";
+import { Swatch } from "../swatch";
+import { directionByIsDark } from "../utilities/direction-by-is-dark";
+
+/**
+ * @internal
+ */
+export function contrastSetRecipe(
+    palette: Palette,
+    reference: Swatch,
+    baseContrast: number,
+    restDelta: number,
+    hoverDelta: number,
+    activeDelta: number,
+    focusDelta: number
+): InteractiveSwatchSet {
+    const direction = directionByIsDark(reference);
+    const accessibleIndex = palette.closestIndexOf(
+        palette.colorContrast(reference, baseContrast)
+    );
+    const accessibleIndex2 =
+        accessibleIndex + direction * Math.abs(restDelta - hoverDelta);
+    const indexOneIsRest =
+        direction === 1
+            ? restDelta < hoverDelta
+            : direction * restDelta > direction * hoverDelta;
+    let restIndex: number;
+    let hoverIndex: number;
+
+    if (indexOneIsRest) {
+        restIndex = accessibleIndex;
+        hoverIndex = accessibleIndex2;
+    } else {
+        restIndex = accessibleIndex2;
+        hoverIndex = accessibleIndex;
+    }
+
+    return {
+        rest: palette.get(restIndex),
+        hover: palette.get(hoverIndex),
+        active: palette.get(restIndex + direction * activeDelta),
+        focus: palette.get(restIndex + direction * focusDelta),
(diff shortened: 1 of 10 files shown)


packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -39,6 +40,12 @@ export type ColorRecipe = Recipe<Swatch>;
 /** @public */
 export type InteractiveColorRecipe = Recipe<InteractiveSwatchSet>;
 
+/** @public */
+export type InteractiveSet = Record<"rest" | "hover" | "active" | "focus", string>;
```
nicholasrice, line 44: what is the `string` type here supporting?

nicholasrice: nevermind I missed the comma

packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -39,6 +40,12 @@ export type ColorRecipe = Recipe<Swatch>;
...
+
+/** @public */
+export type InteractiveRecipe = Recipe<InteractiveSet>;
```
nicholasrice, line 47: I don't see either of these types used anywhere. Are they needed?

khamudom: you are correct. I will remove these for now. thanks!

packages/web-components/fast-components/src/color/recipes/contrast-set-recipe.ts
```diff
@@ -0,0 +1,45 @@
...
+    let restIndex: number;
+    let hoverIndex: number;
+
+    if (indexOneIsRest) {
+        restIndex = accessibleIndex;
+        hoverIndex = accessibleIndex2;
+    } else {
+        restIndex = accessibleIndex2;
+        hoverIndex = accessibleIndex;
+    }
```
robarbms, line 37: It's a little hard to see what is going on, but if I'm reading it correctly, if direction === 1 rest gets the lower number and hover gets the higher number. Otherwise if direction === -1 rest gets the higher number and hover gets the lower number? If that is the case, you could use Math.min and Math.max based on direction to assign these.

```suggestion
    const restIndex: number = Math[direction ? "min" : "max"](accessibleIndex1, accessibleIndex2);
    const hoverIndex: number = Math[direction ? "max" : "min"](accessibleIndex1, accessibleIndex2);
```

packages/web-components/fast-components/src/color/recipes/focus-stroke.ts
```diff
@@ -4,7 +4,7 @@ import { directionByIsDark } from "../utilities/direction-by-is-dark";
 
 /** @internal */
 export function focusStrokeOuter(palette: Palette, reference: Swatch) {
-    return palette.colorContrast(reference, 3.5);
+    return palette.colorContrast(reference, 5.5);
```
robarbms, line 7: Does the focusStrokeInner that uses 3.5 (on line18) need to be updated to 5.5?

bheston (approved):
This is the part of the visual update I suggested we would need to revisit once the color updates were locked down. I believe some of the new palette generation hasn't gone in yet. I'll follow up on that, but I wanted to unblock you for now. Worst case we revisit some deltas, but at least everything else you need is there.

Issue #5344, opened 2021-10-29
Turn on `strictNullChecks`, `strictPropertyInitialization` for `fast-web-utilities` in tsconfig


chrisdholt:
As part of #5303 we turned off `@typescript-eslint/typedef` globally as it was overwritten across multiple packages. As part of that, we turned on `strictNullChecks` and `strictPropertyInitialization` for all packages. Turning this on revealed several issues related to a few tests and function implementations in `fast-web-utilties` and so it was turned off explicitly for that package. The errors should be fixed and addressed as part of this issue and the code should be tested to ensure that there are no breaks or side effects of the change. This issue exists to track turning that back on and resolving the errors.

NiteshSethi09:
Hey @EisenbergEffect, may you assign me this issue. I'd like to work on this!

NiteshSethi09:
Hi @chrisdholt! I've successfully made changes for this issue and the code is merged in the codebase. May you please close this issue?

NiteshSethi09:
Thank you for the information @chrisdholt.

PR #5378, opened 2021-11-09
Turn on strictNullChecks, strictPropertyInitialization for fast-web-utilities

alphashr:
# Pull Request

Couple of changes to be able to turn on ```strictNullChecks```, ```strictPropertyInitialization``` in fast-web-utilities ```tsconfig.json```.

### 🎫 Issues

#5344 

## 👩‍💻 Reviewer Notes

None

## 📑 Test Plan

None

### General

### Component-specific

## ⏭ Next Steps
None


packages/utilities/fast-web-utilities/src/rtl-scroll-converter.spec.ts
```diff
@@ -182,8 +182,8 @@ describe("RtlScrollConverter", (): void => {
     });
 
     it("applyDirectScrollConverters applies correct converters", () => {
-        RtlScrollConverter["getRtlScrollLeftConverter"] = null;
-        RtlScrollConverter["setRtlScrollLeftConverter"] = null;
+        RtlScrollConverter["getRtlScrollLeftConverter"] = RtlScrollConverter["reverseGetRtlScrollConverter"];
```
alphashr, line 185: Setting ```getRtlScrollLeftConverter``` and ```setRtlScrollLeftConverter``` to something other than the defaults (```directGetRtlScrollConverter``` and ```directSetRtlScrollConverter``` respectively)  to test the functionality of ```applyDirectScrollConverters()```

packages/utilities/fast-web-utilities/src/numbers.ts
```diff
@@ -17,6 +25,13 @@ export function wrapInBounds(min: number, max: number, value: number): number {
  * Ensures that a value is between a min and max value. If value is lower than min, min will be returned.
  * If value is greater than max, max will be returned.
  */
-export function limit(min: number, max: number, value: number): number {
+export function limit(
+    min: number | null,
```
nicholasrice, line 29: It seems that these changes and the changes in `query.ts` are only necessary because we have tests that invoke with invalid arguments.

I think we should probably just remove those tests and rely on the compiler to ensure arguments are accurate. If an author provides invalid arguments that isn't the responsibility of these functions to guard against.

packages/utilities/fast-web-utilities/src/strings.ts
```diff
@@ -19,7 +19,7 @@ export function format(formatSpecifier: string, ...parameters: string[]): string
             return match;
         }
 
-        const value: string = parameters[index];
+        const value = parameters[index];
```
alphashr, line 22: Intentional.

packages/utilities/fast-web-utilities/src/numbers.ts
nicholasrice: What I mean is that we shouldn't change these function signatures. Can you revert them and adjust the tests as necessary so that tests compile? Thanks!

nicholasrice (approved):
Thanks!

robarbms (approved):

raghavthind2005 (approved):

bheston:
While the change was handled slightly differently, the core issue has already been addressed in #6662. I confirmed there is significant overlap between this change and that merged solution.

PR #5167, opened 2021-09-08
chore: update visual to web components

khamudom:
# Pull Request

## 📖 Description

This PR is a visual update to the components. Below is a brief overview of what has been done, to hopefully help give an idea of the changes that have been done. 
But please run the branch, if you can, as that will also give you a better look and feel of the components. Feedbacks are welcome.

The majority of the changes on the components are style updates to using the design tokens on different states, like rest, hover, active, and focus. Some of the components are also "grouped" in a way, where they are using similar tokens because from a base level, they look and interact the same. For example, checkbox, radio, and switch are smaller custom inputs so they use the same tokens on the border and background. Another is the menu and list flyouts (from select or combo box) have the same background tokens. Those are just a few examples.
Also updated some SVGs and fixture examples.

- To make sure the colors are in sync with the design, I changed the dark mode in `base-layer-luminance.ts` from `0.23` to `0.09`(#181818).
- Updated the base accent on `color-constants.ts`

- Here are some changes to some recipe files (src/color/recipes)
updated `focus-stroke.ts` changing the color contrast from `3.5` to `5.5`
updated `neutral-fill.ts` changing the direction to use `directionByIsDark(reference)`
updated `neutral-foreground.ts` adding hover, active and focus deltas
updated `neutral-foreground.spec.ts`
added `contrast-set-recipe.ts` - This recipe is used on `stroke-control-strong-recipe`

- Here are some changes to some design token files (src/design-tokens)
updated `baseHeightMultiplier` to set it from 10 to 8
updated `controlCornerRadius` to set it from 4 to 6
updated `layerCornerRadius` new corner radius, set to 6, used on `card` and `menu`
updated `neutral-fill` deltas to (rest: 4, hover: 4, active: 10, focus: 5)
updated `neutral-fill-input-active` delta to (-10)
updated `neutral-fill-stealth` deltas to (hover: 4, active: 10, focus: 4)
updated `neutral-fill-layer-rest` delta to (2)
updated `neutral-stroke` deltas (rest: 49, hover: 49, active: 90, focus: 49)
updated `neutral-foreground-recipe` to create deltas for hover, active, focus
added `neutral-stroke-input-fill` this token is used on inputs with filled appearance
added `stroke-control-strong` this token is used checkbox, radio, slider, outline button, and switch 

### General

- [x] I have included a change request file using `$ yarn change`
- [ ] I have added tests for my changes.
- [x] I have tested my changes.
- [ ] I have updated the project documentation to reflect my changes.
[...]

PR #5382, opened 2021-11-11
chore: visual refresh on buttons and anchors

khamudom:
# Pull Request

## 📖 Description

The default height has been updated to 32px and the corner radius is 6px. The focus border has also been updated to 3px.

Before and After
The order down is rest, hover, active, focus, and disabled states.

Default/Neutral
image

Accent
image

Lightweight
image

Outline
image

Stealth
image

The visual for Anchor is the same for neutral, accent, lightweight, outline, and stealth, except Anchor, does not have a disabled state.

Before and After for Anchor Hypertext
image

 
### General

### Component-specific


janechu:
Should the focus outline (or border) for the anchor be that tight?

khamudom:
> Should the focus outline (or border) for the anchor be that tight?

Yeah, I personally think it is too tight. I kept it as is because there is no design for hypertext. I think we can just keep focus with an underline, which is the current visual. @bheston any suggestion? But for now, I will update to an underline, unless there are any objections or other ideas.

khamudom:
hypertext updated
image

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -38,457 +45,418 @@ import {
...
+                  color: ${SystemColors.HighlightText};
+              }
+              :host .control:${focusVisible} {
+                  forced-color-adjust: none;
+                  border-color: ${SystemColors.ButtonText} !important;
+                  box-shadow: 0 0 0 calc(${focusStrokeWidth} * 1px) ${SystemColors.ButtonText} !important;
+                  background: ${SystemColors.ButtonFace};
+                  color: ${SystemColors.ButtonText};
+              }
+              :host([href]) .control {
```
robarbms, line 146: I couldn't find href used in button.ts or in button.template.ts. Are these shared styles or should all of the :host([href]) styles be in anchor.styles.ts? I get that you didn't put them here, just a broader question. I'll mark this comment as resolved.

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -38,457 +45,418 @@ import {
...
-              fill: currentColor;
-            }
-        `
-    )
-);
+export const BaseButtonStyles = (
+    context: ElementDefinitionContext,
+    definition: FoundationElementDefinition,
+    interactivitySelector: string = ""
+) =>
```
robarbms, line 52: You are calling out the opposite of the interactivitySelector explicitly in buton.styles.ts and anchor.styles.ts above. Would it make sense to create that here and move the styles into BaseButtonStyles?

```
const inactiveSelector = interactivitySelector.indexOf(":not") >= 0 ? 
    interactivitySelector.replace(/(.*):not\(([^)]*)\)(.*)/, '$1$2$3') : 
    `:not(${interactivitySelector})`; 
```

robarbms (commented):
There are are a lot of !important statements included. Is there anyway to remove them. As soon as you add 1, you need to add a hundred. It turns into an arms race for CSS selectors and becomes really hard to manage. Another drawback to !important is that it becomes impossible to override it from inline. Inline styles will supersede stylesheet styles, but for !important statements, the opposite is true. Which means that a `style="color: #000!important"` will not override a `:host { color: #fff!important; }` in a stylesheet.

janechu:
Who's approvals should we consider as necessary (blocking) for these PRs? @bheston? I just want to make sure if I sign off on this, I am not the "authority" on it as I'm not aware of all of the style decisions that have been made.

khamudom:
> Who's approvals should we consider as necessary (blocking) for these PRs? @bheston? I just want to make sure if I sign off on this, I am not the "authority" on it as I'm not aware of all of the style decisions that have been made.

As far as the visual approval, with an exception of a few components like Picker and Calendar, before Jason Reynolds left, we went through all the components and he approved the visual changes. Then I created the PR that was up for review (https://github.com/microsoft/fast/pull/5167) 
Now that we are breaking up the changes into smaller PRs and adding them to a feature branch, maybe the approval is more about the code and not the visual. If we still accept Jason Reynolds as the person who signs off on the visual refresh since that has not changed in these PRs. Not to say, we can't give feedback on the visual, because that is also welcomed. For example, the focus on the hypertext button. That was good feedback. Just my thoughts.

EisenbergEffect:
I think @bheston and @chrisdholt should probably take a look at and be the ones to approve the visual refresh PRs.

packages/web-components/fast-components/src/button/button.styles.ts
```diff
@@ -31,129 +27,58 @@ export const buttonStyles: (
...
-                    color: ${SystemColors.GrayText};
-                    cursor: ${disabledCursor};
+                :host([disabled]) .control {
                     opacity: 1;
+                    background: ${SystemColors.ButtonFace};
+                    border-color: ${SystemColors.GrayText} !important;
+                    box-shadow: none !important;
+                    color: ${SystemColors.GrayText};
+                    fill: currentcolor;
                 }
```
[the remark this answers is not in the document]
khamudom: using `!important` is not necessary, I will remove them

packages/web-components/fast-components/src/styles/patterns/button.ts
khamudom: I think refactoring to move, the selectors with `href`, into anchor style, is a good idea.

janechu (approved):

bheston:
> > Should the focus outline (or border) for the anchor be that tight?
> 
> Yeah, I personally think it is too tight. I kept it as is because there is no design for hypertext. I think we can just keep focus with an underline, which is the current visual. @bheston any suggestion? But for now, I will update to an underline, unless there are any objections or other ideas.

I would really love for text underline to have a setting for how far away from the text it should be in addition to the one that specifies the thickness. Lacking that, I still think it's preferable to faking an underline with a border because the underline is cut for descenders. There's no great way to do that with a border or other treatment.

I'd also really love for comments to be conversations in GitHub just like if they were based on a line of code. Two things to hope for.

chrisdholt (approved):

packages/web-components/fast-components/src/styles/patterns/button.ts
khamudom: updated. thanks

packages/web-components/fast-components/src/anchor/anchor.styles.ts
```diff
@@ -19,11 +29,173 @@ export const anchorStyles: (
...
+            css`
+              :host([href]) .control {
+                  border-color: ${SystemColors.LinkText};
+                  background: ${SystemColors.ButtonFace};
+                  color: ${SystemColors.LinkText};
+                  fill: currentcolor;
+              }
+              :host([href]) .control:hover,
+              :host(.neutral[href]) .control:hover,
+              :host(.outline[href]) .control:hover {
```
bheston, line 52: The `.outline` styles should move down to the outline appearance behavior.

bheston (approved):

PR #5384, opened 2021-11-11
chore: visual refresh on accordion and accordion-item component

khamudom:
# Pull Request

## 📖 Description

Updated the default icons and focus state

Before
image
image

After
image
image

### General

### Component-specific


packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
-    }
-
-    :host([expanded]) .region {
+        ${display("flex")} :host {
+            box-sizing: border-box;
+            font-family: ${bodyFont};
+            flex-direction: column;
+            font-size: ${typeRampBaseFontSize};
+            line-height: ${typeRampBaseLineHeight};
+            color: ${neutralForegroundRest};
```
robarbms, line 43: All of the font styles are being declared in both accordion and accordion-item. This would be a good time to consolidate them so that future updates don't require updating all of these in two places.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
+        }
+        .region {
+            display: none;
+            padding: calc((6 + (${designUnit} * 2 * ${density})) * 1px);
+        }
+        .heading {
+            display: grid;
+            grid-template-columns: auto 1fr auto auto;
+            align-items: center;
+            z-index: 2;
```
robarbms, line 56: Removing `position: relative` might cause problems with the `button::before` element. It uses `div.heading` bounding box to position against.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
+        .button {
+            appearance: none;
+            border: none;
+            background: none;
+            grid-column: 2;
+            grid-row: 1;
+            outline: none;
+            padding: 0 calc((6 + (${designUnit} * 2 * ${density})) * 1px);
+            text-align: left;
+            height: calc(${heightNumber} * 1px);
```
robarbms, line 67: Should line-height be included here with the same value as the height?

I don't think height is needed at all. Button is just a superficial wrapper around the contents of the heading.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
robarbms: Actually, that's probably why the focus is wrapping both the header and the content-region now. This will cause clicking on the content-region to also toggle the expand.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
+        ${display("flex")} :host {
+            box-sizing: border-box;
+            font-family: ${bodyFont};
+            flex-direction: column;
+            font-size: ${typeRampBaseFontSize};
+            line-height: ${typeRampBaseLineHeight};
+            color: ${neutralForegroundRest};
+            border-top: calc(${strokeWidth} * 1px) solid ${neutralFillActive};
+            border-bottom: calc(${strokeWidth} * 1px) solid ${neutralFillActive};
+            position: relative;
```
robarbms, line 46: We shouldn't be adding `position: relative` to the host container. This is causing the `button::before` to position against accordion-item now that it's been removed from `.heading`. It will also cause problems with things like `fast-select` drop downs used in the content region. Developers can apply `position: relative` to `accordion-item` if needed.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
-`.withBehaviors(
+        }
+        .icon {
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            grid-column: 4;
+            z-index: 2;
+            pointer-events: none;
+            border-radius: calc(${controlCornerRadius} * 1px);
```
robarbms, line 98: This isn't needed anymore.
```suggestion
```

robarbms (commented):
There are conflicts with a fix to accordion-item styles that I just merged. You'll need to remove `z-index` from .heading, .start, .end and .icon and add `position: relative` to .start, .end and .icon.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
khamudom: The updated visual has the focus on both heading and content region.
 
image

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
khamudom: Will remove the font styles on accordion

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
khamudom: `position: relative` was moved from heading onto the host is because the focus has been visually updated to be on both heading and content-region

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
khamudom: removing the height

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
-
-    :host([expanded]) .region {
+        ${display("flex")} :host {
+            box-sizing: border-box;
+            font-family: ${bodyFont};
+            flex-direction: column;
+            font-size: ${typeRampBaseFontSize};
+            line-height: ${typeRampBaseLineHeight};
+            color: ${neutralForegroundRest};
+            border-top: calc(${strokeWidth} * 1px) solid ${neutralFillActive};
```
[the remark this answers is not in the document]
khamudom: updated to remove the double lines

packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -70,6 +71,8 @@ export const direction = create<Direction>("direction").withDefault(Direction.lt
 /** @public */
 export const disabledOpacity = create<number>("disabled-opacity").withDefault(0.3);
 /** @public */
+export const layerCornerRadius = create<number>("layer-corner-radius").withDefault(6);
```
bheston, line 74: Please follow https://github.com/microsoft/fluentui/blob/master/packages/web-components/src/design-tokens.ts as much as possible in terms of sequence, commenting, helper functions, etc. We may have new design tokens, which can fall in line alphabetically within "foreground", "fill", and "stroke" sections.

packages/web-components/fast-components/src/design-tokens.ts
```diff
@@ -256,24 +259,41 @@ export const neutralFillStrongFocusDelta = create<number>(
...
-).withDefault(16);
+).withDefault(90);
 /** @public */
 export const neutralStrokeFocusDelta = create<number>(
     "neutral-stroke-focus-delta"
-).withDefault(25);
+).withDefault(49);
+
+/** @public */
+export const neutralStrokeInputFilledRestDelta = create<number>(
```
bheston, line 282: I don't see these used here. Are they related to another component perhaps? This is fine, but I want to be cautious about naming here. We want these recipes to be as common or generic as reasonably possible. For example, is this intended to be paired with the "secondary" fill? If so, could we call this "secondary" stroke instead? Unfortunately the names don't always line up and aren't always used in pairs, but a very specific name becomes hard to use or abused easily.

packages/web-components/fast-components/src/accordion-item/accordion-item.styles.ts
```diff
@@ -35,127 +34,112 @@ export const accordionItemStyles: (
...
+            border-bottom: calc(${strokeWidth} * 1px) solid ${neutralFillActive};
+            position: relative;
+        }
+        .region {
+            display: none;
+            padding: calc((6 + (${designUnit} * 2 * ${density})) * 1px);
```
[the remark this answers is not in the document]
khamudom: updated

packages/web-components/fast-components/src/design-tokens.ts
khamudom: removed

packages/web-components/fast-components/src/design-tokens.ts
khamudom: rearranged and formatted

robarbms (approved):

bheston (approved):

PR #5392, opened 2021-11-16
add size attribute to listbox element

radium-v:
# Pull Request

## 📖 Description

* Adds the `size` attribute to the `<fast-listbox>` element.
* Converts `Listbox` to an abstract base class, and adds `ListboxElement` as a component class derived from that abstract class.

### 🎫 Issues

Overrides work handled in #4974 to unblock individual concerns.

## 👩‍💻 Reviewer Notes

This PR also fixes the `typeaheadExpired` property which was incorrectly spelled `typeAheadExpired`. A deprecated get/set fallback for the original property reflects the value of the new property.

## 📑 Test Plan

`Listbox` fixtures are updated to include specimens with various `size` attributes set. All automated Karma/Playwright tests should pass. `Listbox`, `Select`, and `Combobox` components should all behave as expected.

### General

### Component-specific

## ⏭ Next Steps


scomea:
When using arrow keys to move up/down the list the focused item frequently ends up against the edges of the viewport.  I think it would look better if there was some padding there, or maybe we try to center on the focused item?   

image

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
```diff
@@ -0,0 +1,114 @@
...
+            this.sizeStylesheet = this.$fastController.removeStyles(this.sizeStylesheet);
+        }
+
+        this.sizeStylesheet = css`
+            :host {
+                --size: ${"" + this.size};
+            }
+        `;
+
+        this.$fastController.addStyles(this.sizeStylesheet);
```
chrisdholt, line 112: This ends up tightly coupling styles and the class, which may be ok...but I wonder if this coupling is better done in the _implementation_ of fast-components and others...

sites/site-utilities/statics/assets/components/fast-listbox.schema.json
```diff
@@ -24,6 +24,12 @@
       "mapsToAttribute": "disabled",
       "type": "boolean"
     },
+    "size": {
+      "title": "Size",
+      "description": "The number of visible options",
```
scomea, line 29: Would saying "the "maximum number of visible options" be more accurate?

packages/web-components/fast-foundation/docs/api-report.md
```diff
@@ -1345,42 +1344,61 @@ export class Listbox extends FoundationElement {
...
 
 // @internal (undocumented)
 export interface Listbox extends DelegatesARIAListbox {
 }
 
+// @public
+export class ListboxElement extends Listbox {
+    // @internal @override
+    mousedownHandler(e: MouseEvent): boolean | void;
+    size: number;
```
scomea, line 1392: `size` and `sizeAttribute` need internal or public tags.

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
```diff
@@ -0,0 +1,114 @@
...
+     * The value of the `size` attribute is reflected by the `size` property.
+     *
+     * @public
+     */
+    @attr({
+        attribute: "size",
+        converter: nullableNumberConverter,
+        mode: "fromView",
+    })
+    public sizeAttribute: number;
```
scomea, line 32: `sizeAttribute` feels like an odd name for a prop, not that I necessarily have a better suggestion...

packages/web-components/fast-foundation/src/listbox/listbox.spec.md
```diff
@@ -33,6 +33,7 @@ This component is used as a building block for other components in this library
 - `role` - The role of the element, defaults to "listbox".
 - `selectedOptions` - A collection of the selected options in the `listbox`.
 - `selectedIndex` - The index of the first selected option, or `-1` if nothing is selected. Setting the `selectedIndex` property will update the `selected` state of the option at the new index. Out of range values will reset the `selectedIndex` to `-1`.
+- `size` - The number of options that should be visible in the `listbox` scroll area.
```
scomea, line 36: "maximum number"?

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -333,7 +260,7 @@ export class Listbox extends FoundationElement {
 
         switch (key) {
             // Select the first available option
-            case "Home": {
+            case keyHome: {
                 if (!e.shiftKey) {
                     e.preventDefault();
```
scomea, line 265: I think the calls to preventDefault here are unnecessary as that is already handled by the boolean return value of the function though the template binding.

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @remarks
+     * Setting `size` property does not change the value via the `size` attribute.
+     *
+     * Any value above zero will force the component to display as a listbox.
+     *
+     * @public
+     */
+    @observable
+    public size: number = 0;
```
chrisdholt, line 45: I think I saw a question recently relating to a similar relationship of something which was an attribute and the property didn't map directly. If I want to programmatically set size, setting this will NOT affect the attribute - can you elaborate a bit on your thoughts here?

packages/web-components/fast-foundation/docs/api-report.md
radium-v: They both do - the API extractor only marks non-public properties when they're `@internal`.

packages/web-components/fast-foundation/src/listbox/listbox.ts
radium-v: Since this base method is overridden in `Select` and `Combobox`, the returned value for this function only affects the template binding behavior when called for `ListboxElement`. Relying on the return value to handle `preventDefault` gets really confusing when working with supermethod overrides.

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
radium-v: I wasn't sure if this was convention but it appears that most browsers do bind the `size` attribute to the `size` property. I'll push an update to simplify this.

radium-v: Update: it was due to the interactions between `multiple` and `size`. When `multiple` is present, the `size` property defaults to `4`. This default is set internally, and isn't reflected in the `size` attribute with a native `<select>` element.

chrisdholt: Gotcha - do you think we'll hit issues where a user expects `myListbox.setAttribute("size", "4")` and `myListbox.size = 4` to both update the same thing? Is there a way we can address that or ensure the API works as expected?

packages/web-components/fast-foundation/src/listbox/listbox.ts
```diff
@@ -393,31 +320,188 @@ export class Listbox extends FoundationElement {
...
+            }
+
+            this.selectedIndex = 0;
+        }
+    }
+
+    /**
+     * Sets an option as selected and gives it focus.
+     *
+     * @param index - option index to select
```
alphashr, line 435: ```setSelectedOptions``` takes no param. It uses ```this.selectedIndex```.

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
radium-v: This is now moved to the `components` side.

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
radium-v: This is now updated to remove `sizeAttribute`, in favor of only using the `@attr size` property.

chrisdholt (approved):
I got lost a bit in the reorganization, but it looks like things are just moving around a bit. Thanks for making the change to move the visual behaviors to `fast-components`!

packages/web-components/fast-foundation/src/listbox/listbox.element.ts
radium-v: This PR doesn't deal with the `multiple` case, so I updated it to handle the `size` property and attribute as one `@attr`-decorated property. Changing the property now updates the attribute and vice versa.

packages/web-components/fast-foundation/src/listbox/listbox.ts
scomea: Ok.  I wish we did not have different signatures for the template vs regular event listeners.

scomea (approved):

Issue #5327, opened 2021-10-27
Cannot change fast-tooltip anchor dynamically


MarcSkovMadsen:
Fast-Components: 2.14.2

I'm trying to wrap the `fast` component for use in #python #datascience with [HoloViz Panel](https://panel.holoviz.org/). I would like to add tooltips. During this I noticed I cannot change the `anchor` of the `fast-tooltip` dynamically.

I would expect the tooltip to change to the `button2` when I click `button`. But it does not.

https://user-images.githubusercontent.com/42288570/139015927-023a3a22-f0bb-40b9-8412-012af8484993.mp4

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>title</title>
    <script type="module" src="https://unpkg.com/@microsoft/fast-components@2.14.2"></script>
  </head>
  <body>
    <div>
        <fast-button id="button" onclick='document.getElementById("tooltip").anchor="button2"'>Hover me</fast-button>
        <fast-tooltip id="tooltip" anchor="button">This is a tooltip</fast-tooltip>
        <fast-button id="button2">Hover me 2</fast-button>
    </div>
    <script>

    </script>
  </body>
</html>
```

## Additional Context

See 

- https://github.com/microsoft/fast/pull/5299
- https://github.com/holoviz/panel/issues/2859

PR #5394, opened 2021-11-17
fix: changing tooltip anchor attribute should take effect

scomea:
## 📖 Description
dynamically changing tooltip anchorElement worked, but changing the anchor attribute after the component was created did not.  This change enables anchor attribute changes to take effect.

### 🎫 Issues
fixes https://github.com/microsoft/fast/issues/5327

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
+++ b/packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
@@ -14,7 +14,8 @@ function onShowClick(): void {
     }
 }
 
-function onAnchorMouseEnter(e: MouseEvent): void {
+//check changing anchor by changing anchorElement
+function onAnchorMouseEnterProp(e: MouseEvent): void {
     if (!e.target) {
         return;
     }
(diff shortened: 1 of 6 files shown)


packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
```diff
@@ -25,12 +26,30 @@ function onAnchorMouseEnter(e: MouseEvent): void {
     tooltipInstance.anchorElement = e.target as HTMLElement;
 }
 
+//check changing anchor by setting attribute
+function onAnchorMouseEnterAttribute(e: MouseEvent): void {
```
radium-v, line 30: This duplicative function can be combined with the original by looking at the id of `e.target`

packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
```diff
@@ -25,12 +26,30 @@ function onAnchorMouseEnter(e: MouseEvent): void {
...
+        "tooltip-anchor-switch"
+    ) as FoundationTooltip;
+    tooltipInstance.setAttribute("anchor", (e.target as HTMLElement).id);
+}
+
 addons.getChannel().addListener(STORY_RENDERED, (name: string) => {
     if (name.toLowerCase().startsWith("tooltip")) {
         document
-            .querySelectorAll("fast-button[id^=anchor-anchor-switch]")
+            .querySelectorAll("fast-button[id^=anchor-anchor-switch-prop]")
```
radium-v, line 44: Why use ids here instead of classes?

packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
scomea: It's a sample, not package code, so I prefer the more "obvious" example.

packages/web-components/fast-components/src/tooltip/tooltip.stories.ts
scomea: tooltips/anchors are associated by passing id's so that seemed the logical way to identify elements in the sample

SethDonohue (approved):

robarbms (approved):

radium-v (approved):

EisenbergEffect (approved):

Issue #5385, opened 2021-11-12
fix: disabled attribute didn't block @click handler in fast-flipper


akronbook:
## Issue description
When a fast-flipper sets the "disabled" attribute, the @click handler is still triggered. 

## 💻 Repro or Code Sample

https://github.com/akronbook/flipper-disable-bug

## 🤔 Expected Behavior
When the "disabled" attribute is set, the @click handler should NOT be called. 

## 😯 Current Behavior

In the sample code, the @click handler was triggered even though the flipper was disabled. If you keep clicking on the disabled flipper, the page count will keep increasing, which is surprising.

## 🔦 Context
I am trying to disable the flipper in certain conditions in my application. Currently the workaround is to use "when" directive to remove fast-flipper totally from the DOM tree: this is not an ideal solution for us.

## 🌍 Your Environment

* OS & Device: Mac OS
* Browser Google Chrome
* Version fast-components 2.16.2, fast-element 1.6.2

chrisdholt:
@eljefe223 can you investigate this and see about potentially addressing this holistically across all controls? Are there any scenarios where we would want to invoke click or keydown handlers on disable elements?

PR #5399, opened 2021-11-17
fix: disabled flipper button clicks

eljefe223:
# Pull Request

## 📖 Description
Adds `pointer-events: none` to prevent clicks. 

Other option would be to place a `click` handler on the template and `stopPropagation`. But css approach seems less intrusive.

```ts
public clickHandler(e: Event & MouseEvent) {
        if(this.disabled) {
            e.stopImmediatePropagation();
        }
    }
```

### 🎫 Issues
Closes: #5385 

### General

### Component-specific

### DIFF
--- a/change/@microsoft-fast-components-ce85fe24-cfca-4d69-a8e5-5a4ba3305d77.json
+++ b/change/@microsoft-fast-components-ce85fe24-cfca-4d69-a8e5-5a4ba3305d77.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: diasbled flipper button clicks",
+  "packageName": "@microsoft/fast-components",
+  "email": "jes@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/@microsoft-fast-components-ce85fe24-cfca-4d69-a8e5-5a4ba3305d77.json
+++ b/change/@microsoft-fast-components-ce85fe24-cfca-4d69-a8e5-5a4ba3305d77.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "fix: diasbled flipper button clicks",
+  "comment": "fix: disabled flipper button clicks",
   "packageName": "@microsoft/fast-components",
   "email": "jes@microsoft.com",
   "dependentChangeType": "patch"
--- a/packages/web-components/fast-components/src/flipper/flipper.styles.ts
+++ b/packages/web-components/fast-components/src/flipper/flipper.styles.ts
@@ -77,6 +77,7 @@ export const flipperStyles: (
         cursor: ${disabledCursor};
         fill: currentcolor;
         color: ${neutralForegroundRest};
+        pointer-events: none;
     }
 
     :host([disabled])::before,


scomea (approved):

EisenbergEffect:
I think we have this same problem on other components as well (e.g. button itself). Could we do a quick survey of the components and see if there are other places this fix is needed, and explore what our standard approach should be?

scomea:
I am wondering if this type of "functional" style might best be implemented by setting the attribute from the code side like when the disable prop is toggled, vs. in a style sheet.  Would avoid folks needing to update style sheets for this kind of fix.  Down side is that it would not be easy to override.

eljefe223:
> I am wondering if this type of "functional" style might best be implemented by setting the attribute from the code side like when the disable prop is toggled, vs. in a style sheet. Would avoid folks needing to update style sheets for this kind of fix. Down side is that it would not be easy to override.

@scomea Open to others feedback. We could also make it a style sheet behavior that gets added to the components that need it.

chrisdholt:
> I think we have this same problem on other components as well (e.g. button itself). Could we do a quick survey of the components and see if there are other places this fix is needed, and explore what our standard approach should be?

This is what Jeff is focused on - he’s just taking them individually.

EisenbergEffect (approved):

chrisdholt:
> > I am wondering if this type of "functional" style might best be implemented by setting the attribute from the code side like when the disable prop is toggled, vs. in a style sheet. Would avoid folks needing to update style sheets for this kind of fix. Down side is that it would not be easy to override.
> 
> @scomea Open to others feedback. We could also make it a style sheet behavior that gets added to the components that need it.

If we wanted to handle this behaviorally without styles we can just check before emitting or firing events. I'm open on how we approach this, but pointer-events will fix the current bug. We can definitely consider a larger change as a feature request though.

scomea:
> If we wanted to handle this behaviorally without styles we can just check before emitting or firing events.

I don't think so, click is native so we don't actually emit it.  We could try and stopPropagation on it.

chrisdholt:
> > If we wanted to handle this behaviorally without styles we can just check before emitting or firing events.
> 
> I don't think so, click is native so we don't actually emit it. We could try and stopPropagation on it.

For flipper I don't think the click is native...it doesn't exist, so that is being handled elsewhere. It's the case for key events but it's not actually a native element for this control: https://github.com/microsoft/fast/blob/master/packages/web-components/fast-foundation/src/flipper/flipper.template.ts

bheston (approved):

change/@microsoft-fast-components-ce85fe24-cfca-4d69-a8e5-5a4ba3305d77.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: diasbled flipper button clicks",
```
janechu, line 3:
```suggestion
  "comment": "fix: disabled flipper button clicks",
```

janechu (approved):

PR #5402, opened 2021-11-18
chore: visual refresh on radio component

khamudom:
# Pull Request

## 📖 Description

Updated the size to 18px.

before unchecked
image

after unchecked
image

before checked
image

after checked
image

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
@@ -23,13 +23,16 @@ export const radioGroupStyles: (
         flex-direction: column;
         color: ${neutralForegroundRest};
     }
+
     .positioning-region {
         display: flex;
         flex-wrap: wrap;
     }
+
     :host([orientation="vertical"]) .positioning-region {
         flex-direction: column;
     }
+
     :host([orientation="horizontal"]) .positioning-region {
         flex-direction: row;
     }
--- a/packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
+++ b/packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
@@ -4,7 +4,7 @@ import {
     ElementDefinitionContext,
     FoundationElementDefinition,
 } from "@microsoft/fast-foundation";
-import { designUnit } from "../design-tokens";
+import { designUnit, neutralForegroundRest } from "../design-tokens";
 
 /**
  * Styles for Radio Group
@@ -21,6 +21,7 @@ export const radioGroupStyles: (
         align-items: flex-start;
         margin: calc(${designUnit} * 1px) 0;
         flex-direction: column;
+        color: ${neutralForegroundRest};
     }
     .positioning-region {
         display: flex;
--- a/packages/web-components/fast-components/src/radio/radio.styles.ts
+++ b/packages/web-components/fast-components/src/radio/radio.styles.ts
@@ -41,101 +41,117 @@ export const radioStyles: (
     definition: RadioOptions
 ) => ElementStyles = (context: ElementDefinitionContext, definition: RadioOptions) =>
     css`
-    ${display("inline-flex")} :host {
-        --input-size: calc((${heightNumber} / 2) + (${strokeWidth} * 2));
-        align-items: center;
-        outline: none;
-        margin: calc(${designUnit} * 1px) 0;
-        /*
-        * Chromium likes to select label text or the default slot when
-        * the radio button is clicked. Maybe there is a better solution here?
-        */
-        user-select: none;
-        position: relative;
-        flex-direction: row;
-        transition: all 0.2s ease-in-out;
-    }
-    .control {
-        position: relative;
-        width: calc(var(--input-size) * 1px);
-        height: calc(var(--input-size) * 1px);
-        box-sizing: border-box;
-        border-radius: 50%;
-        border: calc(${strokeWidth} * 1px) solid ${strokeControlStrongRest};
-        background: ${neutralFillInputRest};
-        outline: none;
-        cursor: pointer;
-    }
-    :host(:enabled:hover) .control {
-        border-color: ${strokeControlStrongHover};
-    }
-    :host(:enabled:active) .control {
-        background: ${neutralFillInputActive};
-        border-color: ${strokeControlStrongActive};
-    }
-    :host(:${focusVisible}) .control {
-        border-color: ${strokeControlStrongFocus};
-        outline: calc(${focusStrokeWidth} * 1px) solid ${focusStrokeOuter};
-        outline-offset: calc(${strokeWidth} * 1px);
-    }
-    :host(.checked) .control {
-        background: ${accentFillRest};
-        border-color: ${accentFillRest};
-    }
-    :host(.checked:enabled:hover) .control {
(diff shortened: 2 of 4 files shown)


packages/web-components/fast-components/src/radio-group/radio-group.styles.ts
```diff
@@ -21,6 +21,7 @@ export const radioGroupStyles: (
         align-items: flex-start;
         margin: calc(${designUnit} * 1px) 0;
```
chrisdholt, line 22: Created https://github.com/microsoft/fast/issues/5407 to address this.

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -42,174 +38,155 @@ export const radioStyles: (
...
+    :host(:enabled:hover) .control {
+        border-color: ${strokeControlStrongHover};
+    }
+    :host(:enabled:active) .control {
+        background: ${neutralFillInputActive};
+        border-color: ${strokeControlStrongActive};
+    }
+    :host(:${focusVisible}) .control {
+        border-color: ${strokeControlStrongFocus};
+        outline: calc(${focusStrokeWidth} * 1px) solid ${strokeControlStrongFocus};
```
bheston, line 76: This should be the focus stroke color. Will likely both look the same, but all focus treatments should use the same focus-specific recipes.

bheston (approved):

packages/web-components/fast-components/src/radio/radio.styles.ts
khamudom: updated

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -42,174 +42,155 @@ export const radioStyles: (
...
         outline: none;
         margin: calc(${designUnit} * 1px) 0;
-        /* Chromium likes to select label text or the default slot when
-            the radio is clicked. Maybe there is a better solution here? */
+        ${
+            /*
+             * Chromium likes to select label text or the default slot when
+             * the radio button is clicked. Maybe there is a better solution here?
+             */ ""
+        }
```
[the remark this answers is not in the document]
khamudom: updated

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -41,175 +41,178 @@ export const radioStyles: (
...
+            cursor: ${disabledCursor};
+        }
+
+        :host(.checked) slot[name="checked-indicator"] {
+            opacity: 1;
+        }
+
+        :host(.disabled) {
+            opacity: ${disabledOpacity};
+        }
```
radium-v, line 154:
```suggestion
        :host([disabled]) .label,
        :host([readonly]) .label,
        :host([readonly]) .control,
        :host([disabled]) .control {
            cursor: ${disabledCursor};
        }

        :host(.checked) slot[name="checked-indicator"] {
            opacity: 1;
        }

        :host([disabled]) {
            opacity: ${disabledOpacity};
        }
```

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -41,175 +41,178 @@ export const radioStyles: (
...
+                :host(.checked.disabled:hover) .control,
+                .control:active {
+                    background: ${SystemColors.Field};
+                    border-color: ${SystemColors.GrayText};
+                }
+
+                :host(.disabled) slot[name="checked-indicator"],
+                :host(.checked.disabled) .control:hover slot[name="checked-indicator"] {
+                    fill: ${SystemColors.GrayText};
+                }
```
radium-v, line 215:
```suggestion
                :host([disabled]) {
                    opacity: 1;
                }

                :host([disabled]) .label {
                    color: ${SystemColors.GrayText};
                }

                :host([disabled]) .control,
                :host(.checked[disabled]:hover) .control,
                .control:active {
                    background: ${SystemColors.Field};
                    border-color: ${SystemColors.GrayText};
                }

                :host([disabled]) slot[name="checked-indicator"],
                :host(.checked[disabled]) .control:hover slot[name="checked-indicator"] {
                    fill: ${SystemColors.GrayText};
                }
```

packages/web-components/fast-components/src/radio/radio.styles.ts
```diff
@@ -41,175 +41,178 @@ export const radioStyles: (
...
+            cursor: pointer;
+            font-size: ${typeRampBaseFontSize};
+            line-height: ${typeRampBaseLineHeight};
+        }
+
+        slot[name="checked-indicator"] {
+            flex-shrink: 0;
+        }
+
+        slot[name="checked-indicator"] {
```
radium-v, line 126: These selectors can be combined.

radium-v (changes requested):
I'm not sure which of `.checked`, `[aria-checked="true"]`, and `[checked]` are the preferred way to style the checked state.

packages/web-components/fast-components/src/radio/radio.styles.ts
khamudom: updated

packages/web-components/fast-components/src/radio/radio.styles.ts
khamudom: updated

packages/web-components/fast-components/src/radio/radio.styles.ts
khamudom: combined

EisenbergEffect:
> I'm not sure which of `.checked`, `[aria-checked="true"]`, and `[checked]` are the preferred way to style the checked state.

I think in general we wanted to move away from putting classes on the host or driving anything based on those, in order to not get into potential conflicts with vdom systems or other frameworks that want to manage all aspects of the host element.

chrisdholt:
> > I'm not sure which of `.checked`, `[aria-checked="true"]`, and `[checked]` are the preferred way to style the checked state.
> 
> I think in general we wanted to move away from putting classes on the host or driving anything based on those, in order to not get into potential conflicts with vdom systems or other frameworks that want to manage all aspects of the host element.

I'd personally suggest binding to the `aria-checked="true"` for now. It mirrors how we've approached other instances and it maintains a direct tie to what makes this "checked". Were the API to change at some point away from "checked", the selector would remain for a11y and wouldn't necessarily require changes.

bheston:
> > > I'm not sure which of `.checked`, `[aria-checked="true"]`, and `[checked]` are the preferred way to style the checked state.
> > 
> > 
> > I think in general we wanted to move away from putting classes on the host or driving anything based on those, in order to not get into potential conflicts with vdom systems or other frameworks that want to manage all aspects of the host element.
> 
> I'd personally suggest binding to the `aria-checked="true"` for now. It mirrors how we've approached other instances and it maintains a direct tie to what makes this "checked". Were the API to change at some point away from "checked", the selector would remain for a11y and wouldn't necessarily require changes.

Can we get this written up as a set of principles and priorities. Like use native attributes first, followed by aria, followed by part, then class, etc.? This question keeps coming up and rules are getting applied subjectively. And honestly half the time I'm not sure what's best.

khamudom:
updated to use `[aria-checked="true"]`

PR #5406, opened 2021-11-18
update visuals on listbox and listbox-option

khamudom:
# Pull Request

## 📖 Description

The `listbox-option` visual states are updated. 
The background is using `neutralFillStealth` token and the foreground is using `neutralForeground` token.

listbox-option before (rest, hover, active)
image
listbox-option after(rest, hover, active)
image

Removed the accent background color on the selected state updated to an accent border color.
before
image
after
image

Updated the background on the `listbox` class.

before
image
after
image

### General

### Component-specific


chrisdholt (commented):
@khamudom, it looks like there is a lot of space between the listbox and it's option elements - is that "by design"? It seems like a mistake to have something like 4px between the edge of the listbox and the option.

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -43,117 +38,92 @@ export const optionStyles: (
...
+            border-radius: calc(${controlCornerRadius} * 1px);
+            border: calc(${focusStrokeWidth} * 1px) solid transparent;
+            box-sizing: border-box;
+            color: ${neutralForegroundRest};
+            cursor: pointer;
+            fill: currentcolor;
+            font-size: ${typeRampBaseFontSize};
+            height: calc(${heightNumber} * 1px);
+            line-height: ${typeRampBaseLineHeight};
+            margin: 0 calc((${designUnit} * 3) * 1px);
```
chrisdholt, line 54: This is the space I think I'm looking at - this seems massive. I'll add, I'm not sure we should be adding margin to the element here but rather where it is slotted. Margin should be applied contextually rather than on primitives. Think of what happens if we do this on Button - every button ends up having left/right margin, which seems odd given that margin relates to space between things.

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -43,117 +38,92 @@ export const optionStyles: (
...
+            outline: none;
+            overflow: hidden;
+            align-items: center;
+            padding: 0 calc(${designUnit} * 2.25px);
```
chrisdholt, line 58: Do we really want to use fractional pixels here? If I change design unit to anything else, this is going to end up with fractional values that get parsed oddly by the browser.

chrisdholt (changes requested):

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
bheston: I think you're looking for `9px`, which is `12px` (design unit * 3) - `3px` (focus stroke width).

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
bheston: Agreed this is rather large. Can we go to `2` (or `8px`) instead.

packages/web-components/fast-components/src/listbox/index.ts
```diff
@@ -1,6 +1,25 @@
...
+ * @public
+ */
+export class Listbox extends FoundationListbox {
+    /**
+     * @internal
+     */
+    public connectedCallback(): void {
+        super.connectedCallback();
+
+        fillColor.setValueFor(this, neutralLayerFloating);
```
bheston, line 19: We don't need this in the listbox, only the select and combobox where that's a flyout.

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -43,117 +38,92 @@ export const optionStyles: (
...
+            white-space: nowrap;
+        }
+        :host(:not([disabled]):hover),
+        :host(:not([disabled]):not([aria-selected="true"]):hover) {
+            background: ${neutralFillStealthHover};
+            color: ${neutralForegroundHover};
+        }
+        :host(:not([disabled]):active),
+        :host(:not([disabled]):not([aria-selected='true']):active),
+        :host([aria-selected="true"]),
```
bheston, line 69: For `selected` state we need to have a 3:1 accessible treatment. Please see the updated design that adds the accent border around the selected item and a minor fill. The associated fill should also have a hover and active state, so a selected item should also respond to mouse interaction.

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
khamudom: updated

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
khamudom: removed margin and updated to 8px padding on the listbox

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
khamudom: updated

packages/web-components/fast-components/src/listbox/index.ts
khamudom: removed

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -43,113 +39,93 @@ export const optionStyles: FoundationElementTemplate<
...
+        }
+        :host(:${focusVisible}) {
+            border-color: ${focusStrokeOuter};
+        }
+        :host([disabled]) {
+            cursor: ${disabledCursor};
+            opacity: ${disabledOpacity};
+        }
+        .content {
+            grid-column-start: 2;
```
alphashr, line 85: The ```host``` of content is inline-flex, I am not sure how placing this item through ```grid-column-start``` can be helpful.

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -43,113 +39,93 @@ export const optionStyles: FoundationElementTemplate<
...
+            justify-self: start;
```
alphashr, line 86: ```justify-self``` will be ignored in a flexbox layout.

khamudom: removed

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
khamudom: removed

bheston (approved):
The spacing concern is addressed and the outline addresses the contrast on the selected item. Looks good.

packages/web-components/fast-components/src/listbox/listbox.styles.ts
```diff
@@ -52,21 +51,27 @@ export const listboxStyles: FoundationElementTemplate<ElementStyles> = (
...
-
-            :host(:${focusVisible}) ::slotted([aria-selected="true"][role="option"]) {
-                background: ${SystemColors.Highlight};
-                border-color: ${SystemColors.ButtonText};
-                box-shadow: 0 0 0 calc(${focusStrokeWidth} * 1px) inset ${SystemColors.HighlightText};
-                color: ${SystemColors.HighlightText};
-                fill: currentcolor;
-            }
-        `
+                :host(:${focusVisible}) ::slotted([aria-selected="true"][role="option"]) {
```
chrisdholt, line 54: massive nit - should these attributes be switched? I think specificity is the same, but ideally this conveys an options which is selected...

chrisdholt (approved):

PR #5421, opened 2021-11-29
fix: disabled tree items fire `selected-changed` event

eljefe223:
# Pull Request

## 📖 Description
Stop propagation when the `treeItem` is `disabled`

### 🎫 Issues
Closes: #5322 

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/tree-view/tree-view.ts
+++ b/packages/web-components/fast-foundation/src/tree-view/tree-view.ts
@@ -233,6 +233,9 @@ export class TreeView extends FoundationElement {
 
     private handleItemSelected = (e: CustomEvent): void => {
         const newSelection: TreeItem = e.target as TreeItem;
+        if (newSelection.disabled) {
+            e.stopPropagation();
+        }
         if (newSelection !== this.currentSelected) {
             this.handleSelected(newSelection);
         }
--- a/packages/web-components/fast-foundation/src/tree-view/tree-view.ts
+++ b/packages/web-components/fast-foundation/src/tree-view/tree-view.ts
@@ -233,9 +233,6 @@ export class TreeView extends FoundationElement {
 
     private handleItemSelected = (e: CustomEvent): void => {
         const newSelection: TreeItem = e.target as TreeItem;
-        if (newSelection.disabled) {
-            e.stopPropagation();
-        }
         if (newSelection !== this.currentSelected) {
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-foundation/src/tree-view/tree-view.ts
```diff
@@ -233,6 +233,9 @@ export class TreeView extends FoundationElement {
 
     private handleItemSelected = (e: CustomEvent): void => {
         const newSelection: TreeItem = e.target as TreeItem;
+        if (newSelection.disabled) {
+            e.stopPropagation();
+        }
```
chrisdholt, line 238: Can you share the reasoning for handling this programmatically vs. using `pointer-events: none` elsewhere? This seems reasonable, but it would be good to ensure we have a consistent approach to handling these so there is an intuitive developer experience for library authors and intuitive UX for web authors.

eljefe223: Sure, the reason `pointer-events: none` is a bad approach here is because of nested Tree Items. If a disabled tree is nested the `pointer-events: none` successfully prevents the vent from firing on that item but the event fires on the parent tree item since the parent tree item hit target is the size of itself and all its children.

Screen Shot 2021-11-29 at 11 54 06 AM

eljefe223: Another option would be to move the click event on treeItem from the root template to the `positioning-region` This would ensure that only clicks happen on the item it self and aren't passed from a tree items nested items

scomea: I'm thinking the tree items shouldn't select based on clicks in the nested child region.  If someone introduced styling that added margins between child items you wouldn't expect clicks between two items to select the parent, for example.

scomea:
I've got a conflicting pr that addresses the same bug and some others relating to selection in tree items.  https://github.com/microsoft/fast/pull/5350

scomea:
Wouldn't a better "quick fix" be to not emit from the tree item in the first place, rather than try to intercept in tree view?  
    `private handleSelected(e?: Event): void {
        if (!this.disabled){
        }
    }`

scomea (approved):

scomea (approved):

nicholasrice (approved):

PR #5422, opened 2021-11-29
fix: menu should update child items on connect

scomea:
## 📖 Description

A user encountered an issue where the menu was not reliably updating the properties on child menu items on subsequent connect events for the same menu.  This change specifically updates child menu items when the component connects as well as when the slotted children change.

### 🎫 Issues

user reported

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/menu/menu.spec.ts
+++ b/packages/web-components/fast-foundation/src/menu/menu.spec.ts
@@ -104,6 +104,8 @@ describe("Menu", () => {
 
         await connect();
 
+        await DOM.nextUpdate();
+
         expect(document.getElementById("id1")?.getAttribute("tabindex")).to.equal("0");
 
         await disconnect();
@@ -130,6 +132,7 @@ describe("Menu", () => {
         (menuItem1 as MenuItem).disabled = true;
 
         await connect();
+        await DOM.nextUpdate();
 
         expect(document.getElementById("id1")?.getAttribute("tabindex")).to.equal("0");
 
@@ -157,6 +160,8 @@ describe("Menu", () => {
         element.appendChild(menuItem3);
 
         await connect();
+
+        await DOM.nextUpdate();
 
         expect(
             element.querySelector("[role='menuitem']")?.getAttribute("tabindex")
--- a/packages/web-components/fast-foundation/src/menu/menu.spec.ts
+++ b/packages/web-components/fast-foundation/src/menu/menu.spec.ts
@@ -121,6 +121,8 @@ describe("Menu", () => {
 
         await connect();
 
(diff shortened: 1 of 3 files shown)


packages/web-components/fast-foundation/src/menu/menu.ts
```diff
@@ -51,7 +53,7 @@ export class Menu extends FoundationElement {
      */
     public connectedCallback(): void {
         super.connectedCallback();
-        this.menuItems = this.domChildren();
+        this.setItems();
```
chrisdholt, line 56: I think we're going to see a subsequent call here with `itemsChanged` as it's going to run when connected as well since that is a change - is there a way to minimize the amount of code that runs twice?

scomea: Updated.  Now `setItems` won't run on slot contents changed until after it has been run on connectedCallback first.

robarbms (approved):

packages/web-components/fast-foundation/src/menu/menu.spec.ts
```diff
@@ -130,6 +132,7 @@ describe("Menu", () => {
         (menuItem1 as MenuItem).disabled = true;
 
         await connect();
+        await DOM.nextUpdate();
```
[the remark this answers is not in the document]
scomea: yup

EisenbergEffect (approved):
I think this is fine as a fix. However, I think this is worth revisiting in the future from a collaborating component design perspective. My thought is that it might be better to have menu items register themselves with their parent on connect, rather than having the parent try to find all the children. That would a be a bigger change to this component so I don't want to do that here, but I think it's worth considering for a future update.

chrisdholt:
> I think this is fine as a fix. However, I think this is worth revisiting in the future from a collaborating component design perspective. My thought is that it might be better to have menu items register themselves with their parent on connect, rather than having the parent try to find all the children. That would a be a bigger change to this component so I don't want to do that here, but I think it's worth considering for a future update.

@eisenbergeffect do you have an API shape in mind for this? Would be great to open an issue to track this as an improvement.

EisenbergEffect:
I don't have the details out, but I think what we would do is define an interface that was implemented by components that can have children. So, use the same pattern across menu, accordion, whatever needs this. Then the child simply looks at its parent node and sees if it implements this interface. If it does, it calls it with a ref to itself. This would happen on connect.

Not sure how to name it (so just being silly here), but something like this from the child's perspective:

```ts
connectedCallback() {
  super.connectedCallback();

  if (isParentController(this.parentElement)) {
    this.parentElement.helpMe(this);
  }
}
```

PR #5426, opened 2021-11-30
chore: visual refresh on skeleton component

khamudom:
# Pull Request

## 📖 Description

before
image

after
image

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/skeleton/fixtures/base.html
+++ b/packages/web-components/fast-components/src/skeleton/fixtures/base.html
@@ -1,32 +1,23 @@
 <style>
-    .card-placeholder {
-        background-color: #fff;
+    fast-card {
+        --card-width: 500px;
+        --card-height: auto;
+    }
+
+    .card-padding {
         padding: 20px;
-        width: 500px;
-        border-radius: 4px;
     }
-
-    .card-pattern {
-        border-radius: 4px;
-        width: 500px;
-        height: 250px;
-    }
-
     @media (forced-colors: active) {
-        .card-placeholder {
-            background-color: ButtonText;
-        }
-
-        .card-pattern {
-            background-color: #000;
+        fast-card {
+            background-color: buttontext;
         }
     }
 </style>
 
 <h1>Skeleton</h1>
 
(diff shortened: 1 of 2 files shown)


packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
```diff
@@ -21,41 +25,36 @@ export const skeletonStyles: (
...
             --skeleton-animation-gradient-default: linear-gradient(
                 270deg,
                 var(--skeleton-fill, var(--skeleton-fill-default)) 0%,
-                #f3f2f1 51.13%,
+                ${neutralFillActive} 51.13%,
                 var(--skeleton-fill, var(--skeleton-fill-default)) 100%
             );
             --skeleton-animation-timing-default: ease-in-out;
         }
-
```
radium-v, line 37: Newlines between selector blocks help with readability.

robarbms (approved):
You need a change file, but otherwise looks good.

packages/web-components/fast-components/src/skeleton/skeleton.styles.ts
khamudom: updated

bheston:
I think this is probably fine, but the test fixture is off. Can you make the fixture use a Card or otherwise set the background color into dark mode (fillColor) so we can see that the skeleton colors are correct?

packages/web-components/fast-components/src/skeleton/fixtures/base.html
```diff
@@ -1,32 +1,23 @@
...
+            background-color: buttontext;
         }
     }
 </style>
 
 <h1>Skeleton</h1>
 
-<h2>Used as element blocks</h2>
-<div class="card-placeholder">
+<h4>Used as element blocks</h4>
```
litteredwitherrors, line 19: What did we opt for heading level 4 here and below?

khamudom: that should stay h2

litteredwitherrors (approved):

PR #5429, opened 2021-12-01
fix: focus on combobox input when clicking an associated label

radium-v:
# Pull Request

## 📖 Description

Clicking on a label for a `<fast-combobox>` should move focus to the text input control.

## 👩‍💻 Reviewer Notes

This bug was discovered while updating the fixtures for combobox.

## 📑 Test Plan

Clicking on label which is associated with a combobox should move focus to the text input in the combobox. Clicking away should close the combobox.

### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-foundation/src/combobox/combobox.ts
+++ b/packages/web-components/fast-foundation/src/combobox/combobox.ts
@@ -5,12 +5,12 @@ import {
     SyntheticViewTemplate,
 } from "@microsoft/fast-element";
 import { limit, uniqueId } from "@microsoft/fast-web-utilities";
+import type { FoundationElementDefinition } from "../foundation-element";
 import type { ListboxOption } from "../listbox-option/listbox-option";
 import { ARIAGlobalStatesAndProperties } from "../patterns/aria-global";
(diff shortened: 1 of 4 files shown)


chrisdholt (approved):

packages/web-components/fast-foundation/src/combobox/combobox.ts
```diff
@@ -268,7 +268,9 @@ export class Combobox extends FormAssociatedCombobox {
 
         this.open = !this.open;
 
-        if (!this.open) {
+        if (this.open) {
+            this.control.focus();
+        } else {
             this.updateValue(true);
```
robarbms, line 274: You could move this to line 267 inside of the `if (this.open)` block. Also, when checking if that would cause a problem with how `this.updateValue()` works, I noticed that you only update the value if the controller is connnected. Should the emit be moved inside that block as well? Otherwise it will emit a change event sometimes when it's not changing.

scomea (approved):

robarbms (approved):

robarbms (approved):

PR #5433, opened 2021-12-02
fix: clicking on disabled fast-button events should behave like native button (#4884)

### LINKED ISSUE #4884: fix: Click events on disabled fast-button bubble up the parent hierarchy
# 🐛 Bug Report

When a fast button is in disabled state click events on it are still propagated up the parent hierarchy.

This may be releated to #4698 and #4137, but since the observed behaviour does not affect the event listeners on the fast element, but the event propagation I am creating this separate issue.

The issue propably affects other input elements as well, but I only tested this on fast button for now.

## 💻 Repro or Code Sample

See [https://skaempfer.github.io/fast-button-click/](https://skaempfer.github.io/fast-button-click/) for a behaviour comparison between native button element and fast-button element 

## 🤔 Expected Behavior

Fast button element behaves like native button element: When in disabled state, fast-button elements do not propagate click events up their parent hierarchy. This should be in accordance with the HTML spec: "*A form control that is disabled must prevent any click events that are queued on the user interaction task source from being dispatched on the element.*" ([Source](https://www.w3.org/TR/html52/sec-forms.html#enabling-and-disabling-form-controls-the-disabled-attribute))

## 😯 Current Behavior

When you click on an disabled fast button any click listener on the element is not triggered, but the click event bubbles up the parent hierarchy.

## 💁 Possible Solution

The fast button element wraps a native button element. To behave like its native counterpart the fast button should make a descision on blocking or allowing event propagation based on the disabled state of the wrapped button element.

prudepixie:
# Pull Request

## 📖 Description
This change makes sure that the click events from fast-button in its disabled state behaves the same way as a native html button. 
Specifically, if text is not wrapped around addition html element, clicking on the button should not propagate events to the container div.
```
<div id="container">
   <fast-button disabled>Button</fast-button>
</div>
```

If in the following example, there is a span wrapped around text, then clicking on the span would propagate to the container as a span within a native button would behave.
```
<div id="container">
   <fast-button disabled><span>Button</span></fast-button>
</div>
```

### 🎫 Issues

fixes #4884 
### General

### Component-specific

### DIFF
--- a/packages/web-components/fast-components/src/styles/patterns/button.ts
+++ b/packages/web-components/fast-components/src/styles/patterns/button.ts
@@ -92,6 +92,15 @@ export const BaseButtonStyles = css`
         border: 0;
     }
 
+    .control:disabled span
+    {
+        pointer-events: none;
+    }
+
+    ::slotted(*) {
+        pointer-events: auto;
+    }
+
     .start,
     .end {
         display: flex;
@@ -127,7 +136,7 @@ export const BaseButtonStyles = css`
               color: ${SystemColors.ButtonText};
               fill: currentColor;
             }
-    
+
             :host(:hover) .control {
               forced-color-adjust: none;
               background-color: ${SystemColors.Highlight};
--- a/change/@microsoft-fast-components-0589cbfe-cc68-48de-a2f6-4c70135397c5.json
+++ b/change/@microsoft-fast-components-0589cbfe-cc68-48de-a2f6-4c70135397c5.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: clicking on disabled fast-button events should behave like native button (#4884)",
+  "packageName": "@microsoft/fast-components",
+  "email": "email not defined",
+  "dependentChangeType": "patch"
+}


nicholasrice:
Thanks @prudepixie! can you change the issue reference to say "fixes #4884"? The "fixes" tells github to close the issue when the PR is merged.

nicholasrice (approved):

EisenbergEffect:
Woohoo. 🎊 First official team member PR I think.

EisenbergEffect:
I've reviewed and it looks good to me. I'd like to get @chrisdholt or a member of his team to take a look since they were addressing a few similar issues in other components recently.

chrisdholt:
> I've reviewed and it looks good to me. I'd like to get @chrisdholt or a member of his team to take a look since they were addressing a few similar issues in other components recently.

@eljefe223 is on point to fix these across the components, so I’ve asked him to review this.

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -92,6 +92,15 @@ export const BaseButtonStyles = css`
         border: 0;
     }
 
+    .control:disabled span
```
eljefe223, line 95: Thanks for the PR. Curious thought, this only applies to `span`?

eljefe223 (approved):

packages/web-components/fast-components/src/styles/patterns/button.ts
prudepixie: The button template for the fast-button is defined as:
```
#shadow-root
<button>
  <span part='start></span>
  <span class='content'><slot></slot></span>
  <span part='end'></span>
</button>
```
so the styles are specific to the spans here

eljefe223: Ok, I think I misunderstood the problem you were trying to solve based on the description you provided.

prudepixie:
A potentially more accurate solution for this issue: https://github.com/microsoft/fast/pull/5436

PR #5436, opened 2021-12-04
fix: prevent events to propagate from disabled button (#4884)

### LINKED ISSUE #4884: fix: Click events on disabled fast-button bubble up the parent hierarchy

prudepixie:
# Pull Request

## 📖 Description
This change makes sure that the click events from fast-button in its disabled state behaves the same way as a native html button. 
Specifically, if text is not wrapped around addition html element, clicking on the button should not propagate events to the container div.
```
<div id="container">
   <fast-button disabled>Button</fast-button>
</div>
```

If in the following example, there is a span wrapped around text, then clicking on the span would propagate to the container as a span or any html element within a native button would behave.
```
<div id="container">
   <fast-button disabled><span>Button</span></fast-button>
</div>
```

### 🎫 Issues

fixes #4884 
### General

### Component-specific

## ⏭ Next Steps

There seems to be other components (fast-input?) that have similar issues where because of how the template is defined, the events behave differently as a native HTML element.


packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -157,6 +157,25 @@ export class Button extends FormAssociatedButton {
 
         this.proxy.setAttribute("type", this.type);
         this.handleUnsupportedDelegatesFocus();
+        document.addEventListener('click', this.handleClick, true)
```
nicholasrice, line 160: I don't think we need to watch the entire document to catch this - we could probably just listen on the internal `<button>` or `<span>` elements. That would provide ample opportunity to stop propagation without running this code for *every* click event that bubbles up to the document.

prudepixie: Makes sense, something interesting to note here is that it doesn't seem possible to addEventLister to disabled buttons  , including the fast-button.
Might have something to do with the specs defined here (4.10.18.5): https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-fe-disabled 

So I'm going to add it to the #shadowRoot > button > span instead.

nicholasrice: Yea that makes sense, thanks @prudepixie!

packages/web-components/fast-foundation/src/button/button.spec.ts
```diff
@@ -563,4 +564,50 @@ describe("Button", () => {
             await disconnect();
         });
     });
+
+    describe("of 'disabled'", () => {
```
nicholasrice, line 568: Thanks for adding tests!

nicholasrice (approved):

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -157,8 +157,34 @@ export class Button extends FormAssociatedButton {
 
         this.proxy.setAttribute("type", this.type);
         this.handleUnsupportedDelegatesFocus();
+
+        const spans = this.shadowRoot?.querySelectorAll("span");
```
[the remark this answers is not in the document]
prudepixie: Sounds good, I'll make that change

chrisdholt:
Does this relate to #5433 ?

robarbms (approved):

packages/web-components/fast-foundation/src/button/button.ts
```diff
@@ -157,8 +157,40 @@ export class Button extends FormAssociatedButton {
...
+            });
+        }
+    }
+
+    /**
+     * @internal
+     */
+    public disconnectedCallback(): void {
+        super.disconnectedCallback();
+        const elements = this.shadowRoot?.querySelectorAll("span");
```
nicholasrice, line 175: We spoke offline but I wanted to make the note; We shouldn't access `this.shadowRoot` here because it will be `null` in cases of closed shadow roots and then this won't exhibit the desired behavior.

prudepixie: I've made the update to add the click handler on the spans themselves @nicholasrice please take a look when you get a chance (particularly in the start-end.ts file), thanks!

packages/web-components/fast-foundation/src/patterns/start-end.ts
```diff
@@ -63,6 +63,7 @@ export const endSlotTemplate: (
         part="end"
         ${ref("endContainer")}
         class=${x => (definition.end ? "end" : void 0)}
+        @click="${(x, c) => x.handleClick(c.event as Event)}"
```
nicholasrice, line 66: I think we need to add the event handler imperativly via the `endContainer` ref. Adding the click handler to the template like this adds handle click behavior to *all* components that implement start / end and also assumes they have a method called 'handleClick', so this could easily throw at runtime.

prudepixie: Got it! I'll make the change

prudepixie: Ok I ended up adding the handler imperatively for all spans under the button control, let me know if it looks ok now @nicholasrice

packages/web-components/fast-foundation/src/button/button.ts
nicholasrice: Looks good, thanks!

PR #5440, opened 2021-12-06
chore: visual refresh on number-field, text-area and text-field components

khamudom:
# Pull Request

## 📖 Description

number-field, text-field, and text-area are all using the same common styles in patterns/input.ts

The screenshots are number-field. The visuals are the same on text-field and text-area.

Appearance outline
before
image
after
image

Appearance fill
before
image
after
image

### General

### Component-specific


janechu:
In the before the first example shows without the up/down increment arrows, however in the after image they are there. What's the reasoning behind this? Seems like a behavioral change.

khamudom:
> In the before the first example shows without the up/down increment arrows, however in the after image they are there. What's the reasoning behind this? Seems like a behavioral change.

I don't know the reason why Jason Reynolds made the change to show the arrow on rest. @bheston do you have any comments on this? If we should always show the arrows or revert back to only showing it on hover.

packages/web-components/fast-components/src/number-field/index.ts
```diff
@@ -58,10 +58,14 @@ export const fastNumberField = NumberField.compose<NumberFieldOptions>({
         delegatesFocus: true,
     },
     stepDownGlyph: `
-        <span class="step-down-glyph" part="step-down-glyph"></span>
+        <svg width="10" height="9" viewBox="0 0 10 9" xmlns="http://www.w3.org/2000/svg">
```
bheston, line 61: There are a lot of off-pixel numbers here. I've updated the icons in Figma. Please get the updates and optimize them again. They should be 12x12 containers.

bheston (approved):

bheston:
> > In the before the first example shows without the up/down increment arrows, however in the after image they are there. What's the reasoning behind this? Seems like a behavioral change.
> 
> I don't know the reason why Jason Reynolds made the change to show the arrow on rest. @bheston do you have any comments on this? If we should always show the arrows or revert back to only showing it on hover.

We didn't discuss this specifically, but I think it's fine and potentially beneficial to include. It indicates the behavior of this field similar to the down arrow in a select. I don't think this is a behavioral change as the functionality is still to increment or decrement the value. Displaying the spinner before hover or focus helps avoid accidental activation from trying to focus the field by clicking in that hidden reserved area.

janechu (approved):

packages/web-components/fast-components/src/number-field/index.ts
khamudom: updated

Issue #5444, opened 2021-12-06
feat: add CSP support to fast-element for ElementStyles


nicholasrice:
# 🙋 Feature Request
`@microsoft/fast-element` exposes several API's to programmatically add style elements to light and shadow DOMs when constructable stylesheets are not supported. Because these stylesheets are injected into the document via JavaScript, they are susceptible to Content Security Policy (CSP) restrictions. 

This could take several forms:

**JavaScript configuration**
fast-element should implement a mechanism to provide a nonce attribute and value to all style elements it creates. Something like the following:

```ts
// Configure nonce for fast-element.
FASTElement.styleNonce = "{nonce}";
```

```html
<style nonce="{nonce}"></style>
```
I think this mechanism is preferred because it provides authors more flexibility into how the nonce is retrieved on the client.

**Embed nonce in HTML**
fast-element could look for a nonce meta element, emitting the style element with the nonce value if it finds one.
```html
<meta property="csp-nonce" content="{nonce}" />
...
<style nonce="{nonce}"></style>
```
related discussion: https://github.com/microsoft/fast/issues/4510

awentzel:
we should add this feature for production sites that incorporates nonces to avoid using unsafe-eval (scriptSrc) and unsafe-inline (styleSrc) references.  We do not have this feature today, but, it would improve our security posture.  The solution should also ensure that unsafe is not used in the CSP.

mikecentola:
I was able to use CSP with [vscode-webview-ui-toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) by adding the `<meta property='csp-nonce' content='{cspNonce}'>` to my extension, but I think it should either use `<meta name='csp-nonce' ...` or check for both in https://github.com/microsoft/fast/blob/master/packages/utilities/fast-web-utilities/src/dom.ts

I can create a PR with this change as it is a pretty simple non-breaking change.

awentzel:
> I was able to use CSP with [vscode-webview-ui-toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) by adding the `<meta property='csp-nonce' content='{cspNonce}'>` to my extension, but I think it should either use `<meta name='csp-nonce' ...` or check for both in https://github.com/microsoft/fast/blob/master/packages/utilities/fast-web-utilities/src/dom.ts
> 
> I can create a PR with this change as it is a pretty simple non-breaking change.

[...]

Issue #4510, opened 2021-03-29
Docs: Provide guidance on Content Security Policy rules


Garbee:
Content Security Policy is an important part of web apps being more secure in deployment. Fast should provide a documentation page providing known limitations and guidance for how to avoid CSP issues in restricted environments. It does not need to be all inclusive to start. Just a few of the current known things with examples. The resource is expected to grow as more things are tested and discovered by the community and engineering team.

### Examples

A known limitation is the `:focus-visible` CSS support check needing a nonce to work if `unsafe-inline` is not allowed.

Guidance for components can be if `style-src 'self'` is enforced then the `style` attribute in templates can't be used. Developers will need to use JavaScript to directly update elements through their style property.

Garbee:
Initial rough draft is available for review/commenting. I have a feeling this is a bit out from being decent enough for inclusion. Solid start though.

Reference: https://hackmd.io/@Garbee/Hk8qLm-Su

awentzel:
@garbee this is a great idea.  Any chance, you want to contribute a docs update for this, probably beneath a Security section?    We're also using Docusaurus.

For reference, here's a configuration in use today:
```
// Manage CSP Policy for Creator Application only
if (process.env.WEBSITE_HOSTNAME.indexOf("create") > -1) {
    app.use(
        helmet.contentSecurityPolicy({
            directives: {
                defaultSrc: ["'self'"],
                fontSrc: ["'self' use.typekit.net static.fast.design c.s-microsoft.com"],
                frameAncestors: [`'self' ${process.env.FRAME_ANCESTOR_PARTNER}`],
                imgSrc: ["'self' data: *.fast.design via.placeholder.com"],
                scriptSrc: ["'self' 'unsafe-eval'"],
                styleSrc: ["'self' https: 'unsafe-inline'"],
            },
        })
    );
}```

If you have more insight or ideas on security let us know.

awentzel:
If we can get a few eyes on your changes as published at https://hackmd.io/@Garbee/Hk8qLm-Su if this documentation looks good I can push a PR with these Docusaurus changes.   @EisenbergEffect @nicholasrice are you able to take a look?

nicholasrice:
[...]

PR #5460, opened 2021-12-13
fix: change style Attribute to style tag(to add CSP support)

SilkyFowl:
# Pull Request

## 📖 Description

This PR is to make it possible to set nonce for fast-element.
This solved the problem I encountered, but I couldn't decide if this policy is OK, so I made a Draft.
I hope this helps with #5444.

### Background

This all started when I realized that it was hard to resolve a csp violation in my repository.

https://github.com/SilkyFowl/fable-vscode-extension-samples/pull/17

The [@vscode/webview-ui-toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) used in this repository depends on fast.
By the way, vscode extension [recommends the use of Content security policy when using WebView.](https://code.visualstudio.com/api/extension-guides/webview#content-security-policy)
A sample implementation can be found here.
https://github.com/microsoft/vscode-extension-samples/blob/e958fb7d45abda8177f413062318447ce59c780f/webview-sample/src/extension.ts#L196

Currently, if using Vscode's webview-ui-toolkit, it will violate Vscode's recommendations.
To resolve the inconsistency of webview-ui-toolkit, we need to resolving #5444.

### 🎫 Issues

#5444 <- #4510

This PR was created using #5444 as a hint.

### General

### Component-specific

### DIFF
--- a/packages/utilities/fast-web-utilities/src/dom.ts
+++ b/packages/utilities/fast-web-utilities/src/dom.ts
@@ -38,7 +38,7 @@ export function getKeyCode(event: KeyboardEvent | null): number | null {
  *
  * Based on https://github.com/cssinjs/jss/blob/master/packages/jss/src/DomRenderer.js
  */
-function getNonce(): string | null {
+export function getNonce(): string | null {
     const node = document.querySelector('meta[property="csp-nonce"]');
     if (node) {
         return node.getAttribute("content");
--- a/packages/web-components/fast-element/src/templating/view.ts
+++ b/packages/web-components/fast-element/src/templating/view.ts
@@ -1,3 +1,4 @@
+import { getNonce } from "@microsoft/fast-web-utilities";
 import type { Behavior } from "../observation/behavior";
 import type { ExecutionContext } from "../observation/observable";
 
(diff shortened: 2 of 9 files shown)


EisenbergEffect:
I do have some concerns over the performance of running the query selector on every view, particularly since we don't tend to put style tags in our templates. We prefer to handles styles with the `css` tagged template handler. So, I think there is probably a better way for us to handle this. @nicholasrice may have some thoughts on that.

SilkyFowl:
Thanks for the review.
I'm embarrassed to say that I overlooked the performance degradation caused by query selector.

> @nicholasrice may have some thoughts on that.

Should I wait for him to formulate his thoughts?

EisenbergEffect:
@SilkyFowl No need to be embarrassed. It's a pragmatic solution that looks like it would address the issue. I happen to be very, very familiar with this part of the code and all the implications, so it jumps out to me more than most.

I'd like to wait to hear some thoughts from @nicholasrice since he's done some investigation in this area and we were tracking some improvements to FE 2 for this. With holiday season coming up, it might be a little bit longer than normal for him to get around to it though.

SilkyFowl:
@EisenbergEffect Thank you very much.
If it takes a little time, does that mean I should Close this PR?
If improvements are to be made based on @nicholasrice's findings, then a new PR with that policy will be created, right?
In that case, I'll comment on #5444 if I notice anything.

EisenbergEffect:
@SilkyFowl Yes, I think that would be the best course for now. If you can share your specific scenarios in #5444 then we can all use that issue to discuss together the optimal solution. Once we've arrived at a place we all agree on, then if you would like to contribute through code, we'd love to have you work together with us on a new PR. But I think we should start with a discussion in the issue and also make sure that Nick has a chance to weigh in.

EisenbergEffect:
I'll close this PR in favor of the issue discussion for now.

PR #5467, opened 2021-12-15
fix: listbox-option, listbox, select, and combobox accessibility

### LINKED ISSUE #5243: fix: Screen readers do not announce options in Select listbox
# 🐛 Bug Report

When using keyboard navigation to interact with a select component, options in the list are not announced by the screen reader. Also described in #4514.

## 💻 Repro or Code Sample

1. Active the screen reader on your computer.
2. Focus on a `fast-select` component. The [fast-select example](https://explore.fast.design/components/fast-select) in the component explorer is a reliable repo. 
3. Hit enter to open the listbox
4. Use arrow keys to scan through the options

## 🤔 Expected Behavior

The screen reader should announce each value as the user focuses on it. Native select/options elements do this out of the box. Interestingly enough, the [fast-combobox](https://explore.fast.design/components/fast-combobox) works as expected.

## 😯 Current Behavior

No change is announced when focusing on option elements.

## 🌍 Your Environment

* OS & Device: macOS and Windows 11
* Browser: Microsoft Edge
* Version: 94.0.992.31

cc @hawkticehurst

radium-v:
# Pull Request

## 📖 Description

Improves the ARIA states and property handling for the `<fast-option>`, `<fast-listbox>`, `<fast-select>`, and `<fast-combobox>` components.

* Adds `aria-posinset` and `aria-setsize` attributes to `<fast-option>`
* Moves ARIA properties from the `<template>` to the internal `<input>` in `<fast-combobox>`
* Sets the `aria-controls` to match the internal `listbox` element in `<fast-select>` and `<fast-combobox>`

### 🎫 Issues

Fixes #5243 and #4824.

## 👩‍💻 Reviewer Notes

Screen reader results can vary drastically based on configuration and browser environment, so I compared the output to these examples:

* [Scrollable Listbox](https://w3c.github.io/aria-practices/examples/listbox/listbox-scrollable.html)
* [Select-Only Combobox](https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html)
* [Editable Combobox With Both List and Inline Autocomplete](https://w3c.github.io/aria-practices/examples/combobox/combobox-autocomplete-both.html)

## 📑 Test Plan

Tested with Edge 96 in Windows 11 with Narrator.

### General

### Component-specific

## ⏭ Next Steps

### DIFF
--- a/packages/web-components/fast-foundation/src/combobox/combobox.spec.ts
+++ b/packages/web-components/fast-foundation/src/combobox/combobox.spec.ts
@@ -5,41 +5,52 @@ import { ListboxOption, listboxOptionTemplate } from "../listbox-option";
 import { fixture } from "../test-utilities/fixture";
 import { Combobox, comboboxTemplate as template } from "./index";
 
-const FASTCombobox = Combobox.compose({
-    baseName: "combobox",
-    template
-})
-
-const FASTOption = ListboxOption.compose({
-    baseName: "option",
-    template: listboxOptionTemplate
-})
-
-async function setup() {
-    const { element, connect, disconnect, parent } = await fixture(
-        [FASTCombobox(), FASTOption()]
-    );
-
-    element.id = "combobox";
-
-    const option1 = document.createElement("fast-option") as ListboxOption;
-    option1.textContent = "one";
-
-    const option2 = document.createElement("fast-option") as ListboxOption;
-    option2.textContent = "two";
-
-    const option3 = document.createElement("fast-option") as ListboxOption;
-    option3.textContent = "three";
-
-    element.appendChild(option1);
-    element.appendChild(option2);
-    element.appendChild(option3);
-
-    return { element, connect, disconnect, document, option1, option2, option3, parent };
-}
-
-// TODO: Need to add tests for keyboard handling & focus management
 describe("Combobox", () => {
+    const FASTCombobox = Combobox.compose({
+        baseName: "combobox",
+        template
+    })
+
+    const FASTOption = ListboxOption.compose({
+        baseName: "option",
+        template: listboxOptionTemplate
+    })
+
+    async function setup() {
+        const { element, connect, disconnect, parent } = await fixture(
+            [FASTCombobox(), FASTOption()]
+        );
+
+        element.id = "combobox";
+
+        const option1 = document.createElement("fast-option") as ListboxOption;
+        option1.textContent = "one";
+
+        const option2 = document.createElement("fast-option") as ListboxOption;
+        option2.textContent = "two";
+
+        const option3 = document.createElement("fast-option") as ListboxOption;
+        option3.textContent = "three";
+
+        element.appendChild(option1);
+        element.appendChild(option2);
+        element.appendChild(option3);
+
+        return { element, connect, disconnect, document, option1, option2, option3, parent };
+    }
+
+    it("should include a control with a role of `combobox`", async () => {
+        const { element, connect, disconnect } = await setup();
+
+        await connect();
+
+        expect(element.control).to.exist;
+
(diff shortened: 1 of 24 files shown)


packages/web-components/fast-foundation/src/combobox/combobox.spec.ts
```diff
@@ -40,16 +40,6 @@ async function setup() {
...
-    it("should include a control with a role of `combobox`", async () => {
-        const { element, connect, disconnect } = await setup();
-
-        await connect();
-
-        assert.strictEqual(element.control.getAttribute("role"), "combobox");
-
-        await disconnect();
-    });
-
```
chrisdholt, line 52: This is still true, yes?

radium-v: While it's technically still true, the attribute is now hard-coded in the template instead of reflecting a property binding.

chrisdholt: I think I’m scenarios like these the test is still valid and likely beneficial - this really ensures that the combobox role exists. Whether it’s hard coded or not shouldn’t matter; more that the combobox role is a big part of what makes it a combobox :)

packages/web-components/fast-foundation/src/combobox/combobox.template.ts
```diff
@@ -1,48 +1,47 @@
...
-) => ViewTemplate<Combobox> = (
-    context: ElementDefinitionContext,
-    definition: ComboboxOptions
-) => html`
+export const comboboxTemplate: FoundationElementTemplate<
+    ViewTemplate<Combobox>,
+    ComboboxOptions
+> = (context, definition) => html`
     <template
+        aria-disabled="${x => x.ariaDisabled}"
```
chrisdholt, line 17: does having this at the root create issues with us delegating focus to the input below?

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
```diff
@@ -102,14 +107,14 @@ export class ListboxOption extends FoundationElement {
...
-                this.dirtySelected = true;
-            }
+        this.ariaSelected = this.selected ? "true" : "false";
 
-            if (this.proxy instanceof HTMLOptionElement) {
-                this.proxy.selected = this.selected;
-            }
+        if (!this.dirtySelected) {
+            this.dirtySelected = true;
+        }
```
chrisdholt, line 114: would you mind adding. comment on this change just for future reference?

chrisdholt (approved):

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
radium-v: The `selected` property doesn't rely on the component being connected, so removing the `isConnected` check here allows the `aria-selected` attribute to be set before connecting as well.

packages/web-components/fast-foundation/src/combobox/combobox.template.ts
radium-v: The `disabled` property works on a [small list of elements and form-associated custom elements](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls:-the-disabled-attribute:the-button-element). Some browsers don't support `ElementInternals`, so we can't guarantee that the `disabled` attribute on its own suffices in every scenario.

Regardless, the `disabled` property does work on the shadow `<input>`. And while the `tabindex` prevents the entire component from being focusable while disabled, Narrator does at least announce it when reading the page in scan mode.

scomea (approved):

daviddossett:
Thank you so much!

PR #5470, opened 2021-12-15
Updated to generalized color recipes, mostly from Fluent UI

bheston:
# Pull Request

## 📖 Description

As part of the visual refresh, taking the opportunity to move to the generalized recipes we put in place in the latest Fluent UI update. This supports the longer-term goal of externalizing as much of the design system utilities as possible,

## 👩‍💻 Reviewer Notes

Most recipes updates directly swap and produce the same values. Some recipes will change color values slightly, but this was expected as part of the visual refresh anyway.

### General


EisenbergEffect (approved):
Since this is a breaking change, are we planning to do this as part of the visual refresh major version release?

bheston:
> Since this is a breaking change, are we planning to do this as part of the visual refresh major version release?

@EisenbergEffect, yes, that was my intent. I have it going into the feature branch. Given the recent work and conversations about shared design system recipes though, I think it would be best to put this work into a new fast-design package. If someone could set the shell of that up in the repo I could move things around in this PR. To be clear, it would be highly generalized recipes we think commonly benefit the community through configuration. + @chrisdholt

packages/web-components/fast-components/src/color/recipes/contrast-and-delta-swatch-set.ts
```diff
@@ -0,0 +1,77 @@
...
+    lightHoverDelta: number,
+    lightActiveDelta: number,
+    lightFocusDelta: number,
+    lightDirection: -1 | 1 | undefined | null = undefined,
+    darkBaseContrast: number,
+    darkRestDelta: number,
+    darkHoverDelta: number,
+    darkActiveDelta: number,
+    darkFocusDelta: number,
+    darkDirection: -1 | 1 | undefined | null = undefined
```
chrisdholt, line 52: This is quite a lot of parameters for the function. What are your thoughts on making this a typed object? It represents the input for this recipe (same for others); seems like a good fit for it.

chrisdholt:
> > Since this is a breaking change, are we planning to do this as part of the visual refresh major version release?
> 
> @EisenbergEffect, yes, that was my intent. I have it going into the feature branch. Given the recent work and conversations about shared design system recipes though, I think it would be best to put this work into a new fast-design package. If someone could set the shell of that up in the repo I could move things around in this PR. To be clear, it would be highly generalized recipes we think commonly benefit the community through configuration. + @chrisdholt

I'm wondering if we should discuss an abstraction for these. Approving as the work needs to be done regardless, but I want to kick the conversation off as we recently have had A LOT of interest in consuming the recipes as well as creating custom ones...

chrisdholt (approved):

packages/web-components/fast-components/src/color/recipes/contrast-and-delta-swatch-set.ts
bheston: Agreed. I didn't like having to add those to Fluent and didn't intend to copy them over here. Removed.

PR #5473, opened 2021-12-16
fix(fast-components): add missing type info to registration functions

### LINKED ISSUE #5387: fix: Incomplete component attribute type information when using `fast-react-wrapper`
# 🐛 Bug Report

Hi, over on the [Webview UI Toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit) we're hoping to ship a set of React components using the `fast-react-wrapper`. 

As we've been working towards this we've run into an issue where wrapped toolkit components do not have any type annotations/information for any of their attributes resulting in TypeScript intellisense/build errors.

We've also had some people from our community try out the `fast-react-wrapper` all with the same results.

I'm still not completely sure if this is an issue with the `fast-react-wrapper` itself or how the toolkit is implemented (I suspect it might actually be the latter since the toolkit is closely modeled after the FAST Components repo and that doesn't have the same issue). 

Any help in determining what the issue is and how to fix it would be greatly appreciated! Thanks!

## 💻 Repro or Code Sample

Here's a simple example of wrapping a `<vscode-button>` and using it in a counter component.

```typescript
// Inside toolkit.ts file

import React from "react";
import { provideReactWrapper } from "@microsoft/fast-react-wrapper";
import { provideVSCodeDesignSystem, vsCodeButton } from "@vscode/webview-ui-toolkit";

const { wrap } = provideReactWrapper(React, provideVSCodeDesignSystem());

// Toolkit Components
export const VSCodeButton = wrap(vsCodeButton());
```

```typescript
// Inside Counter.tsx file

import { useState } from "react";
import { VSCodeButton } from "./toolkit";

export function Counter() {
  const [count, setCount] = useState(0);

  function incrementCount() {
    setCount(count + 1);
  }

  // ❗️❗️ ERROR HERE ❗️❗️
  //
  // The `appearance` attribute and any other vscode-button attributes (both custom and
  // default attributes) will have an intellisense type error and will not build
  //
  // Note: A screenshot of the intellisense/build error can be found below
  return (
    <VSCodeButton appearance="secondary" onClick={incrementCount}>
      The count is: {count}
    </VSCodeButton>
  ); 
}
```

## 🤔 Expected Behavior

Wrapped components should have complete/correct type annotations for all of their attributes.

## 😯 Current Behavior

Here's a simple example of a component attribute and the associated error message:

Screen Shot 2021-11-11 at 10 15 56 AM

Screen Shot 2021-11-11 at 10 16 17 AM

## 🔦 Context

Our hope is to ship a set of React components from the Webview UI Toolkit package.

## 🌍 Your Environment

* OS & Device: MacOS on MacBook Pro
* Browser: VS Code Webview (Chromium-based iirc?)
* Version: FAST Element (`v1.6.0`) & FAST Foundation (`v2.21.0`)

EisenbergEffect:
# Pull Request

## 📖 Description

This PR fixes an issue with our component registration functions where type information was being lost in certain cases. This caused downstream effects in our React wrappers, which could not be accurately typed, and were then unusable in React.

### 🎫 Issues

* Fixes #5387

## 👩‍💻 Reviewer Notes

The changes are applied to the `index.ts` files of all the FAST Components. You will see that I have explicitly declared the type params for all calls to `compose`. There are two reasons for this:

1. Some components have custom options, specified with the first type param. Due to a limitation in TypeScript, if there's a function with two type params, once the first param is specified, it can no longer infer the second type param. Because our component registries with custom options were not specifying the second param, the type information was not flowing through the registry. This resulted in an inability to produce type-accurate React wrappers. So, I've added both type params to address this bug.
2. For consistency and to help avoid future bugs, I decided to specify both type params across all components, regardless of whether they have custom options or not.

While doing the work, I noticed that a few components that had special option types weren't using those in their registry code. So, I added those.

Documentation has been updated to call out this scenario.

## 📑 Test Plan

Added some code to the React wrapper tests to ensure that the types flow correctly and React components with all properties are generated.

### General

## ⏭ Next Steps

There are several follow-ups needed:

* The same set of changes need to be applied to the FluentUI Web Components
* I'd like to follow back up at some point and see if I can come up with a nicer alternative to `compose` that does the same thing under the hood. We'd keep the original, but deprecate it, and update code and docs for the new API.
* I noticed a few new components that either haven't shown up in our docs and/or haven't shown up in the component explorer. I think we should fix that. The list is `avatar`, `calendar`, `picker`, and `search`.


EisenbergEffect:
@janechu Any idea what's going on here with the schema generation error that's showing up?

EisenbergEffect:
Hey folks. This really needs reviews because we can't make the React wrappers work in TypeScript with our components without this being merged. /cc @chrisdholt @nicholasrice

nicholasrice (approved):

chrisdholt:
@eisenbergeffect is this still a needed PR? Assuming so, but want to confirm with you.

EisenbergEffect:
> @EisenbergEffect is this still a needed PR? Assuming so, but want to confirm with you.

Yep. Without this, the React wrapper can't infer the types properly and the resultant React component is missing properties. We'll need to make a similar update to our Fluent components as well.

chrisdholt:
> > @EisenbergEffect is this still a needed PR? Assuming so, but want to confirm with you.
> 
> Yep. Without this, the React wrapper can't infer the types properly and the resultant React component is missing properties. We'll need to make a similar update to our Fluent components as well.

Sounds good - looks like there are some issues with the build regarding component schemas, which is odd. We can take a look at this in the next day or so to figure out how to get this unblocked.

chrisdholt:
Closing this as we've removed `fast-components`

Issue #5302, opened 2021-10-20
fix: flippersHiddenFromAT shouldn't map to aria-hidden


robarbms:
# 🐛 Bug Report

The horizontal-scroll attribute `flippersHiddenFromAT` maps to aria-hidden. This will hide the whole component. They should only hide the flippers themselves.

## 🤔 Expected Behavior

`flippersHiddenFromAT` should apply aria-hidden to the flippers. We should use another attribute other than `aria-hidden` to map to the flippers aria-hidden attribute.

## 😯 Current Behavior

`flippersHiddenFromAT` add aria-hidden to the horizontal-scroll component.
## 💁 Possible Solution

`flippersHiddenFromAT` should apply `aria-hidden` to the flippers in the template.

## 🌍 Your Environment

* OS & Device: [e.g. MacOS, iOS, Windows, Linux] on [iPhone 7, PC]
* Browser [e.g. Microsoft Edge, Google Chrome, Apple Safari, Mozilla FireFox]
* Version [e.g. 1.8.0]

PR #5479, opened 2021-12-17
fix: flippersHiddenFromAT shouldn't map to aria-hidden

robarbms:
# Pull Request

## 📖 Description

fillpersHiddenFromAT is mapped to aria-hidden. This will hide the entire component from assistive technology. Changing the attribute to `hide-flippers`.

### 🎫 Issues
https://github.com/microsoft/fast/issues/5302

### General

### Component-specific


packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -140,7 +140,7 @@ export class HorizontalScroll extends FoundationElement {
      * Attribute to hide flippers from assistive technology
      * @public
      */
-    @attr({ attribute: "aria-hidden", converter: booleanConverter })
+    @attr({ attribute: "hide-flippers", converter: booleanConverter })
```
chrisdholt, line 143: Only suggestion here would be to specify that this is for AT (assistive tech). This implies a bit that it also visually hides it, when ideally it wouldn’t prevent render and they’d still show (on hover only technically…)

robarbms: I went back and forth on this and thought the same thing. But I was torn with using a really verbose attribute `flippers-hidden-from-at` and this. I figured that the way to hide flippers would be to slot in an empty element or through CSS so maybe it would be alright to be less explicit here. But it makes sense.

packages/web-components/fast-foundation/src/horizontal-scroll/horizontal-scroll.ts
```diff
@@ -140,7 +140,7 @@ export class HorizontalScroll extends FoundationElement {
      * Attribute to hide flippers from assistive technology
      * @public
      */
-    @attr({ attribute: "aria-hidden", converter: booleanConverter })
+    @attr({ attribute: "flippers-hidden-from-at", converter: booleanConverter })
     public flippersHiddenFromAT: boolean = false;
```
nicholasrice, line 144: Won't this be a breaking change for anyone already using the `aria-hidden` attribute to set this property?

chrisdholt: I think this is fundamentally broken - so I’m not sure we should continue supporting this as it removes the entire control from the DOM. It’s breaking but I think that’s the expected fix as this is broken.

chrisdholt: Similarly, adding aria-hidden to the control would continue to work as expected in that case - I think?

chrisdholt (approved):

EisenbergEffect (approved):

chrisdholt:
Thanks @robarbms!

PR #5485, opened 2021-12-20
docs(badge): describe color

yinonov:
is color a public API? it doesn't appear in the [readme's API](https://github.com/microsoft/fast/blob/b816a09d0cf393567396e3f5a8e06312798dc05e/packages/web-components/fast-foundation/src/badge/badge.spec.md#api)

### DIFF
--- a/packages/web-components/fast-foundation/src/badge/badge.ts
+++ b/packages/web-components/fast-foundation/src/badge/badge.ts
@@ -17,7 +17,7 @@ export class Badge extends FoundationElement {
     public fill: string;
 
     /**
-     * Indicates the badge should have a filled style.
+     * Indicates the badge's color.
      *
      * @public
      * @remarks


packages/web-components/fast-foundation/src/badge/badge.ts
```diff
@@ -17,7 +17,7 @@ export class Badge extends FoundationElement {
     public fill: string;
 
     /**
-     * Indicates the badge should have a filled style.
+     * Indicates the badge's color.
      *
      * @public
```
yinonov, line 22: is this private?
```suggestion
     * @private
```

nicholasrice: I don't think so, I think it should remain public. Is there a reason you're changing it?

yinonov: The reason is it doesn't appear in the [readme's API](https://github.com/microsoft/fast/blob/b816a09d0cf393567396e3f5a8e06312798dc05e/packages/web-components/fast-foundation/src/badge/badge.spec.md#api)

EisenbergEffect: I think we should leave this public and then just back-port this to the spec.

EisenbergEffect (approved):
Thanks for the fix to the documentation!

chrisdholt:
Closing this as stale - feel free to reopen if you feel this is still relevant. Thanks!

PR #5490, opened 2021-12-22
tests: Add contract testing around interpolated HTML tags

ParamagicDev:
# Pull Request

## 📖 Description

When working on a component library, I found a couple use cases for interpolating HTML tags. For example:

Lets say I have `<my-button>` and I want to conditionally render an anchor tag if an `href` is passed in as an attribute:

`<my-button href="#"></my-button>` would then be `<a href="#"><slot></slot></a>` under the hood. If no href, the `<a>` would be a button. So I'd want a template like this:

```js
const template = html`
  ${when(x => x.href == null, html`
    <a href="#"><slot></slot></a>
  `)}
   ${when(x => x.href != null, html`
    <button href="#"><slot></slot></button>
  `)} 
`
```

Additionally, lets say I wanted to pass in an arbitrary tag to a wrapper component.

`<my-wrapper tag="button"></my-wrapper>`

This will not work currently:

```js
const template = html`<${x => x.tag}><slot></slot></${x => x.tag}>`
```

The `button` gets mangled to something like: `<fast-ay31ye{0}fast-ay31ye></fast-ay31ye{0}fast-ay31ye>`

To get around this, you have to call html again:

```js
const template = html`${x => html`<${x.tag}><slot></slot></${x.tag}>`}`
```

## 👩‍💻 Reviewer Notes

There are no changes to implementations. This is purely contract testing to ensure interpolated HTML tags continue working into the future.

## 📑 Test Plan

This is purely backwards compatibility testing.

### General

## ⏭ Next Steps

None that I'm aware of.


claviska:
@EisenbergEffect @chrisdholt we are planning to rely on this behavior pretty heavily, which is where this test comes from. No changes are required, we're just hoping to ensure the current behavior persists.

/cc @benfeely

chrisdholt:
> @EisenbergEffect @chrisdholt we are planning to rely on this behavior pretty heavily, which is where this test comes from. No changes are required, we're just hoping to ensure the current behavior persists.
> 
> /cc @benfeely

Tracking what you all are laying down 👍.

Many of us are laying low during the holidays, but we’ll try and get a solid review on this soon!

claviska:
> Many of us are laying low during the holidays

Says the guy who replies within minutes! 🙃

Thanks for all you do. Enjoy the holidays, everyone!

packages/web-components/fast-element/src/templating/binding.spec.ts
```diff
@@ -189,7 +189,24 @@ describe("The HTML binding directive", () => {
 
             expect(toHTML(parentNode)).to.equal(`This is a template. value`);
         });
-    });
+
+        it("Should allow HTML tags in template bindings", async () => {
```
nicholasrice, line 193: I want to clarify this test-case description but I'm having trouble finding a good one. The expression in this test using the `x.knownValue` isn't a "binding" per-se, just an interpolation. I think what this really tests are that:
a.) Tags can be interpolated during template creation (you added a test specifically for this in template.spec.ts)
b.) Bindings that return templates are evaluated and re-rendered when observable properties used in the returned template are changed.

I actually don't see a test case for *b* in binding.spec.ts (maybe I'm missing or it is covered elsewhere?), so it might be worth generalizing this test-case instead, but mostly I want to avoid any confusion that FAST supports *binding* HTML tags in templates, because it only really supports interpolation.

ParamagicDev: yea perhaps binding is the wrong word. Basically just something so that if I define a function on a FASTElement that returns an HTML tag, I want to be able to use that tag in my template. I'll change this to "Should allow interpolated HTML tags"

chrisdholt (approved):

chrisdholt:
@nicholasrice can you take a look at this again when you are free?

nicholasrice (approved):

EisenbergEffect (approved):
Thank you for your patience!

chrisdholt (approved):
@ParamagicDev this is all set to go with the exception that it needs a change file. If you can run `yarn change` we'll get this right in! Thanks for your patience!

Issue #5349, opened 2021-11-01
fix: template partials in component definitions passed as string literals


radium-v:
# 🐛 Bug Report

Partials that are provided in component definitions are set as template literals, though they should be handled as `html` templates.

## 💻 Repro or Code Sample

```ts
export const fastComponent = Component.compose({
    baseName: "component",
    template,
    styles,
    defaultContent: `
        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
            <path fill-rule="evenodd" clip-rule="evenodd" d="M3.8 3h12.4c.4 0 .8.3.8.8v12.4c0 .4-.3.8-.8.8H3.8a.8.8 0 01-.8-.8V3.8c0-.5.3-.8.8-.8zm12.4-1H3.8C2.8 2 2 2.8 2 3.8v12.4c0 1 .8 1.8 1.8 1.8h12.4c1 0 1.8-.8 1.8-1.8V3.8c0-1-.8-1.8-1.8-1.8zM14 9H6v2h8V9z"/>
        </svg>
    `,
// ...
```

## 💁 Possible Solution

```ts
import { html } from "@microsoft/fast-element";

export const fastComponent = Component.compose({
    baseName: "component",
    template,
    styles,
    defaultContent: html`
        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
            <path fill-rule="evenodd" clip-rule="evenodd" d="M3.8 3h12.4c.4 0 .8.3.8.8v12.4c0 .4-.3.8-.8.8H3.8a.8.8 0 01-.8-.8V3.8c0-.5.3-.8.8-.8zm12.4-1H3.8C2.8 2 2 2.8 2 3.8v12.4c0 1 .8 1.8 1.8 1.8h12.4c1 0 1.8-.8 1.8-1.8V3.8c0-1-.8-1.8-1.8-1.8zM14 9H6v2h8V9z"/>
        </svg>
    `,
// ...
```

## 🔦 Context

Components which have definition properties set as string literals:
* fastAccordionItem
* fastCheckbox
* fastCombobox
* fastFlipper
* fastMenuItem
* fastNumberField
* fastProgress
* fastProgressRing
* fastRadio
* fastSelect
* fastSlider
* fastSwitch
* fastTreeItem

EisenbergEffect:
[...]

PR #5496, opened 2022-01-04
Components use html templates

m4thieulavoie:
# Pull Request

## 📖 Description

This PR adds extra typings by using the `html` template function in `fast-component`'s components parts.

Fixes https://github.com/microsoft/fast/issues/5349

### General

### Component-specific


radium-v (approved):

chrisdholt (approved):
Would love to get @eisenbergeffect to confirm approval here as there is a slight performance cost to using templates.

eljefe223 (approved):

EisenbergEffect:
I'd like to propose an alternative. If the issue is primarily about typing, could we add a cast instead? There is a cost to making things templates, when they could be string. Perhaps switch these back to string but then cast the string as something like this:

```ts
export type TemplatePart = string | ViewTemplate;
```

Thoughts?

radium-v:
@EisenbergEffect the types should already be `SyntheticViewTemplate | string` or something similar, which is what allows the properties to be strings or templates here (see this comment https://github.com/microsoft/fast/issues/5349#issuecomment-1004457459).

EisenbergEffect:
So, is this more about wanting syntax highlighting then?

EisenbergEffect (changes requested):
As much as I'd like to get syntax highlighting here, I'm not sure we should take this approach. The reason is that this has runtime consequences. Where we had strings before, we will now have to compile and create a FAST template instance. And every instance of any of these components will then have to dynamically instantiate that template and compose the views be default. The strings are much cheaper as they just get streamed into the static HTML of the parent template.

I thought there was a way with some of the VS Code plugins to indicate that a string was HTML without using a template tag. Could we leverage that in this case? If not, could we look into adding this as a feature of our own plugin? I think that would be a better approach because it would get us the design-time experience we want without costing us at runtime.

radium-v:
> I thought there was a way with some of the VS Code plugins to indicate that a string was HTML without using a template tag. Could we leverage that in this case?

This would be difficult because there's no way to determine if an untagged template string is any sort of specific language. It would also need to support expressions within that subset.

One option would be to use build-time imports for these partials, where the template content is placed in a raw file with its appropriate extension. This would only work if the content of the string is pure markup - for example, SVG markup in a regular SVG file, injected in-place with a rollup plugin. This would only work for templates with no additional transformations, since template expressions (`${}`) would exhibit the same issue since most languages don't see expressions as a valid construct, and any common directives would exhibit the same non-deterministic behavior but in reverse.

Another would be to modify our syntax highlighting extension to allow something like this:

```ts
    indicator: /* html */ `
        <div part="indicator" class="indicator"></div>
    `,
```

This may work, but it could end up sending the wrong signals when we have ``html`<div></div>` `` and ``/* html */`<div></div>` `` throughout the project.

EisenbergEffect:
Per an offline convo between myself and @radium-v we're going to look into enabling the comment syntax and see if we can leverage that here.

EisenbergEffect (approved):
Thanks for this follow-up!

PR #5500, opened 2022-01-05
add checked property to listbox-option

radium-v:
# Pull Request

## 📖 Description

Adds the `checked` property to the `<fast-option>` component.

### 🎫 Issues

Part of the work needed for #4190.

## 👩‍💻 Reviewer Notes

The `checked` property is only used for multiple selection mode, and is not presented as an author-configurable attribute.

Using `aria-checked` and `aria-selected` comes with _very specific_ requirements, as outlined in the [ARIA best practices document](https://www.w3.org/TR/wai-aria-practices-1.2/#:~:text=Conditions%20that%20would%20permit%20a%20listbox,separate%20method%20for%20controlling%20each%20state.).

> Some design systems use `aria-selected` for single-select widgets and `aria-checked` for multi-select widgets. In the absence of factors that would make an alternative convention more appropriate, this is a recommended convention.

## 📑 Test Plan

* Styles for the checked state should be present in the storybook story.
* Tests that synchronize the `checked` and `aria-checked` properties should pass.

### General

### Component-specific

## ⏭ Next Steps


packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
```diff
@@ -56,10 +64,10 @@ export const optionStyles: FoundationElementTemplate<
         font-size: ${typeRampBaseFontSize};
         height: calc(${heightNumber} * 1px);
         line-height: ${typeRampBaseLineHeight};
-        margin: 0 calc(${designUnit} * 1px);
+        margin: 0 calc((${designUnit} - ${focusStrokeWidth}) * 1px);
         outline: none;
         overflow: hidden;
-        padding: 0 calc(${designUnit} * 2.25px);
+        padding: 0 1ch;
```
eljefe223, line 70: Is this a pattern we will be following? This would be equal to the width of 1 zero from the parent font?

packages/web-components/fast-foundation/src/listbox-option/listbox-option.ts
```diff
@@ -44,6 +44,33 @@ export class ListboxOption extends FoundationElement {
...
 
+    /**
+     * The checked state is used when the parent listbox is in multiple selection mode.
+     * To avoid accessibility conflicts, the checked state should not be present in
+     * single selection mode.
+     *
+     * @public
+     */
+    @observable
+    public checked?: boolean;
```
scomea, line 55: Should this be an attribute as well?

radium-v: It's not intended to be configurable as an attribute since the rules regarding focus wouldn't allow it:

> When a multi-select listbox receives focus:
> * If none of the options are selected before the listbox receives focus, focus is set on the first option and there is no automatic change in the selection state.
> * If one or more options are selected before the listbox receives focus, focus is set on the first option in the list that is selected.

In our case, "focus" will be handled by this checked state (specifically the `aria-checked` attribute).

packages/web-components/fast-components/src/listbox-option/listbox-option.styles.ts
radium-v: I know this has come up before and I'm the one responsible for it 😅

Is there a better design token that would work here? The original value resolves to `9px` with our default `designUnit` value of 4.

eljefe223: @bheston

chrisdholt: We should look at opportunities to use ch where it makes sense though.

chrisdholt (approved):

scomea (approved):

Issue #5510, opened 2022-01-11
rfc: add new adaptive density system


bheston:
# 💬 RFC

Here's a look at some of the history, issues, and updated model for adaptive density.

Note that the examples use current Fluent UI dimensions, but the system works with any values.

# Overview

### Balance between information quantity, visual comfort, and usability.

Density refers to the relationship between visual elements and empty space - How content like text, icons, and images is laid out relative to each other, or how container areas are defined by color, line, or other demarcation.

### What density is not.

It is important to differentiate density from scaling or zooming. Density is the primary adjustment that can be made to fit more or less content in a given area while maintaining its size and readability. Scaling makes content larger or smaller. While this affects how much content is visible, it doesn’t affect the relationship between the content and the empty space. See Density vs. scaling.

### Common motivations.

A few considerations for adjusting interface density are:
- Cognitive preference - people who prefer to see more or less information
- Input mode or device - using a mouse vs. touch
- Feature complexity - providing access to many features in a small space
- Content type - technical control panel vs. marketing site

*A simple example of density configurations affecting vertical spacing:*

![More dense, normal, and less dense button examples](https://user-images.githubusercontent.com/47367562/148861014-5f6bb73a-6a78-4a64-9892-dfc70751cd73.png)

*Due to the nature of horizontal character flow, component width is often a function of its content or layout container, while the height defines its hit target or the number of rows that are visible.*

### Maintaining logic and consistency in component sizing.

The design unit is the number of pixels that form the basis of design decisions made in regards to sizing and layout of components and elements.

In Fluent Design, the design unit is 4 pixels.

![The 4px design unit](https://user-images.githubusercontent.com/47367562/148861614-588fb7c7-1402-447a-9fb7-7e7f2a5827d6.png)

### Describing base density.

The design unit is multiplied with a base value to calculate measurements applied to aspects of interface elements.

For instance, the spacing on the sides of a button, between text and icons, or between buttons are all expressed as a multiple of the design unit.

![A normal density button](https://user-images.githubusercontent.com/47367562/148861690-40fa3b7f-e53d-48d8-b28e-6e6825a54d67.png)

- 2 units vertical outer spacing, applied top and bottom
- 3 units horizontal outer spacing, applied left and right
- 2 units horizontal spacing between internal elements

### Adjusting density.

With the relationships applied to components and layouts, density adjustments are applied to scale the base multipliers.

[...]

PR #5511, opened 2022-01-11
New adaptive density model

bheston:
New adaptive density model
- Added density design tokens
- Updated Button and Text Field as illustration
- Added example Storybook page

# Pull Request

## 📖 Description

An updated adaptive density model. Please see #5510 for full details.

Here's what the test page looks like:

density test page

### 🎫 Issues

This draft accompanies #5510

## 👩‍💻 Reviewer Notes

Updated density model illustrated on Button and Text Field. Would be implemented across the components possibly with visual update.

## 📑 Test Plan

Per visual update and comments from associated RFC.

### General

## ⏭ Next Steps

Address comments on RFC and implement for remaining components.


packages/web-components/fast-components/src/design-system-provider/index.ts
```diff
@@ -227,6 +231,38 @@ export class DesignSystemProvider extends FoundationElement {
...
+    /**
+     *
+     * The density offset, used with designUnit to calculate height and spacing.
+     *
+     * @remarks
+     * HTML attribute: density-component-vertical-adjustment-units
+     *
+     * CSS custom property: --density-component-vertical-adjustment-units
+     */
+    @attr({
```
bheston, line 243: These two attributes were added for example and for testing. To follow the pattern all other density attributes would be added here as well.

packages/web-components/fast-components/src/number-field/number-field.styles.ts
```diff
@@ -44,6 +45,7 @@ export const numberFieldStyles = (context, definition) =>
         .root {
             display: flex;
             flex-direction: row;
+            gap: calc(${densityComponentHorizontalBetweenPadding} * 1px);
```
bheston, line 48: If we can make some changes so `gap` becomes meaningful, this would likely be a common pattern of setting `display: flex` and adding the correct `gap` for either `row` (horizontal) or `column` (vertical) layout.

packages/web-components/fast-components/src/styles/density.ts
```diff
@@ -0,0 +1,249 @@
...
+ * @remarks
+ * A variant on the type ramp for the common usage of a single line of text as in most components, adjusted for visual size rather than leading between multiple lines.
+ *
+ * For text-only and icon-only buttons to be the same height, this value should be the same as the icon size.
+ *
+ * @defaultValue 16px
+ *
+ * @public
+ */
+export const typeRampBaseVisualHeight = create<string>(
```
bheston, line 18: I considered other options for this, but I think however this ends up working there does need to be separate values for handling single vs multiple lines of text. I think the stronger choice is actually something like:

single visual line height (this value) + leading (new property) = multi-line height (existing property)

This is more classically the way text layout is thought of and opens up increased possibilities including https://medium.com/microsoft-design/leading-trim-the-future-of-digital-typesetting-d082d84b202 and https://seek-oss.github.io/capsize/. See the RFC for why text is so hard to size correctly.

packages/web-components/fast-components/src/styles/density.ts
```diff
@@ -0,0 +1,249 @@
...
+ * @remarks
+ * To use relative density do not change this value, only consume it in styles.
+ * If you want to break the chain of relative density, override this default function with a fixed value.
+ *
+ * @public
+ */
+export const densityComponentHorizontalAdjustmentUnitsCumulative = create<number>(
+    "density-component-horizontal-adjustment-units-cumulative"
+);
+densityComponentHorizontalAdjustmentUnitsCumulative.withDefault(
```
bheston, line 81: Nick and I chatted about the idea of adding native support for more complicated design tokens like this in the underlying infrastructure. The token infrastructure is designed to get around limitations of plain css properties, and one of the limitations is no way to access parent properties. I've done that here, but it's somewhat complicated and requires a shadow token `Cumulative` rather than a behavior or mode of the single `Units` token.

On the other hand, this structure does easily allow for `Cumulative` to default to being relative to the parent, or redefining this derived token to be a fixed value. In that usage "Cumulative" is probably not the correct word, so I'm open to other suggestions that mean "this is the actual value you should use within the component", whether that implies a relative or fixed value.

packages/web-components/fast-components/src/styles/density.ts
```diff
@@ -0,0 +1,249 @@
...
+ * Calculated from `densityComponentHorizontalOuterValue` - `strokeWidth`.
+ *
+ * @public
+ */
+export const densityComponentHorizontalOuterPadding = create<number>(
+    "density-component-horizontal-outer-padding"
+).withDefault(
+    (element: HTMLElement) =>
+        densityComponentHorizontalOuterValue.getValueFor(element) -
+        strokeWidth.getValueFor(element)
```
bheston, line 122: I went back and forth between using `strokeWidth` or a new density-specific value I had added. The problem is that the tokens generally apply at a component level, but the border stroke or padding may be applied in css at any of the child elements. Nested layout scenarios may be difficult to describe in this model, say where a border is applied to a container div, and padding isn't applied until a few levels deeper in the dom, or when there may be placeholders for invisible borders or something.

What I was really looking for was a way to set a design token value for a specific element within the styles. Like this `.control` element should have a border for the sake of density, and then I'm also going to apply density padding here.

I had a helper function that applied a property name that was _not_ a full design token, then that property was referenced in these "outer" density values. At that point they had to be `cssPartial`s though, which don't notify on update. That was causing other problems so I went back to the more traditional model.

I do think it would be useful to be able to easily scope a design token value to a particular selector within the styles, but I don't have a great example for it right now. I was considering a new Behavior to set a token value, perhaps with an optional selector to target a child element, but that starts to break the DesignToken update model as well unless the token value can be set in code rather than css.

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -58,9 +60,7 @@ export const BaseButtonStyles = (
             font-family: ${bodyFont};
             outline: none;
             font-size: ${typeRampBaseFontSize};
-            line-height: ${typeRampBaseLineHeight};
-            height: calc(${heightNumber} * 1px);
-            min-width: calc(${heightNumber} * 1px);
+            line-height: ${typeRampBaseVisualHeight};
```
bheston, line 63: Related to the comment above about single line height + leading. Since single line is most common, it may be easier to repurpose what line height means in the token and to have a derived multi-line height for the fewer cases that actually have multiple lines of text.

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -90,7 +91,9 @@ export const BaseButtonStyles = (
             font: inherit;
         }
         .control.icon-only {
-            padding: 0;
+            ${densityComponentHorizontalOuterPadding.cssCustomProperty}: var(${
```
bheston, line 94: One option for making something square. This assumes the vertical measurement should stay consistent, which _implies_ that we think the horizontal measurement probably is different. This works with most buttons I've seen, but may not be a universal application.

Related to the previous comment about not being able to set token values for a particular selector, the more semantically correct alternative would be to set the horizontal `Units` to the same value as vertical and let the `Padding` be calculated.

packages/web-components/fast-components/src/styles/patterns/button.ts
```diff
@@ -118,12 +121,6 @@ export const BaseButtonStyles = (
             display: flex;
             pointer-events: none;
         }
-        .start {
```
bheston, line 121: Replaced with `gap`, which is closer to the intent and safer than needing to style `start` and `end` independently everywhere.

packages/web-components/fast-foundation/src/patterns/start-end.ts
```diff
@@ -59,15 +56,15 @@ export const endSlotTemplate: (
     context: ElementDefinitionContext,
     definition: EndOptions
 ) => html`
-    <span
+    <slot
```
bheston, line 59: I realize this is potentially a large stylistic breaking change. I asked in Discord but didn't get any response about the history of this structure. The issue is that the `span` always *exists* and therefore doesn't work with `flex` layout and `gap` because there is always space between the `span`s even if the `slot` is empty. With only the `slot` if it's empty it's treated correctly in layout and the `gap` is not added.

This seems semantically preferable to me, and seems to work with the existing events and notifications we have setup.

chrisdholt: TLDR - one PR one change is preferable to avoid scenarios like this where an entire change is held up by something unrelated. If it _is_ related, we'll want to look and ensure *every* API continues to work as expected. I'm not sure that's the case unless we actually test for it. Making a break like this and assuming it works doesn't give me much confidence. Can we check that everything *does* work and then we can look at the implications of making a breaking change to a template used by most controls?

packages/web-components/fast-components/src/text-field/text-field.styles.ts
```diff
@@ -38,20 +40,21 @@ export const textFieldStyles = (context, definition) =>
         ${display("inline-block")}
         ${inputStyles(context, definition, ".root")}
         ${inputStateStyles(context, definition, ".root")}
+
         .root {
             display: flex;
             flex-direction: row;
+            gap: calc(${densityComponentHorizontalBetweenPadding} * 1px);
```
chrisdholt, line 47: This could bite us in terms of browser versions supported. Many of the things we are leveraging in terms of modern browsers either have fallback support where not implemented yet (adopted stylesheets) or in the case of things like CSS grid were more robustly included in earlier versions than this. This moves the needle further back and could create issues with commitments we've made. Probably something to discuss.

KingOfTac: Would it be worth it to use gap for browsers that support it and add a matchMedia behavior as a stopgap measure until wider support is available?

This way it would end up being less of a refactor when gap is better supported.

KingOfTac: It actually looks like gap is just about there on support.

https://developer.mozilla.org/en-US/docs/Web/CSS/gap#browser_compatibility

packages/web-components/fast-components/src/styles/density.ts
```diff
@@ -0,0 +1,249 @@
...
+);
+
+/**
+ * The inverse of the outer padding, used to pull child elements out into density padding.
+ *
+ * @remarks
+ * This is a convenience token to apply negative margins to a child element to pull it out over the container's padding.
```
chrisdholt, line 129: Can you help me understand this scenario? Negative margins are typically a code smell. What are we trying to accomplish with negative margin here?

packages/web-components/fast-components/src/styles/density.ts
```diff
@@ -0,0 +1,249 @@
...
+ * @public
+ */
+export const typeRampBaseVisualHeight = create<string>(
+    "type-ramp-base-visual-height"
+).withDefault("16px");
+
+// Component horizontal tokens
+
+/**
+ * The design unit multiplier for component density horizontal outer padding.
```
chrisdholt, line 25: Is this representative of space *outside* the component? If so, I have significant concerns around an approach that specifies anything relating to the positioning or spacing between controls. As a best practice, primitives should make no assumptions about their layout or relationship to other elements. That should be applied contextually.

If this isn't related to things outside the component, we should probably work on the semantics. These are quite a mouthful :)

chrisdholt:
@bheston - closing this for housekeeping as `fast-components` has been removed in favor of the CLI - but we should look to reuse this for the adaptive-ui package.

Issue #5484, opened 2021-12-18
feat: memoize the React wrapper


EisenbergEffect:
# 🙋 Feature Request

## 🤔 Expected Behavior

Passing the same web component or registry to the `wrap` function should return the same React component

## 😯 Current Behavior

A new React component is created every time the `wrap` function is called even if it's called with the same component or registry as previously.

## 💁 Possible Solution

We can have a simple cache lookup based on the passed component.

## 🔦 Context

We've seen some patterns in the community where folks were creating wrappers repeatedly without understanding the implications of that. We can prevent the unintentional error with this simple fix.

## 💻 Examples

```ts
const MyReactComponent = wrap(MyComponent);
const MySameReactComponent = wrap(MyComponent);

assert(MyReactComponent === MySameReactComponent);
```

m4thieulavoie:
@EisenbergEffect is this up for grabs? I'd be happy to look into it :+1:

EisenbergEffect:
@m4thieulavoie Absolutely. Go for it! Thank you 😄

PR #5516, opened 2022-01-12
add a caching system for react wrappers

m4thieulavoie:
# Pull Request

## 📖 Description

This PR adds a simple caching system in the React wrappers. It adds a lookup and returns an existing wrapper if the consumer already previously requested for the same component earlier. That should help prevent multiple expensive `React.forwardRef`s
Fixes https://github.com/microsoft/fast/issues/5484

### 🎫 Issues

* https://github.com/microsoft/fast/issues/5484

### General

### Component-specific


EisenbergEffect:
I think the solution is a bit more trick. I think we want to move the cache check up earlier in the `wrap` function so that we don't even define the class if we've already got one in the cache. But it's complicated by the fact that the "type" that is passed in could be a registry instead. In that case, we need to check the type of that registry to see if it's in the cache as well and not add it to the registry if it is too.

EisenbergEffect (changes requested):
I think the solution is a bit more tricky. I think we want to move the cache check up earlier in the wrap function so that we don't even define the class if we've already got one in the cache. But it's complicated by the fact that the "type" that is passed in could be a registry instead. In that case, we need to check the type of that registry to see if it's in the cache as well and not add it to the registry if it is too.

m4thieulavoie:
> I think the solution is a bit more tricky. I think we want to move the cache check up earlier in the wrap function so that we don't even define the class if we've already got one in the cache. But it's complicated by the fact that the "type" that is passed in could be a registry instead. In that case, we need to check the type of that registry to see if it's in the cache as well and not add it to the registry if it is too.

If I capture properly what you said, [my last commit](https://github.com/microsoft/fast/pull/5516/commits/cd04320db8fe17f789b69d4de994ab81b61c2279) should cover it. However I have the impression I'm missing a key piece of your comment 😅 can you please look at it again?

packages/utilities/fast-react-wrapper/src/index.ts
```diff
@@ -199,6 +199,7 @@ export function provideReactWrapper(React: any, designSystem?: DesignSystem) {
             registrations = [];
         },
     };
+    const wrappersCache = new Map();
```
EisenbergEffect, line 202: I think this needs to be moved up out of the provider function so that these wrappers can be shared across calls to the provider. Basically, if the same type is passed, we want to get back the same React component, even if it's across provider calls.

packages/utilities/fast-react-wrapper/src/index.ts
```diff
@@ -242,109 +243,114 @@ export function provideReactWrapper(React: any, designSystem?: DesignSystem) {
             type = type.type as any;
         }
 
-        class ReactComponent extends React.Component<InternalProps> {
-            private _element: TElement | null = null;
-            private _elementProps!: { [index: string]: unknown };
-            private _userRef?: ReactModule.Ref<unknown>;
-            private _ref?: (element: TElement | null) => void;
+        if (!wrappersCache.has(type)) {
```
EisenbergEffect, line 246: I'm not sure if it's that much more efficient or not, but you could also just do a `get` here and if it is not undefined, just return the result, otherwise create the wrapper and `set` it. This would eliminate needing to have the `has` call.

EisenbergEffect (changes requested):

packages/utilities/fast-react-wrapper/src/index.ts
m4thieulavoie: Rightttt I didn't push my reflexion all the way to multiple providers, that makes sense!

packages/utilities/fast-react-wrapper/src/index.ts
m4thieulavoie: Love it, way cleaner I think!

EisenbergEffect (approved):

nicholasrice (approved):
Looks good to me!

EisenbergEffect:
Merged! Thanks @m4thieulavoie !

Issue #4346, opened 2021-02-05
build tools: test steps cause redundant builds


radium-v:
These packages include `yarn build` as part of their `test` script, which creates unnecessary overhead and time-consuming workflow redundancies:
- `fast-figma-plugin-msft`
- `fast-tooling`
- `fast-tooling-react`
- `fast-animation`
- `fast-color-explorer`
- `fast-component-explorer`
- `fast-creator`
- `fast-tooling-examples`
- `fast-website`
- `site-utilities`

Projects `fast-tooling`, `fast-tooling-react` and `fast-animation` should run their test-step-build as `tsc --noEmit`, though any performance differences are trivial for these packages.

The remaining packages run full builds (usually `webpack`) in their test stages. There are a few options when it comes to addressing this:

- **Skip the test step for packages that don't have tests**: Skipping building entirely is not a viable option, since the underlying problem is that the projects without tests (the sites) are mostly all consumers of the other packages. These projects still need to be built _at some point_ to validate that straightforward build errors don't occur.
- **Add `prepare` steps to packages without tests that need to build**: This makes the most _technical_ sense since they're dependent on the builds from the workspace, but it's extremely time-consuming for everyone to sit through full production builds of all of the sites during our already very long `prepare` step.
- **Add real test steps to all packages**: This is probably the most time-consuming option, and since there's not much in terms of automated options yet, this probably isn't the right solution (playwright would work, but it would be a lot of work).
- **Specify an extended/alternative workflow**: This would work for developers so that we don't have to build everything all at once every time, and would allow for more granular control when running automated workflows.
- change lower level test to "test": "lerna run test",

Ibrahimmaga:
I believe the work recently done by Jane on testing is sufficient for all testing on fast tooling.  Also,  all test are being done line by line. we are not skipping anything. It's time consuming, but it's better to catch the error early than doing damage control. 
I will ping Jane to see if she has something to add.

Ibrahimmaga:
conversation to solve scripts:
- change lower level test to "test": "lerna run test",
- compare all build, test, prepare for all packages 
- make sure they are not duplicating command
- make sure developer workflow is not impacted

awentzel:
This issue should be automatically resolved by completing the webpack upgrade.  Take no action on this PR until final implementation upgrade is complete.

Ibrahimmaga:
[...]

PR #5519, opened 2022-01-12
simplify platforms and browsers workflow

radium-v:
# Pull Request

## 📖 Description

Reduces duplication in the "Validate Platforms/Browsers" workflow.

### 🎫 Issues

Related to #4346.

## 👩‍💻 Reviewer Notes

For pull requests, Lerna will now only run `test` in packages which have changed, and their dependents. For example, a change in `fast-foundation` would cause `fast-website`, `site-utilities`, `fast-components`, and `fast-react-wrapper` to all run their tests. However, if a PR only contains changes to `fast-react-wrapper`, then that would be the only package to be tested since no other packages list it as a dependency. This behavior is only enabled for PRs, and all packages are still tested on `push` to `master`.

### General

## ⏭ Next Steps

Since multiple workflows run the `yarn prepare` and `yarn lerna run test --stream` steps, we should find a way to do this as few times as possible for each PR.


.github/workflows/ci-validate-platforms.yml
```diff
@@ -3,224 +3,109 @@ name: Validate Platforms/Browsers
...
-      uses: github/codeql-action/analyze@v1
-
-  windows_cross-platform_cross-browser:
-    runs-on: [windows-latest]
-    env:
-      PLAYWRIGHT_BROWSERS_PATH: 0
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-11]
```
awentzel, line 22: Do you recall why we're running `macos-11` and not `macos-latest` which would be preferred?

.github/workflows/ci-validate-platforms.yml
```diff
@@ -3,224 +3,109 @@ name: Validate Platforms/Browsers
...
+      - name: Install playwright dependencies and browsers
+        run: |
+          yarn playwright install-deps
+          yarn playwright install
+
+      - name: Build Playwright tests
+        run: yarn workspace @microsoft/fast-components build:pw
+
+      - name: Test Playwright (chromium)
+        run: yarn workspace @microsoft/fast-components test-pw
```
awentzel, line 83: What is `test-pw`?  If related to a password where is it stored?  If not, let's spell out rather than abbreviate for readability and understandability.

.github/workflows/ci-validate-platforms.yml
```diff
@@ -3,224 +3,109 @@ name: Validate Platforms/Browsers
...
+        continue-on-error: true
+        timeout-minutes: 2
+        env:
+          PLAYWRIGHT_BROWSER: chromium
+
+      - name: Test Playwright (firefox)
```
awentzel, line 89: Since we're refactoring a bit here, I probably would separate out each testing browser engine type as their own jobs dependant on a building job.  This will paralyze the testing task, reduce time, and improve the visual graph represented in GitHub Actions.

.github/workflows/ci-validate-platforms.yml
```diff
@@ -3,224 +3,109 @@ name: Validate Platforms/Browsers
...
+          PLAYWRIGHT_BROWSER: firefox
+
+      - name: Test Playwright (webkit)
+        run: yarn workspace @microsoft/fast-components test-pw
+        continue-on-error: true
+        timeout-minutes: 2
+        env:
+          PLAYWRIGHT_BROWSER: webkit
+
+      - name: Initialize CodeQL
```
awentzel, line 103: CodeQL and Analysis could also be broken out into a security test job.  Breaking into separate jobs may require an upload to cloud of the built bits and then download into those jobs so you don't have to rerun build after the initial time.

awentzel (approved):
Approving with comments and hope you can enhance further with my recommendations.

.github/workflows/ci-validate-platforms.yml
radium-v: The scripts in `package.json` (`build:pw, test-chrome:pw, test-pw` etc.), as well as the playwright-specific spec files (`*.pw.spec.ts`) are all referring to Playwright. This was done to differentiate them from the test suite files and scripts that were already present when Playwright was first added to `@microsoft/fast-components`.

.github/workflows/ci-validate-platforms.yml
radium-v: Ideally we only build once, and all other jobs would use those build artifacts. I'll make a follow-up issue.

chrisdholt (approved):

PR #5528, opened 2022-01-19
add findLastIndex and inRange utility functions to fast-web-utilities

radium-v:
# Pull Request

## 📖 Description

* Adds a `findLastIndex` function to `@microsoft/fast-web-utilities`, which returns the index of the last item in an array which causes a predicate function to return `true`. Otherwise, it returns `false`. This is a simplified inversion of the built-in `Array.prototype.findIndex` function.
* Adds a `inRange` function to `@microsoft/fast-web-utilities`, which returns `true` or `false` depending on if the `value` is within the range of the `min` and `max`. If `max` is not provided, the range is between `min` and `0` (`min` and `max` are sorted by value before the test is executed). This is a simplified adaptation of Lodash's `_.inRange` function.

## 👩‍💻 Reviewer Notes

These utility functions are needed for the upcoming listbox multiselect features.

## 📑 Test Plan

All tests should pass. `inRange` and `findLastIndex` will be used in an upcoming pull request.

### General


EisenbergEffect (approved):

packages/utilities/fast-web-utilities/src/array.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    let k = array.length;
+    while (k--) {
+        if (predicate(array[k], k, array)) {
+            return k;
+        }
+    }
+
+    return -1;
```
robarbms, line 22: You'll get better performance using [].reduce.

```suggestion
    return array.reduce((found, value, index, items) => predicate(value, index, items) ? index : found, -1);
```

packages/utilities/fast-web-utilities/src/array.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param array - the array to test
+ * @param predicate - find calls predicate once for each element of the array, in descending order, until it finds one where predicate returns true. If such an element is found, findLastIndex immediately returns that element index. Otherwise, findIndex returns -1.
+ */
+export function findLastIndex<T>(
+    array: Array<T>,
+    predicate: (value: T, index: number, obj: T[]) => unknown
+): number {
+    if (!(predicate instanceof Function)) {
+        throw new TypeError();
+    }
```
robarbms, line 13: Is this necessary with TypeScript?

packages/utilities/fast-web-utilities/src/array.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Returns the index of the last element in the array where predicate is true, and -1 otherwise.
+ *
+ * @param array - the array to test
+ * @param predicate - find calls predicate once for each element of the array, in descending order, until it finds one where predicate returns true. If such an element is found, findLastIndex immediately returns that element index. Otherwise, findIndex returns -1.
+ */
+export function findLastIndex<T>(
+    array: Array<T>,
+    predicate: (value: T, index: number, obj: T[]) => unknown
```
robarbms, line 9:
```suggestion
    predicate: (value: T, index?: number, obj?: T[]) => unknown
```

packages/utilities/fast-web-utilities/src/numbers.ts
```diff
@@ -20,3 +20,15 @@ export function wrapInBounds(min: number, max: number, value: number): number {
...
+/**
+ * Determines if a number value is within a specified range.
+ *
+ * @param value - the value to check
+ * @param min - the range start
+ * @param max - the range end
+ */
+export function inRange(value: number, min: number, max: number = 0): boolean {
+    [min, max] = [min, max].sort((a, b) => a - b);
+    return min <= value && value < max;
```
robarbms, line 33: Sorting is expensive. Have you considered using the limit function to verify that the value is in range?

```suggestion
    return limit(Math.min(min, max), Math.max(min, max), value) === value;
```

robarbms (approved):

packages/utilities/fast-web-utilities/src/array.ts
radium-v: A `while` loop is much faster than a `reduce` here since the `return` will short-circuit as soon as a match is found.

packages/utilities/fast-web-utilities/src/array.ts
radium-v: This matches the behavior of the built-in `Array.prototype.findIndex` function.

packages/utilities/fast-web-utilities/src/array.ts
radium-v: This matches the signature of the built-in `Array.prototype.findIndex` function.

packages/utilities/fast-web-utilities/src/numbers.ts
radium-v: This fails the test "should return `false` when `value` is equal to `max`".

robarbms: So the logic is that it can be equal or greater than min or less than max? Shouldn't min and max logic be handled the same? It seems a little confusing this way.

radium-v: My thinking was to match Lodash's `inRange` behavior, which handles `min` and `max` this way:

> Checks if `n` is between `start` and up to, but not including, `end`. If `end` is not specified, it's set to `start` with `start` then set to `0`. If `start` is greater than `end` the params are swapped to support negative ranges.

scomea (approved):

packages/utilities/fast-web-utilities/src/array.ts
radium-v: I flipped back and forth on making the params optional, but after testing I realized they need to be there to avoid errors when `strictNullChecks` is enabled in #5378

chrisdholt (approved):

